PR #6, opened 2019-07-08
Feat/dotcom react

kennylam:
Adds `jest` testing and React component stubs to project.

#### Changelog

**New**

- `jest` packages and dependencies, coverage, and scripts kennylam@b6adbb36efc38b6ff1deb9702bd515f109d07a26
- adds `Masthead` component stub for React components kennylam@3cd2b5b17d32edcb93c8b6c25909c75173dec84a

**Changed**

- update `.gitignore`


jeffchew:
@kennylam should we remove package-lock.json if we're sticking with yarn?

Also, I know it's not part of this PR but we should add the namespace to the package names right?

I'm looking at eslint and looks like we are forking it from their repo as we talked about. Can we add in the require-jsdoc and valid-jsdoc to that config?

packages/react/scripts/env.js
```diff
@@ -0,0 +1,17 @@
...
+const BABEL_ENV = process.env.BABEL_ENV;
+
+module.exports = () => ({
+  presets: [
+    [
+      require.resolve('@babel/preset-env'),
+      {
+        modules: BABEL_ENV === 'es' ? false : 'commonjs',
+        targets: {
+          browsers: ['extends browserslist-config-carbon'],
```
jeffchew: This looks like an unpublished carbon package, right?

packages/react/src/components/Masthead/Masthead-test.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * Copyright IBM Corp. 2016, 2018
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import Masthead from './Masthead';
+import { shallow } from 'enzyme';
```
jeffchew: Probably not important for this commit, but wondering if we should use storyshots to rely more on snapshot testing over enzyme?

package-lock.json
```diff
@@ -126,6 +126,152 @@
         "trim-right": "^1.0.1"
       }
     },
+    "@babel/helper-annotate-as-pure": {
```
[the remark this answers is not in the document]
kennylam: Yep, removed.

packages/react/scripts/env.js
kennylam: Yes, there are several unpublished packages in the project, leveraging Yarn workspaces.

packages/react/src/components/Masthead/Masthead-test.js
kennylam: Sure we could set that up later. This is just a component stub required for the project to build.

PR #30, opened 2019-07-29
feat(footer): laying down the basework for the footer

photodow:
Starts [digital-design/#1080](https://github.ibm.com/webstandards/digital-design/issues/1080)

This pull request basically initializes the footer component. Since this is my first, please take a look and make sure I did it right before I get too far along. I used the masthead component as a starting place.

#### Changelog

**New**

- initializes new footer component


jeffchew:
@photodow you can set this as a draft pull request for now if this isn't ready to be merged, but thank you for setting this up early.

packages/react/src/components/__tests__/__snapshots__/storyshots.test.js.snap
```diff
@@ -1,5 +1,274 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`Storyshots Footer Default 1`] = `
```
photodow: @kennylam @jeffchew What is this file? I didn't create this file. Should we be tracking it?

jeffchew: @photodow Yes, this is a snapshot file that is autocreated from snapshot testing. Because it's a new story, it generates the snapshot and adds to this file. 

Unit tests would essentially fail if snapshots don't match up between the executed test and the saved snapshot file. You can update the snapshot file by running `test:updateSnapshot` if you know that the snapshot file needs to be updated.

cc: @kennylam @vlad-saling

jeffchew: @photodow more info on snapshot testing in Jest: https://jestjs.io/docs/en/snapshot-testing

PR #51, opened 2019-08-16
Footer

photodow:
### Related Ticket(s)

* [webstandards/digital-design#1368](https://github.ibm.com/webstandards/digital-design/issues/1368)
* [webstandards/digital-design#1080](https://github.ibm.com/webstandards/digital-design/issues/1080)

### Description

The primary purpose of this pull request is to introduce the MVP footer component (not including feedback survey, locale selector, or translations). It includes a new Footer component with two variations, short and tall. Tall is default and does not need to be specified if desired. In addition this pull request also adjusts a couple things in hopes to help improve local development.

### Changelog

**New**

- Footer component
  - Default / Tall
  - Short

**Changed**

- Removed some duplicates from the package dependencies
- Added parent `node_modules` into storybook webpack build
- Configured storybook to allow for full width components


packages/react/src/components/Footer/__stories__/Footer.stories.js
```diff
@@ -2,4 +2,9 @@ import React from 'react';
 import { storiesOf } from '@storybook/react';
 import Footer from '../Footer';
 
-storiesOf('Footer', module).add('Default', () => <Footer />);
+// import './footer.scss';
+import '../../../../../styles/scss/components/footer/_index.scss';
```
kennylam: @photodow You should be able to avoid having to do this by importing your SCSS into a component
```suggestion
import '@ibmdotcom/styles/scss/components/masthead/_masthead.scss';
```

photodow: @kennylam To the component, or the story where it is currently? This requires link `@ibmdotcom/styles`, correct?

packages/react/src/components/Footer/Footer.js
```diff
@@ -17,15 +30,56 @@ class Footer extends React.Component {
...
   render() {
+    const { type } = this.props;
+
     return (
-      <footer>
-        Lorem ipsum dolor sit amet, consectetur adipiscing elit.
-        {localeSelector ? (
-          <LocaleSelector text="Load the Locale Selector!" />
-        ) : null}
+      <footer
```
jeffchew: @photodow can you add in stable selectors here?
https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/docs/stable-selectors.md

packages/react/src/components/Footer/FooterNav.js
```diff
@@ -0,0 +1,100 @@
...
+  /**
+   * Renders the footer nav component
+   *
+   * @returns {object} JSX object
+   */
+  render() {
+    const { groups } = this.props;
+
+    return (
+      <nav className={`${prefix}--footer-nav`}>
```
jeffchew: @photodow can you add in stable selectors here? 
https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/docs/stable-selectors.md

packages/react/src/components/Footer/Footer.js
```diff
@@ -6,8 +6,21 @@
...
+
+import footerMenuData from './__stories__/data/footer-menu';
+import footerLegalData from './__stories__/data/footer-legal';
+
+console.log(footerLegalData);
+
+const { prefix } = settings;
 
 /** Footer component */
 class Footer extends React.Component {
```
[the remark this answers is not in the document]
photodow: Thanks for these comments. It's been a while since I was in React last, and forgot there was a difference. The others we might be able to get away with (as long as children can still have state because the accordion from requires state I believe), but this one it really depends on how our adopters need it. If we are going to just allow adopters pass in whatever, and the with the UI Shell pass in the correct data then that's great. I was assuming the footer would make handle state and make the API call to translation by itself so we get the same data every time and not open that door for the incorrect formation of navigation items. In fact I was working on making that happen now, but maybe I'll pause before I get too far off track.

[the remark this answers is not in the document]
jeffchew: @photodow check out how I handle this in MastheadSearch. I keep it as a stateless component, but use useReducer for handling the API changes with autosuggest. 

cc: @kennylam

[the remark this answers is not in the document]
photodow: I'll check it out.

packages/react/src/components/Footer/__stories__/Footer.stories.js
kennylam: @photodow yeah if you link the packages you can import the css into the component with just a `@ibmdotcom/../../`. Then you're set for use in storybook as well.

jeffchew:
@oliviaflory please also do a design review of this when you get a chance:
https://deploy-preview-51--ibmdotcom-react.netlify.com/

Will ping Wonil and Lara once they have accepted their invites to the Github team.

photodow:
@jeffchew I reviewed it with Wonil before he left, but @oliviaflory & lara definitely don't hesitate to provide feedback. I can also jump on a call and iron out any details real quick if needed too.

packages/react/.storybook/Container.js
```diff
@@ -24,11 +24,8 @@ export default class Container extends Component {
         <div
           data-floating-menu-container=""
           role="main"
+          className="storybook-container"
```
annawen1: Noticed you put the class name as `.storybook-center-container` in the corresponding scss file?

photodow: Great catch, this was leftover from trying to figure out a way to flex center smaller components while allowing full width for something like this footer. I'll remove this, though. thanks.

packages/react/src/components/Footer/__stories__/Footer.stories.js
photodow: So I think I updated it correctly. Let me know if I misunderstood. On another note, this requires extra documentation. As dirty as the original was (I think we could clean it up a little more) it allows contributors to get going without having to link everything up, but this is good for now. It's not like we have a hundred contributors or anything right now. :P

packages/react/src/components/Footer/LegalNav.js
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Loops through and renders list items for legal nav
+ *
+ * @param {Array} links A list of links to be rendered
+ * @returns {object} JSX object
+ */
+function renderListItems(links) {
+  const toRender = [];
+
+  for (let i = 0; i < links.length; i += 1) {
```
jeffchew: Minor thing and general question to all, wondering if we are ok with going with either a for loop, or perhaps something more terse like map:

```
{links.map((item, index) => (
  <li className={`${prefix}--legal-nav__list-item`}>
    <Link data-autoid="footer-legal-nav__link" href={item.url}>
      {item.title}
    </Link>
  </li>
))}
```

@kennylam @photodow @annawen1 

I wouldn't even mind this within the actual render, or having as a separate function works fine as well. Thoughts?

photodow: > Minor thing and general question to all, wondering if we are ok with going with either a for loop, or perhaps something more terse like map:

@jeffchew here's my thinking...

`for` loops are hands down more performant, and probably the most versatile. I've been called out on this before for micro performance/premature optimization, and I kind of disagree with that. I feel like you should be building out the cleanest and most performant codebase from the beginning so you don't have to go back and refactor a lot of things later. If you already have maps in there I don't think there is much to gain by replacing them all with for loops unless you know that's a bottleneck to your code, but I see no problem starting with them.

With that said I get `map()` and some of the other loop methods are easier to read, and specific to certain tasks. If that's the direction we want to go I'm cool with that, but to me, the UX and performance that impacts the user, in my opinion, trumps our coding processes/styles in the end.

> I wouldn't even mind this within the actual render or having as a separate function works fine as well. Thoughts?

If you're referring to including it in the JSX render, I'm personally not a fan because it really muddies the JSX up. I don't want to say there is never a scenario where that doesn't make sense, but as a general rule, I personally try to keep logic out of the JSX as much as possible.

jeffchew: Here's a pretty detailed article on `map` vs `for`, and I'm a bit in favor of `map`. The article even touches on the performance part of `for`, but seems like it makes more sense and see a difference for super large arrays (which we're not really processing here).

I agree with a separate method for rendering the iteration though.

Thoughts @kennylam @annawen1 ?

photodow: Yeah, I've seen that article. It's clearly biased. 😜 haha

I don't agree with everything in it, but I think he does bring out some good points on both sides though. I do agree, the difference is definitely shown in larger arrays, and to be fair the gap in performance has been decreasing in browsers.

> However, this is more preference and working in a standard than a clear-cut victory for one or the other.

I think he's right about this, and when it boils down to preference I take the user's side every time. 🤷‍♂️

photodow: @jeffchew campaigned offline in New York and got the votes for `map()`. So I'll make the updates in this pull request.

packages/react/src/components/Footer/Footer.js
```diff
@@ -6,26 +6,75 @@
...
-        Lorem ipsum dolor sit amet, consectetur adipiscing elit.
-        {localeSelector ? (
-          <LocaleSelector text="Load the Locale Selector!" />
-        ) : null}
-      </footer>
-    );
+import PropTypes from 'prop-types';
+// import { localeSelector } from '../../internal/FeatureFlags';
+// import LocaleSelector from './LocaleSelector';
+import { settings } from 'carbon-components';
```
jeffchew: @photodow this import should point to the es folder so we don't pull in the whole package. I think this is what's causing our artifact to balloon up to be huge. I already asked @kennylam to do the same for the Masthead.
cc: @annawen1

packages/react/src/components/Footer/FooterNav.js
```diff
@@ -0,0 +1,96 @@
...
+ * Copyright IBM Corp. 2016, 2018
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import { settings } from 'carbon-components';
+import { Accordion } from 'carbon-components-react';
```
jeffchew: @photodow change these imports to point to the es folder.
cc: @kennylam @annawen1

packages/react/src/components/Footer/FooterNavGroup.js
```diff
@@ -0,0 +1,79 @@
+/**
+ * Copyright IBM Corp. 2016, 2018
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import { settings } from 'carbon-components';
```
jeffchew: @photodow change these imports to point to the es folder.
cc: @kennylam @annawen1

packages/react/src/components/Footer/LegalNav.js
```diff
@@ -0,0 +1,68 @@
...
+ * Copyright IBM Corp. 2016, 2018
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import { settings } from 'carbon-components';
+import { Link } from 'carbon-components-react';
```
jeffchew: @photodow change these imports to point to the es folder.
cc: @kennylam @annawen1

packages/react/src/components/Footer/README.md
```diff
@@ -0,0 +1,56 @@
...
+> displays consistently at the bottom of each page.
+
+## Getting started
+
+Here's a quick example to get you started.
+
+```javascript
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { Footer } from '@ibmdotcom/react';
```
jeffchew: @photodow we should probably instruct users to point to the es folder if they want to take advantage of tree-shaking. I provided both as an example in the JSDocs for services and utilities:
https://deploy-preview-55--ibmdotcom-services.netlify.com/searchtypeaheadapi

packages/react/src/components/Footer/Footer.js
photodow: like...

```javascript
import { settings } from 'carbon-components/es';
```

or are you asking for this...?

```javascript
```

jeffchew: @photodow the latter. @kennylam already tried this on his end and it drastically reduced the file size.

packages/react/src/components/Footer/__stories__/Footer.stories.js
```diff
@@ -1,5 +1,9 @@
 import React from 'react';
 import { storiesOf } from '@storybook/react';
-import Footer from '../Footer';
+import { Footer } from '@ibmdotcom/react/es/components/Footer';
```
jeffchew: @photodow for the stories, it should be ok to import from the relative path, since it isn't grabbing from an external es module. I think this is breaking the current netlify build actually.

photodow: Ah, just trying to be consistent. I'll try to get that in real quick.

packages/react/src/components/Footer/__stories__/Footer.stories.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import { storiesOf } from '@storybook/react';
-import { Footer } from '@ibmdotcom/react/es/components/Footer';
+import { Footer } from '@ibmdotcom/react/';
```
jeffchew: @photodow you missed the extra slash at the end, shouldn't be `@ibmdotcom/react/`

photodow:
🎉 VICTORY! 🎉

Still not seeing the big green mark, but at least I got past that Artifactory build issue... wow... I hate Artifactory so much.

PR #71, opened 2019-08-30
fix(profile): wrap profile service call in promise

annawen1:
### Related Ticket(s)

- [Profile menu](https://github.com/kennylam/ibm-dotcom-library/commit/388059ff8a0f13fc8e45e2a5f46d3ef672249070)
- [MyIBM profile background services](https://github.ibm.com/webstandards/digital-design/issues/778)

### Description

the `jsonp` npm package doesn't support promises so need to wrap it in a promise 
otherwise wrapping in the async/await function doesn't seem return the value in the promise

### Changelog

**Changed**

- wrap `jsonp` call in promise
- add promise.then() to the status in `Masthead` to set the state


annawen1:
will need to release an alpha release for this? but waiting for any comments before doing so

PR #72, opened 2019-08-30
Feat/dotcom shell

kennylam:
### Related Ticket(s)

https://github.ibm.com/webstandards/digital-design/issues/1456

### Description

Adds the IBM Dotcom shell template. Includes Masthead and Footer components and uses Carbon's grid.

### Changelog

**New**

- Dotcom shell [[ccd4c4b](https://github.com/carbon-design-system/ibm-dotcom-library/commit/ccd4c4bdd06e7a39b2c3af436c9175e5c6d3b45b)]


packages/react/src/components/DotcomShell/__stories__/DotcomShell.stories.js
```diff
@@ -0,0 +1,160 @@
...
+  Button,
+  DataTableSkeleton,
+  StructuredListWrapper,
+  StructuredListHead,
+  StructuredListBody,
+  StructuredListRow,
+  StructuredListCell,
+} from 'carbon-components-react';
+import mastheadLinks from '../../Masthead/__stories__/data/MastheadLinks.js';
+import '@ibmdotcom/styles/scss/components/dotcom-shell/_dotcom-shell.scss';
```
jeffchew: Only minor thing, I think for our stories we can do relative path to our styles package so we will get the styles that aren't pushed up to npm.

jeffchew:
@kennylam I'm going to merge in now, but looks like the left/right sides aren't extending to the edge of the browser (that's the desired behavior, right?). We can address in another PR.

Screen Shot 2019-08-30 at 2 00 09 PM

kennylam:
> I'm going to merge in now, but looks like the left/right sides aren't extending to the edge of the browser (that's the desired behavior, right?). We can address in another PR.

@jeffchew Yeah this is going to get fixed in the sticky Masthead story. I reverted to the incorrect 100% view because the sticky functionality is stlll in the works.

PR #73, opened 2019-09-03
Feat/1545 carbon namespace

jeffchew:
### Related Ticket(s)

https://zenhub.ibm.com/app/workspaces/digital-design-system-5b3a815a8bf6a932efa6fd03/issues/webstandards/digital-design/1545

### Description

This switches the package names from the @ibmdotcom namespace to the @carbon namespace:

- @ibmdotcom/react => @carbon/ibmdotcom-react
- @ibmdotcom/services => @carbon/ibmdotcom-services
- @ibmdotcom/styles => @carbon/ibmdotcom-styles
- @ibmdotcom/utilities => @carbon/ibmdotcom-utilities

In addition to the package name changes, have also added a new root level command:

`yarn reset`

This command will run the following which will perform an overall yarn reset of the installed packages then rebuild:

`yarn cache clean && yarn clean && yarn install && yarn build`

### Changelog

**New**

- `yarn reset` command

**Changed**

- package names switched to the @carbon namespace
- All supporting documentation and package imports
- Renamed lint packages to be consistent


packages/eslint-config-ibmdotcom/package.json
```diff
@@ -1,7 +1,8 @@
 {
-  "name": "@ibmdotcom/eslint-config-ibmdotcom",
+  "name": "@carbon/eslint-config-ibmdotcom",
```
photodow: Just thinking about naming strategy and consistency, all the others start with `ibmdotcom`. Should these be named like... `@carbon/ibmdotcom-eslint-config`?

photodow: One other thought on the naming of the directory this lives in. As an example we don't include the name of the package in the folder name for `react`, but for namespace and context we include it in the package name on `npm`. I think that approach works, and cuts down redundancy since we already know we are within the `ibmdotcom` context when working locally. Should we do something similar for our `config` type packages here as well?

```
|
|-react
|-eslint-config
|-styles
|-stylelint-config
```

```
@carbon/ibmdotcom-react
@carbon/ibmdotcom-eslint-config
@carbon/ibmdotcom-styles
@carbon/ibmdotcom-stylelint-config
```

jeffchew: @photodow good question. I was actually going that route, in fact was going with `config-eslint` and `config-stylelint` so in the package folder they would be essentially next to each other. But then I decided to change it back and have `ibmdotcom` be at the end, since I was looking at the other packages and they seem to have that pattern and probably was why carbon went with that naming pattern as well. e.g. `eslint-config-[name of package]` and `stylelint-config-[name of package]`. Looking at public ones like `eslint-config-airbnb` or `stylelint-config-prettier`, etc. ESlint even has documentation on `shareable configs` that follow this naming pattern, if we ever decide to share ours:

https://eslint.org/docs/developer-guide/shareable-configs

photodow:
Is there any value in passing this naming scheme to our builds?  
(e.g. `ibmdotcom-utilities.min.js --> carbon-ibmdotcom-utilities.min.js`)

Or our hosting builds?
(e.g.  `ibmdotcomlibrary-services.mybluemix.net --> carbon-ibmdotcomlibrary-services.mybluemix.net`)
*I realize that's long, but it's consistent. I wonder if we could also use `carbon-ibmdotcom` subdomain, and each package is in a directory. (e.g. `carbon-ibmdotcom.mybluemix.net/react`)*

jeffchew:
> Is there any value in passing this naming scheme to our builds?
> (e.g. `ibmdotcom-utilities.min.js --> carbon-ibmdotcom-utilities.min.js`)
> 
> Or our hosting builds?
> (e.g. `ibmdotcomlibrary-services.mybluemix.net --> carbon-ibmdotcomlibrary-services.mybluemix.net`)
> _I realize that's long, but it's consistent. I wonder if we could also use `carbon-ibmdotcom` subdomain, and each package is in a directory. (e.g. `carbon-ibmdotcom.mybluemix.net/react`)_

These actually are placeholder for now, we may not even be on IBM Cloud if we can secure free hosting on the EI server. We'd ultimately be pointing the sites to something on www.ibm.com, at least for the cupcake version. The wedding cake version is still up in the air I think.

PR #74, opened 2019-09-04
Feat/863 patterns package

jeffchew:
### Related Ticket(s)

Package | pattern-library architecture
https://github.ibm.com/webstandards/digital-design/issues/863
CI/CD pipeline setup for ibmdotcom-patterns
https://github.ibm.com/webstandards/digital-design/issues/869
Environments setup for pattern-library
https://github.ibm.com/webstandards/digital-design/issues/878

### Description

This sets up the patterns package in the IBM.com Library monorepo. There is a sample `LeadSpace` component, storybook, storyshots, placeholder pattern styles, and netlify setup.

Initial Netlify deploy:
https://5d6fb2f6ddaaa5000840550b--ibmdotcom-patterns.netlify.com

### Changelog

**New**

- Patterns package

**Changed**

- Added global imports in styles


packages/styles/scss/components/footer/index.scss
```diff
@@ -5,16 +5,14 @@
 // LICENSE file in the root directory of this source tree.
 //
 
-@import 'carbon-components/scss/globals/scss/vendor/@carbon/elements/scss/import-once/import-once';
-@import 'carbon-components/scss/globals/scss/vendor/@carbon/layout/scss/mini-unit';
-@import 'carbon-components/scss/globals/scss/vendor/@carbon/layout/scss/breakpoint';
+@import '../../globals/imports';
+//@import 'carbon-components/scss/globals/scss/vendor/@carbon/layout/scss/mini-unit';
```
jeffchew: @photodow I wanted to get your feedback on this one. I created a global imports file, which seemed to align with the other components but noticed that yours is only targeting specific elements in `@carbon/layout`, was wondering if this would be a deal breaker if we switch this out to the global import instead? I commented out for now and didn't remove.
cc: @kennylam @annawen1

jeffchew: I checked and all layout imports are functions and shouldn't affect file size output.

kennylam: @jeffchew @photodow James has suggested in other PRs using `mini-unit` for spacing certain elements, as opposed to some of the `$spacing` tokens. Could go into global.

photodow: So if I understand you correctly you're talking about including something like `@carbon/layout/scss/layout` which includes a lot of things along with the `mini-unit`? I don't think that's a deal-breaker. As you mentioned the layouts, in particular, is only functions, so it won't add anything unless we use those provided. As long as the functions being used are represented in global somehow I think it'll be ok. Personal preference is to be more explicit with things like this though, and don't include a bunch of things we don't need/use so trying to be minimal, but as long as it works at the end of day we should be good.😀

photodow: Are we going to do the same with grid, and type? Those we need to be careful with because they add CSS upon build.

jeffchew: It might be worth adding a note in the global include to be careful about including imports that would add css on builds.

packages/patterns/.cfignore
```diff
@@ -0,0 +1,4 @@
+*
+!manifest.yml
```
photodow: Doesn't cloud foundry need a `manifest.yml` to work? Or does the CLI take care of the `manifest.yml` before uploading to cloud foundry?

packages/patterns/package.json
```diff
@@ -0,0 +1,139 @@
+{
+  "name": "@carbon/ibmdotcom-patterns",
```
photodow: One thought on the name... Should we add `-react` to the end leaving the door open for vanilla, angular and vue if we get the time and resources?

jeffchew: we probably will need to at minimum create the vanilla version eventually. was thinking that name would be `@carbon/ibmdotcom-patterns-vanilla`. We could rename this one to `@carbon/ibmdotcom-patterns-react` or keep it as `@carbon/ibmdotcom-patterns` if react is going to be the "source of truth"? I'm fine either way.
Thoughts @kennylam @annawen1 ?

packages/styles/scss/components/footer/index.scss
jeffchew: @photodow I added comments to the global import, let me know if that looks good.

cc: @kennylam @annawen1

packages/patterns/README.md
```diff
@@ -0,0 +1,76 @@
...
+> A collection of IBM.com patterns implemented using
+> [React](https://reactjs.org/) and
+> [Carbon](https://www.carbondesignsystem.com/).
+
+## Getting Started
+
+Run the following command using [npm](https://www.npmjs.com/):
+
+```bash
+npm i @carbon/ibmdotcom-patterns carbon-components-react carbon-components carbon-icons
```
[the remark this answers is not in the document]
jeffchew: It might be, don't see it as a dependency for now though.

packages/patterns/docs/feature-flags.md
```diff
@@ -0,0 +1,102 @@
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+## Table of Contents
```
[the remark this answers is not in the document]
jeffchew: I was thinking the same thing when I duplicated it. we'd probably need to consolidate it in some way where it points to react and patterns without being too confusing. I haven't even thought about how feature flags will work for vanilla, and that might look pretty different from how the react version is. I think we can ponder on this in another PR though.

packages/patterns/.cfignore
jeffchew: I think this is actually carry over from the carbon repo. we do need a manifest file for cloud foundry, but since we are transitioning to EI we can clean this up later.

packages/patterns/docs/feature-flags.md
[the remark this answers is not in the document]
photodow: Sounds great. Just created an issue to help us keep it in mind moving forward.

https://github.ibm.com/webstandards/digital-design/issues/1605

PR #76, opened 2019-09-05
feat(translation): added translation service

jeffchew:
### Related Ticket(s)

**Translation background services MVP**
https://github.ibm.com/webstandards/digital-design/issues/777

### Description

This pull request includes the translation API that currently grabs the translation data from www.ibm.com. As part of MVP, this goes directly to en-us for now. Ability to switch location/country will be in a future PR.

As there were CORS issues when connecting to www.ibm.com, the code also allows a CORS proxy server to be used to fetch the data, where `CORS_PROXY` is the environment variable to set this configuration. The proxy server will only be used if the request host does not match the responding host (www.ibm.com). 

### Changelog

**New**

- Translation API
- CORS_PROXY environment variable


packages/services/.env.example
```diff
@@ -9,3 +9,6 @@ SEARCH_TYPEAHEAD_HOST=<host for ibm.com search, e.g. https://www-api.ibm.com>
 # Marketing Search
 MARKETING_SEARCH_VERSION=v3
 MARKETING_SEARCH_HOST=<host for ibm.com marketing search, e.g. https://www.ibm.com>
+
+# Translation
+CORS_PROXY=<CORS proxy server for fetching from www.ibm.com, e.g. https://yourproxy.com/>
```
photodow: Do we need to add this to the `react` `.env.example` file as well? It took me a little while to figure out it was required in the package it's running from to make it work. Part of that was my ignorance too. Maybe that or add note here that `.env` in this file won't be enough to get it running locally.

jeffchew: I'll add it to the react .env.example, won't hurt to have it there

jeffchew: @photodow added!

packages/services/src/services/Translation/Translation.js
```diff
@@ -0,0 +1,57 @@
...
+   * @returns {Promise<any>} Translation data
+   * @example
+   * import { TranslationAPI } from '@carbon/ibmdotcom-services';
+   *
+   * async function getTranslation() {
+   *   const response = await TranslationAPI.getTranslation();
+   *   return response;
+   * }
+   */
+  static async getTranslation() {
```
photodow: I wonder if it'd be good to provide may be at least two additional methods. `.getTranslationMasthead()`, and `.getTranslationFooter()`. That might give us flexibility down the road to scale, and depending on how Drupal delivers it'd allow us to split the two calls up if needed/desired.

packages/services/src/services/Translation/Translation.js
```diff
@@ -0,0 +1,57 @@
...
+    const proxy = currenthost !== _host ? _proxy : '';
+    const url = `${proxy}${_endpoint}/${cc}${lc}.json`;
+
+    return await axios
+      .get(url, {
+        headers: {
+          'Content-Type': 'text/plain',
+        },
+      })
+      .then(response => response.data);
```
photodow: Drupal is going to take over this service for us right? Is it too early to think about whether we keep the data structure, or if it requires a more relevant data model? If we know or want to be opinionated about that model this service would be good to do that remodel here. As an immediate example, we might want to combine the separated `socialSharing`, `footerThin` into `footerMenu`. We could also sync up our language here too. Either change the components to match the service API call or vice versa. Just some thoughts as we move forward.

PR #77, opened 2019-09-09
Release/v0.4.0

jeffchew:
### Related Ticket(s)

https://github.ibm.com/webstandards/digital-design/issues/1544

### Description

This include the release v0.4.0 and associated changelogs and tags. The other change that was made was the publish scripts, where it appears that we would need to include the `--conventional-prerelease` and `--conventional-graduate` for the versions to bump properly after a release (from what I can tell). In addition, I have also added the option to create a release candidate for a major bump.

Another change is that I removed `--no-git-tag-version`. This essentially embraces the creation of package level tags, which I think was causing the changelogs from not rendering correctly. this means we will be embracing the creation and auto pushing of the changelogs and package bumps to Git, so I added the documentation back to authenticate with Github in the publishing documentation.

As for the changelogs, I manually cleaned them up as even deleting them entirely then generating again produced odd results. From what I can tell, the changelogs need to be properly derived from the package level tags (e.g. `ibmdotcom-react@vX.Y.Z` and not `vX.Y.Z`), which we are not tagging with currently but hoping with the next release we will see cleaner automated changelogs.

### Changelog

**Changed**

- Changelogs
- Updated publish scripts


jeffchew:
> Aside from the build issue, this looks good to me. 👍

The build issue is because the patterns package doesn't exist yet in this branch 😄

PR #78, opened 2019-09-11
feat(leadspace): leadspace component alpha version

annawen1:
### Related Ticket(s)

[Alpha: Lead Space #1660 ](https://github.ibm.com/webstandards/digital-design/issues/1660)

### Description

Alpha version of the Lead Space pattern

Specs here: [SPEC file](https://ibm.box.com/s/dqndq9s7uqvfsjunwbrn35gsulh43307)

Screen Shot 2019-09-11 at 4 14 32 PM

### Changelog

**New**

- Lead Space component
- import grid styling from carbon


packages/styles/scss/components/leadspace/_leadspace.scss
```diff
@@ -0,0 +1,69 @@
...
+      width: 100%;
+    }
+
+    h1 {
+      @include carbon--type-style(display-01);
+      @include carbon--make-col(7, 16);
+    }
+
+    p {
+      @include carbon--type-style(expressive-heading-03);
```
photodow: I think the expressive type you're using in this file is supposed to be fluid. I think in order to make that happen you need to pass in the `fluid` parameter. I think it's `false` by default.

```scss
@include carbon--type-style(expressive-heading-03, true);
```

I think this would also apply for [#L32](https://github.com/carbon-design-system/ibm-dotcom-library/pull/78/files#diff-581c7770913f29a42c27bb5d30476ce9R32) as well.

packages/styles/scss/components/leadspace/_leadspace.scss
```diff
@@ -0,0 +1,69 @@
...
+      @include carbon--make-col(4, 16);
+    }
+
+    button {
+      margin: carbon--mini-units(4) carbon--mini-units(2) 0 0;
```
photodow: This might be a good use case for using `$layout-01`, and `$layout-03`?

packages/patterns-react/src/patterns/LeadSpace/__stories__/LeadSpace.stories.js
```diff
@@ -14,5 +14,40 @@ storiesOf('LeadSpace', module)
...
+        copy: 'Secondary action',
+        renderArrow: false,
+      },
+    ];
+
+    return (
+      <LeadSpace
+        title={title}
+        copy={copy}
+        buttons={object('buttons', buttons)}
```
photodow: I wonder if the designers are going to have difficulty trying to test this out. Any ideas on an easier way to knob this out?

packages/patterns-react/src/patterns/LeadSpace/__stories__/LeadSpace.stories.js
```diff
@@ -14,5 +14,40 @@ storiesOf('LeadSpace', module)
...
+    const variations = {
+      expressive: '',
+      productive: 'productive',
+    };
+
+    const imageUrl = text('imageUrl', 'https://picsum.photos/id/1025/1000/400');
+
+    const buttons = [
+      {
+        link: '',
```
photodow: Have you tested this by filling this out yet? Apparently it changes to an `a` tag when you add this property, and it changes the design up.

image

packages/patterns-react/src/patterns/LeadSpace/LeadSpace.js
```diff
@@ -5,13 +5,69 @@
...
+ * @param {Array} props.buttons array of buttons for lead space (max 2 buttons)
+ * @returns {*} Lead space component
  */
-const LeadSpace = () => <div data-autoid="leadspace">Lorem Ipsum</div>;
+const LeadSpace = ({ variation, title, copy, buttons, imageUrl }) => (
+  <div
+    className={classnames(`${prefix}--leadspace`, {
+      [`${prefix}--leadspace--productive`]: variation === 'productive',
+    })}>
+    <img className={`${prefix}--leadspace__image`} src={imageUrl} alt="" />
```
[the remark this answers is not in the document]
jeffchew: I'm actually wondering why this isn't a `<picture>` tag.

packages/styles/scss/components/leadspace/_leadspace.scss
```diff
@@ -0,0 +1,69 @@
...
+// LICENSE file in the root directory of this source tree.
+//
+
+@import '../../globals/imports';
+@import 'carbon-components/scss/globals/scss/vendor/@carbon/elements/scss/grid/grid';
+
+@mixin leadspace {
+  .#{$prefix}--leadspace {
+    position: relative;
+    height: carbon--rem(560px);
```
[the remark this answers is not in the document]
annawen1: done

annawen1:
> @annawen1 need to move this into `patterns-react`

done!

packages/styles/scss/patterns/leadspace/_leadspace.scss
```diff
@@ -5,11 +5,135 @@
...
+    @include carbon--type-style(expressive-heading-06, true);
+  }
+
+  @include carbon--breakpoint(md) {
+    .#{$prefix}--leadspace {
+      padding-top: carbon--mini-units(50) / carbon--mini-units(84) * 100%;
+
+      &__overlay {
+        padding: $layout-03;
+        background-image: linear-gradient(
```
photodow: Can you remove the following two properties if you do `background: linear-gradient(to right, $ui-background, transparent 50%);` here instead?

packages/styles/scss/patterns/leadspace/_leadspace.scss
```diff
@@ -5,11 +5,135 @@
...
+      }
+    }
+  }
+
+  @include carbon--breakpoint(lg) {
+    .#{$prefix}--leadspace {
+      padding-top: carbon--mini-units(60) / carbon--mini-units(132) * 100%;
+
+      h1 {
+        @include carbon--make-col(8, 16);
```
photodow: I still don't think this grid is right. You need the `carbon--make-container`, `carbon--make-row`, and `carbon--make-col-ready`.

image

packages/styles/scss/patterns/leadspace/index.scss
```diff
@@ -0,0 +1,11 @@
+@import '../../globals/imports';
+@import '@carbon/themes/scss/themes';
```
photodow: Should these be in the global imports?

packages/styles/scss/patterns/leadspace/index.scss
```diff
@@ -0,0 +1,11 @@
...
+@import 'carbon-components/scss/components/button/_button.scss';
+@import './leadspace';
+
+.bx--leadspace--g100 {
```
photodow: Might want to include the `$prefix`.

packages/patterns-react/src/patterns/LeadSpace/LeadSpace.js
```diff
@@ -5,13 +5,69 @@
...
+ * @param {string} props.copy lead space short copy to support the title
+ * @param {string} props.imageUrl lead space image
+ * @param {Array} props.buttons array of buttons for lead space (max 2 buttons)
+ * @returns {*} Lead space component
  */
-const LeadSpace = () => <div data-autoid="leadspace">Lorem Ipsum</div>;
+const LeadSpace = ({ variation, title, copy, buttons, imageUrl }) => (
+  <div
+    className={classnames(`${prefix}--leadspace`, {
+      [`${prefix}--leadspace--productive`]: variation === 'productive',
```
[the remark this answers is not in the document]
kennylam: @annawen1 [This](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/react/src/components/Masthead/Masthead.js#L63) is how i was doing it, if we want to stay consistent.

packages/styles/scss/components/leadspace/_leadspace.scss
```diff
@@ -0,0 +1,69 @@
...
+
+    &__image {
+      object-fit: cover;
+      width: 100%;
+      height: 100%;
+    }
+
+    &__content-container {
+      padding: carbon--mini-units(4) 0 carbon--mini-units(4)
```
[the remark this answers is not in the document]
kennylam: @photodow I'll circle back once i check, but I had to do similar with grid padding for a few things to match the designs from Wonil & team

packages/patterns-react/src/patterns/LeadSpace/LeadSpace.js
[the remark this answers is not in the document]
jeffchew: I agree, we should break out the logic to keep it clean.

packages/styles/scss/components/leadspace/_leadspace.scss
```diff
@@ -0,0 +1,69 @@
...
+@import '../../globals/imports';
+@import 'carbon-components/scss/globals/scss/vendor/@carbon/elements/scss/grid/grid';
+
+@mixin leadspace {
+  .#{$prefix}--leadspace {
+    position: relative;
+    height: carbon--rem(560px);
+
+    &__image {
+      object-fit: cover;
```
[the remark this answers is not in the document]
annawen1: hmmm good point so based off of [https://caniuse.com/#feat=object-fit](https://caniuse.com/#feat=object-fit) it looks like IE11 does not support it...is IE11 something we need to support? 

@jeffchew @kennylam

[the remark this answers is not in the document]
jeffchew: @annawen1 For the time being, yes. However, we could make an argument to take more of a progressive enhancement approach to this. We would essentially use this feature, but then let IE11 fall back to whatever experience that might be. We would need to probably capture some screenshots of what they are though and then present them to the design team to make sure they are ok with this.
cc: @larahanlon2 @photodow @kennylam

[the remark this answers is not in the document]
jeffchew: To the point of this being alpha though, we probably don't need to worry about this for now, especially if this is behind a feature flag.

packages/patterns-react/src/patterns/LeadSpace/__stories__/LeadSpace.stories.js
jeffchew: I'd say let's hold on this for now, I showed @larahanlon2 and Wonil how to swap out images in storybook so I think they know. In the long term, we are working with Jill and others on what the functional specs for these might be and want to see where they land as far as how these props should be shaped. I still don't know if an array of objects is something that they would want these to be at the end of the day, for example.

packages/patterns-react/src/patterns/LeadSpace/LeadSpace.js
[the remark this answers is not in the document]
annawen1: ok, I'll pull this out like how kenny has done.

packages/patterns-react/src/patterns/LeadSpace/LeadSpace.js
```diff
@@ -5,13 +5,92 @@
...
-const LeadSpace = () => <div data-autoid="leadspace">Lorem Ipsum</div>;
+const LeadSpace = ({ variation, title, copy, buttons, image }) => (
+  <section
+    data-autoid="leadspace"
+    className={classnames(`${prefix}--leadspace`, {
+      [`${prefix}--leadspace--productive`]: variation === 'productive',
+    })}>
+    <div className={`${prefix}--leadspace__container`}>
+      {image && (
+        <picture>
```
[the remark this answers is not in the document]
annawen1: this is a good point - we could consider building out a component that takes in any number of image sources. For the leadspace at different breakpoints we have the different aspect ratios so user will have to put in the 3 sources otherwise it ends up looking wonky.

If we were to build out a component - would that be something that would go under the `react` package?

packages/styles/scss/patterns/leadspace/_leadspace.scss
```diff
@@ -5,11 +5,135 @@
...
+    }
+
+    button:nth-child(2),
+    a:nth-child(2) {
+      margin: $layout-03 0 $layout-01 0;
+    }
+  }
+
+  .#{$prefix}--leadspace--productive h1 {
+    @include carbon--type-style(expressive-heading-06, true);
```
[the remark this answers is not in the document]
annawen1: From my convo with @larahanlon2:
```
"Technically the whole lead space pattern is expressive. So all of the headings are expressive

The display-01 option is just more "expressive". We want to provide this option for key marketing or campaign pages.
The heading-06 option would be for product pages, maybe documentation or commerce.
We have yet to clearly define exactly what page types and where in the journey these would appear but we do want to have the two available."
```

packages/styles/scss/patterns/leadspace/_leadspace.scss
```diff
@@ -5,11 +5,135 @@
...
+      color: $text-01;
+    }
+
+    &__ctas {
+      display: flex;
+      flex-direction: column-reverse;
+    }
+
+    button,
+    a {
```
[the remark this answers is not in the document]
annawen1: 👍simplifies it so much

packages/styles/scss/patterns/leadspace/_leadspace.scss
```diff
@@ -5,11 +5,135 @@
...
+      width: 100%;
+    }
+
+    button:nth-child(2),
+    a:nth-child(2) {
+      margin: $layout-03 0 $layout-01 0;
+    }
+  }
+
+  .#{$prefix}--leadspace--productive h1 {
```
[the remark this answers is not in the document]
annawen1: done, changed the same with `p` tags as well by changing them to `.bx--leadspace__desc`

packages/styles/scss/patterns/leadspace/_leadspace.scss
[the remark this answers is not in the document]
larahanlon2: Hey @photodow @annawen1 yeah, this still needs to be worked out. All of the type-tokens are expressive and that's correct. 

The two options we want to offer have different type styles – one uses `display 01`, the other uses `heading 06`. We would like for the `display 01` to be used in key marketing (or really "expressive") moments, while we would like `heading 06` to be used in more detailed pages like product pages. So...that's where the labeling of expressive Vs productive for the different options came from. But we do need to change so it's less confusing. TBD!

wonilsuhibm:
Hello (test comment)

packages/styles/scss/components/leadspace/_leadspace.scss
```diff
@@ -0,0 +1,69 @@
...
+      width: 100%;
+      height: 100%;
+    }
+
+    &__content-container {
+      padding: carbon--mini-units(4) 0 carbon--mini-units(4)
+        carbon--mini-units(4);
+      position: absolute;
```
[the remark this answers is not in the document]
annawen1: yep with what I'm doing to maintaining the aspect ratio of the component I would have to make the `.bx--leadspace__content` absolute.

```
.#{$prefix}--leadspace {
    // doing this padding thing to maintain the aspect ratio
    padding-top: carbon--rem(370px) / carbon--mini-units(40) * 100%;
}
```

packages/patterns-react/src/patterns/LeadSpace/LeadSpace.js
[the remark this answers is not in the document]
annawen1: sure

packages/styles/scss/patterns/leadspace/_leadspace.scss
```diff
@@ -5,11 +5,135 @@
...
+    }
+
+    button,
+    a {
+      width: 100%;
+    }
+
+    button:nth-child(2),
+    a:nth-child(2) {
+      margin: $layout-03 0 $layout-01 0;
```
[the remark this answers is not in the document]
annawen1: yeah, let me play around with it so we don't have to have the `:nth-child(2)` blocks

packages/styles/scss/patterns/leadspace/_leadspace.scss
annawen1: yep

packages/styles/scss/patterns/leadspace/_leadspace.scss
```diff
@@ -5,11 +5,135 @@
 // LICENSE file in the root directory of this source tree.
 //
 
-@import '../../globals/imports';
-
 @mixin leadspace {
   .#{$prefix}--leadspace {
```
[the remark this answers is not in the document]
annawen1: I actually like having the theme class wrap around the component just to separate the two out, but if there is a push to have it on the same level I don't mind doing the above.

packages/styles/scss/patterns/leadspace/_leadspace.scss
annawen1: I initially added those and I believe the padding was weird but didn't dig further - let me try again and see if I'm getting any issues.

packages/styles/scss/components/leadspace/_leadspace.scss
```diff
@@ -0,0 +1,69 @@
...
+      top: 0;
+      height: 100%;
+      width: 100%;
+    }
+
+    h1 {
+      @include carbon--type-style(display-01);
+      @include carbon--make-col(7, 16);
```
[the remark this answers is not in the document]
annawen1: ok I think I've gotten the grid down - @photodow let me know if it's still off

packages/react/src/components/HorizontalRule/__stories__/HorizontalRule.stories.js
```diff
@@ -52,10 +52,14 @@ storiesOf('HorizontalRule', module)
           <div className="bx--row bx--no-gutter">
             <div className="bx--col">
               <HorizontalRule
-                style={select('style', styles, '')}
-                size={select('size', sizes, '')}
-                contrast={select('contrast', contrasts, '')}
-                weight={select('weight', weights, '')}
+                style={select('style', styles, styles.solid)}
```
photodow: Love these changes.

I'm not sure how the rest of the team feels about this, but changes like this if we were to create a separate pull request we could get it in and review it a lot faster than including it within an already opened but larger pull request like this one.

Just something to think about as we move forward.

cc: @jeffchew @kennylam

annawen1: I thought since this was a minor change I could just pull it in but I can see it interfering with the main goal of this PR - I'll make sure to separate it out next time

packages/styles/scss/patterns/leadspace/_leadspace.scss
```diff
@@ -5,11 +5,133 @@
...
+      height: 100%;
+      width: 100%;
+    }
+
+    &__overlay {
+      @include carbon--make-container;
+
+      padding-top: $layout-03;
+      padding-bottom: $layout-03;
+      position: absolute;
```
photodow: The biggest problem I'm seeing with the grid right now is it's left aligned because of this `position: absolute` and a couple of the properties below it. You might `position: absolute` on a parent, or have the grid container start as a child of this block.


*It should look like this...*
image

*This is what I'm seeing...*
image

packages/react/src/components/HorizontalRule/__stories__/HorizontalRule.stories.js
photodow: You're good! 😄 I just think we can get it in faster outside. Same outcome I think. It's more of a comment to consider down the road.

packages/styles/scss/patterns/leadspace/_leadspace.scss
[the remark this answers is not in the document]
photodow: If we're going to include it as a wrapper then it might be good to include the wrapper inside the component?

packages/styles/scss/patterns/leadspace/_leadspace.scss
annawen1: @photodow I think we are good to go now with the package updates in

packages/styles/scss/patterns/leadspace/_leadspace.scss
```diff
@@ -5,11 +5,141 @@
...
+          transparent 50%
+        );
+      }
+
+      &__content {
+        width: 100%;
+      }
+
+      .#{$prefix}--leadspace__title {
+        @include carbon--make-row;
```
photodow: I think this is causing a slight issue with the grid. With an 8/16 col you should be able to get two of these in a row, and it also wraps. I'm not sure if the wrapping is related, or a separate issue at the moment.

image

packages/styles/scss/patterns/leadspace/_leadspace.scss
```diff
@@ -5,11 +5,141 @@
...
+      padding-top: carbon--mini-units(60) / carbon--mini-units(132) * 100%;
+
+      .#{$prefix}--leadspace__title {
+        @include carbon--make-col(8, 16);
+      }
+
+      .#{$prefix}--leadspace__desc {
+        @include carbon--make-row;
+        @include carbon--make-col-ready;
+        @include carbon--make-col(4, 16);
```
photodow: We're having a slightly different alignment issue down here.

image

packages/styles/scss/patterns/leadspace/_leadspace.scss
```diff
@@ -5,11 +5,141 @@
...
+        @include carbon--make-col-ready;
+        @include carbon--make-col(7, 8);
+      }
+
+      .#{$prefix}--leadspace__desc {
+        width: 50%;
+      }
+
+      &__ctas {
+        flex-direction: row-reverse;
```
photodow: Should the CTA be apart of the same column as the desc? It appears to just be floating outside of the grid.

image

packages/styles/scss/patterns/leadspace/_leadspace.scss
photodow: With that said for some reason when I go to add 4 more columns it just stacks them.

image

packages/styles/scss/patterns/leadspace/_leadspace.scss
annawen1: Is the issue here that I should move the @include carbon--make-row into a parent div?

packages/styles/scss/patterns/leadspace/_leadspace.scss
annawen1: I chose not to put the CTAs in the same column as the desc as the desc can have a wider or smaller width than the cta div. Do I need to wrap a row around the CTAs? It's always going to be positioned bottom left aligned.

packages/styles/scss/patterns/leadspace/_leadspace.scss
annawen1: how are you adding the 4 columns? using the mixins or the grid classes?

packages/styles/scss/patterns/leadspace/_leadspace.scss
```diff
@@ -5,11 +5,141 @@
...
+      max-width: none;
+      background-image: linear-gradient(
+        to bottom,
+        rgba($ui-background, 1),
+        transparent
+      );
+    }
+
+    &__content {
+      position: absolute;
```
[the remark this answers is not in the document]
annawen1: whoops - fixing

packages/styles/scss/patterns/leadspace/_leadspace.scss
annawen1: currently seeing this 
Screen Shot 2019-09-19 at 10 23 28 AM

packages/styles/scss/patterns/leadspace/_leadspace.scss
annawen1: I removed the `absolute` positioning from the `content` div so it looks a bit different...slightly off still

Screen Shot 2019-09-19 at 11 00 10 AM

annawen1:
> This is good to go as an alpha! Looking amazing.
> 
> A few comments towards the final version:
> 
> * needs max width (1584px per 2x grid) and max height (640px?) specs
> * need to allow different icons for the buttons
> * the two buttons need to be the same size (match the wider)
> * the gradient overlay needs to be optional
> * the gradient overlay feels a bit too strong



> This is good to go as an alpha! Looking amazing.
> 
> A few comments towards the final version:
> 
> * needs max width (1584px per 2x grid) and max height (640px?) specs
> * need to allow different icons for the buttons
> * the two buttons need to be the same size (match the wider)
> * the gradient overlay needs to be optional
> * the gradient overlay feels a bit too strong
> 
> ### on `sm` breakpoint:
> Screen Shot 2019-09-19 at 3 13 40 PM
> 
> * `2.` should align with `1.`.
> * break text early instead of spanning the full width, around `3.`
> 
> Thank you!!!

Ah yep, the `2. not aligning with 1.` I realized I hadn't fully pushed my css changes - should be addressed in the latest netilfy build.

For the rest of the comments I'll hold onto and open another PR after this alpha version has been merged. Thanks!

larahanlon2:
@annawen1 yeah, this is amazing, thank you so much! 

One thing we need to agree on is what we're calling the `variation`. They can't be `expressive` and `productive` because it's confusing. Can we change that easily enough?

Is it possible to have the `gradient overlay` layer as a knob? In some cases people may have a well-produced image that doesn't need the gradient overlay. Maybe we could turn the gradient show by default but have a knob to hide?

annawen1:
> One thing we need to agree on is what we're calling the `variation`. They can't be `expressive` and `productive` because it's confusing. Can we change that easily enough?
> 
> Is it possible to have the `gradient overlay` layer as a knob? In some cases people may have a well-produced image that doesn't need the gradient overlay. Maybe we could turn the gradient show by default but have a knob to hide?

@larahanlon2 yep should be an easy change to swap out names. I can add a `gradient overlay` knob so it's optional in my next PR

PR #75, opened 2019-09-04
feat(footer): added locale-button to footer

photodow:
### Related Ticket(s)

- [#1640](https://github.ibm.com/webstandards/digital-design/issues/1640)
- [#1078](https://github.ibm.com/webstandards/digital-design/issues/1078)
- [#1485](https://github.ibm.com/webstandards/digital-design/issues/1485)
- [#1368](https://github.ibm.com/webstandards/digital-design/issues/1368)

### Description

This pull request makes some progress forwards for the local selector and places it inside the footer behind a feature flag. Clearly more work needs to be done, but this has already helped scope down some of the prototype and animation exploration work.

image

### Changelog

**New**

- Feature flagged locale selector button in the footer

PR #81, opened 2019-09-13
Feature Flag Utility

photodow:
### Related Ticket(s)

https://github.com/carbon-design-system/ibm-dotcom-library/pull/75#discussion_r323477375

### Description

This pull request cherry-picked one or two commits from #75 so we can deploy it to alpha faster for further testing, and leveraging. It basically adds a feature flag utility that determines if a feature flag is turned on and decides to render the jsx or not.

### Changelog

**New**

- feature flag utility

### DIFF
--- a/packages/react/src/components/Footer/LocaleButton.js
+++ b/packages/react/src/components/Footer/LocaleButton.js
@@ -1,0 +1,70 @@
+import { FOOTER_LOCALE_BUTTON } from '../../internal/FeatureFlags.js';
+import { featureFlag } from '@carbon/ibmdotcom-utilities';
+
+import React, { useState } from 'react';
+import {
+  Button,
+  ComposedModal,
+  ModalHeader,
+  ModalBody,
+} from 'carbon-components-react';
+import { settings } from 'carbon-components';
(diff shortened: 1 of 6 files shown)


packages/react/src/components/Footer/LocaleButton.js
```diff
@@ -0,0 +1,70 @@
...
+      </ComposedModal>
+    </div>
+  );
+
+  /**
+   * Sets modal state to open
+   *
+   * @private
+   */
+  function open() {
```
kennylam: @photodow Should we standardize placement of functions like this purely for consistency? Only because all [Masthead](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/react/src/components/Masthead/MastheadSearch.js) functions like this come at the start (and mine are longer haha)

jeffchew: @kennylam this was merged in already, I think we can continue the conversation on this in https://github.com/carbon-design-system/ibm-dotcom-library/pull/75 and make any changes if necessary.

PR #120, opened 2019-09-23
Feat/1359 truste

jeffchew:
### Related Ticket(s)

**Add HTML element in the footer for TrustE module injection**
https://github.ibm.com/webstandards/digital-design/issues/1359

### Description

This adds in an additional `<li>` element to the footer navigation that is reserved for injecting the TrustE link for cookie preferences.

### Changelog

**Changed**

- Added cookie preferences placeholder element


packages/react/src/components/Footer/LegalNav.js
```diff
@@ -30,6 +44,7 @@ const LegalNav = ({ links }) => {
       <nav className={`${prefix}--legal-nav`}>
         <ul className={`${prefix}--legal-nav__list`}>
           {renderListItems(links)}
+          {renderTrusteItem()}
```
photodow: I'm not crazy about having an empty `li` in the footer at all times. This also renders the `:last-child` styling here pointless unless the link is injected.

packages/react/src/components/Footer/LegalNav.js
```diff
@@ -12,6 +12,20 @@ import { Link } from 'carbon-components-react';
...
+/**
+ * Placeholder <li/> element for injection of the TrustE cookie preferences link
+ *
+ * @returns {*} JSX object
+ */
+const renderTrusteItem = () => {
+  return (
+    <li
+      className={`${prefix}--legal-nav__list-item`}
+      data-autoid="dds-privacy-cp"
```
photodow: > **From @jeffchew via Slack**
> one thing I'm wondering though is the styling for the link, right now we have the links use `bx--link`, we should probably apply the styling for Juan when they add the link...was thinking possibly setting it in our styles then do an `@extend` of `bx--link` when targeting the `a` element under the nav link item

I'm not crazy about using extend in this scenario or styling the `a` directly. I see no reason why they couldn't inject the link with the class name. They have all of the carbon components, right? Worst case scenario is maybe we can add a utility class where they provide the `url`, and `text`, and we inject it for them?

jeffchew:
> Are we sure this is going to be enough for them? They still need to know when this element is added to the page before they can inject the new link, don't they? Do we also need to emit an event listener for when it's loaded? If that's true, why not just emit the event listener, and let them add the list item with the stable selector as well? Just a thought. Nothing inherently looks wrong with this pull request, I'm just trying to think through the solution a little more here.

Good questions. Let's provide this as-is for now and see if this is enough for his team.

PR #128, opened 2019-09-24
fix(jest): multiple jest fixes for the footer

jeffchew:
### Related Ticket(s)

No related issues

### Description

This fixes a number of jest test issues. With most of them, the footer nav links did not account for the additional placeholder element for the truste module so incremented it by 1.

The other thing is to remove the required PropTypes for the various footer components. I felt it wasn't necessary since the component itself already renders null if no items are passed in, and the shape of the props are already covered and should throw console logs accordingly.

### Changelog

**Changed**

- Jest fixes for the footer

### DIFF
--- a/packages/react/src/components/Footer/FooterNav.js
+++ b/packages/react/src/components/Footer/FooterNav.js
@@ -45,13 +45,26 @@ function renderGroups(groups) {
   ));
 }
 
+/**
+ * @property propTypes
+ * @description Defined property types for component
+ * @type {{groups: shim}}
+ */
 FooterNav.propTypes = {
   groups: PropTypes.arrayOf(
     PropTypes.shape({
-      title: PropTypes.string.isRequired,
-      links: PropTypes.array.isRequired,
-    }).isRequired
-  ).isRequired,
+      title: PropTypes.string,
+      links: PropTypes.array,
+    })
+  ),
+};
+
+/**
+ * @property defaultProps
+ * @type {{groups: Array}}
+ */
+FooterNav.defaultProps = {
+  groups: null,
 };
 
 export default FooterNav;
--- a/packages/react/src/components/Footer/__tests__/LegalNav.test.js
+++ b/packages/react/src/components/Footer/__tests__/LegalNav.test.js
@@ -37,7 +37,7 @@ describe('<LegalNav />', () => {
 
     expect(legalNav.exists('.bx--legal-nav')).toBeTruthy();
     expect(legalNav.find('.bx--legal-nav__list-item')).toHaveLength(
-      FOOTER_NAV_DATA.length - 1
+      FOOTER_NAV_DATA.length
     );
   });
 
@@ -48,7 +48,7 @@ describe('<LegalNav />', () => {
 
     expect(legalNav.exists('.bx--legal-nav')).toBeTruthy();
     expect(legalNav.find('.bx--legal-nav__list-item')).toHaveLength(
-      FOOTER_NAV_DATA.length - 1
+      FOOTER_NAV_DATA.length
     );
   });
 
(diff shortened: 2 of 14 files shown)


packages/react/src/components/Footer/__tests__/LegalNav.test.js
```diff
@@ -37,7 +37,7 @@ describe('<LegalNav />', () => {
 
     expect(legalNav.exists('.bx--legal-nav')).toBeTruthy();
     expect(legalNav.find('.bx--legal-nav__list-item')).toHaveLength(
-      FOOTER_NAV_DATA.length - 1
+      FOOTER_NAV_DATA.length
```
photodow: I understand why you removed the `- 1` here, but I wonder if we should consider adding it back. The reason it was there is that this mock data remains to have the same amount of items, but because the data is malformed the component doesn't render the malformed item leaving it within this scenario one less than the `data.length`.

packages/react/src/components/Footer/__tests__/LegalNav.test.js
```diff
@@ -57,7 +57,7 @@ describe('<LegalNav />', () => {
 
     expect(legalNav.exists('.bx--legal-nav')).toBeTruthy();
     expect(legalNav.find('.bx--legal-nav__list-item')).toHaveLength(
-      FOOTER_NAV_DATA.length
+      FOOTER_NAV_DATA.length + 1
```
photodow: Merely a suggestion, but we might set this `1` to a `const` variable to help remove some of its magic, and help realize why its there?

In the PR I was working on I did the following.

```javascript
const ddsPrivacyCP = 1; // we're adding an additional LI to legalNav

...

expect(legalNav.find('.bx--legal-nav__list-item')).toHaveLength(FOOTER_NAV_DATA.length + ddsPrivacyCP);
```

packages/react/src/components/Footer/FooterNav.js
```diff
@@ -45,13 +45,26 @@ function renderGroups(groups) {
...
+    })
+  ),
+};
+
+/**
+ * @property defaultProps
+ * @type {{groups: Array}}
+ */
+FooterNav.defaultProps = {
+  groups: null,
```
photodow: Are these `null` defaults doing something? It's `undefined` by default. I think that would have the same outcome, unless I'm missing something?

jeffchew: @photodow if a property is not required, linting requires to add in default prop values. We can either go with null, or an empty array. I think either achieves the same thing. But this to get rid of the console errors from tests since it was expecting these fields but was throwing an error when nothing was passed.

cc: @kennylam @annawen1

photodow: Got it, is it necessary to remove `isRequired` then? They are required. If you don't include it the component won't render.

packages/react/src/components/Footer/__tests__/LegalNav.test.js
jeffchew: Perhaps just a comment would suffice?

packages/react/src/components/Footer/FooterNav.js
jeffchew: Your tests were throwing console errors though when you don't. Perhaps adjusting the tests then to not throw the error but expect the error to happen?

packages/react/src/components/Footer/__tests__/LegalNav.test.js
photodow: Perhaps, but isn't it also used multiple times?

packages/react/src/components/Footer/FooterNav.js
photodow: But per react it's supposed to throw those warnings, right? They were throwing those warnings because there are unit tests to test what happens when the props aren't included. Maybe we could use a method similar to #129 but more than just suppressing it we can see how many times it's being executed as well?

jeffchew: That's what I was thinking, where we spy on console errors and would expect the error to throw if a user doesn't pass in the prop. Then we can set things back to required because the test is essentially doing its job.

PR #130, opened 2019-09-25
feat(analytics): hover event triggers for masthead

annawen1:
### Related Ticket(s)

[Analytics: Add event triggers for Masthead #1601](https://github.ibm.com/webstandards/digital-design/issues/1601)

### Description

set up hover event triggers for the masthead items.

You can check if the events are being fired via the networks tab in chrome developer tools:
https://pages.github.ibm.com/digital-behavior/docs/tools-google-developer-tools.html

### Changelog

**New**

- set up analytics utility to check for global `ibmStats` object
- wrap `HeaderMenu` component to detect hover event

**Changed**

- moved the analytics script to preview-head.html to inject into iframe instead of manager-head.html


packages/react/src/components/Masthead/MastheadProfile.js
```diff
@@ -47,6 +48,16 @@ const MastheadProfile = ({
     <HeaderGlobalAction
       aria-label="User Profile"
       data-autoid={`${prefix}--masthead__profile`}
+      onMouseEnter={() =>
```
photodow: We really want to fire an event like this every time we hover? If that's true should we consider adding a delay before triggering it again to avoid a number of hits and impact on performance? I guess I assumed we would want to track clicks.

packages/react/src/components/Masthead/MastheadTopNav.js
```diff
@@ -21,36 +25,62 @@ import {
...
+   * @param {string} type type of element
+   * @returns {object} event fields
+   */
+  const menuItemEventData = (title, key, type) => ({
+    type: 'element',
+    primaryCategory: 'MASTHEAD',
+    eventName: 'HOVER',
+    executionPath: `masthead__l0-nav--nav-${key}`,
+    execPathReturnCode: type,
+    targetTitle: title,
```
photodow: Do we need any conditionals/validation for some of these values in case the information is corrupt?

packages/styles/scss/components/masthead/_masthead.scss
```diff
@@ -232,6 +232,10 @@ $search-transition-timing: 95ms;
       outline: none;
     }
   }
+
+  .#{$prefix}--masthead__nav-item li {
+    height: 100%;
```
photodow: Trying to understand why this was added here?

packages/react/src/components/Masthead/MastheadTopNav.js
```diff
@@ -21,36 +25,62 @@ import {
...
-            return (
-              <HeaderMenuItem
-                href={item.url}
-                data-autoid={`masthead__l0-nav--subnav-${j}`}>
-                {item.title}
-              </HeaderMenuItem>
-            );
-          })}
-        </HeaderMenu>
+        <div
```
photodow: I think this is creating some invalid HTML.

```html
<ul class="bx--header__menu-bar">
  ...
  <div class="bx--masthead__nav-item">
    <li class="bx--header__submenu">
      ...
    </li>
  </div>
  ...
</ul>

File: packages/react/src/components/Masthead/MastheadTopNav.js [hunk 5]
```diff
@@ -21,36 +25,62 @@ import {
...
+          className={`${prefix}--masthead__nav-item`}
+          onMouseEnter={() =>
+            analytics(menuItemEventData(link.title, i, 'dropdown'))
+          }>
+          <HeaderMenu
+            onMouseEnter={() => analytics(menuItemEventData(link.title, i))}
```
photodow: Is there a reason to have two `onMouseEnter` right after each other here like this?

packages/react/src/components/Masthead/MastheadTopNav.js
```diff
@@ -21,36 +25,62 @@ import {
...
+            aria-label={link.title}
+            menuLinkName={link.title}
+            data-autoid={`masthead__l0-nav--nav-${i}`}>
+            {link.menuSections[0].menuItems[
+              i
+            ].megapanelContent.quickLinks.links.map((item, j) => {
+              return (
+                <HeaderMenuItem
+                  href={item.url}
+                  data-autoid={`masthead__l0-nav--subnav-${j}`}>
```
photodow: Did you mean to add one of the above ones here instead?

packages/styles/scss/components/masthead/_masthead.scss
photodow: If I'm understanding correctly, think if you fix the invalid HTML above you won't need this.

packages/react/src/components/Masthead/MastheadProfile.js
photodow: Looking at the documentation a little closer I think they it says hover because the old masthead worked using hover. I would imagine keeping with the same heart/vein because the dropdown interacts via click we would want to use clicks here instead.

cc: @jeffchew @kennylam

packages/utilities/src/utilities/analytics/analytics.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * Helper function to fire a stats/metrics event for an action
+ *
+ * @param {object} eventData Object with standard IBM metric event properties and values to send to Coremetrics
+ * @returns {object} JSX object
+ *
+ */
+function analytics(eventData) {
```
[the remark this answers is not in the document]
jeffchew: I think under services might make more sense. Also, we might need to queue these event triggers if the object doesn't exist yet.

packages/react/src/components/Masthead/MastheadProfile.js
annawen1: yeah it's a bit much to do on hover - a delay might be a good idea.

I talked to @kennylam about it yesterday the dropdown. Currently the carbon masthead dropdowns only appears on click but we're going to create another component or wrapper so that it appears on hover

packages/react/src/components/Masthead/MastheadTopNav.js
annawen1: whoops yep - removing that one

packages/react/src/components/Masthead/MastheadProfile.js
photodow: Taking that information into account I can see an argument for keeping hover around until that comes but seeing that we'll probably release 1.0.0 before that happens I think maybe we should switch it to what's actually happening, and then we could change it when we get the hover states in? Thoughts?

packages/utilities/src/utilities/analytics/analytics.js
```diff
@@ -0,0 +1,14 @@
...
+  if (window.ibmStats) {
```
photodow: Should we limit these triggers to `PRODUCTION` only?

packages/react/src/components/Masthead/MastheadProfile.js
jeffchew: I'm fine with switching to click for now, we just need to let Juan know.
cc: @kennylam @ljcarot @photodow @annawen1

annawen1: ok, I'll have to add some event handlers for keyboard events as well

annawen1:
Closing this for now until hover is implemented in the masthead dropdowns

PR #132, opened 2019-09-26
Feat/global settings

kennylam:
### Related Ticket(s)

https://github.ibm.com/webstandards/digital-design/issues/1700

### Description

This adds global settings to the `@carbon/ibmdotcom-utilities` package.

### Changelog

**New**

- Global settings (includes `dds` prefix).

**Changed**

- Component `autoid` updated to use new `dds` prefix
- Component README updated with new `autoid` values

### DIFF
--- a/packages/react/src/components/DotcomShell/DotcomShell.js
+++ b/packages/react/src/components/DotcomShell/DotcomShell.js
@@ -7,11 +7,13 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
+import { settings } from '@carbon/ibmdotcom-utilities';
 import { settings as carbonSettings } from 'carbon-components';
-import { Footer } from '@carbon/ibmdotcom-react';
-import Masthead from '../Masthead/Masthead';
+import { Masthead, Footer } from '@carbon/ibmdotcom-react';
 
+const { prefix } = settings;
 const cPrefix = carbonSettings.prefix;
+
 /**
  * DotcomShell component
  *
@@ -28,8 +30,14 @@ const DotcomShell = ({
   return (
     <>
       <Masthead navigation={navigation} {...mastheadProps} />
-      <div className={`${cPrefix}--dotcom-shell`}>
-        <div className={`${cPrefix}--dotcom-shell__content`}>{children}</div>
+      <div
+        data-autoid={`${prefix}--dotcom-shell`}
+        className={`${cPrefix}--dotcom-shell`}>
+        <div
+          data-autoid={`${prefix}--dotcom-shell__content`}
+          className={`${cPrefix}--dotcom-shell__content`}>
+          {children}
+        </div>
       </div>
       <Footer type={footerType} />
     </>
--- a/packages/react/src/components/DotcomShell/DotcomShell.js
+++ b/packages/react/src/components/DotcomShell/DotcomShell.js
@@ -7,12 +7,12 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
-import { settings } from '@carbon/ibmdotcom-utilities';
-import { settings as carbonSettings } from 'carbon-components';
+import { settings as ddsSettings } from '@carbon/ibmdotcom-utilities';
+import { settings } from 'carbon-components';
 import { Masthead, Footer } from '@carbon/ibmdotcom-react';
 
+const { stablePrefix } = ddsSettings;
 const { prefix } = settings;
-const cPrefix = carbonSettings.prefix;
 
 /**
  * DotcomShell component
@@ -31,11 +31,11 @@ const DotcomShell = ({
     <>
       <Masthead navigation={navigation} {...mastheadProps} />
       <div
-        data-autoid={`${prefix}--dotcom-shell`}
-        className={`${cPrefix}--dotcom-shell`}>
+        data-autoid={`${stablePrefix}--dotcom-shell`}
+        className={`${prefix}--dotcom-shell`}>
         <div
-          data-autoid={`${prefix}--dotcom-shell__content`}
-          className={`${cPrefix}--dotcom-shell__content`}>
+          data-autoid={`${stablePrefix}--dotcom-shell__content`}
+          className={`${prefix}--dotcom-shell__content`}>
           {children}
         </div>
       </div>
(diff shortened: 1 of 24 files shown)


packages/react/src/components/Footer/LegalNav.js
```diff
@@ -20,8 +22,8 @@ const { prefix } = settings;
 const renderTrusteItem = () => {
   return (
     <li
-      className={`${prefix}--legal-nav__list-item`}
-      data-autoid="dds-privacy-cp"
+      className={`${cPrefix}--legal-nav__list-item`}
+      data-autoid={`${prefix}--dds-privacy-cp`}
```
photodow: `dds--dds-privacy-cp` is this what we want it to be? My guess the original `dds-` here was intended to be the updated prefix.

packages/react/src/components/HorizontalRule/HorizontalRule.js
```diff
@@ -34,9 +36,9 @@ function hrMod(mod) {
  */
 const HorizontalRule = ({ style, size, contrast, weight }) => (
   <hr
-    data-autoid="hr"
+    data-autoid={`${prefix}--hr`}
     className={classnames(
-      `${prefix}--hr`,
+      `${cPrefix}--hr`,
       hrMod(style),
```
photodow: I think `hrMod` needs to be updated?

packages/react/src/components/Masthead/MastheadSearchSuggestion.js
```diff
@@ -47,7 +50,7 @@ const MastheadSearchSuggestion = ({
       className={classNames('container-class', {
         ['container-highlight-class']: isHighlighted,
```
photodow: Should these have prefixes, or are they something external?

packages/react/src/components/DotcomShell/DotcomShell.js
```diff
@@ -7,10 +7,12 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
-import { settings } from 'carbon-components';
+import { settings } from '@carbon/ibmdotcom-utilities';
+import { settings as carbonSettings } from 'carbon-components';
 import { Masthead, Footer } from '@carbon/ibmdotcom-react';
 
 const { prefix } = settings;
+const cPrefix = carbonSettings.prefix;
```
photodow: Again, my destructuring syntax knowledge is a little rusty, but could you do this?

```javascript
const { prefix: cPrefix} = carbonSettings;
```

packages/react/src/components/Footer/Footer.js
```diff
@@ -16,6 +17,7 @@ import LegalNav from './LegalNav';
 import LocaleButton from './LocaleButton';
 
 const { prefix } = settings;
+const cPrefix = carbonSettings.prefix;
```
photodow: My thoughts from Slack. Do with it what you will.

> @kennylam I would almost rather keep `prefix = bx` since that’s what Carbon uses, and I would rather the stable selector’s prefix be something different like `stablePrefix = dds`. Maybe the same with settings `{ settings as ddsSettings }`.

kennylam: @photodow I missed this. PR was practically up already 😭

photodow: I figured it was the case. No worries, sorry I commented so late.

packages/react/src/components/Masthead/MastheadSearchSuggestion.js
kennylam: This isn't something Juan is asking for yet in analytics so we should probably let him make requests as he needs.

packages/react/src/components/Footer/Footer.js
kennylam: Np. @jeffchew and @annawen1 both agree with your suggestion. I do too, so i'll get it in.

packages/react/src/components/Masthead/MastheadSearchSuggestion.js
photodow: Not sure I understand. Why is Juan dictating our `classNames`?

kennylam: He isn't. I mean we don't need to prefix them with stable selectors yet.

packages/react/src/components/DotcomShell/DotcomShell.js
kennylam: We can get rid of this actually, since there won't be a naming conflict now.

packages/react/src/components/Masthead/MastheadSearchSuggestion.js
photodow: I see the confusion. Sorry, I'm not talking stable selectors here `data-autoid`. My question was about the `className` on this line and the one above. `container-highlight-class` Is it supposed to be prefixed liked `#{prefix}--container-highlight-class`?

kennylam: Ah I got you. Yeah we can (should) prefix that.

packages/react/src/components/Footer/LegalNav.js
```diff
@@ -20,8 +22,8 @@ const { prefix } = settings;
 const renderTrusteItem = () => {
   return (
     <li
-      className={`${prefix}--legal-nav__list-item`}
-      data-autoid="dds-privacy-cp"
+      className={`${cPrefix}--legal-nav__list-item`}
+      data-autoid={`${prefix}--privacy-cp`}
```
jeffchew: @kennylam minor thing, let's keep consistent with the name `stablePrefix` here

packages/react/src/components/Masthead/MastheadSearchSuggestion.js
```diff
@@ -44,10 +49,10 @@ const MastheadSearchSuggestion = ({
 
   return (
     <div
-      className={classNames('container-class', {
-        ['container-highlight-class']: isHighlighted,
+      className={classNames(`${prefix}--container-class`, {
+        [`${prefix}--container-highlight-class`]: isHighlighted,
```
photodow: Does anyone know what `${prefix}--container-highlight-class` is doing? I didn't see a reference for it in the `SCSS`.

kennylam: @photodow Looks like this is a leftover from when the autosuggest was first being prototyped. It doesn't need to be there and all the HTML needed for this are already provided by the package. I'll make an issue to clean this up in the future.

PR #134, opened 2019-09-26
feat(analytics): scroll tracking

annawen1:
### Related Ticket(s)

[Analytics: Add event triggers for scroll tracking #1602](https://github.ibm.com/webstandards/digital-design/issues/1602)

### Description

- Add event trigger for scrolling. Fires every 400px interval (only at the deepest scroll depth - does not fire if user scrolls back up)

### Changelog

**New**

- add scroll event listener
- add `RegisterAnalyticsEvent` in `services` package

**Changed**

- move analytics script to the `preview-head.html` file so that the `ibmStats` object is exposed to the iframe

### DIFF
--- a/packages/react/.storybook/manager-head.html
+++ b/packages/react/.storybook/manager-head.html
@@ -1,20 +1,29 @@
-<meta name="keywords" content="IBM, design, system, Carbon, design system, Bluemix, styleguide, style, guide, components, library, pattern, kit, component, cloud, React, React.js">
-<meta name="description" content="A collection of IBM.com components implemented using React and Carbon Design System.">
+<meta
+  name="keywords"
+  content="IBM, design, system, Carbon, design system, Bluemix, styleguide, style, guide, components, library, pattern, kit, component, cloud, React, React.js"
+/>
+<meta
+  name="description"
+  content="A collection of IBM.com components implemented using React and Carbon Design System."
+/>
 
 <!-- Open Graph -->
-<meta property="og:title" content="IBM.com Library React">
-<meta property="og:site_name" content="IBM.com Library React">
-<meta property="og:description" content="A collection of IBM.com components implemented using React and Carbon Design System.">
-<meta property="og:image" content="https://media.github.ibm.com/user/525/files/59e3bfde-b990-11e7-87ef-072e89a87719">
-<meta property="og:url" content="https://ibmdotcom-react.netlify.com">
+<meta property="og:title" content="IBM.com Library React" />
+<meta property="og:site_name" content="IBM.com Library React" />
+<meta
+  property="og:description"
+  content="A collection of IBM.com components implemented using React and Carbon Design System."
+/>
+<meta
+  property="og:image"
+  content="https://media.github.ibm.com/user/525/files/59e3bfde-b990-11e7-87ef-072e89a87719"
+/>
+<meta property="og:url" content="https://ibmdotcom-react.netlify.com" />
 
 <!-- Social -->
-<meta name="twitter:card" content="summary_large_image">
-<meta name="twitter:image:alt" content="IBM.com Library">
-<meta name="twitter:site" content="@_IBM">
-
-<!-- IBM Tag Management and Site Analytics -->
-<script src="//1.www.s81c.com/common/stats/ibm-common.js"></script>
+<meta name="twitter:card" content="summary_large_image" />
+<meta name="twitter:image:alt" content="IBM.com Library" />
+<meta name="twitter:site" content="@_IBM" />
 
 <!-- Storybook override -->
 <script>
--- a/packages/react/src/components/DotcomShell/DotcomShell.js
+++ b/packages/react/src/components/DotcomShell/DotcomShell.js
@@ -5,12 +5,12 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React, { useEffect } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import { settings as ddsSettings } from '@carbon/ibmdotcom-utilities';
 import { settings } from 'carbon-components';
 import { Masthead, Footer } from '@carbon/ibmdotcom-react';
-import globalMethods from '../../global';
+import '../../global';
 
 const { stablePrefix } = ddsSettings;
 const { prefix } = settings;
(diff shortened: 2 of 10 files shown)


packages/react/src/components/Masthead/Masthead.js
```diff
@@ -103,8 +107,27 @@ const Masthead = ({ navigation, ...mastheadProps }) => {
         );
       }
     });
+
```
jeffchew: @annawen1 as discussed, we should move this out into an `AnalyticsAPI`, which will have the `registerEvent` method, as well as a `initScrollTracker` method.

packages/services/src/services/Analytics/Analytics.js
```diff
@@ -0,0 +1,56 @@
...
+ * ibm.com
+ */
+class AnalyticsAPI {
+  /**
+   * service to fire a stats/metrics event for an action
+   *
+   * @param {object} eventData Object with standard IBM metric event properties and values to send to Coremetrics
+   * @returns {object} JSX object
+   * 
+   * example of eventData object expected:
```
jeffchew: @annawen1 can this be placed in an `@example`? How does this generate in JSDocs?

packages/react/src/components/Masthead/Masthead.js
annawen1: done

packages/react/.storybook/manager-head.html
```diff
@@ -1,20 +1,29 @@
-<meta name="keywords" content="IBM, design, system, Carbon, design system, Bluemix, styleguide, style, guide, components, library, pattern, kit, component, cloud, React, React.js">
-<meta name="description" content="A collection of IBM.com components implemented using React and Carbon Design System.">
+<meta
+  name="keywords"
+  content="IBM, design, system, Carbon, design system, Bluemix, styleguide, style, guide, components, library, pattern, kit, component, cloud, React, React.js"
+/>
+<meta
+  name="description"
+  content="A collection of IBM.com components implemented using React and Carbon Design System."
```
photodow: Is there anything we need to look at in here, or is this just formatting mostly?

packages/services/src/services/Analytics/Analytics.js
```diff
@@ -0,0 +1,56 @@
...
+   * method scroll tracking
+   * Only fires the event at 400px intervals and at the deepest scroll depth
+   *
+   **/
+  static initScrollTracker() {
+    let maxScrollDepth = 0;
+    const scrollAnalytics = root.addEventListener('scroll', () => {
+      let scrollDepth = root.pageYOffset;
+
+      if (scrollDepth % 400 === 0 && scrollDepth > maxScrollDepth) {
```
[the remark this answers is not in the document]
jeffchew: I'd mark this as resolved for this PR for now. If there's any changes needed we can address in another PR.

packages/services/src/services/Analytics/Analytics.js
```diff
@@ -0,0 +1,56 @@
...
+        maxScrollDepth = scrollDepth;
+        this.registerEvent({
+          type: 'element',
+          primaryCategory: 'SCROLL DISTANCE',
+          eventName: scrollDepth,
+          executionPath: root.innerWidth,
```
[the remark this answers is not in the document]
annawen1: oh I spoke with Juan EOD yesterday and he said we can get rid of the "grid" and just have the numeric value of the window width

packages/react/.storybook/manager-head.html
annawen1: yep it was just my prettier that reformatted the file after I removed the analytics script from here.

packages/react/src/components/DotcomShell/DotcomShell.js
```diff
@@ -25,6 +26,10 @@ const DotcomShell = ({
   children,
   ...mastheadProps
 }) => {
+  useEffect(() => {
+    AnalyticsAPI.initScrollTracker();
```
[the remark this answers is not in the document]
annawen1: I think for this specific scroll tracking event we discussed only placing it on the `DotcomShell`. If adopters didn't want to use the shell and wanted to pull components a la carte with scroll tracking, they could apply our analytics service themselves. Not sure if every component needs to depend on this analytics service.

packages/services/src/services/Analytics/Analytics.js
annawen1: done

packages/react/src/components/DotcomShell/DotcomShell.js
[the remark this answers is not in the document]
photodow: I guess I don't understand what differentiates the DotcomShell and makes it so special that we only want to track scroll event analytics if that's included on a page vs. from any other component. I thought if you're building an IBM.com page these analytics are required, maybe their not though? If we included a `global.js` for events like these then maybe it's not included in every component, but at least it's one place for users to add all the required things to be on IBM.com

packages/services/src/services/Analytics/Analytics.js
```diff
@@ -0,0 +1,78 @@
+import root from 'window-or-global';
+
+/**
+ * @constant {boolean} scrollTracker determines whether scroll tracking analytics is enabled
+ * @private
+ */
+const scrollTracker = process.env.SCROLL_TRACKING === 'true' || false;
```
photodow: I'm not sure I understand why this was added. Why are we allowing them to opt into something like this with an extra step at that? I feel like this is adding complexity to the adopter's UX unnecessarily. I also feel like scroll tracking is something we want on every page, and leaving this door open for every adopter to opt-in doesn't sound like a great strategy for getting that data. Maybe I'm not aligned on the intent here though?

annawen1: This was something Juan wanted as well - 

Juan: "that will only be enabled on-demand, meaning it is off by default and will be enabled with an option or config function"

packages/react/src/components/DotcomShell/DotcomShell.js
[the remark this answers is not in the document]
jeffchew: I think global.js makes sense, or whatever we decide to call it. I just added a `DDOAPI` that needs to get the version number in the DDO, which is an ask from Juan Leon.

packages/services/src/services/Analytics/Analytics.js
photodow: Ok, cool. I feel like this hurts the quality of our analytics, but ok. It is what it is I guess. I feel like Juan should be doing this work. 😉haha

PR #137, opened 2019-09-27
feat(ddo): adding ddo service

jeffchew:
### Related Ticket(s)

https://github.ibm.com/webstandards/digital-design/issues/1747

### Description

This adds the DDO API service, which includes an overall check if the DDO service is ready, plus a couple of methods:

- setVersion: sets the version of the IBM.com Library based on the global settings
- getAll: gets the full DDO
- getLanguage: gets the language set in the DDO

### Changelog

**New**

- DDO API


packages/utilities/src/utilities/settings/settings.js
```diff
@@ -8,6 +8,7 @@
  *
  */
 const settings = {
+  version: 'dds.v0.5.0',
```
photodow: Just a thought, but could we grab the versions of the packages themselves?

```javascript
const settings = {
   versions: {},
   stablePrefix: 'dds'
}

setPackageVersion('@carbon/ibmdotcom-react');
setPackageVersion('@carbon/ibmdotcom-styles');
setPackageVersion('@carbon/ibmdotcom-services');
setPackageVersion('@carbon/ibmdotcom-utilities');

function setPackageVersion (packageName) {
  try () {
    settings.versions[packageName] = require(`#{packageName}/package.json`).version;
  catch { /* fail silently */ }
}
```

packages/services/src/services/DDO/DDO.js
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  /**
+   * Sets the version of the library to the DDO
+   *
+   * @returns {Promise<any>} Promise object
+   */
+  static async setVersion() {
+    return await this.isReady().then(() => {
+      root.digitalData.page.version = ddsSettings.version;
```
photodow: Should we emit an event to say the DDO object change?

packages/services/src/services/DDO/DDO.js
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  /**
+   * Gets the full digitalData (DDO) object
+   *
+   * @returns {Promise<*>} Promise object
+   */
+  static async getAll() {
+    return await this.isReady().then(() => {
+      return root.digitalData;
```
photodow: Is it possible our versions won't be included in the first this is called?

packages/services/src/services/DDO/DDO.js
```diff
@@ -0,0 +1,62 @@
...
+      }
+    });
+  }
+
+  /**
+   * Gets the full digitalData (DDO) object
+   *
+   * @returns {Promise<*>} Promise object
+   */
+  static async getAll() {
```
photodow: Should adopters be setting the DDO using this service moving forward?

packages/utilities/src/utilities/settings/settings.js
jeffchew: Since we're on independent versioning, this probably won't work after awhile. We will need to be able to grab the global package version somehow.

packages/services/src/services/DDO/DDO.js
jeffchew: we should probably set up setters and getters for them for specific items I think. But if they need to set something specific, they can always use our `isReady` wrapper before they do it.

packages/services/src/services/DDO/DDO.js
jeffchew: not sure if this is necessary...

Issue #126, opened 2019-09-24
Masthead L1: dropdown


oliviaflory:
The sub-links inside the dropdowns are sitting a few pixels too high in the L1 dropdowns. The sub-link should be vertically centered within each option.

<img width="808" alt="Screen Shot 2019-09-24 at 2 32 44 PM" src="https://user-images.githubusercontent.com/45692622/65543811-f15d2f00-dedf-11e9-9177-e3f6796c2659.png">

Issue #125, opened 2019-09-24
Masthead L0: Focus state below rule


oliviaflory:
The focus state is appearing below the bottom rule on all the L0 links, causing the focus state to look uneven with only 1px of blue at the bottom.

Can the focus appear above the rule? 

<img width="755" alt="Screen Shot 2019-09-24 at 3 14 07 PM" src="https://user-images.githubusercontent.com/45692622/65543094-72b3c200-dede-11e9-84cf-cb1579d9706d.png">
<img width="207" alt="Screen Shot 2019-09-24 at 2 32 13 PM" src="https://user-images.githubusercontent.com/45692622/65543104-75aeb280-dede-11e9-8369-0bc64a854223.png">

Issue #124, opened 2019-09-24
Masthead L0: Profile dropdown hidden behind L1


oliviaflory:
The Profile / My IBM dropdown menu is hidden behind L1 when it's open

<img width="207" alt="Screen Shot 2019-09-24 at 2 32 13 PM" src="https://user-images.githubusercontent.com/45692622/65540018-91af5580-ded8-11e9-92cd-a044211cd20b.png">

Issue #119, opened 2019-09-23
Masthead: dropdown box-shadow needs adjustment


larahanlon2:
The `box-shadow` on all dropdown menus for L0 and L1 has too much spread and is slightly too dark. Can we change to these values, please?
<img width="1659" alt="Screen Shot 2019-09-23 at 2 09 33 PM" src="https://user-images.githubusercontent.com/53093550/65451173-9eff0e00-de0c-11e9-8bca-f2284d320242.png">

PR #138, opened 2019-09-27
Fix/masthead design qa

kennylam:
### Related Ticket(s)
#104, #107, #108, #109, #110, #111, #112, #113, #114, #115, #116, #118, #119, #124, #125, #126, #152 

### Description
This fixes the masthead submenus not toggling visibility. Menus will now toggle visibility correctly, with only one expanded at a time.
Oct-03-2019 06-49-42

The masthead platform name now has correct hover/focus states, as well as layout fixes for when it does and does not appear.
Screen Shot 2019-10-03 at 8 56 55 AM

Various CSS fixes to the masthead component based on design review.

### Changelog

**Changed**
- masthead subnavs now toggle correctly
- masthead platform name hover/focus states fixed; layout fixes
- Lots of css fixes


jeffchew:
@kennylam can you retarget this PR to the release branch?

kennylam:
> Everything is almost perfect! Except for 1 issue.
> 
> #152 when turning on the `platform` knob:
> 
> * The gray divider rule after the IBM 8-bar logo shifts 2–3px to the right (it should stay in the same place as without the platform name)
> * The `padding` is too great. It should be:

@larahanlon2 This is fixed.

Re: the ghost spacing, this is an issue with Storybook. It doesn't re-render the page for changes like that and so the injected element is left behind and hidden, introducing that gap. In real-world usage this would be impossible since there is no way to toggle the platform name. Once it's there it's there to stay.

Issue #101, opened 2019-09-21
Make masthead links, Profile icon optional


sanjitbauli:
In URX we don't need any menu on the header, also the profile icon is not required, In Northstar header, there was an option masthead:'popup' this feature will help us to match the same behavior.

Make the masthead configurable,  hide links, profile bypassing props to the component.

A duplicate of this issue has been created in the Digital Design board.  [Make the masthead configurable, hide links, profile bypassing props to the component](https://zenhub.ibm.com/app/workspaces/digital-design-system-5b3a815a8bf6a932efa6fd03/issues/webstandards/digital-design/1772)

ljcarot:
@jeffchew I believe we made masthead configurable.  Can you confirm?

sanjitbauli:
I don't see any option. Maybe i am missing the trick. FYI, i am using own css to hide the menu and profile section at this moment.

ljcarot:
It will be coming out in next release.  That's why you don't see it yet.

Issue #158, opened 2019-10-04
Masthead: L0 dropdown first item focus state 


oliviaflory:
The focus state of the first item in the L0 dropdowns is only showing 1px of the 2px border on the **top side** . Is it possible to fix this? I looks like the rule at the bottom of the L0 is on top of the focus state.

<img width="842" alt="Screen Shot 2019-10-03 at 9 12 45 PM" src="https://user-images.githubusercontent.com/45692622/66174469-21fd4100-e623-11e9-9ef6-1aa24143d09b.png">

photodow:
@RobertaJHahn Does this need to be closed? It looks like it got solved?

RobertaJHahn:
@photodow Thanks for noticing.  Closing.

PR #184, opened 2019-10-10
feat(masthead): makes search and profile optional

kennylam:
### Related Ticket(s)

#101 

### Description

This change allows the Profile and Search components in the masthead to be optional. All masthead components (except the IBM logo) are now optional.

### Changelog

**New**

- Search and profile are now optional

**Changed**

- `Masthead` and `DotcomShell` now share the same knobs for `Masthead` component


packages/react/src/components/Masthead/Masthead.js
```diff
@@ -138,51 +138,59 @@ const Masthead = ({ navigation, ...mastheadProps }) => {
...
-                {navigation !== false ? (
+                {navigation && (
                   <MastheadTopNav
                     {...mastheadProps}
                     navigation={mastheadData}
                   />
-                ) : null}
-                <MastheadSearch />
+                )}
+                {mastheadProps.hasSearch && <MastheadSearch />}
```
jeffchew: @kennylam I couldn't find defaultProps for these, these should be true by default right?

kennylam: @jeffchew Doh! Yes you are right 👍

packages/react/src/components/Masthead/Masthead.js
```diff
@@ -200,4 +208,13 @@ Masthead.propTypes = {
   mastheadProp: PropTypes.object,
 };
 
+/**
+ * @property defaultProps
+ * @type {{hasProfile: boolean, hasSearch: boolean}}
+ */
+Masthead.defaultProps = {
+  hasProfile: true,
+  hasSearch: true,
```
jeffchew: @kennylam propTypes need to be added too...

kennylam:
@oliviaflory This PR addresses #158 also.

PR #526, opened 2019-10-14
feat(locale): locale service

annawen1:
### Related Ticket(s)

[Locale Selector: develop a service that determines the users locale based on ipcinfo or Geo-ip #1817](https://zenhub.ibm.com/app/workspaces/digital-design-system-5b3a815a8bf6a932efa6fd03/issues/webstandards/digital-design/1817)

### Description

Set up a service that uses the stored ipcinfo or the stored geo-ip info plus browser language preference

If cookie did not exist, set the cookie once the locale info has been retrieved

### Changelog

**New**

- set up locale service

**Changed**

- edit the `ipcinfoCookie.set` util to take object as param
- edit the `geolocation` util to only retrieve the `cc` and leave the language to the locale service


packages/services/src/services/Locale/Locale.js
```diff
@@ -0,0 +1,143 @@
...
+   *
+   * @example
+   * import { LocaleAPI } from '@carbon/ibmdotcom-services';
+   *
+   * function async getLocale() {
+   *    const list = await LocaleAPI.getList();
+   * }
+   */
+  static async getList() {
+    const url = `${_proxy}${_endpoint}/usen-locale.json`;
```
jeffchew: @annawen1 shouldn't `getList` and `getLocale` not be hardcoded with `us` and `en`?

annawen1: hmm we only have the us-en country json file up, but also the list is used to verify the `cc-lc` so I don't want to use possibly non-existing `cc-lc` combo to make the call.

annawen1: since we have the redirect happening with `www.ibm.com/common/v18/js/data(([/\?].*)?$)` regex, I'm checking the response for now until we have a regex in place for the `locale` path

kennylam: This redirect has been approved, resolving

packages/services/src/services/Locale/Locale.js
```diff
@@ -0,0 +1,143 @@
...
+   * @param {string} cc country code
+   * @param {string} lc language code
+   * @param {object} list country list
+   *
+   * @returns {object} object with lc and cc
+   * @example
+   * import { LocaleAPI } from '@carbon/ibmdotcom-services';
+   *
+   * async function getLocale() {
+   *   const cc = 'us';
```
jeffchew: @annawen1 can you change this from being hardcoded?

annawen1: sure

Issue #163, opened 2019-10-04
Develop link + icon variant 


oliviaflory:
#### User Story

> As a :

Web simplification designer

> I need:

a variation of the link component

> so that I can:

finalize my designs.

#### Additional information

The web simplification team has created a design that utilizes a link and icon. This is a very common instance on current IBM.com and we need to develop a variation of the link component so it can be standardized.

The web simplification team will be coming back with logic around which icons can be paired in the `link + icon` variant within their pattern. This logic / guideance may or may not apply to the component within the ibm.com library

#### Specs

![Primary_Link_Text-and-icon](https://user-images.githubusercontent.com/45692622/66240163-d6ea3900-e6c9-11e9-947f-6bd3e14badd6.png)

**Note**
This is a variation of the `primary link` component and we **should not** provide a `link + icon` variant of `inline links`.

#### Acceptance criteria
 - [ ] component variation follows Carbon color tokens defined in primary link
 - [ ] component passes accessibility standards on all color themes
 - [ ] testing is done to determine what happens if a link+icon wraps to 2 lines

PR #532, opened 2019-10-16
Feat/link with icon

kennylam:
### Related Ticket(s)

#163 

### Description

Carbon's [link](https://www.carbondesignsystem.com/components/link/code) component with the ability to add an inline icon.

### Changelog

**New**

- `LinkWithIcon` component, tests and story


jeffchew:
> This is really nit picky, but the focus state isn't vertically centered around the link. The top margin is larger than the bottom margin, and ideally it would be equal.
> 
> Let me know if this isn't possible or if it's because of the line height. I would have thought the focus outline would be 22px high to match the line height

Can we merge as is since @kennylam isn't in, then we can report an issue for this?

jeffchew:
Speaking with @oliviaflory , we will merge as-is, then create an issue based on the comment above.

PR #545, opened 2019-10-18
feat(buttongroup): adding ButtonGroup

jeffchew:
### Related Ticket(s)

No related issue

### Description

This creates the generic ButtonGroup that is derived from the `LeadSpace` pattern.

Preview available at:
https://deploy-preview-545--ibmdotcom-react-experimental.netlify.com

NOTE: The Button Group appears to be only legible with a dark background, we probably will need to include a prop to switch between the two.

Feature flag created for this: `BUTTON_GROUP`

### Changelog

**New**

- ButtonGroup component


packages/styles/scss/components/buttongroup/_buttongroup.scss
```diff
@@ -0,0 +1,46 @@
...
+@import '../../globals/imports';
+
+@mixin buttongroup {
+  $btn-min-width: 26;
+
+  .#{$prefix}--buttongroup {
+    @include carbon--make-row;
+
+    &__ctas {
+      @include carbon--make-col-ready;
```
photodow: Do we need to include layout SCSS mixins like this and row above if it's just a full width? If we remove these adopters can put it wherever they need it in their page's layout.

packages/react/src/components/ButtonGroup/__stories__/ButtonGroup.stories.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { withKnobs, object } from '@storybook/addon-knobs';
+import '../../../../../styles/scss/components/buttongroup/_buttongroup.scss';
+import 'carbon-components/scss/globals/grid/_grid.scss';
```
photodow: I see you're pulling in the grid, but it doesn't appear as if you're using it.

packages/react/src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -0,0 +1,106 @@
...
+  }, []);
+
+  return featureFlag(
+    BUTTON_GROUP,
+    <div
+      className={`${prefix}--buttongroup`}
+      data-autoid={`${stablePrefix}--buttongroup`}>
+      <div className={`${prefix}--buttongroup__ctas`}>
+        {buttons.map((button, key) => {
+          if (key > 1) return;
```
photodow: Does it make sense to pull this `1` up and give it a const value to help define what's happening here a little better?

```javascript
const BUTTON_COUNT_ALLOWED = 1; // count starts with 0

... 

if (key > BUTTON_COUNT_ALLOWED) return;

...

File: packages/react/src/components/ButtonGroup/__stories__/ButtonGroup.stories.js (hunk 2 above)
Comment (jeffchew) [reply packages/react/src/components/ButtonGroup/__stories__/ButtonGroup.stories.js]: Good catch, I pulled this from another story and didn't remove that.

File: packages/styles/scss/components/buttongroup/_buttongroup.scss (hunk 1 above)
Comment (jeffchew) [reply packages/styles/scss/components/buttongroup/_buttongroup.scss]: Took a closer look, makes sense. I removed the outer layer and just made it 100% width.
cc: @annawen1 since this would affect your pattern if you're bringing this back in.

File: packages/react/src/components/ButtonGroup/ButtonGroup.js (hunk 3 above)
Comment (jeffchew) [reply packages/react/src/components/ButtonGroup/ButtonGroup.js]: Bringing @wonilsuhibm @larahanlon2 @oliviaflory here, and cc: @annawen1. Is there a requirement to cap the number of allowed buttons? I'm wondering if we should remove entirely?

cc: @photodow

**Comment by jeffchew** [comment]
> 1. I created an issue for the button states on the darker themes:  [carbon-design-system/carbon#4404](https://github.com/carbon-design-system/carbon/issues/4404)

Thanks @photodow .

> 2. Looks like the theme for the button is being set to `g90` via the locale-button (no feature flag in scss currently) in the footer due to `import-once`. https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/styles/scss/components/footer/index.scss
>    
>    * This opens up a larger conversation that maybe we should have during our huddle next week. In how we want to handle defaults, import-once, component mixin, color theming. By default you import a stylesheet for a component, and the mixin is executed for whatever the default theme is unless otherwise specified. To get another theme you have to import the mixin surrounded by the theming mixin. Just in generally it's getting hard to manage, and follow. It might be simplified if we could switch to CSS Custom Properties.

Can you add as a huddle topic?

> 3. Do we need to replace the current button group within the leadspace pattern, and remove that code?

Yes, but I think we can do as a separate PR. cc: @annawen1

File: packages/react/src/components/ButtonGroup/ButtonGroup.js (hunk 3 above)
Comment (annawen1) [reply packages/react/src/components/ButtonGroup/ButtonGroup.js]: when I initially wrote this up as part of the `<LeadSpace />` component, there was a max of 2 buttons we could have for leadspace specifically. But since we're making the buttons more generic we should probably remove that.

Comment (jeffchew) [reply packages/react/src/components/ButtonGroup/ButtonGroup.js]: @annawen1 it could be something you set in the `Leadspace` itself where you only pass in the first two buttons in the array into the `ButtonGroup`.

Comment (jeffchew) [reply packages/react/src/components/ButtonGroup/ButtonGroup.js]: @photodow @annawen1 I removed the cap, and also I removed the JS code for adjusting widths and replaced with a pure flexbox solution.
cc: @wonilsuhibm @larahanlon2 @oliviaflory

File: packages/react/src/components/ButtonGroup/ButtonGroup.js [hunk 4]
```diff
@@ -0,0 +1,76 @@
+/**
+ * Copyright IBM Corp. 2016, 2018
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+import React, { useRef, createRef } from 'react';
+import PropTypes from 'prop-types';
+import { ArrowRight20, ArrowDown20, Pdf20 } from '@carbon/icons-react';
```
photodow: After removing the count does it make sense to also remove the restrictions on the specific icons that can be used within this component, and move that out into the leadspace pattern?

packages/react/src/components/ButtonGroup/README.md
```diff
@@ -0,0 +1,77 @@
+# Button Group
+
+> The button group component is to be utilized within IBM.com for grouping two
```
photodow: I think we can remove "two" based on the latest commit.
```suggestion
> The button group component is to be utilized within IBM.com for grouping
```

packages/react/src/components/ButtonGroup/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+| Name      | Description                                                                                    |
+| --------- | ---------------------------------------------------------------------------------------------- |
+| `buttons` | Array of button objects to render. Each object should contain `link`, `copy`, and `renderIcon` |
+
+## Stable selectors
+
+| Name                        | Description                                 |
+| --------------------------- | ------------------------------------------- |
+| `dds--buttongroup`          | `ButtonGroup` wrapper layer                 |
```
photodow: This is nit picky thing, but does it make sense to add a dash to divide the two words and follow some of the other class naming we have going on `dotcom-shell`, `locale-button`. Just a thought. This alone shouldn't block it going in.

packages/react/src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -0,0 +1,76 @@
...
+ * @returns {*} button components
+ */
+const ButtonGroup = ({ buttons }) => {
+  const icons = {
+    ArrowDown: ArrowDown20,
+    ArrowRight: ArrowRight20,
+    Pdf: Pdf20,
+  };
+
+  const buttonsRef = useRef(buttons.map(() => createRef()));
```
photodow: Is `buttonRef` still necessary? It looks like you removed the resizing javascript that was leveraging this?

packages/react/src/components/ButtonGroup/ButtonGroup.js
jeffchew: I'm actually thinking of cleaning up the options here, because there is a `renderArrow` option here as well. 

I'm wondering if there is a restriction on icons for a reason, and perhaps we want to give them only a few options. Perhaps a question for @wonilsuhibm @oliviaflory @larahanlon2 .

cc: @annawen1

packages/react/src/components/ButtonGroup/ButtonGroup.js
jeffchew: This should be removed in the next commit.

packages/react/src/components/ButtonGroup/README.md
jeffchew: Updated language to say `two or more`

packages/react/src/components/ButtonGroup/ButtonGroup.js
jeffchew: After lengthy discussion, made this button group component more flexible, and allowing the parent pattern to do their own enforcement.

packages/react/src/components/ButtonGroup/README.md
jeffchew: Updated with the hyphen.

jeffchew:
@photodow @annawen1 made a number of changes, the button group is much more flexible now. I also made some edits to the story and README to accommodate.

cc: @wonilsuhibm @larahanlon2 @oliviaflory

packages/react/src/components/ButtonGroup/__stories__/ButtonGroup.stories.js
```diff
@@ -0,0 +1,68 @@
...
+          link: '',
+          copy: text('Primary button copy', 'Primary action button'),
+          renderIcon: primaryIcon || null,
+        },
+        {
+          link: '',
+          copy: text('Secondary button copy', 'Secondary action button'),
+          renderIcon: secondaryIcon || null,
+        },
+      ];
```
photodow: @jeffchew don't kill me for suggesting this, but it might be good if we could provide an option to add or remove buttons.

```javascript
const buttonCount = number('Number of buttons', 2);
const buttons = [];

for (let i = 0; i < buttonCount; i++) {
  buttons.push({
    link: text(`Link ${i+1}`, `https://example.com`),
    copy: text(`Button ${i+1}`, `Button ${i+1}`),
    renderIcon: iconMap[select(`Icon ${i+1}`, iconOptions, iconOptions['Arrow Right'])],
  })
}
```

packages/react/src/components/ButtonGroup/__stories__/ButtonGroup.stories.js
```diff
@@ -0,0 +1,68 @@
...
+
+      return (
+        <div
+          style={{
+            padding: 2 + `rem`,
+            display: 'flex',
+            alignItems: 'center',
+            flexDirection: 'column',
+            backgroundColor: 'black',
```
photodow: I think all you need here is.

```javascript
{
  padding: 2 + `rem`,
  backgroundColor: 'black'
}
```

packages/styles/scss/components/buttongroup/_buttongroup.scss
```diff
@@ -0,0 +1,42 @@
...
+
+    @include carbon--breakpoint(md) {
+      flex-flow: row-reverse wrap-reverse;
+      justify-content: flex-end;
+      width: 100%;
+
+      .#{$prefix}--btn {
+        width: auto;
+        margin-right: $layout-03;
+        flex-grow: 1;
```
photodow: So if I'm not mistaken the buttons will always be most times `width: 20rem;` due to the `Button` `max-width`. Is that expected?

jeffchew: @wonilsuhibm @oliviaflory @larahanlon2 @annawen1 would you know the answer to this?

packages/react/src/components/ButtonGroup/__stories__/ButtonGroup.stories.js
jeffchew: @wonilsuhibm @larahanlon2 @oliviaflory what say you? 😄

packages/react/src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -0,0 +1,62 @@
...
+    <div
+      className={`${prefix}--buttongroup`}
+      data-autoid={`${stablePrefix}--button-group`}>
+      {buttons.map((button, key) => {
+        return (
+          <Button
+            key={key}
+            data-autoid={`${stablePrefix}--button-group-${key}`}
+            renderIcon={button.renderIcon || null}
+            href={button.link}
```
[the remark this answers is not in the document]
jeffchew: I'm looking at the `Button` itself in carbon, and don't see it outwardly supporting onClick. I think however there is an `other` spread operator that gets passed in, which I'm assuming would allow for other props like onClick get passed in. While I'm debating if we should explicitly provide a prop for onClick, we should however think about if we want to support the other built-in props for `Button`, as well as also just passing `others` props through to the Button itself.

packages/react/src/components/ButtonGroup/__stories__/ButtonGroup.stories.js
jeffchew: Done!

packages/react/src/components/ButtonGroup/__stories__/ButtonGroup.stories.js
jeffchew: @photodow regardless of the decision, I don't think this is a deal breaker for this PR, especially with the fact this is still behind a feature flag and should suffice for your Modal story. We can decide this later perhaps.

packages/react/src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -0,0 +1,62 @@
...
+ * @param {Array.<{copy: string}>} props.buttons array of buttons
+ * @param {string} props.buttons[].copy Button copy
+ * @param {string} props.buttons[].link URL for the button item
+ * @param {object} props.buttons[].renderIcon Optional icon type
+ * @returns {*} button components
+ */
+const ButtonGroup = ({ buttons }) => {
+  return featureFlag(
+    BUTTON_GROUP,
+    <div
```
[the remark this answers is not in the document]
jeffchew: I am actually not sure, I know for sure that this helps in a navigation pattern but not sure if the button group counts as one?

@annawen1 @kennylam @andysherman2121 @sannminwin any thoughts on this?

packages/styles/scss/components/buttongroup/_buttongroup.scss
annawen1: that shouldn't be the case...it should only be be hit the `20rem max-width` when the text is too long

packages/react/src/components/ButtonGroup/ButtonGroup.js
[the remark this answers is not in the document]
annawen1: ehhh I don't mind either way

packages/styles/scss/components/buttongroup/_buttongroup.scss
jeffchew: Been playing around with the pure flex way of handling the width growth and it seems to be locked on 20rem. I'll have to play around more with this to see if I can get the behavior to work without resorting to the JS implementation.

packages/react/src/components/ButtonGroup/ButtonGroup.js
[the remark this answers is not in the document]
photodow: Here is my thinking...

1. For me, it's clearly a list of buttons/actions/options for the user to take. It means more to anyone/anything reading it when it's in context to a list instead of floating aimlessly within a random `div`.
2. We are applying a priority to the list by saying the first is primary, and from there the priority diminishes. Because of this, I feel like an ordered list is the right way to go. We can't just throw these in randomly and expect them to perform the same way.
3. It can also benefit accessibility because screen readers will announce to the user that there is a list and how many items are in this list when they reach it.

packages/react/src/components/ButtonGroup/ButtonGroup.js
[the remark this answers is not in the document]
photodow: Yeah, I think spread would allow for onClick and more. I was mistaken. It looks like Carbon's button is handling the validation of the `link`/`href` (which could better align here) so if you set `link` to undefined/null, or just don't provide it it'll turn into a button instead of an anchor which is what I was asking about here.

packages/react/src/components/ButtonGroup/__stories__/ButtonGroup.stories.js
jeffchew: Issue to track this: https://github.com/carbon-design-system/ibm-dotcom-library/issues/558

packages/react/src/components/ButtonGroup/ButtonGroup.js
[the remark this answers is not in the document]
jeffchew: It has been updated to use `<ol>`/`<li>'s`

packages/react/src/components/ButtonGroup/ButtonGroup.js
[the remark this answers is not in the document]
jeffchew: @photodow updated to allow in a spread to go through for each `Button`, also for the `link`. This should accommodate for any of the additional `Button` attributes that the user can pass through for each object in the array.

packages/styles/scss/components/buttongroup/_buttongroup.scss
jeffchew: Issue opened to track this: https://github.com/carbon-design-system/ibm-dotcom-library/issues/559

packages/react/src/components/ButtonGroup/README.md
```diff
@@ -0,0 +1,86 @@
...
+import ReactDOM from 'react-dom';
+import { ArrowRight20, ArrowDown20 } from '@carbon/icons-react';
+import { ButtonGroup } from '@carbon/ibmdotcom-react';
+import '@carbon/ibmdotcom-styles/scss/components/buttongroup/_buttongroup.scss';
+function App() {
+  return;
+  <ButtonGroup
+    buttons={[
+      {
+        link: '',
```
photodow: `href`?

packages/react/src/components/ButtonGroup/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+| Name      | Description                                                 |
+| --------- | ----------------------------------------------------------- |
+| `buttons` | Array of button objects to render. See `Button Item` below. |
+
+### Button Item
+
+| Name         | Description                           |
+| ------------ | ------------------------------------- |
+| `link`       | URL for the button item               |
```
photodow: Is this now `href`?

packages/react/src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -30,22 +30,24 @@ const { prefix } = settings;
...
-            key={key}
-            data-autoid={`${stablePrefix}--button-group-${key}`}
-            renderIcon={button.renderIcon || null}
-            href={button.link}
-            kind={key === 0 ? 'primary' : 'tertiary'}>
-            {button.copy}
-          </Button>
+          <li className={`${prefix}--buttongroup-item`}>
+            <Button
+              key={key}
```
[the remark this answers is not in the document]
jeffchew: Thanks, good catch. This should be fixed now.

packages/react/src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -0,0 +1,64 @@
...
+const { stablePrefix } = ddsSettings;
+const { prefix } = settings;
+
+/**
+ * renders the buttons
+ *
+ * @param {object} props props object
+ * @param {Array.<{copy: string}>} props.buttons array of buttons
+ * @param {string} props.buttons[].copy Button copy
+ * @param {string} props.buttons[].link URL for the button item
```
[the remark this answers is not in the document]
jeffchew: Yup, I changed to `href` here.

packages/react/src/components/ButtonGroup/README.md
jeffchew: Yup!

packages/react/src/components/ButtonGroup/__stories__/ButtonGroup.stories.js
```diff
@@ -0,0 +1,65 @@
...
+          select('Primary button icon', iconOptions, iconOptions['Arrow Right'])
+        ];
+      const secondaryIcon =
+        iconMap[
+          select('Secondary button icon', iconOptions, iconOptions['PDF'])
+        ];
+
+      const buttons = [
+        {
+          link: '',
```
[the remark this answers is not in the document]
jeffchew: Yup, changed. I also added a link to Carbon's documentation for all available props that can be passed in.

jeffchew:
> > 1. Do we need to replace the current button group within the leadspace pattern, and remove that code?
> 
> Yes, but I think we can do as a separate PR. cc: @annawen1

Issue created to track this: https://github.com/carbon-design-system/ibm-dotcom-library/issues/562

packages/react/src/components/ButtonGroup/__stories__/ButtonGroup.stories.js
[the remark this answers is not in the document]
photodow: @jeffchew I didn't see this one get updated.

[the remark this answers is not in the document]
jeffchew: Sorry, forgot to push that file.

Issue #224, opened 2019-10-13
Package | Components architecture (VanillaJS/HTML)


jeffchew:
_Jeff-Chew created the following on May 23:_  

 ### Overview

This is the creation of the package architecture for the Components (VanillaJS/HTML flavor). This will include all IBM.com related components. 

### Additional Information
- Should look into the current `carbon-components` to see what architecture can be lifted from there

### Acceptance criteria
- [ ] Components should reference the `styles` package for CSS styling
- [ ] Storybook HTML as an output, as it will serve as the living documentation for this package
- [ ] Appropriate lint/unit testing functionality
- [ ] Pre-commit/pre-push hooks
- [ ] Publish scripts to generate modular output
- [ ] README on setup and usage
 

_Original issue: https://github.ibm.com/webstandards/digital-design/issues/860_

jeffchew:
_On Jun 25, Jeff-Chew commented:_
@vladislav-saling @Kenny-Lam @james-dow I added in a story to look into Storybook for vanilla JS (https://github.ibm.com/webstandards/digital-design/issues/1120), which I think I'd prefer if it weren't for the Drupal team's ask for twig/json files. But we would also need a way to test the twig/json files themselves, which currently only see that support in pattern lab.

jeffchew:
_On Jun 25, james-dow commented:_
@Jeff-Chew what type of testing are you expecting? Does it go beyond unit testing, or are you looking for interaction/behavior basic QA testing?

jeffchew:
_On Jun 25, Jeff-Chew commented:_
@james-dow The latter. Testing meaning to make sure that the templates render properly. From what I can tell, storybook won't work with twig templates, but pure HTML as part of the stories. It only seems like pattern lab with the twig php engine would be able to render twig templates as part of the output, but I don't want to get tied down to using pattern lab because I honestly prefer Storybook. I'm torn because this is a must have request coming from the Drupal team.

cc: @Kenny-Lam @vladislav-saling @hahnrob @linda-carotenuto @Wonil-Suh1

jeffchew:
_On Jun 25, james-dow commented:_
Correct me if I'm wrong, but we're talking about two separate issues, right?
  1. There is the QA testing process, and then...
  2. Drupal's oddly specific requirements.

[...]

PR #548, opened 2019-10-21
feat(vanilla): added initial vanilla package architecture

jeffchew:
### Related Ticket(s)

#224, #553 

### Description

This adds the initial vanilla package architecture, with Storybook HTML, and the horizontal rule as an example. 

The approach used is ES6 template literals, up for discussion if we want to use another approach.

Deploy preview here:
https://5dae2d51b76cea0008611e0e--ibmdotcom-vanilla.netlify.com/?path=/story/horizontalrule--default

The new URL for the vanilla package storybook is:
https://ibmdotcom-vanilla.netlify.com (not working until this PR is merged to `master`)

### Changelog

**New**

- Vanilla package architecture
- Vanilla version of the horizontal rule


packages/vanilla/src/components/horizontalrule/horizontalrule.template.js
```diff
@@ -0,0 +1,24 @@
...
+const horizontalruleTemplate = ({ style, size, contrast, weight }) => {
+  const hrStyle = style === 'dashed' ? 'bx--hr--dashed' : '';
+  const hrSize = size && size !== '' ? `bx--hr--${size}` : '';
+  const hrContrast =
+    contrast && contrast !== '' ? `bx--hr--${contrast}-contrast` : '';
+  const hrWeight = weight === 'thick' ? 'bx--hr--thick' : '';
+
+  return `<hr
+    data-autoid="dds--hr"
+    class="bx--hr ${hrStyle} ${hrSize} ${hrContrast} ${hrWeight}"
```
photodow: Does it make sense to use the `classnames` package here?

```suggestion
    class="${classnames('bx--hr', hrStyle, hrSize, hrContrast, hrWeight)}"
```

packages/vanilla/src/components/horizontalrule/horizontalrule.template.js
```diff
@@ -0,0 +1,24 @@
...
+ * @param {object} props Property object passed into the template
+ * @param {string} props.style solid vs dashed [''|'dashed']
+ * @param {string} props.size size of the hr [''|'small'|'medium'|'large']
+ * @param {string} props.contrast color of the rule [''|'low'|'high']
+ * @param {string} props.weight weight of the rule [''|'thick']
+ * @returns {string} ES6 template literal of the horizontal rule
+ */
+const horizontalruleTemplate = ({ style, size, contrast, weight }) => {
+  const hrStyle = style === 'dashed' ? 'bx--hr--dashed' : '';
+  const hrSize = size && size !== '' ? `bx--hr--${size}` : '';
```
photodow: I'm thinking `size` would fail if it equals `''`, so I don't think the additional `&& size !== ''` here is necessary. Same comment for `L14`.

```suggestion
  const hrSize = size ? `bx--hr--${size}` : '';
```

packages/vanilla/src/components/horizontalrule/__stories__/horizontalrule.stories.js
```diff
@@ -0,0 +1,54 @@
+import { storiesOf } from '@storybook/html';
+import { withKnobs, select, text } from '@storybook/addon-knobs';
+import wrapper from './templates/wrapper.template';
+import horizontalrule from '../horizontalrule.template';
+import '../../../../../styles/scss/components/horizontalrule/_horizontalrule.scss';
+import 'carbon-components/scss/globals/grid/_grid.scss';
```
photodow: Do we want to import this from `@carbon/grid/scss/_grid.scss`?

packages/vanilla/src/components/horizontalrule/README.md
```diff
@@ -0,0 +1,91 @@
...
+Here's a quick example to get you started:
+
+### ES6 Template Literal
+
+```javascript
+import { horizontalruleTemplate } from '@carbon/ibmdotcom-vanilla';
+import '@carbon/ibmdotcom-styles/scss/components/horizontalrule/index.scss';
+
+console.log(
+  horizontalruleTemplate({
```
photodow: Might I suggest something like adding it into the page instead of consoling?

```javascript
document.body.innerHTML = horizontalruleTemplate();
```

packages/vanilla/src/components/horizontalrule/README.md
```diff
@@ -0,0 +1,91 @@
+# Horizontal Rule
+
+> The horizontal rule component is to be utilized within IBM.com for thematic
+> breaks within the content of the page.
+
+## Getting started
+
+Here's a quick example to get you started:
+
+### ES6 Template Literal
```
photodow: Should we add an area for an HTML example?

````markdown
## Getting started

Here's a quick example to get you started:

### HTML
Copy and paste the following into your HTML `body` to get started.

```html
<hr data-autoid="dds--hr" class="bx--hr" />
```

### ES6 Template Literal
````

File: packages/vanilla/src/components/horizontalrule/horizontalrule.template.js [hunk 6]
```diff
@@ -0,0 +1,24 @@
...
+ * Horizontal Rule template literal
+ *
+ * @param {object} props Property object passed into the template
+ * @param {string} props.style solid vs dashed [''|'dashed']
+ * @param {string} props.size size of the hr [''|'small'|'medium'|'large']
+ * @param {string} props.contrast color of the rule [''|'low'|'high']
+ * @param {string} props.weight weight of the rule [''|'thick']
+ * @returns {string} ES6 template literal of the horizontal rule
+ */
+const horizontalruleTemplate = ({ style, size, contrast, weight }) => {
```
photodow: @jeffchew I think this is a cool addition, but the part I'm struggling with a little in this package is this is less about the `HTML`, and more about the `Javascript`, right? It's hard with the given example `<HorizontalRule />` since it's doesn't have any given Javascript. How do you see components like `Masthead` or `Footer` fitting in with this direction since they'll need to be instantiated?

```html
<footer class="bx--footer">
  ...
</footer>
```

```javascript
dotcom.footer.init({ type: 'short' );
```

Do we add it to the page, assume it's been added via whatever framework before initializing

packages/vanilla/src/components/horizontalrule/horizontalrule.template.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * Horizontal Rule template literal
```
photodow: When I see `horizontalrule.template.js` I'm thinking I'll get back a `<template>...</template/>`

packages/vanilla/src/components/horizontalrule/horizontalrule.template.js
jeffchew: I guess it would fail in the sense where it would add in a class name `bx--hr--`, was trying to avoid that in the output. Just checking for `size` would certainly check for a null value, but doesn't account for empty string. I guess same would be if the user puts in any garbage string though, so we can just remove to be more terse anyway.

packages/vanilla/src/components/horizontalrule/README.md
jeffchew: It's actually below the `ES6 Template Literal` section. Should we bring that above it? I'm fine with either way, but at least both options are already in the README.

packages/vanilla/src/components/horizontalrule/README.md
jeffchew: since it's a README I don't really have much of a preference. At least for the Services and Utilities `@example` sections, we use console.log.

Adding @kennylam @annawen1 @andysherman2121 @sannminwin for any opinions on this.

packages/vanilla/src/components/horizontalrule/horizontalrule.template.js
jeffchew: I don't have a strong opinion either way, but in general I was hesitant to bring in any libraries for this, which is why I even opted against handlebars for the template rendering. 

Adding @annawen1 @kennylam @andysherman2121 @sannminwin for their opinion.

packages/vanilla/.storybook/manager-head.html
```diff
@@ -0,0 +1,31 @@
+<meta
+  name="keywords"
+  content="IBM, design, system, Carbon, design system, Bluemix, styleguide, style, guide, components, library, pattern, kit, component, cloud"
+/>
+<meta
+  name="description"
+  content="A collection of IBM.com components implemented using React and Carbon Design System."
```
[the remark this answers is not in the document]
jeffchew: Went through and removed references to react in the vanilla package.

packages/vanilla/.storybook/webpack.config.js
```diff
@@ -0,0 +1,138 @@
...
+ * @type {boolean}
+ */
+const useStyleSourceMap = process.env.VANILLA_STORYBOOK_SOURCEMAPS === 'true';
+
+/**
+ * Sets the document direction (https://developer.mozilla.org/en-US/docs/Web/API/Document/dir)
+ *
+ * @type {boolean}
+ */
+const useRtl = process.env.CARBON_REACT_STORYBOOK_USE_RTL === 'true';
```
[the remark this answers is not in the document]
jeffchew: Renamed to `VANILLA_STORYBOOK_USE_RTL`

packages/vanilla/src/components/horizontalrule/horizontalrule.template.js
jeffchew: Added additional checks to make sure the proper classes are added.

packages/vanilla/src/components/horizontalrule/horizontalrule.template.js
jeffchew: Since they will ultimately be ES6 classes, they would need to get instantiated. I was digging into the carbon code, and at least for the "kitchen sink" artifact that is generated, there is a flag called "autoinit" which a user can have all of the loaded classes autoinit on page load or not. For our repo, I turned that into an environment variable and called it `VANILLA_AUTOINIT`. 

To this point, it may be worth adding this to the package level README, but at the same time, Carbon team seemed to have purposely left this flag out from their documentation. It's on by default, so that means it would have to be manually set to turn off this behavior.

As for if a user is pulling in these classes adhoc (which will be most cases), it's up to them to instantiate once their application is ready, and for all intents and purposes, they might prefer this. We'd just need to document in the README for the corresponding components that have Javascript functionality to include the init function and how they can trigger it.

cc: @annawen1 @kennylam @sannminwin @andysherman2121

packages/vanilla/src/components/horizontalrule/__stories__/horizontalrule.stories.js
jeffchew: Changed it.

packages/vanilla/src/components/horizontalrule/horizontalrule.template.js
jeffchew: Not sure if we should change the name, as we're still using a way of templating but as a template literal. It's also much more flexible than the template tag. I honestly haven't seen any implementations of the template tag in production.

cc: @kennylam @annawen1 @andysherman2121 @sannminwin

packages/vanilla/src/components/horizontalrule/horizontalrule.template.js
photodow: Hmmm, unless I'm misunderstanding something just checking for `size` will fail if `null`, `undefined`, `false`, or if it's an empty string `''`.

https://jsfiddle.net/r5ockdag/

packages/vanilla/src/components/horizontalrule/README.md
photodow: I took that as what's returning from the ES6 Template Literal console. It might be good to have a specific place called out where the user can copy it from more clearly defined. Just a suggestion though.

packages/vanilla/src/components/horizontalrule/horizontalrule.template.js
photodow: I don't think we want people using the kitchen sink approach, or do we? Just for additional context within the discussion Carbon has adopters initialize it this way...

```scss
@import '~carbon-components/scss/components/card/card';
```
```javascript
import { Card } from 'carbon-components';

const cardElement = document.getElementById('card');
const cardInstance = Card.create(cardElement);
```
https://www.carbondesignsystem.com/get-started/develop/vanilla#javascript

packages/vanilla/src/components/horizontalrule/horizontalrule.template.js
photodow: I'm not necessarily suggesting we change the name. I'm not sure what I'm suggesting really. Just communicating what I expected this to return I suppose. No big deal.

> I honestly haven't seen any implementations of the template tag in production.

I've seen it used in tandem with a templating library like Handlebars. I've also seen it as a main player in building [Web Components](https://css-tricks.com/an-introduction-to-web-components/). 🤷‍♂️

packages/vanilla/src/components/horizontalrule/horizontalrule.template.js
jeffchew: We definitely don't want to encourage the kitchen sink approach. The README for the components will need to document how to instantiate the JS for that specific component.

packages/vanilla/src/components/horizontalrule/horizontalrule.template.js
jeffchew: To be fair, it's also because of the lack of support in IE11. With that said though, you can build out web components with either HTML templates or template literals, and from what I've seen the opinion on approach between one and the other is somewhat split. 

I like the flexibility of template literals, and it can be used today without much fuss. But it is a good discussion topic perhaps in our next huddle, and especially if/when we decide to start thinking about building web components.

Issue #421, opened 2019-10-13
Expand translation service to other languages mimicking the Northstar design


jeffchew:
_james-dow created the following on Sep 10:_  

 #### User Story
> As a `[user role below]`:
An IBM.com user

> I need to:
see IBM.com content in my own language

> so that I can:
understand what I'm reading without having to learn a new language.

#### Additional information
Now that the groundwork for the translation service has been created we need to update it to automatically serve the correct translation based on the user's locale and settings. I think @Jeff-Chew brought up the idea of using a utility to set the correct locale. I don't know if we also want to be able to pass in props to the service?

- [To Do in code #L41-42](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/services/src/services/Translation/Translation.js#L41-L42)
- [Locale Selector API](https://1.www.s81c.com/common/js/dynamicnav/www/countrylist/usen-utf8.js)
- [User Locale API](https://api.www.s81c.com/webmaster/dbip/)

#### Acceptance criteria
 - [ ] Default english translation
 - [ ] Props to override translation?
 - [ ] Automatically serve the correct translation based on user info. 

_Original issue: https://github.ibm.com/webstandards/digital-design/issues/1665_

jeffchew:
_On Sep 10, Jeff-Chew commented:_
@james-dow We will need to sync up with Juan Leon on how to fetch the locale/country. He mentioned they are doing it one way, while northstar is doing it another way. Ideally we will align with fetching this data from his team, we just need details on it.

jeffchew:
_On Sep 10, james-dow commented:_
@Jeff-Chew yeah sounds good. @vladislav-saling  was telling the other day v18 uses what I have there listed as [User Locale API](https://api.www.s81c.com/webmaster/dbip/). One call I was on for the Locale selector a couple weeks back they were throwing out a number of acronyms (gyou, uscn, usdn, ipc info cookie) as different methods to get this information. There may also be room to explore something like geolocation, and the browser language? But that sounds good.

jeffchew:
_On Sep 11, vladislav-saling commented:_
Note: geolocation is done trough Akamai API in v18. See `https://api.www.s81c.com/webmaster/dbip/`

jeffchew:
_On Oct 02, Jeff-Chew commented:_
[...]

PR #549, opened 2019-10-22
feat(locale): hook locale selector service to masthead and footer

annawen1:
### Related Ticket(s)

[Expand translation service to other languages mimicking the Northstar design #421 ](https://github.com/carbon-design-system/ibm-dotcom-library/issues/421)

### Description

Hook the locale service to the footer and masthead. Adding dropdown to the footer locale button modal in order to see the service in action

You can see the language change on refresh

**NOTE** the masthead needs to be adjusted later it's making some checks that don't address all the different scenarios that we get with different languages

Preview Link:
[https://deploy-preview-549--ibmdotcom-react-experimental.netlify.com](https://deploy-preview-549--ibmdotcom-react-experimental.netlify.com)

### Changelog

**New**

- add `<ComboBox />` carbon component to modal
- when item is selected from dropdown, the `ipcInfo` cookie is set
- Footer has locale state and re-renders when locale is set by dropdown
- trigger `LocaleAPI` in `TranslationAPI` service when the feature flag for the footer locale button is true


packages/services/src/services/Locale/Locale.js
```diff
@@ -85,20 +87,24 @@ class LocaleAPI {
...
-    return await axios
+    const list = await axios
       .get(url, {
         headers: {
           'Content-Type': 'application/json; charset=utf-8',
         },
       })
       .then(response => response.data);
+
+    localStorage.setItem('countryList', JSON.stringify(list));
```
jeffchew: @annawen1 wondering if we should either set this to expire, or consider using sessionStorage? Also, the country list doesn't change often, but we'd need a way to possibly cache bust it if necessary as well.

packages/react/src/components/Footer/LocaleButton.js
```diff
@@ -19,11 +20,34 @@ const { prefix } = settings;
...
+ * @param {Function} selectItem method to handle selected item
+ *
  * @returns {object} JSX object
  */
-const LocaleButton = () => {
+const LocaleButton = ({ selectItem }) => {
   const [isOpen, setIsOpen] = useState(false);
 
+  // retrieve list from local storage
+  const list = JSON.parse(localStorage.getItem('countryList'));
```
jeffchew: I know this is still just experimental, but we should probably get this data from the service rather than fetching right from localStorage. Let the service determine if it's a new net call or fetch from localStorage.

packages/services/src/services/Locale/Locale.js
annawen1: yes, I'll need to do that (for the `ipcInfo` cookie as well!) - I believe NorthStar has both set with an expiration of a day...

annawen1: looking at v18, it seems that they set the expiration for the ipcInfo cookie to a year

var allowedTtl = IBMCore.www.module.truste.getUiAllowedStorageTtl(365 * 24 * 3600); in util.setIpcCookie() in misc.js

and then the country list in local storage to a day

var allowedTtl = IBMCore.www.module.truste.getUiAllowedStorageTtl(24 * 3600);

		// Cache translations locally for case of meta change page locale.
		selectorData = serviceData;

		// Store in global data in case page wants to use it.
		IBM.common.util.data.provide("localeSelector", serviceData);

		// Cache in local storage for 24 hours for subsequent page views.
		IBM.common.util.storage.setItem("v18lsdata", selectorData, allowedTtl);

annawen1: I set the cookie to a year and switched to session storage for the country list

packages/react/src/components/Footer/LocaleButton.js
annawen1: The service gets called in the `TranslationAPI` service and does the work of checking sessionStorage/making a new call and setting session storage

jeffchew: Right, but the getter should be also from the service and not in here. Essentially, make a call to `LocaleAPI.getList`, and it should return it. 

In the API, it should check if the localStorage is already set. If it is, return that. If not, make the axios call, set it, then return the list.

annawen1: done, the locale is retrieved by calling `LocaleAPI.getLocale` which will determine the locale from either the `lang` attr or the cookie. I'm keeping the cookie for instances where the `lang` isn't there.

jeffchew:
@annawen1 approved, though one thing I'd possibly like to see if it isn't too much trouble is to refresh the page once a user changes the country/language. Not a huge deal though if we're going to replace with the real deal. I might even just merge this in now and we can do that in a separate PR if it's even worth it.

Issue #392, opened 2019-10-13
Analytics: Add events for the Carbon Tabs component


jeffchew:
_Jeff-Chew created the following on Sep 06:_  

 

#### User Story

> As a `[user role below]`:
IBM Data Analyst

> I need to:
have dynamic event triggers from the Carbon Tabs component

> so that I can:
track dynamic events within the tabs component in my analytics tool

#### Additional information

 - http://ibmdev.webmaster.ibm.com/v18/metrics/
- Will need more details on what events need to be triggered from Juan Leon and team
- https://www.carbondesignsystem.com/components/tabs/code/#events
- Added to the global.js, we will listen to the events emitted from Carbon and add these to those.

#### Acceptance criteria
 - [ ] Event triggers added to the Carbon tabs component per specs provided by Juan Leon's team
(https://ibm.ent.box.com/s/jw44u3vh2cxg0exgjwthvk03do79iki4)
 

_Original issue: https://github.ibm.com/webstandards/digital-design/issues/1604_

jeffchew:
Currently it seems that the global event triggers are initialized only in the Vanilla version of Tabs and  not the React version. Also, there currently isn't an event to capture when Tabs have been initialized.

A ticket has been opened on the Carbon side to hopefully get these enhancements in place:
https://github.com/carbon-design-system/carbon/issues/4454

Issue #391, opened 2019-10-13
Analytics: Add event triggers for the Carbon Modal


jeffchew:
_Jeff-Chew created the following on Sep 06:_  

 

#### User Story

> As a `[user role below]`:
IBM Data Analyst

> I need to:
have dynamic event triggers from the Carbon modal

> so that I can:
track dynamic events for the modal in my analytics tool

#### Additional information

 - http://ibmdev.webmaster.ibm.com/v18/metrics/
- Will need more details on what events need to be triggered from Juan Leon and team
- Carbon's modal apparently has generic event listeners we can subscribe to
- https://www.carbondesignsystem.com/components/tabs/code/#events
- Added to the global.js, we will listen to the events emitted from Carbon and add these to those.

#### Acceptance criteria
 - [ ] Event triggers added to the Masthead per specs provided by Juan Leon's team
(https://ibm.ent.box.com/s/jw44u3vh2cxg0exgjwthvk03do79iki4)
 

_Original issue: https://github.ibm.com/webstandards/digital-design/issues/1603_

jeffchew:
Similar to Tabs, the React Modal is not emitting the global events, while Vanilla Modal is. Created an issue on the Carbon issue board to request for this enhancement:
https://github.com/carbon-design-system/carbon/issues/4455

PR #571, opened 2019-10-24
Analytics[391][392]: modals and dynamic tabs

jeffchew:
### Related Ticket(s)

#391 , #392 

### Description

This creates the event triggers for Carbon's Modal and Tab components. Currently, this only works for Vanilla Tabs. There is an issue opened on the Carbon side to open up these events to the React Modal/Tabs components: 

- https://github.com/carbon-design-system/carbon/issues/4454
- https://github.com/carbon-design-system/carbon/issues/4455

The other change I made was that I made the global init something that can be imported on the component level, and only initializes once. I also moved it into utilities.

### Changelog

**New**

- Analytics triggers for Carbon Tabs

**Changed**

- Moved global init function into utilities


packages/services/src/services/Analytics/Analytics.js
```diff
@@ -86,6 +86,81 @@ class AnalyticsAPI {
...
+        fireEvent({
+          type: 'element',
+          primaryCategory: 'WIDGET',
+          eventName: 'CLICK',
+          eventCategoryGroup: 'TABS DYNAMIC',
+          executionPath: evt.target.id,
+          targetTitle: evt.detail.item.innerText,
+        });
+      } catch (err) {
+        console.error('Error triggering tab event:', err);
```
photodow: Do we want to be consoling this in production?

packages/services/src/services/Analytics/Analytics.js
```diff
@@ -86,6 +86,81 @@ class AnalyticsAPI {
...
+          eventCategoryGroup: 'SHOWHIDE',
+          executionPath: evt.target.id,
+          targetTitle: evt.detail.launchingElement.innerText,
+        });
+      } catch (err) {
+        console.error('Error triggering modal hide event:', err);
+      }
+    });
+
+    root.document.addEventListener('modal-shown', function(evt) {
```
photodow: Does it make sense to bring these out into their methods, with more specific naming `onModalOpen`?

packages/utilities/src/utilities/global/global.js
```diff
@@ -0,0 +1,29 @@
...
+    _initialized = true;
+  }
+
+  // Sets the version of the library in the DDO
+  DDOAPI.setVersion();
+
+  // analytics tracking
+  AnalyticsAPI.initScrollTracker();
+  AnalyticsAPI.initDynamicTabs();
+  AnalyticsAPI.initModals();
```
photodow: Does it make sense to add a method within `AnalyticsAPI` that does this? Something like `AnalyticsAPI.initAll()`?

packages/react/src/components/Footer/Footer.js
```diff
@@ -29,6 +30,11 @@ const Footer = ({ type }) => {
   const [footerMenuData, setFooterMenuData] = useState([]);
   const [footerLegalData, setFooterLegalData] = useState([]);
 
+  useEffect(() => {
+    // initialize global execution calls
+    globalInit();
```
photodow: Are we tracking anything in Footer? Or are you planning to add this to every component? Is this how you're seeing executing #391 within the expressive modal as well?

jeffchew: Good question. We originally had this added to the DotcomShell, but there would be instances possibly where adopters will pull the `Masthead` and/or `Footer` separately. Ultimately though, this is code that needs to be included if any adopter is using the IBM.com Library. Ideally we would have the adopters just load in the globalInit themselves then trigger it, but that most likely won't be always the case. We can include this as part of the getting started documentation on the website, but again we won't be able to reliably capture everyone doing that on the honor system.

Adding this as an include for the Masthead and Footer would be a good place to capture a large percentage of IBM.com. For those that aren't using the Masthead or the Footer, we would need to ask those adopters to call the globalInit directly.

The other option, to your point, is to include it in every component.

Open to thoughts on this.

@annawen1 @kennylam @andysherman2121 @sannminwin

packages/services/src/services/Analytics/Analytics.js
jeffchew: I was just thinking this when reviewing the expressive modal. I think that implementation might have to call these methods directly, so have to probably break this one out.

packages/utilities/src/utilities/global/global.js
jeffchew: Updated in https://github.com/carbon-design-system/ibm-dotcom-library/pull/571/commits/421039ac7f4faef4256f586450b37ea2427a1f36

packages/services/src/services/Analytics/Analytics.js
jeffchew: Updated in https://github.com/carbon-design-system/ibm-dotcom-library/pull/571/commits/421039ac7f4faef4256f586450b37ea2427a1f36

packages/services/src/services/Analytics/Analytics.js
jeffchew: Updated in https://github.com/carbon-design-system/ibm-dotcom-library/pull/571/commits/421039ac7f4faef4256f586450b37ea2427a1f36, now executing if not in production.

packages/react/src/components/Footer/Footer.js
jeffchew: Regardless, I think for now we might be ok with this moving forward in this PR as it will cover all instances of adopters using this. We should possibly add documentation to the README about the globalInit, and/or the cupcake website.

jeffchew: I have added this as a huddle topic for next week.

jeffchew:
> Based on the latest commits these changes look good. The only thing is it's bringing in a bunch of already committed commits. I can't remember if this negatively impacts the logs or if they level out once merged.

Thanks @photodow , I did a squash commit to fix that.

PR #575, opened 2019-10-26
Expressive modal

### LINKED ISSUE #253: Develop the modal w/ margins component
_hahnrob created the following on Jun 20:_  

 

#### User Story

> As a `[user role below]`:
As a developer using the Digital Design System

> I need to:
have an Expressive Modal in the Locale Selector

> so that I can:
include this component in the Locale Selector in the web site I am developing

#### Additional information

- This issue is dependent on the design and the [package architecture https://github.com/carbon-design-system/ibm-dotcom-library/issues/253](https://zenhub.ibm.com/app/workspaces/digital-design-system-5b3a815a8bf6a932efa6fd03/issues/webstandards/digital-design/860) 
- [final design and specs](https://ibm.box.com/s/isie20a89socbo9ael687u6pb75uz9f8)
              - Includes:  Functional labels, Visual redlines, Behavior and interaction specs, Desktop and mobile
 - [Box link to Mary O'Donnell's functional specs](https://ibm.ent.box.com/folder/88263129224)
 - While this expressive modal is included in the Locale Selector, it will be developed as a generic component.
 - Try to leverage the current Carbon modal and applying the expressive theme.  
 - [Carbon Modal](https://www.carbondesignsystem.com/components/modal/code)

#### Acceptance criteria
 - [ ] Component has been tested using Netlify
 - [ ] The component development follows a modular component pattern
 - [ ] The component should be accessible via ES6 import
 - [ ] The component should be visible within the living documentation of the Dotcom components
 - [ ] Presented and approved Product Owner

 

_Original issue: https://github.ibm.com/webstandards/digital-design/issues/1078_

photodow:
### Related Ticket(s)

closes #253

### Description

This pull request creates a new expressive modal leverages the Carbon's modal component. It mostly adds 32px margin around the whole container, pulls out the close button opening up the ability for content transitions and whatever content needs to go in.

image

### Changelog

**New**

- Expressive Modal

**Changed**

- Updated locale button with expressive modal


packages/react/src/components/ExpressiveModal/ExpressiveModal.js
```diff
@@ -0,0 +1,74 @@
...
+
+ExpressiveModal.propTypes = {
+  open: PropTypes.bool,
+  className: PropTypes.string,
+  children: PropTypes.node,
+  onClose: PropTypes.func,
+};
+
+ExpressiveModal.defaultProps = {
+  open: false,
```
jeffchew: @photodow I'm wondering if we should also pass in props for the primary and secondary buttons similar to how the productive modal works. They can be optional attributes, and if they don't pass anything then the ButtonGroup doesn't render.

cc: @annawen1 @sannminwin @andysherman2121 @kennylam

packages/react/src/components/ExpressiveModal/ExpressiveModalCloseBtn.js
```diff
@@ -0,0 +1,55 @@
...
+import { settings } from 'carbon-components';
+import { Close20 } from '@carbon/icons-react';
+
+const { stablePrefix } = ddsSettings;
+const { prefix } = settings;
+
+/**
+ * Expressive Modal
+ *
+ * @param {object} props react proptypes
```
jeffchew: Can you add JSDocs for the props?

packages/react/src/components/ExpressiveModal/ExpressiveModal.js
```diff
@@ -0,0 +1,74 @@
...
+import { ComposedModal } from 'carbon-components-react';
+import ExpressiveModalCloseBtn from './ExpressiveModalCloseBtn';
+
+const { stablePrefix } = ddsSettings;
+const { prefix } = settings;
+
+/**
+ * Expressive Modal
+ *
+ * @param {object} props react proptypes
```
jeffchew: Can you add JSDocs for the props?

packages/react/src/components/ExpressiveModal/ExpressiveModal.js
```diff
@@ -0,0 +1,74 @@
...
+      setIsOpen(false);
+    }
+  }
+
+  /**
+   * Manage open prop and isOpen state
+   */
+  function manageOpenState() {
+    if (prevOpen !== open) {
+      setIsOpen(open);
```
jeffchew: @photodow one thing I neglected to add in the acceptance criteria are the analytics triggered events in this modal. We could probably do this in a separate PR, but that means we'd have to add it post code freeze if we can't sneak this in today.

packages/react/src/components/ExpressiveModal/ExpressiveModalCloseBtn.js
photodow: Sorry, isn't that what this line is?

packages/react/src/components/ExpressiveModal/ExpressiveModal.js
photodow: No, I don't think so. The modal itself isn't supposed to have buttons. When you add buttons it becomes a dialogue. The locale selector and lightbox, for example, should use this component, but won't have buttons.

@wonilsuhibm @oliviaflory please correct me if I misunderstood.

packages/react/src/components/ExpressiveModal/ExpressiveModal.js
photodow: I was assuming that was part of the https://github.com/carbon-design-system/ibm-dotcom-library/issues/391 story. I'm fine with a separate pull request. I don't think code freeze has to be so strict, but it's up to you. At the very least #571 needs to go in first, right?

packages/react/src/components/ExpressiveModal/ExpressiveModalCloseBtn.js
jeffchew: See example here:
https://github.com/carbon-design-system/ibm-dotcom-library/blob/ebc6b1f4dbf82e0f3eefb8bf53b42bff7992edfe/packages/react/src/components/HorizontalRule/HorizontalRule.js#L31

packages/react/src/components/ExpressiveModal/ExpressiveModal.js
jeffchew: @photodow Yes, #571 definitely has to go in first. I'm fine with getting this in separately.

packages/react/src/components/ExpressiveModal/ExpressiveModalCloseBtn.js
photodow: At first, glance that looks inaccurate, but I'll dive in and take a closer look. Thanks for the example!

photodow: I see it now, thanks! I'll make the update.

packages/react/src/components/ExpressiveModal/ExpressiveModal.js
oliviaflory: @jeffchew Wonil and I met with James on Friday and decided the expressive modal should just be a shell so adopters can add whatever they need to it. We will still provide the other modals Carbon has in the theming. Sorry for not updating our convo in the original issue.

After a design critique with Carbon on Thursday we think the expressive modal is different than the current modals that exist in Carbon today, which are like a 'dialogue box' that requires an action / decision. Whereas the expressive modal could be used for lightbox or local selector that could have different requirements or no action required. (We need to work on the definition).

jeffchew: Thanks for bringing me up to speed @oliviaflory . It's too bad we can't just call ours `<Modal />` and rename theirs `<Dialog />`

oliviaflory:
@photodow question about the padding on the expressive modal, is the 30px on the right coming from how Carbon built the modals or us? Carbon's current guidance says text should only span 75% of the modal, is that baked in or do developers have to add that? Only wondering if we should use a percentage vs a fixed pixel size.

Screen Shot 2019-10-29 at 10 18 52 AM

photodow:
@oliviaflory reached out via slack.

> The spacing on the right clears the close button, would you like the 75%? I removed it because of the designs that had the borders going full across, and y’all didn’t mention keeping it when we talked. It sounded more like a pain or restricting, so I removed it, but if not I can add it back :slightly_smiling_face:

image

It's still not perfect because in this design when it scrolls up it'll overlap with the `x`, and the one this pull request has tried to avoid it so there is no overlap. Just let me know, it won't be hard to revert.

PR #577, opened 2019-10-29
fix(featureflag): hiding experimental stories from production

jeffchew:
### Related Ticket(s)

No related issues

### Description

This is a quick PR to hide experimental features from the production storybook output. There were a couple of features behind feature flags that were rendering blank storybook stories. This adds the feature flag around the story itself to only render if the flag is set to true.

I also added the entry for vanilla package in the base README.

### Changelog

**Changed**

- Added feature flags around experimental storybook stories
- Updates to main README

### DIFF
--- a/packages/react/src/components/CardLink/__stories__/Card.stories.js
+++ b/packages/react/src/components/CardLink/__stories__/Card.stories.js
@@ -4,65 +4,68 @@ import { withKnobs, select, text } from '@storybook/addon-knobs';
 import { CardLink } from '../';
 import { ArrowRight20 } from '@carbon/icons-react';
 import readme from '../README.md';
+import { CARD_LINK } from '../../../internal/FeatureFlags';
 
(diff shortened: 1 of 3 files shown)


packages/react/src/components/CardLink/__stories__/Card.stories.js
```diff
@@ -4,65 +4,68 @@ import { withKnobs, select, text } from '@storybook/addon-knobs';
...
 
-storiesOf('Card', module)
-  .addDecorator(withKnobs)
-  .addDecorator(Story => {
-    return (
-      <div className="bx--grid">
-        <div className="bx--row">
-          <div className="bx--col-sm-2 bx--col-md-3 bx--col-lg-6 bx--col-xlg-4 bx--no-gutter">
-            <Story />
+if (CARD_LINK) {
```
photodow: Sorry! Thanks!

PR #580, opened 2019-10-29
feat(altlang): utility for fetching all alternative languages on page

jeffchew:
### Related Ticket(s)

#509 

### Description

This is a new utility that grabs all of the alternative languages on the page and returns as an object. For example, the following on the page:

```html
<link rel="alternate" hreflang="en-us" href="https://www.ibm.com/us-en/">
<link rel="alternate" hreflang="x-default" href="https://www.ibm.com">
<link rel="alternate" hreflang="en-af" href="https://www.ibm.com/af-en">
<link rel="alternate" hreflang="fr-dz" href="https://www.ibm.com/dz-fr">
<link rel="alternate" hreflang="pt-ao" href="https://www.ibm.com/ao-pt">
```

Will return:

```javascript
{
    'en-us': 'https://www.ibm.com/us-en/',
    'x-default': 'https://www.ibm.com',
    'en-af': 'https://www.ibm.com/af-en',
    'fr-dz': 'https://www.ibm.com/dz-fr',
    'pt-ao': 'https://www.ibm.com/ao-pt',
}
```

### Changelog

**New**

- `altlangs` utility

**Changed**

- added alternative langs on the React storybook base page in anticipation of locale selector development


packages/react/.storybook/preview-head.html
```diff
@@ -1,2 +1,137 @@
...
+<link rel="alternate" hreflang="en-ua" href="https://www.ibm.com/ua-en">
+<link rel="alternate" hreflang="en-ug" href="https://www.ibm.com/ug-en">
+<link rel="alternate" hreflang="en-ae" href="https://www.ibm.com/ae-en">
+<link rel="alternate" hreflang="en-gb" href="https://www.ibm.com/uk-en">
+<link rel="alternate" hreflang="es-uy" href="https://www.ibm.com/uy-es">
+<link rel="alternate" hreflang="en-uz" href="https://www.ibm.com/uz-en">
+<link rel="alternate" hreflang="es-ve" href="https://www.ibm.com/ve-es">
+<link rel="alternate" hreflang="en-vn" href="https://www.ibm.com/vn-en">
+<link rel="alternate" hreflang="en-zm" href="https://www.ibm.com/zm-en">
+<link rel="alternate" hreflang="en-zw" href="https://www.ibm.com/zw-en">
```
photodow: 🙀

packages/utilities/src/utilities/altlangs/altlangs.js
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * @returns {object} object of available languages and corresponding URLs
+ *
+ */
+function altlangs() {
+  const links = root.document.querySelectorAll('link[rel="alternate"]');
+  const langs = {};
+
+  links.forEach(link => {
+    langs[link.getAttribute('hreflang')] = link.getAttribute('href');
```
photodow: Is there any value in being explicit with lang and country, and pulling those out for easy access?

Something like this...
```javascript
const hreflang = link.getAttribute('hreflang');
const splitHrefLang = hreflang.split('-');

langs[hreflang] = link.getAttribute('href');
langs['cc'] = splitHrefLang[0];
langs['ll'] = splitHrefLang[1];
```

jeffchew: Good suggestion. I'll add that!

jeffchew: Was thinking a bit on this, the list of available country/language would essentially be cross referenced with the country list here: https://www.ibm.com/common/v18/js/data/jsononly/locale/usen-locale.json

This list already has it in the format of `cc-lc`, so my hunch is that when generating the available list, it would check to see if `altlangs` has those sets of keys available in that object. We might be ok without breaking it out?

cc: @annawen1 @kennylam @andysherman2121 @sannminwin

photodow: That's probably fair.  I'm cool to merge this if y'all are. We can always split them out down the road if the need arises.

packages/react/.storybook/preview-head.html
annawen1: 😯

packages/utilities/src/utilities/altlangs/altlangs.js
annawen1: hmmm the `LocaleAPI.verifyLocale()` method expects the `cc` and `lc` as separate parameters, so if we want to reuse that I think we should break it out.
```
static verifyLocale(cc, lc, list) { ... }
```

jeffchew: Thanks @annawen1 and @photodow , I'll merge in as is then!

Issue #588, opened 2019-11-01
Accessibility issue on the carbon search box on https://ibmdotcom-react.netlify.com/iframe.html?id=dotcom-shell--default


alainmachu:
-->

**## Detailed description**

> Describe in detail the issue you're having.
On https://ibmdotcom-react.netlify.com/iframe.html?id=dotcom-shell--default, expand the search box and type java
It triggers a lot of suggestions
If we use the down arrow on the keyboard to go through the suggestions, we see them in the search box, as we go down, but they are not highlighted in the list below the search box

Best practice for it would be, that as we go through the list with the down or up key, and a suggestion is showing in the search box, the same suggestion should be highlighted in the list below.

> Is this a feature request (new component, new icon), a bug, or a general
> issue? it's an accessibility issue, best practice is that in the scenario above, the suggestions are highlighted as we go through them with the down or up keyword.

> Is this issue related to a specific component? Carbon

> What did you expect to happen? What happened instead? What would you like to
> see changed? See above

> What browser are you working in? Firefox and Chrome

**## Steps to reproduce the issue**

1. Step one: load https://ibmdotcom-react.netlify.com/iframe.html?id=dotcom-shell--default
2. Step two: enter java in the search box
3. Step three: you will see suggestion list below the search box
4. Use down key in the board to go through the list
5. the first suggestion is javascript
6. While using the the down key, javascript shows in the search box in place of java, but it is NOT highlighted in the list
7. And so on for all the keywords

Issue #135, opened 2019-09-27
Clean up unused classes in Masthead


kennylam:
Does anyone know what `${prefix}--container-highlight-class` is doing? I didn't see a reference for it in the `SCSS`.

_Originally posted by @photodow in https://github.com/carbon-design-system/ibm-dotcom-library/pull/132_

kennylam:
TODO: The auto-suggestion drop down in the search component has HTML/CSS that can be removed.

jeffchew:
This looks like a clean up task that almost got pulled away from stalebot. Moving this for triaging.

PR #591, opened 2019-11-01
Search dropdown highlight fix

photodow:
### Related Ticket(s)

fixes: #588 
fixes: #135

### Description

This pull request adds the highlighting to the search suggestion as you key down and up arrows on your keyboard.

Nov-01-2019 11-24-28

### Changelog

**New**

- Added `.bx--container-highlight-class` to adopt focus state


jeffchew:
@photodow I noticed that this is targeting master, which I think I might be ok with. Unless we want to put it in a release candidate? If that's the case, you'd probably need to cherry pick it and commit off of the release branch since master is ahead right now.

Issue #589, opened 2019-11-01
Spacing issue with Carbon search Box on https://ibmdotcom-react.netlify.com/iframe.html?id=dotcom-shell--default


andysherman2121:
Detailed Description:

On https://ibmdotcom-react.netlify.com/iframe.html?id=dotcom-shell--default
When trying to search, the auto complete doesn't put a space between the first and the second word.
Ex: If you search for Cloud, the options with multiple words will show something like this; Cloudcomputing
Cloudobject

![Auto Complete.png](https://images.zenhubusercontent.com/5da883ca0621b7000163aeca/d6fbeff8-b88d-4616-a5fc-5c952ab64698)

PR #594, opened 2019-11-01
fix(masthead): adds space for suggestion item

jeffchew:
### Related Ticket(s)

#589 

### Description

This adds in a visual space for the typeahead suggestion item, as there were no spaces getting added with the way the content was currently structured.

### Changelog

**Changed**

- Masthead search suggestion item visual fix


jeffchew:
> Search terms still aren't being broken up from first to second word

I pushed up an additional fix, you should be able to see it once netlify rebuilds.

Issue #597, opened 2019-11-01
Search not closing


photodow:
## Detailed description

> Describe in detail the issue you're having.

The close button on the masthead search is not responding to close the search feature.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

bug

> Is this issue related to a specific component?

Masthead

> What did you expect to happen? What happened instead? What would you like to
> see changed?

The search should close

PR #598, opened 2019-11-01
fix(masthead): fix to search input not closing

jeffchew:
### Related Ticket(s)

Fixes: #597 

### Description

This is a fix for the search field not closing. The overall fix was to get rid of `useState` and just use `useReducer` for all of the interactions.

### Changelog

**Changed**

- Fixed handling of states in the Masthead search


jeffchew:
> Looks good. only comment I see is you got some extra commits previously merged showing up for some reason. This seemed like an easy fix the other day though.

Yeah I noticed the same thing. I cleaned it up with a squash commit on my end, waiting for it to rebuild to see how it goes.

PR #601, opened 2019-11-01
Fix/unit tests

jeffchew:
### Related Ticket(s)

No related issues

### Description

This fixes various unit tests. The main one it fixes is one in the services package (LocaleAPI), which is breaking the ci-checks for that particular package.

### Changelog

**Changed**

- Various unit test fixes

### DIFF
--- a/packages/services/src/services/Locale/Locale.js
+++ b/packages/services/src/services/Locale/Locale.js
@@ -68,8 +68,7 @@ class LocaleAPI {
     const lang = this.getLang();
     // grab locale from the html lang attribute
     if (lang) {
-      const locale = await this.getList(lang);
-      return locale;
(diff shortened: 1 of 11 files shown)


packages/services/src/services/Locale/Locale.js
```diff
@@ -114,8 +113,8 @@ class LocaleAPI {
       const lang = root.document.documentElement.lang.toLowerCase();
       const codes = lang.split('-');
       return { cc: codes[1], lc: codes[0] };
-    } else {
-      return _localeDefault;
+      //    } else {
+      //      return _localeDefault;
```
photodow: I see we didn't remove `_localeDefault` is that going to cause any linting errors?

PR #623, opened 2019-11-05
[RFC] Expanding featureFlag Utility

### LINKED ISSUE #618: [FeatureFlag] Expand functionality of the featureFlag utility
If we want the feature flag in a broader scope like stories, then maybe we could look at expanding the `featureFlag` utility function. I only used JSX when creating it because that's all I thought we'd be using it for at the time. We could maybe move to a callback approach.

```javascript
function featureFlag(flag, cb) {
  if (flag) {
    if (typeof cb === 'function') {
      return cb(flag)
    } else {
      return cb;
    }
  }

  return null;
}
```

_Originally posted by @photodow in https://github.com/carbon-design-system/ibm-dotcom-library/timeline_

photodow:
### Related Ticket(s)

fixes #618 

### Description

This pull request is more for discussion purposes at our next Huddle. See #618. Ultimately it moves the `DDS_FLAGS_ALL` logic into the `featureFlag` utility, and expands to allow passing in functions as a condition to enable usage with storybook.

### Changelog

**Changed**

- Expands `featureFlag` to allow functions from storybook
- Simplify `DDS_FLAGS_ALL`, so we maintain it once.


jeffchew:
This was discussed in huddle, some thoughts on this:
- Since the DDS_FLAGS_ALL flag would be set on the utility level and not the flag level, code can be introduced where it uses the flag directly and essentially would circumvent the utility
- Not urgent at this time, can come back to this when the design system is more mature
- Using the current approach is fine for now

cc: @kennylam @annawen1 @sannminwin @andysherman2121 @emerson-pereira, please profide any additional feedback

photodow:
> Since the DDS_FLAGS_ALL flag would be set on the utility level and not the flag level, code can be introduced where it uses the flag directly and essentially would circumvent the utility

Can someone elaborate on this first point a little? I don't understand the difference between the utility level and flag level. The `DDS_FLAGS_ALL` flag gets set at the same level as all of the other flags. The utility can definitely be circumvented, but it allows us to keep the logic in one place, reduces the duplications of code, and simplifies the integration of a feature flag in general.

jeffchew:
> > Since the DDS_FLAGS_ALL flag would be set on the utility level and not the flag level, code can be introduced where it uses the flag directly and essentially would circumvent the utility
> 
> Can someone elaborate on this first point a little? I don't understand the difference between the utility level and flag level. The `DDS_FLAGS_ALL` flag gets set at the same level as all of the other flags. The utility can definitely be circumvented, but it allows us to keep the logic in one place, reduces the duplications of code, and simplifies the integration of a feature flag in general.

That's the main point, where a developer could circumvent the utility entirely, whereas if the value of the flag is set primarily on the flag level and not through a utility, then there's no way to avoid a scenario where `DDS_FLAGS_ALL` may not necessary have all feature flag functionality turned on if somewhere in the code the utility is not being used.

jeffchew:
Closing this, don't think this is necessary at this time.

PR #626, opened 2019-11-05
fix(fonts): seperating fonts.scss file from global imports

andysherman2121:
### Related Ticket(s)

https://github.com/carbon-design-system/ibm-dotcom-library/issues/527

### Description

Removes the font packages from the global _imports.scss file. The font packages were being imported each time one of our components were being used, which increased load times. The user is now instructed to only import the separate font file once as needed. 

### Changelog

**New**

added _fonts.scss file
added instructions in component readme files

**Changed**

removed fonts from globals/_imports.scss


andysherman2121:
Still making some updates

packages/styles/scss/globals/_fonts.scss
```diff
@@ -0,0 +1,10 @@
+//
+// Copyright IBM Corp. 2016, 2018
+//
+// This source code is licensed under the Apache-2.0 license found in the
+// LICENSE file in the root directory of this source tree.
+//
+
+@import '@carbon/type/scss/type';
+@import '@carbon/type/scss/font-face/mono';
+@import '@carbon/type/scss/font-face/sans';
```
photodow: I realize y'all had a conversation in New York to go in this direction, but is it necessary to have this file when we could just point adopters to Carbon to do it themselves?

https://www.npmjs.com/package/@carbon/type

andysherman2121: And just remove the font files on our side? @photodow

photodow: Yeah, so remove the imports, remove this new file, and just in our documentation somewhere point people to the link above.

andysherman2121: Fine with me either way, let's get some more feedback and see how everyone else feels

kennylam: > I realize y'all had a conversation in New York to go in this direction, but is it necessary to have this file when we could just point adopters to Carbon to do it themselves?

I don't have a problem with this at all.

Issue #445, opened 2019-10-13
Pattern: develop sticky Table of Contents


jeffchew:
_hahnrob created the following on Sep 16:_  

 

#### User Story

> As a `[user role below]`:
developer using the ibm.com Library

> I need to:
have a sticky table of content pattern that can be used

> so that I can:
leverage it within my application

#### Additional information

- Story within Storybook with corresponding knobs
- minimum 80% unit test coverage
- Utilize Carbon
- [Specs](https://ibm.box.com/s/18w1s9z4j57as682hmkl0du57mwh380u)
- [Spec issue](https://zenhub.ibm.com/app/workspaces/digital-design-system-5b3a815a8bf6a932efa6fd03/issues/webstandards/digital-design/1708)
 - [About virtual server example web page](https://www.ibm.com/cloud/virtual-servers-cloud/about-virtual-machines/index.html)

#### Acceptance criteria
 - [ ] tested React component within the Patterns Storybook

 

_Original issue: https://github.ibm.com/webstandards/digital-design/issues/1709_

wonilsuhibm:
New specs have been added to the [spec issue](https://github.ibm.com/webstandards/digital-design/issues/1708)

jeffchew:
@sannminwin FYI, `Layout` component has been completed and can be used for this story.

sannminwin:
@RobertaJHahn 
@ljcarot 
@jeffchew 
Hello Team,
Sticky table of contents is ready for PR.
Thanks

[Link to PR](https://github.com/carbon-design-system/ibm-dotcom-library/pull/667)

PR #667, opened 2019-11-12
feat(pattern): sticky table of contents

sannminwin:
### Related Ticket(s)

#445   (Pattern: develop sticky table of contents)

Screen Shot 2019-11-12 at 9 18 20 AM

### Description
This pull request creates sticky table of contents pattern.

Deploy Preview:

### Changelog

**New**

- Sticky Table of Contents pattern


wonilsuhibm:
Thank you. This is exactly what we need. The user test results are very positive about this too.

A few comments and questions:

- can the blue bar animate from one spot to the next instead of fading?
- if it has to fade for MVP, can it be faster? I suggest around `110ms` as the duration.
- Is there a hover state for each item?
- When you click on the last item, the blue bar doesn't stop at the last item sometime.
   1. make the window short
   2. Click the last item
   3. See the page scroll to the bottom, but the blue bar does not reach the last item.
   4. Is the focus state correct?

wonilsuhibm:
One more thing - when you refresh the Netlify page, the blue bar doesn't show until you start scrolling.

sannminwin:
> Thank you. This is exactly what we need. The user test results are very positive about this too.
> 
> A few comments and questions:
> 
> * can the blue bar animate from one spot to the next instead of fading?
> * if it has to fade for MVP, can it be faster? I suggest around `110ms` as the duration.
> * Is there a hover state for each item?
> * When you click on the last item, the blue bar doesn't stop at the last item sometime.
>   
>   1. make the window short
>   2. Click the last item
>   3. See the page scroll to the bottom, but the blue bar does not reach the last item.
>   4. Is the focus state correct?

Hey @wonilsuhibm ,
I don't see hover state and focus are included in specs, I am coming to your desk so that we can have conversation.

packages/patterns-react/src/patterns/TableOfContents/TableOfContents.js
```diff
@@ -0,0 +1,271 @@
...
+
+/**
+ * Table of Contents pattern
+ *
+ * @param {object} props props object
+ * @param {string} props.layoutType layout type property ( 1/3 - 1/4 )
+ * @param {object} props.menuItems menu items object which includes title and id
+ * @param {element} props.children children property of table of contents pattern
+ * @param {object} props.marginTop margin top property which adds margin top value to layout
+ * @param {object} props.marginBottom margin bottom property which adds margin top value to layout
```
annawen1: does the value for `marginBottom` ever match the value for `marginTop` in the specs? Can we have one prop instead to set the two?

jeffchew: more of a @wonilsuhibm question, should we explicitly set the margin bottom and top? It's more of a convenience of the layout component, but wondering if it's something that authors would need to ever configure for the table of contents pattern itself?

packages/patterns-react/src/patterns/TableOfContents/TableOfContents.js
```diff
@@ -0,0 +1,271 @@
...
+    selectedTitle: '',
+  };
+
+  /**
+   *
+   *
+   * @memberof TableOfContents
+   */
+  componentDidMount() {
+    window.addEventListener('scroll', this.isInViewport);
```
jeffchew: use `window-or-root` instead of explicit `window`.

packages/patterns-react/src/patterns/TableOfContents/TableOfContents.js
```diff
@@ -0,0 +1,271 @@
...
+    window.removeEventListener('scroll', this.isInViewport);
+  }
+
+  /**
+   * Check whether content is in view
+   *
+   * @memberof TableOfContents
+   */
+  isInViewport = () => {
+    const items = document.querySelectorAll('[data-driverlocation]');
```
jeffchew: use `window-or-root` when referencing `document`

packages/styles/scss/patterns/tableofcontents/_tableofcontents.scss
```diff
@@ -0,0 +1,116 @@
+//
+// Copyright IBM Corp. 2016, 2018
+//
+// This source code is licensed under the Apache-2.0 license found in the
+// LICENSE file in the root directory of this source tree.
+//
+
+@import '@carbon/grid/scss/grid.scss';
+@import '@carbon/ibmdotcom-styles/scss/components/layout/_layout.scss';
```
jeffchew: @sannminwin since you're already in the styles package, you can reference this relatively.

packages/patterns-react/src/patterns/TableOfContents/TableOfContents.js
jeffchew: The name `data-driverlocation` seems odd. Should it be something like `data-toc-anchor`?

I'm wondering also if this should be tightly coupled like this. The `menuItems` prop is passing in the list of menu items with corresponding `id`'s . I'm wondering if we could have them as anchor names instead? The content on the right would simple have `<a name="[anchorID]">`, and for menuItems it would be the `title` and `anchorID`. The AnchorID would also double as a standard HTML jump link.

packages/patterns-react/src/patterns/TableOfContents/TableOfContents.js
```diff
@@ -0,0 +1,271 @@
...
+
+    return featureFlag(
+      TABLE_OF_CONTENTS,
+      <section
+        data-autoid={`${stablePrefix}--tableofcontents`}
+        className={`${prefix}--tableofcontents`}>
+        <div className={`${prefix}--tableofcontents__mobile-menu`}>
+          <div className={`${prefix}--tableofcontents__mobile-menu__wrapper`}>
+            <span className={`${prefix}--tableofcontents__mobile-menu__label`}>
+              Jump to{' '}
```
jeffchew: `Jump to` shouldn't be hardcoded english. We should set as a prop, and can default to `Jump to`.

packages/patterns-react/src/patterns/TableOfContents/TableOfContents.js
sannminwin: Yeah authors may not even need to configure it. I will talk to Ke and propably remove these settings.

packages/patterns-react/src/patterns/TableOfContents/TableOfContents.js
sannminwin: I see what you mean but it is the attribute name that DDP team wants to use for link reference.

packages/patterns-react/src/patterns/TableOfContents/TableOfContents.js
sannminwin: Yeah, We will still need to update it.

packages/patterns-react/src/patterns/TableOfContents/TableOfContents.js
jeffchew: @sannminwin that makes this pattern specific to Drupal then, which is something we should avoid.

sannminwin: > The name `data-driverlocation` seems odd. Should it be something like `data-toc-anchor`?
> 
> I'm wondering also if this should be tightly coupled like this. The `menuItems` prop is passing in the list of menu items with corresponding `id`'s . I'm wondering if we could have them as anchor names instead? The content on the right would simple have `<a name="[anchorID]">`, and for menuItems it would be the `title` and `anchorID`. The AnchorID would also double as a standard HTML jump link.

I see,
content on the right side is unknow to us. The only identifier is data-driverlocation attribute and this generated content won't have id or being able to add it. 
one thing we can do to loosely couple the selector is passing in selector property into pattern. Author will be able to choose appropiate selector and pass in as prop.

sannminwin: and we can use anchorID  as prop name for menuItems.

jeffchew: > content on the right side is unknow to us. The only identifier is data-driverlocation attribute and this generated content won't have id or being able to add it.
> one thing we can do to loosely couple the selector is passing in selector property into pattern. Author will be able to choose appropiate selector and pass in as prop.

I think passing in the selector info as part of the props might work, then that means the left nav doesn't have to worry about what to detect. So the `menuItems` would then pass in the `title` and the `uniqueSelector` for each item instead of `id`. The other thought, it could be anchor name. The left side would then be content with things like `<a name="MyAnchorName"></a><h2>Lorem Ipsum</h2>`, then the `menuItems` would then pass in as:

```javascript
[
  {
    anchorName: 'MyAnchorName',
    title: 'Lorem Ipsum',
  },
  ...
]
```

And query selectors would be `root.document.querySelector("a[name='${menuItems[i].anchorName}']")`

sannminwin: I am coming to your desk to talk about it.

jeffchew: As discussed, we'd like to use anchor links with the name attribute (for pure HTML jump links) as a way to identify the actual jump link in the content section. The TOC component can detect based on the anchor name that's passed in as the menuItems. We could even explore if the TOC component can just detect all of the anchor names in the content well so the TOC links do not have to be manually defined, but that can be handled separately (but can lead to that).

MVIMG_20191113_170710

packages/patterns-react/src/patterns/TableOfContents/TableOfContents.js
sannminwin: removed

jeffchew:
@sannminwin ,
@wonilsuhibm and @larahanlon2 spoke to me yesterday as well, they actually see the table of contents as a component and not a pattern.
cc: @photodow

packages/patterns-react/src/patterns/TableOfContents/TableOfContents.js
```diff
@@ -0,0 +1,159 @@
...
+  layoutProps['type'] = '1-3';
+  layoutProps['marginTop'] = 'none';
+  layoutProps['marginBottom'] = 'none';
+
+  const props = {};
+  props['menuItems'] = menuItems;
+  props['selectedId'] = selectedId;
+  props['selectedTitle'] = selectedTitle;
+  props['menuLabel'] = menuLabel;
+  props['updateState'] = updateState;
```
jeffchew: @sannminwin This might be more of a preference thing, but it may be more terse to do it this way:

const props = {
  menuItems,
  selectId,
  selectedTitle,
  menuLabel,
  updateState,
};

packages/patterns-react/src/patterns/TableOfContents/TableOfContents.js
```diff
@@ -0,0 +1,159 @@
...
+   */
+  const updateState = (id, title) => {
+    setSelectedId(id);
+    setSelectedTitle(title);
+  };
+
+  const layoutProps = {};
+  layoutProps['type'] = '1-3';
+  layoutProps['marginTop'] = 'none';
+  layoutProps['marginBottom'] = 'none';
```
jeffchew: Also preference:

```
const layoutProps = {
  type: '1-3',
  marginTop: 'none',
  marginBottom: 'none',
}
```

packages/patterns-react/src/patterns/TableOfContents/DesktopMenu.js
```diff
@@ -0,0 +1,95 @@
...
+   * @param {Array} items menu items array
+   * @param {string} activeId selected item id
+   * @returns {*} JSX Object
+   */
+  const renderMenuItems = (items, activeId) => {
+    return items.map(item => {
+      if (item) {
+        return (
+          <li
+            key={item.id}
```
jeffchew: @sannminwin let's also add a stable selector for the items as well.

jeffchew:
@sannminwin not urgent, but can you quickly test how this renders on the server side? You can use the `nextjs` example and link the package for testing.

packages/patterns-react/src/patterns/TableOfContents/TableOfContents.js
sannminwin: updated

packages/patterns-react/src/patterns/TableOfContents/TableOfContents.js
sannminwin: updated, thanks

packages/patterns-react/src/patterns/TableOfContents/DesktopMenu.js
sannminwin: added stable selector

sannminwin: added

packages/patterns-react/src/internal/FeatureFlags.js
```diff
@@ -33,9 +33,17 @@ export const LISTSECTION =
...
 export const SIMPLELONGFORM =
   process.env.SIMPLELONGFORM === 'true' || DDS_FLAGS_ALL || false;
+
+/**
+ * This determines if the tableofcontents will be rendered or not
+ *
+ * @type {string | boolean}
+ */
+export const TABLE_OF_CONTENTS =
+  process.env.TABLE_OF_CONTENTS === 'true' || DDS_FLAGS_ALL || false;
```
jeffchew: @sannminwin we can remove from `patterns-react` since this was moved to `react`, right?

packages/react/.env.example
```diff
@@ -23,6 +23,7 @@ FOOTER_LOCALE_BUTTON=<Boolean flag to turn on/off the locale selector>
 BUTTON_GROUP=<Boolean flag to turn on/off the ButtonGroup component>
 CARD_LINK=<Boolean flag to turn on/off the CardLink component>
 MASTHEAD_L1=<Boolean flag to turn on/off the masthead L1>
+TABLE_OF_CONTENTS=<Boolean flag to turn on/off the TableOfContents component>
```
jeffchew: rename to `DDS_TOC`, or `DDS_TABLE_OF_CONTENTS`

packages/react/src/components/TableOfContents/README.md
```diff
@@ -0,0 +1,97 @@
...
+> 💡 Don't forget to import the tableofcontents styles from
+> [@carbon/ibmdotcom-styles](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/styles).
+
+#### Feature Flags
+
+To utilize the following features, set the following variable's to `true` within
+your `.env` file or your application build settings.
+
+```
+TABLE_OF_CONTENTS=true
```
jeffchew: rename to `DDS_TOC`, or `DDS_TABLE_OF_CONTENTS`

sannminwin: I like a shorter one.

packages/react/.env.example
sannminwin: updated

packages/patterns-react/src/internal/FeatureFlags.js
sannminwin: removed

packages/react/.env.example
jeffchew: @sannminwin missed one: `DDS_TOC`

packages/react/src/components/TableOfContents/TableOfContents.js
```diff
@@ -0,0 +1,159 @@
...
+
+import PropTypes from 'prop-types';
+import React, { useState, useEffect } from 'react';
+import {
+  featureFlag,
+  settings as ddsSettings,
+} from '@carbon/ibmdotcom-utilities';
+import { settings } from 'carbon-components';
+import root from 'window-or-global';
+import { TABLE_OF_CONTENTS } from '../../internal/FeatureFlags';
```
jeffchew: @sannminwin need to update this one: `DDS_TOC`

packages/react/src/components/TableOfContents/__stories__/TableOfContents.stories.js
```diff
@@ -0,0 +1,155 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { TABLE_OF_CONTENTS } from '../../../internal/FeatureFlags';
```
jeffchew: @sannminwin need to update this one: `DDS_TOC`

packages/react/src/internal/FeatureFlags.js
```diff
@@ -46,3 +46,11 @@ export const BUTTON_GROUP =
  */
 export const CARD_LINK =
   process.env.CARD_LINK === 'true' || DDS_FLAGS_ALL || false;
+
+/**
+ * This determines if the tableofcontents will be rendered or not
+ *
+ * @type {string | boolean}
+ */
+export const TABLE_OF_CONTENTS =
```
jeffchew: let's name it `DDS_TOC` to be consistent with the environment flag name

packages/react/src/components/TableOfContents/__stories__/TableOfContents.stories.js
sannminwin: hey Jeff,
correct me if I am wrong. 
We are export as TABLE_OF_CONTENTS from FeatureFlads.js right ?

`export const TABLE_OF_CONTENTS =
  process.env.DDS_TOC === 'true' || DDS_FLAGS_ALL || false;
`

packages/react/.env.example
sannminwin: yeah

packages/react/src/components/TableOfContents/TableOfContents.js
sannminwin: updated.

packages/react/src/components/TableOfContents/__stories__/TableOfContents.stories.js
sannminwin: updated.

packages/react/src/internal/FeatureFlags.js
sannminwin: updated.

packages/styles/scss/ibm-dotcom-styles.scss
```diff
@@ -26,3 +26,4 @@
 @import 'patterns/leadspace/leadspace';
 @import 'patterns/listsection/listsection';
 @import 'patterns/simplelongform/simplelongform';
+@import 'patterns/tableofcontents/tableofcontents';
```
jeffchew: @sannminwin isn't this under `components` now?

packages/styles/scss/components/tableofcontents/README.md
```diff
@@ -0,0 +1,24 @@
+### SCSS
+
+#### Usage
+
+Import the package css into the top of your main CSS file.
+
+```css
+@import '@carbon/ibmdotcom-styles/scss/patterns/tableofcontents/index';
```
jeffchew: @sannminwin should update the reference to `components` instead of `patterns` folder

sannminwin: updated.

packages/styles/scss/ibm-dotcom-styles.scss
sannminwin: updated.

packages/react/src/components/TableOfContents/TableOfContents.js
```diff
@@ -30,6 +30,7 @@ const { prefix } = settings;
  * @returns {*} JSX Object
  */
 const TableOfContents = ({ menuItems, children, menuLabel }) => {
+  console.log('menuItems', menuItems);
```
jeffchew: @sannminwin remove this console log

sannminwin: did it

PR #668, opened 2019-11-12
Feat/example apps

jeffchew:
### Related Ticket(s)

No related issues

### Description

This adds a couple of example applications when using the IBM.com Library. Right now there are two, `create-react-app` and `nextjs`.

### Changelog

**New**

- example apps


packages/react/examples/nextjs/pages/index.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+// import Head from 'next/head'
+import '@carbon/ibmdotcom-styles/scss/ibm-dotcom-styles.scss';
+import { DotcomShell } from '@carbon/ibmdotcom-react';
+
+/**
+ * Initializes the homepage
+ */
+const Home = () => <DotcomShell navigation="default"></DotcomShell>;
```
annawen1: Should we add any child content for the DotcomShell? otherwise it'll throw an error since I think children is required

annawen1: Also I think I ran into eslint errors earlier after I pulled your branch and built and tested - should we include the build folder in the eslintignore file?

jeffchew: Sure, I can add that in.

jeffchew: Added with https://github.com/carbon-design-system/ibm-dotcom-library/pull/668/commits/52dd158426410cedde6e9835004b510f990bb9e0

packages/react/examples/create-react-app/scripts/build.js
```diff
@@ -0,0 +1,200 @@
...
+
+      console.log('File sizes after gzip:\n');
+      printFileSizesAfterBuild(
+        stats,
+        previousFileSizes,
+        paths.appBuild,
+        WARN_AFTER_BUNDLE_GZIP_SIZE,
+        WARN_AFTER_CHUNK_GZIP_SIZE
+      );
+      console.log();
```
emerson-pereira: Not sure if `console.log` is here on purpose or leftover?

jeffchew: This is the standard build script when ejecting `create-react-app`, I largely left it alone.

PR #671, opened 2019-11-12
docs(masthead): updated documentation for the masthead to clarify props

jeffchew:
### Related Ticket(s)

No related issue

### Description

These are tweaks to the Masthead README to better clarify how to use the props.

### Changelog

**Changed**

- Masthead README

### DIFF
--- a/packages/react/src/components/Masthead/README.md
+++ b/packages/react/src/components/Masthead/README.md
@@ -37,28 +37,22 @@ MASTHEAD_L1=true
 > [.env.example](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/react/.env.example)
 > for more information
 
-## Navigation Types
+## navigation (String | Object | null)
 
-| Name      | Description                          |
-| --------- | ------------------------------------ |
-| `default` | Default navigation data from IBM.com |
-| `custom`  | Custom navigation data               |
-| `none`    | No navigation                        |
+| Behavior           | Data Type | Description                                 | Example                             |
+| ------------------ | --------- | ------------------------------------------- | ----------------------------------- |
+| default navigation | String    | Default navigation data from IBM.com        | `<Masthead navigation="default" />` |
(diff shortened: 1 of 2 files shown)


packages/react/src/components/Masthead/README.md
```diff
@@ -37,40 +37,42 @@ MASTHEAD_L1=true
 > [.env.example](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/react/.env.example)
 > for more information
 
-## Navigation Types
+## navigation (String | Object | null)
```
emerson-pereira: May be a bit silly, but shouldn't we use backticks like `navigation` and `platform`?

packages/react/src/components/Masthead/README.md
```diff
@@ -37,40 +37,42 @@ MASTHEAD_L1=true
...
+| custom navigation  | Object    | Pass in custom navigation data as an object | `<Masthead navigation={myNavObj}/>` |  |
+| none               | null      | No navigation                               | `<Masthead />`                      |
 
 > 💡 `Custom` navigation data must follow the same structure and key names as
 > `default`. See
 > [this](https://www.ibm.com/common/v18/js/data/jsononly/usen.json) for an
 > example.
 
-## Options
+## platform (Object)
```
emerson-pereira: May be a bit silly, but shouldn't we use backticks like `navigation` and `platform`?

packages/react/src/components/Masthead/README.md
jeffchew: I'm fine with that but as long as we stay consistent. I don't think we backtick other README props that are headlines. [For example](https://github.com/carbon-design-system/ibm-dotcom-library/tree/master/packages/react/src/components/Layout)

We can probably do a huddle topic on README formatting, and see if there's a consistent approach we'd like to take.

jeffchew: @emerson-pereira added as a huddle topic for next week.

Issue #642, opened 2019-11-07
DotcomShell flashes at page load (briefly renders without styles)


jcpalaci:
## Detailed description

> Describe in detail the issue you're having.

When using the DotcomShell as suggested in code examples (https://github.com/carbon-design-system/ibm-dotcom-library/tree/master/packages/react/src/components/DotcomShell#dotcomshell-1), once in a while (specially if cache is disabled), I am able to see the page "flashing" on page load. The page briefly renders without styles and then styles are applied.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

General issue

> Is this issue related to a specific component?

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Load DotcomShell (and page) without the initial "flashing"

> What browser are you working in?

FireFox, Chrome, Safari

> What version of the IBM.com Library are you using?

1.1.0

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

IBM.com search

## Steps to reproduce the issue

1. Load any page using DotcomShell
2. Disable cache
3. Reload several times

## Additional information

- Screenshots or code

![image](https://user-images.githubusercontent.com/57272532/68413017-fd7d1300-0152-11ea-9b69-5f099d30dab9.png)

![image](https://user-images.githubusercontent.com/57272532/68413062-1259a680-0153-11ea-9304-9affcb16c775.png)

- Notes

I'm guessing that this is because styles are being loaded "on demand". Is it a better way to load them to avoid this issue?

annawen1:
This seems to happen on SSR or reload, the DotcomShell looks fine when I tested this out on our CRA and nextJS setup. On storybook, importing `@carbon/ibmdotcom-styles` in the storybook `Container.js` file seems to solve this issue.

jeffchew:
Adding in notes from Nixon on search team:

> This is a cosmetic issue, it can be ugly at times.  Affects slow loading users more severely on first visit/cache busted more than repeat visitors.  Go to the debug window on the browser and in the network tab, tick, disable cache, and reload the page a few times, that way it is more reproducible and you can see more clearly what the impact is cosmetic wise: https://ibm-com-search-beta.mybluemix.net/search?lang=en&cc=us&q=spss
>
> Since it doesn't affect the operation of the site, I don't think it is super urgent, but it is a very nice to have...

PR #672, opened 2019-11-12
fix(storybook): render with styles on SSR/reload

annawen1:
### Related Ticket(s)

[DotcomShell flashes at page load (briefly renders without styles) #642 ](https://github.com/carbon-design-system/ibm-dotcom-library/issues/642)

### Description

In storybook, when refreshing - the DotcomShell/Masthead/Footer renders first without styles

BEFORE:
style issue

AFTER:
with styles

### Changelog

**New**

- import styles in the `Container.js` file

### DIFF
--- a/packages/react/.storybook/Container.js
+++ b/packages/react/.storybook/Container.js
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import './polyfills';
 import './_container.scss';
 import { settings } from 'carbon-components';
+import '@carbon/ibmdotcom-styles';
 
 const { prefix } = settings;
 export default class Container extends Component {


packages/react/.storybook/Container.js
```diff
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import './polyfills';
 import './_container.scss';
 import { settings } from 'carbon-components';
+import '@carbon/ibmdotcom-styles';
```
jeffchew: @annawen1 if we're bringing in the whole styles package, we probably don't need to bring in the individual scss files in the component stories right?

PR #674, opened 2019-11-12
fix(masthead): shrink masthead L1 when it becomes sticky

### LINKED ISSUE #146: Masthead L1 should shrink to 48px height when fixed to top on scroll
When the L1 becomes sticky on scroll (and the L0 disappears), the L1 should shrink to 48px height. Like so:

Screen Shot 2019-10-02 at 3 56 28 PM

When the `eyebrow` is active then leave the L1 at 80px, fixed to the top, just like it is right now.

emerson-pereira:
### Related Ticket(s)

Fixes #146 

### Description

Shrinks down Masthead L1 to the the same height it shrinks when the screen becomes smaller (`carbon--breakpoint-down(lg)`) by sharing some css rules between modifier CSS classes and the media query through Sass mixins.

**Masthead L1 default**

Masthead L1 default

**Masthead L1 short**

Masthead L1 short

### Changelog

**New**

- CSS mixins to share some rules between media query for smaller screens and modifier CSS classes

**Changed**

- Masthead L1 becomes smaller when sticky
- Masthead Eyebrow becomes hidden when Masthead L1 becomes sticky


packages/styles/scss/components/masthead/_masthead-l1.scss
```diff
@@ -12,6 +12,14 @@
...
 $search-transition-timing: 95ms;
 
+@mixin mastheadL1Short {
+  height: 48px;
+}
+
+@mixin eyebrowHidden {
+  display: none;
+}
+
```
emerson-pereira: I thought I'd use mixins here since we have to use the same rule to **media query** + **css modifier classes**. I'm not sure it's the ideal solution thought I wanted to hear thoughts on this.

annawen1: hmm I see what you're doing but it looks weird to me since they're only one line

packages/styles/scss/components/masthead/_masthead-l1.scss
```diff
@@ -12,6 +12,14 @@
 $search-transition: cubic-bezier(0.2, 0, 0.38, 0.9);
 $search-transition-timing: 95ms;
 
+@mixin mastheadL1Short {
+  height: 48px;
```
annawen1: Maybe use `carbon--mini-units(6)` instead of 48px

annawen1:
Screen Shot 2019-11-12 at 4 42 34 PM

The blue border is pushed out when it shrinks

packages/styles/scss/components/masthead/_masthead-l1.scss
emerson-pereira: 48px is actually how it is right now with the media query, but I could implement mini-units, I guess.

packages/styles/scss/components/masthead/_masthead-l1.scss
emerson-pereira: Yeah this was an attempt to use the same rule for both media query and class since it's not possible comma combine the two. But yeah it doesn't feel to good to me either. Maybe re-think the style usage for the sticky nav...

packages/styles/scss/components/masthead/_masthead-l1.scss
emerson-pereira: Done, thanks for the suggestion!

packages/styles/scss/components/masthead/_masthead-l1.scss
```diff
@@ -12,6 +12,28 @@
...
+
+@mixin short-masthead-l1__title {
+  line-height: inherit;
+  padding-bottom: $spacing-05;
+}
+
+@mixin short-masthead-l1__title-after {
+  bottom: 0;
+}
+
```
emerson-pereira: I still can't think of anything other than mixins for sharing the same css without duplicating. I'd appreciate if any one suggests a different approach.

annawen1:
Also I think if we do have an eyebrow above "StockCharts" like we do now, we don't want to shrink it to 48px when on scroll and instead leave as is at 80px

packages/styles/scss/components/masthead/_masthead-l1.scss
jeffchew: Not sure if it would be cleaner to just have a single parent modifier class that targets each of these elements? seems a bit overengineered to have several mixins for this...

emerson-pereira:
> Also I think if we do have an eyebrow above "StockCharts" like we do now, we don't want to shrink it to 48px when on scroll and instead leave as is at 80px

@annawen1 Than this defeats the purpose of the the PR, are saying this is not necessary at all?

jeffchew:
> > Also I think if we do have an eyebrow above "StockCharts" like we do now, we don't want to shrink it to 48px when on scroll and instead leave as is at 80px
> 
> @annawen1 Than this defeats the purpose of the PR, are saying this is not necessary at all?

Can we get design team to weigh in? @wonilsuhibm @larahanlon2 @oliviaflory the intent is for the eyebrow to disappear when the L1 shrinks?

packages/react/src/components/Masthead/MastheadL1.js
```diff
@@ -20,19 +21,31 @@ const { prefix } = settings;
...
+    [`${prefix}--masthead__l1-name-title--shortl1`]: isShort,
   });
 
   return (
     <div className={className}>
-      <div className={`${prefix}--masthead__l1-name`}>
-        <span className={`${prefix}--masthead__l1-name-eyebrow`}>
+      <div className={`${prefix}--masthead__l1-name ${nameShortL1}`}>
+        <span
+          className={`${prefix}--masthead__l1-name-eyebrow ${eyebrowShortL1}`}>
```
annawen1: @emerson-pereira going to leave my comment about the eyebrow shrinking here since it'll be easier to form a thread this way.

From my understanding of the ticket: https://github.com/carbon-design-system/ibm-dotcom-library/issues/146
`When the eyebrow is active then leave the L1 at 80px, fixed to the top, just like it is right now.` 
It seems we don't shrink L1 to 48px for all cases, if there is an eyebrow (it is hardcoded as of now, you can set it up so it is passed through as a prop - that would help with testing) we don't want to shrink it 

@larahanlon2 can you clarify?

packages/react/src/components/Masthead/MastheadL1.js
```diff
@@ -20,19 +21,31 @@ const { prefix } = settings;
...
           <ArrowLeft16 />
           <a href="#">Eyebrow</a>
```
annawen1: can you pass the href values through a prop instead?

packages/react/src/components/Masthead/MastheadL1.js
```diff
@@ -20,19 +21,31 @@ const { prefix } = settings;
...
         </span>
-        <span className={`${prefix}--masthead__l1-name-title`}>
+        <span className={`${prefix}--masthead__l1-name-title ${titleShortL1}`}>
           Stock Charts
```
annawen1: also pass this in through a prop instead

packages/react/src/components/Masthead/MastheadL1.js
annawen1: @jeffchew @emerson-pereira cross out what I said 😅we do always shrink to 48px

from @larahanlon2:
> - we have the eyebrow in the L1 - L1 height will be 80px when not sticky, but will shrink to 48px when sticky (eyebrow is hidden)
> - we don’t have the eyebrow in L1 - L1 height will be 80px when not sticky, but will shrink to 48px when sticky

packages/styles/scss/components/masthead/_masthead-l1.scss
emerson-pereira: @jeffchew Are you suggesting contextual selector here? Like defining css rules for elements under `.bx--masthead__l1--short`?

jeffchew: yes, that's what I was thinking. Adding other folks here to chime in.
cc: @kennylam @annawen1 @photodow @sannminwin @andysherman2121

packages/react/src/components/Masthead/MastheadL1.js
emerson-pereira: done

packages/react/src/components/Masthead/MastheadL1.js
emerson-pereira: done

emerson-pereira:
@jeffchew @annawen1 implemented:

- contextual selector
- eyebrow and title as props

wonilsuhibm:
The shrinking and hiding the eye brow upon scroll was intentional. When scrolling only local navigation should be visible, and since the eye brow takes the user out of the local context, its not local any more and should hide.

- Wonil sent from iPhone.

packages/react/src/components/DotcomShell/__stories__/DotcomShell.stories.js
```diff
@@ -34,7 +34,9 @@ storiesOf('Dotcom Shell', module)
         )}
         footerType={select('Footer', footer, footer.default)}
         hasProfile={boolean('Has profile', true)}
-        hasSearch={boolean('Has search', true)}>
+        hasSearch={boolean('Has search', true)}
+        title={text('Title', 'Stock Charts')}
+        eyebrowLink={text('Eyebrow link', '#')}>
```
jeffchew: @emerson-pereira since the L1 is still behind a feature flag, can you conditionally add these props in the story with the feature flag?

packages/react/src/components/Masthead/MastheadL1.js
```diff
@@ -20,21 +21,20 @@ const { prefix } = settings;
...
   return (
     <div className={className}>
       <div className={`${prefix}--masthead__l1-name`}>
         <span className={`${prefix}--masthead__l1-name-eyebrow`}>
           <ArrowLeft16 />
-          <a href="#">Eyebrow</a>
-        </span>
-        <span className={`${prefix}--masthead__l1-name-title`}>
-          Stock Charts
+          <a href={eyebrowLink}>Eyebrow</a>
```
jeffchew: @emerson-pereira Eyebrow label should probably be a prop as well

packages/react/src/components/Masthead/MastheadL1.js
```diff
@@ -20,21 +21,20 @@ const { prefix } = settings;
...
         </span>
+        <span className={`${prefix}--masthead__l1-name-title`}>{title}</span>
       </div>
       <HeaderNavigation className={`${prefix}--masthead__l1-nav`} aria-label="">
         <HeaderMenuItem href="#">Link 1</HeaderMenuItem>
```
jeffchew: @kennylam @emerson-pereira I was originally going to ask about getting the navigation items as props, but I think is out of the scope of this PR/issue and also it's still not concrete yet, right?

emerson-pereira: @jeffchew @kennylam I rolled back props out of this PR and left it strictly for short L1 fix. I'll create a new PR for props after this.

emerson-pereira:
Merged fixes:

- Implemente props in Masthead L1 component
- Implement conditional props in Masthead L1 storybook per feature flag.

larahanlon2:
Hey @jeffchew great to see L1 shrink now. Thank you @emerson-pereira ! Looking at this, there are a number of things that are not per specs or need to be tweaked. Should we open a new issue or PR for design feedback and approval?

wonilsuhibm:
FYI, here is a recording of an interctive demo that shows how the masthead should shrink.
https://ibm.box.com/s/cr7rqohnsidgonf77zvgwx1ye36ox6pr

annawen1:
> FYI, here is a recording of an interctive demo that shows how the masthead should shrink.
> https://ibm.box.com/s/cr7rqohnsidgonf77zvgwx1ye36ox6pr

looks like we need to add some animation so it shrinks smoothly

jeffchew:
> > FYI, here is a recording of an interctive demo that shows how the masthead should shrink.
> > https://ibm.box.com/s/cr7rqohnsidgonf77zvgwx1ye36ox6pr
> 
> looks like we need to add some animation so it shrinks smoothly

This should be a separate story.

wonilsuhibm:
It's more of not-completely-scrolling-off the screen than animation. Let me know if this makes sense.

Issue #469, opened 2019-10-13
Pattern: develop use cases pattern


jeffchew:
_hahnrob created the following on Sep 19:_  

 

#### User Story

> As a `[user role below]`:
developer using the ibm.com Library

> I need to:
have a use case pattern that can be used

> so that I can:
leverage it within my application

#### Additional information

 - Story within Storybook with corresponding knobs
 - minimum 80% unit test coverage
 - Utilize Carbon
- [Web Simplification design & specs](https://ibm.ent.box.com/folder/93008842052)
 - [Pattern: Use cases specs from web simplification #468](https://app.zenhub.com/workspaces/ibmcom-library-5d449f3642eb1962336cbe52/issues/carbon-design-system/ibm-dotcom-library/468)

#### Acceptance criteria
 - [ ] tested React component within the Patterns Storybook

 

_Original issue: https://github.ibm.com/webstandards/digital-design/issues/1738_

RobertaJHahn:
@sannminwin I have updated the use case dev issue with the box folder link provided by Wonil.

sannminwin:
Thanks @RobertaJHahn

sannminwin:
@RobertaJHahn 
@ljcarot 
@jeffchew 

Hello Team,
Pull request is created for use cases pattern.

[PR Link](https://github.com/carbon-design-system/ibm-dotcom-library/pull/689)

PR #689, opened 2019-11-14
feat(pattern): use cases pattern

sannminwin:
### Related Ticket(s)

#469 Develope Use Cases pattern

use-cases-pattern

### Description
This pull request creates use cases pattern.

Deploy Preview:

### Changelog

**New**

- use cases pattern


packages/patterns-react/src/patterns/UseCases/UseCasesGroup.js
```diff
@@ -0,0 +1,101 @@
...
+        srcSet={`${img.uri.md}, ${img.uri.lg} 2x`}
+      />
+      <source
+        media="(min-width: 320px)"
+        srcSet={`${img.uri.sm}, ${img.uri.md} 2x`}
+      />
+      <img src={img.uri.md} srcSet={`${img.uri.sm}`} alt={img.alt}></img>
+    </picture>
+  );
+};
```
jeffchew: Not urgent for this PR, but I can see this as a standalone component that can be reused.

packages/patterns-react/src/internal/FeatureFlags.js
```diff
@@ -33,9 +33,17 @@ export const LISTSECTION =
...
 export const SIMPLELONGFORM =
   process.env.SIMPLELONGFORM === 'true' || DDS_FLAGS_ALL || false;
+
+/**
+ * This determines if the usecases will be rendered or not
+ *
+ * @type {string | boolean}
+ */
+export const USECASES =
+  process.env.USECASES === 'true' || DDS_FLAGS_ALL || false;
```
jeffchew: Per our huddle this morning, let's rename to `DDS_USECASES`

packages/patterns-react/src/patterns/UseCases/UseCasesGroup.js
sannminwin: Created new issue.
https://github.com/carbon-design-system/ibm-dotcom-library/issues/699

packages/patterns-react/src/internal/FeatureFlags.js
sannminwin: updated.

ltitle:
It looks like the title and subheads are not using the right text sizes. The section title should be $expressive-heading-05 and the subheads should be $expressive-heading-04. 

It is hard to tell but also please confirm that the image is spanning 8 columns on desktop.

I do not see the card CTAs for the section. Are you coding that into this pattern?

Vertical spacing looks right on desktop but needs to be reduced for medium and small breakpoints (see specs)

jeffchew:
> It looks like the title and subheads are not using the right text sizes. The section title should be $expressive-heading-05 and the subheads should be $expressive-heading-04.
> 
> It is hard to tell but also please confirm that the image is spanning 8 columns on desktop.
> 
> I do not see the card CTAs for the section. Are you coding that into this pattern?
> 
> Vertical spacing looks right on desktop but needs to be reduced for medium and small breakpoints (see specs)

Thank you @ltitle , @sannminwin can you look into these?

@ltitle I am thinking of merging this as-is and then these feedback can be addressed in a separate PR, just so that the DPP Pilot team can be unblocked for this pattern. Think that would be ok?

sannminwin:
> It looks like the title and subheads are not using the right text sizes. The section title should be $expressive-heading-05 and the subheads should be $expressive-heading-04.
> 
> It is hard to tell but also please confirm that the image is spanning 8 columns on desktop.
> 
> I do not see the card CTAs for the section. Are you coding that into this pattern?
> 
> Vertical spacing looks right on desktop but needs to be reduced for medium and small breakpoints (see specs)

Hi Lila,
Thanks for having a look.
There were 4 versions of specs. This pattern is updated based on the latest version of specs that you provided. Headings are coded as mentioned in specs. Card Link is there  but it seems not rendering because of feature flag bug. I will try to fix now. Here I will include some screenshots of code for typography that you mention. Please let me know if you see any specific discrepency so that we can address them as new issues in the future.

.
xlg-1312px-use case section

Screen Shot 2019-11-19 at 9 56 18 AM

Screen Shot 2019-11-19 at 9 57 03 AM

jeffchew:
@ltitle updates appear to be in now, can you double check? If all looks good we will merge in.

Issue #634, opened 2019-11-06
Masthead (and DotcomShell) briefly renders and disappears


jcpalaci:
## Detailed description

> Describe in detail the issue you're having.

When I try to implement `DotcomShell` or `Masthead`, the component doesn't render properly, it briefly shows on screen but then it disappears. I can also replicate this behavior in https://ibmdotcom-react.netlify.com/

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

Masthead (and DotcomShell)

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Masthead (and DotcomShell) rendering properly

> What browser are you working in?

Safari, Chrome, Firefox

> What version of the IBM.com Library are you using?

1.1.0

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

IBM.com search

## Steps to reproduce the issue

See Additional information

## Additional information
 - **The scope of this ticket will be to create a hotfix to default to us-EN if there are any rendering issues based on invalid translation data.**

- Screenshots or code

Using this test page:
```
import React from 'react';
import { DotcomShell } from '@carbon/ibmdotcom-react';
import '@carbon/ibmdotcom-styles/scss/components/masthead/index.scss';
import '@carbon/ibmdotcom-styles/scss/components/footer/index.scss';
import '@carbon/ibmdotcom-styles/scss/components/dotcom-shell/_dotcom-shell.scss';

export default class App extends React.Component {
  render() {
    return (
      <DotcomShell navigation={'default'} footerType="short">
        Hi there!
      </DotcomShell>
    );
  }
}
```
I'm getting this error (`Cannot read property 'megapanelContent' of undefined`):
![image](https://user-images.githubusercontent.com/57272532/68311906-df8eb000-0077-11ea-90a2-57b2690084e0.png)

If I use `navigation={null}` that seems to clear the error. But if I add a `lang="en-US"` in the HTML tag, then I see a different error (Cannot read property 'trim' of undefined):
![image](https://user-images.githubusercontent.com/57272532/68313585-6d6b9a80-007a-11ea-988e-be12a118e833.png)

If I disable the search box (`hasSearch={false}`), then the Masthead shows OK this time.

- Notes

I'm located in Guadalajara, Mexico; working from home. Later today will test in the office (where usually I got an IP from the US)

jeffchew:
[...]

PR #691, opened 2019-11-14
Fix/masthead json render

kennylam:
### Related Ticket(s)

#634 

### Description

**Staging:**
https://deploy-preview-693--ibmdotcom-react-staging.netlify.com/?path=/story/masthead--default

This fixes an issue in rendering `Masthead`/`MastheadLeftNav` navigation from translation files from IBM.com. In this update, navigation subnavs are populated from top-level megamenu items. Non-megamenu subnavs are unchanged.

Screen Shot 2019-11-14 at 4 10 24 PM

Screen Shot 2019-11-14 at 4 10 00 PM

### Changelog

**Changed**

- Update navigation rendering to use top-level megamenu links
- Update `MastheadLeftNav` with the same render changes
- CSS updates to IBM logo padding and link color in SideNav.

### DIFF
--- a/packages/react/src/components/Masthead/MastheadLeftNav.js
+++ b/packages/react/src/components/Masthead/MastheadLeftNav.js
@@ -36,7 +36,7 @@ const MastheadLeftNav = ({ navigation, isSideNavExpanded }) => {
    * @returns {*} Left side navigation
    */
   const sideNav = navigation.map((link, i) => {
-    if (link.hasMenu && link.hasMegapanel) {
+    if (link.hasMenu) {
       return (
         <SideNavMenu title={link.title} key={i}>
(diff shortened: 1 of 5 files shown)


kennylam:
**Staging:** https://deploy-preview-693--ibmdotcom-react-staging.netlify.com/?path=/story/masthead--default

kennylam:
> @kennylam looks good for the patch, though eventually we should transform the data from the service instead. Perhaps we should do that before hitting up the vanilla masthead.

@jeffchew Yep I'll make a story for that and get it in. Only reason I didn't for this PR is because I'll probably need to transform some of the Profile data into the Masthead data as well and I'd rather get this issue fixed first.

PR #695, opened 2019-11-15
fix(buttongroup): adding columns to ButtonGroup story

andysherman2121:
### Related Ticket(s)

https://github.com/carbon-design-system/ibm-dotcom-library/issues/559

### Description

Adding columns to the ButtonGroup story to show how to resize the buttons when adjusting to different screen sizes

### Changelog

**New**

- Columns

**Changed**

- Added columns to the ButtonGroup story file, and added instructions to the README.md file


packages/react/src/components/ButtonGroup/README.md
```diff
@@ -22,29 +22,38 @@ import React from 'react';
...
+      </div>
+    </div>
+  </div>;
 }
 ReactDOM.render(<App />, document.querySelector('#app'));
 ```
 
-> 💡 Don't forget to import the button group styles from
+> Utilize the grid system to resize the buttons to fit the desired sizes 💡
+> Don't forget to import the button group styles from
```
jeffchew: @andysherman2121 I'm wondering if this needs a better explanation of what's going on here. There's nothing stopping users from using the `ButtonGroup` as-is, the only effect they are getting with the grid wrapper is to let the buttons shrink when the text is smaller. We can have one example without any grid wrapper, and another with it. And in tandem, show the difference between the two (add some text above each example)

andysherman2121: Sounds good, i'll adjust

packages/react/src/components/ButtonGroup/README.md
```diff
@@ -15,7 +15,47 @@ Here's a quick example to get you started.
 @include carbon--font-face-sans();
 ```
 
-> 💡 Only import font's once per usage
+> 💡 Only import font's once per usage 💡 This example utilizes the grid system
+> in order to adjust the button size based on the text size 💡 Adjust columns as
+> needed
```
jeffchew: Sort of nitpicky, can the two examples be separated by perhaps H3 titles, e.g.:

```
### Example

This is the base example `ButtonGroup`. Note, the buttons will not resize based on text size.

...code snippet...

### Example with Buttons based on text size

This example utilizes the grid system in order to adjust the button size based on the text size. Adjust the columns as needed.

...code snippet...
```

PR #696, opened 2019-11-15
fix(typeahead): dynamically set lc/cc from getLang for search typeahead

jeffchew:
### Related Ticket(s)

No related issues

### Description

This is a quick fix to dynamically set the lc/cc combination for the search typeahead service.

### Changelog

**Changed**

- Set lc/cc from LocaleAPI.getLang in SearchTypeaheadAPI

### DIFF
--- a/packages/services/src/services/Locale/Locale.js
+++ b/packages/services/src/services/Locale/Locale.js
@@ -112,8 +112,12 @@ class LocaleAPI {
   static getLang() {
     if (root.document.documentElement.lang) {
       const lang = root.document.documentElement.lang.toLowerCase();
-      const codes = lang.split('-');
-      return { cc: codes[1], lc: codes[0] };
+      if (lang.indexOf('-') === -1) {
+        return _localeDefault;
+      } else {
+        const codes = lang.split('-');
+        return { cc: codes[1], lc: codes[0] };
+      }
     } else {
       return _localeDefault;
     }
--- a/packages/services/src/services/Locale/Locale.js
+++ b/packages/services/src/services/Locale/Locale.js
@@ -21,8 +21,8 @@ const _proxy = process.env.CORS_PROXY || '';
  * @private
  */
 const _localeDefault = {
-  lc: 'us',
-  cc: 'en',
+  lc: 'en',
+  cc: 'us',
 };
 
 /**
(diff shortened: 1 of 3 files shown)


packages/services/src/services/Locale/Locale.js
```diff
@@ -114,8 +114,8 @@ class LocaleAPI {
       const lang = root.document.documentElement.lang.toLowerCase();
       const codes = lang.split('-');
       return { cc: codes[1], lc: codes[0] };
-      //    } else {
-      //      return _localeDefault;
```
annawen1: oh can you also add check here that both `cc` and `lc` are defined from the `lang` attr? If either are undefined then we'll return default for now. For cases where the `lang` is just "en" like we saw from one of the adopters.

jeffchew: Yup, added here: https://github.com/carbon-design-system/ibm-dotcom-library/pull/696/commits/e5d31ec15b476461aa0a5d9ee3899b85d83db647

packages/services/src/services/SearchTypeahead/SearchTypeahead.js
```diff
@@ -39,11 +40,10 @@ class SearchTypeaheadAPI {
...
   static async getResults(query) {
-    const lc = 'en'; // TODO: create utility for fetching lc
-    const cc = 'us'; // TODO: create utility for fetching cc
-    const url = `${_endpoint}?lang=${lc}&cc=${cc}&query=${encodeURIComponent(
-      query
-    )}`;
+    const lang = LocaleAPI.getLang();
+    const url = `${_endpoint}?lang=${lang.lc}&cc=${
+      lang.cc
+    }&query=${encodeURIComponent(query)}`;
```
emerson-pereira: Optionally, maybe make url query a bit easier to read. Not sure if is over-engineering though.

```
const urlQuery = [
      `lang=${lang.lc}`,
      `cc=${lang.cc}`,
      `query=${encodeURIComponent(query)}`,
    ].join('&');
const url = `${_endpoint}?${urlQuery}`;
```

jeffchew: No all good, I like that pattern. I made the adjustment in https://github.com/carbon-design-system/ibm-dotcom-library/pull/696/commits/93ab1a74917d702fc7eefd7c6d569fe9c07c17d5.

PR #715, opened 2019-11-20
feat(footer): Vanilla Footer (updated) (with readme)

andysherman2121:
### Related Ticket(s)

https://github.com/carbon-design-system/ibm-dotcom-library/issues/356

### Description

Creates vanilla versions of tall and short footers. 

Screen Shot 2019-11-20 at 2 33 54 PM
Screen Shot 2019-11-20 at 2 34 02 PM

### Changelog

Vanilla footer


packages/vanilla/src/components/footer/__stories__/footer.stories.js
```diff
@@ -0,0 +1,28 @@
...
+  //     sidebar: readme
+  //   },
+  // })
+  .add('Default', () => {
+    const footerTypeOptions = {
+      tall: '',
+      short: 'short',
+    };
+    setTimeout(() => {
+      console.log('DOMContentLoaded');
```
jeffchew: @andysherman2121 can you remove the console log here

andysherman2121: 👍

packages/vanilla/src/components/footer/footer.template.js
```diff
@@ -0,0 +1,87 @@
...
+}
+/**
+ * renders logo and bottom section of footer
+ *
+ * @param {string} type type of footer in use
+ * @returns {object} JSX object
+ */
+const footerTemplate = ({ type }) => {
+  return `
+  <footer data-autoid="dds--footer" class="bx--footer ${setFooterType(type)}">
```
jeffchew: @andysherman2121 can you pull in prefix for all of the `bx--` classnames here, as well as `dds--` for the stableselector prefix

packages/vanilla/src/components/footer/footerNav.template.js
```diff
@@ -0,0 +1,98 @@
+/**
+ * renders footer nav for tall
+ *
+ * @returns {object} JSX object
+ */
+function footerNavTemplate() {
+  return `
+  <nav data-autoid="dds--footer-nav" class="bx--footer-nav">
+    <ul class="bx--accordion bx--footer-nav__container">
```
jeffchew: @andysherman2121 update all of the classname and stable selector prefixes here to pull from settings.

packages/vanilla/src/components/footer/footer.template.js
```diff
@@ -0,0 +1,87 @@
...
+ * @param {string} El type of footer in use
+ * @returns {object} JSX object
+ */
+function footerCreate(El) {
+  const accordionEl = El.querySelector('.bx--accordion');
+  if (accordionEl) {
+    // eslint-disable-next-line
+    const accordionInit = Accordion.create(accordionEl);
+  }
+}
```
jeffchew: I'm wondering if we should pull the initialize code into its own file, and keep the templates in the `.template.js` files. We should probably follow that model I think especially when we hit more complex funcitonality like in the Masthead. We can decide if we want to create as an ES6 class with an init, or maybe something simpler like this

packages/vanilla/src/components/footer/__stories__/footer.stories.js
```diff
@@ -0,0 +1,28 @@
+import { storiesOf } from '@storybook/html';
+import { withKnobs, select } from '@storybook/addon-knobs';
+import footer from '../footer.template';
```
jeffchew:
```
import footerTemplate from '../footer.template';
import Footer from '../footer';
```

packages/vanilla/src/components/footer/__stories__/footer.stories.js
```diff
@@ -0,0 +1,28 @@
...
+      tall: '',
+      short: 'short',
+    };
+    setTimeout(() => {
+      console.log('DOMContentLoaded');
+      const footerElement = document.querySelector('.bx--footer');
+      // eslint-disable-next-line
+      const footerInstance = footer.create(footerElement);
+    }, 5000);
+    return footer.template({
```
jeffchew:
```
setTimeout(() => {
 const footerElement = document.querySelector('.bx--footer');
 Footer.init(footerElement);
}, 5000);'
return footerTemplate({ ... });

File: packages/vanilla/src/components/footer/index.js [hunk 7]
```diff
@@ -0,0 +1,10 @@
+/**
+ * Copyright IBM Corp. 2016, 2018
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+export { default as Footer } from './footer';
```
jeffchew: @andysherman2121 make sure to export from `/src/components/index.js` as well.

packages/vanilla/src/components/footer/footer.template.js
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * renders optional footer nav for tall
+ *
+ * @param {string} type type of footer in use [tall|short]
+ * @param {object} footerMenu Nav items
+ * @returns {object} JSX object
+ * @private
+ */
+function _optionalFooterNav(type, footerMenu) {
+  console.log('optionalFooterNav', type);
```
jeffchew: @andysherman2121 get rid of this console log.

packages/vanilla/src/components/footer/README.md
```diff
@@ -0,0 +1,119 @@
...
+@include carbon--font-face-mono();
+@include carbon--font-face-sans();
+```
+
+> 💡 Only import font's once per usage
+
+```javascript
+import { Footer } from '@carbon/ibmdotcom-react';
+import 'yourapplication.scss';
+import '@carbon/ibmdotcom-styles/scss/components/footer/index.scss';
```
jeffchew: @andysherman2121 move this to under `yourapplication.scss` instead:

```
```scss
// yourapplication.scss
@import '@carbon/type/scss/font-face/mono';
@import '@carbon/type/scss/font-face/sans';
@include carbon--font-face-mono();
@include carbon--font-face-sans();

@import '@carbon/ibmdotcom-styles/scss/components/footer/index';
```

packages/vanilla/src/components/footer/README.md
```diff
@@ -0,0 +1,119 @@
...
+    globalInit();
+
+    if (El) {
+      Accordion.create(El);
+    }
+  }
+}
+```
+
+#### Feature Flags
```
jeffchew: @andysherman2121 Vanilla will probably never have feature flags. This is only for React.

packages/vanilla/src/components/footer/README.md
```diff
@@ -0,0 +1,119 @@
...
+FOOTER_LOCALE_BUTTON=true
+```
+
+> See
+> [feature-flags.md](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/react/docs/feature-flags.md)
+> and
+> [.env.example](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/react/.env.example)
+> for more information
+
+## Types (optional)
```
jeffchew: @andysherman2121 so @annawen1 just updated a bunch of READMEs, can you doublecheck the latest on `master`?

packages/vanilla/src/components/footer/README.md
```diff
@@ -0,0 +1,119 @@
...
+| `dds--footer-nav-group`       | Component   |
+| `dds--footer-nav-group__link` | Interactive |
+| `dds--footer-logo`            | Component   |
+| `dds--footer-logo__link`      | Interactive |
+| `dds--footer-locale-btn`      | Interactive |
+| `dds--legal-nav`              | Component   |
+| `dds--legal-nav__link`        | Interactive |
+| `dds--locale-modal`           | Component   |
+
+## Server Side Rendering
```
jeffchew: @andysherman2121 this is the React version of the server side rendering. You should add in details of how to make the service call then applying to the ES6 template literal. I would even not call it `server-side rendering` and have more of a section on how to fetch the navigation data (probably reference and show how to use the `getFooterWithData` method in the Footer class)

packages/vanilla/src/components/footer/README.md
```diff
@@ -0,0 +1,119 @@
...
+  const response = await TranslationAPI.getTranslation(lang);
+
+  return footerTemplate({
+    type,
+    footerMenu: response.footerMenu,
+    footerThin: response.footerThin,
+  });
+}
+}
+
```
jeffchew: @andysherman2121 rather than including the whole method, I would say just to tell adopters to use the method itself:

```
import { Footer } from '@carbon/ibmdotcom-vanilla';

const content = Footer.getFooterWithData(); // fetches the default footer content, then returns the footer markup
const elem = document.getElementById('yourFooterDiv');
elem.innerHTML = content; // sets the footer content into the element
Footer.init(elem); // initializes the footer
```

andysherman2121: Ah gotcha, makes sense

packages/vanilla/src/components/footer/README.md
```diff
@@ -0,0 +1,119 @@
...
+```
+
+> 💡 Don't forget to import the footer styles from
+> [@carbon/ibmdotcom-styles](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/styles).
+
+### Initialize
+
+> Init example below
+
+```javascript
```
jeffchew: I wouldn't print the init function here, but tell the user what the init function does (it initializes the accordion view for mobile). Then tell the user how to initialize it:

```
import { Footer, footerTemplate } from '@carbon/ibmdotcom-vanilla';

const content = footerTemplate({
  type: 'tall',
  footerMenu: { ...footer menu content ... }
  footerThin: { ... footer thin content ... }
}); // returns the markup
const elem = document.getElementById('yourFooterDiv');
elem.innerHTML = content; // sets the footer content into the element
Footer.init(elem); // initializes the footer
```

packages/vanilla/src/components/footer/README.md
```diff
@@ -0,0 +1,119 @@
...
+@include carbon--font-face-mono();
+@include carbon--font-face-sans();
+
+import '@carbon/ibmdotcom-styles/scss/components/footer/index.scss';
+```
+
+> 💡 Only import font's once per usage
+
+```javascript
+import { Footer } from '@carbon/ibmdotcom-react';
```
jeffchew: @andysherman2121 this is referencing React. Also, I think that the initialize example below should suffice for the description for the non-data fetching. You should probably bring up the data fetch approach up, and even I would say that should be the preferred method over the pure template one.

andysherman2121: @jeffchew is there any reason they should be using the pure template approach ever?

jeffchew: some teams might want to use it. Drupal team for instance is doing that for the Masthead already, and is asking to do that for the footer too. They might want to have their own way of fetching the navigation data and injecting it.

andysherman2121: got it

packages/vanilla/src/components/footer/README.md
```diff
@@ -17,39 +17,61 @@ Here's a quick example to get you started.
...
+  type: 'tall',
+  footerMenu: { ...footer menu content ... }
+  footerThin: { ... footer thin content ... }
+}); // returns the markup
+const elem = document.getElementById('yourFooterDiv');
+elem.innerHTML = content; // sets the footer content into the element
+Footer.init(elem); // initializes the footer
+```
 
 ### Initialize
```
jeffchew: @andysherman2121 you can probably remove `initialize` and `fetch navigation data` sections.

andysherman2121: Yeah i was thinking that was redundant after adding the methods

andysherman2121: @jeffchew updated

Issue #560, opened 2019-10-23
[CardLink] CSS inheritance issues with the CardLink and Link


jeffchew:
@jeffchew I'm already aware of that issue, and figured since it's behind a feature flag, and @sannminwin needs it sooner rather then later I'll figure it out in a seperate PR. It appears to be less of an issue with the card itself and more with how storybook is building it, but I'm still looking into it. You can see my comment above in the description.

> *Note: there is a known bug in this at the moment. Pulling in the `CardLink` by itself should be fine, but the way/order storybook is pulling in the `footer` with themed `g90` `bx--link` component it's overriding some of the styling since Carbon added the `.bx--link` class to the clickable tiles. I'm looking into how we can fix this, and will put in a follow up pull request.*
>
> *To test with proper styling pull down locally and temporarily comment out [this line](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/styles/scss/components/link/_link-expressive.scss#L8)*
> ```scss
> ...
> // @import '~carbon-components/scss/components/link/link';
> @import '../../globals/vars.scss';
> @import '../../globals/mixins.scss';
> ...
> ```

_Originally posted by @photodow in https://github.com/carbon-design-system/ibm-dotcom-library/pull/554#issuecomment-545027396_

photodow:
I put together a reduced test case of what appears to be happening.

Link is being loaded after tile is getting loaded in CSS.
https://codesandbox.io/s/zealous-austin-voxbe?fontsize=14

Here is what happens when it gets loaded before the tile.
https://codesandbox.io/s/strange-night-12bgk?fontsize=14

it's messing with margins, colors and text-decorations.

RobertaJHahn:
@kennylam This is a sprint 22 issues that is still open.  Can I close it or should I move it to sprint 23?

Issue #507, opened 2019-10-13
Locale Selector:  develop the geographic overlay 


jeffchew:
_hahnrob created the following on Oct 04:_  

 

#### User Story

> As a `[user role below]`:
As a developer using the Digital Design System

> I need to:
have a geographic overlay that enables a user to select their geography

> so that I can:
include this component in the Locale Selector in the web site I am developing

#### Additional information

- This issue is dependent on the design and the [package architecture https://github.com/carbon-design-system/ibm-dotcom-library/issues/507](https://zenhub.ibm.com/app/workspaces/digital-design-system-5b3a815a8bf6a932efa6fd03/issues/webstandards/digital-design/860) 
- [final design and specs](https://ibm.box.com/s/isie20a89socbo9ael687u6pb75uz9f8)
              - Includes:  Functional labels, Visual redlines, Behavior and interaction specs, Desktop and mobile
 - [Box link to Mary O'Donnell's functional specs](https://ibm.ent.box.com/folder/88263129224)
 - [develop the clickable tile component #504](https://app.zenhub.com/workspaces/ibmcom-library-5d449f3642eb1962336cbe52/issues/carbon-design-system/ibm-dotcom-library/504)
 - [Locale Selector: develop a service that determines the users locale based on ipcinfo or Geo-ip #515](https://app.zenhub.com/workspaces/ibmcom-library-5d449f3642eb1962336cbe52/issues/carbon-design-system/ibm-dotcom-library/515)
 - [Expand translation service to other languages mimicking the Northstar design #421](https://app.zenhub.com/workspaces/ibmcom-library-5d449f3642eb1962336cbe52/issues/carbon-design-system/ibm-dotcom-library/421)

#### Acceptance criteria
 - [ ] Component has been tested using Netlify
 - [ ] The component development follows a modular component pattern
 - [ ] The component should be accessible via ES6 import
 - [ ] The component should be visible within the living documentation of the Dotcom components
 - [ ] Presented and approved Product Owner

 

_Original issue: https://github.ibm.com/webstandards/digital-design/issues/1808_

PR #725, opened 2019-11-21
Feat/locale filter

kennylam:
### Related Ticket(s)

#506, #507, #560

### Description

Adds Locale Modal component. The component uses the TranslationAPI to compare against locales included in a page's `<link />` tags and creates a list of regions, countries, and languages based on what is available.

### Changelog

**New**

- Locale modal with region, country, and language filters

**Changed**

- `LocaleModal`, `Footer`, and `CardLink` stories now use local `scss` imports to avoid duplicates. Once fully tested all components will import styles this way.

**Removed**

- `LocaleModal`, `Footer`, and `CardLink` direct `scss` imports from stories file removed. All now import one local file for import.


packages/react/src/components/LocaleModal/LocaleModal.js
```diff
@@ -0,0 +1,199 @@
...
+   *
+   * @returns {object} list item
+   */
+  const sortList = list => {
+    const pageLangs = altlangs();
+    const filterList = [];
+
+    list.regionList &&
+      list.regionList.map((region, index) => {
+        let isAmericas = region.name.indexOf('America') > -1;
```
jeffchew: @kennylam I think this would only work for the english translated countrylist, right? `America` is translated into the corresponding language in other countrylist files.

kennylam: @jeffchew At the moment, yes. Not sure how we're going to match 'North America` and 'Latin America' in dozens of languages to merge the lists. Been giving it some thought but haven't come up with a good solution yet.

kennylam: @jeffchew actually one thing I can do is change it to match region keys instead, `na` and `la`

jeffchew: One thing I would even do is to not worry about regions, and just have the front-end be a dumb interface that iterates through the regions. We will eventually go back to the data and fix the data to have NA and LA to be merged together, I don't think that the UI layer should be doing that logic. I believe I already merged usen to have NA and LA merged already on the data side (unless you're pointing to an older data file), and we will round back to the other country list files and merge those later.

jeffchew: @kennylam https://1.www.s81c.com/common/js/dynamicnav/www/countrylist/jsononly/usen-utf8.json

packages/react/src/components/LocaleModal/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Stable selectors
+
+| Name                                   | Description |
+| -------------------------------------- | ----------- |
+| `dds--locale-modal`                    | Component   |
+| `dds--locale-modal__geo-btn-${region}` | Interactive |
+| `dds----locale-modal__filter`          | Interactive |
+| `dds----locale-modal__locales`         | Interactive |
+| `dds----locale-modal__region-back`     | Interactive |
```
jeffchew: there are double dashes here

packages/styles/scss/components/card-link/index.scss
```diff
@@ -9,6 +9,7 @@
 @import '../../globals/vars';
 @import '../../globals/mixins.scss';
 
+@import '~carbon-components/scss/components/link/link';
```
jeffchew: @kennylam we can't import with `~`, as that's a webpack specific thing

packages/styles/scss/components/locale-modal/_locale-modal.scss
```diff
@@ -0,0 +1,183 @@
...
+// This source code is licensed under the Apache-2.0 license found in the
+// LICENSE file in the root directory of this source tree.
+//
+
+@import '../../globals/imports';
+@import '../../globals/vars.scss';
+@import '../../globals/mixins.scss';
+@import '../card-link/index';
+
+@import '~carbon-components/scss/components/search/search';
```
jeffchew: Need to remove the `~`

larahanlon2:
@kennylam BTW if you go to Carbon pictograms and select `Location` from the dropdown you'll find the correct globes for locale. https://www.carbondesignsystem.com/guidelines/pictograms/library

Is it possible to load a different version of the icon depending on the users location? So if someone is in France the Europe Africa icon would show but if someone is in Chicago the Americas icon would show?

packages/react/src/components/Footer/LocaleButton.js
```diff
@@ -67,27 +33,11 @@ const LocaleButton = ({ selectItem }) => {
         United States — English
```
jeffchew: @kennylam as discussed, let's get this as well as the modal heading to be dynamically set based on the html lang attribute.

jeffchew:
@kennylam There may be a bug with the way the country/language combinations are filtering. For countries that have more than one language option, only one seems to be showing up in the search results (e.g. Canada should have English and French).

kennylam:
> @kennylam There may be a bug with the way the country/language combinations are filtering. For countries that have more than one language option, only one seems to be showing up in the search results (e.g. Canada should have English and French).

@jeffchew Good catch, thanks. This is fixed.

oliviaflory:
> @kennylam BTW if you go to Carbon pictograms and select `Location` from the dropdown you'll find the correct globes for locale. https://www.carbondesignsystem.com/guidelines/pictograms/library
> 
> Is it possible to load a different version of the icon depending on the users location? So if someone is in France the Europe Africa icon would show but if someone is in Chicago the Americas icon would show?

You'll probably need the icon version, Pictograms aren't meant to be smaller than 32px or something like that. In the [Icon page](https://www.carbondesignsystem.com/guidelines/icons/library) search "Earth" and it has the different versions Lara mentioned. 

Or [Carbon icons repo](https://github.com/carbon-design-system/carbon/tree/master/packages/icons/svg/32) has the SVGs

packages/styles/scss/components/card-link/index.scss
jeffchew: @kennylam not sure if it's something on my end, this was marked as resolved but I don't see a change applied.

packages/services/src/services/Locale/Locale.js
```diff
@@ -118,8 +117,6 @@ class LocaleAPI {
         const codes = lang.split('-');
         return { cc: codes[1], lc: codes[0] };
       }
-    } else {
-      return _localeDefault;
```
jeffchew: @kennylam this should be added back if the lang attribute is not set, no? the `getLang` method is being used in other places as well (e.g. setting the `lc/cc` for the Masthead SearchTypeahead API)

cc: @annawen1

packages/react/src/components/LocaleModal/LocaleModal.js
```diff
@@ -0,0 +1,165 @@
...
+/**
+ * @property propTypes
+ * @description Defined property types for component
+ * @type {{isOpen: boolean, setIsOpen: boolean, headerLabel: string, headerTitle: string}}
+ */
+LocaleModal.propTypes = {
+  isOpen: PropTypes.bool,
+  setIsOpen: PropTypes.bool,
+  headerLabel: PropTypes.string,
+  headerTitle: PropTypes.string,
```
jeffchew: @kennylam The `headerTitle` will need to be set based on the `getLang` display name, as well as in the button label in the footer. If this can't make code freeze, we at least need to track this as a hotfix before we go live with `v1.2.0`.

cc: @annawen1

kennylam: @jeffchew yes all of these are going to be put into the translation files and were not captured in the scope of this story. Need to get all of translatable fields into the json

packages/services/src/services/Locale/Locale.js
kennylam: @jeffchew @annawen1  Returning a default language there will result in the first condition as always true. Cookies will never be get/set from the second condition.
https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/services/src/services/Locale/Locale.js#L70

I'll need to take a look and see how it affects MastheadSearch

kennylam: Cleared this up for @annawen1 and she explained that the second condition is sort of old functionality that is being left there for the time being. Will revert to default locale

Issue #713, opened 2019-11-19
Fix dependency cycle in Services/Utilities


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

Lerna is throwing the following error during builds:

```
lerna notice cli v3.16.4
lerna info versioning independent
lerna info Executing command in 6 packages: "yarn run ci-check"
lerna WARN ECYCLE Dependency cycles detected, you should fix these!
lerna WARN ECYCLE @carbon/ibmdotcom-services -> @carbon/ibmdotcom-utilities -> @carbon/ibmdotcom-services
```

This probably has to do with the two packages probably importing classes from each other. We may need to think about how we can approach fixing this.

## Steps to reproduce the issue

1. Run `yarn lerna run ci-check`
2. See warning near beginning of build

PR #726, opened 2019-11-22
Fix/713 circular dependencies

jeffchew:
### Related Ticket(s)

#713 

### Description

This fixes the circular dependency issues between services and utilities. The major change is the global initializer for the services, which has been moved into the services package instead. 

In addition, prefixes have been added to the horizontal rule for the vanilla component. This was added in to test and confirm that adding the utilities package to the vanilla package would work without any issues.

### Changelog

**Changed**

- services now has the global init for the services called
- prefixes added to the horizontal rule in the vanilla package


jeffchew:
@andysherman2121 once this is in we can apply this to the footer and merge in as well.

jeffchew:
Not sure why system check hasn't come back, but it shows as clear here: 
https://app.circleci.com/jobs/github/jeffchew/ibm-dotcom-library/450

PR #731, opened 2019-11-22
fix(leadspace): move theming to pattern level as prop

annawen1:
### Related Ticket(s)

[[LeadSpace] Update styles to apply color theme as a container modifier #732](https://github.com/carbon-design-system/ibm-dotcom-library/issues/732)

### Description

Pass theme to pattern as prop instead of using a wrapper

### Changelog

**Changed**

- create mixin for themed items and to use in the carbon theme mixin
- pass theme as a prop


packages/styles/scss/patterns/leadspace/_leadspace.scss
```diff
@@ -87,6 +119,12 @@
     }
   }
 
+  .#{$prefix}--leadspace--g100 {
+    @include carbon--theme($carbon--theme--g100) {
+      @include themed-items;
+    }
+  }
+
```
jeffchew: greatsuccess

Is there a default in case the adopter doesn't include the modifier class?

annawen1: you mean if they don't pick the `g100` theme? it defaults to the white theme

PR #741, opened 2019-11-25
fix(locale): update locale endpoint

kennylam:
### Description

Update `LocaleAPI` endpoint to use latest JSON. New JSON merges North/Latin America regions into one region.

### Changelog

**Changed**

- update endpoint to use latest JSON
- update `LocaleAPI` tests to use new endpoint
- update `LocaleAPI` mock data to reflect new JSON changes


packages/services/src/services/Locale/Locale.js
```diff
@@ -6,7 +6,7 @@ import { ipcinfoCookie, geolocation } from '@carbon/ibmdotcom-utilities';
  * @constant {string | string} Host for the Locale API call
  * @private
  */
-const _host = process.env.TRANSLATION_HOST || 'https://www.ibm.com';
+const _host = process.env.TRANSLATION_HOST || 'https://1.www.s81c.com';
```
jeffchew: The `TranslationAPI` also uses the `TRANSLATION_HOST`, I don't think this affects it since it's pointing to what looks like an akamai host, but wanted to make sure?

jeffchew: Doublechecked, it seems to work:
https://1.www.s81c.com/common/v18/js/data/jsononly/usen.json

Issue #712, opened 2019-11-19
[Styles] Foundation of proxy approach for the expressive theme


jeffchew:
#### User Story

> As a `[user role below]`:
IBM.com Library developer

> I need to:
have the footprint/facility for proxying carbon's styles

> so that I can:
begin applying the expressive theme across all IBM.com Library components/patterns and Carbon Components within my application

#### Additional information

 - after working with the Carbon team, the initial approach of theming within Carbon core has more fundamental issues which will require more strategy in approach
- Carbon team suggested a way to "proxy" the carbon packages within IBM.com Library, where the theme can be applied through our library before adopters begin using carbon components. This is essentially an override layer, but within IBM.com Library
 - #709 has been created to bring in the carbon packages as package dependencies, which will aid in locking down carbon versions that will be proxied
- Will need to tap Josh Black from Carbon team for consultation/support to ensure that the approach is aligned with eventually "ejecting" and going with the core approach with Carbon once that is ready
- Can possibly choose an example component that can be themed using this approach
- May want to consider a separate storybook of carbon components using the expressive theme, perhaps a separate section within the react storybook rather than a new one

#### Acceptance criteria
 - [ ] Foundation approach for the proxied style theme
 - [ ] Storybook of Carbon components, ready to take in the expressive theme

jeffchew:
@joshblack pinging you here to track based on what we discussed, I added some additional thoughts on this after our meeting.

cc: @RobertaJHahn @wonilsuhibm @ljcarot

PR #750, opened 2019-11-27
Feat/712 expressive theme

jeffchew:
### Related Ticket(s)

#712 

### Description

This adds the foundation of the expressive theming for Carbon components. This will also apply to the IBM.com Library, though will need to determine what the level of effort is involved for making adjustments after this is put in place.

The current approach is to use CSS custom properties, as I found it was difficult to override the tokens in pre-compiled SASS after some struggling with Carbon.

### Changelog

**New**

- Storybook for Carbon with the expressive theme
- enabled custom properties feature flag

**Changed**

- Renamed temporary expressive styling of the carbon components to a temporary file
- moved temporary override styles into its own folder, also renamed the mixins
- added the expressive theme to the `patterns-react` and `react` storybooks, with environment variable to enable (off by default)
- adjusted temporary overrides to be disabled if the expressive theme feature flag is enabled


jeffchew:
@kennylam @annawen1 @photodow @emerson-pereira @andysherman2121 if you guys can take a look at this and start providing reviews, I think this should be close to what we can have as a foundation for the expressive theme proxy.

jeffchew:
@kennylam @photodow @annawen1 @emerson-pereira @andysherman2121 this PR is getting pretty big, so I'm setting as `ready for review`, we can iterate much more on this in future PRs.

cc: @wonilsuhibm @oliviaflory @larahanlon2

PR #761, opened 2019-12-02
feat(vanilla-masthead): implement vanilla masthead

emerson-pereira:
### Related Ticket(s)

Vanilla masthead


packages/vanilla/src/components/masthead/mastheadHeader.template.js
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @param {object} props Property object passed into the template
+ * @param {string} props.className Custom css classes passed into the root element
+ * @param {object} props.attrs Object with key/value HTML attributes passed into the root element
+ * @param {string} props.children ES6 template literal of the chidren content passed inside the root element
+ * @returns {string} ES6 template literal of the horizontal rule
+ *
+ */
+const mastheadHeaderTemplate = (props = {}) => {
+  const { className: customClassName, attrs, children } = props;
```
jeffchew: the pattern we've been using is have it be in the props and not deconstructed inside, so that it matches the jsdocs

packages/vanilla/package.json
```diff
@@ -42,6 +42,9 @@
     "@babel/runtime": "^7.5.5",
     "@carbon/ibmdotcom-styles": "1.1.2",
     "@carbon/ibmdotcom-utilities": "1.2.0-alpha.1070",
+    "@carbon/icon-helpers": "^10.4.0",
+    "@carbon/icons": "^10.6.1",
```
jeffchew: for now, these should be `peerDependencies` and not `dependencies`. There is a separate story to address the dependencies and tree shaking issues.

packages/vanilla/src/components/masthead/mashteadLeftNav.template.js
```diff
@@ -0,0 +1,11 @@
+/**
```
jeffchew: The filename appears to be misspelled.

packages/vanilla/src/components/masthead/masthead.template.js
```diff
@@ -0,0 +1,94 @@
...
+  `;
+
+  return `
+    <div class="${prefix}--masthead">
+      <div class="${prefix}--masthead__l0">
+        ${headerTemplate({
+          attrs: {
+            'aria-label': 'IBM',
+            'data-autoid': `${stablePrefix}--masthead`,
+          },
```
jeffchew: @emerson-pereira I think there's too much attribute passing where it's not necessary. Even children, those can probably be imported/references in the child templates and not passed in at the parent level.

packages/vanilla/src/components/masthead/mastheadHeaderMenuButton.template.js
```diff
@@ -0,0 +1,62 @@
...
+  const {
+    className: customClassName,
+    attrs = {},
+    isActive,
+    isCollapsible,
+  } = props;
+  const { 'aria-label': ariaLabel } = attrs;
+
+  // TODO: Check if should use root
+  window.onHeaderMenuButtonClick = () => {
```
jeffchew: Should use `window-or-root` to help determine. Should assume that adopters would be using this either client side or server side.

packages/vanilla/src/components/masthead/mastheadHeaderMenuButton.template.js
```diff
@@ -0,0 +1,62 @@
...
+
+  const icon = isActive ? toString(Close20) : toString(Menu20);
+
+  return `
+    <button
+      ${attrsTemplate}
+      ${titleTemplate}
+      class="${className}"
+      type="button"
+      onclick="onHeaderMenuButtonClick()">
```
jeffchew: I wouldn't bind this to window unless it's necessary

packages/vanilla/src/components/masthead/mastheadHeaderMenuItem.template.js
```diff
@@ -0,0 +1,40 @@
...
+ * @param {string} props.role Role for the underlying <li> node
+ * @param {string} props.children ES6 template literal of the chidren content passed inside the <span> element
+ * @returns {string} ES6 template literal of the Masthead Header Menu Item
+ *
+ */
+const mastheadHeaderMenuItemTemplate = (props = {}) => {
+  const { className: customClassName, attrs = {}, role, children } = props;
+
+  const attrsTemplate = Object.entries(attrs).reduce((acc, [key, value]) => {
+    return `${acc} ${key}="${value}"`;
```
jeffchew: as mentioned, I don't think passing attributes from parent is necessary.

Issue #637, opened 2019-11-06
Add data-autoid attributes to optional fields for Patterns


kexuibm:
### The problem
When the Drupal Core Team is generating twig templates from the IDL React Pattern Apps, some of the knobs in the pattern markup are optional fields.  When we convert these optional fields into Drupal Twig Templates we would need a selector for our automation to inject twig conditional statements to render that field's markup depending on whether the content for that field exists in Drupal.

### The solution
Per conversation with @jeffchew  the solution we propose is for IDL team to inject data-autoid for each optional field/knob at 2 places:

1.  The tag where that optional knob is being rendered
2.  The nearest parent div for that tag.

So for example in the Lead Space pattern there is the following markup generated by the react pattern:

<div class="bx--leadspace__row" > <p class="bx--leadspace__desc" data-autoid="leadspace__description">[KNOB VALUE]</p></div>

we would like to modify it to look like this:

<div class="bx--leadspace__row"  **data-autoid="leadspace__description_container"**> <p class="bx--leadspace__desc" **data-autoid="leadspace__description**">[KNOB VALUE]</p></div>
          

### Additional information
We would like for this feature to be propagated to all 6 patterns being delivered in the 12/6 release.

nmangold:
I'm not sure we need the data-autoid attribute on the (1) tag where the optional field is being rendered, or necessarily (2) the nearest parent. 

We need the data-autoid attribute, or a unique selector, on the outermost container that contains only the optional field. For several of the patterns I've seen so far, this is usually the row container. However, that would not be the case with inline elements.

The goal is to hide the container markup for the optional field if the field is empty, but not if the container contains other elements that may or may not  be empty.

jeffchew:
@RobertaJHahn fyi, Drupal team is requesting if we can fit this into our next sprint.

cc: @ljcarot @wonilsuhibm

PR #774, opened 2019-12-04
refactor(pattern): added data-autoid to optional fields and it's parent

sannminwin:
### Related Ticket(s)

#637 Add data-autoid attributes to optional fields for Patterns

### Description
This pull request adds data-autoid to optional fields and it's parent element.

### Changelog

**New**

- added data-autoid attribute to optional fields and it's parent element of  patterns.


packages/patterns-react/src/patterns/LeadSpace/LeadSpace.js
```diff
@@ -97,7 +97,9 @@ const LeadSpace = ({
     <section
       data-autoid={`${stablePrefix}--leadspace`}
       className={className(variation, theme)}>
-      <div className={`${prefix}--leadspace__container`}>
+      <div
+        data-autoid={`${stablePrefix}--leadspace__container`}
```
annawen1: This here seems kinda excessive to me, we already have an autoid in for the `section` element above

sannminwin: Yeah, I see your point. We already have it at parent level. And I just talked with Ke and confirmed that they need `data-autoid` on all optional fields and it's immediate parent.

jeffchew:
@sannminwin you probably need to update the corresponding READMEs too, right?

packages/patterns-react/src/patterns/LeadSpaceCentered/LeadSpaceCentered.js
```diff
@@ -98,12 +98,23 @@ const LeadSpaceCentered = ({
...
-          <p className={`${prefix}--leadspace--centered__desc`}>{copy}</p>
-          {buttons && buttons.length > 0 && <ButtonGroup buttons={buttons} />}
+          <p
+            data-autoid={`${stablePrefix}--leadspace--centered__desc`}
+            className={`${prefix}--leadspace--centered__desc`}>
+            {copy}
+          </p>
+          {buttons && buttons.length > 0 && (
+            <ButtonGroup
+              data-autoid={`${stablePrefix}--leadspace--centered__button-group`}
```
annawen1: I was going to do this a bit later but can you edit `ButtonGroup` to accept this prop? 😁 It doesn't actually set the data id here. Probably pass it in as `autoid` prop and then in `ButtonGroup` you set `data-autoid` there.

packages/patterns-react/src/patterns/LeadSpace/LeadSpaceButtons.js
```diff
@@ -40,7 +40,6 @@ const ButtonGroup = ({ buttons }) => {
     const button2 = buttonsRef.current[1];
 
     if (button1.current.offsetWidth != button2.current.offsetWidth) {
-      console.log('triggered');
```
annawen1: Thanks, that's been hanging out there for a while...😅

packages/patterns-react/src/patterns/LeadSpaceCentered/LeadSpaceCentered.js
sannminwin: The end goal is to add `autoid` to each button and it's parent.
It seems I have to work on ButtonGroup.

annawen1: Right, yeah you can pass in the base part of the autoid then.

```
<ButtonGroup autoid={`${stablePrefix}--leadspace--centered`) ..... />
```

and then in ButtonGroup.js:
```
const ButtonGroup = ({ buttons, autoid }) => {
  return featureFlag(
    DDS_BUTTON_GROUP,
    <ol
      className={`${prefix}--buttongroup`}
      data-autoid={`${autoid}--button-group`}>
      {buttons.map((button, key) => {
        return (
          <li key={key} className={`${prefix}--buttongroup-item`}>
            <Button
              data-autoid={`${autoid}--button-group-${key}`}
              {...button}
              kind={key === 0 ? 'primary' : 'tertiary'}>
              {button.copy}
            </Button>
          </li>
        );
      })}
    </ol>
  );
};
```

PR #776, opened 2019-12-05
fix(usecase): add theme modifier class to usecase pattern

annawen1:
### Related Ticket(s)

[[UseCases] Update styles to apply color theme as a container modifier #734](https://github.com/carbon-design-system/ibm-dotcom-library/issues/734)

### Description

Add theme modifier class to UseCase pattern

### Changelog

**Changed**

- pass `theme` as prop to UseCase pattern
- set `themed-items` mixin for styles that are using color tokens


packages/styles/scss/patterns/usecases/index.scss
```diff
@@ -3,10 +3,6 @@
...
-.#{$prefix}--usecases--g100 {
-  @include carbon--theme($carbon--theme--g100) {
-    @include usecases;
-    @include usecases-group;
-    @include usecases-item;
-  }
-}
+@include usecases;
+@include usecases-group;
+@include usecases-item;
```
jeffchew: Actually, do we need these if they are already getting included at the bottom of the files themselves?

sannminwin: I think this file is exporting all three scss files as one and we still need it. what we inlcuded at the bottom of each scss file is theme specific scss to apply to `themed items`

annawen1: Right, the UseCase story is still importing this index.scss file that includes these 3 files

jeffchew: Each file is doing their own `@include` using the `import-once` wrapper like this:

```scss
@include exports('usecases-item') {
  @include usecases-item;
}
```
 
So once you import the file, the mixin already gets fired off at the end of the file, right?

annawen1: But the story is referencing this index file here - don't we need to include all related usecase css files?

PR #777, opened 2019-12-05
fix(listsection): add theme modifier to ListSection pattern

annawen1:
### Related Ticket(s)

[[ListSection] Update styles to apply color theme as a container modifier #735](https://github.com/carbon-design-system/ibm-dotcom-library/issues/735)

### Description

Add theme modifier class to ListSection pattern

### Changelog

**Changed**

- pass `theme` as prop to ListSection pattern
- set `themed-items` mixin for styles that are using color tokens

### DIFF
--- a/packages/styles/scss/patterns/listsection/index.scss
+++ b/packages/styles/scss/patterns/listsection/index.scss
@@ -3,10 +3,6 @@
 @import './listsection-group';
 @import './listsection-item';
 
-.#{$prefix}--listsection--g100 {
-  @include carbon--theme($carbon--theme--g100) {
-    @include listsection;
-    @include listsection-group;
-    @include listsection-item;
-  }
-}
+@include listsection;
+@include listsection-group;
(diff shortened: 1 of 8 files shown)


packages/styles/scss/patterns/listsection/index.scss
```diff
@@ -3,10 +3,6 @@
...
-.#{$prefix}--listsection--g100 {
-  @include carbon--theme($carbon--theme--g100) {
-    @include listsection;
-    @include listsection-group;
-    @include listsection-item;
-  }
-}
+@include listsection;
+@include listsection-group;
+@include listsection-item;
```
jeffchew: I don't think these are necessary since they are already getting exported in the files themselves...

annawen1: donee

annawen1: donee

Issue #411, opened 2019-10-13
Centralize the IBM Logo for IBM.com


jeffchew:
_james-dow created the following on Sep 10:_  

 ### Detailed description

Not sure if this is necessarily a bug, but maybe more of a technical debt item that I wanted to make sure we didn't forget.

Ultimately we need to figure out how to best leverage and unify the IBM logo across IBM.com. Especially for our own components like the `masthead`, and `footer`. Right now we are using two different IBM logos for those components.  **When consolidating the logos, the banding issue must also be resolved.**

#### Additional Information
Currently, the SVG logo is showing banding issue on some monitors. It depends on the monitor's DPI, but here is a screen capture:

![Screen Shot 2019-10-04 at 6.26.04 PM.png](https://images.zenhubusercontent.com/5ad8a7dd4b5806bc2bcfcb61/a97a448b-435a-4ebc-9f4d-cd3db4921c28)

 - [Closed Banding issue #165](https://app.zenhub.com/workspaces/ibmcom-library-5d449f3642eb1962336cbe52/issues/carbon-design-system/ibm-dotcom-library/165)

#### Two logos
- [FooterLogo.js](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/react/src/components/Footer/FooterLogo.js)
- [IbmLogo](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/react/src/components/Icon/IbmLogo.js)

#### Questions
 - Can we leverage `@carbon/icons-react` for this?
 - Can we explore using an [SVG Sprite](https://css-tricks.com/svg-sprites-use-better-icon-fonts/)

#### Acceptance Criteria
 - [ ] Single logo is used for the Masthead and Footer
 - [ ] The IBM logo banding issue on some monitors is resolved.  
 

_Original issue: https://github.ibm.com/webstandards/digital-design/issues/1639_

jeffchew:
@photodow why was this pulled into this sprint? We didn't triage this or set a priority for this yet.

cc: @RobertaJHahn

photodow:
@jeffchew sorry if I did something out of order. I was told to cherry-pick items that could be done quickly. If that's not true, then I can return it to be triaged, no problem.

jeffchew:
[...]

PR #783, opened 2019-12-06
fix(logo): update masthead logo svg

kennylam:
### Related Ticket(s)

#411 

### Description
This updates the IBM Logo in the masthead component with a new asset to address banding issues when icon is scaled. Masthead/Footer icons have also been moved out of the `react` package and into `styles` 
so they are framework agnostic.

### Changelog

**New**

- add `babel-plugin-inline-react-svg` to allow direct import of svgs

**Changed**

- Updated Masthead IBM logo asset - no more banding
- masthead/footer logos moved to svg files in `styles/icons/svg`

**Removed**

- inline `svg` from masthead/footer components


packages/react/src/components/Footer/FooterLogo.js
```diff
@@ -8,6 +8,7 @@
 import React from 'react';
 import { settings as ddsSettings } from '@carbon/ibmdotcom-utilities';
 import { settings } from 'carbon-components';
+import FooterLogo from '../../../../styles/icons/svg/IBM-8bar-logo--h65-white.svg';
```
jeffchew: @kennylam not sure if this import will work since it's pulling in relatively across packages. If for some reason the adopter doesn't pull in the styles package this could break. The reference would have to be:
```
import { FooterLogo } from '@carbon/ibmdotcom-styles/icons/svg/IBM-8bar-logo--h65-white.svg'
```

packages/react/src/components/Icon/IbmLogo.js
```diff
@@ -8,6 +8,7 @@
 import React from 'react';
 import { settings as ddsSettings } from '@carbon/ibmdotcom-utilities';
 import { settings } from 'carbon-components';
+import MastheadLogo from '../../../../styles/icons/svg/IBM-8bar-logo--h23.svg';
```
jeffchew: Same here, needs to reference the styles package.

packages/styles/icons/svg/IBM-8bar-logo--h23.svg
```diff
@@ -0,0 +1,8 @@
...
+        </g>
+    </g>
+</svg>
```
jeffchew: @kennylam you will need to add the `icons` folder under `package.json` to be available in node_modules.

jeffchew:
```
"files": [
    "dist/**/*",
    "icons/**/*",
    "scss/**/*"
  ],
```

PR #784, opened 2019-12-06
fix(component): table-of-contents fixed dark theme class on parent element

RaphaelAmadeu:
### Related Ticket(s)

#733 

### Description

The dark theme class was moved from the parent element to be a prop of the component.
Both README files were updated to reflect this change.

### Changelog

**New**

- prop.theme

**Changed**

- Storybook story
- Component props
- README files
**Removed**

- Class from the parent element.


packages/styles/scss/components/tableofcontents/index.scss
```diff
@@ -3,6 +3,6 @@
 
 .#{$prefix}--tableofcontents--g100 {
   @include carbon--theme($carbon--theme--g100) {
-    @include tableofcontents;
+    @include themed-items;
   }
 }
```
jeffchew: @RaphaelAmadeu this wrapper can actually go away, since we have shifted to creating the theme modifiers as a modifier class and not a parent class. 

cc: @annawen1

packages/styles/scss/components/tableofcontents/_tableofcontents.scss
```diff
@@ -103,6 +121,7 @@
         display: block;
       }
     }
+    @include themed-items;
```
jeffchew: @RaphaelAmadeu you will also need to add the theme modifier class here, and not in the parent.
cc: @annawen1

packages/react/src/components/TableOfContents/README.md
```diff
@@ -67,10 +80,9 @@ DDS_TOC=true
 
 ## menuItems
 
-| Properties Name | Data Type | Description     |
-| --------------- | --------- | --------------- |
-| title           | String    | Menu title text |
-| id              | String    | Menu id         |
+| Name    | Data Type | Default value                                  | Description |
+| ------- | --------- | ---------------------------------------------- | ----------- |
+| `theme` | String    | `` | Set the dark theme if the value is `g100` |
```
annawen1: Can you create another table for theme instead and keep the original `menuItems` table?

like https://github.com/carbon-design-system/carbon-for-ibm-website/pull/149

Oh and add `theme` prop to the `Props` table above with the corresponding data (required, data type, default value, description)

packages/react/src/components/TableOfContents/TableOfContents.js
```diff
@@ -26,10 +27,11 @@ const { prefix } = settings;
  *
  * @param {object} props props object
  * @param {object} props.menuItems menu items object
+ * @param {string} props.theme selected theme string
  * @param {*} props.children children property of component
  * @returns {*} JSX Object
  */
-const TableOfContents = ({ menuItems, children, menuLabel }) => {
+const TableOfContents = ({ menuItems, children, theme }) => {
```
annawen1: are we not using `menuLabel` anymore?

annawen1: We need it for mobile

Screen Shot 2019-12-06 at 2 31 19 PM

annawen1: We need it for mobile

Screen Shot 2019-12-06 at 2 31 19 PM

packages/react/src/components/TableOfContents/TableOfContents.js
```diff
@@ -109,6 +111,16 @@ const TableOfContents = ({ menuItems, children, menuLabel }) => {
...
   };
 
+  /**
+   * Add the dark theme class if it's selected
+   *
+   * @param {string} selected selected theme string, g100 if dark empty if light
+   * @returns {string} get dark theme class if it's selected
+   */
+  const getSelectedTheme = selected => {
+    if (selected == 'g100') return `${prefix}--tableofcontents--g100`;
```
annawen1: Can we just copy reuse the same logic that we have for other patterns [here](https://github.com/carbon-design-system/ibm-dotcom-library/blob/287fde438bee8e1f3d973fea05e366399b7216d9/packages/patterns-react/src/patterns/UseCases/UseCases.js#L89)

So if other themes need to be added, we don't have to have another conditional for it

packages/react/src/components/TableOfContents/__stories__/TableOfContents.stories.js
```diff
@@ -46,124 +44,105 @@ if (DDS_TOC) {
...
-              scelerisque vel. Vestibulum ante ipsum primis in faucibus orci
-              luctus et ultrices posuere cubilia Curae; Donec fringilla dui
-              tellus, a pretium diam vehicula et. Etiam non vulputate augue.
-              Morbi laoreet diam dapibus sapien pellentesque tristique. Morbi id
-              nibh metus. Integer non scelerisque nisl.
-            </p>
-          </TableOfContents>
-        </div>
+        <TableOfContents
+          theme={select('theme', themes, themes['light (white)'])}
```
annawen1: I changed the object key name to just `white` but forgot to change the field here. 😅 Can you change `themes['light (white)']` to `themes.light`?

packages/styles/scss/components/tableofcontents/index.scss
annawen1: Right if you take a look at https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/styles/scss/patterns/usecases/_usecases.scss

and https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/styles/scss/patterns/usecases/index.scss

packages/react/src/components/TableOfContents/README.md
```diff
@@ -59,11 +72,12 @@ DDS_TOC=true
...
-| `menuLabel` | NO       | String    | `Jump to`     | Placeholder value for menu label                                                 |
+| Name        | Required | Data Type | Default Value                                | Description                                                                      |
+| ----------- | -------- | --------- | -------------------------------------------- | -------------------------------------------------------------------------------- |
+| `children`  | NO       | Array     | null                                         | Content to display next to the side nav.                                         |
+| `menuItems` | NO       | Array     | null                                         | Array of menu item objects to render within the side nav. See `menuItems` below. |
+| `menuLabel` | NO       | String    | `Jump to`                                    | Placeholder value for menu label                                                 |
+| `theme`     | NO       | String    | `` | Defines the color theme for the pattern |
```
annawen1: Can you add "See `Theme` below." in the description?

RaphaelAmadeu: @annawen1, I've updated the readme with this, thanks!

PR #804, opened 2019-12-10
Fix/locale display

kennylam:
### Description

Show localised country/language of user in region selection, if available.
image

**Changed**

- Region selection country/language is now dynamic based on `<html lang />`
- `LocaleModal` css overrides moved to `CardLink`

**Removed**

- `CardLink` featureFlag removed


kennylam:
> @kennylam looks good! Need to remove `DDS_CARD_LINK` from the following files as well:
> 
> * FeatureFlags.js
> * .env.example
> * CardLink.test.js
> * CardLink README.md

@jeffchew good catch thanks

Issue #409, opened 2019-10-13
Duplicate CSS in Storybook


jeffchew:
_james-dow created the following on Sep 10:_  

 ### Detailed description

I've noticed a lot of duplicate CSS being built out in Storybook. Not sure if this impacts our releases at the moment, but at the very least we should look into how we can clean this up on our end.

### Steps to reproduce the issue

1. Open it up in chrome dev tools (or similar tool) in Storybook
2. It should be relatively obvious from there.

<img src="https://media.github.ibm.com/user/19/files/3016b500-d336-11e9-9082-6303627159ba" width="350px" />

#### Additional Information
 -  Try the global CSS approach first to see whether it works.  If not, additional exploration will be needed.   -  The work is time boxed at 3 points.

#### Acceptance Criteria
 - [ ] Each component renders 1 CSS
 - [ ] If the first approach does not work, then the steps taken are documented.

_Original issue: https://github.ibm.com/webstandards/digital-design/issues/1636_

PR #818, opened 2019-12-11
fix(css): move storybook css out of story file into local index.scss

kennylam:
### Description

Removes individual imports of css in Storybook in favor of a single `index.scss`. This prevents styles from being overwritten. This does not address #409 

### Changelog

**Changed**

- Storybook story css imports consolidated into a single `index.scss`


packages/react/src/components/ButtonGroup/__stories__/ButtonGroup.stories.js
```diff
@@ -3,10 +3,11 @@ import { storiesOf } from '@storybook/react';
 import { withKnobs, text, select } from '@storybook/addon-knobs';
 import { ArrowRight20, ArrowDown20, Pdf20 } from '@carbon/icons-react';
 import '@carbon/grid/scss/grid.scss';
```
jeffchew: @kennylam should this be added to the `index.scss` file too?

kennylam: Doh, yep I missed that one. Thanks @jeffchew

Issue #562, opened 2019-10-23
[LeadSpace] Update to use the ButtonGroup


jeffchew:
This work is to capture the updates to the `LeadSpace` patter to use the `ButtonGroup` instead of the current implementation within the `LeadSpace`.

> 3. Do we need to replace the current button group within the leadspace pattern, and remove that code?
> 
> Yes, but I think we can do as a separate PR. cc: @annawen1
> 
> _Originally posted by @jeffchew in https://github.com/carbon-design-system/ibm-dotcom-library/pull/545#issuecomment-544663344_

### Acceptance Criteria
- [ ] Remove current button group and replace with `ButtonGroup` component
- [ ] Add button limit restriction within the `LeadSpace` pattern (e.g. only pass the first two button objects in the array into the `ButtonGroup`)
- [ ] ButtonGroup feature flag added to LeadSpace README

PR #841, opened 2019-12-12
fix(pattern): buttongroup applied to the leadspace pattern

RaphaelAmadeu:
### Related Ticket(s)

#562 

### Description

The leadspace pattern and it's story has been updated to use the global component ButtonGroup instead of the custom one.

### Changelog

**Changed**

- README.md file updated
- Story updated to use the global ButtonGroup component
- Pattern components updated to use the global ButtonGroup component

**Removed**

- Previous button group component


packages/patterns-react/src/patterns/LeadSpace/__stories__/LeadSpace.stories.js
```diff
@@ -33,16 +34,53 @@ if (DDS_LEADSPACE) {
         productive: 'productive',
       };
 
+      /**
+       * Return icon component based on the parameter
+       *
+       * @param {string} icon icon name parameter
+       * @returns {*} Icon JSX component or null
+       */
+      const selectIcon = icon => {
```
annawen1: can you follow what we have for LeadSpaceCentered.stories.js to keep consistency?
```
const iconMap = {
        ArrowRight20,
        ArrowDown20,
        Pdf20,
      };

      const icons = {
        ArrowRight: 'ArrowRight20',
        ArrowDown: 'ArrowDown20',
        Pdf: 'Pdf20',
        none: null,
      };

      const buttons = [
        {
          link: '',
          copy: 'Primary action button',
          renderIcon:
            iconMap[select('primary button icon', icons, icons.ArrowRight)],
        },
        {
          link: '',
          copy: 'Secondary',
          renderIcon:
            iconMap[select('secondary button icon', icons, icons.ArrowRight)],
        },
      ];
```

packages/patterns-react/src/patterns/LeadSpace/__stories__/LeadSpace.stories.js
```diff
@@ -33,16 +34,53 @@ if (DDS_LEADSPACE) {
...
+      );
+
       const buttons = [
         {
-          link: '',
-          copy: 'Primary action button',
-          renderIcon: 'ArrowDown',
+          link: text('Button2 link:', ''),
+          copy: text('Button2 label:', 'Primary action'),
+          renderIcon: selectIcon(button2Icon),
```
annawen1: Hmmm I don't think we can have these fields and the ` buttons={object('buttons', buttons)}` at the same time - notice if you modify in your text field it doesn't actually change

packages/patterns-react/src/patterns/LeadSpace/__stories__/LeadSpace.stories.js
RaphaelAmadeu: Sure, I'll apply this logic, thanks for sharing it

packages/patterns-react/src/patterns/LeadSpace/__stories__/LeadSpace.stories.js
RaphaelAmadeu: I've changed these in the latest commit, now the only error we are getting is from the buttongroup component, one as the page loads in, and one as we change the icon to none, this is also present in the buttongroup story, so I don't know exactly how to proceed

annawen1: what errors are you getting? I'm not seeing any

Also noticed you removed the button object knob altogether - if that's the case can we put in fields so that users can change the button text?

annawen1: what errors are you getting? I'm not seeing any

Also noticed you removed the button object knob altogether - if that's the case can we put in fields so that users can change the button text?

annawen1: what errors are you getting? I'm not seeing any

Also noticed you removed the button object knob altogether - if that's the case can we put in fields so that users can change the button text?

packages/styles/scss/patterns/leadspace/_leadspace.scss
```diff
@@ -153,7 +133,8 @@ $btn-min-width: 26;
         @include carbon--make-col(8, 16);
       }
 
-      .#{$prefix}--leadspace__desc {
+      .#{$prefix}--leadspace__desc,
+      .#{$prefix}--leadspace__ctas {
         @include carbon--make-col-ready;
         @include carbon--make-col(6, 16);
```
annawen1: I think we have to increase the column span for the `ctas` since they are breaking into the second line 
maybe to `carbon--make-col(8,16)`? Can we also try to make the button stack instead of being squished?

https://ibmdotcom-patterns-react-experimental.netlify.com/?path=/story/leadspace--default-with-no-image

packages/styles/scss/patterns/leadspace/_leadspace.scss
```diff
@@ -6,9 +6,7 @@
 //
 
 @import '../../globals/imports';
-@import 'carbon-components/scss/components/button/button';
-
-$btn-min-width: 26;
```
annawen1: We need this variable, each button has a `min-width: carbon--mini-units($btn-min-width);`

Issue #782, opened 2019-12-06
Masthead: remove platform name from L0 on mobile and move to side menu


larahanlon2:
In the current masthead, the `platform name` cannot be supported in the L0. When the platform name is longer than IBM Cloud, for example, it breaks:

![Screen Shot 2019-12-05 at 6.00.54 PM.png](https://images.zenhubusercontent.com/5d96606084daa600013e3afb/ce292143-f804-47de-ab33-ce02f7b406ab)

![Screen Shot 2019-12-05 at 6.01.40 PM.png](https://images.zenhubusercontent.com/5d96606084daa600013e3afb/fc47996c-7cf4-4810-95a4-1c173dab5ba7)

![Screen Shot 2019-12-05 at 6.02.30 PM.png](https://images.zenhubusercontent.com/5d96606084daa600013e3afb/02f27f1e-e992-4c29-9bfe-a8993d381952)

I spoke with @kennylam and we agree that moving the `platform name` to the top of the menu list on mobile is the best direction right now. For future release, I opened #781 feature request to align the menu style and behaviour to the original masthead specs: https://github.com/carbon-design-system/ibm-dotcom-library/issues/781

larahanlon2:
FYI @kennylam

larahanlon2:
The tactical change would look like this.
- Move the `platform name` down to be the first menu item.
- Add a 1px `gray-70` border-bottom

<img width="461" alt="Screen Shot 2019-12-11 at 9 47 23 AM" src="https://user-images.githubusercontent.com/53093550/70632625-33536280-1bfd-11ea-9e83-be7c01075eb4.png">

Issue #781, opened 2019-12-06
Masthead: mobile menu should be full-width and cover L0


larahanlon2:
Because we're using Carbon UI header, the [hamburger] menu doesn't behave/appear the same as ibm.com masthead specs. 

Current implementation:
<img width="406" alt="Screen Shot 2019-12-06 at 11 30 47 AM" src="https://user-images.githubusercontent.com/53093550/70338963-e138c880-181b-11ea-8fde-fda866c5d8fa.png">

In the original design,`mobile` menu should slide over the entire screen, covering the L0. The `platform name` would appear at the top of the menu list. Like so:

<img width="805" alt="Screen Shot 2019-12-06 at 11 16 59 AM" src="https://user-images.githubusercontent.com/53093550/70337924-f3b20280-1819-11ea-8d2d-4b125d21571f.png">

The `platform name` is used as a title and navigational link.
The `close-icon` is used to bring the user back to the previous page view and show L0 again.
Like so:

<img width="1225" alt="Screen Shot 2019-12-06 at 11 20 27 AM" src="https://user-images.githubusercontent.com/53093550/70338234-89e62880-181a-11ea-866b-cb3aa9f60404.png">

**This model allows the user to focus more on menu navigation. It prevents them from tapping on IBM logo, Search or Login which is distracting them from the main task at hand.**

RobertaJHahn:
@kennylam @larahanlon2 I am moving this epic back into the Prioritized Epics pipeline.  During the Triage meeting, we decided to turn this original issue into an epic and use linked issues to manage the multiple tasks associated with this work.  We also agreed that the first issue would be to work with Carbon regarding the strategic solution.  I have created [Issue 824](https://app.zenhub.com/workspaces/ibmcom-library-5d449f3642eb1962336cbe52/issues/carbon-design-system/ibm-dotcom-library/824) to capture this work.  I assigned it to the first sprint in 2020 as many people will begin taking holiday vacation.

PR #842, opened 2019-12-12
Fix/masthead layout fixes

kennylam:
### Related Ticket(s)

#781 #782 

### Description
Masthead platform name, if it exists, is now hidden in mobile views and will appear in the sidenav instead. 

### Changelog

**Changed**

- Masthead platform name hidden in mobile view and appears in sidenav instead
- In mobile breakpoints (`md`) the sidenav is now 100% wide and covers search/profile buttons
- LocaleModal region buttons now scroll if viewport is too short

### DIFF
--- a/packages/styles/scss/components/masthead/_masthead.scss
+++ b/packages/styles/scss/components/masthead/_masthead.scss
@@ -55,6 +55,7 @@ $search-transition-timing: 95ms;
     margin-right: auto;
     margin-bottom: 1px;
     position: relative;
+    z-index: 5999;
 
     &__logo {
       height: 100%;
@@ -203,7 +204,9 @@ $search-transition-timing: 95ms;
       border: 2px solid transparent;
       > {
         svg {
-          fill: $text-01;
+          position: relative;
+          fill: currentColor;
+          top: 2px;
         }
       }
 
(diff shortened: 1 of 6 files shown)


packages/styles/scss/components/masthead/_masthead.scss
```diff
@@ -245,6 +248,12 @@ $search-transition-timing: 95ms;
       &--active {
         border: 2px solid $interactive-01;
         background-color: $ui-01;
+
+        @include carbon--breakpoint-up(sm) {
+          position: absolute;
+          top: 0;
+          z-index: 6001;
```
jeffchew: @kennylam 6001?

kennylam: @jeffchew Yes, styles have changed a bit to allow the overlay/nav to cover the L1 in that breakpoint. Default carbon z-index is 6000 and the close button gets covered. So it's now 6001

larahanlon2:
Looking great. Is it possible to keep the L0 border-bottom when the menu opens? It currently disappears so there is nothing visually separating:

Screen Shot 2019-12-18 at 12 30 43 PM

PR #848, opened 2019-12-13
fix(locale): fix locale filter availability text; fixes #847

### LINKED ISSUE #847: [LocaleSelector] Availability text doesn't change for input values that return no results
## Detailed description

If user types a value in the country/lang search input field that produces no results (ie. "korea" in the Americas region), the availability text does not change until user adds a space after value. 

> see changed?

Expected result when there is no matching country or lang, text should change to "This page is unavailable in your preferred location or language" immediately. However text stays the same until user types in additional character.

Chrome

v1.2.0-rc.1

1. open locale modal from footer
2. click on "Americas" Region link
3. Type "Korea" in the search input field - notice there are no results but the text underneath the input says "This page is available in the following locations and languages"
4. Type additional space - notice the text then changes to "This page is unavailable in your preferred location or language"

## Additional information
search input field

kennylam:
### Related Ticket(s)

#847 

### Description

Fixes an issue where filter was including countries from all regions

### Changelog

**Changed**

- fix filter list to only include countries of selected region

### DIFF
--- a/packages/react/src/components/LocaleModal/LocaleModalRegions.js
+++ b/packages/react/src/components/LocaleModal/LocaleModalRegions.js
@@ -10,9 +10,10 @@ import PropTypes from 'prop-types';
 import { settings as ddsSettings } from '@carbon/ibmdotcom-utilities';
 import { settings } from 'carbon-components';
 import { CardLink } from '../CardLink';
+import { ArrowRight20, Error20 } from '@carbon/icons-react';
+
 const { stablePrefix } = ddsSettings;
 const { prefix } = settings;
-import { ArrowRight20, Error20 } from '@carbon/icons-react';
 
 /**
  * LocaleModalRegions component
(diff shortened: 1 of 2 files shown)


packages/react/src/components/LocaleModal/LocaleModalRegions.js
```diff
@@ -10,9 +10,10 @@ import PropTypes from 'prop-types';
 import { settings as ddsSettings } from '@carbon/ibmdotcom-utilities';
 import { settings } from 'carbon-components';
 import { CardLink } from '../CardLink';
+import { ArrowRight20, Error20 } from '@carbon/icons-react';
```
jeffchew: @kennylam have we added `@carbon/icons-react` as a dependency?

jeffchew: nevermind, validated with you that it is.

Issue #816, opened 2019-12-11
[Utility] Same Height


photodow:
#### User Story
> As a `[user role below]`:
Carbon for IBM.com developer

> I need to:
apply the same height across a number of elements based on the largest element

> so that I can:
deliver my work without reinventing the wheel

#### Additional information

We have a need that keeps arising where a list of components need to be the same size as the largest in the list. This work should create a utility to manage that.

###### CardArray example
![image](https://user-images.githubusercontent.com/3793636/70641224-e5d9f400-1c02-11ea-8ef3-1bb6934e7048.png)

> Reach out if you need any additional information.

#### Acceptance criteria
 - [ ] Target a specific list of elements to compare
 - [ ] Identify the largest element
 - [ ] Resize all smaller elements to match the largest in the list
 - [ ] Handle events like `resize` and `onload`?
 - [ ] Development approval
 - [ ] `@carbon/ibmdotcom-utility` that can be used in react and vanilla.

jeffchew:
@photodow I'm going to pull this into sprint 24.

PR #853, opened 2019-12-16
feat(utility): first version of the sameheight utility

RaphaelAmadeu:
### Related Ticket(s)

Related to ticket #816 

### Description

This is the first version of the sameheight utility,
in order to maintain it compatible with both, React and Vanilla I've built it in a function.

### Changelog

**New**

- sameheight utility added


packages/utilities/src/utilities/sameheight/sameheight.js
```diff
@@ -0,0 +1,58 @@
...
+ * sameheight();
+ */
+function sameheight() {
+  /**
+   * Internal function made for avoiding adding more eventlisteners
+   */
+  function setHeight() {
+    if (window.innerWidth > 671) {
+      const sameheightContainers = Array.prototype.slice.call(
+        document.querySelectorAll('[sameheight]')
```
photodow: It also looks like this is going to need to be executed at a global level. I wonder if it would make more sense to allow a class name to be passed into the utility so this could be used at the component level. That way we could better leverage react hooks and not have to worry as much about racing issues.

packages/utilities/src/utilities/sameheight/sameheight.js
```diff
@@ -0,0 +1,58 @@
...
+      sameheightContainers.forEach(container => {
+        const selectedElements = Array.prototype.slice.call(
+          container.querySelectorAll(container.attributes['sameheight'].value)
+        );
+        let targetHeight = 0;
+        selectedElements.forEach(elem => {
+          elem.offsetHeight > targetHeight
+            ? (targetHeight = elem.offsetHeight)
+            : false;
+        });
```
photodow: Would there be any value in creating a separate utility, or exposing this as a method of this utility to grab the tallest item selected. `height.getMax`, `height.setHeight`. That way we could potentially set the height via State in ReactJS if possible? Idk... just thinking out loud. There may be a better way to manage that.

packages/utilities/src/utilities/sameheight/sameheight.js
```diff
@@ -0,0 +1,58 @@
...
+
+        selectedElements.forEach(elem => {
+          elem.offsetHeight == targetHeight
+            ? (elem.style.height = 'auto')
+            : (elem.style.height = targetHeight + 'px');
+        });
+      });
+    }
+  }
+  setHeight();
```
photodow: If we move forward with this being a global utility only executed once, then we should probably add some sort of onload event for this. I would still try to steer aware from doing this though. In fact, it might be good to let the component/pattern handle the events too so it can manage the cleanup properly.

packages/utilities/src/utilities/sameheight/sameheight.js
jeffchew: I actually made the suggestion in the channel to pass in the array of elements instead, that way a lot of this classname/queryselectorall logic is taken out of the utility.

packages/utilities/src/utilities/sameheight/sameheight.js
```diff
@@ -0,0 +1,58 @@
...
+ * import {sameheight} from '@carbon/ibmdotcom-utilities';
+ *
+ * sameheight();
+ */
+function sameheight() {
+  /**
+   * Internal function made for avoiding adding more eventlisteners
+   */
+  function setHeight() {
+    if (window.innerWidth > 671) {
```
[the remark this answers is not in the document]
jeffchew: we could have an optional field that sets when this logic will start to kick off (s, m, lg, etc), and default logic is that it will always be on regardless of breakpoint.

packages/utilities/src/utilities/sameheight/sameheight.js
RaphaelAmadeu: The logic was changed, as @jeffchew said in some comments up, now it's a normal function.

packages/utilities/src/utilities/sameheight/sameheight.js
RaphaelAmadeu: I've done this in the last commit

packages/utilities/src/utilities/sameheight/sameheight.js
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @param {Array} elemCollection Html objects array
+ */
+function sameheight(elemCollection) {
+  /**
+   * Internal function made for avoiding adding more eventlisteners
+   *
+   * @param {Array} elemArr array of elements to be alligned
+   */
+  function setHeight(elemArr) {
```
jeffchew: I typically don't like to see nested functions. Can you create this as perhaps a private function outside of this?

packages/utilities/src/utilities/sameheight/sameheight.js
photodow: I still think we should still maybe avoid event listeners in this utility. It seems like an easy way to get a memory leak with no way to clean any old events up.

jeffchew: Agreed, but let's add some documentation on how an adopter would use it

packages/utilities/src/utilities/sameheight/sameheight.js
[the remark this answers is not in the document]
RaphaelAmadeu: I guess this approach makes sense, so we can have the logic as a second parameter, I'll apply this to the current logic

Issue #823, opened 2019-12-11
Expressive theme – Icon-only Buttons


oliviaflory:
# Icon-only buttons

### Default
- Update icon to 20x20px
- Icon looks like it has a 16x20px area, should it be 20x20?
<img width="227" alt="Screen Shot 2019-12-11 at 1 52 21 PM" src="https://user-images.githubusercontent.com/45692622/70650697-92c16a80-1c1d-11ea-8e3b-00513d818628.png">
- should have equal padding on the right, left, top and bottom, design intent is a square 48x48px button. Current size is 47x48px (also that way in Carbon, should we contribute that change back?)

### Field
- Update icon to 20x20px
- Icon looks like it has a 16x18.4px area, should it be 20x20?
<img width="227" alt="Screen Shot 2019-12-11 at 1 55 44 PM" src="https://user-images.githubusercontent.com/45692622/70650952-0499b400-1c1e-11ea-9e3a-5ad3276dc9b4.png">

- should have equal padding on the right, left, top and bottom, design intent is a square 40x40px button. Current size is 47x40px (also that way in Carbon, should we contribute that change back?)

### Small
- Update icon to 20x20px
- Icon looks like it has a 16x20px area, should it be 20x20?
<img width="226" alt="Screen Shot 2019-12-11 at 2 00 08 PM" src="https://user-images.githubusercontent.com/45692622/70651318-90abdb80-1c1e-11ea-9e6b-28dab1acd107.png">
<img width="102" alt="Screen Shot 2019-12-11 at 1 58 39 PM" src="https://user-images.githubusercontent.com/45692622/70651329-9a354380-1c1e-11ea-943d-845fb0850a5b.png">

- should have equal padding on the right, left, top and bottom, design intent is a square 40x40px button. Current size is 47x40px (also that way in Carbon, should we contribute that change back?)

Issue #821, opened 2019-12-11
Expressive theme – Button


oliviaflory:
# Button

## Default 
**Primary, Secondary, Danger, Tertiary**

- Button height is currently 50px, needs to be 48px.

- Padding right is 60px + 3px border = 63px, Carbon also does this on their side, but design intent was 64px padding right.

- Padding left is 12px + 3px border = 15px, Carbon also does this on their side, but design intent was 16px padding left.

- Icon is 16x16px, need to be 20x20 px

**Ghost**
- Padding right and left is 12px + 3px border = 15px, Carbon also does this on their side, but design intent was 16px padding right and left.

- Icon is 16x16px, need to be 20x20 px

## Field 
**Primary, Secondary, Danger, Tertiary**
- In the first example the text "button" and icon need to be vertically centered
<img width="253" alt="Screen Shot 2019-12-11 at 1 07 44 PM" src="https://user-images.githubusercontent.com/45692622/70647587-45da9580-1c17-11ea-8e6d-bd88c5557cd2.png">

- It looks like text line height is currently 18.4, needs to be 22px. Should be to spec with $body-short-01, please confirm
<img width="231" alt="Screen Shot 2019-12-11 at 12 56 16 PM" src="https://user-images.githubusercontent.com/45692622/70646837-bb456680-1c15-11ea-8a9a-277826046fa3.png">

- Button height should maintain 40px if line height is updated.

- Icon is 16x16px, need to be 20x20 px

- Padding left is 60px + 3px border = 63px, Carbon also does this on their side, but design intent was 64px padding left.

**Ghost**
- Padding right is 16px + 3px border = 19px, Carbon also does this on their side, but design intent was 16px padding right.

## Small
**Primary, Secondary, Danger, Tertiary**
- It looks like text line height is currently 28, needs to be 22px. Should be to spec with $body-short-01, please confirm
<img width="226" alt="Screen Shot 2019-12-11 at 1 00 12 PM" src="https://user-images.githubusercontent.com/45692622/70647103-3b6bcc00-1c16-11ea-8a66-dcac962ea93a.png">
- Button height should maintain 40px if line height is updated. 
- Icon is 16x16px, need to be 20x20 px
- Padding left is 60px + 3px border = 63px, Carbon also does this on their side, but design intent was 64px padding left.

**Ghost**
- Padding right is 16px + 3px border = 19px, Carbon also does this on their side, but design intent was 16px padding right.

## Sets of buttons
[...]

PR #866, opened 2019-12-16
feat(expressive): update button component to expressive theme

annawen1:
### Related Ticket(s)

Expressive theme – Icon-only Buttons #823
Expressive theme – Button #821

### Description

Edit the button padding to fit the dimensions expected for expressive theme.

Details in the related tickets.

### Changelog

**Changed**

- changed the storybook to import 20x20 icons instead of 16x16
- adjusted the padding for the buttons


packages/styles/scss/themes/expressive/components/_button.scss
```diff
@@ -15,7 +15,43 @@
...
+    }
+  }
+
+  .#{$prefix}--btn--field,
   .#{$prefix}--btn--sm {
-    min-height: rem(40px);
+    padding: 6px 61px 6px 13px;
+
+    &.#{$prefix}--btn--ghost {
+      padding: 6px 13px;
```
jeffchew: @annawen1 @oliviaflory this is some pretty specific padding, will we run into issues when we try to contribute this back to carbon?

annawen1: hmm I could convert it to rems? 

Although I do see specific padding on their end too:
```
/// @type Value
/// @access public
/// @group button
$button-padding-field: calc(0.675rem - 3px) 60px calc(0.675rem - 3px) 12px !default;
```

jeffchew: Looks like they set a default $button-padding-field variable that's used on their end. Would we be able to set variables on our end (to match whatever variables they are using) and set it on our end?

annawen1: yeah I could do that

annawen1: yeah i could do that

Issue #743, opened 2019-11-27
Markdown utility


photodow:
#### User Story
> As a `[user role below]`:
IBM.com developer

> I need to:
repeat common markdown conversions

> so that I can:
meet the pattern requirements for rich text

#### Additional information

We might have more needs arise from this as we learn more about our rich text needs, but for now the most common thing we're seeing is the need for italics within some of our patterns following markdown syntax. In other places we need a little more rich text including links and such.

 - [Pictogram Array Functional Specs](https://ibm.box.com/s/xngulixjv6oem8sdgreoslw2tu9d6udr)
 - [Card Array Functional Specs](https://ibm.box.com/s/px2557guu928pwu3umr71t255u409teu)
 - [Markdown guidelines](https://www.markdownguide.org/basic-syntax/)
 - [Markdown processors](https://github.com/markdown/markdown.github.com/wiki/Implementations)

#### Acceptance criteria
 - [ ] Convert Markdown to HTML
 - [ ] Turn on/off specific rules (e.g. italics turned on while everything else is turned off)
 - [ ] `@carbon/ibmdotcom-utility` that can be used in react and vanilla.

photodow:
@jeffchew do we need functional specs for utilities like this?

jeffchew:
No, I think we should just describe in the acceptance criteria what we need here and then engineering can carry out.

jeffchew:
@photodow is this ready for dev? I want to pull this into sprint 24 for the long break.

photodow:
@jeffchew yeah, this should be ready to pick up.

PR #872, opened 2019-12-17
feat(utility): markdown to html converter

emerson-pereira:
### Related Ticket(s)

Addresses #743 

### Description

Implement a technology-agnostic markdown-to-html converter in the utilities package.

### Changelog

**New**

- Markdown converter in the utilities package
- Accepts parameters to choose specific markdown designators to be converted
- Currently accepts _italic_ and **bold**
- Can be easily scaled to accept more markdown features
- Removes html tags from string to avoid improper usage
- Fixes double spaces


packages/utilities/src/utilities/markdownconverter/markdownconverter.js
```diff
@@ -0,0 +1,53 @@
...
+ * @returns {string} String with html tags stripped out
+ */
+const _removeHtmlTags = str => str.replace(_htmlTagRegex, '');
+
+/**
+ *
+ * @param {string} str String to be checked for double spaces
+ * @returns {string} String with html double spaces fixed
+ */
+const _fixDoubleSpaces = str => str.replace(_doubleSpaceRegex, ' ');
```
jeffchew: Add descriptions for these methods

packages/utilities/src/utilities/index.js
```diff
@@ -11,4 +11,5 @@ export * from './featureflag';
 export * from './geolocation';
 export * from './ipcinfoCookie';
 export * from './serialize';
+export * from './markdownconverter';
```
jeffchew: can this be alphabetized in the list?

packages/utilities/src/utilities/markdownconverter/markdownconverter.js
```diff
@@ -0,0 +1,53 @@
...
+ * @param {boolean} options.italic Defines if should convert italic
+ * @param {boolean} options.bold Defines if should convert bold
+ * @returns {string} String converted to html
+ * @example
+ * import { markdownConverter } from '@carbon/ibmdotcom-utilities';
+ *
+ * markdownConverter('Lorem _ipsum_ __dolor__ *sit* **amet**.')
+ * // Lorem <em>ipsum</em> <strong>dolor</strong> <em>sit</em> <strong>amet</strong>.'
+ */
+function markdownConverter(str, { italic, bold } = {}) {
```
jeffchew: Thinking about the name, perhaps be more explicit and call it `markdownToHtml`?

packages/utilities/src/utilities/markdownconverter/markdownconverter.js
```diff
@@ -0,0 +1,53 @@
...
+ */
+const _fixDoubleSpaces = str => str.replace(_doubleSpaceRegex, ' ');
+
+/**
+ * Utiltity function for converting markdown into html
+ *
+ * @param {string} str String to convert to html
+ * @param {object} options Object with options for the conversion
+ * @param {boolean} options.italic Defines if should convert italic
+ * @param {boolean} options.bold Defines if should convert bold
```
jeffchew: Are these default as true or false?

packages/utilities/src/utilities/markdownconverter/markdownconverter.js
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ *
+ * @param {string} str String to be checked for double spaces
+ * @returns {string} String with html double spaces fixed
+ */
+const _fixDoubleSpaces = str => str.replace(_doubleSpaceRegex, ' ');
+
+/**
+ * Utiltity function for converting markdown into html
```
jeffchew: We should probably add more notes on what exactly this tool converts. It doesn't do the full markdown conversion, only particular formatting like italic, bold, etc. Also should mention that any injected HTML is stripped out. Perhaps there should be a flag to allow html to not be stripped?

packages/utilities/src/utilities/markdownconverter/markdownconverter.js
photodow: Either that or simply `markdown`. I feel like the HTML part is given, but I'm fine either way.

packages/utilities/src/utilities/markdownconverter/markdownconverter.js
```diff
@@ -0,0 +1,53 @@
...
+function markdownConverter(str, { italic, bold } = {}) {
+  const isAllStyles = !italic && !bold;
+  let converted = _removeHtmlTags(str);
+
+  if (italic || isAllStyles) {
+    converted = converted.replace(_italicRegex, (match, p1) => {
+      if (!p1.length) {
+        return match;
+      }
+      return `<em>${p1}</em>`;
```
photodow: Does it make sense to use BEM on these to avoid styling `HTML` directly? For example... `bx--md--italic`, `bx--md--bold` or something?

jeffchew: Why not make it an option? something like `useCarbonClasses`, we can set the default to be true or false?

packages/utilities/src/utilities/markdownconverter/markdownconverter.js
emerson-pereira: These are default as false, but then if none is declared it means all "styles" are applied. May be a little controversial in the way i coded it out but it seemed the simpler way to do it. I could default it to true and and change the logic though... Do you think that would be better?

packages/utilities/src/utilities/markdownconverter/markdownconverter.js
emerson-pereira: Thanks, I added some more info to the functions description!

Regarding the html stripping flag, do we want to allow that? I thought we wanted to restricted it to just the html tags coming from the markdown converter. Since we will be opening the gates to html usage inside some components text-only areas, allowing for html could result in improper usage of the component I would guess...
But that's just what I assumed, If that's not an issue I could add the flag or remove the html stripping altogether. Let me know your thoughts...

packages/utilities/src/utilities/markdownconverter/markdownconverter.js
emerson-pereira: Added!

packages/utilities/src/utilities/index.js
emerson-pereira: Done!

packages/utilities/src/utilities/markdownconverter/markdownconverter.js
emerson-pereira: Updated to `markdownToHtml`

packages/utilities/src/utilities/markdownconverter/markdownconverter.js
emerson-pereira: Added `useCarbonClasses ` option and updated to use `${prefix}--type-light` and `${prefix}--type-semibold` per carbon type documentation

packages/utilities/src/utilities/markdownToHtml/markdownToHtml.js
```diff
@@ -0,0 +1,71 @@
...
+      }
+      return useCarbonClasses
+        ? `<span class="${prefix}--type-light">${p1}</span>`
+        : `<em>${p1}</em>`;
+    });
+  }
+  if (bold || isAllStyles) {
+    converted = converted.replace(_boldRegex, (match, p1) => {
+      return useCarbonClasses
+        ? `<span class="${prefix}--type-semibold">${p1}</span>`
```
photodow: I would still encourage you to use the syntactically correct `HTML` markup here. You could just do something like the following...

```jsx
return <strong className={useCarbonClasses ? '${prefix}--type-semibold'  : ''}></strong>;

Comment (jeffchew) [reply packages/utilities/src/utilities/markdownToHtml/markdownToHtml.js]: Agreed, this.

Comment (emerson-pereira) [reply packages/utilities/src/utilities/markdownToHtml/markdownToHtml.js]: Update to semantic html

File: packages/utilities/src/utilities/markdownToHtml/markdownToHtml.js [hunk 8]
```diff
@@ -0,0 +1,71 @@
...
+const _doubleSpaceRegex = /\s{2,}/g;
+const _italicRegex = /[_*](.*?)[_*]/g;
+const _boldRegex = /[_*]{2}(.*?)[_*]{2}/g;
+
+/**
+ * Removes any html tags from a string and keeps inner text if any
+ *
+ * @param {string} str String to be checked for html tags
+ * @returns {string} String with html tags stripped out
+ */
```
jeffchew: Add `@private` so this doesn't show up under jsdoc output.

packages/utilities/src/utilities/markdownToHtml/markdownToHtml.js
```diff
@@ -0,0 +1,71 @@
...
+const _removeHtmlTags = str => str.replace(_htmlTagRegex, '');
+
+/**
+ * Checks for double spaces between words in a string and replaces for single spaces
+ *
+ * @param {string} str String to be checked for double spaces
+ * @returns {string} String with html double spaces fixed
+ */
```
jeffchew: Add @private so this doesn't show up under jsdoc output.

packages/utilities/src/utilities/markdownToHtml/markdownToHtml.js
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Converts some markdown syntaxes into html
+ * It's not a full markdown-to-html converter
+ * It currently supports two syntaxes: Bold and Italic
+ * Bold: Double asterisk (**) or double underscore (__).
+ * Bold examples: **Lorem ipsum** __dolor__
+ * Italic: Single asterisk (*) or single underscore (_)
+ * Italic examples: _Lorem ipsum_ *dolor*
+ *
```
jeffchew: @emerson-pereira just checked out the JSDoc output, the formatting seems a bit hard to read:
https://deploy-preview-872--ibmdotcom-utilities.netlify.com/global.html#markdownToHtml

I believe you can add in HTML formatting so you can retain formatting in descriptions, might be worth trying:

```
/**
 * Converts some markdown syntaxes into html
 * <p>It's not a full markdown-to-html converter</p>
 * <p>It currently supports two syntaxes: <strong>Bold</strong> and <em>Italic</em></p>
 * <ul>
 * <li>Bold: Double asterisk (**) or double underscore (__).</li>
 * <li>Bold examples: **Lorem ipsum** __dolor__</li>
 * <li>Italic: Single asterisk (*) or single underscore (_)</li>
 * <li>Italic examples: _Lorem ipsum_ *dolor*</li>
 * </ul>
 */
```

packages/utilities/src/utilities/markdownToHtml/markdownToHtml.js
```diff
@@ -0,0 +1,71 @@
...
+ * @param {string} str String to convert to html
+ * @param {object} options Object with options for the conversion
+ * @param {boolean} options.italic Defines if should convert italic
+ * @param {boolean} options.bold Defines if should convert bold
+ * @param {boolean} options.useCarbonClasses If true uses carbon typography classes
```
jeffchew: If these are optional you should update the JSDoc to show those. Also show the default values:

```
 * @param {boolean=} [options.italic=false] Defines if should convert italic
 * @param {boolean=} [options.bold=false] Defines if should convert bold
 * @param {boolean=} [options.useCarbonClasses=true] If true uses carbon typography classes
```

See: https://jsdoc.app/tags-param.html#optional-parameters-and-default-values

packages/utilities/src/utilities/markdownToHtml/markdownToHtml.js
emerson-pereira: Oh yeah, makes sense! Added.

packages/utilities/src/utilities/markdownToHtml/markdownToHtml.js
emerson-pereira: Done

packages/utilities/src/utilities/markdownToHtml/markdownToHtml.js
emerson-pereira: Good to know! Implemented that, let's see...

packages/utilities/src/utilities/markdownToHtml/markdownToHtml.js
emerson-pereira: Right, also the options object itself is optional so it would look like the below right?

```
* @param {string} str String to convert to html
 * @param {object} [options={}] Object with options for the conversion
 * @param {boolean} [options.italic=false] Defines if should convert italic
 * @param {boolean} [options.bold=false] Defines if should convert bold
 * @param {boolean} [options.useCarbonClasses=true] If true uses carbon typography classes
```

emerson-pereira: And I had the `useCarbonClasses ` be default as `true` as you put out in the example. I think it makes!

packages/utilities/package.json
```diff
@@ -35,6 +35,7 @@
   },
   "dependencies": {
     "axios": "^0.19.0",
+    "carbon-components": "^10.9.0",
     "js-cookie": "^2.2.1",
     "window-or-global": "^1.0.1"
   },
```
emerson-pereira: Also for the prefix thing I'm using the `carbon-components` package as a dependency in the utilities package.json. Can anyone advise of this is the correct approach? This way it worked well in the test I made locally and automated tests passed fine too.

packages/utilities/src/utilities/markdownToHtml/markdownToHtml.js
jeffchew: @emerson-pereira for optional, you would need to add an `=` equal sign after the param type (e.g. `{boolean=}`)

packages/utilities/src/utilities/markdownconverter/markdownconverter.js
jeffchew: I honestly don't have a preference either way. Perhaps true by default? Otherwise passing nothing doesn't do anything, right?

packages/utilities/src/utilities/markdownconverter/markdownconverter.js
jeffchew: right, I guess maybe it strips by default, you have to explicitly set a flag to turn it off? Remember this is a utility that can feasibly be used anywhere. Maybe an adopter would want to use the utility directly in their application.

packages/utilities/package.json
jeffchew: Hmm, it makes sense to pull it in since it needs the settings for the prefix. It's just a little odd for a utility package to pull that in just for the prefix, lol! I don't think it is doing any harm though (at least I hope not)

packages/utilities/src/utilities/markdownconverter/markdownconverter.js
emerson-pereira: Right, ok I can add that

packages/utilities/src/utilities/markdownconverter/markdownconverter.js
emerson-pereira: Exactly, that's why I chose this approach, it was the one that most sense from the adopter experience pov. Any other approach would take an extra level of complexity that I'm not would be good for the adopter, be it adding extra flag to allow allow, accepting string/array of strings and so on...

packages/utilities/src/utilities/markdownconverter/markdownconverter.js
emerson-pereira: Flag added

packages/utilities/src/utilities/markdownToHtml/markdownToHtml.js
emerson-pereira: as talked on slack equal sign is not necessary in this syntax

Issue #877, opened 2019-12-17
Footer – secondary button theme is incorrect


oliviaflory:
The locale selector button is not using the correct theme. It looks like the secondary button from the white or gray 10 theme.
![Screen Shot 2019-12-17 at 11.59.39 AM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/5f02f9d5-f626-4e1b-84af-49aa3df606af)

In the Gray 90 theme the secondary button should be lighter than the background color, example below
![Screen Shot 2019-12-17 at 12.45.16 PM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/5e51baa5-e6e8-464a-b2c1-419dc1288fb3)

Issue #876, opened 2019-12-17
Footer Accordion Theme Slightly Off


photodow:
## Detailed description

It appears as if the footer accordion's theme has been reset.

![image](https://user-images.githubusercontent.com/3793636/71010954-3e513b80-20b2-11ea-840c-eb8b987a42f1.png)

PR #881, opened 2019-12-17
fix(footer): theme bugs

annawen1:
### Related Ticket(s)

Footer Accordion Theme Slightly Off #876
Footer – secondary button theme is incorrect #877

### Description

import the carbon styles for accordion and button and stick them within the theme mixins

### Changelog

**Changed**

- add carbon accordion and button under the theme mixin

Screen Shot 2019-12-17 at 12 05 30 PM

### DIFF
--- a/packages/styles/scss/components/footer/_footer-nav.scss
+++ b/packages/styles/scss/components/footer/_footer-nav.scss
@@ -6,6 +6,7 @@
 //
 
 @import '../../temp-carbon-expressive/accordion/accordion-expressive';
+@import 'carbon-components/scss/components/accordion/accordion';
 
 /// Footer nav styles
 /// @access private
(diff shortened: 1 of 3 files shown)


packages/styles/scss/components/footer/_footer-nav.scss
```diff
@@ -54,6 +55,7 @@
 
 @include exports('footer-nav') {
   @include carbon--theme($carbon--theme--g90) {
+    @include accordion;
```
jeffchew: @annawen1 I'm thinking through this, I'm wondering if we should scope these styles, so we would have these in the corresponding wrapper class to be sure, and that this doesn't affect other accordions or buttons outside of the component:

```
.${prefix}--footer {
  @include accordion;
  @include button;
}
```

annawen1: ah ok sure

Issue #765, opened 2019-12-03
Component:  Develop and alpha version of the lightbox component


RobertaJHahn:
#### User Story

> As a `[user role below]`:
DDS developer

> I need to:
work closely with the Functional Analyst and designer to begin developing the Lightbox component

> so that I can:
become involved earlier in the design process and functional documentation to compress the schedule and develop a higher quality pattern

#### Additional information

- Story within Storybook with corresponding knobs
- minimum 80% unit test coverage
- Utilize Carbon
- [Lightbox MVP design specs #326](https://app.zenhub.com/workspaces/ibmcom-library-5d449f3642eb1962336cbe52/issues/carbon-design-system/ibm-dotcom-library/326)
- [Lightbox MVP functional specs  #688](https://app.zenhub.com/workspaces/ibmcom-library-5d449f3642eb1962336cbe52/issues/carbon-design-system/ibm-dotcom-library/688)
#### Acceptance criteria
 - [ ] alpha react component within the Patterns storybook
 - [ ] create a feature flag for this component
 - [ ] Gap analysis of alpha code and design & functional specs
 - [ ] If applicable, create and size stories for additional work

Issue #767, opened 2019-12-03
Component:  Image Functional Specs - alpha version


photodow:
#### User Story
> As a `[user role below]`:
IBM.com developer for Carbon

> I need to:
build a responsive picture component

> so that I can:
be sure I'm following the standards when creating patterns and other components.

#### Additional information
As mentioned in issue #699 we need a component for responsive images, and thus need some functional specs. In a call, I had the other day with the web simplification team they mentioned certain limitations to help standardize the image sizes, aspect ratios, number of images allowed, aligned to what breakpoints, and possibly more. I'll be reaching out to the team to help define some of this.

![image](https://user-images.githubusercontent.com/3793636/70094323-e8779080-15e7-11ea-9736-d1b0edfb1bb9.png)

 - [Request for Responsive Picture Component #699](https://github.com/carbon-design-system/ibm-dotcom-library/issues/699)

###### Components/patterns already using something similar
 - [pattern/useCasesGroup](https://github.com/carbon-design-system/ibm-dotcom-library/blob/13e21b52b71a9a00057b14d73db5df0af3efcbf5/packages/patterns-react/src/patterns/UseCases/UseCasesGroup.js#L62-L72)
 - [pattern/leadspaceImage](https://github.com/carbon-design-system/ibm-dotcom-library/blob/0aecec572ee6d408b5d228588183da0d13cd39f8/packages/patterns-react/src/patterns/LeadSpace/LeadSpaceImage.js#L40-L55)

###### Patterns with imagery
- [Card with image](https://ibm.box.com/s/riny5aabsllfxf8dyn2g2b9ykpnejya7)
- Leadspace
   - [Centered leadspace](https://ibm.box.com/s/w9g5arb8pcrw2sheyonuolaukqvjgcly)
- [Use case](https://ibm.box.com/s/9khxckrhvj93y7os2gzbkuuu9e32aka4)
- [Featured link](https://ibm.box.com/s/22eumjqkfj273y9s3251zwgyz3zc7q0n)
- *[Logo grid](https://ibm.box.com/s/vha7x9514lnuqw1t2vwrw23xqij49zo4) (maybe not needed)*
- ~*[Pictogram array](https://ibm.box.com/s/iyrjxcp1lwymla8891hgsyl8bl4m0fvn) (maybe not needed)*~
- [Embeddable image](https://ibm.box.com/s/dtn9t9o1iy8h26rb66mr05gns98v42st)
[...]

PR #882, opened 2019-12-17
feat(component): implement Lightbox component

sannminwin:
### Related Ticket(s)

#765 Component: Develop alpha version of the Lightbox Media Viewer component.

LightboxMediaViewer

### Description
This pull request creates Lightbox Media Viewer component
 
Deploy Preview:

Design Reference:

### Changelog

**New**

- LightboxMediaViewer component


packages/react/src/components/Lightbox/README.md
```diff
@@ -0,0 +1,104 @@
...
+          sm: 'https://via.placeholder.com/640x320',
+          md: 'https://via.placeholder.com/768x384',
+          lg: 'https://via.placeholder.com/1024x512',
+        },
+        alt: 'Placeholder Image',
+      }
+    open={true}
+    onClose={() => alert('<<< model closed >>>')}
+  />;
+}
```
emerson-pereira: Looks like it's not returning the actual LightBox in the example

packages/react/src/components/Lightbox/Lightbox.js
```diff
@@ -0,0 +1,116 @@
...
+  );
+};
+
+Lightbox.PropTypes = {
+  title: PropTypes.string,
+  copy: PropTypes.string,
+  image: PropTypes.shape({
+    uri: PropTypes.shape,
+    alt: PropTypes.string,
+  }),
```
emerson-pereira: Is it missing the actual shaping of the uri prop here?

packages/react/src/components/Lightbox/README.md
jeffchew: Right, and are any of these props required? How about default props?

packages/react/src/components/Lightbox/Lightbox.js
jeffchew: Right, and are any of these props required? How about default props?

packages/react/src/components/Lightbox/README.md
```diff
@@ -0,0 +1,104 @@
...
+@include carbon--font-face-sans();
+```
+
+> 💡 Only import font's once per usage
+
+```javascript
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { LightBox } from '@carbon/ibmdotcom-react';
+import '@carbon/ibmdotcom-styles/scss/components/lightbox/_lightbox.scss';
```
jeffchew: This needs to be moved up to the `scss` block

packages/react/src/components/Lightbox/README.md
jeffchew: Right, seems to be a closing `return;` Can you see if you can reconcile this against any of the other READMEs? @annawen1 did a refactor of the READMEs a little while ago.

packages/react/src/components/Lightbox/README.md
```diff
@@ -0,0 +1,104 @@
...
+| `copy`    | NO       | String    | null          | Short description text for lightbox image               |
+| `image`   | YES      | Object    | null          | Image object for lightbox. see `Image` below for detail |
+| `open`    | YES      | Boolean   | null          | sets `true | false` whether the lightbox is open/close  |
+| `onClose` | NO       | Function  | null          | Run the function to do something on close               |
+
+### Image
+
+| Name  | Data Type | Description                 |
+| ----- | --------- | --------------------------- |
+| `uri` | object    | Urls for images             |
```
jeffchew: @photodow @sannminwin would this be using the responsive image component? we should align on the prop names.

packages/react/src/components/Lightbox/__stories__/Lightbox.stories.js
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { withKnobs, text, object, boolean } from '@storybook/addon-knobs';
+import '@carbon/grid/scss/grid.scss';
+import '../../../../../styles/scss/components/lightbox/_lightbox.scss';
```
jeffchew: @sannminwin create an `index.scss` file within the story folder, which should then import these.

packages/react/src/components/Lightbox/Lightbox.js
sannminwin: Yeah, For alpha phase I , it will work with minimum one url althought we can pass all three sm, md, lg and  we can make lg as required one.

packages/react/src/components/Lightbox/README.md
sannminwin: I think that's way to go. Where can I find specs for responsive image component ?

photodow: @jeffchew yes, that's the intent at the moment.

@sannminwin loosely working on this sprint. Will probably be pushing that particular work into some of the extended days of this sprint.

photodow: I think this is the issue though where you can track those specs. #767

packages/react/src/components/Lightbox/README.md
sannminwin: updated with [this commit](https://github.com/carbon-design-system/ibm-dotcom-library/pull/882/commits/afbd414592d1948f0861ff4866931e632f6389c6)

packages/react/src/components/Lightbox/Lightbox.js
sannminwin: updated with [this commit](https://github.com/carbon-design-system/ibm-dotcom-library/pull/882/commits/afbd414592d1948f0861ff4866931e632f6389c6)

packages/react/src/components/Lightbox/__stories__/Lightbox.stories.js
sannminwin: updated with [this commit](https://github.com/carbon-design-system/ibm-dotcom-library/pull/882/commits/afbd414592d1948f0861ff4866931e632f6389c6)

packages/react/src/components/Lightbox/README.md
sannminwin: updated with [this commit](https://github.com/carbon-design-system/ibm-dotcom-library/pull/882/commits/afbd414592d1948f0861ff4866931e632f6389c6)

packages/react/src/components/Lightbox/Lightbox.js
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * @param {object} props props object
+ * @param {string} props.title LightBox media title
+ * @param {string} props.copy Lightbox media short description
+ * @param {string} props.image Lightbox responsive image object
+ * @param {boolean} props.open sets whether the modal is open/close
+ * @param {Function} props.onClose do something on close in addition, return false to completely replace
+ * @returns {*} JSX Object
+ */
+const Lightbox = ({ title, copy, image, open, onClose }) => {
```
photodow: I know it's a little longer, but the name `Lightbox` itself doesn't really mean much here. Technically the `ExpressiveModal` is a "lightbox" already. I think this should be called `LightboxMediaViewer`

packages/react/src/components/Lightbox/Lightbox.js
```diff
@@ -0,0 +1,120 @@
...
+  /**
+   * Set default image
+   *
+   * @param {*} image responsive image object
+   * @returns {string} default image url
+   */
+  const defaultImage = image => {
+    return image.uri.md
+      ? image.uri.md
```
photodow: I think we've been more consistently using `src`, so maybe something like `image.src.md` would be closer. I think this will be more defined once the responsive image component begins taking shape.

sannminwin: I think this is one of the things that need refactoring and standardizing. For this alpha pattern , I am using same image object format that I used in use cases. 
@jeffchew @photodow 
I can refactor for both this component and use cases  pattern if we all want it.

photodow: It's probably not worth refactoring anything until it's a little more concrete.

packages/react/src/components/Lightbox/Lightbox.js
jeffchew: I would actually like @wonilsuhibm 's opinion on this as well. Do we envision having this component be specific to being a media viewer, or do we want to leave it open for other things in the future?

photodow: We can totally pull in @wonilsuhibm to get his input, but `LightboxMediaViewer` is what he and I aligned on. Images for now, possibly video, and audio (not sure about pdfs and such, maybe?) in the future. @jeffchew Is there something other than those you're thinking of could go in here? Anything else would most likely just use the `ExpressiveModal`.

sannminwin: it sounds like we are good to go with the name LightboxMediaViewer.
@wonilsuhibm @jeoffw @photodow 
I am cool with new name too. Should we be moving forward ?

jeffchew: @sannminwin let's go with it. And I think you tagged the wrong Jeff.

sannminwin: updated

packages/react/src/components/LightboxMediaViewer/LightboxMediaViewer.js
```diff
@@ -0,0 +1,120 @@
...
+      ? image.uri.lg
+      : image.uri.sm;
+  };
+
+  if (!image.uri.lg) {
+    return null;
+  }
+  return featureFlag(
+    DDS_LIGHTBOX_MEDIA_VIEWER,
+    <section
```
photodow: Not sure if a section is necessary. This is in a lot of ways just a wrapper for the `ExpressiveModal` with a very specific layout and data model. We should be able to just pass in the `className` without a problem. I wonder if it makes sense to override the `data-autoid` for the `ExpressiveModal` since it's technically different, either that or just keep it the same, and `LightboxMediaViewer` doesn't get a new a specific stable selector because its shared. 🤔idk... @jeffchew any thoughts?

```jsx
return (
  <ExpressiveModal
    data-autoid={`${stablePrefix}--lightbox`}
    className={`${prefix}--lightbox`}>
    <MediaViewer />
  </ExpressiveModal>
);
```

packages/react/src/components/LightboxMediaViewer/LightboxMediaViewer.js
```diff
@@ -0,0 +1,120 @@
...
+      data-autoid={`${stablePrefix}--lightbox-media-viewer`}
+      className={`${prefix}--lightbox-media-viewer`}>
+      <ExpressiveModal open={open} fullwidth={true} onClose={onClose}>
```
photodow: I wonder if it would reduce complexity in documentation, and possibly code if we just use an object spread and allow users to pass in props based on the `ExpressiveModal `dependency documentation.

```jsx
<ExpressiveModal fullwidth={true} {...modalProps}>
```

packages/react/src/components/LightboxMediaViewer/LightboxMediaViewer.js
```diff
@@ -0,0 +1,120 @@
...
+                    />
+                  )}
+                  {image.uri.sm && (
+                    <source
+                      media="(max-width: 320px)"
+                      srcSet={`${image.uri.sm}`}
+                    />
+                  )}
+                  <img src={defaultImage(image)} alt={image.alt}></img>
+                </picture>
```
photodow: While this work might still be worked on I'm pretty sure it'll just be an array of image objects. `[{ src: '', alt: '' }, { src: '', alt: '' }]`, and this component for the time being only needs to take in one image. The object and logic you have above isn't really necessary at this time, especially with the responsive image coming down the road soon.

packages/react/src/components/LightboxMediaViewer/__stories__/LightboxMediaViewer.stories.js
```diff
@@ -0,0 +1,51 @@
...
+        'Curabitur malesuada varius mi eu posuere'
+      );
+
+      const copy = text(
+        'copy (required)',
+        `
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aenean et ultricies est. 
+        Mauris iaculis eget dolor nec hendrerit. Phasellus at elit sollicitudin, sodales 
+        nulla quis, consequat libero. Here are 
+        some common categories:
```
photodow: The way you're using your template literal you're getting this extra space in the input field.

image

packages/react/src/components/LightboxMediaViewer/__stories__/LightboxMediaViewer.stories.js
```diff
@@ -0,0 +1,51 @@
...
+        },
+        alt: 'Placeholder Image',
+      };
+      return (
+        <LightboxMediaViewer
+          title={title}
+          copy={copy}
+          image={object('image', image)}
+          open={boolean('open', true)}
+          onClose={() => alert('<<< model closed >>>')}
```
photodow: Not sure if this is necessary for the story.

packages/react/src/components/LightboxMediaViewer/LightboxMediaViewer.js
jeffchew: @photodow @sannminwin I would say keep it the same.

jeffchew:
@photodow Sann is technically off the project now. I'm wondering if we should merge in as-is then have someone else address feedback?

photodow:
Yeah, that makes sense.

Issue #874, opened 2019-12-17
CardLink Focus is Off


photodow:
## Detailed description

It appears as if the CardLink Focus state is off at the moment.

Here is what I'm seeing in Chrome.

![image](https://user-images.githubusercontent.com/3793636/71009886-9e46e280-20b0-11ea-891f-5ca75232ecfa.png)

https://ibmdotcom-react.netlify.com/?path=/story/card--link

I think it should be `2px`, but something else might be overwriting it.

![image](https://user-images.githubusercontent.com/3793636/71009962-c0406500-20b0-11ea-8516-a0fde6fcb3dd.png)

https://www.carbondesignsystem.com/components/tile/code

photodow:
> My concern here is if Carbon updates their code to say, `3px` we'd be left in the dust. I opened a ticket with Carbon to get to the root of the problem.
> 
> https://github.com/carbon-design-system/carbon/issues/4937

https://github.com/carbon-design-system/ibm-dotcom-library/pull/897#discussion_r361229793

PR #897, opened 2019-12-20
fix(component): fixed focus outline style

RaphaelAmadeu:
### Related Ticket(s)

Related to ticket #874 
### Description

I've updated the component styling to fix the outline size, setting it to 2px.
Screen Shot 2019-12-20 at 14 28 30
Screen Shot 2019-12-20 at 14 28 39

### Changelog

**Changed**

- CardLink Outline size


packages/styles/scss/components/card-link/index.scss
```diff
@@ -52,7 +52,7 @@
     }
 
     &:focus {
-      outline-color: $focus;
+      outline: 2px solid $focus;
```
photodow: My concern here is if Carbon updates their code to say, `3px` we'd be left in the dust. I opened a ticket with Carbon to get to the root of the problem.

https://github.com/carbon-design-system/carbon/issues/4937

jeffchew: Part of me wants to just track in Carbon, but seems like they have a huge backlog of tickets they aren't getting to. Maybe we should just contribute it directly @RaphaelAmadeu @photodow ?

jeffchew:
Since we will be contributing back directly to Carbon, will close out this PR.

Issue #540, opened 2019-10-18
[LinkWithIcon] Alignment of focus state


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

The focus state isn't vertically centered around the link. The top margin is larger than the bottom margin, and ideally it would be equal.

Let me know if this isn't possible or if it's because of the line height. I would have thought the focus outline would be 22px high to match the line height

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

LinkWithIcon

## Steps to reproduce the issue

1. Go to https://ibmdotcom-react.netlify.com
2. View the LinkWithIcon story
3. Keyboard tab to the LinkWithIcon

## Additional information

<img width="354" alt="67017678-1bada100-f0c8-11e9-9ca6-bb57acacc7f6" src="https://user-images.githubusercontent.com/1641214/67119792-d9628d80-f1b5-11e9-9d60-1363cc5c0557.png">

Related to PR: https://github.com/carbon-design-system/ibm-dotcom-library/pull/532

RobertaJHahn:
@oliviaflory engineering would like to see examples of this button within multiple contexts to prioritize and size.  Jeff will work with you on this.

RobertaJHahn:
Will not be concerned about the inline treatment.  Will merge without the inline design.

PR #899, opened 2019-12-23
Fix/link with icon

RaphaelAmadeu:
### Related Ticket(s)

Related to ticket #540 
### Description

I've tried to align the text to the center of the element more properly by using flex, so it would be displayed properly on the focus state.

Screen Shot 2019-12-20 at 11 35 19


photodow:
@oliviaflory we need your sharp eyes on this. Please! 😄

jeffchew:
> @RaphaelAmadeu @wonilsuhibm @oliviaflory is it an issue that the link icon is now a block element? Meaning it won't sit inline to other text and inline elements?
> 
> image
> 
> vs...
> 
> image
> 
> > P.S. ignore the misalignment, I think that's different types paired up. Sorry, but hopefully you still get what I'm trying to say! 😅

Were we always allowing a link with icon to be inline? I guess that's more of a question for @wonilsuhibm @oliviaflory @larahanlon2

oliviaflory:
@photodow @jeffchew @RaphaelAmadeu 

The first issue I see is that the link text is 14px/18px, and should be 16px/22px, that might help with the misaligned line heights?


Ideally we would not have a text+icon component inline, but as James has pointed out before, there's nothing stopping our adopters 😅 

We will hopefully be able to write guidelines to steer adopters away from this implementation *within a paragraph* like this
image

However, I can see this example where multiple links could be stacked on top of each other. 

image

The focus state extending so far past the end of the icon is not great. The focus state should end right after the icon, like this:
Screen Shot 2020-01-02 at 5 00 34 PM

wonilsuhibm:
re preventing link with icon from being used inline, us there anything tbat can be done via `block` dispkay setting?

annawen1:
> @photodow @jeffchew @RaphaelAmadeu
> 
> The first issue I see is that the link text is 14px/18px, and should be 16px/22px, that might help with the misaligned line heights?
> 
> Ideally we would not have a text+icon component inline, but as James has pointed out before, there's nothing stopping our adopters 😅
> 
> We will hopefully be able to write guidelines to steer adopters away from this implementation _within a paragraph_ like this
> image
> 
> However, I can see this example where multiple links could be stacked on top of each other.
> ...

So the font-size and line-height will adjust to 16px/22px once we turn the `REACT_STORYBOOK_CARBON_EXPRESSIVE` env var flag on

Screen Shot 2020-01-03 at 2 53 51 PM

Issue #826, opened 2019-12-11
[Pictogram Array] Use new markdown utility


photodow:
## Detailed description

The [PictogramArray](https://ibmdotcom-patterns-react-experimental.netlify.com/?path=/story/content-array-with-pictograms--default) isn't converting markdown italics into markup.

###### Example of it not converting markdown
![image](https://user-images.githubusercontent.com/3793636/70664644-6d8a2780-1c30-11ea-8e7b-14c512bc8f4c.png)

###### Example of what it should look like
![image](https://user-images.githubusercontent.com/3793636/70665150-9959dd00-1c31-11ea-900e-c3a94b4f0d93.png)

## Additional information

- [PictogramArray Functional Specs](https://ibm.box.com/s/xngulixjv6oem8sdgreoslw2tu9d6udr)
- [PictogramArray](https://ibmdotcom-patterns-react-experimental.netlify.com/?path=/story/content-array-with-pictograms--default)

PR #901, opened 2019-12-26
fix(pictogram-array): use markdown utility

emerson-pereira:
### Related Ticket(s)

Addresses #826 

### Description

Adds new markdown-to-html utility to copy of the pictogram array pattern

image

### Changelog

**Changed**

- Changed copy of pictogram array pattern to use markdown converter utility


packages/patterns-react/src/patterns/PictogramArray/README.md
```diff
@@ -73,6 +73,27 @@ ReactDOM.render(<App />, document.querySelector('#app'));
...
+```javascript
+{
+  copy: '_Lorem_ ipsum __dolor__ sit amet.',
+  ...
+}
+```
+
+Renders:
+
+_Lorem_ ipsum **dolor** sit amet.
```
photodow: All we need is italic. Bold isn't required and is in fact deeply frowned upon by the web simplification team. It denotes/changes content hierarchy they said.

emerson-pereira: Oh right, I forgot. Fixing...

packages/patterns-react/src/patterns/PictogramArray/README.md
```diff
@@ -73,6 +73,24 @@ ReactDOM.render(<App />, document.querySelector('#app'));
...
+{
+  copy: '_Lorem_ ipsum *dolor* sit amet.',
+  ...
+}
+```
+
+Renders:
+
+_Lorem_ ipsum _dolor_ sit amet.
+
```
emerson-pereira: For some reason Storybook will add bolder font-weight to their <em> tag.
Apparently others have this issue: https://github.com/tuchk4/storybook-readme/issues/156
I know that there's a story to update storybook, maybe that solves this issue.
Otherwise I can follow up with that open issue in storybook github...

packages/patterns-react/src/patterns/PictogramArray/README.md
emerson-pereira: Fixed

jeffchew:
> Personally I feel like we need to open the flood gate for all markdown, and allow adopters like Drupal to handle the limitations they need. With that said I left a change to align with expectations from the web simplification team, and the functional specs.

Agreed. All options should be turned on, and for this pattern in particular you are manually setting bold to false.

emerson-pereira:
> > Personally I feel like we need to open the flood gate for all markdown, and allow adopters like Drupal to handle the limitations they need. With that said I left a change to align with expectations from the web simplification team, and the functional specs.
> 
> Agreed. All options should be turned on, and for this pattern in particular you are manually setting bold to false.

That's what I had previously, should I just roll back to have it allow all or are you suggesting something else? would it make sense to pass in a flag to the component of accepted styles?

photodow:
> Agreed. All options should be turned on, and for this pattern in particular you are manually setting bold to false.

Just for clarity on my "flood gate" comment above... I was more so referring to even all patterns allowing the use of all markdown features within these rich text copy areas. That way everything is consistent and easier to maintain, and so then adopters like drupal would control what kind of markdown they would like to pass into it by limiting or stripping before rendering. That's my personal opinion at the moment though.

packages/patterns-react/src/patterns/PictogramArray/PictogramArrayItem.js
```diff
@@ -9,12 +9,27 @@ import { ArrowRight20 } from '@carbon/icons-react';
...
+/**
+ * Converts markdown to html and returns object to be used in `dangerouslySetInnerHTML`
+ *
+ * @param {string} string string to be converted to html
+ * @returns {object} object with a __html key containing converted string
+ */
+const convertMarkdown = string => ({
+  __html: markdownToHtml(string, {
+    italic: true,
+  }),
```
jeffchew: @emerson-pereira as discussed, update the utility so that this would be `bold: false` instead.

emerson-pereira: @jeffchew By default every style is false. The only way of limiting is opting-in. At least that's how the converter is designed right now.

jeffchew: @emerson-pereira can the utility be updated to be opt-out instead?

emerson-pereira: @jeffchew Sure, I'll work on it.

emerson-pereira: @jeffchew Created PR #977

emerson-pereira: @jeffchew updated to `bold: false`

Issue #745, opened 2019-11-27
Content Array Group


photodow:
#### User Story
> As a `[user role below]`:
As an IBM.com developer

> I need to:
create a content group component

> so that I can:
reuse it across multiple layout patterns

#### Additional information
This is a common pattern I'm seeing throughout multiple of these layout patterns. I think we could cut down some complexity if we broke this out into a separate component and reuse it. Basically it can be used multiple times throughout the content array section containing one content entity type (e.g. pictogram, card, simple). It contains a title, and area below to pass in content (.e.g a list of content entities as mentioned above)

![image](https://user-images.githubusercontent.com/3793636/70340297-2823bf80-1816-11ea-8c9c-67f6a6b0fe77.png)

 - [Pictogram Array Functional Specs](https://ibm.box.com/s/xngulixjv6oem8sdgreoslw2tu9d6udr)
 - [Card Array Functional Specs](https://ibm.box.com/s/px2557guu928pwu3umr71t255u409teu)
 - Pictogram Array Visual Specs [[sketch](https://ibm.box.com/s/iyrjxcp1lwymla8891hgsyl8bl4m0fvn)][[png](https://ibm.box.com/s/ai326isog3uovcgvhuq82e26x6fwb0j0)]
 - Simple Array Visual Specs [[sketch](https://ibm.box.com/s/00upfv34of6qgeqyil13jjengaefxll9)] [[png](https://ibm.box.com/s/cvcc83kpy8ekcaoa5lz6edjiksmw0izr)]
- Card Array Visual Specs [[sketch](https://ibm.box.com/s/fx8odkidz7h63yh242p4k8wrtz77gxj1)] [[png](https://ibm.box.com/s/gy1nsj9em79yi0zdn0oi7k8as934lx9i)]
 - Investigate the [Pictogram Array spacing issue #833](https://app.zenhub.com/workspaces/ibmcom-library-5d449f3642eb1962336cbe52/issues/carbon-design-system/ibm-dotcom-library/833) when updating the Pictogram Array pattern to be sure that this update resolves the documented spacing issues.  If resolved, document this in Issue 833 and close it. 

#### Acceptance criteria
 - [x] Common title included
 - [x] Section to pass in children (e.g. JSX/HTML)
 - [ ] Design, Development, and OM approval
 - [x] Apply to existing patterns/components
    - [x] CardyArray
    - [x] PictogramArray
    - [x] ListSection

PR #902, opened 2019-12-27
Content Group component created

RaphaelAmadeu:
### Related Ticket(s)

#745 
### Description

The Content Group component was created and added to the CardArray pattern.

### Changelog

**New**

- ContentGroup component

**Changed**

- CardArray changed to use the new component

**Removed**

- Unnecessary scss that was used in CardArray


packages/react/src/components/ContentGroup/README.md
```diff
@@ -0,0 +1,67 @@
...
+> 💡 Don't forget to import the content group styles from
+> [@carbon/ibmdotcom-styles](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/styles).
+
+## Props
+
+| Name           | Required | Data Type | Default Value                   | Description                                     |
+| -------------- | -------- | --------- | ------------------------------- | ----------------------------------------------- |
+| `heading`      | YES      | Object    | {copy: null, type: 'heading-4'} | Heading prop object                             |
+| `heading.copy` | YES      | String    | null                            | Content Group section title                     |
+| `heading.type` | NO       | String    | 'heading-4'                     | Title heading type, can be changed to heading-5 |
```
photodow: Generally speaking, we need to align on our data modal a little better, and it's naming conventions. For example in this pull request alone we call it heading, and we call it title. Just look at lines 50-54 below here. In the functional specs I've been trying to call these titles across the board. Does heading make more sense to y'all?

packages/patterns-react/src/patterns/CardArray/CardArray.js
```diff
@@ -28,46 +30,38 @@ const CardArray = ({ title, content }) => {
...
-    <section
-      className={`${prefix}--cardarray`}
-      data-autoid={`${stablePrefix}--cardarray`}>
-      <div className={`${prefix}--cardarray__container`}>
-        <div className={`${prefix}--cardarray__row`}>
-          <div className={`${prefix}--cardarray__col`}>
-            <h3 className={`${prefix}--cardarray__title`}>{title}</h3>
-          </div>
+    <section data-autoid={`${stablePrefix}--cardarray`}>
+      <ContentGroup heading={{ copy: title }}>
```
photodow: @RaphaelAmadeu and I talked a little about this earlier this week. My comment here is similar to the https://github.com/carbon-design-system/ibm-dotcom-library/pull/882#discussion_r361229946 I gave to Sann earlier in the week as well. We've wrapped this `ContentGroup` with a `section` element. `ContentGroup` in a way is the containing component. That's what we're building here, and that's what `"CardArray"` is. It's a `ContentGroup`. Does the CardArray need a wrapper just so it can have a stable selector, is its stable selector technically shared with the `ContentGroup`, or is it completely a separate thing? It's almost as if we are saying from a hierarchical standpoint that `CardArray` is greater than `ContentGroup` when they are both one in the same.

@jeffchew maybe this is something we can discuss in our next huddle? I'm going to add it for now.

packages/patterns-react/src/patterns/CardArray/CardArray.js
```diff
@@ -28,46 +30,38 @@ const CardArray = ({ title, content }) => {
...
     DDS_CARD_ARRAY,
-    <section
-      className={`${prefix}--cardarray`}
-      data-autoid={`${stablePrefix}--cardarray`}>
-      <div className={`${prefix}--cardarray__container`}>
-        <div className={`${prefix}--cardarray__row`}>
-          <div className={`${prefix}--cardarray__col`}>
-            <h3 className={`${prefix}--cardarray__title`}>{title}</h3>
-          </div>
+    <section data-autoid={`${stablePrefix}--cardarray`}>
```
photodow: Let's say we are going to use a wrapper to do this, I'm still not sure `section` would be the appropriate one to use. I almost feel like `article` would be better because this is a self-contained piece of content. I still kind of feel like that belongs in the `ContentGroup` though.

packages/patterns-react/src/patterns/CardArray/CardArray.js
```diff
@@ -28,46 +30,38 @@ const CardArray = ({ title, content }) => {
   const containerRef = useRef();
 
   useEffect(() => {
-    setCardsHeight();
+    setSameHeight();
+    window.addEventListener('resize', setSameHeight);
```
photodow: How many times is this getting called in the lifecycle? I know for the footer I had pass in an empty array `[]` to `useEffect()` so it only called it on the first render.

packages/patterns-react/src/patterns/CardArray/CardArray.js
```diff
@@ -28,46 +30,38 @@ const CardArray = ({ title, content }) => {
...
-      <div className={`${prefix}--cardarray__container`}>
-        <div className={`${prefix}--cardarray__row`}>
-          <div className={`${prefix}--cardarray__col`}>
-            <h3 className={`${prefix}--cardarray__title`}>{title}</h3>
-          </div>
+    <section data-autoid={`${stablePrefix}--cardarray`}>
+      <ContentGroup heading={{ copy: title }}>
+        <div
+          className={`${prefix}--cardarray`}
+          data-autoid={`${stablePrefix}--cardarray`}>
```
photodow: You have the same stable selector twice. Also, would this `className` need to go up above too?

packages/react/src/components/ContentGroup/ContentGroup.js
```diff
@@ -0,0 +1,64 @@
...
+              'heading-4'}`}>
+            {heading.copy}
+          </h3>
+        </div>
+        <div
+          data-autoid={`${stablePrefix}--content-group__children`}
+          className={`${prefix}--content-group__children`}>
+          {children}
+        </div>
+      </div>
```
photodow: I need to add it to the functional specs, but talking to the Web Simplification team I think we're going to need an optional CTA here at the end as well.

image

packages/styles/scss/components/content-group/_content-group.scss
```diff
@@ -0,0 +1,73 @@
...
+      padding-bottom: $carbon--layout-05;
+
+      &__title {
+        margin-bottom: $carbon--layout-05;
+      }
+
+      &__container {
+        @include carbon--make-col(4, 4);
+      }
+    }
```
photodow: I don't think there is a `sm` breakpoint in the code. I think you could remove the `@include carbon--breakpoint(sm) {}` and move all of it's content up a level and it would do the same thing.

packages/styles/scss/components/content-group/_content-group.scss
```diff
@@ -0,0 +1,73 @@
...
+// LICENSE file in the root directory of this source tree.
+//
+
+@import '../../globals/imports';
+
+@mixin themed-items {
+  .#{$prefix}--content-group__title {
+    color: $text-01;
+  }
+}
```
photodow: I'm seeing this pattern used multiple times now, and I'm a little worried we're not thinking about the scope of our `SCSS`. If I'm not mistaken this is just overwriting the other previously imported mixins we created called `themed-items`. I think we can scope it to this component a little better with a better naming convention like `content-group-theme`, or maybe just include its contents within the mixin below. Also, if we continue down this path does it make sense to break this mixin out into its own file? That way we have a single file handling the theming aspect.

photodow: @jeffchew do we need to include `carbon--type-style('expressive-heading-04');` in here to align with the new theming strategy? I don't know, would also some of the `box-modal` styling be needed as well?

packages/styles/scss/components/content-group/_content-group.scss
```diff
@@ -0,0 +1,73 @@
...
+    @include carbon--breakpoint(sm) {
+      padding-top: $carbon--layout-03;
+      padding-bottom: $carbon--layout-05;
+
+      &__title {
+        margin-bottom: $carbon--layout-05;
+      }
+
+      &__container {
+        @include carbon--make-col(4, 4);
```
photodow: Should this be a column, or should it be a container? Also, I'm not seeing the row in this pull request either.

packages/patterns-react/src/patterns/CardArray/CardArray.js
```diff
@@ -28,46 +30,38 @@ const CardArray = ({ title, content }) => {
...
           <div
             data-autoid={`${stablePrefix}--cardarray-group`}
             ref={containerRef}
-            className={`${prefix}--cardarray__col ${prefix}--cardarray-group`}>
+            className={`${prefix}--cardarray__col ${prefix}--cardarray-group ${prefix}--grid--condensed`}>
```
photodow: I see you're using `${prefix}--grid--condensed`, but I'm a little concerned about where it's coming from. I don't see it in your `SCSS` anywhere.

packages/styles/scss/components/content-group/_content-group.scss
```diff
@@ -0,0 +1,73 @@
...
+      max-width: 90%;
+    }
+
+    &__heading-4 {
+      @include carbon--type-style('expressive-heading-04');
+    }
+
+    &__heading-5 {
+      @include carbon--type-style('expressive-heading-05');
+    }
```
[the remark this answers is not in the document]
RaphaelAmadeu: I've seen both heading sizes across the designs pointed on the zenhub issue, that's why I've made it this way

packages/styles/scss/patterns/cardarray/_cardarray.scss
```diff
@@ -16,68 +16,16 @@
...
-        }
-      }
-    }
-    @include carbon--breakpoint('max') {
-      &__col {
-        max-width: carbon--rem(640px);
-      }
+
+    .#{$prefix}--link {
+      outline: 1px solid $ui-background;
```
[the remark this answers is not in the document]
RaphaelAmadeu: This is being used to make the white border around the CardLinks

packages/react/src/internal/FeatureFlags.js
```diff
@@ -16,6 +16,14 @@
  */
 export const DDS_FLAGS_ALL = process.env.DDS_FLAGS_ALL === 'true' || false;
 
+/**
+ * This determines if the tableofcontents will be rendered or not
+ *
+ * @type {string | boolean}
+ */
+export const DDS_CONTENT_GROUP =
+  process.env.DDS_CONTENT_ARRAY_GROUP === 'true' || DDS_FLAGS_ALL || false;
```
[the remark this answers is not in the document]
RaphaelAmadeu: This was fixed in a newer commit

packages/patterns-react/src/patterns/CardArray/CardArray.js
```diff
@@ -80,7 +74,7 @@ const CardArray = ({ title, content }) => {
  */
 const _renderCardArrayItems = contentArray =>
   contentArray.map(elem => (
-    <CardArrayItem title={elem.title} copy={elem.copy} link={elem.link} />
+    <CardArrayItem title={elem.title} copy={elem.copy} href={elem.href} />
```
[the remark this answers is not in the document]
RaphaelAmadeu: This will be applied on the newer commit

packages/react/src/components/ContentGroup/ContentGroup.js
RaphaelAmadeu: Is this going to be another `CardLink`?

packages/patterns-react/src/patterns/CardArray/CardArray.js
RaphaelAmadeu: I've taken this out of the code, as it was for testing

packages/styles/scss/patterns/cardarray/_cardarray.scss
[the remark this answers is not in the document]
photodow: So that's what the condensed grid should be managing.

packages/styles/scss/components/content-group/_content-group.scss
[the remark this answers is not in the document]
photodow: I think we talked about this last week, but I think we're confusing the content section and the content group. From what I'm seeing the content group only has one token.

packages/patterns-react/.env.example
```diff
@@ -1,17 +0,0 @@
...
-DDS_LEADSPACE=<Boolean flag to turn on/off the lead space pattern>
-DDS_LEADSPACE_CENTERED=<Boolean flag to turn on/off the LeadSpaceCentered component>
-DDS_LISTSECTION=<Boolean flag to turn on/off the ListSection component>
-DDS_PICTOGRAM_ARRAY=<Boolean flag to turn on/off the PictogramArray component>
-DDS_SIMPLELONGFORM=<Boolean flag to turn on/off the SimpleLongForm component>
-DDS_TOC=<Boolean flag to turn on/off the TableOfContents component>
-DDS_USECASES=<Boolean flag to turn on/off the UseCases component>
-
-# Carbon Expressive Theme
-PATTERNS_STORYBOOK_CARBON_EXPRESSIVE=<Boolean to turn on/off the carbon expressive theme in storybook>
```
jeffchew: @RaphaelAmadeu why was this file deleted?

packages/styles/scss/components/content-group/_content-group.scss
jeffchew: > I'm seeing this pattern used multiple times now, and I'm a little worried we're not thinking about the scope of our `SCSS`. If I'm not mistaken this is just overwriting the other previously imported mixins we created called `themed-items`. I think we can scope it to this component a little better with a better naming convention like `content-group-theme`, or maybe just include its contents within the mixin below. Also, if we continue down this path does it make sense to break this mixin out into its own file? That way we have a single file handling the theming aspect.

Adding @annawen1 here. I think we're ok for now since the theme mixin is getting referenced right away in the same file, but to be on the safe side it might be worth creating different mixin names for the theming. Not sure if it's necessary to break them out into their own file though.

packages/react/src/components/ContentGroup/ContentGroup.js
photodow: The only examples I have in front of me indicate it's only a `CardLink`, but based on my understanding of how the CTA section works it might be smart to add the `linkWithIcon` as an optional CTA because it all depends on the hierarchy of the page.

`LinkWithIcon` -> `CardLink` -> `FeaturedLink`.

packages/patterns-react/.env.example
RaphaelAmadeu: I've restored it on the last commit, It was deleted by mistake

packages/styles/scss/components/content-group/_content-group.scss
photodow: Yeah, it's getting referenced right before it's used, but this doesn't allow us to call it again in the future. As an example, if this theme needs to be changed in a different context using `SCSS` we might have some issues.

Here is just a quick example of what I mean...
```scss
@import 'components/scss/component-1';
@import 'components/scss/component-2';

.component-3 {
  @include carbon--theme($carbon--theme--g90) {
    @include themed-items;
  }
}
```

Which `themed-items` are we theming in `component-3`? Based on the order it would be `component-2`, and there is now no way to theme `component-1` if that's what we were wanting to do within this context. We use a similar technique in theming the footer accordion and button.

packages/styles/scss/components/content-group/_content-group.scss
RaphaelAmadeu: I've restructured the code in later commits

Issue #875, opened 2019-12-17
Footer Grid is Slightly Off


photodow:
## Detailed description

It looks like the footer alignment got messed up a little in Chrome. I think they should all be lined up on every 4 columns.

![image](https://user-images.githubusercontent.com/3793636/71010419-668c6a80-20b1-11ea-8d35-ab5c0434151c.png)

![image](https://user-images.githubusercontent.com/3793636/71010591-b53a0480-20b1-11ea-963c-5930753d93c0.png)

PR #904, opened 2019-12-30
fix(scss): aligned footer items to the grid

RaphaelAmadeu:
### Related Ticket(s)

Related to ticket: #875 
### Description

Screen Shot 2019-12-30 at 11 27 56
Screen Shot 2019-12-30 at 11 28 09

The SCSS code has been edited in order to fix the alignment of the footer items, so they are within the grid spacing.

### Changelog

**Changed**

- SCSS adjustments


packages/styles/scss/components/footer/_footer-nav.scss
```diff
@@ -45,11 +45,12 @@
 
         display: block;
         column-count: 2;
-        column-gap: 0;
+        column-gap: 32px;
```
annawen1: maybe we could use the `$spacing-07` carbon token here https://www.carbondesignsystem.com/guidelines/spacing

photodow: Maybe better would be `$carbon--grid-gutter` if we do have to go this route.

RaphaelAmadeu: I've applied this change on the latest commit

PR #961, opened 2020-01-06
feat(component): image component

carthick:
(Please Review the files from second commit).
THIS PR IS A COMPANION PR FOR THE FEATURED LINK [FEATURED LINK PATTERN](https://github.com/carbon-design-system/ibm-dotcom-library/pull/852). SHOULD BE REVIEWED ALONG WITH  [FEATURED LINK PATTERN](https://github.com/carbon-design-system/ibm-dotcom-library/pull/852)

### Related Tickets 

[FEATURED LINK PATTERN](https://github.com/carbon-design-system/ibm-dotcom-library/pull/852)

### Description

Creation of the reusable component for the image component.

Screen Shot 2020-01-06 at 11 24 06 AM


packages/react/src/components/ImageComponent/ImageComponent.js
```diff
@@ -0,0 +1,64 @@
...
+        alt={alt}
+      />
+    </picture>
+  );
+};
+
+Image.propTypes = {
+  images: PropTypes.arrayOf(
+    PropTypes.shape({
+      minWidth: PropTypes.number,
```
jeffchew: I'm wondering if we should be referencing the breakpoint names rather than actual numbers here (e.g. `sm`, `md`, `lg` `xlg`, 'max`), then we would render out the mappings rather than have the user rely on getting the numbers right?

https://www.carbondesignsystem.com/guidelines/2x-grid/implementation#responsive-options

photodow: We should probably allow for both. There are moments where the adopters might need to break something down in between the standard breakpoints. So something like a `switch` or `if` statement that allows us to catch the keywords first and set the appropriate breakpoint, but then on `default` set the breakpoint to whatever the user passed in.

packages/react/src/components/ImageComponent/ImageComponent.js
```diff
@@ -0,0 +1,64 @@
...
+ * renders background image
+ *
+ * @param {object} props props object
+ * @param {object} props.images array of images used for diff breakpoints
+ * @param {string} props.defaultImage default image (usually image for largest breakpoint)
+ * @param {string} props.alt alt of the image
+ * @returns {*} picture element
+ */
+const Image = ({ images, defaultImage, alt }) => {
+  const sortedImages = sortSources(images);
```
photodow: Is the order of these an important factor? I thought if the media doesn't match the browser will just go onto the next one, but maybe I missed something in my understanding.

packages/react/src/components/ImageComponent/ImageComponent.js
```diff
@@ -0,0 +1,64 @@
...
+
+  return (
+    <picture data-autoid={`${stablePrefix}--imagecomponent`}>
```
photodow: I don't think we need the word "component" in the name of the component. I'm double-checking with @wonilsuhibm to see if we aligned on a name for this yet. I threw out just `Picture`, or  I think `ResponsivePicture` was suggested, although it seems like there are bigger plans for this component. I think @wonilsuhibm suggested `EmbeddedPicture`. I kind of either like just the simple `Picture`, or maybe a `RichPicture` because I think there might be more features to come on this down the road.

jeffchew: Agreed, `component` is probably not necessary here.

packages/react/src/components/ImageComponent/ImageComponent.js
jeffchew: No, my understanding is that order isn't important.

packages/react/src/components/ImageComponent/ImageComponent.js
```diff
@@ -0,0 +1,64 @@
...
+      {sortedImages.map((imgSrc, key) => {
+        return <source key={key} srcSet={imgSrc.url} />;
```
photodow: I'm not seeing a `media` attribute applied here. Isn't that needed to make this work?

packages/react/src/components/ImageComponent/ImageComponent.js
carthick: @jeffchew @photodow I am not sure about the order of these but i will update them in order ....

packages/react/src/components/ImageComponent/ImageComponent.js
carthick: @jeffchew @photodow Should i update that to RichPicture or Embedded picture or Just picture something like that?

packages/react/src/components/ImageComponent/ImageComponent.js
```diff
@@ -0,0 +1,64 @@
...
+const sortSources = sources => {
+  return sources.sort((a, b) => (a.minWidth > b.minWidth ? -1 : 1));
+};
+
+/**
+ * renders background image
+ *
+ * @param {object} props props object
+ * @param {object} props.images array of images used for diff breakpoints
+ * @param {string} props.defaultImage default image (usually image for largest breakpoint)
```
photodow: Is the `defaultImage` necessary? It looks like we're already looping through the array, maybe we could just pull out the largest/smallest, first, or last image, or maybe allow them to specify? Something like this maybe?

```jsx
<Picture
  alt=""
  breakpoints={[{
    breakpoint: 'lg',
    src: 'https://example.com/lg.jpg'
  }, {
    breakpoint: 'xlg',
    src: 'https://example.com/xlg.jpg'
  }, {
    breakpoint: 'md',
    src: 'https://example.com/md.jpg'
  }]}
/>
```

```html
<picture>
  <source srcset="https://example.com/lg.jpg" media="(min-width: 1056px)" />
  <source srcset="https://example.com/xlg.jpg" media="(min-width: 1312px)" />
  <source srcset="https://example.com/md.jpg" media="(min-width: 675px)" />
  <img class="bx--picture__img" src="https://example.com/lg.jpg" alt="" />
</picture>
```

Grabbing the first one would probably be the easiest for now, and we could always expand from there if we need to. What do you think?

jeffchew: Perhaps the logic should be in the pre-defined breakpoints, grab those in the order of availability (`lg`, `md`, etc) and if none are available, use the first one?

packages/react/src/components/ImageComponent/ImageComponent.js
jeffchew: Good catch, yes `media` is important here.

packages/react/src/components/ImageComponent/ImageComponent.js
photodow: I added this as a comment over in #970, and thought it might be useful in this scenario for you to consider as well. Carbon exports their `breakpoints` in Javascript allowing us to use them for things like this.

https://github.com/carbon-design-system/carbon/blob/master/packages/layout/src/index.js#L45-L71

packages/react/src/components/ImageComponent/ImageComponent.js
annawen1: the order does matter for the <picture> element!

packages/react/src/components/ImageComponent/__stories__/ImageComponent.stories.js
```diff
@@ -0,0 +1,33 @@
...
+  .add('Default', () => {
+    const defaultImages = text(
+      'Default image:',
+      'https://picsum.photos/id/2/672/672'
+    );
+    const imageObject = object('Images Object:', [
+      'https://picsum.photos/id/2/320/160',
+      'https://picsum.photos/id/2/400/400',
+      'https://picsum.photos/id/2/672/672',
+    ]);
```
annawen1: image object should be an array of objects
```
image: [
  {
     minWidth: 0,
     url: .....
  },
 {
     minWidth: 320,
     url: .....
  },
 {
     minWidth: 1024,
     url: .....
  }
]
```

packages/react/src/components/ImageComponent/ImageComponent.js
carthick: updated @photodow

packages/react/src/components/ImageComponent/ImageComponent.js
carthick: still waiting for the suggestions

packages/react/src/components/ImageComponent/ImageComponent.js
carthick: updated updated the media attribute

packages/react/src/components/ImageComponent/ImageComponent.js
carthick: added break points to the logic

packages/react/src/components/ImageComponent/__stories__/ImageComponent.stories.js
carthick: updated the object with array of objects

packages/react/src/components/ImageComponent/ImageComponent.js
jeffchew: Let's go with `Picture`, thoughts?

photodow: Yeah, let's go with `Picture`. Sorry, I was working to get a response from Wonil, but I guess he's been too busy to respond.

packages/react/src/components/ImageComponent/ImageComponent.js
photodow: @annawen1 I think I see now. I had to code it out to visualize it. haha

https://codepen.io/photodow/pen/RwNMJBx?editors=1000

If you use `min and max` (not sure about browser support) it doesn't matter as much as long as you're consistent, but if you use either `min`, or `max` then you have to be more conscientious about it.

packages/react/src/components/ImageComponent/ImageComponent.js
photodow: @jeffchew @carthick I finally heard back from Wonil today. Could we go with `Image` instead? Wonil was worried `Picture` sounds too much like something like a product image, and doesn't account for things like charts and such. It has a previous meaning to some of the other designers.

carthick: NP @photodow I have updated that to Image ....

packages/react/src/components/Image/README.md
```diff
@@ -0,0 +1,78 @@
...
+| Name       | Data Type | Description        |
+| ---------- | --------- | ------------------ |
+| `src`      | String    | Url of Image.      |
+| `minWidth` | int       | min Width of image |
+
+## Stable selectors
+
+| Name                         | Description |
+| ---------------------------- | ----------- |
+| `dds--imagecomponent__image` | Interactive |
```
jeffchew: @carthick you're missing `$dds--imagecomponent`

packages/react/src/components/Image/__stories__/index.scss
```diff
@@ -0,0 +1,6 @@
+//
+// Copyright IBM Corp. 2016, 2018
+//
+// This source code is licensed under the Apache-2.0 license found in the
+// LICENSE file in the root directory of this source tree.
+//
```
annawen1: might want to import the the associated styles here:
```
@import '../../../../../styles/scss/components/imagecomponent/imagecomponent.scss';
```

PR #967, opened 2020-01-07
fix(utility): fixed resizing issues on the sameheight utility

RaphaelAmadeu:
### Description

Emerson and Mahesh found some issues regarding the `sameheight` utility, so I've tried to improve it and correct the problems on resize.

### Changelog

**Changed**

- JS function rewritten


mkothur:
@RaphaelAmadeu One miner thing could you change the utility function name as camelcase like `sameHeight` thanks!

PR #970, opened 2020-01-07
fix(utility): changed same-height utility name to camel case

RaphaelAmadeu:
### Description

Emerson and Mahesh found some issues regarding the `sameheight` utility, so I've tried to improve it and correct the problems on resize.

### Changelog

**Changed**

- JS function rewritten


packages/utilities/src/utilities/sameheight/sameheight.js
```diff
@@ -20,7 +20,7 @@
  * @param {string} minSize Minimum size for the utility to be activated, empty for small,
  *  md for medium, lg for large, xlg for xlarge, max for maximum
  */
-function sameheight(elemCollection, minSize) {
+function sameHeight(elemCollection, minSize = false) {
   const elemArr = Array.prototype.slice.call(elemCollection);
   let targetWidth = 0;
   if (minSize) {
```
photodow: I'm looking at lines `26-41`, and I'm a little nervous we've hard coded the breakpoints in here. Can we leverage the `@carbon/layout` package to get the proper breakpoints from the source?

https://github.com/carbon-design-system/carbon/blob/master/packages/layout/src/index.js#L45-L71

photodow: If I'm understanding what you're doing here you could just do something along the lines of the following...

```javascript
import { breakpoints } from '@carbon/layouts';

function sameheight(elemCollection, minSize) {
  const target = breakpoints[minSize];
  const targetWidth = target ? parseFloat(target.width) * baseFontSize : 0;

  ...
}
```

`parseFloat()` and `baseFontSize` takes `1rem` and converts to `16` for our needs.

packages/utilities/src/utilities/sameheight/sameheight.js
```diff
@@ -41,19 +41,26 @@ function sameheight(elemCollection, minSize) {
...
         : false;
     });
 
     elemArr.forEach(elem => {
-      elem.offsetHeight == targetHeight
-        ? (elem.style.height = 'auto')
-        : (elem.style.height = targetHeight + 'px');
+      elem.offsetHeight == targetHeight;
+      elem.style.height = targetHeight + 'px';
+    });
```
photodow: Help me understand why we're looping through this three times. I'm concerned about performance here...

```javascript
let targetHeight = 0;

elemArr.forEach(elem => {
  elem.style.height = 'auto';
  if (elem.offsetHeight > targetHeight) {
    targetHeight = elem.offsetHeight;
  }
});

elemArr.forEach(elem => {
  if (elem.offsetHeight !== targetHeight) {
    elem.style.height = targetHeight + 'px';
  }
});
```

photodow: While on the topc of performance, something else that might help reduce some unnecessary calculations and loops is if we add a timeout to this function.

```javascript
window.sameHeightProgress = {};

function sameheight(uniqueID, elemCollection, minSize) {
  clearInterval(window.sameHeightProgress[uniqueID]);
  window.sameHeightProgress[uniqueID] = setInterval(() => {
    // do stuff
  }, 200);
}
```

I also wonder if we passed in the selector instead of the elements themselves if we could prevent resizing the same element more than once, and maybe use that as the `uniqueID` as well?

jeffchew: Would `requestAnimationFrame` work out better here actually?

jeffchew: was just talking about this with @RaphaelAmadeu , while this is definitely valid, it seems like it's more on the execution side rather than the utility itself. We could possibly add it in the jsdoc, something like :
```
window.addEventListener('resize', function() {
    window.requestAnimationFrame(function() {
        sameheight(ElementArray, 'md');
    });
}, true);
```

but might be ok the way it is...

photodow: @jeffchew @RaphaelAmadeu just trying to think through this a little. Is there ever a scenario where we wouldn't want to use `sameHeight`, and `requestAnimationFrame` at the same time? If we take the function and want to use it as a one off (say `onLoad`) it still executes once? Does it slow the `onLoad` creating a noticeable "flash"?

If there is no difference I feel like we would be reducing some overhead if we included it within the function.

Is there any performance hits for running multiple `requestAnimationFrame` at a time? If so I could see the suggestion above is better because if you use this in multiple "sections" as-is you'd call it once for each one.

```javascript
window.addEventListener('resize', function() {
    window.requestAnimationFrame(function() {
        sameheight(section1, 'md');
        sameheight(section2, 'md');
        sameheight(section3, 'md');
    });
}, true); 
```

If we don't include it in the function we're opening the risk of adopters using it wrong and creating performance issues, right? I guess at the end of the day, that's not our problem really...

I lean towards including it personally, but either way works in the end, right?

jeffchew: At the same time, we should put our shoes into the adopter. Generally speaking, when I look for a plugin or a utility, I don't look inside to see what it's doing unless something is wrong. If there is already a `requestAnimationFrame` set up in a method that in general doesn't need it, I wouldn't expect it to be there. And if I then set up my listener, what if I set it up on my end? That makes it two `requestAnimationFrames`, one in my adopter side event listener, and another one in the function itself. I wonder if there would be any side effects if we had nested requestAnimationFrame methods.

I think I am leaning more towards adding it in the JSDoc and be part of the documentation, and not in the function.

Issue #705, opened 2019-11-19
Pattern: Develop an alpha version of the cards with images


RobertaJHahn:
#### User Story

> As a `[user role below]`:
DDS developer

> I need to:
work closely with the Functional Analyst and designer to begin developing the cards with images

> so that I can:
become involved earlier in the design process and functional documentation to compress the schedule and develop a higher quality pattern

#### Additional information

- Story within Storybook with corresponding knobs
- minimum 80% unit test coverage
- Utilize Carbon
- [Content array with cards design story #656](https://app.zenhub.com/workspaces/ibmcom-library-5d449f3642eb1962336cbe52/issues/carbon-design-system/ibm-dotcom-library/656)
- [Content array with cards functional specs #682](https://app.zenhub.com/workspaces/ibmcom-library-5d449f3642eb1962336cbe52/issues/carbon-design-system/ibm-dotcom-library/682)

#### Acceptance criteria
 - [ ] alpha react component within the Patterns storybook
 - [ ] create a feature flag for this component
 - [ ] Gap analysis of alpha code and design & functional specs
 - [ ] If applicable, create and size stories for additional work

mkothur:
PR Link (https://github.com/carbon-design-system/ibm-dotcom-library/pull/803)

PR #983, opened 2020-01-08
feat(pattern): created an alpha version of CardSection pattern layout

mkothur:
### Related Ticket(s)

Related to issue #705
### Description

Created the alpha version of the Cards with images and made changes in `/packages/react/src/components/CardLink` added one pore property `source` and function to render an image in `CardsWithImageGroup`

image
image
image

### Changelog

**Changed**

- Added CardWithImage component and styles updated the CardLink react componet.
- Added a new feature flag for the created component.


packages/styles/scss/components/card-link/index.scss
```diff
@@ -8,21 +8,30 @@
...
     display: flex;
     flex-direction: column;
-    padding: $carbon--spacing-05;
 
     > *:not(:last-child) {
       margin-bottom: $carbon--spacing-07;
     }
 
+    &__wrapper {
+      padding: $spacing-05;
```
annawen1: can you change the token to `$carbon--spacing-05`? That may fix the build error

mkothur: yeah, might be.

PR #1019, opened 2020-01-14
feat(pattern): updated list-section to use the content-group

RaphaelAmadeu:
### Related Ticket(s)

#745 

### Description

I've updated the `ListSection` pattern, so it uses the `ContentGroup` component

### Changelog

**Changed**

- JS and  SCSS code for the `ListSection`

**Removed**

- SCSS code that was not being used


packages/patterns-react/src/patterns/ListSection/ListSectionGroup.js
```diff
@@ -24,14 +25,22 @@ const { prefix } = settings;
  */
 const ListSectionGroup = ({ listGroup: { title, lists } }) => {
```
photodow: Is `ListSectionGroup`, and `ContentGroup` kind of the same thing? Does it make sense to remove this file, and use `ContentGroup similarly to how `ListSectionGroup` was being used in `ListSection`?

packages/styles/scss/patterns/listsection/_listsection-group.scss
```diff
@@ -6,16 +6,14 @@
...
+  }
 
-      margin-top: $carbon--layout-05;
-    }
-    @include carbon--breakpoint('lg') {
-      &__title {
-        margin-top: $carbon--layout-06;
-      }
+  .#{$prefix}--content-group {
+    padding-bottom: $carbon--layout-03;
```
photodow: Is this the right place for this? Should this be apart of the `ContentGroup`?

packages/styles/scss/patterns/listsection/_listsection-group.scss
```diff
@@ -6,16 +6,14 @@
 //
 
 @mixin listsection-group {
-  .#{$prefix}--listsection-group {
-    &__title {
-      @include type-style('expressive-heading-04');
+  .#{$prefix}--content-group__title {
+    padding-left: 0;
```
photodow: This shouldn't be necessary... Looking at the `ContentGroup` it looks as if there are still some grid stuff in there?

packages/patterns-react/src/patterns/ListSection/ListSectionGroup.js
jeffchew: Right...it seems like `ListSectionGroup` can be removed entirely and just use `ContentGroup` directly?

photodow: @jeffchew yeah, that's what I was trying to suggest. Sorry if it came out a little confusing.

packages/styles/scss/patterns/listsection/_listsection-group.scss
jeffchew: I removed it and looks to be adding some padding to the left of the title by default. I say let's leave this in here for now, and come back to it when we refactor the MixedArray.

jeffchew:
@photodow fortunately, @RaphaelAmadeu gave me contributor access to his fork before he left on vacation. I consolidated the `ListSectionGroup` and think this should be good for now until we refactor the `MixedArray`.

Issue #976, opened 2020-01-08
[ImageCards] Updates to condensed grid per PR feedback


jeffchew:
@mkothur naw, @annawen1's right. We need the spacing inbetween. That comes from the Condensed Grid. We don't have to build that into each component, we just have to use what's already there. If Carbon decides one day they want to change the condensed grid to `3px`, then we have to go to a bunch of components and patterns to make sure we updated.

Here are a couple resources...
https://github.com/carbon-design-system/carbon/blob/master/packages/grid/scss/_mixins.scss#L19-L44
https://carbon-elements.netlify.com/grid/examples/preview/#condensed-grid
https://www.carbondesignsystem.com/guidelines/2x-grid/implementation/#the-condensed-grid

Reach out if you have any questions or concerns.

_Originally posted by @photodow in https://github.com/carbon-design-system/ibm-dotcom-library/pull/803_

photodow:
Here's a reduced case using codesandbox to show how to use the condensed grid. It's already in the `CSS`, so you just have to use the proper class name `bx--grid--condensed` at a parent level...

https://codesandbox.io/s/pedantic-galileo-o7ei3

PR #1034, opened 2020-01-14
fix(pattern): fixed gutter between cards using gridCondensed property

mkothur:
### Related Ticket(s)

Related to ticket #976

### Description

Added the grid condensed property in cardSection pattern which will help to create spacing between cards. And restructure the JSX little bit to match the CSS changes.

And also added margin in logo_grid CSS to look nicer.

### Changelog

**Changed**

CardSection - js and scss
logGrid - scss

### DIFF
--- a/packages/patterns-react/src/patterns/CardSection/CardSection.js
+++ b/packages/patterns-react/src/patterns/CardSection/CardSection.js
@@ -67,24 +67,36 @@ const CardSection = ({ title, cards, theme }) => {
       className={classNames(`${prefix}--card-section`, setTheme(theme))}>
       <div className={`${prefix}--card-section__container`}>
         <div className={`${prefix}--card-section__row`}>
-          <div className={`${prefix}--card-section__col`}>
+          <div className={`${prefix}--card-section__col--left`}>
             <h2 className={`${prefix}--card-section__title`}>{title}</h2>
+          </div>
+          <div className={`${prefix}--card-section__col--right`}>
             <div
-              className={`${prefix}--card-section__cards`}
+              className={classNames(
+                `${prefix}--card-section__cards`,
+                `${prefix}--grid--condensed`
+              )}
               ref={containerRef}>
-              {cards.map((card, index) => {
-                return (
-                  <CardLink
-                    key={index}
-                    image={card.image}
-                    title={card.title}
-                    content={card.copy}
-                    href={card.link.href}
-                    target={card.link.target}
-                    icon={<ArrowRight20 />}
-                  />
-                );
-              })}
+              <div className={`${prefix}--card-section__cards__container`}>
+                <div className={`${prefix}--card-section__cards__row`}>
+                  {cards.map((card, index) => {
+                    return (
+                      <div className={`${prefix}--card-section__cards__col`}>
+                        <CardLink
+                          key={index}
+                          imgSrc={card.imgSrc}
+                          altText={card.altText}
+                          title={card.title}
+                          content={card.copy}
+                          href={card.link.href}
+                          target={card.link.target}
+                          icon={<ArrowRight20 />}
+                        />
+                      </div>
+                    );
+                  })}
+                </div>
+              </div>
             </div>
           </div>
         </div>
--- a/packages/patterns-react/src/patterns/CardSection/CardSection.js
+++ b/packages/patterns-react/src/patterns/CardSection/CardSection.js
@@ -67,36 +67,25 @@ const CardSection = ({ title, cards, theme }) => {
       className={classNames(`${prefix}--card-section`, setTheme(theme))}>
       <div className={`${prefix}--card-section__container`}>
         <div className={`${prefix}--card-section__row`}>
-          <div className={`${prefix}--card-section__col--left`}>
-            <h2 className={`${prefix}--card-section__title`}>{title}</h2>
-          </div>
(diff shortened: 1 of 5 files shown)


packages/styles/scss/patterns/card-section/_card-section.scss
```diff
@@ -50,96 +36,61 @@
...
+          @include carbon--make-col(4, 8);
+        }
+        @include carbon--breakpoint('lg') {
+          @include carbon--make-col(1, 3);
+        }
 
-      > *:not(:last-child) {
-        margin-bottom: 0;
+        margin-top: $carbon--grid-gutter--condensed / 2;
+        margin-bottom: $carbon--grid-gutter--condensed / 2;
```
photodow: I think I saw https://github.com/carbon-design-system/carbon/pull/4960 got merged in. These lines of code may not be necessary anymore.

mkothur: you mean we could remove these props and just have `${prefix}--grid--condensed` this in JSX right?

mkothur: But we do need `margin-bottom: $carbon--grid-gutter--condensed / 2` this prop in that  I guess. Because when I remove both of them it looks like this.
image

photodow: Looks like they might not have releasee it yet for some reason. I thought they were doing nightly builds. I guess we can wait a little longer.

mkothur: Okay.

photodow:
This is looking a lot cleaner! 🎉

With that said there are some minor grid alignment issues we need to resolve.

It doesn't have a `max-width` because it looks like you didn't include the `bx--make-container` mixin.
image

Unless it's supposed to be full width, in that case it's still slightly off.
image


------


I think the learn template maxes out its width out, but in general should we plan for when the adopter uses the full-width grid modifier? @jeffchew @annawen1 @kennylam @RaphaelAmadeu

packages/styles/scss/patterns/card-section/_card-section.scss
```diff
@@ -50,96 +36,61 @@
...
-      @include carbon--make-col(8, 8);
+      &__container {
+        @include carbon--make-container;
+      }
+      &__row {
+        @include carbon--make-row;
+      }
+      &__col {
+        @include carbon--make-col-ready;
+        @include carbon--make-col(4, 4);
```
photodow: When we added this line earlier I wasn't really thinking. All it does is set 100%. Since it's 100% be default I don't think it's needed.

packages/patterns-react/src/patterns/CardSection/CardSection.js
```diff
@@ -67,25 +67,36 @@ const CardSection = ({ title, cards, theme }) => {
...
-                    title={card.title}
-                    content={card.copy}
-                    href={card.link.href}
-                    target={card.link.target}
-                    icon={<ArrowRight20 />}
-                  />
-                );
-              })}
+              <div className={`${prefix}--card-section__cards__container`}>
+                <div className={`${prefix}--card-section__cards__row`}>
```
photodow: So we talked a little about this on the call this morning. I think this can be simplified. Now that we have it working let's try to get rid of what we don't need. I think reducing some of this will also resolve some of the alignment issues I'm seeing.

Take a look at the following...

```jsx
<div className= className={`${prefix}--card-section__container`}>
  <div className={`${prefix}--card-section__row`}>
    <h2 className={`${prefix}--card-section__title`}>{title}</h2>
    <div className={`${prefix}--card-section__cards`}>
      <div className={`${prefix}--card-section__cards ${prefix}--row--condensed`}>
      {cards.map((card, index) => {
        return (
          <div className={`${prefix}--card-section__cards__col`}>
            <CardLink />
          </div>
        );
      })}
    </div>
  </div>
</div>
```

```scss
.card-section {
  @include carbon--make-container;

  &__row {
    @include carbon--make-row;
  }

  &__title {
    @include carbon--make-col(4, 16);
    // add your breakpoints and spans.
  }

  &__cards {
    @include carbon--make-col(12, 16);
    // add your breakpoints and spans.

    &__row {
      @include carbon--make-row;
    }
    &__col {
      @include carbon--make-col(1, 3);
      // add your breakpoints and spans.
    }
  }
}
```

Notice how `col--left` and `col--right` was moved onto `title` and `cards`, and how we removed `cards__container`.

packages/patterns-react/src/patterns/CardSection/__stories__/CardSection.stories.js
```diff
@@ -77,7 +77,7 @@ const cardsGroup = {
     },
     {
       image: {
-        defaultImage: 'http://picsum.photos/id/1003/1056/480',
+        defaultImage: 'https://picsum.photos/id/1018/1056/480',
```
photodow: I can't add a comment, but on line `L#67` above. The image isn't rendering. I added `https`, and it seemed to work for me?

mkothur: yeah @photodow: the `s` is missing from the image URL  because of the feature link  commit from @carthick 
image
Let me update the PR.

Issue #1047, opened 2020-01-15
Expressive theme – Notification


oliviaflory:
### Inline
The Action button isn't vertically centered / aligned with the heading and close icon
![Screen Shot 2020-01-15 at 11.06.56 AM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/a8ce3f5f-c3d3-49e6-8b06-61270a52dc44)

 *Note that the Action button is meant to remain at the top when the notification text extends to multiple lines, see example from the Carbon storybook below*
![Screen Shot 2020-01-15 at 11.10.29 AM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/3e2ed9c6-9f6d-4a05-bbb4-7d38b3a65c8c)

### Example link

We are missing the "Example link" that is present in the Carbon storybook, I want to make sure it's aligned correctly on our side. I found it in the Story tab on the Carbon storybook, but couldn't edit the story in our storybook. I was able to edit in HTML, and it looked ok but want to confirm :) 

**Carbon**
![Screen Shot 2020-01-15 at 10.54.24 AM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/2c1926db-ad51-4f3a-92b1-b3153391a177)

![Screen Shot 2020-01-15 at 10.54.46 AM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/a76ec446-2492-49e6-bd39-3cdb0ec93b6d)

**Expressive**
![Screen Shot 2020-01-15 at 10.55.51 AM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/19c35158-2b0a-406e-b255-d035e3043633)

PR #1053, opened 2020-01-16
feat(expressive): inline notifications action button styles

mkothur:
### Related Ticket(s)

Expressive theme – Inline notification #1047

### Description
Expressive theme – Notification

The Action button isn't vertically centered/aligned with the heading and close icon

image

### Changelog

**Changed**

  Adjust the padding of Action button and heading and subtitle styles.
 image


packages/styles/scss/themes/expressive/components/_inline-notifications.scss
```diff
@@ -0,0 +1,31 @@
...
+/// Inline Notifications styles (expressive)
+/// @access private
+/// @group inline-notifications-expressive
+@mixin inline-notifications-expressive {
+  .#{$prefix}--inline-notification {
+    .#{$prefix}--inline-notification__action-button.#{$prefix}--btn--ghost {
+      padding: calc(0.375rem - 3px) 1rem;
+    }
+    .#{$prefix}--inline-notification__title {
+      @include type-style('productive-heading-01');
```
annawen1: So we probably don't want to set type styles here since they should already be set from Jeff's type tokens (`_tokens.scss`). You may have to do some digging to see why the font type does not match up

packages/styles/src/carbon-stories/Notification/Notification-story.js
```diff
@@ -27,6 +27,11 @@ const notificationProps = () => ({
   role: text('ARIA role (role)', 'alert'),
   title: text('Title (title)', 'Notification title'),
   subtitle: text('Subtitle (subtitle)', 'Subtitle text goes here.'),
```
annawen1: Can you remove this `subtitle` field since you are adding a new `subtitle` field?

packages/styles/scss/themes/expressive/components/_inline-notifications.scss
annawen1: Looking at the ticket, I don't see anything about the font type being incorrect - do we need to change the title and subtitle?

mkothur:
yeah, @annawen1 the reason I added the font type for subtitle and title when I see the actual carbon-design-system/carbon storybook font styles are different so that reason added. Yeah correct not mentioned in ticket will remove them.

Issue #1054, opened 2020-01-16
Expressive theme – Code snippet


oliviaflory:
### Inline
:star: inline variation looks good! No changes at this time.

### Single line
![Screen Shot 2020-01-15 at 5.43.59 PM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/5d7f0f1d-0cdf-4a3d-9d4c-ea70eda9973c)

- [ ] update container height to 48px

![Screen Shot 2020-01-15 at 5.44.08 PM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/953bec29-ba3d-41d7-942a-b715c5cc58da)

- [ ] update copy icon to 20x20 px
- *Note this might be using the `copy button`, if it is we should update that first, [seen here](https://carbon-expressive.mybluemix.net/?path=/story/copybutton--default)

### Multi line
![Screen Shot 2020-01-16 at 10.50.15 AM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/b63f239f-bd08-4d50-80cd-736d59aad255)

- [ ] update copy icon to 20x20px
- [ ] once icon is updated let's look at updating the hover state to 48x48 (it might have to be 40x40 because it's only 32x32 right now)

![Screen Shot 2020-01-16 at 10.50.38 AM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/230399af-4360-4cde-bdf9-612fecdbc956)

- [ ] update chevron icon to 20x20px
- [ ] I thought this button would have remained 40px tall if it's using the `small button` variation, but now it's 44px. Can we look into if this is using a unique version of the small button?

PR #1060, opened 2020-01-16
feat(expressive): styles added for code snippet

mkothur:
### Related Ticket(s)

Expressive theme –  #1054

### Description
Expressive theme – Code snippet: update copy icon size and container height and show more hover state size
  single-line
  Multiline
image
image
image
image
image

### Changelog

**Changed**
     code-snippet.scss


annawen1:
@oliviaflory can you take a look at the changes? Sounded like you might want to adjust some of the sizing

packages/styles/scss/themes/expressive/components/_code-snippet.scss
```diff
@@ -0,0 +1,43 @@
...
+    height: rem(20px);
+    width: rem(20px);
+  }
+  // Multi Line Snippet Copy Button
+  .#{$prefix}--snippet--multi .#{$prefix}--snippet-button {
+    height: $carbon--spacing-08;
+    width: $carbon--spacing-08;
+  }
+  .#{$prefix}--btn--ghost {
+    border-width: 1px;
```
annawen1: why are we changing the border width here?

packages/styles/scss/themes/expressive/components/_code-snippet.scss
```diff
@@ -0,0 +1,43 @@
...
+    width: 3rem;
+  }
+  //Single Line Snippet Copy Button
+  .#{$prefix}--snippet__icon {
+    height: rem(20px);
+    width: rem(20px);
+  }
+  // Multi Line Snippet Copy Button
+  .#{$prefix}--snippet--multi .#{$prefix}--snippet-button {
+    height: $carbon--spacing-08;
```
annawen1: @oliviaflory did you want only the hover state of the copy button to be 48px by 48px? or did you want the button to always be 48px regardless of what state it is in?

packages/styles/scss/themes/expressive/components/_code-snippet.scss
mkothur: @annawen1 In the ticket description saying the button height would be 40px but it has 44px, when I looked it has 1px border is adding from four sides it is becoming 44px so I reduced the border-width to 1px.

packages/styles/scss/themes/expressive/components/_code-snippet.scss
annawen1: Also looking at the carbon code, it is using the `<CopyButton />` component so @mkothur we should probably move those changes to the `<CopyButton />` button instead like mentioned in the ticket.

mkothur: you mean to create another component for _copy-button.scss to have them in this file or change the file name _code-snippet.scss to _copy-button.scss?

packages/styles/scss/themes/expressive/components/_code-snippet.scss
annawen1: mmm we probably don't want to do that because it will affect the focus state of the buttons

ie: left is border-width of 3px and middle button is border-width of 1px
Screen Shot 2020-01-17 at 11 04 15 AM

You might want to check the `<Button />` component (small variation) styles and see why it's not 40px

packages/styles/scss/themes/expressive/components/_code-snippet.scss
annawen1: I mean create another file `_copy-button.scss` for the CopyButton component yes

packages/styles/scss/themes/expressive/components/_code-snippet.scss
mkothur: Okay

Issue #1085, opened 2020-01-17
Expressive theme – time picker


oliviaflory:
### Default
![Screen Shot 2020-01-17 at 9.38.58 AM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/50a1944c-ac8a-4478-a5ff-7e60fe7504e2)

- [ ] update chevron icon to match the size we used in [dropdown](https://carbon-expressive.mybluemix.net/?path=/story/dropdown--default)

annawen1:
Looking at the codebase, `<TimePicker />` uses a different icon compared to `<DropDown />` component. It uses `<ChevronDownGlyph />` which sets the svg at width=10 and height=6. This makes bumping up the icon to 20x20 look weird.

We are thinking of opening a ticket with Carbon or contributing back changes to update the icon to `<ChevronDown16 />`

annawen1:
PR to change icon submitted to Carbon: https://github.com/carbon-design-system/carbon/pull/5115

PR #1087, opened 2020-01-17
feat(expressive): updated the chevron icon in timepicker as per dropdown

mkothur:
### Related Ticket(s)

Expressive theme - #1085

### Description
Updated the chevron icon in time picker as per the drop-down component.
image
image

### Changelog

**Changed**

  height of chevron inside the select of time-picker.

### DIFF
--- a/packages/styles/scss/themes/expressive/components/_time-picker.scss
+++ b/packages/styles/scss/themes/expressive/components/_time-picker.scss
@@ -19,4 +19,9 @@
     height: carbon--rem(48px);
     width: carbon--rem(85px);
   }
+  .#{$prefix}--time-picker__select {
+    svg {
+      height: 100%;
+    }
+  }
 }


packages/styles/scss/themes/expressive/components/_time-picker.scss
```diff
@@ -19,4 +19,9 @@
     height: carbon--rem(48px);
     width: carbon--rem(85px);
   }
+  .#{$prefix}--time-picker__select {
+    svg {
+      height: 100%;
```
annawen1: Not sure if setting the `height` changes anything

I think what @oliviaflory was trying to say in the ticket - if you compare the dropdown chevron and the time picker chevron, the time picker chevron looks bigger


Screen Shot 2020-01-17 at 11 36 23 AM
Screen Shot 2020-01-17 at 11 36 39 AM

mkothur: Okay got it

mkothur: @annawen1: after looking into this, the `Dropdown` component is using the lighter version of the chevron icon. And time picker is using a darker version icon. Where these icons are being referenced in the themes.

annawen1:
Don't think this PR is necessary anymore as we are switching the icon to use the same one as the `<Select />` component which already has expressive styles applied to it. 

PR submitted to Carbon: https://github.com/carbon-design-system/carbon/pull/5115

PR #1118, opened 2020-01-21
feat(storybook): breaking out sections in storybook

jeffchew:
### Related Ticket(s)

No related issues

### Description

This is a slight modification to Storybook where we can break out our stories into sections. I think it will be easier to navigate through the patterns now that we're thinking about the hierarchy of these patterns and separating them out as such in Storybook (sections vs blocks vs sub-patterns).

### Changelog

**Changed**

- Adding separators/sections in Patterns storybook

### DIFF
--- a/packages/patterns-react/src/patterns/CardSection/__stories__/CardSection.stories.js
+++ b/packages/patterns-react/src/patterns/CardSection/__stories__/CardSection.stories.js
@@ -5,133 +5,12 @@ import { DDS_CARD_SECTION } from '../../../internal/FeatureFlags';
 import ImageCards from '../ImageCards';
 import React from 'react';
 import SimpleCards from '../SimpleCards';
+import cards from './data/cards.json';
 import readme from '../README.md';
 import { storiesOf } from '@storybook/react';
 
-const cardsGroup = {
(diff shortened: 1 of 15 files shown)


packages/patterns-react/src/patterns/CardSection/__stories__/CardSection.stories.js
```diff
@@ -5,133 +5,12 @@ import { DDS_CARD_SECTION } from '../../../internal/FeatureFlags';
...
-        text: 'Learn more',
-        target: '_self',
-      },
-    },
-  ],
-};
-
 if (DDS_CARD_SECTION) {
-  storiesOf('CardSection', module)
+  storiesOf('Sub-Patterns|CardSection', module)
```
photodow: This is the only one that I think is incorrect at the moment. I think it belongs in `Section`, but I understand if we want to keep this the way it is until we bring in other sub-patterns so we have this placeholder as an example.

jeffchew: Groovy yeah, I will fix this in the other PR where I re-org everything.

PR #1134, opened 2020-01-22
feat(patterns): patterns package re-org

jeffchew:
### Related Ticket(s)

#1135 

### Description

This is a massive re-organization of the patterns, which breaks it down into `sections`, `blocks`, and `sub-patterns`. Another **VERY** major change is that patterns are now merged into the same package as `react`. Patterns will remain in the `patterns` subfolder, but will be in the same `src` folder as `components`. 

Patterns in general have been also separated into the corresponding subfolders to match their new categorizations. The following components have also been re-classified as `sub-patterns`:

- ButtonGroup
- CardLink
- ContentGroup
- Layout
- TableOfContents

### Changelog

**Changed**

- Major folder/pattern/architectural changes

**Removed**

- `patterns-react` package


mkothur:
Reviewing...

photodow:
> Deploy preview created for package "React (experimental)":
> https://ibmdotcom-react-experimental.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/1134/index.html
> 
> Built with commit: b49b3a1131fbe1b6b57682adf1ac3605585a8d32

Just a thought here...

We could reduce confusion on what URL to visit when it comes to experimental if we just turn all flags on for `ibmdotcom-react` and either bucket all experimentals into their own bucket or add something like (experimental) in the name.

Issue #993, opened 2020-01-10
Footer SSR breaks


sanjitbauli:
## Detailed description

```
TypeError: Cannot read property 'querySelectorAll' of undefined
    at altlangs (/Users/sanjitbauli/workspace/urx-nxt/node_modules/@carbon/ibmdotcom-utilities/lib/utilities/altlangs/altlangs.js:28:48)
    at sortList (/Users/sanjitbauli/workspace/urx-nxt/node_modules/@carbon/ibmdotcom-react/lib/components/LocaleModal/LocaleModal.js:157:54)
    at LocaleModal (/Users/sanjitbauli/workspace/urx-nxt/node_modules/@carbon/ibmdotcom-react/lib/components/LocaleModal/LocaleModal.js:216:17)
    at processChild (/Users/sanjitbauli/workspace/urx-nxt/node_modules/react-dom/cjs/react-dom-server.node.development.js:3090:14)
```
> Its a bug 
> issue?

> The problem is with the footer component
>  The expected behavior is, the footer should render without any errors.

> happening in all browsers

> What version of the IBM.com Library are you using?

    "@carbon/ibmdotcom-react": "^1.2.1",
    "@carbon/ibmdotcom-services": "^1.2.1",
    "carbon-components": "^10.9.0",
    "carbon-components-react": "^7.9.0",

> What offering/product do you work on? Any pressing ship or release dates we
URX

When i am using the component only for the Front end, then the dds proxy returns 404
<img width="1036" alt="Screenshot 2020-01-10 at 4 51 39 PM" src="https://user-images.githubusercontent.com/272230/72149572-7dc92a00-33c9-11ea-86f0-c0460d4aa9aa.png">

<img width="1162" alt="Screenshot 2020-01-10 at 4 52 15 PM" src="https://user-images.githubusercontent.com/272230/72149620-92a5bd80-33c9-11ea-8a92-ebe2b8cdaa57.png">

photodow:
Feel free to correct me, but this looks like it's a translation issue, and not an issue with any of our components?

In the example given it appears as if `inen` doesn't show up. `404`
https://1.www.s81c.com/common/js/dynamicnav/www/countrylist/jsononly/inen-utf8.json

But if you use `usen` it doesn't give a `404`, and works as expected?
https://1.www.s81c.com/common/js/dynamicnav/www/countrylist/jsononly/usen-utf8.json

Am I way off base with that?

[...]

PR #1181, opened 2020-01-24
fix(footer): render locale modal on client side when button is clicked

annawen1:
### Related Ticket(s)

Footer SSR breaks #993

### Description

When SSR, footer breaks when trying to retrieve `altlangs`. Add check to `altlangs` utility to detect if `root.document` exists. 

Removed hardcoded lang-region text for the locale button and use `LocaleAPI.getLangDisplay` service instead.

Make sure `<LocaleModal />` is rendered on click of the locale button instead.

### Changelog

**Changed**

- render `<LocaleModal />` based on `isOpen` state
- replace hardcoded lang-region text of localeButton with LocaleAPI.getLangDisplay` response
- add check to `altlangs` utility.


packages/react/examples/nextjs/package.json
```diff
@@ -8,8 +8,8 @@
     "start": "next start"
   },
   "dependencies": {
-    "@carbon/ibmdotcom-react": "1.2.0-rc.0",
-    "@carbon/ibmdotcom-styles": "1.2.0-rc.0",
+    "@carbon/ibmdotcom-react": "1.3.0-alpha.7",
+    "@carbon/ibmdotcom-styles": "1.3.0-alpha.7",
```
jeffchew: I think once we get this in `rc.1`, we should update the dependency to that so when we push full release the dependencies will update automatically by lerna.

jeffchew:
For some reason the PR didn't get notified, but the system check seemed to have passed so I'm just going to merge this: https://app.circleci.com/jobs/github/annawen1/ibm-dotcom-library/281

Issue #1176, opened 2020-01-24
Create the Content Item sub-pattern


photodow:
#### User Story
> As a `[user role below]`:
IBM.com Library Developer

> I need:
a reusable `ContentItem`

> so that I can:
use the same code across multiple patterns and components

#### Additional information
 - [Functional Specs](https://ibm.box.com/s/b64hvlc187xs653x07loh9zwh5nl8arf)
 - [Image Functional Specs](https://ibm.box.com/s/1bbrspqkipecv2o010l8czzese084cix)
 - [CTA Functional Specs](https://ibm.box.com/s/ceu99rjzd0e87sm51n9cv550yssockme)

#### Acceptance criteria
 - [ ] Added to sub-pattern bucket in storybook
 - [ ] Reusable content item sub-pattern create
 - [ ] heading, image, copy, cta

PR #1237, opened 2020-01-29
feat(content-item): content item sub-pattern

annawen1:
### Related Ticket(s)

Create the Content Item sub-pattern #1176

### Description

Content Item Sub-pattern

Screen Shot 2020-02-04 at 5 30 01 AM

General styling based off https://ibm.box.com/s/owlpnu8303gg6nse6yvtjyjow90e6tb7

**** has dependency on the CTA component which I will add once it is done ****

### Changelog

**New**

- `<ContentItem />` files

**Changed**

- pass `classname` prop to the `<Image />` component to allow for styling
- move `localeModalProps` knob object within story so it doesn't appear on other stories


packages/react/src/internal/FeatureFlags.js
```diff
@@ -142,6 +142,14 @@ export const DDS_FEATURED_LINK =
...
   process.env.DDS_CONTENT_GROUP === 'true' || DDS_FLAGS_ALL || false;
 
+/**
+ * This determines if the ContentItem component will be rendered or not
+ *
+ * @type {string | boolean}
+ */
+export const DDS_CONTENT_ITEM =
+  process.env.DDS_CONTENT_ITEM === 'true' || DDS_FLAGS_ALL || false;
+
```
jeffchew: @annawen1 we're going straight to prod qa with this one, let's exclude the feature flag.

annawen1: ok, removed!

packages/react/src/patterns/sub-patterns/ContentItem/ContentItem.js
```diff
@@ -0,0 +1,50 @@
...
+    data-autoid={`${stablePrefix}--content-item`}>
+    {heading && (
+      <h4 className={`${prefix}--content-item__heading`}>{heading}</h4>
+    )}
+    {image && <Image classname={`${prefix}--content-item__image`} {...image} />}
+    {copy && (
+      <span
+        className={`${prefix}--content-item__copy`}
+        dangerouslySetInnerHTML={{
+          __html: markdownToHtml(copy),
```
photodow: So right now this looks like it's on for everything. I think we italic is only allowed, and I think @jeffchew put in an issue
#1256  to allow new lines to create paragraphs. Probably should change the container here to a `div` as well since it'll have block elements here soon.

packages/react/src/patterns/sub-patterns/ContentItem/__stories__/ContentItem.stories.js
```diff
@@ -0,0 +1,36 @@
...
+        { src: 'https://picsum.photos/id/2/448/224', minWidth: 'md' },
+        { src: 'https://picsum.photos/id/2/352/176', minWidth: 'lg' },
+      ]),
+      alt: text('alt', 'content item image'),
+      defaultImage: text(
+        'default image:',
+        'https://picsum.photos/id/2/352/176'
+      ),
+    };
+    return <ContentItem heading={heading} copy={copy} image={image} />;
```
photodow: Could we put this within a grid just for the story to make it easier to visualize?

packages/styles/scss/patterns/sub-patterns/content-item/_content-item.scss
```diff
@@ -0,0 +1,40 @@
...
+
+@mixin content-item {
+  .#{$prefix}--content-item {
+    display: flex;
+    flex-direction: column;
+
+    &__heading {
+      margin-top: $carbon--spacing-07;
+      @include carbon--type-style('heading-02');
+    }
```
photodow: A couple general comments on styling. We need to make sure we add the properly themed tokens like `$text-01` to the content. I also think we need to add the `95%` width here, and max-width of `640px`. As of right now that's pretty consistent across the board it looks like. This all goes for copy as well.

packages/react/src/patterns/sub-patterns/ContentItem/__stories__/ContentItem.stories.js
photodow: A suggestion based on the specs you provided. This will also help design see the percentages and max-widths on the content.

```html
<div class="bx--grid">
  <div class="bx--row">
      ...
    </div>
  </div>
</div>
```

annawen1: ok! done

packages/styles/scss/patterns/sub-patterns/content-item/_content-item.scss
annawen1: done!

packages/react/src/patterns/sub-patterns/ContentItem/ContentItem.js
annawen1: ok! changed to `div`

annawen1:
@photodow @jeffchew thinking we can merge it in with the CTA in for now to unblock some folks, I can add the CTA in later once that has been done.

jeffchew:
> @photodow @jeffchew thinking we can merge it in without the CTA in for now to unblock some folks, I can add the CTA in later once that has been done.
> 
> CC @andysherman2121 @RaphaelAmadeu

Agreed, let's get this in!

Issue #1168, opened 2020-01-23
Create the CTA component


photodow:
#### User Story
> As a `[user role below]`:
IBM.com Library Developer

> I need to:
provide different cta options

> so that I can:
can easily allow the proper cta's for a given pattern

#### Additional information
 - [Functional Specs](https://ibm.box.com/s/ceu99rjzd0e87sm51n9cv550yssockme)
 - [LinkWithIcon](https://ibmdotcom-react-experimental.mybluemix.net/?path=/story/components-link-with-icon--default)
 - [ButtonGroup](https://ibmdotcom-react-experimental.mybluemix.net/?path=/story/patterns-sub-patterns-buttongroup--default)
- [Card](https://ibmdotcom-react-experimental.mybluemix.net/?path=/story/patterns-sub-patterns-card--link)
- [FeatureCard](https://ibmdotcom-react-experimental.mybluemix.net/?path=/story/patterns-blocks-feature-card--default)
- [Carbon Icons](https://www.carbondesignsystem.com/guidelines/icons/library)
 - Scope for this version does not include video and download type

#### Acceptance criteria
 - [ ] Types supported (jump, local, external)
 - [ ] Styles (text, card, feature)
 - [ ] FeatureCard allows icon to be changed
 - [ ] Single CTA component switches between multiple CTA styles and types.

PR #1268, opened 2020-01-30
feat(component): created cta component and changing the data model

mkothur:
### Related Ticket(s)

Related issue: #1168 

### Description

Added the CTA component and added the knobs to switch b/w different cta styles and types and also 

### Change log

**New**

Created CTA in `react components`
Updated the ContentGorup as per functional specs and added the CTA.

image
image


packages/react/src/components/Cta/Cta.js
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * @param {object} props props object
+ * @param {string} props.style style ( text | card | button | feature ).
+ * @param {object} props.type tyle ( jump | local | external ).
+ * @param {object} props.cta cta object which includes href,text, copy, card and image, button  properties.
+ * @returns {*} CTA Component
+ */
+const CTA = ({ style, type, ...cta }) =>
+  style === 'card' ? (
+    <CardLink {...cta} icon={iconSelector(type)} />
```
photodow: I'm wondering if we should lock down the props for this given style. I'm not sure we want images, and other things in here. Only maybe a heading, and icon. Thoughts?

packages/react/src/components/Cta/Cta.js
```diff
@@ -0,0 +1,83 @@
...
+ * @returns {*} cta type component
+ */
+const iconSelector = type =>
+  type === 'external' ? (
+    <Launch20 />
+  ) : type === 'jump' ? (
+    <ArrowDown20 />
+  ) : (
+    <ArrowRight20 />
+  );
```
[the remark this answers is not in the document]
mkothur: Yeah @photodow working on smooth scrolling for `jump` and new tab for `external`.

packages/react/src/components/Cta/README.md
```diff
@@ -0,0 +1,54 @@
+# Cta
```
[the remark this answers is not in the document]
mkothur: yeah.

packages/react/src/components/Cta/Cta.js
jeffchew: Yes, I would not pass all props in.

packages/react/src/components/Cta/__stories__/Cta.stories.js
```diff
@@ -0,0 +1,127 @@
...
+    let cta, type;
+    const copy = ['Lorem ipsum dolor sit amet', 'Consectetur adipisicing elit'];
+    const types = ['local', 'jump', 'external'];
+    const styles = ['text', 'card', 'feature', 'button'];
+    const style = select('style', styles, styles[0]);
+
+    const target = {
+      external: '_blank',
+      jump: '_self',
+    };
```
photodow: So I'm afraid there might be a little confusion here. We don't want people passing in this object. The `type` prop we're passing in should decide whether it's `_blank`, smooth scroll, or none.

packages/react/src/components/Cta/Cta.js
```diff
@@ -0,0 +1,82 @@
...
+    <CardLink {...cta} icon={iconSelector(type)} />
+  ) : style === 'button' ? (
+    <ButtonGroup buttons={renderButtons(cta)} />
+  ) : style === 'feature' ? (
+    <FeaturedLink
+      {...cta}
+      //icon={iconSelector(type)}
+    />
+  ) : (
+    <LinkWithIcon href={cta.href} target={cta.target}>
```
photodow: This may be like this because you're still working on it, but I just want to reiterate in case there is a misunderstanding that the `target` here shouldn't always be here, we only need to define if the `type==="jump"`, and we need to apply it to all of the other ctas as well.

mkothur: Yeah @photodow still working on it.

packages/react/src/components/Cta/__stories__/Cta.stories.js
mkothur: But we should mention somewhere to assign a target value based on type right?

photodow: Maybe i'm not understanding your question, but they won't assign the target value. We don't want them to assign the target. The component will assign it them. The adopter will pass in `type`, and then we'll use a conditional statement to decide what target value to provide, or if we need to provide one at all.

mkothur:
@photodow: added the smooth scroll for `jump` type.

packages/react/src/components/Cta/Cta.js
```diff
@@ -0,0 +1,83 @@
...
+ * @param {string} type cta type ( external | jump | local)
+ * @returns {*} cta type component
+ */
+const iconSelector = type =>
+  type === 'external' ? (
+    <Launch20 />
+  ) : type === 'jump' ? (
+    <ArrowDown20 />
+  ) : (
+    <ArrowRight20 />
```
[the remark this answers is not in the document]
RaphaelAmadeu: I guess that we should, but only for internal links, I think anchors and external links should always be identified.

packages/react/src/components/Cta/Cta.js
```diff
@@ -0,0 +1,139 @@
...
+        <CardLink
+          {...cta}
+          icon={iconSelector(type)}
+          onClick={e => {
+            handleOnClick(e, cta.id);
+          }}
+        />
+      );
+    }
+    return <CardLink {...cta} icon={iconSelector(type)} />;
```
photodow: Wondering if we could simplify a lot of this logic with a simple tertiary, and only retain a single component to return?

```jsx
<CardLink
  {...cta}
  icon={iconSelector(type)}
  onClick={type === 'jump' ? handleOnClick : null}
  }}
/>
```

packages/react/src/components/Cta/index.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * Copyright IBM Corp. 2016, 2018
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+export { default as CTA } from './Cta';
```
photodow: Again, we can probably keep this all capital, and it'll be ok.

`Cta --> CTA`

packages/react/src/components/Cta/Cta.js
photodow: @jeffchew good opportunity for those card decorator patterns we were talking about last week.

packages/react/src/components/Cta/index.js
mkothur: yeah @photodow when I make change the file name `Cta to CTA ` I am getting an error Module not found: Error: [CaseSensitivePathsPlugin] `/Users/mkothur@us.ibm.com/ibm-dotcom-library/packages/react/src/components/Cta/Cta.js` does not match the corresponding path on disk `CTA.js`.
 @ ./src/components/Cta/__stories__/Cta.stories.js 14:34-51,

packages/react/src/components/Cta/Cta.js
mkothur: Okay

packages/react/src/components/Cta/Cta.js
```diff
@@ -0,0 +1,139 @@
...
+  );
+};
+
+/**
+ * Handle OnClick
+ *
+ * @param {*} e event object
+ * @param {*} id element id
+ */
+const handleOnClick = (e, id) => {
```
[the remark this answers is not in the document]
mkothur: that make sense.

packages/react/src/components/Cta/Cta.js
```diff
@@ -0,0 +1,139 @@
...
+ * CTA component
+ *
+ * @param {object} props props object
+ * @param {string} props.style style ( text | card | button | feature ).
+ * @param {object} props.type tyle ( jump | local | external ).
+ * @param {object} props.cta cta object which includes href,text, copy, card and image, button  properties.
+ * @returns {*} CTA Component
+ */
+const CTA = ({ style, type, ...cta }) => {
+  if (style === 'card') {
```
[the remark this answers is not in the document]
mkothur: yeah think so.

packages/react/src/components/Cta/index.js
photodow: 🤔got it. So it's an eslint error?

mkothur: no its git issue, trying to fix this.

mkothur:
@photodow addressed the above changes except Cta --> CTA(for this I will remove and add again). Please do take a look the changes.

packages/react/src/components/Cta/CTA.js
```diff
@@ -0,0 +1,93 @@
...
+          onClick={type === 'jump' ? smoothScrolling : null}
+        />
+      );
+    case 'feature':
+      return <FeaturedLink {...cta} />;
+    default:
+      return (
+        <LinkWithIcon
+          href={cta.href}
+          target={type === 'external' ? '_blank' : null}
```
photodow: This should be applied to the other styles I believe. I wonder if it makes sense to pull this logic out into a separate function so we don't have to repeat all of this logic 3 times.

```javascript
<CardLink
  {...cta}
  icon={iconSelector(type)}
  target={external()}
  onClick={type === 'jump' ? smoothScrolling : null}
/>

function external (type) {
  return type === 'external' ? '_blank' : null;
}
```

I guess you could even do this with the `onClick` as well, that way the logic only has to be updated once. I keep going back and forth on this one. It's almost easier to just write the `onClick` out on its own because you have to pass in all the props and execute an additional two functions.


```javascript
<CardLink
  {...cta}
  icon={iconSelector(type)}
  target={external()}
  onClick={e => { jump(e, type) }}
/>

function jump (e, type) {
  return type === 'jump' ? SmoothScroll(e) : null;
}
```

packages/react/src/components/Cta/CTA.js
```diff
@@ -0,0 +1,93 @@
...
+  }
+};
+/**
+ * TEMPORARY sets icon based on link type
+ *
+ * @param {string} type cta type ( external | jump | local)
+ * @returns {*} cta type component
+ */
+const TEMP_iconSelector = type =>
+  type === 'external' ? Launch20 : type === 'jump' ? ArrowDown20 : ArrowRight20;
```
photodow: Can we get `ButtonGroup` updated so we don't need these two nearly identical functions?

packages/utilities/src/utilities/smoothScrolling/smoothScrolling.js
```diff
@@ -0,0 +1,14 @@
...
+ * @param {*} e event object
+ */
+const smoothScrolling = e => {
+  e.preventDefault();
+  const id = e.currentTarget.getAttribute('href');
+  document.querySelector(id).scrollIntoView({
+    behavior: 'smooth',
+    block: 'start',
+  });
+};
```
photodow: Are we going to update `TableOfContents.js`, or save that for another pull request?

packages/react/src/components/Cta/README.md
```diff
@@ -0,0 +1,54 @@
...
+> 💡 Only import font's once per usage
+
+```javascript
+import React from 'react';
+import ReactDOM from 'react-dom';
+import 'yourapplication.scss';
+
+function App() {
+  return <CTA style={style} type={type} {...cta} />;
+}
```
photodow: I don't think an adopter can just copy and paste this into their application and get it working. `cta` isn't a prop.

```javascript
function App() {
  return <CTA
    style='text'
    type='local'
    copy='IBM Homepage'
    href='www.ibm.com'
  />;
}
```

packages/react/src/components/Cta/README.md
```diff
@@ -0,0 +1,54 @@
...
+@include carbon--font-face-mono();
+@include carbon--font-face-sans();
+@import '@carbon/ibmdotcom-styles/scss/components/imagecomponent/imagecomponent';
+```
+
+> 💡 Only import font's once per usage
+
+```javascript
+import React from 'react';
+import ReactDOM from 'react-dom';
```
photodow: Also need to import the `CTA` component.

packages/react/src/components/Cta/CTA.js
```diff
@@ -0,0 +1,93 @@
...
+ * @param {object} cta object with buttons array
+ * @returns {*} object
+ */
+const renderButtons = ({ buttons }) =>
+  buttons.map(button => {
+    button.renderIcon = TEMP_iconSelector(button.type);
+    return button;
+  });
+CTA.propTypes = {
+  cta: PropTypes.object,
```
photodow: CTA isn't actually a prop. Not sure if there is a way to define what this actually is. I'd just leave it blank, and let the CTA components themselves handle the proptypes.

packages/react/src/components/Cta/README.md
```diff
@@ -0,0 +1,54 @@
...
+> 💡 Don't forget to import the image styles from
+> [@carbon/ibmdotcom-styles](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/styles).
+
+## Props
+
+| Name    | Required | Data Type | Default Value | Description                                    |
+| ------- | -------- | --------- | ------------- | ---------------------------------------------- |
+| `style` | NO       | String    | text          | Deafult style is text (LinkWithIcon component) |
+| `cta`   | NO       | Object    | null          | Contains type, href,copy title, heading, image |
+| `type`  | NO       | String    | local         | Contains type, href,copy title, heading, image |
```
photodow: CTA isn't actually a prop, checkout the functional specs for these definitions. It adopts whatever the props are for the given CTA.

image

packages/utilities/src/utilities/smoothScrolling/smoothScrolling.js
mkothur: save for another full request.

packages/react/src/components/Cta/CTA.js
mkothur: Sure.

packages/react/src/components/Cta/README.md
mkothur: so, we need add these {title, text, href} in  # props section of README.md  file?

photodow: You might add a comment and or link to the other CTA components saying use their props here. I can try to add a more specific comment within the functional specs.

mkothur: Okay.

mkothur:
@photodow updated the PR.

packages/react/src/components/Cta/__stories__/Cta.stories.js
```diff
@@ -0,0 +1,118 @@
...
+        type = select('type', types, types[0]);
+        cta = {
+          heading: copy[0],
+          card: {
+            type: type,
+            href: urlBy[type],
+            title: copy[1],
+            image: {
+              defaultImage: 'https://picsum.photos/id/672/672',
+              alt: '...',
```
photodow: Might be good to add an actual alt text here. `Placeholder image`.

packages/react/src/components/Cta/CTA.js
```diff
@@ -0,0 +1,115 @@
...
+const CTA = ({ style, type, ...cta }) => {
+  switch (style) {
+    case 'card':
+      return (
+        <CardLink
+          {...cta}
+          icon={iconSelector(type)}
+          target={external(type)}
+          onClick={e => jump(e, type)}
+        />
```
photodow: For this one we need to lock it down until we can figure out card decorators.

I believe all we need is the heading, and we should probably give it a `className`.

```jsx
<CardLink
  copy={cta.copy}
  className={`${prefix}--card-cta`}
  icon={iconSelector(type)}
  target={external(type)}
  onClick={e => jump(e, type)}
/>
```

Then we might want to add some `SCSS` (it could even be apart of the card sub-pattern possibly for now where we define `heading-02` for the copy.

I think you can use copy or heading here, but I'm recommending `copy` since this is essentially just like a large text link.

packages/react/src/components/Cta/CTA.js
```diff
@@ -0,0 +1,115 @@
...
+const external = type => (type === 'external' ? '_blank' : null);
+
+/**
+ * TEMPORARY sets icon based on link type
+ *
+ * @param {string} type cta type ( external | jump | local)
+ * @returns {*} type of icon component
+ */
+const TEMP_iconSelector = type =>
+  type === 'external' ? Launch20 : type === 'jump' ? ArrowDown20 : ArrowRight20;
```
photodow: I'll add this comment again, I think we should do the work to remove this temporary function.

packages/react/src/components/Cta/README.md
```diff
@@ -0,0 +1,65 @@
...
+| `type`  | YES      | String    | local         | Describes which type `local`,`jump`,`external` icon.                 |
+
+## Style type's
+
+| Style Type | Component Name | Description                                                               |
+| ---------- | -------------- | ------------------------------------------------------------------------- |
+| `text`     | LinkWithIcon   | See the documentation of LinkWithIcon component and use thier props here. |
+| `button`   | ButtonGroup    | See the documentation of ButtonGroup component and use thier props here.  |
+| `card`     | Card           | See the documentation of Card component and use thier props here.         |
+| `feature`  | FeatureCard    | See the documentaion of FeatureCard component and user their props here.  |
```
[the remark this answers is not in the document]
mkothur: You want me to reference the github path of the specific component?

packages/react/src/components/Cta/README.md
```diff
@@ -0,0 +1,65 @@
...
+> [@carbon/ibmdotcom-styles](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/styles).
+
+## Props
+
+| Name    | Required | Data Type | Default Value | Description                                                          |
+| ------- | -------- | --------- | ------------- | -------------------------------------------------------------------- |
+| `style` | YES      | String    | text          | Describes style type, for more information See `Style type's` below. |
+| `type`  | YES      | String    | local         | Describes which type `local`,`jump`,`external` icon.                 |
+
+## Style type's
```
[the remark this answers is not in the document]
mkothur: Okay

packages/react/src/components/Cta/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+> 💡 Don't forget to import the image styles from
+> [@carbon/ibmdotcom-styles](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/styles).
+
+## Props
+
+| Name    | Required | Data Type | Default Value | Description                                                          |
+| ------- | -------- | --------- | ------------- | -------------------------------------------------------------------- |
+| `style` | YES      | String    | text          | Describes style type, for more information See `Style type's` below. |
+| `type`  | YES      | String    | local         | Describes which type `local`,`jump`,`external` icon.                 |
```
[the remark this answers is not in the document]
mkothur: Ok will do as like style types.

packages/react/src/components/Cta/__stories__/Cta.stories.js
mkothur: done

packages/react/src/components/Cta/CTA.js
mkothur: here I have a doubt will sync up with you on slack.

packages/react/src/components/Cta/CTA.js
mkothur: here also I have a doubt will sync up with you on slack.

photodow: Put in a pull request to make the change to the CardLink here: #1327 

Once that gets merged you should be able to replace `iconsSelector` with this method, and for `LinkWithIcon` you can use a technique demoed in this [codesandbox](https://codesandbox.io/s/clever-williamson-dw9us)

packages/react/src/components/Cta/CTA.js
```diff
@@ -0,0 +1,122 @@
...
+import { smoothScroll } from '@carbon/ibmdotcom-utilities';
+
+const { prefix } = settings;
+
+/**
+ * CTA component
+ *
+ * @param {object} props props object
+ * @param {string} props.style style ( text | card | button | feature ).
+ * @param {object} props.type tyle ( jump | local | external ).
```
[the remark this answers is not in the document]
mkothur: yeah it is, while reverting the contentGroup changes it messed sorry. will update.

packages/react/src/components/Cta/CTA.js
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * sets button
+ *
+ * @param {object} cta object with buttons array
+ * @returns {*} object
+ */
+const renderButtons = ({ buttons }) =>
+  buttons.map(button => {
+    button.renderIcon = TEMP_iconSelector(button.type);
+    //buttons.renderIcon = button.type === 'external' ? Launch20 : button.type === 'jump' ? ArrowDown20 : ArrowRight20;
```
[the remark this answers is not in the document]
mkothur: sorry, forgot to remove. will do

packages/react/src/components/Cta/CTA.js
```diff
@@ -0,0 +1,122 @@
...
+
+const { prefix } = settings;
+
+/**
+ * CTA component
+ *
+ * @param {object} props props object
+ * @param {string} props.style style ( text | card | button | feature ).
+ * @param {object} props.type tyle ( jump | local | external ).
+ * @param {object} props.cta cta object which includes href,text, copy, card and image, button  properties.
```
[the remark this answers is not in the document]
mkothur: removing this.

packages/react/src/components/Cta/CTA.js
```diff
@@ -0,0 +1,122 @@
...
+    case 'button':
+      return (
+        <ButtonGroup
+          buttons={renderButtons(cta)}
+          target={external(type)}
+          onClick={e => jump(e, type)}
+        />
+      );
+    case 'feature':
+      return <FeaturedLink {...cta} />;
```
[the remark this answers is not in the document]
mkothur: will add.

packages/react/src/components/Cta/CTA.js
```diff
@@ -0,0 +1,122 @@
...
+import React from 'react';
+import { settings } from 'carbon-components';
+import { smoothScroll } from '@carbon/ibmdotcom-utilities';
+
+const { prefix } = settings;
+
+/**
+ * CTA component
+ *
+ * @param {object} props props object
```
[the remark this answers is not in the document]
mkothur: yeah @photodow if remove this I am getting an error
image
And other components are being used.the above line.

mkothur:
@photodow: Updated the PR with above changes except this
image

packages/react/src/components/Cta/CTA.js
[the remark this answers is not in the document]
photodow: Oh, I see it now. My bad

packages/react/src/components/Cta/Cta.js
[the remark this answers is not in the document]
photodow: See #1327

packages/react/src/components/Cta/CTA.js
photodow: See #1327

mkothur:
@photodow : In codeSandbox, you have `others.js` and importing that in App and returning the other file value in another function of App file and rendering. 
In this scenario LinkWithIcon is other and App would be Stroy right?

packages/react/src/components/Cta/CTA.js
photodow: > @photodow : In codeSandbox, you have others.js and importing that in App and returning the other file value in another function of App file and rendering.
In this scenario LinkWithIcon is other and App would be Stroy right?

No, `App` would be like the default switch statement. `Other` would be more like the given icon. Look at the below to see what I'm talking about.

Essentially you need to set the `iconSelector(type)` return to a variable with a capital letter before you can render it as a component.

```javascript
  default:
    const Icon = iconSelector(type);
    return (
      <LinkWithIcon
        href={cta.href}
        target={external(type)}
        onClick={e => jump(e, type)}>
        {cta.copy}
        <Icon />
      </LinkWithIcon>
    );
  }
};

const iconSelector = type =>
  type === 'external' ? Launch20 : type === 'jump' ? ArrowDown20 : ArrowRight20;
```

mkothur:
@photodow: Never mind I got.

mkothur:
And @photodow can I include this change into this PR or want to wait me till this #1327 get merged.

packages/react/src/components/Cta/index.js
jeffchew: this is resolved, it was an import reference issue when I walked through this with @mkothur

packages/styles/scss/patterns/sub-patterns/card-link/index.scss
```diff
@@ -46,6 +46,11 @@
         fill: currentColor;
       }
     }
+    &--CTA {
+      .#{$prefix}--card-link__title {
+        @include carbon--type-style('expressive-heading-02');
+      }
+    }
```
jeffchew: folder name should be `card`, it's still called `card-link`

packages/utilities/src/utilities/smoothScroll/smoothScroll.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * Handle OnClick
+ *
+ * @param {*} e event object
+ */
```
jeffchew: The documentation on this seems very light. The code itself seems to be looking for the event object's target href then doing something with it. I think there needs to be more instruction on what this common utility is doing and what you need to set it to. Include an `@example` too.

mkothur: Okay.

packages/react/src/components/CTA/CTA.js
```diff
@@ -0,0 +1,113 @@
...
+ * Copyright IBM Corp. 2016, 2018
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+import { ArrowDown20, ArrowRight20, Launch20 } from '@carbon/icons-react';
+import { ButtonGroup } from '../../patterns/sub-patterns/ButtonGroup';
+import { CardLink } from '../../patterns/sub-patterns/CardLink';
+import { FeaturedLink } from '../../patterns/blocks/FeaturedLink';
+import { LinkWithIcon } from '../LinkWithIcon';
```
jeffchew: Although might be ok for now, but one thing we might want to think about is dynamic imports around these. If someone imports `CTA`, their build artifact will include all of these components, even if they may be using just one, for example. I don't think it's going to be a huge performance difference, but something to think about.

packages/react/src/components/CTA/CTA.js
```diff
@@ -0,0 +1,113 @@
...
+ * sets button
+ *
+ * @param {object} cta object with buttons array
+ * @returns {*} object
+ */
+const renderButtons = ({ buttons }) =>
+  buttons.map(button => {
+    button.renderIcon = iconSelector(button.type);
+    return button;
+  });
```
jeffchew: These all appear to be private methods, so we should name and document them as such (include underscores in front of the const names, and add `@private` in the jsdocs.

And while it's a preference thing, I've been adding private constants and methods at the top of the file rather than below, but might be something we can discuss as a team what our standard should be.

mkothur: yeah sure, will update.

mkothur:
@jeffchew updated the above comments.

packages/utilities/src/utilities/smoothScroll/smoothScroll.js
```diff
@@ -0,0 +1,29 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * Utility handles smoothScroll to the target element after OnClick
+ *
+ * @example
+ * import {smoothScroll} from '@carbon/ibmdotcom-utilities';
+ *
+ * smoothScroll(e);
```
jeffchew: I think this is still a bit fuzzy. this is specifically for smooth scrolling to an anchor link on the page, so something like:

```
<a href="#anchorlinkname" onclick="smoothScroll(e)">lorem ipsum</a>
```

packages/react/src/components/CTA/CTA.js
```diff
@@ -35,7 +35,7 @@ const CTA = ({ style, type, ...cta }) => {
           href={cta.href}
           icon={_iconSelector(type)}
           target={_external(type)}
-          handleClick={e => _jump(e, type)}
+          onMouseDown={e => _jump(e, type)}
```
jeffchew: @photodow @mkothur is switching to mousedown going to have any accessibility impact? seems like it would...

photodow: Yeah, we were just talking about that.

photodow: he's going to revert it to handleClick again, and we'll work with Carbon on a fix.

packages/utilities/src/utilities/smoothScroll/smoothScroll.js
```diff
@@ -6,12 +6,12 @@
  */
 
 /**
- * Utility handles smoothScroll to the target element after OnClick
+ * Utility handles smoothScroll to the target element after onMouseDown
```
jeffchew: Wonder if we should explicitly say this is for anchor links?

packages/react/src/components/CTA/CTA.js
jeffchew: ok, yeah let's get this in as-is with click events, and track separately.

packages/utilities/src/utilities/smoothScroll/smoothScroll.js
mkothur: updated.

Issue #1064, opened 2020-01-16
Pattern: Develop Table of Contents pattern - Updates from W/S - Sprint 20-02


RobertaJHahn:
#### User Story

> As a `[user role below]`:
IBM.com Library developer

> I need to:
work closely with the Business Analyst and designer to begin developing the (pattern name) pattern

> so that I can:
become involved earlier in the design process and functional documentation to compress the schedule and develop a higher quality pattern

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- Functional specs issue (#1063  )
 - [Functional Specs](https://ibm.ent.box.com/file/599375190897)
 - Design QA issue (#1203)
 - Prod QA issues (#1142)

#### Acceptance criteria
 - [x] Update the TOC with the option to dynamically generate the links based on the page content
 - [x] Remove the feature flag for this component
 - [x] Additional knobs or story necessary in Storybook to switch between dynamic and manual generation of the links
 - [x] Update the README for the react pattern and corresponding styles
- [ ] A comment is posted in the Design QA issue, tagging Wonil and Roberta, when development is finished
 - [ ] A comment is posted in the Prod QA issue, tagging Praveen and Chetan, when development is finished

larahanlon2:
FYI Table of Contents design specs are now available on Abstract: https://share.goabstract.com/dbaeb532-b588-4138-8a07-3d04f01b4f34?

PR #1269, opened 2020-01-30
Pattern: Update Table of Contents pattern 

carthick:
### Related Ticket(s)

Services - #1064 
### Description

Update the TOC with the option to dynamically generate the links based on the page content


packages/react/src/patterns/sub-patterns/TableOfContents/TableOfContents.js
```diff
@@ -33,11 +31,49 @@ const { prefix } = settings;
...
-
+  const [selectedTitle, setSelectedTitle] = useState(menuItems[0].name);
+  const [autoMenuItems, setmenuItems] = useState([]);
   useEffect(() => {
+    /**
+     * Set selected id & title
+     *
+     */
+    if (menuItems) {
+      var setSelectedItem = () => {
```
jeffchew: @carthick we don't use `var` in ES6.

packages/react/src/patterns/sub-patterns/TableOfContents/TableOfContents.js
```diff
@@ -33,11 +31,49 @@ const { prefix } = settings;
...
-  const [selectedTitle, setSelectedTitle] = useState(menuItems[0].title);
-
+  const [selectedTitle, setSelectedTitle] = useState(menuItems[0].name);
+  const [autoMenuItems, setmenuItems] = useState([]);
   useEffect(() => {
+    /**
+     * Set selected id & title
+     *
+     */
+    if (menuItems) {
```
jeffchew: The order of this seems all off. If anything, if `menuItems` doesn't exist, then you should be running `findMenuItems`, correct? It seems like `setSelectedItem` should be run regardless of whether `menuItems` exists or not. The only difference is that if it already is passed in as props, it won't run `findMenuItems`, otherwise it will.

packages/react/src/patterns/sub-patterns/TableOfContents/TableOfContents.js
```diff
@@ -63,23 +99,6 @@ const TableOfContents = ({ menuItems, children, menuLabel, theme }) => {
...
-    const id = elems[0] || menuItems[0].id;
-    const filteredItems = menuItems.filter(menu => {
-      if (id !== 'undefined') {
-        return menu.id === id;
-      }
-    });
-    const title = filteredItems[0].title;
-    setSelectedId(id);
-    setSelectedTitle(title);
-  };
```
jeffchew: This was probably fine where it was before.

packages/react/src/patterns/sub-patterns/TableOfContents/TableOfContents.js
```diff
@@ -33,11 +31,49 @@ const { prefix } = settings;
...
+        setSelectedTitle(title);
+      };
+    }
+    /**
+     * loops into the array of elements and returns the values
+     *
+     * @private
+     * @returns {Array} returns elemenrt name and data title
+     */
+    const _find_menu_items = () => {
```
jeffchew: The name of this doesn't follow convention. Should be something like `findMenuItems` instead.

jeffchew: Also, this can remain as a private method, but can be outside of the Class (doesn't need to be inside of it).

packages/react/src/patterns/sub-patterns/TableOfContents/TableOfContents.js
```diff
@@ -152,8 +170,8 @@ const TableOfContents = ({ menuItems, children, menuLabel, theme }) => {
           className={`${prefix}--tableofcontents__sidebar`}
           data-sticky="true">
           <div className={`${prefix}--tableofcontents__mobile-top`}></div>
-          <TOCDesktop {...props} />
-          <TOCMobile {...props} />
+          <TOCDesktop autoMenuItems={autoMenuItems} {...props} />
+          <TOCMobile autoMenuItems={autoMenuItems} {...props} />
```
jeffchew: What is `autoMenuItems`? Is this even necessary? They don't appear to be props for TOCDecktop or TOCMobile unless I missed something?

packages/react/src/patterns/sub-patterns/TableOfContents/TableOfContents.js
```diff
@@ -58,7 +58,7 @@ const TableOfContents = ({ menuItems, children, menuLabel, theme }) => {
      * @private
      * @returns {Array} returns elemenrt name and data title
      */
-    const _find_menu_items = () => {
+    const findMenuItems = () => {
```
jeffchew: @carthick I also wouldn't define a method inside of a `useEffect`. It should be outside of it, but can be called from it.

packages/react/src/patterns/sub-patterns/TableOfContents/TableOfContents.js
```diff
@@ -33,11 +31,49 @@ const { prefix } = settings;
  */
 const TableOfContents = ({ menuItems, children, menuLabel, theme }) => {
   const [selectedId, setSelectedId] = useState(menuItems[0].id);
-  const [selectedTitle, setSelectedTitle] = useState(menuItems[0].title);
-
+  const [selectedTitle, setSelectedTitle] = useState(menuItems[0].name);
+  const [setmenuItems] = useState([]);
```
jeffchew: @carthick this line doesn't look necessary anymore.

packages/react/src/patterns/sub-patterns/TableOfContents/TableOfContents.js
carthick: @jeffchew If  I am trying to move this to the same place .. I am getting the below error
Screen Shot 2020-02-03 at 2 22 11 PM

carthick: Never mind .... updated that and move back

packages/react/src/patterns/sub-patterns/TableOfContents/TableOfContents.js
```diff
@@ -33,34 +31,29 @@ const { prefix } = settings;
  */
 const TableOfContents = ({ menuItems, children, menuLabel, theme }) => {
   const [selectedId, setSelectedId] = useState(menuItems[0].id);
-  const [selectedTitle, setSelectedTitle] = useState(menuItems[0].title);
-
+  const [selectedTitle, setSelectedTitle] = useState(menuItems[0].name);
   useEffect(() => {
+    if (!menuItems || menuItems === undefined || menuItems === null) {
+      findMenuItems();
```
jeffchew: @carthick one thing I'm not a fan about here is that you're mutating the prop `menuItems` coming into the component. If anything, you might want to do something like:
```
const useMenuItems = menuItems ? [...menuItems] : _findMenuItems();
```

In this case, you check if `menuItems` is defined or not, and if not, it calls the `_findMenuItems` method (which I still think should be taken completely outside of the class), and use the return of that to set `useMenuItems` within the class. If  `menuItems` is defined, you're using the spread operator to do a safe clone of the array instead of manipulating `menuItems` within the class. After that, I'd swap out usage of `menuItems` in here with `useMenuItems`.

carthick: @jeffchew as pre my understanding i have updated the .. Please let me know if i need to update anything else....

packages/react/src/patterns/sub-patterns/TableOfContents/TableOfContents.js
```diff
@@ -33,34 +31,31 @@ const { prefix } = settings;
...
-
+  const [selectedTitle, setSelectedTitle] = useState(menuItems[0].name);
+  const useMenuItems = menuItems ? [...menuItems] : findMenuItems();
   useEffect(() => {
     scrollStop(setSelectedItem);
   });
 
+  if (!useMenuItems) {
+    findMenuItems();
+  }
```
jeffchew: @carthick move this to inside the `useEffect`. Also, as I mentioned, move the method `findMenuItems` itself to outside of `const TableOfContents`, and rename to `_findMenuItems`.

In general, anything that needs to be called on the client side needs to be inside the `useEffect`.

packages/react/src/patterns/sub-patterns/TableOfContents/TableOfContents.js
```diff
@@ -33,34 +31,31 @@ const { prefix } = settings;
  */
 const TableOfContents = ({ menuItems, children, menuLabel, theme }) => {
   const [selectedId, setSelectedId] = useState(menuItems[0].id);
-  const [selectedTitle, setSelectedTitle] = useState(menuItems[0].title);
-
+  const [selectedTitle, setSelectedTitle] = useState(menuItems[0].name);
+  const useMenuItems = menuItems ? [...menuItems] : findMenuItems();
```
jeffchew: You may want to set `useMenuItems` as a let, but define it inside the `useEffect`. If you try to load `TableOfContents` on the server side, the method `_findMenuItems` would fail.

packages/react/src/patterns/sub-patterns/TableOfContents/TableOfContents.js
```diff
@@ -33,34 +31,31 @@ const { prefix } = settings;
...
-        bounding.right <=
-          (root.innerWidth || document.documentElement.clientWidth)
-      ) {
-        elesInView.push(element.getAttribute('name'));
-      }
+      useMenuItems.push({
+        id: element.getAttribute('name'),
+        title: element.getAttribute('data-title'),
+      });
     });
```
jeffchew: Move this to outside of `TableOfContents`, above as a private method. You also should not reference `useMenuItems` in here, it should be a standalone function that grabs the elements, and returns back the array. useMenuItems will be defined by either the passed in prop `menuItems`, or will be set as the returned value from the `findMenuItems` method.

packages/react/src/patterns/sub-patterns/TableOfContents/TableOfContents.js
```diff
@@ -72,14 +67,38 @@ const TableOfContents = ({ menuItems, children, menuLabel, theme }) => {
     const id = elems[0] || menuItems[0].id;
     const filteredItems = menuItems.filter(menu => {
       if (id !== 'undefined') {
-        return menu.id === id;
+        return menu.id == id;
```
jeffchew: don't use double equal, should use triple equal for comparisons.

packages/react/src/patterns/sub-patterns/TableOfContents/TableofContentsDesktop.js
```diff
@@ -23,7 +23,7 @@ const { prefix } = settings;
  * @param {*} props.updateState function to update parent state.
  * @returns {*} JSX Object
  */
-const TOCDesktopMenu = ({ menuItems, selectedId, updateState }) => {
+const TOCDesktop = ({ menuItems, selectedId, updateState }) => {
```
jeffchew: Why was this filename renamed to `TableOfContentsDesktop`, but `TOCMobile` remained the same?

carthick: Build is failing because of the prettier if i change the file name to TOCDesktop ... Please suggest a better file name so that i can go and update ......
Screen Shot 2020-01-31 at 1 14 51 PM

packages/react/src/patterns/sub-patterns/TableOfContents/TableOfContents.js
```diff
@@ -33,12 +31,33 @@ const { prefix } = settings;
...
 const TableOfContents = ({ menuItems, children, menuLabel, theme }) => {
   const [selectedId, setSelectedId] = useState(menuItems[0].id);
-  const [selectedTitle, setSelectedTitle] = useState(menuItems[0].title);
+  const [selectedTitle, setSelectedTitle] = useState(menuItems[0].name);
 
   useEffect(() => {
+    let useMenuItems = menuItems ? [...menuItems] : _findMenuItems();
+    if (!useMenuItems) {
+      _findMenuItems();
+    }
```
jeffchew: The oneliner already covers what's happening, so this cane be removed.

packages/react/src/patterns/sub-patterns/TableOfContents/TableOfContents.js
```diff
@@ -165,9 +166,27 @@ const TableOfContents = ({ menuItems, children, menuLabel, theme }) => {
...
+  const menuItems = [];
+  eles.forEach(element => {
+    menuItems.push({
+      id: element.getAttribute('name'),
+      title: element.getAttribute('data-title'),
+    });
+  });
+  return menuItems;
+};
+
```
jeffchew: Can you move this to below the imports at the top of the file?

packages/react/src/patterns/sub-patterns/TableOfContents/TableofContentsDesktop.js
jeffchew: Have you found all references for TOCDesktop or TableOfContentsDesktop? There might be an import that's out of sync.

jeffchew: Also, the name of the file should be `TOCDesktop.js` if that is the file that `TableOfContents.js` is pointing to.

larahanlon2:
FYI Table of Contents design specs are now available on Abstract: https://share.goabstract.com/dbaeb532-b588-4138-8a07-3d04f01b4f34?

carthick:
> FYI Table of Contents design specs are now available on Abstract: https://share.goabstract.com/dbaeb532-b588-4138-8a07-3d04f01b4f34?

@larahanlon2 
Screen Shot 2020-02-04 at 3 23 42 PM I had sent a request to access

packages/react/src/patterns/sub-patterns/TableOfContents/__stories__/TableOfContents.stories.js
```diff
@@ -1,23 +1,33 @@
...
-import { select, text, withKnobs } from '@storybook/addon-knobs';
+import { object, select, text, withKnobs } from '@storybook/addon-knobs';
+import dataContent from './data/dataContent';
 import { DDS_TOC } from '../../../../internal/FeatureFlags';
 import React from 'react';
 import readme from '../README.md';
 import { storiesOf } from '@storybook/react';
 import TableOfContents from '../TableOfContents';
 
 if (DDS_TOC) {
```
jeffchew: @carthick almost missed this. Should remove the feature flag here.

jeffchew: Also, need to remove `DDS_TOC` overall (in `FeatureFlags.js`, etc)

carthick: @jeffchew removed and updated Feature Flag... Please let me know if anything needs to be changed

Issue #1283, opened 2020-01-31
Pattern:  Card - Enable Markdown in Copy


photodow:
## Detailed description
Right now the card doesn't use markdown within the main copy of the card. We should enable this only for italics at the moment. `*text*` `_text_`

#### Additional information
- [Functional Specs](https://ibm.box.com/s/y1h8lgvn8zjdzr9r0owi7gvy5mjz7lcd)

#### Acceptance criteria

- [ ] Italic markdown compiles to HTML
- [ ] Italics enabled in card main copy

PR #1307, opened 2020-02-04
fix(card): add markdown utility to card copy

annawen1:
### Related Ticket(s)

Pattern: Card - Enable Markdown in Copy #1283

### Description

Use markdown for the main copy of the card. Enabled only for italics.

Screen Shot 2020-02-04 at 2 50 37 PM

### Changelog

**Changed**

- import markdown utility and enable italics and allowHTML


packages/react/src/patterns/sub-patterns/CardLink/CardLink.js
```diff
@@ -58,7 +61,11 @@ const CardLink = ({
  */
 function optionalContent(content) {
   return !content ? null : (
-    <p className={`${prefix}--card-link__content`}>{content}</p>
+    <div
+      className={`${prefix}--card-link__content`}
+      dangerouslySetInnerHTML={{
+        __html: markdownToHtml(content, { italic: true, allowHtml: true }),
```
jeffchew: the markdown utility is opt out, so you would need to do `bold: false`, and `italic: true` isn't needed here...

annawen1: ah right, done!

packages/react/src/patterns/sub-patterns/CardLink/CardLink.js
```diff
@@ -58,7 +61,11 @@ const CardLink = ({
  */
 function optionalContent(content) {
   return !content ? null : (
-    <p className={`${prefix}--card-link__content`}>{content}</p>
+    <div
+      className={`${prefix}--card-link__content`}
+      dangerouslySetInnerHTML={{
+        __html: markdownToHtml(content, { bold: false, allowHtml: true }),
```
photodow: What does the `allowHtml` do?

photodow:
Should we try to remove other references where markdown is already being passed into `CardLink`?


https://github.com/carbon-design-system/ibm-dotcom-library/blob/392e82b8408f909b45fa7471916a45d380e88981/packages/react/src/patterns/blocks/CardArray/CardArray.js#L97-L100

packages/react/src/patterns/sub-patterns/CardLink/CardLink.js
jeffchew: From the specs:
```Determines if HTML markup is permitted. If disabled, HTML tags will be stripped out.```
By default, this is false. I think we should keep it off, honestly.

Comment (annawen1) [reply packages/react/src/patterns/sub-patterns/CardLink/CardLink.js]: oh ok I thought we wanted it though,

`Italic markdown compiles to HTML` <- acceptance criteria from the ticket

Comment (jeffchew) [reply packages/react/src/patterns/sub-patterns/CardLink/CardLink.js]: there is another story to update the markdown utility to handle paragraphs, but otherwise we are only converting italics and nothing else: https://github.com/carbon-design-system/ibm-dotcom-library/issues/1256

Comment (annawen1) [reply packages/react/src/patterns/sub-patterns/CardLink/CardLink.js]: gotcha ok, removed `allowHtml`

**Comment by annawen1** [comment]
> Should we try to remove other references where markdown is already being passed into `CardLink`?
> 
> https://github.com/carbon-design-system/ibm-dotcom-library/blob/392e82b8408f909b45fa7471916a45d380e88981/packages/react/src/patterns/blocks/CardArray/CardArray.js#L97-L100

good point, I've removed the other instances where patterns using `<CardLink />` are using the markdown utility

File: packages/react/src/patterns/blocks/CardArray/CardArray.js [hunk 3]
```diff
@@ -0,0 +1,114 @@
+/**
```
jeffchew: @annawen1 was this here before? I think @RaphaelAmadeu already refactored this...

annawen1: uhhhh not sure, there was a bunch of updating from master that I pulled in a few times

annawen1: ok removed

Issue #1281, opened 2020-01-31
Pattern:  Card - Image interaction


photodow:
## Detailed description

We need the image to show interaction just as the rest of the card does. As the user hovers over any part of the card the image should darken slightly by 20%.

#### Additional information

- [Functional Specs](https://ibm.box.com/s/y1h8lgvn8zjdzr9r0owi7gvy5mjz7lcd)

#### Acceptance criteria

- [ ] card link image gets 20% darker when any part of the card is hovered on

PR #1308, opened 2020-02-04
fix(card): darken image for hover on card

annawen1:
### Related Ticket(s)

Pattern: Card - Image interaction #1281

### Description

Darken image when user hovers over card

Screen Shot 2020-02-04 at 3 20 42 PM

Feb-05-2020 16-18-16

### Changelog

**Changed**

- set image brightness to 80% when card is hovered


packages/styles/scss/patterns/sub-patterns/card-link/index.scss
```diff
@@ -54,6 +54,10 @@
 
     &:hover {
       text-decoration: none;
+
+      .#{$prefix}--image__img {
+        filter: brightness(80%);
```
photodow: Looks great, only needs the correct transition from the [@carbon/motion](https://www.npmjs.com/package/@carbon/motion) I believe.

mkothur:
LGTM!

packages/styles/scss/patterns/sub-patterns/card-link/index.scss
annawen1: ah was wondering about why there was a small bit of lag on the brightness, will look into how they're using @carbon/motion

annawen1: ok I added the same transition I see in the carbon repo

packages/styles/scss/patterns/sub-patterns/card-link/index.scss
```diff
@@ -17,6 +18,7 @@
     text-decoration: none;
     display: flex;
     flex-direction: column;
+    transition: $duration--moderate-01 motion(standard, productive);
```
photodow: Sorry Anna, but I think `bx--card-link` already has a transition — infact it has two of them thanks to `bx--link` — but get the transition to work on the image I think you need to wrap this with the image selector similar to how you did it below.

```scss
.#{$prefix}--card-link {
  .#{$prefix}--image__img {
    transition: $duration--moderate-01 motion(standard, productive);
  }
}
```

> Does the image component not let us pass in a custom `className` so we could use `bx--card-link__img`? It would help reduce the specificity a little.

annawen1: Oh the image component does let us pass a classname, let me do that and wrap the transition like you mentioned above

packages/styles/scss/patterns/sub-patterns/card-link/README.md
```diff
@@ -21,4 +21,5 @@ Import the package css into the top of your main CSS file.
 | `.bx--card-link__title`   | Title element within the CardLink component |
 | `.bx--card-link__content` | Paragraph of text styling for CardLink      |
 | `.bx--card-link__footer`  | Footer aligned bottom with icon             |
-| `.bx--card-link__image`   | Footer aligned bottom with icon             |
+| `.bx--card-link__img`     | Footer aligned bottom with icon             |
```
photodow: Maybe we could give this an accurate description before merging?

annawen1: oh whoops forgot to edit the desc, done!

Issue #1164, opened 2020-01-23
Pictogram Item: Updates to the Pictogram Item per Functional Specifications


jeffchew:
#### User Story

> As a `[user role below]`:
IBM.com Library user

> I need to:
have a PIctogramItem component available

> so that I can:
use this sub-pattern in pattern blocks/sections

#### Additional information

 - Functional specs: https://ibm.ent.box.com/file/604131694379
- Component currently is nested under the `PictogramArray` (dev story to update to `ContentGroupPictograms`: https://github.com/carbon-design-system/ibm-dotcom-library/issues/1165)

#### Acceptance criteria
 - [x] `PictogramItem` pattern added under `sub-patterns` in Storybook
 - [x] `PictogramItem` moved into the `sub-patterns` folder in the React package
 - [x] Minimum 80% unit test coverage

RobertaJHahn:
@RaphaelAmadeu Moving this issue to sprint 20-03 so that I can close out sprint 20-02.  You should continue your work with the expectation to finish it before the end of sprint 20-03, Friday Feb 14.

PR #1317, opened 2020-02-05
feat(sub-pattern): PictogramItem component created

RaphaelAmadeu:
### Related Ticket(s)

#1164 

### Description

I've separated the `PictogramItem` component from the former `PictogramArray` component.
It was updated with the latest functional specs, unit testing and documentation.

(CTA is not working for now, and will be working once the `ContentItem` is updated to work with the upcoming CTA component)

### Changelog

**New**

- Documentation and Unit testing
- `PictogramItem` created


packages/react/src/patterns/sub-patterns/PictogramItem/PictogramItem.js
```diff
@@ -0,0 +1,72 @@
...
+const PictogramItem = ({ heading, copy, Pictogram, cta }) => (
+  <div
+    data-autoid={`${stablePrefix}--pictogram-item`}
+    className={`${prefix}--pictogram-item`}>
+    <div className={`${prefix}--pictogram-item__row`}>
+      <div className={`${prefix}--pictogram-item__pic-col`}>
+        <Pictogram
+          data-autoid={`${stablePrefix}--pictogram-item_pictogram`}
+          className={`${prefix}--pictogram-item__pictogram`}
+          viewBox="8 8 32 32"
```
photodow: Is this not defined in the SVG itself?

packages/react/src/patterns/sub-patterns/PictogramItem/PictogramItem.js
```diff
@@ -0,0 +1,72 @@
...
+          height="80"
+          width="80"
```
photodow: Should this be in the `SCSS`?

packages/react/src/patterns/sub-patterns/PictogramItem/PictogramItem.js
```diff
@@ -0,0 +1,72 @@
...
+ * Converts markdown to html and returns object to be used in `dangerouslySetInnerHTML`
+ *
+ * @param {string} string string to be converted to html
+ * @returns {object} object with a __html key containing converted string
+ */
+//   const convertMarkdown = string => ({
+//     __html: markdownToHtml(string, {
+//       bold: false,
+//     }),
+//   });
```
photodow: I'm sure you know this stuff and other comments will need to go?

packages/react/src/patterns/sub-patterns/PictogramItem/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+> 💡 Don't forget to import the pictogram item styles from
+> [@carbon/ibmdotcom-styles](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/styles).
+
+## Props
+
+| Name        | Required | Data Type | Default Value | Description                                                          |
+| ----------- | -------- | --------- | ------------- | -------------------------------------------------------------------- |
+| `heading`   | YES      | String    | null          | Heading text                                                         |
+| `copy`      | YES      | String    | null          | Copy text                                                            |
```
photodow: Down here you have this all required, but in proptypes you do not.

packages/react/src/patterns/sub-patterns/PictogramItem/README.md
```diff
@@ -0,0 +1,87 @@
...
+import ReactDOM from 'react-dom';
+import { PictogramItem } from '@carbon/ibmdotcom-react';
+import 'yourapplication.scss';
+
+function App() {
+  const heading = 'Lorem ipsum dolor sit amet.';
+  const copy = 'Lorem ipsum dolor sit amet, consectetur adipiscing elit.';
+
+  const cta = {
+    type: 'text',
```
photodow: This should be defined inside the component once we get the CTA.

packages/react/src/patterns/sub-patterns/PictogramItem/README.md
```diff
@@ -0,0 +1,87 @@
...
+| `Pictogram` | YES      | Component | null          | Pictogram component imported from `@carbon/pictograms-react` package |
+| `cta`       | NO       | Object    | null          | Object with cta data, see `cta` bellow                               |
+
+### cta
+
+| Name   | Data Type | Description   |
+| ------ | --------- | ------------- |
+| `type` | String    | CTA Type      |
+| `href` | String    | CTA href      |
+| `copy` | String    | CTA copy text |
```
photodow: Wonder if we need to spell it out, or if we can just link them to the text link cta option for more information.

packages/react/src/patterns/sub-patterns/PictogramItem/PictogramItem.js
RaphaelAmadeu: I've moved this into the scss

packages/react/src/patterns/sub-patterns/PictogramItem/PictogramItem.js
RaphaelAmadeu: The `Pictogram` is the SVG imported by the adopter.

packages/react/src/patterns/sub-patterns/PictogramItem/PictogramItem.js
RaphaelAmadeu: I've cleaned the comments up

packages/react/src/patterns/sub-patterns/PictogramItem/README.md
RaphaelAmadeu: I've added the verification in the Proptypes

packages/react/src/patterns/sub-patterns/PictogramItem/README.md
RaphaelAmadeu: I've moved the type to inside of the `PictogramItem.js`

packages/react/src/patterns/sub-patterns/PictogramItem/README.md
RaphaelAmadeu: maybe as we'll be only dealing with href and copy this would be better to have just these 2 options on the documentation?

packages/react/src/patterns/sub-patterns/PictogramItem/PictogramItem.js
photodow: So we're passing in the `SVG`, and not the Pictogram component coming from Carbon?

[Pictogram Library](https://www.carbondesignsystem.com/guidelines/pictograms/library)
[@carbon/pictograms](https://www.npmjs.com/package/@carbon/pictograms)
[@carbon/pictograms-react](https://www.npmjs.com/package/@carbon/pictograms-react)

packages/react/src/patterns/sub-patterns/PictogramItem/README.md
photodow: Only problem I see with that is if it is defined by another component `LinkWithIicon` those props could change, and then now we have to track down everywhere it's mentioned like here in the documentation instead of having one source of truth and updating it there.

packages/react/src/patterns/sub-patterns/PictogramItem/PictogramItem.js
RaphaelAmadeu: Actually, I've misunderstood, this pictogram is received by a prop, the user needs to import the Pictogram and pass it to the component

packages/react/src/patterns/sub-patterns/PictogramItem/README.md
RaphaelAmadeu: makes sense, I'll link the CTA documentation, thanks!

RaphaelAmadeu: @photodow I've placed a link to the path where the CTA documentation will be available, once the block is finished.

packages/react/src/patterns/sub-patterns/PictogramItem/PictogramItem.js
photodow: Ok, so if we're pulling it from @carbon/pictograms-react — as we should — then I believe the viewBox is already included. Am I seeing that correctly?

https://raw.githubusercontent.com/carbon-design-system/carbon/master/packages/pictograms/svg/agriculture.svg

RaphaelAmadeu: The default viewbox does not match our sizing requirements, that's why I overwrote it.

packages/react/src/patterns/sub-patterns/PictogramItem/__stories__/PictogramItem.stories.js
```diff
@@ -0,0 +1,66 @@
...
+// import readme from '../README.md';
+import { storiesOf } from '@storybook/react';
+
+storiesOf('Patterns (Sub-Patterns)|PictogramItem', module)
+  .addDecorator(withKnobs)
+  //   .addParameters({
+  //     readme: {
+  //       sidebar: readme,
+  //     },
+  //   })
```
jeffchew: README should be turned back on

RaphaelAmadeu: @jeffchew I've turned it back on, thanks!

packages/react/src/patterns/sub-patterns/PictogramItem/README.md
jeffchew: We should probably be linking to the CTA README file, that's still in PR though with @mkothur .

jeffchew: but yes, in general I agree with @photodow , we are following this same methodology with our functional specs where we reference other components/patterns/etc, we should do the same here in our READMEs.

RaphaelAmadeu: @jeffchew  I've linked the readme file from Mahesh PR.

Issue #1282, opened 2020-01-31
Pattern:  Card - Types Link & Default


photodow:
## Detailed description

The card should have two types of options. One is where the whole card is clickable, and the other is where it's a static card with a CTA (that later is the default).

#### Additional information

- [Functional Specs](https://ibm.box.com/s/y1h8lgvn8zjdzr9r0owi7gvy5mjz7lcd)

#### Acceptance criteria

- [ ] Card link option by type
- [ ] Card link has optional icon as visual indicator
- [ ] Static card option by type
- [ ] Static card has optional CTA

PR #1320, opened 2020-02-05
fix(card): adding static/unclickable option to card based off type prop

annawen1:
### Related Ticket(s)

Pattern: Card - Types Link & Default #1282

### Description

The card should have two types of options. One is where the whole card is clickable, and the other is where it's a static card with a CTA (that later is the default).

This will be determined by the `type` prop. if `type === 'link'` then render `<ClickableTile />`, if not render `<Tile />`

### Changelog

**Changed**

- add `type` prop
- edit patterns using the `<CardLink />` to pass in `type` prop
- add `type` prop desc to CardLink README


packages/react/src/patterns/sub-patterns/CardLink/CardLink.js
```diff
@@ -47,6 +48,17 @@ const CardLink = ({
         {renderFooter(icon)}
       </div>
     </ClickableTile>
+  ) : (
+    <Tile
+      data-autoid={`${stablePrefix}--card-link`}
```
photodow: I see the stable selectors, file names, imports, and CSS classnames are still `cardlink`. Should we rename the component to just `card`?

packages/react/src/patterns/sub-patterns/CardLink/__stories__/Card.stories.js
```diff
@@ -27,6 +27,7 @@ storiesOf('Patterns (Sub-Patterns)|Card', module)
   .add('Link', () => {
```
photodow: Do we want to add a `default` or `static` story to storybook?

packages/react/src/patterns/sub-patterns/CardLink/CardLink.js
annawen1: yeah let's do that

packages/react/src/patterns/sub-patterns/CardLink/CardLink.js
```diff
@@ -47,6 +48,17 @@ const CardLink = ({
...
+      className={classNames(`${prefix}--card-link`, className)}>
+      <Image {...image} />
+      <div className={`${prefix}--card-link__wrapper`}>
+        <h3 className={`${prefix}--card-link__title`}>{title}</h3>
+        {optionalContent(content)}
+        {renderFooter(icon)}
+      </div>
```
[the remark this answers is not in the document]
annawen1: @photodow not sure what you mean, can you explain?

packages/react/src/patterns/sub-patterns/Card/Card.js
```diff
@@ -24,32 +24,40 @@ const { prefix } = settings;
...
-    return null;
-  }
-
-  return (
+  return type === 'link' ? (
     <ClickableTile
-      data-autoid={`${stablePrefix}--card-link`}
-      className={classNames(`${prefix}--card-link`, className)}
+      data-autoid={`${stablePrefix}--card`}
+      className={classNames(`${prefix}--card`, className)}
```
photodow: Do we need a modifier for link?

packages/react/src/patterns/sub-patterns/Card/Card.js
```diff
@@ -24,32 +24,40 @@ const { prefix } = settings;
...
+        <h3 className={`${prefix}--card__title`}>{title}</h3>
         {optionalContent(content)}
         {renderFooter(Icon)}
       </div>
     </ClickableTile>
+  ) : (
+    <Tile
+      data-autoid={`${stablePrefix}--card`}
+      className={classNames(`${prefix}--card`, className)}>
+      <Image {...image} />
```
photodow: You added the class name above, but I don't see it down here. Another reason it might be good to extract this section out so they are consistent as we discussed yesterday.

packages/react/src/patterns/sub-patterns/Card/Card.js
```diff
@@ -0,0 +1,97 @@
...
+const { prefix } = settings;
+
+/**
+ * Card Link Component
+ *
+ * @param {object} props react proptypes
+ * @returns {object} JSX object
+ */
+const Card = ({ title, copy, cta, image, className, type, ...props }) => {
+  const CardTile = type === 'link' ? ClickableTile : Tile;
```
photodow: :+1:

packages/react/src/patterns/sub-patterns/Card/Card.js
```diff
@@ -0,0 +1,97 @@
...
+  return (
+    <CardTile
+      data-autoid={`${stablePrefix}--card`}
+      className={classNames(`${prefix}--card`, className)}
+      href={cta.href}
```
photodow: Does this cause any problem staying here if only `Tile` is used?

packages/react/src/patterns/sub-patterns/Card/Card.js
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * Render footer with icon
+ *
+ * @param {object} cta cta object
+ * @param {string} type type of card (clickable/static)
+ * @returns {object} JSX object
+ */
+function renderFooter(cta, type) {
+  if (type !== 'link') {
+    return <CTA style="text" type={cta.type} href={cta.href} copy={cta.copy} />;
```
photodow: This might still need to be wrapped in footer so it sticks to the bottom.

packages/react/src/patterns/sub-patterns/Card/Card.js
```diff
@@ -0,0 +1,97 @@
...
+  title: PropTypes.string.isRequired,
+  icon: PropTypes.object,
+  copy: PropTypes.string,
+  cta: PropTypes.shape({
+    href: PropTypes.string.isRequired,
+    type: PropTypes.string,
+    icon: PropTypes.element,
+    copy: PropTypes.string,
+  }),
+  image: PropTypes.object,
```
photodow: Kind of waiting to see where https://github.com/carbon-design-system/ibm-dotcom-library/pull/1363#discussion_r377167071 lands.

packages/react/src/patterns/sub-patterns/Card/README.md
```diff
@@ -39,21 +37,34 @@ ReactDOM.render(<App />, document.querySelector('#app'));
...
+| Name        | Required | Data Type | Default Value | Description                                                                  |
+| ----------- | -------- | --------- | ------------- | ---------------------------------------------------------------------------- |
+| `copy`      | NO       | String    | null          | Paragraph of text that further describing the resource with added detail.    |
+| `className` | NO       | String    | null          | Classname to be assigned to the Card component                               |
+| `title`     | YES      | String    | n/a           | Concise yet descriptive string of text describing the linked resource.       |
+| `type`      | NO       | String    | n/a           | Determines whether card is clickable or static.                              |
+| `imgSrc`    | NO       | String    | null          | Image source to be passed as a property to the to the Card component         |
+| `altText`   | NO       | String    | null          | Image alternate text to be passed as a property to the to the Card component |
```
photodow: Might want to handle this in another pull request because it seems you're not handling prop and documentation updates, but I think this is outdated documentation.

packages/styles/scss/patterns/sub-patterns/card/index.scss
```diff
@@ -12,8 +12,8 @@
 @import 'carbon-components/scss/components/tile/tile';
 @import 'carbon-components/scss/components/link/link';
 
-@mixin card-link {
-  .#{$prefix}--card-link {
+@mixin card {
+  .#{$prefix}--card {
```
photodow: Seeing a couple visual issues we need to look into.

packages/react/src/patterns/sub-patterns/Card/__stories__/Card.stories.js
```diff
@@ -0,0 +1,111 @@
...
+          'Ratio',
+          ratio,
+          ratio['none']
+        )}`}>
+        <Card
+          title={title}
+          copy={copy}
+          cta={cta}
+          target={target}
+          type={type}
```
photodow: Would it be easier to change the type with a `select` rather than a whole new story? Just trying to think through this for a minute.

packages/react/src/patterns/sub-patterns/Card/__stories__/Card.stories.js
```diff
@@ -0,0 +1,111 @@
...
+  })
+  .add('Static', () => {
+    const title = text('title (required)', 'Lorem ipsum dolor sit amet');
+    const type = text('type', 'static');
+    const copy = text('copy', '');
+    const cta = {
+      icon: ArrowRight20,
+      type: 'local',
+      href: text('href (required)', 'https://example.com'),
+    };
```
photodow: I think we might be missing the `copy` here.

packages/react/src/patterns/sub-patterns/Card/Card.js
annawen1: ack yeah, let me stick that in the footer wrapper

packages/react/src/patterns/sub-patterns/Card/Card.js
annawen1: i don't think so

packages/react/src/patterns/sub-patterns/Card/Card.js
kennylam: @photodow no, it shouldn't.

packages/react/src/patterns/sub-patterns/Card/__stories__/Card.stories.js
annawen1: I want to keep it different since the cta options change depending on the type

photodow: I see. I guess it's concerns me we have so much of the repeated logic and content between stories. I guess adding a conditional with an either or CTA seems simpler to me, but I can live with it since it's just the story and not the actual component.

```javascript
const types = ['link', 'static'];
const type = select('type', types, types[0]);
const icons = {ArrowRight20, ArrowDown20, Launch20};

let cta;

if (type === 'link') {
  cta = {
    icon: select('icon', icons, ArrowRight20),
    href: text('href (required)', 'https://example.com')
  };
} else {
  cta = {
    type: 'local',
    copy: text('copy (required)', 'hello'),
    href: text('href (required)', 'https://example.com')
  };
}

return (
  <Card
    type={type}
    cta={cta[type]}
  />
)'

```

packages/react/src/patterns/sub-patterns/Card/Card.js
annawen1: yeah it's fine

Issue #1306, opened 2020-02-04
[SSR Test] Create SSR test environment


jeffchew:
#### User Story

> As a `[user role below]`:
IBM.com Library Tester

> I need to:
have an environment where I can test the server-side rendered React components

> so that I can:
quickly identify any rendering issues with the IBM.com Library via SSR

#### Additional information

 - A nextjs example exists, but should create a new test application to make room for testing tools and functionality that can be expanded in this environment (e.g. language/locale switcher, etc).

#### Acceptance criteria
 - [ ] Test environment that includes the React DotdomShell with sample content
 - [ ] Should deploy after every new available canary artifact

PR #1322, opened 2020-02-05
feat(ssr-test): adding language switching capabilities to nextjs test

jeffchew:
### Related Ticket(s)

#1306 

### Description

This adds the ability to switch languages from the locale selector in the NextJS SSR test environment

### Changelog

**New**

- Language support


packages/react/ssr-tests/nextjs/pages/_app.js
```diff
@@ -17,16 +21,42 @@ const langCode = {
...
    * @returns {*} Page wrapper JSX
    */
   render() {
-    const { Component, pageProps } = this.props;
+    const { Component, useLang, host, pageProps } = this.props;
     return (
       <Container>
-        <DotcomShell navigation="default" langCode={langCode}>
+        <Head>
+          <Altlang host={host} lang={useLang} />
```
annawen1: we don't need the `lang` prop here right? Don't see it being used in `<Altlang />`

jeffchew: thanks, good catch! pushing a fix.

PR #1329, opened 2020-02-06
fix(locale):; fixes #1300

### LINKED ISSUE #1300: Locale selector is not working for any locale other than en-us
## Detailed description

In ibm.com/search we deployed an update to use the locale selector (using release 1.3.0), the locale selector is only working for en-us, all other locales are failing, having an empty locale display in the footer and also an empty locale modal. Locally, the locale selector is working but all the locales are in English (we are in the process of reverting our changes)

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

Locale selector (LocaleButton/LocaleModal)

> see changed?

Locale selector showing up at least in English

All of them

1.3.0

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

IBM.com Search

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

- Screenshots or code

image

image

- Notes

Seems like all the requests for countrylist JSON files (example: /common/js/dynamicnav/www/countrylist/jsononly/mxes-utf8.json) are being redirected to /common/js/dynamicnav/www/usen-utf8.js, but that’s a JSONP file, not a valid JSON file. locally, seems to work bc there’s a CORS error early in the process and then the us-en JSON file is used by default, and that JSON file stills works (no redirect): https://www.ibm.com/common/js/dynamicnav/www/countrylist/jsononly/usen-utf8.json

kennylam:
### Related Ticket(s)

#1300 

### Description

Fixes fallback when translation response fails. 

### Changelog

**Changed**

- `us-en` is now always called in addition to user `cc-lc`, as opposed to only calling if user `cc-lc` fails.

### DIFF
--- a/packages/services/src/services/Locale/Locale.js
+++ b/packages/services/src/services/Locale/Locale.js
@@ -182,25 +182,28 @@ class LocaleAPI {
     if (sessionList) {
       return sessionList;
     } else {
-      let url;
-      try {
-        await axios.get(`${_endpoint}/${cc}${lc}-utf8.json`);
-        url = `${_endpoint}/${cc}${lc}-utf8.json`;
-      } catch (error) {
-        // use _localeDefault if 404
-        url = `${_endpoint}/${_localeDefault.cc}${_localeDefault.lc}-utf8.json`;
-      }
+      const axiosList = axios.create({
+        headers: {
+          'Content-Type': 'application/json; charset=utf-8',
+        },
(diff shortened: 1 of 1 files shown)


jeffchew:
@kennylam this needs to target the release branch.

kennylam:
> @kennylam this needs to target the release branch.

Doh!

PR #1338, opened 2020-02-07
fix(scss): fixed wrong scss selector issues in storybook

RaphaelAmadeu:
### Description

I've found a scss problem related to the `ContentGroup` on the `ListSection` that was affecting another components, I've fixed this and also changed the selector made for testing purposes on the `ContentGroup` story scss file.

### Changelog

**Changed**

- SCSS code


packages/styles/scss/patterns/blocks/listsection/_listsection.scss
```diff
@@ -76,6 +76,17 @@
...
+      padding-left: 0;
+    }
+
+    .#{$prefix}--content-group {
+      padding-bottom: $carbon--layout-03;
+      &:last-of-type {
+        padding-bottom: 0;
+      }
+    }
+
```
jeffchew: Adding @kennylam here, as he's working on the content block at the moment which includes refactoring listsection.

packages/styles/scss/patterns/blocks/listsection/_listsection.scss
```diff
@@ -76,6 +76,17 @@
       }
     }
 
+    .#{$prefix}--content-group__title {
+      padding-left: 0;
```
kennylam: @RaphaelAmadeu I'm seeing a lot of this in the `UseCases` pattern I'm currently refactoring. Styles like this aren't really necessary when using Carbon's grid. I've been removing stuff like this wherever I see it.

packages/react/src/patterns/sub-patterns/ContentGroup/__stories__/index.scss
```diff
@@ -1,6 +1,6 @@
 @import '../../../../../../styles/scss/patterns/sub-patterns/content-group/content-group';
```
photodow: Isn't this already being imported?

packages/react/src/patterns/sub-patterns/ContentGroup/__stories__/ContentArrayGroup.stories.js
```diff
@@ -19,5 +19,9 @@ storiesOf('Patterns (Sub-Patterns)|ContentGroup', module)
       'Lorem ipsum dolor sit amet.'
     );
 
-    return <ContentGroup heading={copy} />;
+    return (
+      <div className="content-group-story">
+        <ContentGroup heading={copy} />
+      </div>
```
photodow: Could we leverage the grid here instead? Similar to how we're doing in `FeatureCard`?

```html
<div class="bx--grid ">
  <div class="bx--row">
      ...
    </div>
  </div>
</div>
```

packages/styles/scss/patterns/blocks/listsection/_listsection.scss
photodow: 👆

packages/react/src/patterns/sub-patterns/ContentGroup/__stories__/ContentArrayGroup.stories.js
```diff
@@ -19,5 +19,9 @@ storiesOf('Patterns (Sub-Patterns)|ContentGroup', module)
       'Lorem ipsum dolor sit amet.'
```
photodow: Could we rename `ContentArrayGroup.stories.js` to `ContentGroup.stories.js`?

packages/styles/scss/patterns/blocks/listsection/_listsection.scss
```diff
@@ -76,6 +76,17 @@
       }
     }
 
+    .#{$prefix}--content-group__title {
```
jeffchew: Is this file getting renamed or that's being addressed in another PR?

PR #1343, opened 2020-02-07
docs(footer): adding data structure docs for the footer

jeffchew:
### Related Ticket(s)

No related issues

### Description

The data structure for passing in the data manually for the footer is unclear, adding this to the docs.

### Changelog

**Changed**

- Footer README


packages/vanilla/src/components/footer/README.md
```diff
@@ -70,6 +70,161 @@ Footer.init(elem); // initializes the footer
 | `tall`  | Default footer variant includes additional navigation taking up more space. |
 | `short` | Short footer variant reduces space by removing any additional navigation.   |
 
+### footerMenu (data structure)
```
photodow: Instead of spelling out the whole json model, does it make sense to link to the data model examples we're using for testing?

[footer-menu.json](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/react/src/components/Footer/__data__/footer-menu.json)
[footer-legal.json](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/react/src/components/Footer/__data__/footer-legal.json) (we could rename to thin possibly?)

jeffchew: We could, but the `data` part might throw people off.

jeffchew: I took a closer look, and was able to strip out the `data` part of the sample json and link directly instead.

photodow: I see your concern now. I for some reason thought you were talking about the directory at first. I think this is a cleaner and more scaleable approach though.

jeffchew: Agreed, should always avoid duplication in the docs for sure. Just waiting for the builds to go through and make sure everything looks ok.

Issue #1256, opened 2020-01-29
[Markdown Utility] Update utility to handle double new line


jeffchew:
#### User Story

> As a `[user role below]`:
IBm.com Library user

> I need to:
have the markdown utility automatically create paragraph tags

> so that I can:
utilize this tool in my application for string handling

#### Additional information

 - Functional specifications: https://ibm.ent.box.com/file/604090824435

#### Acceptance criteria
 - [ ] Markdown Utility is able to handle double new lines in strings and converts to `<p>` tags
 - [ ] Additional unit test coverage

PR #1363, opened 2020-02-10
feat(content-item): add cta component to content-item

annawen1:
### Related Ticket(s)

Create the Content Item sub-pattern #1176

### Description

Add `<CTA />` component to `<ContentItem />`

Screen Shot 2020-02-10 at 10 14 29 AM

### Changelog

**Changed**

- Add `<CTA />`
- only allow italicizing for the markdown
- add cta prop to content item README


packages/react/src/patterns/sub-patterns/ContentItem/ContentItem.js
```diff
@@ -40,16 +42,28 @@ const ContentItem = ({ copy, heading, image }) => (
         data-autoid={`${stablePrefix}--content-item__copy`}
         className={`${prefix}--content-item__copy`}
         dangerouslySetInnerHTML={{
-          __html: markdownToHtml(copy),
+          __html: markdownToHtml(copy, { bold: false }),
         }}></div>
     )}
+    {cta && <CTA style="text" type={cta.type} copy={cta.copy} />}
```
photodow: Are we missing an `href`?

photodow: I wonder if it makes sense to add this logic to the CTA component. Thoughts?

packages/react/src/patterns/sub-patterns/ContentItem/README.md
```diff
@@ -35,7 +35,12 @@ function App() {
     defaultImage: 'https://picsum.photos/id/2/352/176',
   };
 
-  return <ContentItem heading={heading} copy={copy} image={image} />;
+  const cta = {
+    type: 'jump',
+    copy: 'Click here',
+  };
```
photodow: I think we're missing an `href`

packages/react/src/patterns/sub-patterns/ContentItem/README.md
```diff
@@ -50,8 +55,16 @@ ReactDOM.render(<App />, document.querySelector('#app'));
...
 | `copy`    | NO       | String    | null          | Copy text                        |
+| `cta`     | NO       | object    | null          | CTA object. See `cta` below.     |
 | `image`   | NO       | object    | null          | Image object. See `image` below. |
 
+### cta (optional)
+
+| Name   | Data Type | Description                            |
+| ------ | --------- | -------------------------------------- |
+| `type` | String    | Type of CTA (jump, local, or external) |
+| `copy` | String    | Text of the CTA component              |
```
photodow: I think th emore important thing we outline in the documentation is the `style` being used, and possibly `types` allowed and then we can link to the  `CTA` component for documentation on how to use thosee styles and types.

packages/react/src/patterns/sub-patterns/ContentItem/__stories__/ContentItem.stories.js
```diff
@@ -20,6 +20,17 @@ storiesOf('Patterns (Sub-Patterns)|ContentItem', module)
...
+    const types = {
+      jump: 'jump',
+      local: 'local',
+      external: 'external',
+    };
+
+    const cta = {
+      copy: text('cta.copy', 'Lorem ipsum dolor sit amet'),
+      type: select('cta.type', types, types.local),
+    };
```
photodow: I think we're missing the `href` here.

packages/styles/scss/patterns/sub-patterns/content-item/_content-item.scss
```diff
@@ -28,6 +28,10 @@
       color: $text-01;
       @include carbon--type-style('body-long-02', true);
     }
+
+    .#{$prefix}--link-with-icon__container {
+      margin-top: $carbon--spacing-05;
+    }
```
photodow: What's the intent behind this here? I ask because I'm wondering if we should apply it somewhere else instead of inside of the `CTA` component if possible. If that's not going to work can we get a class name passed into it?

packages/react/src/patterns/sub-patterns/ContentItem/ContentItem.js
```diff
@@ -40,16 +42,28 @@ const ContentItem = ({ copy, heading, image }) => (
         data-autoid={`${stablePrefix}--content-item__copy`}
         className={`${prefix}--content-item__copy`}
         dangerouslySetInnerHTML={{
-          __html: markdownToHtml(copy),
+          __html: markdownToHtml(copy, { bold: false }),
```
photodow: I realize this is not apart of this issue, but I just want to point out #1256 is still outstanding for this sub-pattern so we keep it fresh in the forefront.

packages/react/src/patterns/sub-patterns/ContentItem/ContentItem.js
annawen1: yep forgot, adding it in now!

packages/styles/scss/patterns/sub-patterns/content-item/_content-item.scss
annawen1: It's still within the scope of `.#{$prefix}--content-item` so it's not going to affect the `CTA` component being used in other places right? I'm fine with having classname passed to it

photodow: Yeah, I guess my concern is the `link-with-icon` specific call out here. What happens if in the future that name changes or we decide to allow cards here. Is that spacing then going to be applied still (assuming it's all still relevant of course)?

packages/react/src/patterns/sub-patterns/ContentItem/ContentItem.js
```diff
@@ -40,16 +42,28 @@ const ContentItem = ({ copy, heading, image }) => (
...
   heading: PropTypes.string,
-  image: PropTypes.object,
+  image: PropTypes.shape({
+    images: PropTypes.shape({
+      src: PropTypes.string,
+      minWidth: PropTypes.string,
+    }),
+    alt: PropTypes.string,
+    defaultImage: PropTypes.string,
+  }),
```
photodow: I'm not sure I would worry about the image, or cta shape here since that's being defined by those components.

packages/styles/scss/patterns/sub-patterns/content-item/_content-item.scss
```diff
@@ -28,6 +28,10 @@
       color: $text-01;
       @include carbon--type-style('body-long-02', true);
     }
+
```
photodow: One note on the styling is the CTA extends the whole block, so I wonder if we need to add a float, or `inline-block` to contain it a little better.

image

packages/react/src/patterns/sub-patterns/ContentItem/ContentItem.js
jeffchew: hmm, might be overdoing it, should just see `cta` as its own component that can be used wherever. If we added a null check on the root level, would we do that for every component we make?

packages/react/src/patterns/sub-patterns/ContentItem/ContentItem.js
jeffchew: Hmm, we might want to if we are adding restrictions to certain props on a wrapper/decorator level, right? e.g. no passing of certain props...

photodow: I don't think we're doing much restriction outside of `cta.style` for this component. I feel like if you wanted to get more specific than object you could use `PropTypes.instanceOf()`. In this given scenario I wouldn't be worried about style because it'll just be ignored since the component itself is hard coding it.

```javascript
ContentItem.propTypes = {
  cta: PropTypes.instanceOf(CTA),
  image: PropTypes.instanceOf(Image),
  ...
}
```

You could use `exact` if you want to get a warning thrown when cta.styles is defined, but you'll still have to keep up with those props if anything changes down the road. There might be a scenario for one of our decorators, but I'm personally not seeing it here.

packages/react/src/patterns/sub-patterns/ContentItem/ContentItem.js
photodow: I would think we should. I guess it depends on the experience we want to build for our adopters. Do we want to throw a nasty error if the data is missing and incorrect, or do we just not render a component if the data is malformed. I've generally leaned towards the later throughout most of my development practice because then the application is at least still usable, but I can see a case for throwing errors too.

jeffchew: Not throwing errors in this case would make it tough for folks to debug I would think. It's also a new practice we haven't applied to our other components. Not to mention, it helps our own development when we're building out decorator patterns that make use of these sub-patterns and components, and making sure that we are utilizing them correctly. 

I think throwing errors is something we should keep in place.

packages/styles/scss/patterns/sub-patterns/content-item/_content-item.scss
annawen1: yeah I think this will need to be addressed in the CTA component

packages/react/src/patterns/sub-patterns/ContentItem/ContentItem.js
jeffchew: I like `instanceOf`.

annawen1: ok, changing to `instanceOf`!

packages/react/src/patterns/sub-patterns/ContentItem/ContentItem.js
jeffchew: @kennylam already completed this.

packages/styles/scss/patterns/sub-patterns/content-item/_content-item.scss
annawen1: Screen Shot 2020-02-11 at 10 26 27 AM

fixed

packages/styles/scss/patterns/sub-patterns/content-item/_content-item.scss
annawen1: wrapped in it a custom class

Issue #1342, opened 2020-02-07
Content group cards – text should span 90% width


oliviaflory:
## Detailed description

> Describe in detail the issue you're having.
The spec states that the text within the cards should only span 90% of the card container width, currently it spans 100% width.
**Spec**
![Screen Shot 2020-02-07 at 11.25.37 AM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/68f71d08-d135-4775-84a6-934c998696aa)
**Current**
![Screen Shot 2020-02-07 at 11.06.02 AM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/066ecee0-806b-4c56-8358-586f9c2812a9)

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

> Is this issue related to a specific component?
ContentGroupCards pattern

> What did you expect to happen? What happened instead? What would you like to
> see changed?
Update cards so text spans 90% container width instead of 100%

> What browser are you working in?
Chrome

> What version of the IBM.com Library are you using?
[ContentGroupCards](https://ibmdotcom-react.mybluemix.net/?path=/story/patterns-blocks-contentgroupcards--default)

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?
Web simplification

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

- Screenshots or code
- Notes

RobertaJHahn:
@RaphaelAmadeu Hi Raph, this bug is a low severity, but apparently it is an easy fix.  Can you fix this while you are working on the pattern in this sprint?

RaphaelAmadeu:
@RobertaJHahn Hi Roberta, sure thing, I can fix this in this sprint.

PR #1371, opened 2020-02-11
fix(scss): limited text wrapper size for 90% of card

RaphaelAmadeu:
### Related Ticket(s)

#1342 

### Description

I've added a 90% width limit to the texts inside the cards.
Screen Shot 2020-02-11 at 11 24 19
Screen Shot 2020-02-11 at 11 23 56

### Changelog

**New**

- added 90% width limit to the texts.

**Removed**

- Unnecessary scss breakpoint handling


packages/styles/scss/patterns/blocks/contentgroupcards/_contentgroupcards.scss
```diff
@@ -66,6 +62,11 @@
       }
     }
 
+    .#{$prefix}--card-link__title,
+    .#{$prefix}--card-link__content {
+      width: 90%;
+    }
```
photodow: This should be at the card level I would imagine. I think we want this 90% across all cards. @oliviaflory correct me if I'm wrong.

packages/styles/scss/patterns/blocks/contentgroupcards/_contentgroupcards.scss
```diff
@@ -40,12 +38,10 @@
...
-      @include carbon--breakpoint(sm) {
-        @include carbon--make-col(4, 4);
-        .#{$prefix}--content-group-cards__row {
-          padding-left: $carbon--layout-01;
-          padding-right: $carbon--layout-01;
-        }
+      @include carbon--make-col(4, 4);
+      .#{$prefix}--content-group-cards__row {
+        padding-left: $carbon--layout-01;
+        padding-right: $carbon--layout-01;
```
photodow: For another pull request we should create an issue to refactor this SCSS so it doesn't retain so much grid in it. It still needs to be able to fit inside of `ContentBlockMixed`

packages/styles/scss/patterns/blocks/contentgroupcards/_contentgroupcards.scss
```diff
@@ -66,6 +62,11 @@
       }
     }
 
+    .#{$prefix}--card-link__title,
+    .#{$prefix}--card-link__content {
+      width: 90%;
```
photodow: I wonder if there is a way we could create a or some mixins to help with some of this? Either just a single mixin to help with these couple of values, or maybe our own type mixins that wrap in this kind of stuff? Just thinking out loud at this point I guess.

packages/styles/scss/patterns/blocks/contentgroupcards/_contentgroupcards.scss
photodow: I believe there is also a `max-width: 640px` / 80 mu.

photodow: There is also a grid inside of the card that I'm personally choosing to ignore for the moment because I don't think it's correct. (if we follow the grid I believe it would look something like this...)

```scss
{
  width: 67.5%; // 90% of 75% or (6, 8)
  max-width: carbon--mini-unit(80);
}
```

@oliviaflory @wonilsuhibm any thoughts on the grid inside of the card?

image

packages/styles/scss/patterns/blocks/contentgroupcards/_contentgroupcards.scss
```diff
@@ -66,6 +62,11 @@
       }
     }
 
+    .#{$prefix}--card-link__title,
+    .#{$prefix}--card-link__content {
```
photodow: Something to keep in mind is I think @annawen1's pull request with her renaming of the card got merged yesterday.

```scss
.#{$prefix}--card__title,
.#{$prefix}--card__content { ... }
```

packages/styles/scss/patterns/blocks/contentgroupcards/_contentgroupcards.scss
```diff
@@ -40,12 +38,10 @@
     }
 
     .#{$prefix}--content-group {
-      @include carbon--breakpoint(sm) {
-        @include carbon--make-col(4, 4);
-        .#{$prefix}--content-group-cards__row {
-          padding-left: $carbon--layout-01;
-          padding-right: $carbon--layout-01;
-        }
+      @include carbon--make-col(4, 4);
```
photodow: Doesn't this just set 100%? Do we need that when a block element by default is 100%?

packages/styles/scss/patterns/blocks/contentgroupcards/_contentgroupcards-item.scss
```diff
@@ -5,13 +5,12 @@
...
-@import '../../../components/link-with-icon/link-with-icon';
-
 @mixin content-group-cards-item {
   .#{$prefix}--content-group-cards-item {
+    h3,
     p {
-      margin-bottom: $carbon--layout-03;
-      margin-top: $carbon--layout-03;
+      width: 90%;
+      max-width: carbon--mini-units(80);
```
photodow: I think we should create a mixin for this, and attack the root sub-patterns. Not the decorators. Also, are there no classnames we can use to style with instead of the `html`?

packages/styles/scss/patterns/blocks/contentgroupcards/_contentgroupcards.scss
photodow: Wondering if this got missed.

RaphaelAmadeu: I've removed it in the latest commit, thanks!

packages/styles/scss/patterns/blocks/contentgroupcards/_contentgroupcards-item.scss
RaphaelAmadeu: I've updated the selectors so they use the `Card` classes.

packages/styles/scss/globals/utils/_card-content-width.scss
```diff
@@ -0,0 +1,13 @@
+//Sets the Width and Max-Width of the Card to 90% / carbon--mini-units(80)
+
+@mixin card-content-width {
```
photodow: This is most likely going to go beyond just the card. You might just call it `content-width` because of that.

packages/styles/scss/globals/utils/_card-content-width.scss
```diff
@@ -0,0 +1,13 @@
...
+  .#{$prefix}--card__title,
+  .#{$prefix}--card__content {
```
photodow: We can remove the class names here. Just...

```scss
@mixin content-width {
  width: 90%;
  max-width: carbon--mini-units(80);
}
```

packages/styles/scss/patterns/blocks/contentgroupcards/index.scss
```diff
@@ -7,7 +7,7 @@
 
 @import '../../../globals/imports';
 @import 'contentgroupcards';
-@import 'contentgroupcards-item';
 @import '../../sub-patterns/content-group/content-group';
 @import '@carbon/grid/scss/grid';
 @import '../../sub-patterns/card/';
+@import '../../../globals/utils/card-content-width';
```
photodow: Probably could add this to `globals/imports` instead?

packages/styles/scss/patterns/sub-patterns/card/index.scss
```diff
@@ -28,12 +28,14 @@
 
     &__title {
       @include carbon--type-style('expressive-heading-03');
+      @include content-width;
```
photodow: @annawen1 keep an eye on these changes with your pull request.

Issue #1093, opened 2020-01-17
Card: Text and icon should not change to blue


larahanlon2:
## Detailed description

The text and should remain **white** on default and on hover/focus (or any interactive state). 
Please note that if this card becomes available in other themes, eg. light, the text and icon color will need to change from white to gray-100.

#### As-is

Currently, both the text and icon change to `blue` on hover:
![Screen Shot 2020-01-17 at 10.57.38 AM.png](https://images.zenhubusercontent.com/5d96606084daa600013e3afb/fa95d912-b6d2-4e0f-a63d-070d4256f3e0)

photodow:
@RobertaJHahn this needs to be applied to the card work.

Issue #1070, opened 2020-01-16
Pattern:  Card High Contrast Modifier


photodow:
## Detailed description

The current alpha work has it themed using the basic colors. We need a high contrast version that we can use in scenarios like this, and the CardSection where the background color is dark on a light theme, or light on a dark theme. It leverages different color tokens.

##### As-is
![image](https://user-images.githubusercontent.com/3793636/72571471-4c4ad580-3885-11ea-90d2-b3f44163b864.png)

##### To-be
![image](https://user-images.githubusercontent.com/3793636/72571495-5ec50f00-3885-11ea-827d-97fc51bf74da.png)

Notice how the background is dark, and the text is light. This can be achieved through inverse colors. Check out the card functional specs and design specs for more information.

> Created a story to align icon, and style "button"

#### Additional information

- [FeaturedLink Functional Specs](https://ibm.box.com/s/tlv7h1zgc886l5ovcap08uc1x97jpdo2)
- [Card Functional Specs](https://ibm.box.com/s/y1h8lgvn8zjdzr9r0owi7gvy5mjz7lcd)
- [Carbon Colors](https://www.carbondesignsystem.com/guidelines/color/overview)
- [CardSection Functional Specs](https://ibm.box.com/s/6lxup3zn1mbjgas1fv8ky19ntg7unpyv)

#### Acceptance Criteria

- [ ] Card has a inverse option for all themes, true or false
- [ ] Card follows functional specs data model, props, and naming

ljcarot:
FYI @chetankan

PR #1387, opened 2020-02-12
fix(card): add high contrast modifier to card

annawen1:
### Related Ticket(s)

Card: Text and icon should not change to blue #1093
Pattern: Card High Contrast Modifier #1070

### Description

A high contrast version (inverse) of the card that can be used with different color themes.

### Changelog

**New**

- add theme option to storybook so we can toggle and see how the card looks in different themes

**Changed**

- utilize different color tokens based based on the `inverse` flag


annawen1:
@photodow talked to @oliviaflory @larahanlon2 and @wonilsuhibm yesterday about the color tokens and we made some changes different from the tokens mentioned in the func specs

packages/react/src/patterns/sub-patterns/Card/__stories__/Card.stories.js
```diff
@@ -32,6 +32,8 @@ storiesOf('Patterns (Sub-Patterns)|Card', module)
       type: 'local',
       copy: text('cta.copy', ''),
     };
+    const inverse = boolean('inverse', true);
```
photodow: Can we have `false` be the default?

packages/react/src/patterns/sub-patterns/Card/__stories__/Card.stories.js
```diff
@@ -1,5 +1,5 @@
 import './index.scss';
```
photodow: We need to add an issue to add `image` to the knob options here in storybook.

packages/react/src/patterns/sub-patterns/Card/Card.js
```diff
@@ -40,11 +42,16 @@ const Card = ({
...
+        `${prefix}--card`,
+        { [`${prefix}--card--inverse`]: inverse },
+        className
+      )}
       href={href}
       {...props}>
       <Image {...image} classname={`${prefix}--card__img`} />
       <div className={`${prefix}--card__wrapper`}>
+        <p className={`${prefix}--card__eyebrow`}>{eyebrow}</p>
         <h3 className={`${prefix}--card__title`}>{title}</h3>
```
photodow: Since both of these are optional, can we render them `null` if not present?

packages/react/src/patterns/sub-patterns/Card/Card.js
```diff
@@ -27,10 +27,12 @@ const { prefix } = settings;
  */
 const Card = ({
   title,
```
photodow: It's fine if we're planning this for another pull request, but we haven't forgot about updating the props, correct?

packages/react/src/patterns/sub-patterns/Card/Card.js
```diff
@@ -82,16 +89,23 @@ function renderFooter(cta, type, href, Icon) {
   return (
     <footer className={`${prefix}--card__footer`}>
       {type !== 'link' ? (
-        <CTA style="text" type={cta.type} href={href} copy={cta.copy} />
+        <CTA
+          style="text"
+          type={cta.type}
+          href={href}
+          copy={cta.copy}
+          customClassname={`${prefix}--card__cta`}
```
photodow: We need to align on the naming of this prop. Some places we're calling it `customClassname`, and other places we're calling it `className`.

packages/react/src/patterns/sub-patterns/Card/Card.js
```diff
@@ -82,16 +89,23 @@ function renderFooter(cta, type, href, Icon) {
...
       ) : (
-        Icon && <Icon />
+        Icon && <Icon className={`${prefix}--card__cta`} />
       )}
     </footer>
   );
 }
 
 Card.propTypes = {
   title: PropTypes.string.isRequired,
```
photodow: Not required anymore.

packages/react/src/patterns/sub-patterns/Card/Card.js
```diff
@@ -100,6 +114,7 @@ Card.propTypes = {
     copy: PropTypes.string,
   }),
```
photodow: Can we maybe use instanceOf?

packages/react/src/patterns/sub-patterns/Card/__stories__/index.scss
```diff
@@ -5,5 +5,27 @@
...
 
-@import '../../../../../../styles/scss/patterns/sub-patterns/card/index';
+@import '../../../../../../styles/scss/patterns/sub-patterns/card';
 @import '@carbon/grid/scss/index.scss';
+
+.#{$prefix}--card--white {
+  @include carbon--theme($carbon--theme--white) {
+    @include card;
+  }
+}
```
photodow: Is this one necessary? Doesn't it build out in white by default?

packages/styles/scss/patterns/sub-patterns/card/index.scss
```diff
@@ -78,6 +92,28 @@
...
+      color: $inverse-01;
+    }
+
+    .#{$prefix}--card__eyebrow {
+      color: $text-03;
+    }
+
+    &:hover {
+      background-color: $inverse-hover-ui;
+    }
```
photodow: Double check your `:hover` and make sure it's not being applied to the static card. I think I had a comment in a previous pull request suggesting to add a modifier for `link` type.

packages/styles/scss/patterns/sub-patterns/card/index.scss
```diff
@@ -27,16 +27,29 @@
...
     &__title {
+      margin-bottom: $carbon--layout-05;
       @include carbon--type-style('expressive-heading-03');
+    }
 
-      margin-bottom: $carbon--layout-05;
+    &__title,
+    &__content {
+      color: $text-01;
     }
```
photodow: Just a thought. Instead of spelling these classnames out fully down below it might be good to do the following.

```scss
&__title,
&__content {
  ...
  .#{$prefix}--card--inverse  & {
    ....
  }
}

File: packages/styles/scss/patterns/sub-patterns/card/index.scss [hunk 11]
```diff
@@ -27,16 +27,29 @@
...
 
     &__content {
+      margin-bottom: $carbon--layout-01;
       @include carbon--type-style('body-short-02');
+    }
 
-      margin-bottom: $carbon--layout-01;
+    &__eyebrow {
       color: $text-02;
+      margin-bottom: $carbon--spacing-03;
```
photodow: Should we be consistent with using layout vs. spacing? Change to one or the other. I feel like spacing might be more appropriate here. Can't remember why I chose layout originally.

packages/react/src/patterns/sub-patterns/Card/__stories__/index.scss
```diff
@@ -5,5 +5,27 @@
...
+  @include carbon--theme($carbon--theme--g90) {
+    @include card;
+  }
+}
+
+.#{$prefix}--card--g100 {
+  @include carbon--theme($carbon--theme--g100) {
+    @include card;
+  }
+}
```
photodow: Since this is all apart of just the story, could we change the background so that can be clearly seen?

packages/react/src/patterns/sub-patterns/Card/Card.js
```diff
@@ -27,10 +27,12 @@ const { prefix } = settings;
...
   title,
+  eyebrow,
   copy,
   href,
   cta,
   image,
+  inverse,
   className,
   type,
   icon: Icon,
```
photodow: We'll probably need to updated this to the new SVG data model we were talking about.

```js
icon: {
  src: Icon,
  ...iconProps
}
```

packages/styles/scss/patterns/sub-patterns/card/index.scss
```diff
@@ -14,7 +14,7 @@
 
 @mixin card {
   .#{$prefix}--card {
-    color: $text-01;
```
photodow: By removing this color you've fixed #1404 as well it seems like. (although I'm not a fan of how the blue is coming from `bx--link` instead of our `scss`).

This introduces a new problem within the locale selector...



The icons are supposed to be `$text-01` (this card was originally built for the local selector which is why it was `$text-01` to begin with.)

packages/react/src/patterns/sub-patterns/Card/__stories__/index.scss
annawen1: @photodow change the background?

packages/styles/scss/patterns/sub-patterns/card/index.scss
annawen1: I thought for styling, carbon doesn't go more than two levels deep?

packages/styles/scss/patterns/sub-patterns/card/index.scss
annawen1: got it

packages/styles/scss/patterns/sub-patterns/card/index.scss
photodow: Oh if that's what you're going for I suppose. I'm not a believer of that restriction I suppose. 😉

packages/react/src/patterns/sub-patterns/Card/__stories__/index.scss
photodow: Right, `g10` for instance if you don't change the background of the story as well it just looks like some text sitting on a white background. If we could add a background to the story based on the theme I think it might go a long way helping to visualize it.

image

packages/styles/scss/patterns/sub-patterns/card/index.scss
annawen1: the blue is explicitly defined now like so
```
&__cta,
&__cta a:visited {
      color: $interactive-04;
}
```

packages/styles/scss/patterns/sub-patterns/card/index.scss
annawen1: hmm looking at carbon's website, it seems to me that `spacing` is used more for within a component and `layout` is for between components

https://www.carbondesignsystem.com/guidelines/spacing#spacing-scale

I'll change the rest of the tokens to use `spacing` instead

packages/react/src/patterns/sub-patterns/Card/Card.js
annawen1: @kennylam @andysherman2121 @RaphaelAmadeu @mkothur what do you guys think? when passing a classname to a child component, should we name the prop `customClassname` or `className`?

packages/react/src/patterns/sub-patterns/Card/Card.js
annawen1: I think I want to put it in another PR

packages/react/src/patterns/sub-patterns/Card/Card.js
annawen1: let's put this in another PR

packages/react/src/patterns/sub-patterns/Card/Card.js
mkothur: > @kennylam @andysherman2121 @RaphaelAmadeu @mkothur what do you guys think? when passing a classname to a child component, should we name the prop `customClassname` or `className`?

For me `customClassname` is make sense to me to differentiate with actual `className` for child component, I gave for `ContentGroup`.

annawen1: yeah I agree with that as well

packages/react/src/patterns/sub-patterns/Card/__stories__/index.scss
annawen1: done

annawen1:
@photodow forgot this needs to go into release, changing the base branch somehow throws a 500 error so opening a new PR

Issue #1381, opened 2020-02-12
Pattern-Table of Content- In iPhone with iOS 12 'Jump to' functionality is not working


chetankan:
## Detailed description
Issue:- In 'Table of Content' pattern 'jump to' functionality is not working on iPhone's which are running on iOS 12 version.On selection options from 'jump to' nothing happens.

## Steps to reproduce the issue

1. Enter the pattern url in the mobile browser
2. Click on 'Jump to' option
3. Select an 3rd or 4th option from the list.

## Additional information

Pattern url :- http://ibmdotcom-react-experimental.mybluemix.net/?path=/story/patterns-sub-patterns-table-of-contents--manually-define-menu-items

Devices :- iPhone xs, xs Max, XR, 8
OS:- iOS 12![Jumnp to section on mobile device.png](https://images.zenhubusercontent.com/5e21a841639f517d456f7e43/a3ab8bab-2e62-4f17-9c3b-0bf46a3d414a)

RaphaelAmadeu:
@chetankan ,
Could you please test this again using this link,
https://ibmdotcom-react.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/1423/index.html,
Thank you!

chetankan:
@RaphaelAmadeu :- I checked the link which you have provided on iOS 12 'Jump to' functionality is working fine.

Issue #1171, opened 2020-01-23
Smooth Scrolling Utility


photodow:
#### User Story
> As a `[user role below]`:
IBM.com Library Developer

> I need:
a smooth scrolling utility

> so that I can:
I can use it across multiple patterns and components.

The logic exists within the [Table of Content component](https://ibmdotcom-react-experimental.mybluemix.net/?path=/story/patterns-sub-patterns-table-of-contents--default), and could potentially be moved into its own utility.

#### Additional information
 - [Table of contents functional specs](https://ibm.box.com/s/3jy1tuk7zz6do61jfj55n7x7b2jpbzou)
 - [CTA Functional Specs](https://ibm.box.com/s/ceu99rjzd0e87sm51n9cv550yssockme)
 - [TableOfContents.js#L83-L101](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/react/src/patterns/sub-patterns/TableOfContents/TableOfContents.js#L83-L101)

#### Acceptance criteria
 - [x] Apply smooth scrolling logic to new component
 - [ ] Replace smooth scrolling logic from table of contents with new utility

PR #1389, opened 2020-02-12
fix(mobile): first attempt to fix target-scrolling on ios 12

RaphaelAmadeu:
### Related Ticket(s)

#1381 
### Description

I've changed the method used from `scrollIntoView` to `scrollTo`

### Changelog

**Changed**

- Mobile "Scroll to" handling


photodow:
@RaphaelAmadeu one suggestion is @mkothur I believe got his smooth scroll utility merged in. Maybe we can switch it out, and work on improving the utility instead?

#1171

RaphaelAmadeu:
> @RaphaelAmadeu one suggestion is @mkothur I believe got his smooth scroll utility merged in. Maybe we can switch it out, and work on improving the utility instead?
> 
> #1171


@photodow From what I could investigate until now, the problem of the scroll not happening on the iphone is due to the`scrollIntoView` not being supported by mobile safari, the fix I've tried was changing it to `scrollTo`

photodow:
Right, so if we make that change in the utility then the fix can be applied across the board.

RaphaelAmadeu:
First, let's ask for the QA team to test if this solution works with IOS 12, if it does, I can work along with @mkothur in order to implement this and solve this problem on future components that could use the scrolling.

Issue #1198, opened 2020-01-24
Content Section: Create the sub-pattern per functional specifications


jeffchew:
#### User Story

> As a `[user role below]`:
IBM.com Library developer

> I need to:
have a core sub-pattern available for Content sections

> so that I can:
build section level patterns with this core consistent pattern

#### Additional information

 - Functional specifications: https://ibm.ent.box.com/file/604984417521
 - Currently baked into the `CardSection`, part of the work should abstract this out of the `CardSection` and then have that pattern utilize the `ContentSection`

#### Acceptance criteria
 - [ ] Create the `ContentSection` per functional specifications
 - [ ] Separate out from the `CardSection` and use the `ContentSection` within that pattern
 - [ ] Storybook stories added with corresponding knobs
 - [ ] README documentation added
 - [ ] Minimum 80% unit test coverage

photodow:
@jeffchew it might be worth using the Content Block sub-patter on the right side of this layout because I've seen designs where their is also an optional title, and the text on the right side line up with the text within the card block, also I've seen they allow for a CTA at the bottom. Here's an example...

![image](https://user-images.githubusercontent.com/3793636/73182674-15738b80-40df-11ea-967b-13cd09da4b60.png)

PR #1406, opened 2020-02-13
Feat/1198 content section

mkothur:
### Related Ticket(s)

Related issue: #1198 

### Description

Created a contentSection sub pattern and added unit test and once it is approved will utilize this pattern inside CardSection .
### Changelog

**New**

- ContentSection
image

### DIFF
--- a/packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
+++ b/packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
@@ -21,33 +21,34 @@ const { prefix } = settings;
  * @param {string} props.heading  Heading objects
  * @param {string} props.theme theme name
  * @param {*} props.children JSX Components
+ * @param {string} props.className optional class to be applied to the containing node
  * @returns {*} JSX ContentSection component
  */
-const ContentSection = ({ heading, theme, children }) => {
+const ContentSection = ({ heading, theme, children, customClassName }) => {
+  const className = classNames(`${prefix}--content-section`, customClassName);
   /**
    * sets the class name based on theme type
    *
-   * @param {string} theme theme type ( g10 | white/default )
+   * @private
+   * @param {string} theme theme type
    * @returns {string} theme css class names
    */
-  const setTheme = theme => {
+  const _setTheme = theme => {
     return theme && `${prefix}--content-section--${theme}`;
   };
+
   return (
-    <div
-      className={classNames(`${prefix}--content-section`, setTheme(theme))}
+    <section
+      className={classNames(className, _setTheme(theme))}
       data-autoid={`${stablePrefix}--content-section`}>
       <div className={`${prefix}--content-section__heading__col`}>
-        <h3
-          data-autoid={`${stablePrefix}--content-section__heading`}
-          className={`${prefix}--content-section__heading`}>
-          {heading}
-        </h3>
+        <h2 className={`${prefix}--content-section__heading`}>{heading}</h2>
       </div>
       <div className={`${prefix}--content-section__children__col`}>
         <div className={`${prefix}--content-section__children`}>{children}</div>
       </div>
-    </div>
+      {/* <div className={`${prefix}--content-section__children`}>{children}</div> */}
+    </section>
   );
 };
 
@@ -55,6 +56,7 @@ ContentSection.propTypes = {
   heading: PropTypes.string.isRequired,
   theme: PropTypes.string,
   children: PropTypes.element,
+  customClassName: PropTypes.string,
 };
 
 export default ContentSection;
--- a/packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
+++ b/packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
@@ -25,7 +25,6 @@ const { prefix } = settings;
  * @returns {*} JSX ContentSection component
  */
 const ContentSection = ({ heading, theme, children, customClassName }) => {
-  const className = classNames(`${prefix}--content-section`, customClassName);
   /**
    * sets the class name based on theme type
    *
@@ -36,6 +35,8 @@ const ContentSection = ({ heading, theme, children, customClassName }) => {
   const _setTheme = theme => {
     return theme && `${prefix}--content-section--${theme}`;
   };
+
+  const className = classNames(`${prefix}--content-section`, customClassName);
 
   return (
     <section
--- a/packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
+++ b/packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
@@ -41,13 +41,16 @@ const ContentSection = ({ heading, theme, children, customClassName }) => {
     <section
       className={classNames(className, _setTheme(theme))}
       data-autoid={`${stablePrefix}--content-section`}>
-      <div className={`${prefix}--content-section__heading__col`}>
-        <h2 className={`${prefix}--content-section__heading`}>{heading}</h2>
+      <div className={`${prefix}--content-section__grid`}>
+        <div className={`${prefix}--content-section__row`}>
+          <div className={`${prefix}--content-section__left`}>
+            <h2 className={`${prefix}--content-section__heading`}>{heading}</h2>
+          </div>
+          <div className={`${prefix}--content-section__children`}>
+            {children}
+          </div>
+        </div>
       </div>
-      <div className={`${prefix}--content-section__children__col`}>
-        <div className={`${prefix}--content-section__children`}>{children}</div>
-      </div>
-      {/* <div className={`${prefix}--content-section__children`}>{children}</div> */}
     </section>
   );
 };
--- a/packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
+++ b/packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
@@ -15,13 +15,13 @@ const { stablePrefix } = ddsSettings;
 const { prefix } = settings;
 
 /**
- * ContentArrayGroup Component, for use with content arrays
+ * ContentSection Component, for use with cardSection
  *
  * @param {object} props props object
- * @param {string} props.heading  Heading objects
+ * @param {string} props.heading  Heading
  * @param {string} props.theme theme name
  * @param {*} props.children JSX Components
- * @param {string} props.className optional class to be applied to the containing node
+ * @param {string} props.customClassName optional class to be applied to the containing node
  * @returns {*} JSX ContentSection component
  */
 const ContentSection = ({ heading, theme, children, customClassName }) => {
@@ -36,11 +36,13 @@ const ContentSection = ({ heading, theme, children, customClassName }) => {
     return theme && `${prefix}--content-section--${theme}`;
   };
 
-  const className = classNames(`${prefix}--content-section`, customClassName);
-
   return (
     <section
-      className={classNames(className, _setTheme(theme))}
+      className={classNames(
+        `${prefix}--content-section`,
+        customClassName,
+        _setTheme(theme)
+      )}
       data-autoid={`${stablePrefix}--content-section`}>
       <div className={`${prefix}--content-section__grid`}>
         <div className={`${prefix}--content-section__row`}>
--- a/packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
+++ b/packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
@@ -1,0 +1,52 @@
+/**
+ * Copyright IBM Corp. 2016, 2018
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import classNames from 'classnames';
+import { settings as ddsSettings } from '@carbon/ibmdotcom-utilities';
+import PropTypes from 'prop-types';
+import React from 'react';
+import { settings } from 'carbon-components';
+
+const { stablePrefix } = ddsSettings;
+const { prefix } = settings;
+
+/**
+ * ContentArrayGroup Component, for use with content arrays
+ *
+ * @param {object} props props object
+ * @param {*} props.children JSX Components
+ * @param {string} props.heading  Heading objects
+ * @returns {*} JSX ContentArrayGroup component
+ */
+const ContentSection = ({ children, heading }) => {
+  return (
+    <div
+      className={`${prefix}--content-section`}
+      data-autoid={`${stablePrefix}--content-section`}>
+      <h3
+        data-autoid={`${stablePrefix}--content-section__title`}
+        className={`${prefix}--content-section__title`}>
+        {heading}
+      </h3>
+      <div
+        data-autoid={`${stablePrefix}--content-section__children`}
+        className={classNames(
+          `${prefix}--content-section__col`,
+          `${prefix}--content-section__children`
+        )}>
+        {children}
+      </div>
+    </div>
+  );
+};
+
+ContentSection.propTypes = {
+  heading: PropTypes.string,
+  children: PropTypes.element,
+};
+
+export default ContentSection;
--- a/packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
+++ b/packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
@@ -18,34 +18,42 @@ const { prefix } = settings;
  * ContentArrayGroup Component, for use with content arrays
  *
  * @param {object} props props object
- * @param {*} props.children JSX Components
  * @param {string} props.heading  Heading objects
- * @returns {*} JSX ContentArrayGroup component
+ * @param {string} props.theme theme name
+ * @param {*} props.children JSX Components
+ * @returns {*} JSX ContentSection component
  */
-const ContentSection = ({ heading, children }) => {
+const ContentSection = ({ heading, theme, children }) => {
+  /**
+   * sets the class name based on theme type
+   *
+   * @param {string} theme theme type ( g10 | white/default )
+   * @returns {string} theme css class names
+   */
+  const setTheme = theme => {
+    return theme && `${prefix}--content-section--${theme}`;
+  };
   return (
     <div
-      className={`${prefix}--content-section`}
+      className={classNames(`${prefix}--content-section`, setTheme(theme))}
       data-autoid={`${stablePrefix}--content-section`}>
-      <h3
-        data-autoid={`${stablePrefix}--content-section__title`}
-        className={`${prefix}--content-section__title`}>
-        {heading}
-      </h3>
-      <div
-        data-autoid={`${stablePrefix}--content-section__children`}
-        className={classNames(
-          `${prefix}--content-section__col`,
-          `${prefix}--content-section__children`
-        )}>
-        {children}
+      <div className={`${prefix}--content-section__heading__col`}>
+        <h3
+          data-autoid={`${stablePrefix}--content-section__heading`}
+          className={`${prefix}--content-section__heading`}>
+          {heading}
+        </h3>
+      </div>
+      <div className={`${prefix}--content-section__children__col`}>
+        <div className={`${prefix}--content-section__children`}>{children}</div>
       </div>
     </div>
   );
 };
 
 ContentSection.propTypes = {
-  heading: PropTypes.string,
+  heading: PropTypes.string.isRequired,
+  theme: PropTypes.string,
   children: PropTypes.element,
 };
 
--- a/packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
+++ b/packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
@@ -22,7 +22,7 @@ const { prefix } = settings;
  * @param {string} props.heading  Heading objects
  * @returns {*} JSX ContentArrayGroup component
  */
-const ContentSection = ({ children, heading }) => {
+const ContentSection = ({ heading, children }) => {
   return (
     <div
       className={`${prefix}--content-section`}
--- a/packages/react/src/patterns/sub-patterns/ContentSection/__stories__/ContentSection.stories.js
+++ b/packages/react/src/patterns/sub-patterns/ContentSection/__stories__/ContentSection.stories.js
@@ -3,7 +3,10 @@ import { text, withKnobs } from '@storybook/addon-knobs';
 import ContentSection from '../ContentSection';
 import React from 'react';
 import readme from '../README.md';
+import { settings } from 'carbon-components';
 import { storiesOf } from '@storybook/react';
+
+const { prefix } = settings;
 
 storiesOf('Patterns (Sub-Patterns)|ContentSection', module)
   .addDecorator(withKnobs)
@@ -18,5 +21,10 @@ storiesOf('Patterns (Sub-Patterns)|ContentSection', module)
       'Lorem ipsum dolor sit amet.'
     );
 
-    return <ContentSection heading={copy} />;
+    //return <ContentSection heading={copy} />;
+    return (
+      <div className={`${prefix}--grid`}>
+        <ContentSection heading={copy}></ContentSection>
+      </div>
+    );
   });
--- a/packages/react/src/patterns/sub-patterns/ContentSection/__stories__/ContentSection.stories.js
+++ b/packages/react/src/patterns/sub-patterns/ContentSection/__stories__/ContentSection.stories.js
@@ -29,11 +29,12 @@ storiesOf('Patterns (Sub-Patterns)|ContentSection', module)
     };
 
     return (
-      <div className={`${prefix}--grid ${prefix}--content-section-story`}>
+      <div className={`${prefix}--content-section-story`}>
         <ContentSection
           heading={heading}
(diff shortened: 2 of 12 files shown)


packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
```diff
@@ -0,0 +1,64 @@
...
+@import '../../../globals/imports';
+
+@mixin themed-items {
+  color: $text-01;
+  background: $ui-background;
+}
+
+@mixin content-section {
+  .#{$prefix}--content-section {
+    @include carbon--make-row;
```
photodow: You're missing a `@include carbon--make-container` I believe.

packages/react/src/patterns/sub-patterns/ContentSection/__stories__/ContentSection.stories.js
```diff
@@ -0,0 +1,36 @@
...
+      'Component title(required):',
+      'Lorem ipsum dolor sit amet'
+    );
+    const themes = {
+      g10: 'g10',
+      white: '',
+    };
+
+    return (
+      <div className={`${prefix}--grid`}>
```
photodow: This isn't needed.

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
```diff
@@ -0,0 +1,64 @@
...
+
+    &__heading {
+      margin-bottom: $carbon--layout-03;
+      @include carbon--type-style('expressive-heading-04', true);
+
+      &__col {
+        @include carbon--make-col-ready;
+        @include carbon--make-col(4, 4);
```
photodow: I think this just adds a 100%. Is that necessary for a block element that's already 100%?

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
```diff
@@ -0,0 +1,64 @@
...
+      padding-top: $carbon--layout-05;
+      padding-bottom: $carbon--layout-07;
+    }
+  }
+
+  .#{$prefix}--content-section--g10 {
+    @include carbon--theme($carbon--theme--g10) {
+      @include themed-items;
+    }
+  }
```
photodow: @annawen1 I noticed you were adding this to your storybook. Should we follow that same pattern here?

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
```diff
@@ -0,0 +1,64 @@
...
+  background: $ui-background;
+}
+
+@mixin content-section {
+  .#{$prefix}--content-section {
+    @include carbon--make-row;
+
+    &__heading {
+      margin-bottom: $carbon--layout-03;
+      @include carbon--type-style('expressive-heading-04', true);
```
photodow: Looks like you got this correct based on the typed out specs, but I don't think that's the correct token. My guess is it is supposed to be `expressive-heading-02`. @wonilsuhibm @larahanlon2 @oliviaflory can you confirm?

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
```diff
@@ -0,0 +1,64 @@
...
+
+        @include carbon--breakpoint(lg) {
+          @include carbon--make-col(4, 16);
+        }
+      }
+    }
+    &__children {
+      &__col {
+        @include carbon--make-col-ready;
+        @include carbon--make-col(4, 4);
```
photodow: See the above comment about 100%. I'm not sure this is necessary.

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
```diff
@@ -0,0 +1,64 @@
...
+  color: $text-01;
+  background: $ui-background;
+}
+
+@mixin content-section {
+  .#{$prefix}--content-section {
+    @include carbon--make-row;
+
+    &__heading {
+      margin-bottom: $carbon--layout-03;
```
photodow: What specs were you referring to? The ones I'm looking at says `64px`, not `32px`. So that would be `layout-05` instead.

image

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
```diff
@@ -0,0 +1,64 @@
...
+        @include carbon--make-col-ready;
+        @include carbon--make-col(4, 4);
+
+        @include carbon--breakpoint(lg) {
+          @include carbon--make-col(12, 16);
+        }
+      }
+    }
+
+    padding-top: $carbon--layout-05;
```
photodow: According to the specs this should be `32px` not `64px` here. So that would be a `layout-03` I believe.

image

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
```diff
@@ -0,0 +1,64 @@
...
+    padding-bottom: $carbon--layout-05;
+
+    @include carbon--breakpoint('md') {
+      padding-bottom: $carbon--layout-06;
+    }
+    @include carbon--breakpoint('lg') {
+      padding-top: $carbon--layout-05;
```
photodow: According to spec I'm seeing a `48px` here instead of `32px`. So that would be a `layout-04` I believe.

image

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
```diff
@@ -0,0 +1,64 @@
...
+@mixin themed-items {
+  color: $text-01;
+  background: $ui-background;
+}
+
+@mixin content-section {
+  .#{$prefix}--content-section {
+    @include carbon--make-row;
+
+    &__heading {
```
photodow: Now there is an interesting relationship between spacing that we need to take into account between the `heading`, and the pattern as a whole.

`@lg` breakpoint the pattern as a whole has only `48px` spacing top. While the heading has as `64px` margin-top. Because of the `48px` that's actually more like `16px` top that needs to be added here. I believe that would be a `layout-01`.

image

With that said that changes things for the smaller breakpoints as well. We could update this on a different break point, or we could just take it into account when adding spacing to the pattern top as a whole. Something to think about.

`@sm --> layout-01`
`@lg --> layout-03`

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
photodow: Taking that last comment, and the heading margin-top into account this would become `layout-01`

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
photodow: Taking the heading margin into account this would become `layout-03` instead.

mkothur:
@photodow: updated the PR for above comments

packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * @param {object} props props object
+ * @param {string} props.heading  Heading objects
+ * @param {string} props.theme theme name
+ * @param {*} props.children JSX Components
+ * @param {string} props.className optional class to be applied to the containing node
+ * @returns {*} JSX ContentSection component
+ */
+const ContentSection = ({ heading, theme, children, customClassName }) => {
+  const className = classNames(`${prefix}--content-section`, customClassName);
```
photodow: I don't think this line is needed. Either that, or put it below `_setTheme`, and add that piece here instead. That way you only use `classNames` once.

packages/react/src/patterns/sub-patterns/ContentSection/__stories__/ContentSection.stories.js
```diff
@@ -0,0 +1,41 @@
...
+    readme: {
+      sidebar: readme,
+    },
+  })
+  .add('Default', () => {
+    const heading = text(
+      'Component heading(required):',
+      'Lorem ipsum dolor sit amet'
+    );
+    const children = text('Grey box represents:', '');
```
photodow: What does the label here supposed to mean? Why not just `children`?

packages/react/src/patterns/sub-patterns/ContentSection/__stories__/ContentSection.stories.js
```diff
@@ -0,0 +1,41 @@
...
+      white: '',
+    };
+
+    return (
+      <div className={`${prefix}--content-section-story`}>
+        <ContentSection
+          heading={heading}
+          theme={select('theme', themes, themes.white)}
+          children={children}
+          customClassName={`${prefix}--col-lg-sm-8`}
```
photodow: Why are you passing `col-lg-sm-8` here? You might be able to remove the extra div, and add `content-section-story` here instead.

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
```diff
@@ -0,0 +1,83 @@
...
+      @include carbon--make-col-ready;
+    }
+    &__left {
+      @include carbon--breakpoint(lg) {
+        @include carbon--make-col(4, 16);
+      }
+    }
+    &__heading {
+      margin-bottom: $carbon--layout-05;
+      @include type-style('expressive-heading-02');
```
photodow: should fluid be set to `true`?

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
photodow: Take a look at this.

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
```diff
@@ -0,0 +1,83 @@
...
+        padding-top: $carbon--layout-03;
+      }
+
+      padding-top: $carbon--layout-03;
+      padding-bottom: $carbon--layout-07;
+    }
+
+    @include carbon--breakpoint('sm') {
+      &__heading {
+        padding-top: $carbon--layout-01;
```
photodow: `margin-top` I feel like would be more appropriate here, and this could be merged with lines 32-35 if i'm not mistaken.

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
```diff
@@ -0,0 +1,83 @@
...
+    @include carbon--breakpoint('lg') {
+      &__heading {
+        padding-top: $carbon--layout-03;
+      }
+
+      padding-top: $carbon--layout-03;
+      padding-bottom: $carbon--layout-07;
+    }
+
+    @include carbon--breakpoint('sm') {
```
photodow: I don't think the `sm` breakpoint does anything.

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
```diff
@@ -0,0 +1,83 @@
...
+      }
+    }
+
+    padding-top: $carbon--layout-01;
+    padding-bottom: $carbon--layout-05;
+
+    @include carbon--breakpoint('lg') {
+      &__heading {
+        padding-top: $carbon--layout-03;
+      }
```
photodow: This is just me personally, but I would put this under `&__heading` above on lines 32-35, and just include a breakpoint within there.

```scss
&__heading {
  ...
    padding-top: $carbon--layout-03;
  }
}
```

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
```diff
@@ -0,0 +1,83 @@
...
+        @include carbon--make-col(12, 16);
+      }
+    }
+
+    padding-top: $carbon--layout-01;
+    padding-bottom: $carbon--layout-05;
+
+    @include carbon--breakpoint('lg') {
+      &__heading {
+        padding-top: $carbon--layout-03;
```
photodow: I think you might be adding too much spacing here on `lg`. I don't think this `padding-top` is necessary.

photodow: I also think it's getting overridden from the one below it.

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
```diff
@@ -0,0 +1,83 @@
...
+      }
+
+      padding-top: $carbon--layout-03;
+      padding-bottom: $carbon--layout-07;
```
photodow: I think this is getting overwritten because `md` is after it.

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
```diff
@@ -0,0 +1,83 @@
...
+
+    padding-top: $carbon--layout-01;
+    padding-bottom: $carbon--layout-05;
+
+    @include carbon--breakpoint('lg') {
+      &__heading {
+        padding-top: $carbon--layout-03;
+      }
+
+      padding-top: $carbon--layout-03;
```
photodow: I think this should be `$carbon--layout-04`. You need `48px` above the children.

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
```diff
@@ -0,0 +1,83 @@
...
+    &__left,
+    &__children {
+      @include carbon--make-col-ready;
+    }
+    &__left {
+      @include carbon--breakpoint(lg) {
+        @include carbon--make-col(4, 16);
+      }
+    }
+    &__heading {
```
photodow: I believe this has a width of `90%`. See about syncing up with @RaphaelAmadeu and see if we can get a mixin or something for this. He's working on it for the Card #1342 .

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
annawen1: doesn't matter really, talked to @oliviaflory and 01, 02, 03 of expressive types are all fixed and won't scale anyways

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
annawen1: So the drupal team wants the option to change the theme through props which is why I think it's fine here?

For the `card` component, I set it through storybook since `card` being used in sections and blocks which will handle the theming

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
mkothur: Ok

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
mkothur: Ah I see. Will move below of breakpoints.

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
mkothur: Okay.

mkothur: @photodow updated the PR for above comments except this, will sync up with @RaphaelAmadeu on this.

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
photodow: Thanks @annawen1

packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
```diff
@@ -0,0 +1,66 @@
...
+import { settings } from 'carbon-components';
+
+const { stablePrefix } = ddsSettings;
+const { prefix } = settings;
+
+/**
+ * ContentArrayGroup Component, for use with content arrays
+ *
+ * @param {object} props props object
+ * @param {string} props.heading  Heading objects
```
photodow: Heading string

packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
```diff
@@ -0,0 +1,66 @@
...
+import { settings as ddsSettings } from '@carbon/ibmdotcom-utilities';
+import PropTypes from 'prop-types';
+import React from 'react';
+import { settings } from 'carbon-components';
+
+const { stablePrefix } = ddsSettings;
+const { prefix } = settings;
+
+/**
+ * ContentArrayGroup Component, for use with content arrays
```
photodow: Not sure this description is accurate at all.

packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * @param {object} props props object
+ * @param {string} props.heading  Heading objects
+ * @param {string} props.theme theme name
+ * @param {*} props.children JSX Components
+ * @param {string} props.className optional class to be applied to the containing node
```
photodow: It looks like you changed this prop name to align with others.

packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
```diff
@@ -0,0 +1,66 @@
...
+   */
+  const _setTheme = theme => {
+    return theme && `${prefix}--content-section--${theme}`;
+  };
+
+  const className = classNames(`${prefix}--content-section`, customClassName);
+
+  return (
+    <section
+      className={classNames(className, _setTheme(theme))}
```
photodow: You don't need to use `classNames` this many times.

```javascript
const _setTheme = theme => {
  return theme && `${prefix}--content-section--${theme}`;
};

return (
  <section
    className={classNames(`${prefix}--content-section`, customClassName, _setTheme(theme))}
    ...
  />
);
```

packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
```diff
@@ -0,0 +1,66 @@
...
+        </div>
+      </div>
+    </section>
+  );
+};
+
+ContentSection.propTypes = {
+  heading: PropTypes.string.isRequired,
+  theme: PropTypes.string,
+  children: PropTypes.element,
```
photodow: Should this be required? Thinking out loud here. What do y'all think?

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
photodow: ok, but let's not neglect this.

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
```diff
@@ -0,0 +1,82 @@
...
+      margin-bottom: $carbon--layout-05;
+
+      @include type-style('expressive-heading-02');
+
+      @include carbon--breakpoint('sm') {
```
photodow: I don't think breakpoint `sm` does anything.

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
```diff
@@ -0,0 +1,82 @@
...
+        margin-top: $carbon--layout-01;
+      }
+      @include carbon--breakpoint('lg') {
+        margin-top: $carbon--layout-03;
+      }
```
photodow: This is adding too much spacing. I don't see this in the specs.

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
photodow: I'm not seeing this update.

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
photodow: This looks like it's in the same place.

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
mkothur: So can I reduce the margin or remove totally?

packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
mkothur: done

packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
mkothur: done

packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
mkothur: done

packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
mkothur: done

packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
mkothur: you mean children is required?

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
photodow: If I'm reading it right this needs to be removed entirely. you should already have `$carbon--layout-01` which is all you need I think.

packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
photodow: yes

mkothur:
@photodow Updated the PR

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
mkothur: Okay

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
```diff
@@ -0,0 +1,74 @@
...
+    }
+    &__heading {
+      margin-top: $carbon--layout-01;
+      margin-bottom: $carbon--layout-05;
+
+      @include type-style('expressive-heading-02');
+      @include carbon--breakpoint('lg') {
+        margin-top: $carbon--layout-02;
+      }
```
photodow: This needs to be removed.

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
```diff
@@ -0,0 +1,74 @@
...
+        @include carbon--make-col(12, 16);
+      }
+    }
+
+    @include carbon--breakpoint('md') {
+      padding-bottom: $carbon--layout-06;
+    }
+
+    padding-top: $carbon--layout-01;
+    padding-bottom: $carbon--layout-05;
```
photodow: This should go above `md`.

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
```diff
@@ -0,0 +1,74 @@
...
+    &__children {
+      @include carbon--breakpoint(lg) {
+        @include carbon--make-col(12, 16);
+      }
+    }
+
+    @include carbon--breakpoint('md') {
+      padding-bottom: $carbon--layout-06;
+    }
+
```
photodow: Where did the `lg` `padding-top: $carbon--layout-04` go?

packages/react/src/patterns/sub-patterns/ContentSection/ContentSection.js
mkothur: here required make sense to me.

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
mkothur: removed.

photodow: Why? It's part of the design spec. I think you might have removed the wrong one. The one above was supposed to be removed.

mkothur: updated.

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
```diff
@@ -0,0 +1,74 @@
...
+    padding-top: $carbon--layout-01;
+    padding-bottom: $carbon--layout-05;
+
+    @include carbon--breakpoint('md') {
+      padding-bottom: $carbon--layout-06;
+    }
+    @include carbon--breakpoint('lg') {
+      padding-top: $carbon--layout-04;
+    }
```
photodow: I think we lost `padding-bottom: $carbon--layout-07;` at some point through all the shuffle.

mkothur: yeah, my bad will update.

Issue #1395, opened 2020-02-13
Pattern: Table of contents - Accessibility - Focus doesn't switch to the section heading on click of table of content (anchor) links


Praveen-sr:
## Detailed description
Device / browser :- JAWS Firefox

Issue :- Upon clicking of each link on left navigation, the Screen reader focus is not set on the beginning of the relevant content. Focus remains on the triggered link itself. This would confuse the Screen reader user.
Expected :- Upon clicking of each link, set the screen reader focus on the relevant heading.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/b58916fc-5836-4376-9d97-80c92076c15e)

RaphaelAmadeu:
@Praveen-sr ,
Could you please test using this link https://ibmdotcom-react.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/1413/index.html?path=/story/patterns-sub-patterns-table-of-contents--manually-define-menu-items ? Thanks!

Praveen-sr:
@RaphaelAmadeu @RobertaJHahn Verified this on link you provided, its working now. (focus goes to the target section)

RaphaelAmadeu:
@photodow Maybe should bring the discussion about the behaviour of these anchor buttons, as focussing the target section would not be a common behaviour in cases like this.

jeffchew:
I like the simplicity of aria tagging as a navigation element (e.g. aria landmarks), I think whatever we decide on we should document in the functional specs.

photodow:
Agreed, that's why I'm pushing for us to try and figure out how this should work.

RobertaJHahn:
@photodow I have added you to this issue to keep track the decision needed to close this issue.  I have also moved it to sprint 20-05 since it was not resolved in the previous sprint.

RobertaJHahn:
After talking to Ram:  We need to provide focus when the user clicks on a link.  The user needs to traverse the rest of the page to return.  A functional spec story is needed.  

Raph will store the logic that he had developed and approved by Ram.

PR #1413, opened 2020-02-14
fix(sub-pattern): screen reader content targeting

RaphaelAmadeu:
### Related Ticket(s)

#1395 
### Description

I've implemented a function that helps the screen readers by focusing the content when the scroll is triggered by the nav buttons.

This was tested and it's working with ChromeVox

### Changelog

**New**

- Focus on the target section

### DIFF
--- a/packages/react/src/patterns/sub-patterns/TableOfContents/TOCDesktop.js
+++ b/packages/react/src/patterns/sub-patterns/TableOfContents/TOCDesktop.js
@@ -64,11 +64,11 @@ const TOCDesktop = ({ menuItems, selectedId, updateState }) => {
     });
     const title = filteredItems[0].title;
     updateState(id, title);
+    triggerFocus(`a[name="${id}"]`);
     document.querySelector(`a[name="${id}"]`).scrollIntoView({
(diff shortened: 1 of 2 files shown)


packages/react/src/patterns/sub-patterns/TableOfContents/TOCDesktop.js
```diff
@@ -64,12 +64,25 @@ const TOCDesktop = ({ menuItems, selectedId, updateState }) => {
...
+   *
+   * @param {*} elem Selector to find the item
+   */
+  function triggerFocus(elem) {
+    const element = document.querySelector(elem);
+    element.setAttribute('tabindex', '0');
+    element.focus({ preventScroll: true });
+    element.removeAttribute('tabindex');
+  }
+
```
photodow: We should apply this to the utility function. The more add here to fix bugs the further the two become away from each other.

RaphaelAmadeu: So we should turn this into a utility?

Issue #1383, opened 2020-02-12
Pattern: Content Group Card - Accessibility - beginning and ending boundary for each card is not conveyed for the screen reader user.


Praveen-sr:
## Detailed description

Device - JAWS FF, VO iPhone Safari

Issue:- The beginning and ending boundary for each card is not conveyed for the screen reader user.
Suggestion:- Provide each card within the landmark region and give appropriate label. 

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/3fe0f7b4-f238-4916-a339-c25b8be641a9)

RaphaelAmadeu:
@Praveen-sr I've updated the outline, could you please confirm if it is okay now? thanks!

Praveen-sr:
We tested this on staging (https://ibmdotcom-react-experimental.mybluemix.net/) but issue still persists. Please let us know if its deployed on staging. Thanks

RaphaelAmadeu:
@Praveen-sr , could you please try again? I've just seem the link you used and the outline is bigger than before.

RaphaelAmadeu:
@Praveen-sr , could you please test it with this link from (PR Build)[https://ibmdotcom-react.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/1414/index.html?path=/story/patterns-blocks-contentgroupcards--default].

Thank you!

RobertaJHahn:
@RaphaelAmadeu Could you provide an estimate for your work.  Thanks.

Praveen-sr:
@RaphaelAmadeu Tested this on  PR build (https://ibmdotcom-react.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/1414/index.html?path=/story/patterns-blocks-contentgroupcards--default) its fine now.

PR #1414, opened 2020-02-17
Fix/content group cards accessibility

RaphaelAmadeu:
### Related Ticket(s)

#1383 
### Description

I've added `role` and `aria-labelledby` attributes to the cards, so they are better understood by screen readers.

### Changelog

**New**

- added `role` and `aria-labelledby`

**Changed**

- SCSS import order

### DIFF
--- a/packages/react/src/patterns/blocks/ContentGroupCards/ContentGroupCards.js
+++ b/packages/react/src/patterns/blocks/ContentGroupCards/ContentGroupCards.js
@@ -89,6 +89,8 @@ const _renderCards = items =>
         icon={ArrowRight20}
         href={elem.cta.href}
         type="link"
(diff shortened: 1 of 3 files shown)


photodow:
I'd love to understand the issue this is solving more. Where is the requirement coming from? Is it coming from WCAG, 508, or IBM A11y? Can we get a link to the documentation where it's documented. If this is a requirement we should share the documentation so we can all learn from this for the future.

I'm pushing back here a little because I'm not familiar with this requirement, and can't find it anywhere for block links. A quick test I did for VoiceOver on MacOs it works as I'm understanding it is supposed to work. We're using basic and semantic HTML5 here which should be accessible defining roles and labels and such.

Issue #1275, opened 2020-01-30
[ContentBlockSimple] Rename "SimpleLongForm" to "ContentBlockSimple" and update per functional specifications


jeffchew:
#### User Story

> As a `[user role below]`:
IBM.com Library user

> I need to:
have a `ContentBlockSimple` block pattern available

> so that I can:
utilize within my application

#### Additional information

 - Functional specifications: https://ibm.ent.box.com/file/595011203539
 - Video is out of scope for this story.
- Design QA testing issue (#1036)
- Prod QA testing issue (#1293)
 
#### Acceptance criteria
 - [x] Rename `SimpleLongForm` to `ContentBlockSimple`
 - [x] Remove feature flag
 - [x] Update propTypes and defaultProps per functional specifications (required vs not required)
 - [x] Minimum 80% unit test coverage
 - [x] Update Design QA / Production QA issues when code is merged and ready for testing
- [x] A comment is posted in the Design QA issue, tagging Wonil and Roberta, when development is finished
- [x] Assign Wonil to the Design QA issue
 - [x] A comment is posted in the Prod QA issue, tagging Praveen and Chetan, when development is

larahanlon2:
FYI `Content block – simple` design specs are now updated and available on Abstract: https://share.goabstract.com/06a41475-72e4-45d0-9bef-3fb0158bcdca?

PR #1429, opened 2020-02-19
Feat/content block simple

kennylam:
### Related Ticket(s)

#1275 
### Description

`SimpleLongForm` is now `ContentBlockSimple` with updated functional specifications.

### Changelog

**Changed**

- `SimpleLongForm` is now `ContentBlockSimple`


packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
```diff
@@ -0,0 +1,94 @@
...
+    copy={copy}
+    items={items}
+    cta={cta}
+    style
+    data-autoid={`${stablePrefix}--content-block-simple`}
+    customClassName={`${prefix}--content-block-simple ${prefix}--col-lg-8`}>
+    <div
+      data-autoid={`${prefix}--content-block-simple__content`}
+      className={`${prefix}--content-block-simple__content`}>
+      {_renderMedia(mediaType, mediaData)}
```
photodow: Looking at the design, the media belongs below the content.

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
```diff
@@ -0,0 +1,94 @@
...
+}) => (
+  <ContentBlock
+    heading={heading}
+    copy={copy}
+    items={items}
+    cta={cta}
+    style
+    data-autoid={`${stablePrefix}--content-block-simple`}
+    customClassName={`${prefix}--content-block-simple ${prefix}--col-lg-8`}>
+    <div
```
photodow: Is this `div` wrapper necessary`?

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
```diff
@@ -0,0 +1,94 @@
...
+  heading,
+  items,
+  mediaType,
+  mediaData,
+  cta,
+}) => (
+  <ContentBlock
+    heading={heading}
+    copy={copy}
+    items={items}
```
photodow: These aren't needed.

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
```diff
@@ -0,0 +1,94 @@
...
+    cta={cta}
+    style
```
photodow: Is this necessary?

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
```diff
@@ -0,0 +1,94 @@
...
+    data-autoid={`${stablePrefix}--content-block-simple`}
+    customClassName={`${prefix}--content-block-simple ${prefix}--col-lg-8`}>
```
photodow: Shouldn't use grid in the component.

photodow: Again, can we try to align on what we're calling the `customClassName`, `className`, or `customClassname`?

@annawen1 @mkothur

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
```diff
@@ -0,0 +1,94 @@
...
+ * @param {object} props.mediaData Data for renderimg the media
+ * @param {Array} props.items Array of objects with data for ContentItems
+ * @returns {*} Content block simple pattern
+ */
+const ContentBlockSimple = ({
+  copy,
+  heading,
+  items,
+  mediaType,
+  mediaData,
```
photodow: I understand this was in the functional specs, but I don't know if we should include it right now. All it should be is an `Image` object because we're not ready to bring in video yet. I don't know. Just my thoughts.

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+};
+
+/**
+ *
+ * @param {Array} items Array of data for ContentItems to be rendered
+ * @returns {*} Array of ContentItem Components
+ */
+const _renderContent = items =>
+  items.map((item, index) => <ContentItem {...item} key={index} />);
```
photodow: I think we're overcomplicating this.

First off I'm seeing the `heading` being passed through. There are no headings here.



Second, the `copy` prop should just be passed down here. With markdown we should be able to format multiple paragraphs here.

```jsx
<ContentBlock heading={heading} cta={cta}>
  <ContentItem copy={copy} />
  <Image {...image} />
</ContentBlock>
```

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
```diff
@@ -0,0 +1,94 @@
...
+
+ContentBlockSimple.propTypes = {
+  copy: PropTypes.string.isRequired,
+  heading: PropTypes.string.isRequired,
+  items: PropTypes.array.isRequired,
+  cta: PropTypes.object,
```
photodow: Can we use `instanceOf(CTA)` here?

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
```diff
@@ -0,0 +1,94 @@
...
+  mediaType: PropTypes.string,
+  mediaData: PropTypes.object,
```
photodow: If we can see to bring this into a single prop `image` we could use `instanceOf(Image)` here too.

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
```diff
@@ -0,0 +1,94 @@
...
+ * @param {Array} items Array of data for ContentItems to be rendered
+ * @returns {*} Array of ContentItem Components
+ */
+const _renderContent = items =>
+  items.map((item, index) => <ContentItem {...item} key={index} />);
+
+ContentBlockSimple.propTypes = {
+  copy: PropTypes.string.isRequired,
+  heading: PropTypes.string.isRequired,
+  items: PropTypes.array.isRequired,
```
photodow: This shouldn't be apart of this component.

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ *
+ * @param {string} mediaType Media type, video or image
+ * @param {object} mediaData Data for renderimg the media
+ * @returns {*} JSX Component with the media
+ */
+const _renderMedia = (mediaType, mediaData) => {
+  if (mediaData) {
+    if (mediaType === 'image') {
```
photodow: Can we combine these?

```javascript
if (mediaData && mediaType === 'image') { ... }
```

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
```diff
@@ -0,0 +1,94 @@
...
+      return (
+        <div data-autoid={`${stablePrefix}--content-block-simple__media`}>
```
photodow: Is it necessary to have this?

packages/react/src/patterns/blocks/ContentBlockSimple/README.md
```diff
@@ -0,0 +1,103 @@
...
+import ReactDOM from 'react-dom';
+import { SimpleLongForm } from '@carbon/ibmdotcom-react';
+import 'yourapplication.scss';
+
+function App() {
+  return (
+    <ContentBlockSimple
+      heading={heading}
+      copy={copy}
+      items={items}
```
photodow: Shouldn't be necessary.

packages/react/src/patterns/blocks/ContentBlockSimple/README.md
```diff
@@ -0,0 +1,103 @@
...
+      mediaType={mediaType}
+      cta={cta}
+    />
+  );
+}
+
+ReactDOM.render(<App />, document.querySelector('#app'));
+```
+
+> 💡 Don't forget to import the simplelongform styles from
```
photodow: Wrong name `simplelongform`.

packages/react/src/patterns/blocks/ContentBlockSimple/README.md
```diff
@@ -0,0 +1,103 @@
...
+> [@carbon/ibmdotcom-styles](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/styles).
+
+## Props
+
+| Name        | Required | Data Type | Default Value | Description                                                              |
+| ----------- | -------- | --------- | ------------- | ------------------------------------------------------------------------ |
+| `heading`   | YES      | String    | n/a           | Title of the content block.                                              |
+| `copy`      | YES      | String    | n/a           | Introduction copy briefly describing the section.                        |
+| `items`     | YES      | Array     | n/a           | Array of content items for use as content body.                          |
```
photodow: Shouldn't be necessary.

packages/react/src/patterns/blocks/ContentBlockSimple/README.md
```diff
@@ -0,0 +1,103 @@
...
+| `mediaData` | NO       | Array     | n/a           | Array of image sizes for use with content items.                         |
+| `mediaType` | NO       | String    | 'image'       | Type of media. Currently only supports `image`.                          |
+| `cta`       | NO       | Object    | n/a           | CTA used at the end of content body. `Text` and `Card` styles supported. |
```
photodow: Can we link to the CTA documentation?

packages/react/src/patterns/blocks/ContentBlockSimple/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+### items
+
+| Name      | Data Type | Description          |
+| --------- | --------- | -------------------- |
+| `heading` | String    | Item section heading |
```
photodow: Definitely not apart of this pattern.

packages/react/src/patterns/blocks/ContentBlockSimple/README.md
```diff
@@ -0,0 +1,103 @@
...
+| `copy`    | String    | Item section copy    |
+
+### mediaData
+
+| Name           | Data Type | Description                                            |
+| -------------- | --------- | ------------------------------------------------------ |
+| `src`          | String    | Url of image.                                          |
+| `minWidth`     | String    | Minimum width of image. Supports multiple breakpoints. |
```
photodow: I don't think this is accurate. Check the Image documentation if not just link out to it.

https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/react/src/components/Image/Image.js#L39-L49

packages/react/src/patterns/blocks/ContentBlockSimple/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+### CTA
+
+| Name    | Data Type | Description                                                  |
+| ------- | --------- | ------------------------------------------------------------ |
+| `style` | String    | CTA style. Supports `text` and `card`.                       |
+| `type`  | String    | CTA link behavior. Supports `external`, `jump`, and `local`. |
+| `title` | String    | CTA title.                                                   |
+| `href`  | String    | CTA url.                                                     |
+| `copy`  | String    | CTA copy.                                                    |
```
photodow: I wouldn't bother repeating this documentation here. `ContentBlock` should indicate how it's limited, and `CTA` should offer what props are.

packages/react/src/patterns/blocks/ContentBlockSimple/__stories__/ContentBlockSimple.stories.js
```diff
@@ -0,0 +1,85 @@
...
+    const ctaProps = {
+      style: select('CTA style', ctaStyles, ctaStyles.text),
+      type: select('CTA types', ctaTypes, ctaTypes.local),
+      title: 'Lorem ipsum dolor sit amet',
+      href: 'https://www.ibm.com',
+      copy: 'Lorem ipsum dolor sit ametttt',
+    };
+
+    return (
+      <div className={`${prefix}--grid`}>
```
photodow: So you're missing a row and column. I'd suggest following the pattern we've been using throughout the rest of the patterns.

```html
<div class="bx--grid ">
  <div class="bx--row">
      ...
    </div>
  </div>
</div>
```

packages/react/src/patterns/blocks/ContentBlockSimple/__stories__/ContentBlockSimple.stories.js
```diff
@@ -0,0 +1,85 @@
...
+          copy={select('Copy (required)', copy, copy['single paragraph'])}
+          heading={text(
+            'Heading (required)',
+            'Curabitur malesuada varius mi eu posuere'
+          )}
+          items={items}
+          mediaData={mediaData}
+          mediaType={mediaType}
+          cta={ctaProps}
+        />
```
photodow: Again, I think we're overcomplicating this.

packages/react/src/patterns/blocks/ContentGroupSimple/__stories__/ContentGroupSimple.stories.js
```diff
@@ -1,6 +1,7 @@
 import './index.scss';
 import { text, object, withKnobs, select } from '@storybook/addon-knobs';
 import ContentGroupSimple from '../ContentGroupSimple';
+import ContentGroupSimpleKnobs from './data/ContentGroupSimple.knobs';
```
photodow: I think I see what you're trying to do, but we should probably link to it in the documentation at the least so people can easily see how the data needs to be structured.

packages/styles/scss/patterns/blocks/content-block-simple/_content-block-simple.scss
```diff
@@ -0,0 +1,21 @@
...
+// LICENSE file in the root directory of this source tree.
+//
+@import '../../../globals/imports';
+@import '../../sub-patterns/card/index';
+@import '../../../components/link-with-icon/link-with-icon';
+
+@mixin content-block-simple {
+  .#{$prefix}--content-block-simple {
+    &__content {
+      padding-bottom: $carbon--spacing-07;
```
photodow: Is there a reason to apply it here instead of applying it to `ContentBlock`?

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
kennylam: @photodow I've changed this, but this is different in other func specs
image

Which is correct, or is it variable?

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
kennylam: Hm, yes, to give it a `data-autoid`.

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
kennylam: This is contrary to what I've been told by @wonilsuhibm. Are we not dictating what the maximum size of these patterns should be? Otherwise authors can place it in a 10-col if they wanted. The visual designs we use show these patterns laid out in an 8-col grid. Are we missing something?

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
kennylam: I... have no idea what that is doing there lol. Removed.

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
kennylam: If that's the case then we should update the func specs to not even mention video being an option, since as far as I know (and have been doing) all the patterns accommodate `video` because it's listed in the specs

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
kennylam: I don't understand, what do you mean they aren't needed? Look at the spec maybe the `ContentBlock` `copy` prop isn't needed. But the rest?
image

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
kennylam: I've been using `customClassName` everywhere. Happy to change to a standard, and agree that we should.

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
kennylam: I think @RaphaelAmadeu mentioned it was done this way in anticipation of a `video` type. But can condense for now.

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
kennylam: For the purpose of including a stable selector, yes. Unless we want to modify the media components to accept a `data-autoid` as a prop? That would work too.

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
kennylam: Ok I will change to receive only one `ContentItem` (I think I got my wires crossed on that one), but are you saying that `ContentBlock` should not have `heading` and `cta`? Per the spec:

image

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
kennylam: I don't understand. What props would be passed, and to where?

packages/react/src/patterns/blocks/ContentGroupSimple/__stories__/ContentGroupSimple.stories.js
kennylam: @photodow Yes for sure. I need to do this

packages/react/src/patterns/blocks/ContentBlockSimple/__stories__/ContentBlockSimple.stories.js
kennylam: This goes back to the question of whether the columns should be dictated by authors or the component. I feel like it should be at the component level, at least for the largest size. @wonilsuhibm is this correct?

packages/react/src/patterns/blocks/ContentBlockSimple/__stories__/ContentBlockSimple.stories.js
kennylam: For `ContentItems` right, and I'll simplify. But for `ContentBlock`?
https://github.com/carbon-design-system/ibm-dotcom-library/pull/1429#discussion_r382014098

packages/styles/scss/patterns/blocks/content-block-simple/_content-block-simple.scss
kennylam: This is a conversation I want to have with the team in terms of where spacing should be applied when nesting patterns. Right now we have block, group, and item levels all specifying their own spacing, and when used in another container it doesn't display correctly. This is even more apparent for optional items, e.g. if a heading with `padding-top` is optional, and is not present in a content block, the padding goes away with it and the above/below elements could now be crashing.

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
wonilsuhibm: I agree with both perspectives—I'm wondering if we can control how wide or narrow this can be—for example, this pattern should never be in a 4 column layout or 16 column layout. Can we do something like this: we offer two choices: this pattern can be either 8 or 6 cols wide, and let the adopter pick one, instead of letting them specify a random number of columns? And, with this, we do not have an actual need for anything other than 8 columns at the moment—maybe we cross this bridge of multiple options when there is an actual need...? What do you guys think?

packages/react/src/patterns/blocks/ContentBlockSimple/README.md
kennylam: I don't understand, can you explain?

packages/react/src/patterns/blocks/ContentBlockSimple/README.md
kennylam: Sure. I think it's a good idea to go back through all the READMEs and do this.

packages/react/src/patterns/blocks/ContentBlockSimple/README.md
kennylam: I think I'm misunderstanding the functional specs and how patterns are being wrapped in others. Is `heading` not part of `ContentItem`?
https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/react/src/patterns/sub-patterns/ContentItem/ContentItem.js#L24

packages/react/src/patterns/blocks/ContentBlockSimple/README.md
kennylam: I think I see what I did here... The CTA for this pattern comes from `ContentBlock` yeah? The children have no CTA at all.

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
kennylam: `CTA` isn't being imported here though

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
photodow: Please correct me if I'm becuase I'm still learning on this. Aren't they supposed to be added to the highest container, and interactive/dynamic elements? I neither here.

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
photodow: That type of layout is dictated at the section level. So for instance this will need to sit inside of something like table of contents. That's what will dictate the layout for this component. If we start adding grid elements to this it won't be able to sit where we need it to sit.

photodow: I get the impression this is what the team as aligned on. I added the comment here for everyone to see.

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
photodow: Only commenting on this because it didn't get resolved. ContentBlock only has heading `1a`, children `2`, and cta `not sure number right now`.

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
photodow: valid point.

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
photodow: Hmm... I wish I could see video coming, but i'm not sure how that's going to change everything yet.

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
photodow: I guess the purpose of stable selectors just isn't 100% clear with me. I thought it was for dynamic content or interactive content. Aside from that I believe the Image component itself already has a stable selector.

packages/react/src/patterns/blocks/ContentBlockSimple/README.md
photodow: Trying to catch them as we go.

packages/react/src/patterns/blocks/ContentBlockSimple/__stories__/ContentBlockSimple.stories.js
photodow: It's dictated by the section patterns.

packages/react/src/patterns/blocks/ContentBlockSimple/__stories__/ContentBlockSimple.stories.js
photodow: I should have added more detail here I think. I think I was referring to `items`, `mediaData`, `mediaType` as over complicating it mostly.

packages/styles/scss/patterns/blocks/content-block-simple/_content-block-simple.scss
photodow: We need to take a look at this across the board. This is on my to do for this sprint. Agreed it is a problem, but the solution I feel is to fix it at the sub-pattern level instead of having these one offs.

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
```diff
@@ -0,0 +1,74 @@
...
+ * @param {string} props.copy Content block simple short copy to support the heading
+ * @param {string} props.heading Content block simple heading
+ * @param {object} props.cta cta object
+ * @returns {*} Content block simple pattern
+ */
+const ContentBlockSimple = ({ copy, heading, image, cta }) => (
+  <ContentBlock
+    heading={heading}
+    cta={cta}
+    data-autoid={`${stablePrefix}--content-block-simple`}
```
photodow: Is this passing the `data-autoid` down?

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
```diff
@@ -0,0 +1,74 @@
...
+const { prefix } = settings;
+
+/**
+ * Content Block - Simple pattern
+ *
+ * @param {object} props props object
+ * @param {string} props.copy Content block simple short copy to support the heading
+ * @param {string} props.heading Content block simple heading
+ * @param {object} props.cta cta object
+ * @returns {*} Content block simple pattern
```
photodow: I see image as a prop, but i'm not seeing it in the docs here.

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
```diff
@@ -0,0 +1,74 @@
...
+ */
+const _renderMedia = (mediaType, mediaData) => {
+  if (mediaData && mediaType === 'image') {
+    return (
+      <div data-autoid={`${stablePrefix}--content-block-simple__media`}>
+        <Image {...mediaData} />
+      </div>
+    );
+  }
+};
```
photodow: Can we remove this section now?

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ *
+ * @param {Array} items Array of data for ContentItems to be rendered
+ * @returns {*} Array of ContentItem Components
+ */
+const _renderContent = items =>
+  items.map((item, index) => <ContentItem {...item} key={index} />);
```
photodow: Can we remove this?

packages/react/src/patterns/blocks/ContentBlockSimple/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+@import '@carbon/ibmdotcom-styles/scss/patterns/blocks/content-block-simple/content-block-simple';
+```
+
+> 💡 Only import fonts once per usage
+
+```javascript
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { SimpleLongForm } from '@carbon/ibmdotcom-react';
```
photodow: SimpleLongForm exists no more.

packages/react/src/patterns/blocks/ContentBlockSimple/README.md
```diff
@@ -0,0 +1,84 @@
...
+import 'yourapplication.scss';
+
+function App() {
+  return (
+    <ContentBlockSimple
+      heading={heading}
+      copy={copy}
+      items={items}
+      mediaData={mediaData}
+      mediaType={mediaType}
```
photodow: Need to pass in image now, right?

packages/react/src/patterns/blocks/ContentBlockSimple/README.md
```diff
@@ -0,0 +1,84 @@
...
+import ReactDOM from 'react-dom';
+import { SimpleLongForm } from '@carbon/ibmdotcom-react';
+import 'yourapplication.scss';
+
+function App() {
+  return (
+    <ContentBlockSimple
+      heading={heading}
+      copy={copy}
+      items={items}
```
photodow: I think we can get rid of this now.

packages/react/src/patterns/blocks/ContentBlockSimple/README.md
```diff
@@ -0,0 +1,84 @@
...
+> 💡 Don't forget to import the ContentBlockSimple styles from
+> [@carbon/ibmdotcom-styles](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/styles).
+
+## Props
+
+| Name      | Required | Data Type | Default Value | Description                                                              |
+| --------- | -------- | --------- | ------------- | ------------------------------------------------------------------------ |
+| `heading` | YES      | String    | n/a           | Title of the content block.                                              |
+| `copy`    | YES      | String    | n/a           | Simple content item.                                                     |
+| `image`   | NO       | Array     | n/a           | Optional image to use after content item. See `image` section.           |
```
photodow: Instead of an image section, maybe we can link to the Image documentation.

packages/react/src/patterns/blocks/ContentBlockSimple/__stories__/ContentBlockSimple.stories.js
```diff
@@ -0,0 +1,89 @@
...
+      sidebar: readme,
+    },
+  })
+  .add('Default', () => {
+    const ctaProps = {
+      style: 'card',
+      type: 'external',
+      title: 'Lorem ipsum dolor sit amet',
+      href: 'https://www.ibm.com',
+      copy: 'Lorem ipsum dolor sit ametttt',
```
photodow: I'd keep an eye on some of these to changes with @annawen1's pull request for high contrast.

packages/react/src/patterns/blocks/ContentBlockSimple/__stories__/ContentBlockSimple.stories.js
```diff
@@ -0,0 +1,89 @@
...
+          minWidth: 320,
+        },
+        {
+          src: 'http://fpoimg.com/400x400?bg_color=0f62fe&text_color=ffffff',
+          minWidth: 400,
+        },
+        {
+          src: 'http://fpoimg.com/672x672?bg_color=0f62fe&text_color=ffffff',
+          minWidth: 672,
+        },
```
photodow: This should work, but I don't know if it's necessary.

packages/react/src/patterns/blocks/ContentBlockSimple/__stories__/ContentBlockSimple.stories.js
```diff
@@ -0,0 +1,89 @@
...
+      Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aenean et ultricies est. Mauris iaculis eget dolor nec hendrerit. Phasellus at elit sollicitudin, sodales nulla quis, consequat libero.
+
+      Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aenean et ultricies est. Mauris iaculis eget dolor nec hendrerit. Phasellus at elit sollicitudin, sodales nulla quis, consequat libero.
+      `,
+    };
+
+    return (
+      <div className={`${prefix}--grid`}>
+        <ContentBlockSimple
+          copy={select('Copy (required)', copy, copy['single paragraph'])}
```
photodow: Could we just add a block of text instead of a select?

packages/styles/scss/patterns/blocks/content-block-simple/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+#### Modifiers
+
+Use these modifiers with `.bx--content-block-simple` class.
+
+| Class                                    | Description                                            |
+| ---------------------------------------- | ------------------------------------------------------ |
+| `.bx--content-block-simple--icon-link`   | The class name for simple long form with icon link     |
+| `.bx--content-block-simple--card`        | The class name for simple long form with card link     |
+| `.bx--content-block-simple--with-border` | The class name for simple long form with border bottom |
```
photodow: I'm not seeing these anywhere else, am I overlooking something?

packages/styles/scss/patterns/blocks/content-block-simple/README.md
```diff
@@ -0,0 +1,38 @@
...
+| ---------------------------------------- | ------------------------------------------------------------------------------- |
+| `.bx--content-block-simple__container`   | Element containing all simple long form content                                 |
+| `.bx--content-block-simple__row`         | Vertical divisions across the viewport                                          |
+| `.bx--content-block-simple__col`         | Horizonal divisions across the viewport and change in width at different screen |
+| `.bx--content-block-simple__title`       | Title for simple long form pattern                                              |
+| `.bx--content-block-simple__content`     | Content holder for short description                                            |
+| `.bx--content-block-simple__link`        | Wrapper for link component                                                      |
+| `.bx--content-block-simple__divider_col` | Horizonal division for divider element                                          |
+| `.bx--content-block-simple__divider`     | Divider element                                                                 |
```
photodow: Looks like in general this file needs to be updated. I don't think most of these are used.

packages/styles/scss/patterns/blocks/content-block-simple/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+#### Blocks
+
+| Class                       | Description                                               |
+| --------------------------- | --------------------------------------------------------- |
+| `.bx--content-block-simple` | Highest level class name for the simple long form pattern |
```
photodow: Still says simple long form.

packages/styles/scss/patterns/blocks/content-block-simple/_content-block-simple.scss
```diff
@@ -0,0 +1,37 @@
+//
+// Copyright IBM Corp. 2016, 2018
+//
+// This source code is licensed under the Apache-2.0 license found in the
+// LICENSE file in the root directory of this source tree.
+//
+@import '../../../globals/imports';
+@import '../../sub-patterns/card/index';
+@import '../../../components/link-with-icon/link-with-icon';
```
photodow: Should this be here?

packages/react/src/patterns/blocks/ContentBlockSimple/__stories__/ContentBlockSimple.stories.js
kennylam: I think we should keep this to show that it allows multiple paragraphs/styling.

packages/styles/scss/patterns/blocks/content-block-simple/README.md
kennylam: No, you're right. I didn't realize we even had READMEs for css...

packages/styles/scss/patterns/blocks/content-block-simple/_content-block-simple.scss
```diff
@@ -0,0 +1,37 @@
...
+
```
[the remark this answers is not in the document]
kennylam: > The image needs a `max-width` of `640px` or `80mu`.

Should this not be taken care of in the `Image` component?

kennylam:
> Is it the `ContentBlock` that's having CTA issues?
> 
> image

Hm, is it a `ContentBlock` issue? It's just importing CTA. This needs to be laid out at the parent level?

http://ibmdotcom-react.mybluemix.net/?path=/story/components-cta--default

packages/react/src/patterns/blocks/ContentBlockSimple/README.md
kennylam: Yes this has already been updated.

packages/react/src/patterns/blocks/ContentBlockSimple/README.md
kennylam: Yes this has already been updated.

packages/react/src/patterns/blocks/ContentBlockSimple/README.md
kennylam: I totally agree with this. The only reason I'm reluctant to do it now is because pretty much all the other READMEs are broken up into sections. I'll do it now but we should keep track of this and change them all in the future.

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
kennylam: No, good catch.

packages/styles/scss/patterns/blocks/content-block-simple/_content-block-simple.scss
kennylam: I agree. But with the total inconsistency from one block/group/component to the next, it is impossible not to have them in some cases. Will leave it as is for now and let's get an issue where we fix globally.

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
kennylam: Related to your comment below about `data-autoid` being passed down to `ContentBlock`, which it isn't. This has been fixed.

packages/react/src/patterns/blocks/ContentBlockSimple/ContentBlockSimple.js
kennylam: It isn't just for dynamic/interactive content. One of the original asks was whether content could be identified by sections, not simply by events.

packages/react/src/patterns/blocks/ContentBlockSimple/README.md
kennylam: This has been removed

packages/react/src/patterns/blocks/ContentBlockSimple/README.md
photodow: The README here hasn't been updated though.

packages/react/src/patterns/blocks/ContentBlockSimple/README.md
photodow: The README here hasn't been updated yet.

packages/react/src/patterns/blocks/ContentBlockSimple/README.md
photodow: This is part of the ontology work. Every README moving forward we've been trying to making this happen.

packages/react/src/patterns/blocks/ContentBlockSimple/__stories__/ContentBlockSimple.stories.js
photodow: Right, but all it needs is multiple paragraphs in a text field (if even that). Not all of these select options.

packages/styles/scss/patterns/blocks/content-block-simple/_content-block-simple.scss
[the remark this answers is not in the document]
photodow: Not from what I can see at the moment. Each pattern handles Images slightly different. For instance it would be weird for a card or leadspace to have a max-width.

photodow:
> Is it the ContentBlock that's having CTA issues?
> 
> image
> 
> Hm, is it a ContentBlock issue? It's just importing CTA. This needs to be laid out at the parent level?
> 
> http://ibmdotcom-react.mybluemix.net/?path=/story/components-cta--default

Yeah, I think so. The CTA shouldn't have any grid in it. It's literally just a switcher.

packages/styles/scss/patterns/blocks/content-block-simple/_content-block-simple.scss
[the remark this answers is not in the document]
kennylam: Ok sorry, I don't think i know what you mean then?

packages/styles/scss/patterns/blocks/content-block-simple/_content-block-simple.scss
```diff
@@ -0,0 +1,35 @@
...
+@import '../../../globals/imports';
+
+@mixin content-block-simple {
+  .#{$prefix}--content-block-simple {
+    &__content {
+      padding-bottom: $carbon--spacing-07;
+    }
+
+    .#{$prefix}--content-item {
+      margin-bottom: $carbon--spacing-07;
```
photodow: margin bottom here I think might be more determined based on the CTA. `32` is correct unless the CTA is a `.bx--card` in which then it needs to be `48px`. The `48px` I think is only for `@md` and higher.

I wonder if that's necessary, or if we could see if design is willing to having only one spacing between either CTA. @wonilsuhibm @oliviaflory @larahanlon2

packages/styles/scss/patterns/blocks/content-block-simple/_content-block-simple.scss
```diff
@@ -0,0 +1,35 @@
+//
+// Copyright IBM Corp. 2016, 2018
+//
+// This source code is licensed under the Apache-2.0 license found in the
+// LICENSE file in the root directory of this source tree.
+//
+@import '../../../globals/imports';
+
```
photodow: There are a couple issues with the ContentBlock spacing that maybe we can address in a seperate PR.

photodow: The heading margin bottom should be 32@sm and 48@md, not 48@sm, and 64@md

packages/react/src/patterns/blocks/ContentBlockSimple/__stories__/ContentBlockSimple.stories.js
```diff
@@ -0,0 +1,93 @@
...
+      type: 'external',
+      title: 'Lorem ipsum dolor sit amet',
+      href: 'https://www.ibm.com',
+      copy: 'Lorem ipsum dolor sit ametttt',
+    };
+
+    const cta = {
+      cta: ctaProps,
+      none: null,
+    };
```
photodow: I believe card, and text are options here. Might consider adding a text style option.

packages/styles/scss/patterns/blocks/content-block-simple/_content-block-simple.scss
photodow: The padding at the bottom should have a `64px@sm`.

packages/styles/scss/patterns/blocks/content-block-simple/_content-block-simple.scss
```diff
@@ -0,0 +1,35 @@
...
+      &__heading {
+        margin-top: 0;
+      }
+
+      p {
+        margin-bottom: $carbon--spacing-07;
+
+        &:last-of-type {
+          margin-bottom: 0;
```
photodow: I'm not sure this is necessary? Especially if you use margin up above. then margins can collapse.

packages/styles/scss/patterns/blocks/content-block-simple/_content-block-simple.scss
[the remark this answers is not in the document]
photodow: Spacing around the image is also incorrect. `48px` on top and bottom except on small where it's `32px` top and bottom.

packages/styles/scss/patterns/blocks/content-block-simple/_content-block-simple.scss
```diff
@@ -0,0 +1,35 @@
...
+
+    .#{$prefix}--content-item {
+      margin-bottom: $carbon--spacing-07;
+
+      &__heading {
+        margin-top: 0;
+      }
+
+      p {
+        margin-bottom: $carbon--spacing-07;
```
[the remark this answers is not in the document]
kennylam: I'll change to `06` but I purposely didn't change it in `contentItem` because i don't want to create issues in other patterns/groups that are using it. We can make this an issue to fix

packages/react/src/patterns/blocks/ContentBlockSimple/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+> 💡 Don't forget to import the ContentBlockSimple styles from
+> [@carbon/ibmdotcom-styles](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/styles).
+
+## Props
+
+| Name      | Required | Data Type | Default Value | Description                                                                                                                                                     |
+| --------- | -------- | --------- | ------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `heading` | YES      | String    | n/a           | Title of the content block.                                                                                                                                     |
+| `copy`    | YES      | String    | n/a           | Simple content item.                                                                                                                                            |
```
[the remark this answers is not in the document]
kennylam: I'll add a link to the utility

packages/styles/scss/patterns/blocks/content-block-simple/_content-block-simple.scss
kennylam: Same reason as https://github.com/carbon-design-system/ibm-dotcom-library/pull/1429#discussion_r382719090

packages/react/src/patterns/blocks/ContentBlockSimple/__stories__/ContentBlockSimple.stories.js
```diff
@@ -0,0 +1,94 @@
...
+      some common categories:
+
+      Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aenean et ultricies est. Mauris iaculis eget dolor nec hendrerit. Phasellus at elit sollicitudin, sodales nulla quis, consequat libero.
+
+      Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aenean et ultricies est. Mauris iaculis eget dolor nec hendrerit. Phasellus at elit sollicitudin, sodales nulla quis, consequat libero.
+      `,
+    };
+
+    return (
+      <div className={`${prefix}--grid`}>
```
photodow: Shouldn't be a deal breaker, but you're using prefix here, and not below. I don't think prefix is necessary since we're bringing the grid into storybook as is.

Issue #1235, opened 2020-01-29
Pattern:  Content group - with Pictograms - Optional learn more links at item level in patterns


ltitle:
When looking at both the List Section pattern and the Pictogram Array pattern I noticed that there is no ability in Storybook to remove the item level "Learn more ->" links. These should be optional links so the authors can choose whether or not to have them.

Similarly, when viewing the Use Case pattern I do not see any "Learn more ->" links or the ability to turn them on in Storybook.   This issue will be resolved in Storybook.

Expected outcome of implementation:
![image](https://user-images.githubusercontent.com/57198117/73383557-8b225780-4297-11ea-98af-68fc8923b93b.png)

![image](https://user-images.githubusercontent.com/57198117/73383595-9bd2cd80-4297-11ea-86f7-d7be7ae5c021.png)

![image](https://user-images.githubusercontent.com/57198117/73383620-a7be8f80-4297-11ea-9516-fd50f95b48f7.png)

![image](https://user-images.githubusercontent.com/57198117/73383638-af7e3400-4297-11ea-8416-bb7f72e5c784.png)

Acceptance criteria:
Item level text lists are optional for the following patterns:
- [ ] Content block - mixed groups
- [x] Content group - with pictograms
- [ ] Content block - with media

photodow:
Just to reiterate based on the conversation in our meeting today.

The `ContentItem` is used throughout all of these patterns, and handles the optional CTA. So to be clear the CTAs are optional in these patterns.

It's only on the story side where we've hardcoded them with the adopter dummy data we're passing in. We could possibly add a toggle that allows these to be turned on or off or something similar so they can be played with and tested in that way using storybook.

RaphaelAmadeu:
@oliviaflory , I was talking with @kennylam about these changes to the stories, and we concluded that as the `ContentBlock` will be refactored in order to use the `ContentGroup` and the Docs will be updated to reference the Sub-patterns capabilities on having the optional CTAs, updating the storybook now does not seem viable for the `ContentBlock`.

For the `ContentGroupPictograms`, I've updated the story adding a knob for optional CTA, and it's currently on PR state.

Thanks!

oliviaflory:
@RaphaelAmadeu thanks for the update!

PR #1431, opened 2020-02-19
feat(pattern): cta is now optional for content-group-pictograms

RaphaelAmadeu:
### Related Ticket(s)

#1235 

### Description

I've updated the component and story in order to make the CTAs optional.

### Changelog

**Changed**

- `PictogramItem` and `ContentGroupPictograms` updated


packages/react/src/patterns/sub-patterns/PictogramItem/PictogramItem.js
```diff
@@ -47,7 +47,7 @@ const PictogramItem = ({
         <ContentItem
           heading={heading}
           copy={copy}
-          cta={{ style: 'text', ...cta }}
+          cta={cta != null ? { style: 'text', ...cta } : null}
```
annawen1: does it work if we do `{cta && {style: 'text', ...cta}}`

packages/react/src/patterns/blocks/ContentGroupPictograms/__stories__/ContentGroupPictograms.stories.js
```diff
@@ -109,11 +119,7 @@ storiesOf('Patterns (Blocks)|ContentGroupPictograms', module)
           'Item 3 Copy (required)',
           'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aenean et ultricies est. Mauris iaculis eget dolor nec hendrerit. Phasellus at elit sollicitudin, sodales nulla quis, consequat libero.'
         ),
-        cta: object('Item 3 Link', {
-          type: 'local',
-          href: 'https://www.example.com',
-          copy: 'Lorem ipsum dolor',
-        }),
+        cta: toggleCta(boolean('Item 3 Cta')),
```
annawen1: don't you have to specify a boolean value?

`cta: toggleCta(boolean('Item 3 Cta', false))`

packages/react/src/patterns/sub-patterns/PictogramItem/PictogramItem.js
RaphaelAmadeu: yes, it does, I've updated it on the latest version

packages/react/src/patterns/blocks/ContentGroupPictograms/__stories__/ContentGroupPictograms.stories.js
RaphaelAmadeu: I've added it on the latest commit, Thanks!

packages/react/src/patterns/sub-patterns/PictogramItem/PictogramItem.js
```diff
@@ -47,7 +47,7 @@ const PictogramItem = ({
         <ContentItem
           heading={heading}
           copy={copy}
-          cta={{ style: 'text', ...cta }}
+          cta={cta && { style: 'text', ...cta }}
```
photodow: Maybe I'm missing something, but I think this is redundant.

https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/react/src/patterns/sub-patterns/ContentItem/ContentItem.js#L48-L56

packages/react/src/patterns/blocks/ContentGroupPictograms/__stories__/ContentGroupPictograms.stories.js
```diff
@@ -71,11 +89,7 @@ storiesOf('Patterns (Blocks)|ContentGroupPictograms', module)
           'Item 1 Copy (required)',
           'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aenean et ultricies est. Mauris iaculis eget dolor nec hendrerit. Phasellus at elit sollicitudin, sodales nulla quis, consequat libero.'
         ),
-        cta: object('Item 1 Link', {
-          type: 'local',
-          href: 'https://www.example.com',
-          copy: 'Lorem ipsum dolor',
-        }),
+        cta: toggleCta(boolean('Item 1 Cta', false)),
```
photodow: Instead of repeating all of this could we just repeat it?

```javascript
const numOfPictograms = number("Pictogram count", 35);
const items = [];

for (let i = 0; i < count; i++) {
  items.push({
    heading: text(`Heading ${i}`, '...');
    copy: text(`Copy ${i}` '...'),
    cta: toggleCta(boolean(`CTA ${i}`, false)),
  });
}
```

packages/react/src/patterns/blocks/ContentGroupPictograms/__stories__/ContentGroupPictograms.stories.js
```diff
@@ -61,6 +61,24 @@ storiesOf('Patterns (Blocks)|ContentGroupPictograms', module)
       'Lorem ipsum dolor sit amet, consectetur adipiscing elit.'
     );
 
+    /**
+     * Toggles CTA data based on item value
+     *
+     * @param {boolean} item defines if cta will be rendered
+     * @returns {*} if true returns cta data, if false, returns null
+     */
+    const toggleCta = item => {
```
photodow: I'd go all caps on `CTA` since it's an acronym.

packages/react/src/patterns/sub-patterns/PictogramItem/PictogramItem.js
RaphaelAmadeu: As this pattern only supports the `text` style, we need to pass it as a prop inside of the component, and due to this, even if the another `CTA` props are not passed, it will render an incomplete `CTA`.

packages/react/src/patterns/blocks/ContentGroupPictograms/__stories__/ContentGroupPictograms.stories.js
RaphaelAmadeu: I've updated it so it looks more like you pointed, Thanks!

packages/react/src/patterns/sub-patterns/PictogramItem/PictogramItem.js
photodow: ah, I think I see what you're saying. Interesting dilemma.

packages/react/src/patterns/blocks/ContentGroupPictograms/__stories__/ContentGroupPictograms.stories.js
```diff
@@ -61,65 +49,50 @@ storiesOf('Patterns (Blocks)|ContentGroupPictograms', module)
...
-          src: selectPictogram(pictogram3),
-          'aria-label': text('Aria-label 3:', 'Pattern'),
+          src: selectPictogram(
+            select(
+              `Item ${i + 1} Pictogram (required)`,
+              pictograms,
+              pictograms.Desktop
+            )
+          ),
+          'aria-label': text('Aria-label 1:', 'Desktop'),
```
photodow: I wouldn't add the aria label as a storybook knob. It just makes it harder to see and update what's going on.

RaphaelAmadeu: I've removed it from the story

Issue #1393, opened 2020-02-13
Pattern: Table of contents - Accessibility(keyboard) - On Firefox, each menu item getting dotted lines as visual focus indicator


Praveen-sr:
## Detailed description
Device : Windows 10
Browser : Firefox

Issue:- While navigating using keyboard, each menu item getting  dotted lines as visual focus indicator 
Suggestion : 
If, Visual Focus Indicator with 3 CSS px stroke then provide a sufficient contrast that is equal to 3:1. 2. Visual focus indicator for a link that is a 1 CSS px then provide a sufficient contrast greater than 4.5:1 
For more information please find the below link, https://www.w3.org/WAI/GL/low-vision-a11y-tf/wiki/Contrast_(Minimum)

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/165439ee-1296-42f3-9184-be7c750ce7c0)

Expected : (This is on Chrome browser)
![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/6c9ac95b-d00d-4363-96b3-b68e68921656)

PR #1443, opened 2020-02-20
fix(scss): updated focus state for the buttons on firefox

RaphaelAmadeu:
### Related Ticket(s)

#1393

### Description

I've updated the scss code for `TableOfContents` in order to improve the visibility of the focus on the buttons.

### Changelog

**New**

- Mozilla focus improved visibility


packages/styles/scss/patterns/sub-patterns/tableofcontents/_tableofcontents.scss
```diff
@@ -97,6 +97,14 @@
...
           cursor: pointer;
         }
+
+        @supports (-moz-appearance: none) {
+          a:focus {
+            outline: none;
+            border: 1px solid $interactive-04;
+            box-shadow: 0 0 4px 0 $interactive-04;
+          }
+        }
```
photodow: Did we reach out to design to see what the proper focus state is? I'm not seeing any focus state on here. Are we using the proper component? I say this because the problem may have already been solved with a different component, or we maybe able to solve it for all browsers instead of just Firefox.

RaphaelAmadeu: I'll reach out to them, this css is just a try of mimic the chrome focus state

photodow: Right, but Carbon has its own focus state that we should probably be following.

oliviaflory: There was no interactive specs for TOC in the initial specs. I just spoke to Raph and we decided he would enable Carbon focus state using Carbon's `$focus` token so it's aligned with the system.

RaphaelAmadeu: @photodow I've updated the SCSS in order to be compliant with Carbon system.

PR #1449, opened 2020-02-24
feat(utility): make use of smoothScroll util in tableOfContents

mkothur:
### Related Ticket(s)

Related issue: #1171 

### Description

Make use of smoothscroll utility function in TableOfContents pattern
### Changelog

**New**

- smoothScroll

**Changed**
 - using smoothScroll in tableOfContent


packages/utilities/src/utilities/smoothScroll/smoothScroll.js
```diff
@@ -11,19 +11,27 @@
...
  *
- * <a href="#anchorlinkname" onClick={smoothScroll(e)}>lorem ipsum</a>
+ * Here e is event param and seletor is param where you want to apply smoothscroll
+ * <a href="#anchorlinkname" onClick={smoothScroll({ e, selector })}>lorem ipsum</a>
  *
  * You can use this for jump to target element by providing event object.
  * it will scroll into view of target by selecting attribute and assigning to id.
  *
- * @param {*} e event object
+ * @param {*} e event object and seletor
```
RaphaelAmadeu: Maybe adding a description like
`@param {object} {e: Event, Selector: QuerySelector for the target element}`

packages/utilities/src/utilities/smoothScroll/smoothScroll.js
```diff
@@ -11,19 +11,27 @@
...
  */
-const smoothScroll = e => {
-  e.preventDefault();
-  const id = e.currentTarget.getAttribute('href');
-  document.querySelector(id).scrollIntoView({
-    behavior: 'smooth',
-    block: 'start',
-  });
+const smoothScroll = ({ e, selector }) => {
+  if (e != null) {
```
jeffchew: `!==` not `!=`

packages/utilities/src/utilities/smoothScroll/smoothScroll.js
mkothur: done

packages/utilities/src/utilities/smoothScroll/smoothScroll.js
mkothur: done

packages/utilities/src/utilities/smoothScroll/smoothScroll.js
```diff
@@ -11,19 +11,27 @@
...
+    document.querySelector(id).scrollIntoView({
+      behavior: 'smooth',
+      block: 'start',
+    });
+  } else if (selector !== null) {
+    document.querySelector(selector).scrollIntoView({
+      behavior: 'smooth',
+      block: 'start',
+    });
+  }
```
photodow: We can probably simplify this and reduce some of the redundancy down.

```javascript
const smoothScroll = ({ e, selector }) => {
  let getSelector;

  if (e !== null) {
    e.preventDefault();
    getSelector = e.currentTarget.getAttribute('href');
  } else if (selector !== null) {
    getSelector = selector;
  } else {
    return null;
  }

  document.querySelector(getSelector).scrollIntoView({
    behavior: 'smooth',
    block: 'start',
  });
};
```

packages/utilities/src/utilities/smoothScroll/smoothScroll.js
```diff
@@ -11,19 +11,27 @@
...
  *
- * <a href="#anchorlinkname" onClick={smoothScroll(e)}>lorem ipsum</a>
+ * Here e is event param and seletor is param where you want to apply smoothscroll
+ * <a href="#anchorlinkname" onClick={smoothScroll({ e, selector })}>lorem ipsum</a>
  *
  * You can use this for jump to target element by providing event object.
  * it will scroll into view of target by selecting attribute and assigning to id.
  *
- * @param {*} e event object
+ * @param {object} { e: Event, Selector: QuerySelector for the target element }
```
photodow: Are we gaining any benefits from using an object over two parameters?

packages/react/src/patterns/sub-patterns/TableOfContents/TOCMobile.js
```diff
@@ -92,14 +93,8 @@ const TOCMobile = ({ menuItems, selectedId, menuLabel, updateState }) => {
...
-    document.querySelector(`a[name="${id}"]`).scrollIntoView(true);
-    const scrolledY = root.scrollY;
-    if (scrolledY) {
-      root.scroll({
-        top: scrolledY - 48,
-        behavior: 'smooth',
-      });
-    }
+    const selector = `a[name="${id}"]`;
+    smoothScroll({ e: null, selector });
```
photodow: Based on the logic below I don't think `e: null` is necessary. Is that right?

packages/react/src/patterns/sub-patterns/TableOfContents/TOCDesktop.js
```diff
@@ -64,25 +65,10 @@ const TOCDesktop = ({ menuItems, selectedId, updateState }) => {
     });
     const title = filteredItems[0].title;
     updateState(id, title);
-    document.querySelector(`a[name="${id}"]`).scrollIntoView({
-      behavior: 'smooth',
-      block: 'start',
-    });
-    triggerFocus(`a[name="${id}"]`);
+    const selector = `a[name="${id}"]`;
+    smoothScroll({ e: null, selector });
```
photodow: See comment...

packages/utilities/src/utilities/smoothScroll/smoothScroll.js
mkothur: Okay

packages/utilities/src/utilities/smoothScroll/smoothScroll.js
mkothur: yeah @photodow  I believe CTA needs an event parameter  and  TOC needs a selector which we need to pass an id.

packages/react/src/patterns/sub-patterns/TableOfContents/TOCMobile.js
mkothur: @photodow If we did not pass an event with null argument in the object it is taking event as not null so and then looking for `href`. And I believe `e: null` is necessary that is how I am handling in utility.

packages/react/src/patterns/sub-patterns/TableOfContents/TOCDesktop.js
mkothur: @photodow If we did not pass an event with null argument in the object it is taking event as not null so and then looking for href. And I believe e: null is necessary that is how I am handling in utility.

packages/react/src/patterns/sub-patterns/TableOfContents/TOCMobile.js
photodow: I see then what if we removed the `null` requirement in the utility and just check if it was passed or not `!e`?

packages/utilities/src/utilities/smoothScroll/smoothScroll.js
photodow: Do we really need two different options, or can we make a decision on a consistent/opinionated way to do it?

mkothur: Sure

packages/react/src/patterns/sub-patterns/TableOfContents/TOCMobile.js
mkothur: @photodow updated the PR

packages/utilities/src/utilities/smoothScroll/smoothScroll.js
jeffchew: I think passing an object doesn't seem necessary, seems simpler to be:

`smoothScroll(e, selector)` and not `smoothScroll({e, selector})`

mkothur: @jeffchew first I tried without object `smoothScroll(e, selector) it is not working for TOC . so the reason I wrapped them in an object.

mkothur:
@jeffchew @photodow : updated the PR  and now utility being invoked with arguments without wrapping them in an object.

Issue #1170, opened 2020-01-23
Feature Card: Allow Different Icons


photodow:
Based on the CTA #1168 story we should look into opening up the icon to be changed by passing in an Icon similar to how CardLink works today.

#### Additional information
 - [Functional specs](https://ibm.box.com/s/tlv7h1zgc886l5ovcap08uc1x97jpdo2)
 - [CardLink Functional Specs](https://ibm.box.com/s/y1h8lgvn8zjdzr9r0owi7gvy5mjz7lcd)
 - [cta functional specs](https://ibm.box.com/s/ceu99rjzd0e87sm51n9cv550yssockme)

#### Acceptance criteria
 - [x] FeatureCard accepts prop to change icon

photodow:
@RobertaJHahn Removed this from Content Block epic, and moved into Card epic.

Issue #1071, opened 2020-01-16
Feature Card: Icon Position


photodow:
## Detailed description

The Icon Button on the featured link needs to be updated.

#### As-is

![image](https://user-images.githubusercontent.com/3793636/72571028-1822e500-3884-11ea-993d-3dec094257cb.png)

#### Expected
> 👀 see updated comments below from @larahanlon2 

Notice how the "button" needs to align right. ~Keep in mind this looks like a button, but does not function like a button because the whole thing is a clickable card link.~

#### Additional information

- [FeaturedLink Functional Specs](https://ibm.box.com/s/tlv7h1zgc886l5ovcap08uc1x97jpdo2)
- [Card Functional Specs](https://ibm.box.com/s/y1h8lgvn8zjdzr9r0owi7gvy5mjz7lcd)
- [Carbon Button Mixins](https://github.com/carbon-design-system/carbon/blob/master/packages/components/src/components/button/_mixins.scss)
- [Carbon Colors](https://www.carbondesignsystem.com/guidelines/color/overview)

#### Acceptance criteria

- [ ] Align icon right
- [ ] Add "button" visuals and states

larahanlon2:
After speaking with Wonil, Olivia and Lila, it's agreed that the blue button background should be **removed**. The icon will stay in the same position: bottom right of card.

**Design update. To be:**
<img width="970" alt="Screen Shot 2020-01-17 at 1 14 46 PM" src="https://user-images.githubusercontent.com/53093550/72635711-722ab580-392b-11ea-9909-c501f98f8294.png">

larahanlon2:
In the current Storybook version, the icon is in the wrong place.

**Current:**
<img width="817" alt="Screen Shot 2020-01-17 at 10 57 43 AM" src="https://user-images.githubusercontent.com/53093550/72635915-ee24fd80-392b-11ea-9e8b-8f4206a7549a.png">

It should be placed to the bottom right. See above.

photodow:
Thanks @larahanlon2 !

photodow:
@RobertaJHahn Moving this under the feature card epic

PR #1456, opened 2020-02-24
fix(feature-card): change icon position and accept icon as prop

annawen1:
### Related Ticket(s)

Feature Card: Allow Different Icons #1170
Feature Card: Icon Position #1071

### Description

1. Change icon position of the FeatureCard to the right.
2. Instead of importing the icon within the FeatureCard component itself, allow the icon to be passed in through the `cta` prop

Screen Shot 2020-02-25 at 12 44 04 PM

### Changelog

**Changed**

- float featurecard svg icon to the right
- import the icon through the stories and pass in to FeatureCard component via the `cta` prop


packages/styles/scss/patterns/blocks/FeaturedLink/_featuredlink.scss
```diff
@@ -12,13 +12,18 @@
     color: $text-01;
     background: $ui-background;
 
+    &__card svg {
+      float: right;
+    }
```
photodow: Can we also add the color change here? It's supposed to be a white arrow, and I think it makes sense to override it as a one off in this pattern since it's the only time I'm seeing it at the moment.

packages/react/src/patterns/blocks/FeaturedLink/FeaturedLink.js
```diff
@@ -27,7 +25,7 @@ const { prefix } = settings;
 const FeaturedLink = ({ heading, card }) => {
   const cardProps = {
     ...card,
-    cta: { ...card.cta, icon: { src: ArrowRight20, ...card.cta.icon } },
+    cta: { ...card.cta, icon: { src: card.cta.src, ...card.cta.icon } },
```
photodow: I think this should be `card.cta.icon.src`.

photodow: Since we're not hardcoding this anymore do we stille need `cardProps`? We might be able to just pass in `card` down below now instead.

```jsx
<Card
  heading={card.heading}
  cta={card.cta}
  inverse={true}
  type="link"
/>
```

The functional specs don't appear to be super clear here, but I think all that's required is the `heading`, and `cta`. Should we add a validation to make sure those two are present for this pattern?

packages/react/src/components/CTA/__stories__/CTA.stories.js
```diff
@@ -51,6 +52,7 @@ storiesOf('Components|CTA', module)
             heading: copy[1],
             cta: {
               href: urlBy[type],
+              src: ArrowRight20,
```
photodow: This and others should technically be `cta: { icon: { src: ArrowRight20 } }`. This is to keep the data models consistent and allow for passing props down like `aria-label`.

packages/styles/scss/patterns/blocks/FeaturedLink/_featuredlink.scss
annawen1: yep good catch

packages/react/src/patterns/blocks/FeaturedLink/FeaturedLink.js
annawen1: argh yep

packages/react/src/components/CTA/__stories__/CTA.stories.js
annawen1: yep completely missed this - will fix

packages/react/src/patterns/blocks/FeaturedLink/FeaturedLink.js
annawen1: yeah I can add validation here

annawen1: yeah I can add validation here

Issue #1067, opened 2020-01-16
Feature card: Card Needs to Hang


photodow:
## Detailed description

If you look at the as-is below you'll see that the text aligns flush with the edge of the card. What needs to happen is the card needs to hang into the gutters, so that the text inside the card lines up with the text above (assuming there were text in the left hand side, look at example instead).

##### As-is
![image](https://user-images.githubusercontent.com/3793636/72572268-a0ef5000-3887-11ea-9a1f-0744e64ab831.png)

##### To-be
![image](https://user-images.githubusercontent.com/3793636/72572453-2f63d180-3888-11ea-95d0-e5218268bfb4.png)

#### Additional information
- [FeaturedLink Functional Specs](https://ibm.box.com/s/tlv7h1zgc886l5ovcap08uc1x97jpdo2)

#### Acceptance Criteria
- [ ] Card lines hangs into gutter
- [ ] text remains in column
- [ ] text aligns with content above

RobertaJHahn:
@jeffchew @photodow @larahanlon2 @wonilsuhibm @oliviaflory Could you decide off-line how this treatment will be handled and document the decision in this issue?  Thanks

PR #1474, opened 2020-02-25
fix(feature-card): set card to hang into gutters

annawen1:
### Related Ticket(s)

Feature card: Card Needs to Hang #1067

### Description

Set card in the `FeatureCard` component to hang into the gutters as per [design specs](https://ibm.ent.box.com/file/608050432455)

Screen Shot 2020-02-25 at 3 53 35 PM

### Changelog

**New**

- add `hang` mixin to global utils folder

**Changed**

- set the content group title in `FeatureCard` to full width
- in `md` breakpoint and higher, have the card hang

### DIFF
--- a/packages/styles/scss/patterns/blocks/feature-card/_feature-card.scss
+++ b/packages/styles/scss/patterns/blocks/feature-card/_feature-card.scss
@@ -37,6 +37,7 @@
       bottom: 0;
       @include carbon--breakpoint('md') {
         flex-direction: row;
+        @include hang;
       }
 
       &__wrapper,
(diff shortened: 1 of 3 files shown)


packages/styles/scss/patterns/blocks/feature-card/_feature-card.scss
```diff
@@ -48,6 +49,16 @@
...
       }
     }
+
+    @include carbon--breakpoint('md') {
+      .#{$prefix}--content-group__title {
+        max-width: 100%;
+      }
+
+      .#{$prefix}--content-group__children {
+        @include hang;
```
photodow: I'd personally put it on the card itself since that's what's doing the hanging.

annawen1: ok!

Issue #1481, opened 2020-02-26
Pattern: Card Section Simple -Accessibility- screen reader - beginning and ending boundary for each card is not conveyed for the user.


Praveen-sr:
## Detailed description

Device - JAWS FF, VO iPhone Safari

Issue:- The beginning and ending boundary for each card is not conveyed for the screen reader user.
Suggestion:- Provide each card within the landmark region and give appropriate label.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/2f274907-7ead-4d4c-ac4a-43db62467abf)

mkothur:
@Praveen-sr , I have added the role and aria-labelledby attributes  could you please test it here from [PR deploy preview](https://ibmdotcom-react.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/1489/index.html?path=/story/patterns-sections-cardsectionsimple--default). And let me is it behaves as expected or not. thank you!

Praveen-sr:
@mkothur @RobertaJHahn Re tested this on the link provided above, issue still persists.
Recommendation:
provide region role for the div which contains the card with aria-labelledby. and associate its id with the heading for that particular card.
below is an example link for reference.
https://www.w3.org/TR/wai-aria-practices/examples/landmarks/region.html 
Thanks

mkothur:
@Praveen-sr: update the changes could test with the same above build URL.

Praveen-sr:
@mkothur @RobertaJHahn Re tested this on the [PR build URL](https://ibmdotcom-react.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/1489/index.html?path=/story/patterns-sections-cardsectionsimple--default) provided above, issue still persists.

#1383 - Is the same issue, its fixed and working.

Praveen-sr:
@mkothur Re tested this with the new link, its working now. Cards beginning and ending boundary is conveyed to the user.
@RobertaJHahn We can close this issue. Thanks

PR #1489, opened 2020-02-26
fix(pattern): added role and aria-labelledby in card of cardSection

mkothur:
### Related Ticket(s)

Related issue: #1481 

### Description
 Added the role and aria-labelledby attributes to the cards of cardSection pattern, so that the screen readers could understand.

### Changelog

**New**

- added role and aria-labelledby

### DIFF
--- a/packages/react/src/patterns/sections/CardSection/CardSection.js
+++ b/packages/react/src/patterns/sections/CardSection/CardSection.js
@@ -87,7 +87,10 @@ const _renderCards = (cards, containerRef, cta) => (
     ref={containerRef}>
     {cards.map((card, index) => {
       return (
-        <div className={`${prefix}--card-section__cards__col`}>
+        <div
+          className={`${prefix}--card-section__cards__col`}
+          role="region"
+          aria-labelledby={`${card.heading}`}>
           <Card
             key={index}
             image={card.image}
@@ -101,8 +104,7 @@ const _renderCards = (cards, containerRef, cta) => (
               },
             }}
             type="link"
-            role="region"
-            aria-labelledby={`region${index}`}
+            id={`${card.heading}`}
           />
         </div>
       );
--- a/packages/react/src/patterns/sections/CardSection/CardSection.js
+++ b/packages/react/src/patterns/sections/CardSection/CardSection.js
@@ -4,6 +4,7 @@
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
  */
+
 import React, { useEffect, useRef } from 'react';
 import { ArrowRight20 } from '@carbon/icons-react';
 import { Card } from '../../sub-patterns/Card';
--- a/packages/react/src/patterns/sections/CardSection/CardSection.js
+++ b/packages/react/src/patterns/sections/CardSection/CardSection.js
@@ -100,8 +100,9 @@ const _renderCards = (cards, containerRef, cta) => (
                 src: ArrowRight20,
               },
             }}
-            target={card.cta.target}
             type="link"
+            role="region"
+            aria-labelledby={`region${index}`}
           />
         </div>
       );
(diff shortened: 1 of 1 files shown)


packages/react/src/patterns/sections/CardSection/CardSection.js
```diff
@@ -98,8 +99,9 @@ const _renderCards = (cards, containerRef) => (
                 src: ArrowRight20,
               },
             }}
-            target={card.cta.target}
             type="link"
+            role="region"
```
photodow: Maybe this is just my accessibility ignorance showing, but I don't think we want to change the role of these. The role is a "link", and I'm not sure why we're wanting to change that. The `<a>` by default gets a a `Link` role. Based on my understanding we'd be hurting the accessibility of this if we change its role.

I'd personally love more information and reasoning from the QA team on why this is necessary.

packages/react/src/patterns/sections/CardSection/CardSection.js
```diff
@@ -98,8 +99,9 @@ const _renderCards = (cards, containerRef) => (
...
+            aria-labelledby={`region${card.heading}`}
```
photodow: I think your syntax is wrong here. If you're trying to pass a string of text into this to label you need to use `aria-label`. If you're trying to use `aria-labeledby` you need to pass in an `ID`, and that `ID` needs to be attached to something else.

photodow: If we're moving forward with this is this something that's going to continue to be a problem, and should it actually be solved at the `Card` sub-pattern level?

packages/react/src/patterns/sections/CardSection/CardSection.js
mkothur: I have checked with @RaphaelAmadeu before making this change because he also got accessibility issue in `ContentGroupCards` it has `role=region`.

mkothur:
@photodow updated the PR.

packages/react/src/patterns/sections/CardSection/CardSection.js
```diff
@@ -98,8 +102,8 @@ const _renderCards = (cards, containerRef) => (
                 src: ArrowRight20,
               },
             }}
-            target={card.cta.target}
             type="link"
+            id={`region-${card.heading}`}
```
jeffchew: wouldn't card heading potentially have spaces?

mkothur: @jeffchew - card heading has spaces I added hyphen for readable

jeffchew: Having an id with spaces is generally not a good idea though.

jeffchew: I added a uniqueid utility in this PR which might help: https://github.com/carbon-design-system/ibm-dotcom-library/pull/1547

mkothur: yeah @jeffchew but I need to use the  div which contains the card with aria-labelledby. and associate its id with the heading for that particular card.

mkothur:
@jeffchew updated the PR.

packages/react/src/patterns/sections/CardSection/CardSection.js
jeffchew: Right, what I'm suggesting is use the uniqueid utility to generate an id, and use that as the id. Not the card heading, since those can introduce any tyles of characters that would make the id name invalid (special characters, spaces, etc).

mkothur: ok. @jeffchew- I believe the utility is not merged?

jeffchew: Yes, it needs to be reviewed and merged.

PR #1492, opened 2020-02-26
fix(pattern): fix block-level autoid structure

kennylam:
### Related Ticket(s)

#1490 

### Description

This fixes an issue where block-level stable selectors were being repeated.

### Changelog

**Changed**

- block-level stable selectors hoisted to wrap entire pattern
- update `content-block-mixed` css folder and filename


packages/react/src/patterns/sub-patterns/ContentBlock/ContentBlock.js
```diff
@@ -19,7 +19,7 @@ const { stablePrefix } = ddsSettings;
 const { prefix } = settings;
 
 /**
- * ContentArrayBlock Component, for use with content arrays
+ * ContenBlock Sub-pattern
```
jeffchew: @kennylam minor mispelling here

Issue #1477, opened 2020-02-25
Card section with images – images are the wrong ratio


oliviaflory:
#### Describe in detail the issue you're having.
The card images are not using the intended ratio in the Card section with images pattern. 

| Current | Intent |
|---|---|
| 16:9 ? | 4:3|
|![Screen Shot 2020-02-25 at 4.46.24 PM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/9eb5044c-37ba-4b60-ba47-9477b4d14f4d)  |![Screen Shot 2020-02-25 at 4.45.38 PM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/fbfcfbd6-bb67-440e-b7b6-eb3c79d7897b)  |

#### Is this a feature request (new component, new icon), a bug, or a general issue?
Bug

#### Is this issue related to a specific component?
[Card section with images](https://ibmdotcom-react.mybluemix.net/?path=/story/patterns-sections-cardsectionimages--default)

#### What browser are you working in?
Chrome

#### What offering/product do you work on? Any pressing ship or release dates we should be aware of?
Web simplification

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

- Screenshots or code
![Screen Shot 2020-02-25 at 4.40.15 PM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/0f8dcc1c-9bf7-4575-abbf-9242e09c5961)

- Notes

PR #1495, opened 2020-02-26
feat(pattern): added aspect ratio for cardsection images pattern

mkothur:
### Related Ticket(s)

Related issue: #1477 

### Description

Added the extra mixin for card image aspect ratio and wrapped the Image component in another div in Card pattern.

### Changelog

**New**

- new mixin card-image and new modifier image-ratio

**Changed**

- Card jsx
- card-section .scss
image


packages/styles/scss/patterns/sections/card-section/_card-section.scss
```diff
@@ -28,6 +30,20 @@
...
+      padding-top: aspect-ratio(4, 3);
+      position: relative;
+    }
+    &__img {
+      position: absolute;
+      top: 0;
+    }
+  }
+}
+
```
jeffchew: I'm not sure absolutely positioning the image is the best approach here, should these perhaps be a background image instead?

cc: @annawen1

mkothur: @jeffchew the image is being rendered `Image` component  from Crad pattern, and `_image.scss` has object-fit. If I use `background-image` property here how could I pass image src url to this? help me to understand

jeffchew: @annawen1 @andysherman2121 any thoughts on this? I know LeadSpace work is in flight now and the original LeadSpaceCentered used a background image...

photodow: @jeffchew @mkothur I have a pull request over in carbon that's almost ready to be merged with a couple mixins and bug fixes for their current implementation of aspect ratio.

https://github.com/carbon-design-system/carbon/pull/5255/files#diff-986aa34d579220df16c7b49cfda45a2dR252-R265

Here is the crux of the mixin I submitted, and no position absolute is necessary...

```scss
@mixin ratio () {
 display: flex;

 &::before {
  content: '';
  width: 1px;
  margin-left: -1px;
  float: left;
  height: 0;
 }

 &::after {
  content: '';
  display: table;
  clear: both;
 }
}

.#{$prefix}--card-section {
 &__card {
  .#{$prefix}--image {
   @include ratio;

   &--img {
    height: auto;
    flex: 1;
   }
  }
 }
}
```

> Side note, why is the customClassName for the Image component being applied to children elements?

mkothur: yeah @photodow the image is rendering by Image component inside the Card sub-pattern and from where we are passing image prop to card pattern to handle aspect ratio wrapped that Image component in a div inside of Card pattern and added new scss modifier for that in _card-section.scss.

photodow: @mkothur the example code I provided shouldn't need an extra `div`, or any absolute positioning.

packages/react/src/patterns/sub-patterns/Card/Card.js
```diff
@@ -50,7 +50,9 @@ const Card = ({
       )}
       href={cta.href}
       {...props}>
-      <Image {...image} classname={`${prefix}--card__img`} />
+      <div className={`${prefix}--card__image-ratio`}>
```
photodow: You shouldn't need this wrapper.

packages/styles/scss/patterns/sections/card-section/_card-section.scss
mkothur: @photodow I have tried with your above snippet the aspect ratio is coming as expected. my be I am using wrong way.

packages/styles/scss/globals/utils/_ratio-calc.scss
```diff
@@ -0,0 +1,10 @@
+// Ratio calculation
+// function to reuse the ratio-calc
+// @param $width
+//   width of the element
+// @param $height
+//   height of the element
+// @return percentage of height vs width
+@function ratio-calc($width, $height) {
+  @return percentage($height / $width);
```
photodow: I think you copied my example one for one when I was trying to demonstrate some of the logic rather than focusing on being syntactically correct. I don't think you're using the right function here.

https://sass-lang.com/documentation/modules/math

packages/styles/scss/patterns/sections/card-section/_card-section.scss
```diff
@@ -6,7 +6,8 @@
 //
 @import '../../../components/link-with-icon/link-with-icon';
 @import '../../../globals/imports';
-@import '../../../globals/utils/aspect-ratio';
+// @import '../../../globals/utils/aspect-ratio';
+@import '../../../globals/utils/ratio-calc';
```
photodow: If I'm not mistaken your new `ratio-calc` does the same thing the original `aspect-ratio` does.

packages/styles/scss/patterns/sections/card-section/_card-section.scss
```diff
@@ -31,21 +32,49 @@
...
+  &::after {
+    content: '';
+    display: table;
+    clear: both;
   }
 }
 
 @mixin card-section {
   .#{$prefix}--card-section {
+    &__card {
```
photodow: Are you using `bx--card-section__card`? I don't see it in the rendered `HTML`.

mkothur:
@photodow updated the PR.

packages/react/src/patterns/sections/CardSection/CardSection.js
```diff
@@ -87,7 +87,8 @@ const _renderCards = (cards, containerRef, cta) => (
     ref={containerRef}>
     {cards.map((card, index) => {
       return (
-        <div className={`${prefix}--card-section__cards__col`}>
+        <div
+          className={`${prefix}--card-section__cards__col ${prefix}--card-section__card`}>
```
photodow: Small thing, but should `card-section__card` be passed in as a `customClassName` for the `Card`?

packages/styles/scss/patterns/sections/card-section/_card-section.scss
```diff
@@ -43,8 +44,36 @@
...
+@mixin ratio-base() {
+  display: flex;
+
+  &::before {
+    content: '';
+    width: 1px;
+    margin-left: -1px;
+    float: left;
+    height: 0;
+    padding-top: aspect-ratio(4, 3);
```
photodow: Small thing, but this probably shouldn't be apart of this mixin.

packages/styles/scss/patterns/sections/card-section/_card-section.scss
```diff
@@ -43,8 +44,36 @@
   }
 }
 
+@mixin ratio-base() {
```
photodow: I wonder if we should pull this out as a utility until carbon get's their. mixins merged in.

packages/styles/scss/patterns/sections/card-section/_card-section.scss
```diff
@@ -43,8 +44,36 @@
...
 @mixin card-section {
   .#{$prefix}--card-section {
+    &__card {
+      .#{$prefix}--image {
+        @include ratio-base;
+
+        &__img {
+          height: auto;
+          flex: 1;
+        }
```
photodow: I believe the aspect ratio is only on `@lg` breakpoint and higher. I think all other times it has a height of `240px`. Migth just want to double check with @larahanlon2 or someone.

packages/react/src/patterns/sections/CardSection/CardSection.js
photodow: Small thing, but should `card-section__card` be passed in as a `customClassName` for the `Card`?

mkothur: Okay. done

mkothur:
@photodow updated the PR.

packages/styles/scss/patterns/sections/card-section/_card-section.scss
mkothur: sure

packages/styles/scss/patterns/sections/card-section/_card-section.scss
mkothur: @photodow  I did not understand this?

photodow: Looking at what carbon is implementing this aspect is in a seperate mixin. Maybe we shouldn't focus on it, but at the least if you include it in this mixin you should allow the height and width there to be passed in at the mixin level above.

mkothur:
@photodow updated the PR for aspect-ratio.

packages/styles/scss/patterns/sections/card-section/_card-section.scss
```diff
@@ -45,6 +46,20 @@
...
+    &__card {
+      .#{$prefix}--image {
+        &__img {
+          @include carbon--breakpoint('lg') {
+            @include ratio-base(4, 3);
+
+            height: auto;
+          }
+
+          height: 240px;
```
jeffchew: Use `carbon--mini-units(30)`

mkothur: sure.

mkothur: @jeffchew updated the PR.

PR #1547, opened 2020-02-28
feat(uniqueid): adding uniqueid utility

jeffchew:
### Related Ticket(s)

No related issue

### Description

This adds a uniqueid utility in the utilities package

### Changelog

**New**

- uniqueid utility


packages/utilities/src/utilities/uniqueid/uniqueid.js
```diff
@@ -0,0 +1,27 @@
...
+ * import { uniqueid } from '@carbon/ibmdotcom-utilities';
+ *
+ * const id1 = uniqueid(); // id1
+ * const id2 = uniqueid(); // id2
+ * const id3 = uniqueid('prefix'); // prefix3
+ * const id4 = uniqueid('prefix-'); // prefix-4
+ */
+function uniqueid(prefix = 'id') {
+  _lastId++;
+  return `${prefix}${_lastId}`;
```
photodow: Thinking out loud here, but if we start adding generated ids to our components wouldn't that make it hard to re-use them in other situations if the need arose? (e.g. styling, selecting, aria-, and labels just to name a few)

jeffchew: There are certainly situations where you wouldn't want to use this tool, things like react keys, etc. Using it for label for seems like it'd be a good case for this though, especially if you want to add for components that may appear more than once on the page.

jeffchew: That said, this PR is purely for the generic utility itself. We can decide on the situation in the component or pattern if it's a good fit or not.

Issue #1512, opened 2020-02-27
Content group - simple: spacing issues


jacobottesen-dgc:
## Content group - simple spacing issues: 
There are spacing issues on above the first heading, under the image and above the link.

-->

## Detailed description

> Describe in detail the issue you're having.
1. Spacing issues above the first heading. The spacing should be 96px or 6rem not 32px.

<img width="784" alt="Screen Shot 2020-02-27 at 9 08 51 AM" src="https://user-images.githubusercontent.com/56400049/75461105-e5badc00-5947-11ea-9c38-cf0425151e6c.png">

2. Spacing below the images should be 32px not 48px.

<img width="860" alt="Screen Shot 2020-02-27 at 9 07 09 AM" src="https://user-images.githubusercontent.com/56400049/75461278-29154a80-5948-11ea-9687-19b12cbc4618.png">

3. Spacing above the link should be 16px not 32px.

<img width="814" alt="Screen Shot 2020-02-27 at 9 09 56 AM" src="https://user-images.githubusercontent.com/56400049/75461326-3df1de00-5948-11ea-83d5-56ecb91ab625.png">

> Is this a feature request (new component, new icon), a bug, or a general
> issue?
Bug

> Is this issue related to a specific component?
Content group - simple

> What did you expect to happen? What happened instead? What would you like to
> see changed?

> What browser are you working in?
Firefox

## Steps to reproduce the issue

1. Open content group - simple in storybook: https://ibmdotcom-react.mybluemix.net/?path=/story/patterns-blocks-contentgroupsimple--default
2. Right click to inspect 
3. View spacing issues mentioned above.

RobertaJHahn:
@RaphaelAmadeu Can you fix this tomorrow so that it can go into the March release.  Needs to be retested too.

RaphaelAmadeu:
@RobertaJHahn , jsut opened a PR for this issue, as soon it's merged in it will be ready for testing. Thanks!

RobertaJHahn:
@RaphaelAmadeu I think you are supposed to add the assign the people who should do the retesting to the PR.  You should verify this process with Jeff.

PR #1553, opened 2020-02-28
fix(scss): adjusted padding to look as design specs

RaphaelAmadeu:
### Related Ticket(s)

#1512 

### Description

I've adjusted the paddings and margins so the pattern is compliant with the design specs

### Changelog

**Changed**

- Pattern SCSS


photodow:
Sorry @RaphaelAmadeu I think this got solved as a part of #1556

jeffchew:
Closing this per @photodow 's comment.

Issue #1196, opened 2020-01-24
Pattern:  Card Section - Updates per functional specifications


jeffchew:
#### User Story

> As a `[user role below]`:
IBM.com Library user

> I need to:
have a `CardSection` section pattern available

> so that I can:
utilize it within my application

#### Additional information

 - Functional specifications: https://ibm.ent.box.com/file/608969589570
- Content Section story: https://github.com/carbon-design-system/ibm-dotcom-library/issues/1198
- The high contrast card is included in the scope.

#### Acceptance criteria
 - [ ] Update to utilize the `ContentSection`
 - [ ] Update propType definitions (required vs. not required) and any necessary defaultProps
 - [ ] Update prop names per data model
 - [ ] Update all corresponding documentation and Storybook stories

PR #1561, opened 2020-03-02
fix(pattern): validate and do not pass cta into cardsectionimages

mkothur:
### Related Ticket(s)

Related issue: #1196 

### Description

validating the otherProps and making sure don’t pass it into the CardSectionImages component. 

### Changelog

 - remove specific property from object of spread syntax.


packages/react/src/patterns/sections/CardSectionImages/__stories__/CardSectionImages.stories.js
```diff
@@ -20,11 +20,18 @@ storiesOf('Patterns (Sections)|CardSectionImages', module)
...
+      cta: {
+        href: 'https://www.example.com',
+      },
+    };
     return (
       <CardSectionImages
         heading={text('Heading (required):', 'Read more about it')}
         theme={select('theme', themes, themes.white)}
         cards={object('Data', cards.CardSectionImages)}
+        cta={cta}
```
photodow: Shouldn't we not be passing this in here?

packages/react/src/patterns/sections/CardSectionImages/CardSectionImages.js
```diff
@@ -21,7 +21,8 @@ const CardSectionImages = ({ cards, ...otherProps }) => {
     ({ image, eyebrow, heading, copy, cta: { href } }) =>
       image && eyebrow && heading && !copy && href
   );
-  return <CardSection {...otherProps} cards={cardsWithImages} />;
+  const { ['cta']: _, ...newObj } = otherProps;
```
photodow: This is written out in a confusing way. Is there anyreason why we couldn't just use `delete`?

```
delete otherPropse.cta;
return <CardSection {...otherProps} cards={cardsWithImages} />;
```

packages/react/src/patterns/sections/CardSectionImages/__stories__/CardSectionImages.stories.js
mkothur: Ok will remove.

packages/react/src/patterns/sections/CardSectionImages/CardSectionImages.js
mkothur: I also thought the delete operator, but I may get any comments from you on that delete operator , so tried another syntax to remove specific prop from spread operator.

mkothur:
@photodow updated the PR.

jeffchew:
> This works for me. Not sure if @jeffchew or anyone else has concerns about using `delete` here though. I'd love to get their thoughts on it.

I think should be fine.

Issue #1505, opened 2020-02-27
Pattern:- Table of Content - Blank space with scroll indicator is displayed on removing link title from table of content


chetankan:
Issue:- When link title is removed from 'Table of Content' links and  pattern is opened in new tab, here link title is not shown on left side of the page but still scroll indicator is displayed in that removed link place.

## Steps to reproduce the issue

1. Enter table of content url in the browser.
2. In knobs, remove 'Title'.
3. Open the pattern in new tab.

## Additional information

OS:- Windows, Mac
Browsers:- Chrome, Firefox, Safari, Edge

![Table of content without title.jpg](https://images.zenhubusercontent.com/5e21a841639f517d456f7e43/12b50ec3-6db2-4e47-aa57-315131ea3f0f)

PR #1582, opened 2020-03-04
fix(pattern): added validation for menu-items props

RaphaelAmadeu:
### Related Ticket(s)

#1505 

### Description

I've updated the `Proptypes` on `TableOfContents` and coded a new validation function, that prevents `menuItem` entries with blank `id` or `title` from being rendered

### Changelog

**Changed**

- Added new function and edited `Proptypes`


packages/react/src/patterns/sub-patterns/TableOfContents/TableOfContents.js
```diff
@@ -171,6 +171,18 @@ const TableOfContents = ({ menuItems, children, menuLabel, theme }) => {
...
+   *
+   * @param {Array} menuItems array of Items
+   * @returns {Array} filtered array of items
+   */
+  const validateMenuItems = menuItems => {
+    return menuItems.filter(
+      item => item.title.length > 0 && item.id.length > 0
+    );
+  };
+
```
jeffchew: @RaphaelAmadeu I feel like this might be a bit overkill, and we probably if anything would like to lean more towards properly defining the PropTypes, as well as `defaultProps` as much as possible. The warnings would come in through the console, and we're essentially leaning on prop-types to do its thing instead of building extra guard rails like this in all of our patterns.

kennylam: @RaphaelAmadeu I tend to agree with @jeffchew. I've thought about this a lot myself, and I think letting prop-types handle this is probably the way to go for now. Unless I'm missing something and this isn't simply props validation?

RaphaelAmadeu: @jeffchew , @kennylam ,

I thought that only the Proptype to be set required would prevent this from happening, but if an empty string is provided, Proptypes don't point it as an issue.

jeffchew:
@RaphaelAmadeu coming back to this one and thought about it for a bit. I think one solution might be to do a trim on the link title and see if it's blank. If it is, then don't include it in the final menuItems. Thoughts?

cc: @kennylam @annawen1 @reesedownes @andysherman2121

Issue #1478, opened 2020-02-26
Card section with images - minimum height


oliviaflory:
#### Describe in detail the issue you're having.
The minimum height of the card section with images at the small breakpoint is incorrect, the height is too short, and doesn't follow the design specs.

| Current | Intent |
|---|---|
| ![Screen Shot 2020-02-25 at 4.53.58 PM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/ea9f2d07-48fc-495f-ba1a-a4b941e84274)  |  ![Screen Shot 2020-02-25 at 4.54.26 PM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/21cddf11-d0e8-45cf-acd7-2c8998ae425e) |

#### Is this a feature request (new component, new icon), a bug, or a general issue?
design defect / bug

#### Is this issue related to a specific component?
[Card section with images](https://ibmdotcom-react.mybluemix.net/?path=/story/patterns-sections-cardsectionimages--default)

#### What did you expect to happen? What happened instead? What would you like to see changed?
Minimum space between body copy and icon should be 64px

#### What browser are you working in?
Chrome

#### What offering/product do you work on? Any pressing ship or release dates we should be aware of?
Web simplification

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

- Screenshots or code
- Notes

mkothur:
@RobertaJHahn I have opened a [PR](https://github.com/carbon-design-system/ibm-dotcom-library/pull/1584) for this issue but later confirmed with @photodow  it is resolved here [1556](https://github.com/carbon-design-system/ibm-dotcom-library/pull/1556), so closing the issue and PR.

PR #1556, opened 2020-03-02
Spacing Fixes

photodow:
### Related Ticket(s)

This closes #1470, closes #1550, closes #1512, closes #1506, closes #1498, closes #1480, closes #1476, closes #1452, closes #1402, closes #1368, closes #1409, closes #1480, closes #1368, closes #1549, closes #1460, closes #1452, closes #1401, closes #1409, closes #1190, closes #1195, closes #1402, closes #1405, closes #1450, closes #1483, closes #1517, closes #1557 

> Let me know if I missed any issues...

### Description

With the latest release we didn't have a clear picture on how to implement the spacing, as I looked across to document this information I went ahead and put the pull request together since it helped me in visualizing the work. I think I got it all, will need design to review.

### Changelog

**Changed**

- spacing across all patterns
- card `heading` prop to `copy` prop
- cleaned up some storybook content to look closer to designs.

PR #1584, opened 2020-03-04
fix(pattern): updated margin in small breakpoint as per design

mkothur:
### Related Ticket(s)

Related issue: #1478 

### Description
Updated the styles of minimum height of the card section with images at the small breakpoint and also medium and large breakpoint.

### Changelog
**Changed**

  - updated the content section scss because we are using this sub-pattern in CardSection pattern

### DIFF
--- a/packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
+++ b/packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
@@ -30,14 +30,31 @@
       }
     }
     &__heading {
-      margin-top: $carbon--layout-01;
+      margin-top: $carbon--layout-03;
       margin-bottom: $carbon--layout-05;
 
       @include type-style('expressive-heading-02');
+
+      @include carbon--breakpoint('md') {
+        margin-top: $carbon--layout-03;
+      }
+      @include carbon--breakpoint('lg') {
+        margin-top: $carbon--layout-01;
+      }
     }
     &__children {
       @include carbon--breakpoint(lg) {
         @include carbon--make-col(12, 16);
+      }
+    }
+    .#{$prefix}--card {
+      &__eyebrow {
+        margin-top: $carbon--layout-01;
+      }
+      &__heading {
+        @include carbon--breakpoint(lg) {
+          margin-bottom: $carbon--layout-03;
+        }
       }
     }
 
--- a/packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
+++ b/packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
@@ -35,9 +35,6 @@
 
       @include type-style('expressive-heading-02');
 
-      @include carbon--breakpoint('md') {
-        margin-top: $carbon--layout-03;
-      }
       @include carbon--breakpoint('lg') {
         margin-top: $carbon--layout-01;
       }


packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
```diff
@@ -30,16 +30,33 @@
...
+    .#{$prefix}--card {
+      &__eyebrow {
+        margin-top: $carbon--layout-01;
+      }
+      &__heading {
+        @include carbon--breakpoint(lg) {
+          margin-bottom: $carbon--layout-03;
+        }
+      }
+    }
```
photodow: Didn't this get solved in #1556? Have you pulled down the latest yet? Also, I think if this is still an issue we'd want to solve this at the card level most likely.

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
```diff
@@ -30,16 +30,33 @@
...
     &__heading {
-      margin-top: $carbon--layout-01;
+      margin-top: $carbon--layout-03;
       margin-bottom: $carbon--layout-05;
 
       @include type-style('expressive-heading-02');
+
+      @include carbon--breakpoint('md') {
+        margin-top: $carbon--layout-03;
+      }
```
photodow: Is this redundant?

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
```diff
@@ -30,16 +30,33 @@
...
+      @include carbon--breakpoint('lg') {
+        margin-top: $carbon--layout-01;
+      }
```
photodow: Why are you changing all of this? I think you had it correct in the first place. The issue doesn't talk about needing to fix this either.

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
mkothur: yes @photodow I have latest and I got your spacing PR changes , #1478 has spacing issue in small break point, when I added 
`@include carbon--breakpoint('sm') {
  margin-top:  $carbon--layout-03;
}` and it does nothing so made the above changes.

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
mkothur: yeah it is will remove.

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
mkothur: I changed because 'sm' breakpoint does nothing so did like above

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
mkothur: updated

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
photodow: Looking at the latest [here](http://ibmdotcom-react.mybluemix.net/iframe.html?id=patterns-sections-cardsectionimages--default) at the smallest breakpoint it looks correct to me.

image

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
photodow: I'm not sure I understand what you're saying. There is no `sm` breakpoint mixin here. You added `md` and `lg`. You now have too much spacing at the `sm` and `md` breakpoints. It looks like it's supposed to be `32px` and now based on this code (still waiting for it build) it looks like you have `48px`.

packages/styles/scss/patterns/sub-patterns/content-section/_content-section.scss
oliviaflory: I think @photodow is right, the spacing seems to be fixed after the global changes from #1556 
We must have missed closing this one when those were applied?

mkothur: yeah, @oliviaflory I close this PR. thx @photodow

PR #1603, opened 2020-03-05
fix(pattern): fixed the LinkList hover state width to edges

mkothur:
### Description

Fixed the width of LinkList when hover on the CTA item.

### Changelog
**Changed**

 - Added max width of CTA modifier for @lg breakpoint.
Before:
image
Now:
image

### DIFF
--- a/packages/styles/scss/patterns/sub-patterns/link-list/_link-list.scss
+++ b/packages/styles/scss/patterns/sub-patterns/link-list/_link-list.scss
@@ -42,6 +42,8 @@
 
         color: $link-01;
       }
+
+      max-width: none;
     }
 
     @include carbon--breakpoint('md') {


jeffchew:
This is what I'm seeing now:
Screen Shot 2020-03-06 at 11 58 28 AM

Is max width the solution here?

cc: @photodow @wonilsuhibm @larahanlon2 @oliviaflory

mkothur:
> This is what I'm seeing now:
> Screen Shot 2020-03-06 at 11 58 28 AM
> 
> Is max width the solution here?
> 
> cc: @photodow @wonilsuhibm @larahanlon2 @oliviaflory

are seeing this in @lg breakpoint?

jeffchew:
> are seeing this in @lg breakpoint?

This is in smaller than lg. For lg, it seems to be locked at a max width of 320px. Asking the designers, is this the design intent? It is possible that adopters can put the LinkList in a column larger than the right column in the current designs, and in those cases, would we just want the LinkList to take the width of its container?

mkothur:
Okay. right now the Card CTA modifier  has 320px max-width, that was missed when I a made PR for LinkList pattern.

jeffchew:
Closing this, this PR seems to fix the issue: https://github.com/carbon-design-system/ibm-dotcom-library/pull/1675

Issue #1581, opened 2020-03-04
[Layout] Nested layout sub-pattern modifications


photodow:
### The problem
The drupal team is having an issue implementing to the Link list in relation to the content block segmented and simple.

A secondary issue we're having is we have no where exactly for the bottom divider to really live.

> 👀See #1541, #1514

### The solution 
The proposed solution for the moment is to modify the Layout sub-pattern to allow nesting, and a 3 column option. And then an optional dividing line.

### Additional information
 - Designs based on...
    - [Content Block — Simple](https://ibm.box.com/s/5jyo2ml6as6stjctavqivjhhhub6zwwt)
    - [Content Block — Segmented](https://ibm.box.com/s/7hg6ro0zkx7bb3ghut7c4attwbn625yt)
 - Functional specs to come later (👀see #1544)
 - This change is needed urgently by Drupal.  This change has to be included in the v1.5.0 release.  Code Freeze is Friday, March 6, 2020

#### example of layout
![image](https://user-images.githubusercontent.com/3793636/75892399-dd0a5000-5df6-11ea-9fcc-5bd548f84dbc.png)

#### example of divider
![image](https://user-images.githubusercontent.com/3793636/75891904-0ecee700-5df6-11ea-9cde-f3be9dea41ed.png)

### Acceptance criteria

- [ ] 3 column layout (specifically 2 column and 1 column layout [👀see image above])
- [x] Add optional divider to the bottom of the Layout sub-pattern
- [x] Ability to nest inside of the grid (might need to remove container (e.g. `.bx--grid`)
- [ ] Aligns with the 2x grid.

> Reach out if you have any questions!

PR #1606, opened 2020-03-06
fix(pattern): updated layout pattern

RaphaelAmadeu:
### Related Ticket(s)

#1581 

### Description

The Layout pattern was updated with the new 2/3 format, along with the optional bottom border and nested styling

### Changelog

**New**

- Optional border and nested styling


packages/styles/scss/patterns/sub-patterns/layout/_layout.scss
```diff
@@ -28,6 +28,10 @@
         }
       }
     }
+
+    &-border {
```
photodow: Should this be a modifier `--`?

packages/react/src/patterns/sub-patterns/Layout/Layout.js
```diff
@@ -22,29 +22,34 @@ const { prefix } = settings;
  */
 const _types = {
   '1-3': [`${prefix}--col-lg-4`, `${prefix}--col-lg-12`], // 1/4 - 3/4
+  '2-3': [`${prefix}--col-lg-8`, `${prefix}--col-lg-4`], // 2/3 - 1/3
```
photodow: I don't think this is going to align with the grid properly. These classes are based on a 16 column grid, and you can't divide that into 3.

jeffchew: @photodow @RaphaelAmadeu would `2-1` be a more appropriate type?

photodow: Does `Layout` have the ability to be responsive? Do they start off as a single column all the way up to `lg`?

jeffchew:
@RaphaelAmadeu doesn't look like the README has been updated yet?

packages/react/src/patterns/sub-patterns/Layout/Layout.js
photodow: We need a better way to name these the type names here aren't clear to me... `4-12`, `2-1` maybe? It also doesn't account for any differences in responsiveness.

RaphaelAmadeu: I agree, I've just changed the naming on the latest commit

packages/react/src/patterns/sub-patterns/Layout/Layout.js
```diff
@@ -65,7 +70,10 @@ function _updateChild(type, stickyOffset, children) {
         React.cloneElement(child, {
           className: classnames(
             child.props.className,
-            _types[type] && _types[type][i] ? _types[type][i] : `${prefix}--col`
+            _types[type] && _types[type][i]
+              ? _types[type][i]
+              : `${prefix}--col`,
+            border ? `${prefix}--layout-border` : ''
```
photodow: Should this be applied to the row instead?

packages/react/src/patterns/sub-patterns/Layout/Layout.js
```diff
@@ -22,29 +22,34 @@ const { prefix } = settings;
...
  * @param {object} children child nodes
  * @returns {*} modified child content
  * @private
  */
-function _updateChild(type, stickyOffset, children) {
+function _updateChild(type, stickyOffset, border, children) {
   const final = [];
 
-  children.map((child, i) => {
+  const componentArr = type === '2-1' ? [children[1], children[0]] : children;
```
photodow: Can you help me understand what's happening here?

packages/react/src/patterns/sub-patterns/Layout/__stories__/Layout.stories.js
```diff
@@ -44,111 +47,115 @@ storiesOf('Patterns (Sub-Patterns)|Layout', module)
...
-            layoutOptions,
-            layoutOptions['layout-06']
-          )}
-          stickyOffset={number('Sticky offset (in pixels)', 0)}>
-          <div
-            data-sticky={select(
-              'Sticky left column',
-              ['true', 'false'],
-              'true'
+        <div className={nested ? 'bx--grid' : ''}>
```
photodow: To test whether it nests properly it needs to fit inside of a `bx--grid` `bx--row` and some sort of `bx--col` like `bx--col-lg-12`.

packages/react/src/patterns/sub-patterns/Layout/Layout.js
RaphaelAmadeu: I've implemented a new proportion based styling

packages/react/src/patterns/sub-patterns/Layout/Layout.js
RaphaelAmadeu: I've changed this to be applied to the row.

packages/react/src/patterns/sub-patterns/Layout/Layout.js
RaphaelAmadeu: The `Layout` uses the fist component as the size navigation, as the new proportion uses it on the right size, this ensures that the components are ordered properly.

packages/styles/scss/patterns/sub-patterns/layout/_layout.scss
```diff
@@ -28,6 +28,20 @@
...
+    &--border {
+      border-bottom: solid 1px $ui-04;
+    }
+
+    &-2-3 {
+      flex: 0 0 66.66%;
+      max-width: 66.66%;
+    }
+
+    &1-3 {
```
jeffchew: Does this need a dash here?

packages/react/src/patterns/sub-patterns/Layout/README.md
```diff
@@ -54,6 +54,7 @@ ReactDOM.render(<App />, document.querySelector('#app'));
 | Name  | Description             |
 | ----- | ----------------------- |
 | `1-3` | 1/4 - 3/4 column layout |
+| `2-1` | 2/3 - 1/3 column layout |
```
jeffchew: `border` and `nested` need to be documented as well, right?

packages/react/src/patterns/sub-patterns/Layout/Layout.js
jeffchew: I think `Layout` doesn't really care what column is what, it's pretty generic. We can probably keep the order in which the columns come in, and Drupal team will need to do something like this:

```
<Layout>
  <div><ContentBlockSegment /></div>
  <div><LinkList /></div>
</Layout>
```

packages/styles/scss/patterns/sub-patterns/layout/_layout.scss
```diff
@@ -28,6 +28,20 @@
...
+
+    &-2-3 {
+      flex: 0 0 66.66%;
+      max-width: 66.66%;
+    }
+
+    &-1-3 {
+      flex: 0 0 33.33%;
+      max-width: 33.33%;
+    }
```
photodow: I'd use carbon's grid mixins here. 

```scss
@include carbon--make-col(2, 3);
@include carbon--make-col(1, 3);
```

packages/styles/scss/patterns/sub-patterns/layout/_layout.scss
```diff
@@ -28,6 +28,20 @@
         }
       }
     }
+
+    &--border {
+      border-bottom: solid 1px $ui-04;
+    }
+
+    &-2-3 {
```
photodow: How is the responsive being handled. Is it 3 all the way down?

packages/react/src/patterns/sub-patterns/Layout/__stories__/Layout.stories.js
RaphaelAmadeu: currently, it is inside a `bx--col-lg-12`

packages/styles/scss/patterns/sub-patterns/layout/_layout.scss
```diff
@@ -10,9 +10,11 @@
 @mixin layout {
   .#{$prefix}--layout {
     &--sticky {
-      position: sticky;
-      top: 0;
-      z-index: 1;
+      @include carbon--breakpoint(lg) {
+        position: sticky;
```
jeffchew: @RaphaelAmadeu mobile needs to be sticky as well.

jeffchew:
Checking out the deploy preview, the desktop version seems to be pushing to the right:

Screen Shot 2020-03-06 at 12 57 01 PM

packages/react/src/patterns/sub-patterns/Layout/__stories__/Layout.stories.js
```diff
@@ -78,78 +85,124 @@ storiesOf('Patterns (Sub-Patterns)|Layout', module)
...
-              habitasse. Interdum velit laoreet id donec ultrices tincidunt arcu
-              non sodales. Interdum posuere lorem ipsum dolor sit. Quam quisque
-              id diam vel quam elementum pulvinar etiam non. Massa enim nec dui
-              nunc mattis enim. Ac tortor vitae purus faucibus ornare
-              suspendisse sed. Non pulvinar neque laoreet suspendisse.
-            </p>
+
+          <div className="bx--grid">
+            <div className="bx--row">
+              <div className="bx--col">
```
photodow: This is already included with the parent `Layout` above. Not needed for demonstration purposes.

packages/react/src/patterns/sub-patterns/Layout/__stories__/Layout.stories.js
```diff
@@ -78,78 +85,124 @@ storiesOf('Patterns (Sub-Patterns)|Layout', module)
...
+              </div>
+            </div>
           </div>
         </Layout>
 
+        <div className="bx--grid">
+          <div className="bx--row">
+            <div className="bx--col-lg-12 bx--col-sm-4 bx--offset-lg-4"></div>
+          </div>
+        </div>
```
photodow: What is this doing down here?

packages/react/src/patterns/sub-patterns/Layout/__stories__/Layout.stories.js
```diff
@@ -78,78 +85,124 @@ storiesOf('Patterns (Sub-Patterns)|Layout', module)
...
-              suspendisse sed. Non pulvinar neque laoreet suspendisse.
-            </p>
+
+          <div className="bx--grid">
+            <div className="bx--row">
+              <div className="bx--col">
+                <Layout
+                  type={'2-1'}
+                  stickyOffset={number('Sticky offset (in pixels)', 0)}
+                  border={boolean('Optional border 2-1:', false)}
```
photodow: I think the knobs just need to be cleaned up a little.

packages/react/src/patterns/sub-patterns/Layout/__stories__/Layout.stories.js
RaphaelAmadeu: This was removed in the latest commit

packages/react/src/patterns/sub-patterns/Layout/Layout.js
```diff
@@ -116,7 +116,7 @@ const Layout = ({
   <section
     data-autoid={`${stablePrefix}--layout`}
     className={classnames(
-      nested ? '' : `${prefix}--grid`,
+      nested ? `${prefix}--col` : `${prefix}--grid`,
```
photodow: Why did you make this change? This looked correct before you added `${prefix--col}`. This whole pattern has to sit inside of a `--col`, not be a `--col`.

packages/react/src/patterns/sub-patterns/Layout/__stories__/Layout.stories.js
```diff
@@ -86,123 +77,101 @@ storiesOf('Patterns (Sub-Patterns)|Layout', module)
...
-                      <li className="bx--list__item">Item 3</li>
-                      <li className="bx--list__item">Item 4</li>
-                      <li className="bx--list__item">Item 5</li>
-                      <li className="bx--list__item">Item 6</li>
-                      <li className="bx--list__item">Item 7</li>
-                    </ul>
-                  </div>
-                </Layout>
-              </div>
+          <Layout type={'2-1'} nested={true}>
```
photodow: This is where the optional border needs to go.

packages/react/src/patterns/sub-patterns/Layout/__stories__/Layout.stories.js
```diff
@@ -56,7 +52,9 @@ storiesOf('Patterns (Sub-Patterns)|Layout', module)
             layoutOptions,
             layoutOptions['layout-06']
           )}
-          stickyOffset={number('Sticky offset (in pixels)', 0)}>
+          stickyOffset={number('Sticky offset (in pixels)', 0)}
+          border={boolean('Optional border:', false)}
+          nested={true}>
```
photodow: This should not be nested based on my understanding of what you're doing here because you're not nesting it anywhere in the grid. it stands on its own.

packages/react/src/patterns/sub-patterns/Layout/__stories__/Layout.stories.js
```diff
@@ -78,76 +76,100 @@ storiesOf('Patterns (Sub-Patterns)|Layout', module)
...
+            </div>
+            <div
+              data-sticky={select(
+                'Sticky right column',
+                ['true', 'false'],
+                'true'
+              )}
+              style={{
+                backgroundColor: 'white',
+              }}>
```
photodow: Why are we setting the background color here?

packages/react/src/patterns/sub-patterns/Layout/__stories__/Layout.stories.js
```diff
@@ -78,76 +76,100 @@ storiesOf('Patterns (Sub-Patterns)|Layout', module)
...
-              pellentesque elit. Est ultricies integer quis auctor elit sed.
-              Quis risus sed vulputate odio ut. Varius sit amet mattis
-              vulputate. Enim ut tellus elementum sagittis vitae et leo. Posuere
-              ac ut consequat semper viverra nam libero. Habitant morbi
-              tristique senectus et netus et malesuada fames. Enim neque
-              volutpat ac tincidunt vitae semper quis lectus nulla.
-            </p>
+          <Layout type={'2-1'} nested={true}>
+            <div>
+              <h3>Column 2</h3>
```
photodow: Just a thought... `Column 2.1`

packages/react/src/patterns/sub-patterns/Layout/__stories__/Layout.stories.js
```diff
@@ -78,76 +76,100 @@ storiesOf('Patterns (Sub-Patterns)|Layout', module)
...
+            <div
+              data-sticky={select(
+                'Sticky right column',
+                ['true', 'false'],
+                'true'
+              )}
+              style={{
+                backgroundColor: 'white',
+              }}>
+              <h3>Column 1</h3>
```
photodow: continuing that thought `Column 2.2`

packages/react/src/patterns/sub-patterns/Layout/__stories__/Layout.stories.js
jeffchew: I think because in mobile the content starts to overlap and looks a little weird. This is more on the story level.

packages/react/src/patterns/sub-patterns/Layout/__stories__/Layout.stories.js
RaphaelAmadeu: I removed this on the latest version

packages/react/src/patterns/sub-patterns/Layout/__stories__/Layout.stories.js
RaphaelAmadeu: I applied this on the latest version

Issue #1621, opened 2020-03-09
Card Section Simple has a generic data-autoid


kexuibm:
Card Section Simple has a generic data-autoid

## Detailed description
Using, both v1.4.1 and v1.5.0-rc.0, the Card Section Simple Pattern renders what appears to be a generic data-autoid:

    <section class="bx--content-section bx--card-section" data-autoid="**dds--content-section**">
Can this be fixed to use a unique autoid? (e.g dds--card-section-simple-section)

PR #1625, opened 2020-03-09
fix(autoid): make autoids more specific for card section

annawen1:
### Related Ticket(s)

Card Section Simple has a generic data-autoid #1621

### Description

make autoid for card section patterns more specific.

Screen Shot 2020-03-09 at 12 34 16 PM

### Changelog

**Changed**

- pass `autoid` prop to `CardSection` and `ContentSection` components

### DIFF
--- a/packages/react/src/patterns/sections/CardSection/CardSection.js
+++ b/packages/react/src/patterns/sections/CardSection/CardSection.js
@@ -28,7 +28,7 @@ const { prefix } = settings;
  * @param {Array} props.cards Array of card
  * @returns {object} JSX Object
  */
-const CardSection = ({ heading, theme, cards, cta }) => {
+const CardSection = ({ heading, theme, cards, cta, ...otherProps }) => {
   const containerRef = useRef();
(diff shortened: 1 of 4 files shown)


packages/react/src/patterns/sections/CardSection/CardSection.js
```diff
@@ -67,6 +67,7 @@ const CardSection = ({ heading, theme, cards, cta }) => {
   return (
     <ContentSection
       heading={heading}
+      autoid={otherProps.autoid}
```
jeffchew: Approved, but I'm wondering if `autoid` should be a first class prop for the `CardSection`, but default as `null` under `defaultProps`?

jeffchew: Nevermind, I think we'll get the same effect with this implementation, right?

annawen1: yep it'll default to `null` anyways

Issue #1562, opened 2020-03-03
Pattern:- Card- On setting aspect ratio to 2:1, 4:3, and 16:9 title and copy text goes out of card space.


chetankan:
Issue:- In card patter when aspect ratio is set to 2:1 or 4:3 or 16:9 in that case if title and copy text is displayed out of card space.

## Steps to reproduce the issue

1. Enter the url in the browser
2. Open card pattern(Static or Link)
3. Add some extra text in title and copy section
4. Set aspect ratio to 2:1 or 4:3 or 16:9 
5. Open pattern in new page.

OS:- Windows 10, Mac, iOS, Android 
Browsers:- Chrome, Firefox, Safari, UC, Samsung.![Card with 2-1 aspect ratio.png](https://images.zenhubusercontent.com/5e21a841639f517d456f7e43/80f6d15b-c515-43b3-a30a-b767a394aa87)![Card with image in 2-1 ratio.png](https://images.zenhubusercontent.com/5e21a841639f517d456f7e43/273b66f8-c01a-4914-874d-0afb0d019075)

RobertaJHahn:
This issue is a blocker for the QA team regression testing.   Must be fixed by EOD Friday, 3/6.

RobertaJHahn:
Keeping this in v1.5.0 to hopefully patch the fix in next week.

jeffchew:
Hey @RaphaelAmadeu do you have time to get this in before the next release candidate? @kennylam took a brief look and thinks might be a simple fix, you can reach out to him with any questions.

cc: @RobertaJHahn @ljcarot

jeffchew:
@RobertaJHahn going to remove this from `v1.5.0` as the engineers discussed and isn't an issue with the Card itself but for the storybook story. 

cc: @RaphaelAmadeu @kennylam @photodow @annawen1 @andysherman2121

RobertaJHahn:
@jeffchew I have added it to the April release.

PR #1649, opened 2020-03-10
fix(pattern): fixed aspect-ratio on card pattern;

RaphaelAmadeu:
### Related Ticket(s)

#1562 
### Description

Aspect handling was changed from the card to the image level, so the content fits in the card regardless of the selected aspect.

### Changelog

**New**

- imageAspect prop

### DIFF
--- a/packages/styles/scss/patterns/sub-patterns/card/index.scss
+++ b/packages/styles/scss/patterns/sub-patterns/card/index.scss
@@ -8,6 +8,7 @@
 @import '../../../globals/imports';
 @import '../../../temp-carbon-expressive/temp-carbon-expressive';
 @import '@carbon/motion/scss/motion.scss';
+@import '../../../globals/utils/ratio-base';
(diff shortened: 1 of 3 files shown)


packages/styles/scss/patterns/sub-patterns/card/index.scss
```diff
@@ -124,6 +125,28 @@
...
+    }
+
+    &4x3 {
+      @include ratio-base(4, 3);
+    }
+
+    &1x1 {
+      @include ratio-base(1, 1);
+    }
+  }
```
jeffchew: Just checking, it looks like we're adding these as options to the card itself. We're good with this? 
cc: @photodow @kennylam @annawen1 @andysherman2121

andysherman2121: I don't see a problem with adding those options

Issue #1652, opened 2020-03-11
Learn Page: iOS 12 - Jump to option is not stuck to top of the page.


Praveen-sr:
## Detailed description
Devices - iPhone and iPad

Issue :- On iOS 12 version - Jump to option doesn't stick to the top of the page. On scroll of the page it isn't seen.
Note: Works as expected on others devices and on storybook.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/709ed32f-b3f2-42bb-8681-5250fd8e7218)

PR #1660, opened 2020-03-11
Fix/mobile toc sticky

kennylam:
### Related Ticket(s)

#1652 

### Description

Adds `-webkit` vendor prefix for `position: sticky`, which is required for iOS 12 Safari.

Mar-11-2020 09-41-17

### Changelog

**New**

- adds `autoprefixer` plugin to `NextJS` webpack config

**Changed**

- `.bx--layout--sticky` now uses `vendor-prefix`util for `position: sticky`

**Removed**
- `vendor-prefix` util mixin


packages/styles/scss/globals/utils/_vendor-prefix.scss
```diff
@@ -0,0 +1,22 @@
...
+/// @param {List} $prefixes (()) - List of prefixes to print
+/// @example scss
+/// .selector {
+///   @include prefix((
+///     column-count: 3,
+///     column-gap: 1.5em,
+///     column-rule: 2px solid hotpink
+///   ), webkit moz);
+/// }
+@mixin prefix($declarations, $prefixes: ()) {
```
jeffchew: @kennylam I'm, wondering if we should introduce an autoprefixer in gulp instead?

jeffchew: @kennylam actually it looks like we're already using autoprefixer, but I'm wondering if it's how it's configured that we're missing iOS12:

https://github.com/carbon-design-system/ibm-dotcom-library/blob/989093d00ab2ca8f723d4b9530e3162e2d7d0481/packages/styles/gulp/tasks/sass.js#L33

kennylam: @jeffchew I considered this also but decided against it because I feel like we should allow this to be on a as-needed basis. The latest browsers no longer need vendor prefixes at all and we probably don't need to add prefixes to everything by default. What do you think?

jeffchew: well at least for iOS Safari 12, I'm seeing why it's possibly causing an issue:

https://browserl.ist/?q=%3E+1%25%2C+Last+2+versions

Based on this, Safari on iOS 12.0-12.1 isn't included, and would probably see this issue. Should verify what version of safari is being used in browserstack.

But I hear you, it's a tactical fix just for this instance too so the risk is super low. I think I'm ok with pushing this specific fix.

jeffchew: As discussed, @kennylam is looking into adding autoprefixer in storybook's webpack configuration to see if that fixes the issue.

jeffchew:
Closing this out after speaking with @kennylam , the NextJS app has been moved to https://github.com/carbon-design-system/ibm-dotcom-library-nextjs-test and the change is already been added there.

Issue #1108, opened 2020-01-20
Create an inline embedded (default) version of the video player


RobertaJHahn:
#### User Story

> As a `[user role below]`:
IBM.com Library adopter developer

> I need to:
have an inline embedded version of the video player  

> so that I can:
include it in the ibm.com web pages I am creating

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- Functional specs issue (#1066  )
- Design QA testing issue (#1349)
- Prod QA testing issue (#1143)

#### Acceptance criteria
 - [x] Include README for the react pattern and corresponding styles
 - [x] Minimum 80% unit test coverage
- [x] A comment is posted in the Design QA issue, tagging Wonil and Roberta, when development is finished
 - [x] A comment is posted in the Prod QA issue, tagging Praveen and Chetan, when development is finished

PR #1699, opened 2020-03-13
Feat/video player

kennylam:
### Related Ticket(s)

#1108

### Description

Adds the embedded `VideoPlayer` component, utilizing Kaltura's video platform.

### Changelog

**New**

- Kaltura `VideoPlayer` component


kennylam:
> 
> 
> Approved, looks good! Though one thing, I noticed there are no knobs for this. Should probably add for at least the `showDescription` flag, and maybe it's worth opening up for the video ID as well so people can test their own videos

@jeffchew Good call, I'll add a knob for `showDescription`. The `videoId` is a little trickier, since it doesn't always render with any valid ID you pass, not sure why yet. Using a valid video ID can still return an embed saying "Video not found". Maybe tied to `partnerID`?

jeffchew:
> > Approved, looks good! Though one thing, I noticed there are no knobs for this. Should probably add for at least the `showDescription` flag, and maybe it's worth opening up for the video ID as well so people can test their own videos
> 
> @jeffchew Good call, I'll add a knob for `showDescription`. The `videoId` is a little trickier, since it doesn't always render with any valid ID you pass, not sure why yet. Using a valid video ID can still return an embed saying "Video not found". Maybe tied to `partnerID`?

Hey @kennylam one minor thing, when the video is loading I noticed that you see a `NaN: NaN` flash before the video itself loads. We probably should suppress the description from loading until the video is loaded.

kennylam:
> Hey @kennylam one minor thing, when the video is loading I noticed that you see a `NaN: NaN` flash before the video itself loads. We probably should suppress the description from loading until the video is loaded.

@jeffchew Good catch. I added a check to only return when `msDuration` exists. It tries to call it in the hook even before `videoData` is done so returns `NaN` the first time round.

PR #1710, opened 2020-03-16
"fix(leadspace): optional buttons"

andysherman2121:
### Related Ticket(s)

https://github.com/carbon-design-system/ibm-dotcom-library/issues/1568

### Description

Adds optional button feature to LeadSpace


andysherman2121:
Spoke with @larahanlon2 about changing Button Copy from "Primary/Secondary" to "Button1/2" to match how the current Button Group component works

larahanlon2:
Looks good. Thanks! 

Once more than 6 buttons are added, it starts to break but we will never want that to happen so we will handle that in the documentation, CMS, and design guidelines for now.

PR #1729, opened 2020-03-18
Fix/issue 1663 masthead

reesedownes:
### Related Ticket(s)

https://app.zenhub.com/workspaces/ibmcom-library-5d449f3642eb1962336cbe52/issues/carbon-design-system/ibm-dotcom-library/1663

### Description

Search icon is misaligned, because the styling targets all svgs in that masthead to be pushed down a few pixels. Targeting specifically the search icon by using the data-autoid as a selector.

misaligned-icon

### Changelog

**New**

- added the following to _masthead.scss
[data-autoid$='header__search--search'] {
   svg {
     top: 0;
   }
}


jeffchew:
@reesedownes make sure to add the label `patterns: react` so the deploy preview is generated for Storybook React.

packages/styles/scss/components/masthead/_masthead.scss
```diff
@@ -44,6 +44,12 @@ $search-transition-timing: 95ms;
       left: 0;
       background-color: $ui-03;
     }
+
+    [data-autoid$='header__search--search'] {
+      svg {
+        top: 0;
+      }
+    }
```
kennylam: @reesedownes There's already a class for the search icon here where we can add the fixes we need. I don't think we should be using `autoid` in the css.

https://github.com/carbon-design-system/ibm-dotcom-library/blob/dd055e17f8f189765e9d4b376600fd50c7f57ea0/packages/styles/scss/components/masthead/_masthead-search.scss#L28-L32

reesedownes: @kennylam i agree, but i couldn't find how else to target it before. made the change and i am putting in another pull request.

Issue #1206, opened 2020-01-27
Pattern: Using special characters (&, ",', <>) in 'Title' and 'Copy',  characters are shown as there hexadecimal values.


Praveen-sr:
## Detailed description

Device:- Windows, Mac
Browser:- Chrome, Safari

Issue :- When we enter special characters in Title and copy, special characters are shown as there hexadecimal values.  
Expected :- Only the Special characters needs to shown.

This is not happening in Drupal.  This is only happening for adopters directly using the React patterns.

## Steps to reproduce the issue

1. Open the browser ( Firefox, chrome or Safari)
2. Open the Storybook URL on the browser
3. Select Lead Space Centered section
4. Edit knobs and enter special character values for Title, Copy
4. Verify the entered special characters view on the lead Space section.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/f000d760-d62d-4c24-b3db-739f9b943bb2)

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/764e0e9f-b7b7-4866-aa42-5c7ad2c3f92c)

RobertaJHahn:
Wonil, heard from Jill that the patterns must allow special characters.

RaphaelAmadeu:
This is only happening in storybook, due to the way it handles strings,
After testing it on an isolated environment, the usage of the HTML entities is working properly.

As we can see in this example I made:
https://codesandbox.io/s/cocky-napier-7t40y,

@RobertaJHahn I think we can close this issue,
Thanks!

PR #1734, opened 2020-03-19
fix(pattern): html entities now work in leadspace pattern

RaphaelAmadeu:
### Related Ticket(s)

#1206 

### Description

I've updated the copy and h1 tags for the `Leadspace` pattern, so it now can use HTML entities as part of the text.

### Changelog

**New**

- HTML entity support added


annawen1:
I don't know if we want to allow `html` for the title and copy, we just want to make sure that the special characters aren't escaped - I think it's more so happening from storybook than the actual pattern

RaphaelAmadeu:
> I don't know if we want to allow `html` for the title and copy, we just want to make sure that the special characters aren't escaped - I think it's more so happening from storybook than the actual pattern

I agree, maybe we would want just to document that the entities will not work on storybook

Issue #1741, opened 2020-03-20
Update the DDS documentation with the workaround for the missing CSS styles


RobertaJHahn:
Bug (#1676) was created by Ke about missing layout and link-list syles.  A fix is needed in Carbon.  The issue is to update the DDS documentation with the work around so that it is clear until the final solution is available from Carbon.

PR #1750, opened 2020-03-23
chore(docs): updated docs with scss issue workaround

RaphaelAmadeu:
### Related Ticket(s)

#1741 

### Description

I've updated the `README` files, so they include the fix for the scss import issue

### DIFF
--- a/packages/styles/README.md
+++ b/packages/styles/README.md
@@ -33,8 +33,9 @@ application:
 import '@carbon/ibmdotcom-styles';
 ```
 
-For Carbon component styles to properly import, the following should be added to
-your environment variables:
+💡 There may be times pathing errors are encountered when importing certain
+stylesheets. We're working hard to fix these, but in the meantime you can add
+the following to your .env file to resolve:
 
 ```
 SASS_PATH=node_modules:src
--- a/packages/styles/README.md
+++ b/packages/styles/README.md
@@ -33,6 +33,13 @@ application:
 import '@carbon/ibmdotcom-styles';
 ```
 
+For Carbon component styles to properly import, the following should be added to
+your environment variables:
+
+```
+SASS_PATH=node_modules:src
+```
+
 ## Expressive Theme
 
 In addition to styles for IBM.com Library components and patterns, this package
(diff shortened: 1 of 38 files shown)


packages/styles/README.md
```diff
@@ -33,6 +33,13 @@ application:
 import '@carbon/ibmdotcom-styles';
 ```
 
+For Carbon component styles to properly import, the following should be added to
+your environment variables:
+
+```
+SASS_PATH=node_modules:src
+```
```
jeffchew: @kennylam wondering if this is enough context for adopters or if we should provide a more detailed explanation around this, and if this is an issue on the Carbon side or not that we're trying to do a workaround for?

cc: @RaphaelAmadeu

kennylam: @jeffchew @RaphaelAmadeu Yes, agree a little more detail could be helpful.
 maybe something like 
```suggestion
> 💡 There may be times pathing errors are encountered when importing certain stylesheets. We're working hard to fix
> these, but in the meantime you can add the following to your `.env` file to resolve.

```
SASS_PATH=node_modules:src
```
```

RaphaelAmadeu: @kennylam Thanks for the explanation text, I've added it to my latest PR

packages/styles/README.md.rej
```diff
@@ -0,0 +1,6 @@
+diff a/packages/styles/README.md b/packages/styles/README.md	(rejected hunks)
+@@ -36 +36,3 @@ import '@carbon/ibmdotcom-styles';
+-💡 There may be times pathing errors are encountered when importing certain stylesheets. We're working hard to fix these, but in the meantime you can add the following to your .env file to resolve:
++💡 There may be times pathing errors are encountered when importing certain
++stylesheets. We're working hard to fix these, but in the meantime you can add
++the following to your .env file to resolve:
```
jeffchew: Hey @RaphaelAmadeu , what is this file?

RaphaelAmadeu: @jeffchew , I've removed it

Issue #1619, opened 2020-03-09
Text length in cards should be 90% of the available space


wonilsuhibm:
When text is in a container, they should span only 90% of the available space instead of the full width. See the comment below.

wonilsuhibm:
![Screen Shot 2020-03-09 at 10 03 16 AM](https://user-images.githubusercontent.com/30183447/76220080-40c4bc80-61ed-11ea-9a78-dba0e41bde46.png)

PR #1779, opened 2020-03-24
fix(scss): added max-width to card content

RaphaelAmadeu:
### Related Ticket(s)

#1619 

### Description

I've added the max-width to the `Card` pattern scss, so it better fits the design specs.

### Changelog

**New**

- Max-width added

### DIFF
--- a/packages/styles/scss/patterns/sub-patterns/card/index.scss
+++ b/packages/styles/scss/patterns/sub-patterns/card/index.scss
@@ -25,6 +25,7 @@
       flex-direction: column;
       padding: $carbon--spacing-05;
       min-height: carbon--mini-units(20);
+      max-width: 90%;
     }
 
     &__heading {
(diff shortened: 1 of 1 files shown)


packages/styles/scss/patterns/sub-patterns/card/index.scss
```diff
@@ -25,6 +25,7 @@
       flex-direction: column;
       padding: $carbon--spacing-05;
       min-height: carbon--mini-units(20);
+      max-width: 90%;
```
photodow: Don't we have a utility for this? Also, I would add this to the copy and heading instead of the wrapper.

kennylam: @photodow do you mean `max-width`? We have no utility for that, that I know of.

Issue #1111, opened 2020-01-20
Create the link version of the video component that opens into the Lightbox media viewer


RobertaJHahn:
#### User Story

> As a `[user role below]`:
IBM.com Library adopter developer

> I need to:
have a link version of the video player  

> so that I can:
include it in the ibm.com web pages I am creating

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- Functional specs issue (#1066  )
- [Functional specs](https://ibm.ent.box.com/file/600045806474):  details documented in section 2.2
- [CTA functional specs](https://ibm.box.com/s/ceu99rjzd0e87sm51n9cv550yssockme)
- Design QA testing issue (#1640)
- Prod QA testing issue (#1641)

#### Acceptance criteria
 - [ ] create a feature flag for this component
 - [ ] Include README for the react pattern and corresponding styles
 - [ ] Additional video type added to CTA.
- [ ] A comment is posted in the Design QA issue, tagging Wonil and Roberta, when development is finished
 - [ ] A comment is posted in the Prod QA issue, tagging Praveen and Chetan, when development is finished

RobertaJHahn:
This story will be applying the work to the CTA component.  The functional specs will be updated.

[CTA functional specs](https://ibm.ent.box.com/file/603914748878)

PR #1865, opened 2020-03-26
Feat(cta): render lightbox video player to ctas

annawen1:
### Related Ticket(s)

Create the link version of the video component that opens into the Lightbox media viewer #1111

### Description

add `video` type to the `CTA` and render lightbox when clicked.

Mar-30-2020 14-01-22

### Changelog

**Changed**

- call videoplayer api in `CTA` component 
- add `PlayOutline20` icon to be used for `video` types


packages/react/src/components/CTA/CTA.js
```diff
@@ -26,11 +33,46 @@ const { prefix } = settings;
...
+const CTA = ({ style, type, customClassName, ...otherProps }) => {
+  const [renderLightBox, openLightBox] = useState(false);
+  const [videoTitle, setVideoTitle] = useState([{ title: '', key: 0 }]);
+  const [mediaData, setMediaData] = useState({});
+
+  useEffect(() => {
+    video();
+  }, [video, style, type]);
+
+  const video = useCallback(async () => {
```
jeffchew: The const `video` seems pretty generic, this method seems to be making the call to get the meta data then setting the title. Perhaps rename to `getVideoData` or something along those lines?

packages/react/src/components/CTA/CTA.js
```diff
@@ -136,6 +347,7 @@ const _renderButtons = ({ buttons }) =>
  * @returns {*} object
  */
 const _renderFeatureCard = featureCard => {
+  if (featureCard.type === 'video') featureCard.cta.href = '#';
```
jeffchew: Would this run into issues if there were actual anchor links on the page? Like if the link was inside of the Table of Contents?

annawen1:
> @annawen1 had an overall question regarding the anchor link being set as `#`, I think that might pose issues on pages with anchor links and can take users to the top of the page unintentionally?
> 
> Also, because CTA is starting to look a little loaded, I'm wondering if it's worth thinking about breaking down CTA and have sub components with corresponding type handlers in their own subcomponent files within CTA?

@jeffchew yeah I was setting the hrefs to `#` because if I don't pass anything, the links won't render clickable. 

And yep I was talking to @kennylam earlier and I think we should probably go back in to refactor since it seems like it's only going to get more complicated

packages/react/src/components/CTA/CTA.js
annawen1: yep gotcha

jeffchew:
> @jeffchew yeah I was setting the hrefs to `#` because if I don't pass anything, the links won't render clickable.

Is there another solution we could use?
 
> And yep I was talking to @kennylam earlier and I think we should probably go back in to refactor since it seems like it's only going to get more complicated

How much time would that take?

jeffchew:
> > And yep I was talking to @kennylam earlier and I think we should probably go back in to refactor since it seems like it's only going to get more complicated
> 
> How much time would that take?
> 
> @jeffchew I can open another branch based off of this one - I'm thinking a day or two to separate things up?

Wish we could do threads on the main comments. I'm fine with having it in a separate PR. The other question about the anchor links, I doublechecked and it looks like the `preventDefault`'s you have should prevent the anchor jump behavior from what I can tell...

packages/react/src/components/CTA/CTA.js
jeffchew: Resolving, looks like `preventDefault` will take care of this.

photodow:
> Also, because CTA is starting to look a little loaded, I'm wondering if it's worth thinking about breaking down CTA and have sub components with corresponding type handlers in their own subcomponent files within CTA?

I can't think of any issues off hand with having sub-components. I think the important thing to maintain are the types and styles, so we don't have to repeat that logic everywhere.

packages/react/src/components/CTA/CTA.js
```diff
@@ -26,11 +33,46 @@ const { prefix } = settings;
...
+const CTA = ({ style, type, customClassName, ...otherProps }) => {
+  const [renderLightBox, openLightBox] = useState(false);
+  const [videoTitle, setVideoTitle] = useState([{ title: '', key: 0 }]);
+  const [mediaData, setMediaData] = useState({});
+
+  useEffect(() => {
+    getVideoData();
+  }, [getVideoData, style, type]);
+
+  const getVideoData = useCallback(async () => {
```
jeffchew: Minor thing, this could use JSDocs.

annawen1: coming right up

PR #1879, opened 2020-03-26
fix(leadspace): fixed tabbing order for buttons on leadspace

andysherman2121:
### Related Ticket(s)

https://github.com/carbon-design-system/ibm-dotcom-library/issues/1229

Switched the tab order so that the secondary button is focused before the primary button

Screen Shot 2020-03-26 at 4 46 11 PM


packages/react/src/patterns/sub-patterns/ButtonGroup/ButtonGroup.js
```diff
@@ -31,6 +31,7 @@ const ButtonGroup = ({ buttons }) => (
       return (
         <li key={key} className={`${prefix}--buttongroup-item`}>
           <Button
+            tabIndex={key === 0 ? 2 : 1}
```
jeffchew: If it's primary, shouldn't it be a lower tab index than higher?

andysherman2121: > If it's primary, shouldn't it be a lower tab index than higher?

@jeffchew The QA guys and Wonil confirmed they want the secondary button focused first because visually it's before the primary button

jeffchew: I'm curious why tabindex needs to be set at all then if it looks to be the natural tab order now?

andysherman2121: The natural tab order focuses on primary first

jeffchew: Is this the way the ButtonGroup is coded then?

andysherman2121: Yeah i agree, me and @RaphaelAmadeu were talking about this earlier

jeffchew: @andysherman2121 @RaphaelAmadeu @annawen1 
Going to approve for now, but one thought I have is that we might want to come back and readdress how the props are pulled in. My initial thought is to have a separate prop for a single primary button, and another prop for an array of secondary buttons.

Issue #1620, opened 2020-03-09
Learn Page: table of contents - Jump to option on mobile, blocks the heading of the content section.


Praveen-sr:
## Detailed description

Issue:- When user selects option from Jump to, the heading of the content section is not seen.

## Additional information

Actual :- 
![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/c665cfc1-9e74-4804-aca8-bbba18afdc28)

Excepted :- 
![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/0ce6bf70-8d1a-4d3a-881c-09f8df348e4a)

PR #1919, opened 2020-03-30
fix(util): added offset to the smooth-scroll util

RaphaelAmadeu:
### Related Ticket(s)
#1620 

### Description

An offset to the scrolling was added, so the entire heading appears properly after scrolling on mobile devices.

### Changelog

**New**

- Added scroll offset


packages/utilities/src/utilities/smoothScroll/smoothScroll.js
```diff
@@ -31,9 +31,13 @@ const smoothScroll = (e, selector) => {
   } else {
     return null;
   }
-  document.querySelector(getSelector).scrollIntoView({
+
+  window.scroll({
+    top:
+      document.querySelector(getSelector).getBoundingClientRect().top -
+      50 +
```
jeffchew: I feel like this should be an optional parameter, where the default is 0. Thoughts?

RaphaelAmadeu: This is a nice idea, i've implemented it on my latest commit

packages/utilities/src/utilities/smoothScroll/smoothScroll.js
```diff
@@ -31,9 +32,13 @@ const smoothScroll = (e, selector) => {
   } else {
     return null;
   }
-  document.querySelector(getSelector).scrollIntoView({
+
+  window.scroll({
```
jeffchew: @RaphaelAmadeu are we switching from `scrollIntoView`?

RaphaelAmadeu: @jeffchew ,
`scrollIntoView` does not support offset, and it was causing the heading to not be shown completely on mobile devices.

Issue #1780, opened 2020-03-24
Leadspace background image not rendering properly


diegomarcuz:
## Detailed description

> Describe in detail the issue you're having.

* Leadspace background image is not covering full width on big resolutions, like 2560 X 1440

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

* Bug

> Is this issue related to a specific component?
* Leadspace with images

> What did you expect to happen? What happened instead? What would you like to
> see changed?

* Leadspace needs to cover full width.

> What browser are you working in?
* I have tested on Chrome 80, Firefox 74 and Safari 13

> What version of the IBM.com Library are you using?
*  @carbon/ibmdotcom-react": "^1.6.0-alpha.3070",
*  @carbon/ibmdotcom-styles": "^1.6.0-alpha.3066",
*  @carbon/icons-react": "^10.9.1",
*  @carbon/pictograms-react": "^10.9.0",

## Steps to reproduce the issue

1. Step one
* Remove for bx--leadspace class: 
    * max-width;
    * margin-right;
    * margin-left.

2. Step two
* Add for bx--leadspace class:
   * background-size: cover;
   * background-position: center;
   * background-repeat: no-repeat.

## Additional information

![Screen Shot 2020-03-24 at 12 03 21](https://user-images.githubusercontent.com/37422384/77440831-81b0f980-6dc7-11ea-909c-a1b48eead967.png)

jeffchew:
@diegomarcuz can you provide the image url you're using?

diegomarcuz:
Sure, http://webdev01.webmaster.ibm.com/events/testing-images/2625-801-leadspace.jpg

RobertaJHahn:
Reach out to Diego to have him use Backgound cover for the image.  DDS dev should also investigate why we are not using the image component.  @annawen1 @andysherman2121

PR #1922, opened 2020-03-30
fix(scss): leadspace now fill wider resolutions

RaphaelAmadeu:
### Related Ticket(s)

#1780 

### Description

The leadspace image now spread to fill wider resolution.


packages/styles/scss/patterns/sections/leadspace/_leadspace.scss
```diff
@@ -72,7 +72,6 @@ $btn-min-width: 26;
   .#{$prefix}--leadspace {
     margin-right: auto;
     margin-left: auto;
-    max-width: 99rem;
```
jeffchew: @andysherman2121 wasn't there a change you made that added this recently?

cc: @RaphaelAmadeu

andysherman2121: Yeah the leadspace is supposed to have a max width so that when it reaches 99rem's. It has white gutters on either side of it

jeffchew: @larahanlon2 can you weigh in on the max width treatment of the leadspace?

photodow: @andysherman2121 are we not using the grid handle that `max-width`?

larahanlon2: @andysherman2121 is correct in saying that the lead space needs to max out with the grid – just like the masthead. Here's the visual spec that we confirmed before:  
Screen Shot 2020-04-01 at 9 07 43 AM

jeffchew: @larahanlon2 @kennylam does the masthead/footer also max out at that width? Should we be maxing the width on the dotcom shell level instead?

larahanlon2: The **masthead** has the same max behavior except that the HR at the bottom of the L0 continues to span full width. I believe the content in the **footer** also does the same except the background color continues to span the width of the browser. All of this is to ensure that the content (text, components etc.) are aligned from pattern to pattern.

andysherman2121: @photodow we're not using the grid to handle the max width. The max width is set at the max width breakpoint

kennylam: > @larahanlon2 @kennylam does the masthead/footer also max out at that width? Should we be maxing the width on the dotcom shell level instead?

@jeffchew we're already doing that. All content minus masthead/footer in `Dotcomshell` sits inside `bx--grid`, which maxes at 99rem. I didn't get a chance to familiarize myself with `leadspace` yet. Not sure why we need to specify a max width for that, instead of 100% of the parent.

image

jeffchew: @RaphaelAmadeu @kennylam @andysherman2121 So where did we land on this, we're removing the max-width for the leadspace? I'm assuming it still maxes out at 99rems if it's sitting in the DotcomShell, right?

jeffchew: I actually don't see a max-width happening for the dotcom shell:

Screen Shot 2020-04-03 at 8 39 39 AM

Perhaps this PR should also add the max-width to the `bx--dotcom-shell`?

jeffchew: Resolving this, we know we need to remove max-width for the leadspace, plus @RaphaelAmadeu just added the grid wrapper to storybook.

Issue #1777, opened 2020-03-24
Accessibility - Locale Selector - Screen reader is not identifying this (back button) control. Hence the user does not know its presence. 


Praveen-sr:
## Detailed description

Issue - 
Select by location or by language : back button - Screen reader is not identifying this control. Hence the user does not know its presence. 

Expected -
Screen reader should identify this control with appropriate Role/Label. And also this has to be made keyboard operable.

RaphaelAmadeu:
@Praveen-sr 
Could you please test the back button usage again?

Thank you!

Praveen-sr:
@RaphaelAmadeu Yes, ram verified on the link you provided, but Screen reader is not identifying the link in each card, all are announced only as text.

Praveen-sr:
@RaphaelAmadeu Sure, can you provide link to re test? 
Thanks!

RaphaelAmadeu:
@Praveen-sr 
You can use staging for the test,
Thanks!

Praveen-sr:
@RaphaelAmadeu Now screen reader (jaws) is identifying the button role. But, the label is given as "select region". Since the triggering control also has the same label, this is confusing. 
Hence recommend to label this as "back".
Voiceover iPhone: VO is not identifying the button role still.

RaphaelAmadeu:
@Praveen-sr 
We updated this component some days ago,
we thought that hardcoding the label with 'back' is not a good approach as it would not adapt for language changes,
Can you test this again please?
Thanks!

Praveen-sr:
@RaphaelAmadeu Re tested this, now screen reader is announcing the label as "select region" which is inappropriate. To avoid confusion, the label can be given as "Go back to Select region" ?

wonilsuhibm:
I suggest "Select a different region" for the aria-label for the back button.

ljcarot:
Since the icon is suggesting go back, how about 'Go back to select a different region'.

PR #1934, opened 2020-03-31
fix(accessibility): locale-modal back button now works with keyboard

RaphaelAmadeu:
### Related Ticket(s)

#1777 

### Description

I've added a role, tabindex and a listener to handle keyboard inputs on the modal header back button.

### Changelog

**New**

- Keyboard input handling


packages/react/src/components/LocaleModal/LocaleModalRegions.js
```diff
@@ -60,12 +60,28 @@ const LocaleModalRegions = ({
...
+          btn.setAttribute('tabindex', '1');
+          btn.setAttribute('role', 'button');
           btn.addEventListener('click', () => {
-            setIsFiltering(false);
-            setClearResults(true);
-            document.getElementById(`${prefix}--locale-modal__filter`).value =
-              '';
+            localeBackPressed(btn);
+          });
+          btn.addEventListener('keyup', e => {
```
jeffchew: @RaphaelAmadeu do we need to remove the event listeners when `localeBackPressed` is called?

RaphaelAmadeu: I've implemented the removal of the listeners as it is activated.

Issue #1600, opened 2020-03-05
Content block, request to change type token for copy


larahanlon2:
### The problem
The [Content block](https://ibmdotcom-react.mybluemix.net/?path=/story/patterns-sub-patterns-contentblock--default) copy (main paragraph after the heading) is currently using `expressive-heading-03` as per design specs. However, this type token was not intended for long reading purposes and should only be used for headings. As a result, the line height is very tight and makes reading a lot of text more difficult. 

Current:
<img width="1643" alt="Screen Shot 2020-03-03 at 5 42 49 PM" src="https://user-images.githubusercontent.com/53093550/76017163-225c7980-5eec-11ea-980a-a8b9b98caa71.png">

### The solution
Design team and Lila agree to replace the current type token with `expressive-paragraph-01` which is available via [Carbon](https://www.carbondesignsystem.com/guidelines/typography/expressive).

Desired:
<img width="974" alt="Screen Shot 2020-03-05 at 2 19 56 PM" src="https://user-images.githubusercontent.com/53093550/76017383-7f582f80-5eec-11ea-9f19-5369c18d93c6.png">

### Additional information
Making this change will improve the overall reading experience for viewers.

PR #1957, opened 2020-04-03
fix(issue-1600-type-token): updated content block

reesedownes:
### Related Ticket(s)

#1600 

### Description

Update the styling of Content Block copy from `'expressive-heading-03'` to `'expressive-paragraph-01'`

### Changelog

```
    &__copy {
      p {
        @include carbon--type-style('expressive-paragraph-01', true);
```


jeffchew:
@reesedownes fyi, I removed the `package:styles` label, that builds the carbon expressive deploy preview which we don't need here.

@larahanlon2 I already approved but wanted to get your eye on this before we merge this in.

kennylam:
> Hey, looks good but one question. **Can someone confirm that the line-height is correct?** According to Carbon, the line-height should be 40px / 2.5rem on max, 36px / 2.25rem on lg and xlg breakpoints, 30px / 1.875rem on med and sm. It's looking a little tight right now in my opinion but I want to confirm it's using the token correctly.

@larahanlon2 Confirmed those are the `line-height` values. Not exact, since the values scale based on the viewport size, but all within a pixel or two of what you listed.

jeffchew:
Merging, looks like the token is being used properly and PR is approved.

PR #1965, opened 2020-04-06
fix(css): fixes profile menu scroll; fixes #1960

### LINKED ISSUE #1960: [Masthead] IBM Login menu moves with scroll (Vanilla)
## Detailed description

When scrolling with the IBM Cloud menu open, it scrolls with the page

Feb-06-2020 12-45-53

This fix was made in React (https://github.com/carbon-design-system/ibm-dotcom-library/issues/1332), but we need to apply the same fix in Vanilla.

Masthead (Vanilla Package)

Chrome

`v1.5.1`

1. Go to http://ibmdotcom-vanilla.mybluemix.net/
2. View the Dotcom Shell
3. Open menu in upper right
4. Scroll the page

kennylam:
### Related Ticket(s)

#1960 

### Description

Fixes profile menu scrolling in the vanilla `Masthead` component. This is the vanilla version of #1947 

### Changelog

**New**

- sets profile menu list position to `fixed`


packages/styles/scss/components/masthead/_masthead.scss
```diff
@@ -309,6 +309,7 @@ $search-transition-timing: 95ms;
   // masthead profile menu
   .#{$prefix}--overflow-menu-options {
     background-color: $ui-background;
+    position: fixed;
```
jeffchew, line 312: Approved. @kennylam , with this are the hooks you added for the React one still needed? I think I recall you were adding the position fixed dynamically or something?

kennylam: @jeffchew Yes, this is the first thing i tried on the `react` package, thinking it should be this simple. But for some reason the React version inlines all the styles on the component, making this impossible. Only way to override that one is to either go `!important` or rewrite the actual property, which is what I did.

https://github.com/carbon-design-system/carbon/blob/0719ef89fb25fc34655abed891257d85e7ff1264/packages/react/src/internal/FloatingMenu.js#L350-L358

jeffchew: Got it, I guess this is React being React.

Issue #1933, opened 2020-03-31
Refactor CTA component


jeffchew:
#### User Story

> As a `[user role below]`:
IBM.com Library Developer

> I need to:
have a CTA component that is separated into multiple sub components

> so that I can:
easily manage, maintain, and scale the CTA component

#### Additional information

 - CTA is currently a single file: https://github.com/carbon-design-system/ibm-dotcom-library/tree/master/packages/react/src/components/CTA
 - Should break down to multiple sub files within this folder, possibly to manage each corresponding CTA type.
- Might be worth having the QA team test the deploy preview version of this before merging in.
- Should test if patterns using the CTA component are not impacted

#### Acceptance criteria
 - [ ] CTA broken down to multiple subfiles/components
 - [ ] Should have QA team retest to make sure all looks good

PR #1982, opened 2020-04-07
fix(cta): refactor of cta into subcomponents

annawen1:
### Related Ticket(s)

Refactor CTA component #1933

### Description

Refactor the CTA into subcomponents as it was becoming logic heavy

### Changelog

**New**

- add `TextCTA`, `FeatureCTA`, `ButtonCTA`, and `CardCTA`

**Changed**

- add `await` to the api call and ensure all Promises are resolved for the `videoTitle`


packages/react/src/components/CTA/ButtonCTA.js
```diff
@@ -0,0 +1,117 @@
...
+  jump: PropTypes.func,
+  type: PropTypes.string,
+  openLightBox: PropTypes.func,
+  setLightBox: PropTypes.func,
+  renderLightBox: PropTypes.bool,
+  launchLightBox: PropTypes.func,
+  videoTitle: PropTypes.array,
+  mediaData: PropTypes.object,
+  setMediaData: PropTypes.func,
+};
```
jeffchew, line 115: Should we set `defaultProps` for these? Looks like all of these are optional unless are some supposed to be required?

PR #1987, opened 2020-04-07
fix(tabs): fixed padding for selected tabs

andysherman2121:
### Related Ticket(s)

https://github.com/carbon-design-system/ibm-dotcom-library/issues/1356

### Description

Fixed issue where text is moving location when you select a tab. Below is how it used to look

tabbing


RaphaelAmadeu:
@andysherman2121 ,
I noticed that the label goes a little bit up when it's in the active state, is this expected?

andysherman2121:
> @andysherman2121 ,
> I noticed that the label goes a little bit up when it's in the active state, is this expected?

Thanks Raph, just fixed that issue

PR #1991, opened 2020-04-07
fix(select): fixed inline select when invalid

andysherman2121:
### Related Ticket(s)

https://github.com/carbon-design-system/ibm-dotcom-library/issues/1355

The inline select labels and helper text were moving position when the invalid state is activated.

### DIFF
--- a/packages/styles/scss/themes/expressive/components/_select.scss
+++ b/packages/styles/scss/themes/expressive/components/_select.scss
@@ -26,4 +26,12 @@
     height: rem(20px);
     width: rem(20px);
   }
+  .#{$prefix}--select--inline.#{$prefix}--select--invalid .#{$prefix}--label,
+  .#{$prefix}--select--inline.#{$prefix}--select--invalid
+    .#{$prefix}--form__helper-text {
+    margin-top: 1rem;
+  }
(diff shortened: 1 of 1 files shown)


packages/styles/scss/themes/expressive/components/_select.scss
```diff
@@ -26,4 +26,12 @@
     height: rem(20px);
     width: rem(20px);
   }
+  .#{$prefix}--select--inline.#{$prefix}--select--invalid .#{$prefix}--label,
+  .#{$prefix}--select--inline.#{$prefix}--select--invalid
+    .#{$prefix}--form__helper-text {
+    margin-top: 1rem;
+  }
+  .#{$prefix}--form__helper-text {
+    line-height: 1rem;
```
annawen1, line 35: can we combine these two under the same selector? or is it different?

andysherman2121: they're different, i tried combining already

annawen1: ok, I think the helper text originally supposed to be 1rem line-height but because carbon hardcoded it we didn't touch the type-face of it. I guess we can just merge it in as it is now

Issue #1930, opened 2020-03-31
Lightbox media viewer: When the image is tall or window is short, the text is hidden and the user is not able to see that there is text.


wonilsuhibm:
## Detailed description

> Describe in detail the issue you're having.

[storybook](https://ibmdotcom-react-experimental.mybluemix.net/?path=/story/components-lightboxmediaviewer--default)
When the image is tall or window is short, the text is hidden and the user is not able to see that there is text. 

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

Lightbox media viewer.

> What did you expect to happen? What happened instead? What would you like to
> see changed?

The image should be shorter and the text should be visible.

> What browser are you working in?

Chrome 80.0.3987.149 (Official Build) (64-bit) on Mac.

> What version of the IBM.com Library are you using?

1.5.1

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. Open [lightbox media viewer](https://ibmdotcom-react-experimental.mybluemix.net/?path=/story/components-lightboxmediaviewer--default)
2. Select the 200x750 image from the knobs

## Additional information

- See the screenshot below.

wonilsuhibm:
<img width="1029" alt="Screen Shot 2020-03-31 at 10 13 19 AM" src="https://user-images.githubusercontent.com/30183447/78036408-44ed8100-7338-11ea-8e61-1c75e3f6c38c.png">

PR #1998, opened 2020-04-08
fix(scss): image now scale down on medium and smaller devices

RaphaelAmadeu:
### Related Ticket(s)

#1930 

### Description

I've changed how the image size is handled in small and medium devices, so the images reduces in order to text to appear on these devices.


RaphaelAmadeu:
@wonilsuhibm 
I want to include you, so we can have the opinion of the design team on this

RaphaelAmadeu:
@kennylam ,
Just fixed this issue with the centering,
Thanks!

PR #1999, opened 2020-04-08
fix(accessibility): added aria-label to back button

RaphaelAmadeu:
### Related Ticket(s)

#1777 

### Description

Added label for the back button on the `locale selector` so it helps the screen reader

### DIFF
--- a/packages/react/src/components/LocaleModal/LocaleModalRegions.js
+++ b/packages/react/src/components/LocaleModal/LocaleModalRegions.js
@@ -71,11 +71,13 @@ const LocaleModalRegions = ({
           document.getElementById(`${prefix}--locale-modal__filter`).value = '';
           btn.removeAttribute('tabindex');
           btn.removeAttribute('role');
+          btn.removeAttribute('aria-label');
         };
 
         [...localeBackBtn].forEach(btn => {
           btn.setAttribute('tabindex', '1');
           btn.setAttribute('role', 'button');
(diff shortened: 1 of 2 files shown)


packages/react/src/components/LocaleModal/LocaleModalRegions.js
```diff
@@ -71,11 +71,13 @@ const LocaleModalRegions = ({
           document.getElementById(`${prefix}--locale-modal__filter`).value = '';
           btn.removeAttribute('tabindex');
           btn.removeAttribute('role');
+          btn.removeAttribute('aria-label');
         };
 
         [...localeBackBtn].forEach(btn => {
           btn.setAttribute('tabindex', '1');
           btn.setAttribute('role', 'button');
+          btn.setAttribute('aria-label', 'back');
```
jeffchew, line 80: @kennylam do you recall if there is a "Back" label we can leverage here instead of hardcoding the english only label here?

cc: @RaphaelAmadeu

kennylam: @jeffchew Yep you're right. That label is already provided by the translation file.

@RaphaelAmadeu `headerTitle` (Select region) is the one you want.

RaphaelAmadeu: @kennylam , Thanks, I'll change the label with the `headerTitle`

PR #2014, opened 2020-04-09
fix(scss): added max-width to dotcom-shell content

RaphaelAmadeu:
### Related Ticket(s)

#1780 

### Description

The `Dotcom-Shell` scss was updated, so it have the default `max-width` based on the masthead, in addition to that, the `Dotcom-Shell` has been added to the `Leadspace` story.


packages/react/src/patterns/sections/LeadSpace/__stories__/LeadSpace.stories.js
```diff
@@ -68,15 +69,15 @@ storiesOf('Patterns (Sections)|LeadSpace', module)
     };
 
     return (
-      <div className="bx--grid">
+      <DotcomShell footerType="short">
```
jeffchew, line 72: @RaphaelAmadeu I think in huddle we decided to keep the LeadSpace isolated in storybook.

Issue #1920, opened 2020-03-30
Linklist is rendering too high on a pattern


kexuibm:
Using v1.5.1-rc.1, the Link List is rendering too high on a pattern.

Here is what it looks like in the Visual Specs:
<img width="744" alt="Screen Shot 2020-03-30 at 3 13 04 PM" src="https://user-images.githubusercontent.com/55464272/77952027-00f76f00-7299-11ea-810a-56227b56c26a.png">

Here is what it looks like on NextJS:

<img width="1549" alt="Screen Shot 2020-03-30 at 3 14 02 PM" src="https://user-images.githubusercontent.com/55464272/77952109-1d93a700-7299-11ea-97a7-46c3e4e8f636.png">

RobertaJHahn:
Jeff has added this bug to the March 6, 2020 engineering huddle meeting agenda to see how this issue can be fixed.  A new decorator pattern may be needed to fix this issue.

jeffchew:
@wonilsuhibm @photodow 
Some notes we have from our engineering huddle to handle this:
- We are thinking of enhancing the ContentBlock sub-pattern to accept a LinkList which is optional, we would need to modify all of the decorator patterns to accommodate this
- One thought would be to perhaps allow any optional content on the right column? Question for design if this is something that's been considered for future state of the ContentBlock in general?

photodow:
@jeffchew can we sync up on this over slack, or a separate meeting? I don't know if it makes sense to do that, I'm a little concerned we're overloading and changing the purpose of content block here. I'm wondering if it would make more sense to add to something like the content block segmented, and content block simple before updating content block across the board. I also wonder if we should consider the possibility of a seperate type of decorator pattern to handle this instead so we can keep segmented and simple seperated.

jeffchew:
> @jeffchew can we sync up on this over slack, or a separate meeting? I don't know if it makes sense to do that, I'm a little concerned we're overloading and changing the purpose of content block here. I'm wondering if it would make more sense to add to something like the content block segmented, and content block simple before updating content block across the board. I also wonder if we should consider the possibility of a seperate type of decorator pattern to handle this instead so we can keep segmented and simple seperated.

This might be better discussed over a meeting. I'll try to find a time that works for us, but I'm not convinced we'll find a solution for this by `v1.6.0` code freeze.

cc: @RobertaJHahn @wonilsuhibm @ljcarot @photodow

jeffchew:
[...]

PR #2023, opened 2020-04-10
fix(content-block): render optional linklist in layout

annawen1:
### Related Ticket(s)

Linklist is rendering too high on a pattern #1920

### Description

wrap `ContentBlock` content and heading in `Layout` component to ensure the `LinkList` aligns properly with the copy.

Screen Shot 2020-04-10 at 12 15 46 PM

### Changelog

**Changed**

- wrap the `heading` and `content` in `Layout` component if `LinkList` is passed in

### DIFF
--- a/packages/react/src/patterns/sub-patterns/ContentBlock/ContentBlock.js
+++ b/packages/react/src/patterns/sub-patterns/ContentBlock/ContentBlock.js
@@ -11,8 +11,7 @@ import {
 } from '@carbon/ibmdotcom-utilities';
 import { CTA } from '../../../components/CTA';
 import cx from 'classnames';
-import Layout from '../Layout/Layout';
-import LinkList from '../LinkList/LinkList';
+import { Layout } from '../Layout';
 import PropTypes from 'prop-types';
 import React from 'react';
 import { settings } from 'carbon-components';
@@ -29,7 +28,7 @@ const { prefix } = settings;
  * @param {*} props.children JSX Components
  * @param {string} props.customClassName allows user to pass in custom class name
  * @param {*} props.cta CTA props object
- * @param {object} props.linkList LinkList props object
+ * @param {object} props.aside components to be passed into the right panel
  * @returns {*} JSX ContentBlock component
  */
 const ContentBlock = ({
@@ -38,7 +37,7 @@ const ContentBlock = ({
   children,
   customClassName,
   cta,
-  linkList,
+  aside,
 }) => {
   const content = (
     <>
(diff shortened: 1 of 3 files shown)


jeffchew:
@annawen1 can the stories for the content block decorator patterns be added to include the LinkList to optionally turn on or off? @photodow which patterns would need to have this as an example?

cc: @kennylam

kennylam:
@jeffchew @photodow Should we leave that right column open to accept any component and not just `LinkList`? So instead of taking `LinkList` as a prop it would just be generic child content for that right col?

(This looks great btw, @annawen1)

photodow:
> @jeffchew @photodow Should we leave that right column open to accept any component and not just LinkList?

I believe so. Other things will eventually need to live in the right hand side of the content block.

packages/react/src/patterns/sub-patterns/ContentBlock/ContentBlock.js
```diff
@@ -55,10 +56,58 @@ const ContentBlock = ({ heading, copy, children, customClassName, cta }) => {
...
+              <div></div>
+            </>
+          )
+        : title}
+      {linkList
+        ? _layoutWrap(
+            <>
+              <div>{content}</div>
+              <div>
+                <LinkList {...linkList} />
```
photodow, line 91: Should link list be built directly into content block, or should it just allow an adopter to pass in link list and other items?

packages/react/src/patterns/sub-patterns/ContentBlock/ContentBlock.js
```diff
@@ -55,10 +56,58 @@ const ContentBlock = ({ heading, copy, children, customClassName, cta }) => {
...
 
+/**
+ * wraps content in layout component
+ *
+ * @private
+ * @param {Element} content content elements
+ * @returns {*} jsx cta component
+ */
+const _layoutWrap = content => (
+  <Layout type="2-1">{content.props.children}</Layout>
```
photodow, line 108: I believe the nested option would need to be passed in to make this work. Probably should also allow for other props like border to be turned on as well.

packages/react/src/patterns/sub-patterns/ContentBlock/ContentBlock.js
annawen1: ok, let me try that out

packages/react/src/patterns/sub-patterns/ContentBlock/ContentBlock.js
annawen1: done

annawen1:
me trying to update my branch + lag = accidentally merging this PR in to master...have a PR to revert the changes if necessary...

jeffchew:
> me trying to update my branch + lag = accidentally merging this PR in to master...have a PR to revert the changes if necessary...

@annawen1 I'll take a look, but one thing I noticed, there's no documentation on how to do the new `aside` feature. I also mentioned before that we should add this to the storybook stories for the decorator patterns to show how it works in those contexts.

PR #2024, opened 2020-04-10
fix(issue-1695): using DDO to get language

reesedownes:
### Related Ticket(s)

https://app.zenhub.com/workspaces/ibmcom-library-5d449f3642eb1962336cbe52/issues/carbon-design-system/ibm-dotcom-library/1695

### Description

Editing the getLang() method of the Locale service to use the DDO first, with ```html lang="##"``` as a fallback. I took what I thought were the appropriate parts from the meta.js logic.

### Changelog

**Changed**

 - ```
    static async getLang() {
    let ddoLocal = await DDOAPI.getAll();

    if (ddoLocal && ddoLocal.page && ddoLocal.page.pageInfo) {
      const lang = root.document.documentElement.lang || undefined;
      let pageInfoIBM = ddoLocal.page.pageInfo.ibm; ...
```

<!-- Deploy Previews are enabled by applying the following labels for the corresponding package: -->
<!-- *** "package: react" -->
<!-- *** "package: services" -->

### CODE REVIEW

File: packages/services/src/services/Locale/Locale.js [hunk 1]
```diff
@@ -152,8 +154,46 @@ class LocaleAPI {
...
+
+        // Account for more incorrect coding.
+        if (!pageInfoIBM.lc) pageInfoIBM.lc = 'en';
+        if (!pageInfoIBM.cc) pageInfoIBM.cc = 'us';
+
+        return {
+          cc: pageInfoIBM.cc,
+          lc: pageInfoIBM.lc,
+        };
+      } else return _localeDefault;
```
jeffchew, line 195: The logic seems a bit off here. We should probably try grabbing from the html lang attribute before going to default.

packages/services/src/services/Locale/Locale.js
```diff
@@ -152,8 +154,46 @@ class LocaleAPI {
    *    const locale = await LocaleAPI.getLang();
    * }
    */
-  static getLang() {
-    if (root.document.documentElement.lang) {
+  static async getLang() {
+    let ddoLocal = await DDOAPI.getAll();
+
+    if (ddoLocal && ddoLocal.page && ddoLocal.page.pageInfo) {
+      const lang = root.document.documentElement.lang || undefined;
```
jeffchew, line 161: @reesedownes we might want to separate out the logic here a bit. Perhaps a private method for `getFromDDO`, and `getFromLang`, then `getLang` would just be doing one or the other, or ultimately `_localeDefault` if both fail.

reesedownes:
@jeffchew agreed. did some refactoring, so take a look, please.

packages/services/src/services/Locale/Locale.js
```diff
@@ -85,6 +85,76 @@ const _axiosConfig = {
...
+      }
+
+      if (lang) {
+        pageInfoIBM.lc = pageInfoIBM.lc || lang.substring(0, 2).toLowerCase();
+        pageInfoIBM.cc = pageInfoIBM.cc || lang.substring(3, 5).toLowerCase();
+      }
+
+      // Account for more incorrect coding.
+      if (!pageInfoIBM.lc) pageInfoIBM.lc = 'en';
+      if (!pageInfoIBM.cc) pageInfoIBM.cc = 'us';
```
jeffchew, line 147: Let's kill this block (lines 140-147). It looks like if DDO fails, it will default to en/us. We should have en/us be the fallback if both `_getLocaleFromDDO` and `_getLocaleByLangAttr` fail.

packages/services/src/services/Locale/Locale.js
```diff
@@ -85,6 +85,76 @@ const _axiosConfig = {
...
+ * Return a locale object based on the DDO API, or "false"
+ * so the consumer can decide what to do next
+ *
+ * @returns {(object | boolean)}
+ * @private
+ */
+async function _getLocaleFromDDO() {
+  const ddoLocal = await DDOAPI.getAll();
+
+  const lang = root.document.documentElement.lang || undefined;
```
jeffchew, line 118: Can remove this line, let's keep lang logic in its corresponding method `_getLocaleByLangAttr`.

packages/services/src/services/Locale/Locale.js
```diff
@@ -85,6 +85,76 @@ const _axiosConfig = {
...
+  let pageInfoIBM = ddoLocal.page.pageInfo.ibm;
```
jeffchew, line 119: Let's move this inside the if block right below in case the ddo object is malformed or nonexistent.

packages/services/src/services/Locale/Locale.js
```diff
@@ -85,6 +85,76 @@ const _axiosConfig = {
...
+
+  if (ddoLocal && ddoLocal.page && ddoLocal.page.pageInfo) {
+    if (pageInfoIBM) {
+      // Set proper CC for us to use.
+      if (pageInfoIBM.country) {
```
jeffchew, line 124: Is this from existing logic in v18? Wonder if the line above should be compressed to `if (pageInfoIBM && pageInfoIBM.cc && pageInfoIBM.lc) { ... }`

jeffchew:
@reesedownes unit tests will need to get updated as well: https://app.circleci.com/pipelines/github/carbon-design-system/ibm-dotcom-library/1080/workflows/77c21466-c7dd-4198-a501-0b49a9617635/jobs/2059

kennylam:
@reesedownes the CI is failing because both `cc` and `lc` are `undefined` in the unit tests.

image

reesedownes:
thanks for the help, @kennylam 

@jeffchew fixed the tests and did a refactor... and no, i don't want to commit 105 things. that just happened when i rebased trying to see if that would make "react - experimental" work. if i just have to make another branch, i will.

jeffchew:
@

> thanks for the help, @kennylam
> 
> @jeffchew fixed the tests and did a refactor... and no, i don't want to commit 105 things. that just happened when i rebased trying to see if that would make "react - experimental" work. if i just have to make another branch, i will.

You can do a git reset to before the rebase and then do a rebase via the Github PR instead.

reesedownes:
@jeffchew ok, looks good now. should say "8 commits, 4 files changed".

Issue #1776, opened 2020-03-24
Accessibility - Locale Selector - Keyboard - Focus visible issues 


Praveen-sr:
## Detailed description

Issue :-

1.All the Card present in model page (For example:- Americas, Asia Pacific, Europe, Americas, Middle East and Africa)
Focus is going to the card/element but it is not visible at all.

2.Scroll bar/frame (Location and languages page)
Focus is going to the scroll bar/frame but it is not visible at all.

Expected Result:
Provide thick and strong visual focus
1. If, Visual Focus Indicator with 3 CSS px stroke then provide a sufficient contrast that is equal to 3:1. 
2. Visual focus indicator for a link that is a 1 CSS px then provide a sufficient contrast greater than 4.5:1

For more information please find the below link,
https://www.w3.org/WAI/GL/low-vision-a11y-tf/wiki/Contrast_(Minimum)

RaphaelAmadeu:
@Praveen-sr 

Could you please provide a video or a screenshot of this issue?
So we can a better idea of the issue,

Thanks!

Praveen-sr:
@RaphaelAmadeu Screenshot for this issue:
![image](https://user-images.githubusercontent.com/60002341/78795863-22f89c00-79d3-11ea-9cfd-49824a9c7e67.png)

![image](https://user-images.githubusercontent.com/60002341/78795910-360b6c00-79d3-11ea-8b7e-15997c08e6e9.png)

RaphaelAmadeu:
I've just tested it and the focus is working fine on 
https://ibmdotcom-react.mybluemix.net/?path=/story/components-locale-modal--default

ljcarot:
Works for me too.  Here is recording of the focus appearing https://recordit.co/L9G3Zv52NL

Praveen-sr:
@RaphaelAmadeu 
Tested this on [Storybook ](https://ibmdotcom-react.mybluemix.net/?path=/story/components-locale-modal--default) 
When we open the locale selector, we are getting the visual focus on the cards and its **working fine**.

Tested this on [Learn Page ](http://ibmdotcom-nextjs-staging.mybluemix.net/learn.html)

When we open the locale selector from the footer region we are not getting the visual focus on the cards. **Not working**.

cc @ljcarot

PR #2045, opened 2020-04-13
fix(scss): fixed footer styles overriding locale-modal

RaphaelAmadeu:
### Related Ticket(s)

#1776 

### Description

I've updated the scss selectors in order to fix a bug where the styles aimed for the footer would override the focus state on the locale modal tiles.


packages/styles/scss/components/footer/_footer.scss
```diff
@@ -41,7 +41,7 @@
       }
     }
 
-    .#{$prefix}--link {
+    li a.#{$prefix}--link {
```
jeffchew, line 44: @RaphaelAmadeu wonder if we should add some specificity rather than target the element:
```
.#{$prefix}--footer.#{$prefix}--link {
```

RaphaelAmadeu: @jeffchew 
Just updated the PR with some changes regarding this,
Thanks!

RaphaelAmadeu:
@jeffchew 
Just added the new class to the Vanilla footer as well

jeffchew:
> @jeffchew
> Just added the new class to the Vanilla footer as well

Thanks @RaphaelAmadeu , it looks like you need to update your jest snapshot file for this to pass CI checks.

Issue #1769, opened 2020-03-24
Locale selector:- Keyboard - Select region modal dialog


chetankan:
Issue :- Unable to close the dialog by pressing escape key.

Expected:- Apart from close button, user would also expect the dialog to be closed by pressing the "escape" key.

URL:- http://ibmdotcom-nextjs-test.mybluemix.net/learn.html

jeffchew:
From the PR:

https://github.com/carbon-design-system/ibm-dotcom-library/pull/2046#issuecomment-613609879

```@chetankan @jeffchew @guilhermelMoraes Sorry I'm a little late to this but I wasn't aware of #1769. I'm not sure this is the correct interaction. This code is working in the LocaleModal story, but will fail in the Footer story since esc closes the entire modal, which is pretty much the standard behavior for all modals.

We should either trap the modal esc while filtering country/lang, or better yet, leave this as it is. This isn't an accessibility issue. What do you guys think?```

I'm inclined to leave this alone, thoughts?

cc: @ljcarot @kennylam @chetankan @Praveen-sr @guilhermelMoraes

Praveen-sr:
@RobertaJHahn Ram re tested this on staging and Learn page, It is working fine. We are able to close the dialog by pressing escape key.

jeffchew:
Per @Praveen-sr 's comment we are closing this issue.

PR #2046, opened 2020-04-13
fix(LocaleModal): esc key closes the Locale Modal filtering

guilhermelMoraes:
### Related Ticket(s)

Locale selector:- Keyboard - Select region modal dialog [#1769](https://github.com/carbon-design-system/ibm-dotcom-library/issues/1769)

### Description

If the user is filtering the region at the `LocaleModal` and press the Escape key on the keyboard it closes the filter.

### Changelog

**New**

- keydown eventListener in `LocaleModal` to detect the Escape key press.


packages/react/src/components/LocaleModal/LocaleModal.js
```diff
@@ -71,6 +71,21 @@ const LocaleModal = ({ isOpen, setIsOpen }) => {
...
 
+  useEffect(() => {
+    /**
+     * @param {number} keyCode - The code of the keyboard key pressed.
+     */
+    function closeFilterOnEscapeKeyPress({ keyCode }) {
+      if (keyCode === 27 && isFiltering) {
+        setIsFiltering(false);
+      }
+    }
```
jeffchew, line 82: Minor thing, let's make this a private method outside of the `useEffect`, and include the underscore in front of the method name, maybe something like `_closeModalOnEscape`. Be sure to include the `@private` flag in the JSDocs.

packages/react/src/components/LocaleModal/LocaleModal.js
```diff
@@ -162,6 +168,17 @@ const LocaleModal = ({ isOpen, setIsOpen }) => {
   function close() {
     setIsOpen(false);
   }
+
+  /**
+   * @private
```
jeffchew, line 173: Can you add a description for this private method too?

packages/react/src/components/LocaleModal/LocaleModal.js
```diff
@@ -162,6 +168,17 @@ const LocaleModal = ({ isOpen, setIsOpen }) => {
...
+   *
+   * @param {number} keyCode - The code of the keyboard key pressed.
+   */
+  function _closeFilterOnEscapeKeyPress({ keyCode }) {
```
jeffchew, line 177: Does this need to be passed in as an object?

guilhermelMoraes: The `keyCode` property came from the keydown event, which is an object.

packages/react/src/components/LocaleModal/LocaleModal.js
guilhermelMoraes: For sure!

packages/react/src/components/LocaleModal/LocaleModal.js
jeffchew: ok. For JSDocs, we'd want to probably document that then:

```
/**
 * Detects for the esc key press then closes the modal
 * 
 * @param {object} event Event object from the event listener
 * @param {number} event.keyCode code of the keyboard key pressed
 * @private
 */
```

guilhermelMoraes: Thanks, Jeff!

kennylam:
@chetankan @jeffchew @guilhermelMoraes Sorry I'm a little late to this but I wasn't aware of #1769. I'm not sure this is the correct interaction. This code is working in the `LocaleModal` story, but will fail in the `Footer` story since `esc` closes the entire modal, which is pretty much the standard behavior for all modals.

We should either trap the modal `esc` while filtering country/lang, or better yet, leave this as it is. This isn't an accessibility issue. What do you guys think?

Praveen-sr:
@jeffchew @guilhermelMoraes Ram re tested this on staging and Learn page, It is working fine. We are able to close the dialog by pressing escape key.

guilhermelMoraes:
> @jeffchew @guilhermelMoraes Ram re tested this on staging and Learn page, It is working fine. We are able to close the dialog by pressing escape key.

@Praveen-sr what exactly is closing when do you press the Esc key? The entire modal or only the filter part?

jeffchew:
@guilhermelMoraes sounds like we should just close this PR.

Issue #1193, opened 2020-01-24
Content block - simple: End of section line width


ltitle:
Viewed in Chrome

NOTE: THESE SPECS APPLY TO ALL PATTERNS WITH A HORIZONTAL RULE AT THE END OF IT

The horizontal rule at the end of the pattern should extend all the way to the right edge of the grid for Large, XL and Max breakpoints.
The horizontal rule at the end of the pattern should extend all the way to both edges of the screen for Medium and Small breakpoints.

As seen in Storybook in XL breakpoint:
![image](https://user-images.githubusercontent.com/57198117/73103644-8b4edb80-3ec2-11ea-8dde-9f7673e51e94.png)

Expected result for L, XL and Max (note: on Max size the line will still stop at the end of the grid):
![image](https://user-images.githubusercontent.com/57198117/73103704-aa4d6d80-3ec2-11ea-8b40-0b12f1a24b00.png)

As seen in Storybook for Medium breakpoint:
![image](https://user-images.githubusercontent.com/57198117/73103751-c6e9a580-3ec2-11ea-8974-351ff9615cd7.png)

Expected result for Medium:
![image](https://user-images.githubusercontent.com/57198117/73103732-ba654d00-3ec2-11ea-8122-22dddfcb5ad5.png)

jeffchew:
@ltitle this may have been solved in a different issue, can you take a look at our NextJS instance here:

https://ibmdotcom-nextjs-test.mybluemix.net

jeffchew:
From @ltitle : 
```
It doesnt look resolved on the test site
on desktop sizes the line should extend all the way to the right and on med and small it should extend the full width of the screen
```

ljcarot:
@jeffchew I'm on desktop and it looks like the line extends all the way to the right.

ltitle:
@jeffchew Maybe I am not looking at the right thing but I am still not seeing the line go all the way to the right edge.

![image](https://user-images.githubusercontent.com/57198117/77780770-e4ee9600-702a-11ea-8889-2346809ee698.png)

jeffchew:
Unassigning and keeping in the backlog for now.

PR #2090, opened 2020-04-15
fix(scss): fixed border size for layout pattern

RaphaelAmadeu:
### Related Ticket(s)

#1193 

### Description

The border for the `Layout` pattern was updated on `md`, `lg`, `xlg` and `max` breakpoints, so it better fits the design specs.


jeffchew:
Approved and merging this. The failure to build for react experimental is a false positive as I was in the middle of adding node engine detection in the devops task.

Issue #1946, opened 2020-04-01
Component:  Masthead - Update the user documentation regarding the css loading order


RobertaJHahn:
#### User Story

> As a `[user role below]`:
ibm.com website developer

> I need to:
know how to load the Carbon styles and ibm.com Library scss order

> so that I can:
prevent writing over the ibm.com Library scss with the Carbon styles 

#### Additional information

 - This issue is to update the adopter documentation.  See additional information in the closed issue (#673)

#### Acceptance criteria
 - [ ] Updated user documentation

PR #2096, opened 2020-04-15
fix(docs): updated documentation for masthead

RaphaelAmadeu:
### Related Ticket(s)

#1946 

### Description

The documentation was updated so we included that our import must be done after the `carbon-components` one if both are being used.


jeffchew:
> @RaphaelAmadeu Just a thought, but should we add the Carbon-specific documentation to the main
> [styles](https://github.com/carbon-design-system/ibm-dotcom-library/tree/master/packages/styles#usage) README instead?

@RaphaelAmadeu can you take a look? Might make sense to have in the styles README, perhaps the main one.

cc: @kennylam

RaphaelAmadeu:
@jeffchew 
I've added it to the styles main README some days ago,
CC: @kennylam

packages/react/src/components/Masthead/README.md
```diff
@@ -17,7 +17,7 @@ Here's a quick example to get you started.
 @import '@carbon/ibmdotcom-styles/scss/components/masthead/masthead.scss';
 ```
 
-> 💡 Only import fonts once per usage. Don't forget to import the Masthead
+> `💡 Only import fonts once per usage. Don't forget to import the Masthead`
```
jeffchew, line 20: Screen Shot 2020-04-28 at 9 27 10 AM

The change from this seems odd, is this necessary?

PR #2107, opened 2020-04-16
fix(issue-2103): isReady() catch, new test

reesedownes:
### Related Ticket(s)

https://github.com/carbon-design-system/ibm-dotcom-library/issues/2103

### Description

DDOAPI.isReady() gets a catch clause so it knows how to handle no DDO being available. LocaleAPI._getLocaleFromDDO() now falls back immediately to LocaleAPI._getLocaleByLangAttr() on receiving the null response.

This will let the TranslationAPI function correctly and render the nav (and other components that rely on LocaleAPI.getLang()).

### Changelog

**Changed**

updated DDO and Locale services, and added an additional test to Locale.test


packages/services/src/services/Locale/Locale.js
```diff
@@ -144,14 +144,14 @@ async function _getLocaleFromDDO() {
       }
     }
 
-    if (!pageInfoIBM.lc || !pageInfoIBM.cc) return false;
+    if (!pageInfoIBM.lc || !pageInfoIBM.cc) return _getLocaleByLangAttr();
```
jeffchew, line 147: @reesedownes looking at the logic in getLang, I wonder if it would be ok to leave it as-is with returning `false`, just so we can keep the logic of fetching from DDO and fetching from lang separate.

packages/services/src/services/Locale/__tests__/Locale.test.js
```diff
@@ -12,6 +12,7 @@ jest.mock('@carbon/ibmdotcom-utilities', () => ({
   },
   geolocation: jest.fn(() => Promise.resolve('us')),
 }));
+jest.setTimeout(20000);
```
jeffchew, line 15: What is this doing here?

packages/services/src/services/Locale/Locale.js
reesedownes: @jeffchew yeah, that's pretty heavy handed. i'll try it that way quickly.

packages/services/src/services/Locale/__tests__/Locale.test.js
reesedownes: leaving it the default timeout of 5000 ms didn't give it enough the test enough time to wait for the instances of ```const lang = await LocaleAPI.getLang();``` to complete when there was no html lang attr and no DDO. it seemed to need an average of ~12 seconds.

reesedownes:
@jeffchew the new fixes are in

PR #2131, opened 2020-04-20
fix(leadspace): properly point to defaultSrc for the Image

jeffchew:
### Related Ticket(s)

No related issue

### Description

I can't seem to find the issue that this was related to, but wondering if this should get patched in. The Leadspace is pointing to `image.default` and not `image.defaultSrc`. 

### Changelog

**Changed**

- Fix to Leadspace to properly point to `defaultSrc`


jeffchew:
@Praveen-sr @chetankan fyi, this is getting hotfixed for the next release candidate, which affects the Leadspace.

PR #2135, opened 2020-04-20
fix(tooltip): fixed tooltip link fontsize

andysherman2121:
### Related Ticket(s)

https://github.com/carbon-design-system/ibm-dotcom-library/issues/1274

### Description

Updated tooltip link to expressive size.

### DIFF
--- a/packages/styles/scss/themes/expressive/components/_file-uploader.scss
+++ b/packages/styles/scss/themes/expressive/components/_file-uploader.scss
@@ -28,4 +28,7 @@
     height: rem(20px);
     width: rem(20px);
   }
+  .#{$prefix}--tooltip .#{$prefix}--link {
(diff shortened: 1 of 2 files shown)


packages/styles/scss/themes/expressive/components/_file-uploader.scss
```diff
@@ -28,4 +28,7 @@
     height: rem(20px);
     width: rem(20px);
   }
+  .#{$prefix}--tooltip .#{$prefix}--link {
+    font-size: 1rem;
+  }
```
annawen1, line 33: We should put this in the tooltip file no? Doesn't seem like it's part of file-uploader

andysherman2121: you're on to something !

Issue #1656, opened 2020-03-11
Content Section: Theme issue in Storybook: On applying g100, g90 , g10 text in children section is not visible


chetankan:
Issue:- On applying 'g100' and 'g90' theme text on children section is not visible.
On applying 'g10' theme visually user cannot differentiate between 'Component Heading' and 'Children' section
 
## Steps to reproduce the issue

1. Enter url in the browser
2. Click on 'Content Section'
3. Set theme to 'g100' , 'g90' and 'g10' .

## Additional information

OS:- Windows 10, Mac, iOS(12,13) , Android (8,9)
Browsers:- Chrome, Safari, FF, Samsung, UC browser, Edge

URL:- https://ibmdotcom-react-staging.mybluemix.net/?path=/story/patterns-sub-patterns-contentsection--default![G100 theme.png](https://images.zenhubusercontent.com/5e21a841639f517d456f7e43/0a4b23c2-3782-4751-8ada-d9a6532d44e5)![G90 theme.png](https://images.zenhubusercontent.com/5e21a841639f517d456f7e43/18244dd9-b0fb-465b-b2c0-6a3a0c7a88ed)![g10 theme.png](https://images.zenhubusercontent.com/5e21a841639f517d456f7e43/c95f0cb4-5876-486d-9775-a8060b87171b)

PR #2161, opened 2020-04-21
style(content-section): fixing theme error on storybook

guilhermelMoraes:
### Related Ticket(s)

[Content Section: Theme issue in Storybook: On applying g100, g90 , g10 text in children section is not visible #1656](https://github.com/carbon-design-system/ibm-dotcom-library/issues/1656)

### Description

The themes were not being applied correctly. This problem seems to related only with the storybook appearance.

### Changelog

**Changed**

- `packages/react/src/patterns/sub-patterns/ContentSection/__stories__/index.scss`

### DIFF
--- a/packages/react/src/patterns/sub-patterns/ContentSection/__stories__/index.scss
+++ b/packages/react/src/patterns/sub-patterns/ContentSection/__stories__/index.scss
@@ -8,23 +8,23 @@
 .#{$prefix}--content-section-story .#{$prefix}--content-section__children {
   height: carbon--mini-units(60);
   padding: $carbon--spacing-05;
-  @include themed-items;
+  @include content-section-themed-items;
 }
 
 .#{$prefix}--content-section--g10 .#{$prefix}--content-section__children {
   @include carbon--theme($carbon--theme--g10) {
-    @include themed-items;
+    @include content-section-themed-items;
   }
 }
 
 .#{$prefix}--content-section--g90 .#{$prefix}--content-section__children {
   @include carbon--theme($carbon--theme--g90) {
-    @include themed-items;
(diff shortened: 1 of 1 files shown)


packages/react/src/patterns/sub-patterns/ContentSection/__stories__/index.scss
```diff
@@ -1,7 +1,30 @@
 @import '../../../../../../styles/scss/patterns/sub-patterns/content-section/content-section';
 
+@mixin themed-items {
```
jeffchew, line 3: Let's rename this to something specific to the content section, perhaps `content-section-theme-items`. Remember that mixins are available outside of this file if compiled.

guilhermelMoraes: Ok, all right. I did it

packages/react/src/patterns/sub-patterns/ContentSection/__stories__/index.scss
```diff
@@ -1,7 +1,30 @@
...
+@mixin content-section-themed-items {
+  color: $text-01;
+  background: $ui-01;
+}
+
 .#{$prefix}--content-section-story .#{$prefix}--content-section__children {
   height: carbon--mini-units(60);
-  background-color: #f4f4f4;
   padding: $carbon--spacing-05;
+  @include themed-items;
```
jeffchew, line 11: @guilhermelMoraes you'll have to update these includes too...

guilhermelMoraes: My bad!

Issue #2162, opened 2020-04-21
Video title/description in LightboxMediaViewer should be populated by Kaltura API


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

The video in lightboxmediaviewer is getting populated by the title/description props, but should be by the API

> Is this issue related to a specific component?

LightboxMediaViewer

> What version of the IBM.com Library are you using?

v1.6.0-rc.0

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. Go to: https://ibmdotcom-react-staging.mybluemix.net/?path=/story/components-lightboxmediaviewer--embedded-video-player
2. Clear title and description

jeffchew:
Side note, layout looks way off when there's no title or description, and the video itself is super small:

![Screen Shot 2020-04-21 at 1 15 07 PM](https://user-images.githubusercontent.com/1641214/79896699-b1f3b400-83d6-11ea-9872-ab81db2fbb33.png)

@annawen1 do you want to create a separate issue for this?

cc: @RaphaelAmadeu @wonilsuhibm @ljcarot @RobertaJHahn

PR #2164, opened 2020-04-21
fix(cta): pass in title and desc to lightboxviewer from video api

annawen1:
### Related Ticket(s)

Video title/description in LightboxMediaViewer should be populated by Kaltura API #2162

### Description

Populate the `lightboxmediaviewer` title and description from the Kaltura API when setting video media type in `CTA` component

The `LightboxMediaViewer` now makes call to `VideoPlayerAPI` service to retrieve title, alt, and description. To prevent multiple calls to the api, object is set up within the service to cache video data.

### Changelog

**Changed**

- set `alt`, `title`, and `description` from KalturaAPI in `LightboxMediaViewer`
- set object within videoplayer service to cache video data to prevent multiple api calls for the same video id

### DIFF
--- a/packages/services/src/services/VideoPlayer/VideoPlayer.js
+++ b/packages/services/src/services/VideoPlayer/VideoPlayer.js
@@ -190,8 +190,8 @@ class VideoPlayerAPI {
    */
   static async api(videoId) {
     return await this.checkScript().then(() => {
-      if (videoData && videoData.id === videoId) {
-        return videoData;
+      if (videoData && videoData[videoId]) {
+        return videoData[videoId];
       } else {
         return new Promise(resolve => {
           return new root.kWidget.api({ wid: _partnerId }).doRequest(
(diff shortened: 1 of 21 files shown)


jeffchew:
@annawen1 looks like you might still be working on this, wanted to make sure LightboxMediaViewer is also on your radar: https://ibmdotcom-react.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/2164/index.html?path=/story/components-lightboxmediaviewer--embedded-video-player

annawen1:
@jeffchew yep done

packages/services/src/services/VideoPlayer/VideoPlayer.js
```diff
@@ -182,18 +190,23 @@ class VideoPlayerAPI {
...
+      } else {
+        return new Promise(resolve => {
+          return new root.kWidget.api({ wid: _partnerId }).doRequest(
+            {
+              service: 'media',
+              action: 'get',
+              entryId: videoId,
+            },
+            function(jsonObj) {
+              videoData = jsonObj;
```
jeffchew, line 204: should this be set actually as an associative array in case multiple videos are on the page?
```videoData[jsonObj.id] = jsonObj;```

PR #2165, opened 2020-04-21
fix(modal): added missing gradient to overflow of modal

andysherman2121:
### Related Ticket(s)

https://github.com/carbon-design-system/ibm-dotcom-library/issues/1318

### Description

The modal was missing a gradient on the bottom of the text which was causing the overflow text to look cut off completely


packages/styles/scss/themes/expressive/components/_modal.scss
```diff
@@ -0,0 +1,25 @@
...
+
+/// Modal styles (expressive)
+/// @access private
+/// @group modal-expressive
+@mixin modal-expressive {
+  .#{$prefix}--modal-content--overflow-indicator {
+    background-image: linear-gradient(
+      to bottom,
+      rgba(244, 244, 244, 0),
+      #f4f4f4
```
annawen1, line 22: can we use a color-token here instead?

annawen1: looks like in carbon, they have `background-image: linear-gradient(to bottom, rgba($ui-01, 0), $ui-01);` in `_modal.scss` 

have you tried updating to the latest carbon dependencies to see if this issue would be resolved?

andysherman2121: Yeah updating carbon dependencies doesn't fix the issue

Issue #2099, opened 2020-04-16
React error in masthead


kyjak:
"@carbon/ibmdotcom-react": "^1.5.1-rc.1",
"@carbon/ibmdotcom-styles": "^1.5.1-rc.1",
"@carbon/icons": "^10.9.2",
"@carbon/type": "^10.9.2",
"carbon-components": "^10.10.3",
"carbon-components-react": "^7.10.3",

![image](https://user-images.githubusercontent.com/15310250/79442874-65c10200-7fd9-11ea-86bb-fdd64dadce3d.png)

jeffchew:
Hi @kyjak , is this impacting your application at all?

PR #2172, opened 2020-04-22
Fix/masthead react errors

kennylam:
### Related Ticket(s)

#2099 

### Description

Fixes various React errors in `Masthead`

### Changelog

**Changed**

- no more `javascript:void(0);` for `href`'s. React no longer allowing for versions > 16.9.0
- updated propType types
- small css changes to `MastheadLeftNav`

### DIFF
--- a/packages/react/src/components/carbon-components-react/UIShell/SideNavMenu.js
+++ b/packages/react/src/components/carbon-components-react/UIShell/SideNavMenu.js
@@ -66,7 +66,7 @@ export class SideNavMenu extends React.Component {
     /**
      * For submenu back button to toggle expand/collapse
      */
-    isBackButton: PropTypes.bool,
+    isbackbutton: PropTypes.bool,
   };
 
   static defaultProps = {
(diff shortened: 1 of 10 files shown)


packages/react/src/components/carbon-components-react/UIShell/SideNavMenu.js
```diff
@@ -177,7 +177,7 @@ export class SideNavMenu extends React.Component {
     if (item) {
       return React.cloneElement(item, {
         onClick:
-          item.props.isBackButton === true
+          item.props.isbackbutton === 'true'
```
jeffchew, line 180: it's a boolean value but is checking against a string value?

kennylam: @jeffchew Yeah that was one of the React errors. It didn't like that a prop was placed on a DOM element and checking for a boolean. This is their suggestion to get around that.

jeffchew: ok, so the proptype should be PropType.string instead of PropType.bool?

Otherwise looks good, I already approved it.

kennylam: @jeffchew Doh! Let me fix that...

PR #2173, opened 2020-04-22
fix(lightbox): display videoplayer and desc copy without distortion

annawen1:
### Related Ticket(s)

### Description

If description or title was too short, the video player would shrink

BEFORE: 
Screen Shot 2020-04-22 at 1 42 38 PM

AFTER: videoplayer doesn't shrink even if there is no title or description
Screen Shot 2020-04-22 at 1 43 04 PM

### Changelog

**Changed**

- override carbon's `padding-right: 20%` on the `modal-content`

**Removed**

- styles setting media to the top of the modal, media should be vertically centered

### DIFF
--- a/packages/services/src/services/VideoPlayer/VideoPlayer.js
+++ b/packages/services/src/services/VideoPlayer/VideoPlayer.js
@@ -190,8 +190,8 @@ class VideoPlayerAPI {
    */
   static async api(videoId) {
     return await this.checkScript().then(() => {
-      if (videoData && videoData.id === videoId) {
-        return videoData;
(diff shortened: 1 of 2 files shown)


annawen1:
@RaphaelAmadeu can you take a look at this? I removed some of your code and want to make sure I didn't revert any important bug fixes...

annawen1:
so seems that removing the `height: 100%` from the row, reverts changes for this bug: https://github.com/carbon-design-system/ibm-dotcom-library/issues/1931

However by setting the row height to 100%, the copy and image become misaligned when modal is at full height

Issue #1703, opened 2020-03-16
Learn Page:- Clicking on back button in browser, mismatch in page focus and table of content scroll indicator 


chetankan:
Issue:- when user clicks on any card or cta, user is navigated to respective landing page but on clicking back button(browser) here page focus stays on where user clicked but table of content scroll indicator (blue line) always sticks to first title

## Steps to reproduce the issue

1. Enter the url in the browser
2. Click on 'Learn More' card.
3. Click on 'Content Block - With Media'
4. Click on 'Feature Card'.
5. Once user is navigated to feature card link, click on browser back button.
6. Check the page focus and table of content blue tick.

## Additional information

URL:- http://ibmdotcom-nextjs-staging.mybluemix.net/learn.html

OS:- Windows 10, Mac
Browsers:- Chrome, Firefox, Safari, Edge![Back button issue on learn page.png](https://images.zenhubusercontent.com/5e21a841639f517d456f7e43/98ccf5c0-3b4a-4791-b680-a33b7ce52b7d)

RaphaelAmadeu:
@chetankan 
Could you please test this again using our canary staging? thanks!
https://ibmdotcom-react-canary.mybluemix.net/

PR #2182, opened 2020-04-23
fix(pattern): table-of-contents now identify the whole section

RaphaelAmadeu:
### Related Ticket(s)

#1703 

### Description

A bug was happening in our next js page where if we navigated to a section of the `TableOfSection` and them click the link and click the back arrow key the blue marker would be wrong until you reach another section title,
The fix for this was identifying the whole section as a waypoint, by adding a div around it.


jeffchew:
Was looking through the logic, this seems to require the content to be contained in a container to properly get the bounding area, but if the anchor links are siblings it won't work. I'm wondering if we need to calculate based on scroll position of elements and calculating between scroll positions of each anchor.

RaphaelAmadeu:
Closing this for now.

Issue #1668, opened 2020-03-11
Component:  Make dev updates to footer component - legal changes


RobertaJHahn:
#### User Story

> As a `[user role below]`:
IBM.com Library developer

> I need to:
work closely with the Business Analyst and designer to make the Legal changes to the Footer component based on the latest designs and functional specs

> so that I can:
create a component that works as designed

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- Design specs issue (#860)
- Functional specs issue (#1661) and (#1314)
- [Footer functional specs Box link ](https://ibm.ent.box.com/file/639660622895)
 - Design QA Issue (#1670)
 - Prod QA Issue (#1671) 

#### Acceptance criteria
- [ ] Include README for the react pattern and corresponding styles
- [ ] Minimum 80% unit test coverage
- [ ] A comment is posted in the Design QA issue, tagging Wonil and Roberta, when development is finished
 - [ ] A comment is posted in the Prod QA issue, tagging Praveen and Chetan, when development is finished

PR #2200, opened 2020-04-24
feat(footer): Updated with new design specs

RaphaelAmadeu:
### Related Ticket(s)

#1668 

### Description

I've updated the React Footer so it matches the designs specs provided.

OLD: 
Screen Shot 2020-04-24 at 6 33 07 PM

NEW:
Screen Shot 2020-04-24 at 6 32 37 PM


packages/react/src/components/Footer/LegalNav.js
```diff
@@ -76,6 +75,30 @@ function renderListItems(links) {
       </li>
     );
   });
+
+  return renderedLinks.map((elem, index, origin) => {
+    if (index == 0) {
```
jeffchew, line 80: Should use `===` and not `==`

packages/react/src/components/Footer/LegalNav.js
```diff
@@ -76,6 +75,30 @@ function renderListItems(links) {
...
+      return (
+        <ul className={`${prefix}--legal-nav__holder`} key={index}>
+          {elem}
+        </ul>
+      );
+    } else if (index % 2 != 0 && origin[index + 1] != undefined) {
```
jeffchew, line 86: `!==` and not `!=`

packages/react/src/components/Footer/LegalNav.js
```diff
@@ -76,6 +75,30 @@ function renderListItems(links) {
...
+        <ul className={`${prefix}--legal-nav__holder`} key={index}>
+          {elem}
+          {origin[index + 1]}
+        </ul>
+      );
+    } else if (origin[index + 1] != undefined) {
+      return (
+        <ul className={`${prefix}--legal-nav__holder`} key={index}>
+          {origin[index + 1]}
+          {renderTrusteItem()}
```
jeffchew, line 97: Just trying to check the logic on this, the `renderTrusteItem` always gets rendered in any cases? Since this is a legal thing we really need to make sure that this is always here (and only rendered once).

packages/react/src/components/Footer/LegalNav.js
```diff
@@ -76,25 +76,40 @@ function renderListItems(links) {
...
-      );
-    } else if (index % 2 != 0 && origin[index + 1] != undefined) {
+  const chunked_arr = [];
+  chunked_arr.push(renderedLinks[0]);
+  let index = 1;
+  while (index < renderedLinks.length) {
+    chunked_arr.push(renderedLinks.slice(index, 2 + index));
+    index += 2;
+  }
+  console.log('chunked', chunked_arr);
```
asudoh, line 86: Weighing in; Just wondered if this line is meant to be kept.

jeffchew:
The logic seems off, is the first row supposed to be just one item? 

Screen Shot 2020-04-28 at 8 41 27 AM

And just a thought, why not use columns? http://w3schools-fa.ir/cssref/css3_pr_columns.html

RaphaelAmadeu:
@jeffchew 
The first column is supposed to have only one item, as done in the latest design specs.
Screen Shot 2020-04-28 at 10 32 08 AM

jeffchew:
> @jeffchew
> The first column is supposed to have only one item, as done in the latest design specs.
> Screen Shot 2020-04-28 at 10 32 08 AM

hmm, not sure if this is the best idea. If users have their adblocker turned on, then the cookie preferences link isn't there and the way the links will look would be weird. @wonilsuhibm @larahanlon2 @oliviaflory would it be ok to have this treatment be natural column sorting?

cc: @RaphaelAmadeu

larahanlon2:
> > @jeffchew
> > The first column is supposed to have only one item, as done in the latest design specs.
> > Screen Shot 2020-04-28 at 10 32 08 AM
> 
> hmm, not sure if this is the best idea. If users have their adblocker turned on, then the cookie preferences link isn't there and the way the links will look would be weird. @wonilsuhibm @larahanlon2 @oliviaflory @jacobottesen-dgc would it be ok to have this treatment be natural column sorting?
> 
> cc: @RaphaelAmadeu

My suggestion is that the items should be split as evenly as possible, starting on the far-left/first column. In that case, I would position 2 items in the first col, 2 in the second, 1 in the third:

1: Contact IBM, Terms of Use
2: Accessibility, Privacy
3: Cookie Preferences 

If there is an adblocker, then cookie preferences would not appear in the third col, it would just be empty. The cols are always balanced. My two cents.

oliviaflory:
Agreed with @jeffchew and @larahanlon2 suggestion of natural sorting and starting at the far left.

jeffchew:
Thank you @oliviaflory and @larahanlon2 !

@RaphaelAmadeu I think this would simplify the logic, right?

wonilsuhibm:
Sounds good to me too. One minor suggestion: I would group them slightly differently.
1: Contact IBM, Accessibility
2: Terms of use, Privacy
3: Cookie Preferences

oliviaflory:
@RaphaelAmadeu I noticed the interactive states within the Sketch file and PDF are not following the color tokens or color values for the Gray 90 theme and also have outdated HEX values listed. Some of the incorrect specs will not pass accessibility, the design team needs to look into updating the spec file, can you revert any interactive style updates you made? 

Please keep the layout updates to the legal links though!

cc @larahanlon2 @wonilsuhibm @jacobottesen-dgc @jeffchew

packages/react/src/components/Footer/__tests__/Footer.test.js
```diff
@@ -63,5 +63,8 @@ xdescribe('<Footer />', () => {
     expect(footer.querySelectorAll('.bx--legal-nav__list-item')).toHaveLength(
       MOCK_DATA.footerThin.length + 1
     );
+
+    console.log(footer.querySelectorAll('.bx--footer-nav-group'));
```
jeffchew, line 67: @RaphaelAmadeu probably need to remove this console log

packages/react/src/components/Footer/LegalNav.js
```diff
@@ -76,6 +75,24 @@ function renderListItems(links) {
...
     );
   });
+
+  renderedLinks.push(renderTrusteItem());
+
+  const chunked_arr = [];
+  let index = 0;
+
+  while (index < renderedLinks.length) {
+    chunked_arr.push(renderedLinks.slice(index, 2 + index));
```
jeffchew, line 85: @RaphaelAmadeu just thinking about this iteration, if there's for some reason more than 6 links does this then give an unwanted column? I'm wondering if we should instead of have a hardcoded `2`, we should take the length of the `renderedLinks`, and do something like `Math.ceil(renderedLinks.length / 3)`. That way, it would be always 3 columns, but can support an infinite number of legal links. Not saying we'd ever reach that point, but the legal links can also be customized with custom navigation data.

RaphaelAmadeu: @jeffchew , I've applied this logic on the latest PR thank you!

packages/react/src/components/Footer/__tests__/Footer.test.js
RaphaelAmadeu: This was removed on a latest PR

Issue #1952, opened 2020-04-02
Adjust Masthead to accommodate latest carbon-components update to 10.11.0


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

With the latest version of Carbon, the following is happening in the Masthead:

![Screen Shot 2020-04-02 at 11.56.07 AM.png](https://images.zenhubusercontent.com/5cf80f82b494896be7bc635b/4d82d603-aaff-4f6d-ae49-cb03565124f2)

This was highlighted in a previous issue: https://github.com/carbon-design-system/ibm-dotcom-library/issues/1328

And a PR opened by Josh Black shows the deploy preview with the latest release candidate:
https://github.com/carbon-design-system/ibm-dotcom-library/pull/1950

> Is this issue related to a specific component?

Masthead

## Steps to reproduce the issue

1. Go to the deploy preview in https://github.com/carbon-design-system/ibm-dotcom-library/pull/1950
2. View Masthead

jeffchew:
@guilhermelMoraes latest carbon has been updated and can see the issue now: https://ibmdotcom-react-canary.mybluemix.net/?path=/story/components-dotcom-shell--default

cc: @kennylam

PR #2214, opened 2020-04-27
chore(react-package): upgrade the carbon-components version

guilhermelMoraes:
### Related Ticket(s)

[Adjust Masthead to accommodate latest update to 10.11.0 #1952](https://github.com/carbon-design-system/ibm-dotcom-library/issues/1952)

### Description

Following the discussion on the #1952 issue ticket, upgrade the carbon-components version to the `10.11.0-rc.0` would break the `Masthead` due to an css selector mismatching. I did the update, but the problems related in the ticket were not showing up. I did an inspection and there are no problems so far. I would like to make a double check with you all if I'm missing something here. 

#### expected problem:
Captura de Tela 2020-04-27 às 17 04 06

#### current: 
Captura de Tela 2020-04-27 às 17 04 28

### Changelog

**Changed**

- carbon-components upgraded to: `10.11.0-rc.0`;


jeffchew:
Hi @guilhermelMoraes , you would need to update `carbon-components-react` as well. I think in general, we would want to update anything from carbon, at least in the react/styles/vanilla packages.

Let me actually take this upgrade on. We might need to also do some light regression around this.

guilhermelMoraes:
> Hi @guilhermelMoraes , you would need to update `carbon-components-react` as well. I think in general, we would want to update anything from carbon, at least in the react/styles/vanilla packages.
> 
> Let me actually take this upgrade on. We might need to also do some light regression around this.

Ohh, I see. Thanks for clarifying, @jeffchew

jeffchew:
Closing this, have a separate PR open: https://github.com/carbon-design-system/ibm-dotcom-library/pull/2248

PR #2225, opened 2020-04-28
Feat/language selector

jeffchew:
### Related Ticket(s)

https://github.com/carbon-design-system/ibm-dotcom-library/issues/2199

### Description

This adds a flag to be able to utilize a language dropdown instead of the locale button/selector. This swaps out the button with a ComboBox. The only thing missing here is changing the ComboBox direction to top, which appears to be a new feature we need to pull in when upgrading Carbon. But main functionality is all in place for this PR. 

Right now, this is behind a feature flag, so we can at least get this main functionality into the repo and try to address the combobox direction separately.

I also cleaned up a lot of the Footer docs, which looked to be very out of date.

### Changelog

**New**

- Language selector feature in the Footer


packages/styles/scss/components/footer/_language-selector.scss
```diff
@@ -0,0 +1,94 @@
...
+    @include combo-box;
+    @include listbox;
+    @include text-input;
+    @include combo-box-expressive;
+    @include list-box-expressive;
+    @include text-input-expressive;
+    @include carbon--make-col-ready;
+    @include carbon--make-col(4, 4);
+
+    margin-top: carbon--mini-units(2);
```
kennylam, line 31: Just for consistency, we've been using spacing tokens for margins/paddings

```suggestion
    margin-top: $spacing-05;
```

packages/react/src/components/Footer/__data__/language-items.json
```diff
@@ -0,0 +1,38 @@
+[
```
kennylam, line 1: @jeffchew Do we need to include full cc-lc for these IDs?

packages/react/src/components/Footer/LanguageSelector.js
```diff
@@ -0,0 +1,83 @@
...
+  return (
+    <div className={`${prefix}--language-selector__container`}>
+      <ComboBox
+        id="dds-language-selector"
+        data-autoid={`${stablePrefix}--language-selector`}
+        className={`${prefix}--language-selector`}
+        onChange={({ selectedItem }) => _setSelectedItem(selectedItem)}
+        items={items}
+        itemToString={item => (item ? item.text : '')}
+        // initialSelectedItem={initialSelectedItem}
```
kennylam, line 51: @jeffchew Did you mean to comment this out? Clearing the selected language leaves the combobox empty.

packages/react/src/components/Footer/__data__/language-items.json
jeffchew: no, these are only targeting language

packages/react/src/components/Footer/LanguageSelector.js
jeffchew: right, let me uncomment that

Issue #1859, opened 2020-03-26
Pattern: Develop Quote - Prod level version


RobertaJHahn:
#### User Story

> As a `[user role below]`:
IBM.com Library developer

> I need to:
create/change the `Quote`

> so that I can:
provide the ibm.com adopter developers patterns/components they can use to build ibm.com web pages

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- Design specs issue (#2011  )
 - [Visual Designs & Design Specs Box folder link](https://ibm.ent.box.com/folder/110531199928)
- Functional specs issue (#1836  )
- Design QA testing issue (#2159 )
 - Prod QA testing issue (#2160 )
- [Functional specs](https://ibm.ent.box.com/file/653670368937)

#### Acceptance criteria
- [x] Include README for the react pattern and corresponding styles
- [x] Minimum 80% unit test coverage

PR #2257, opened 2020-04-30
feat(pattern): quote pattern created

RaphaelAmadeu:
### Related Ticket(s)

#1859 

### Description

The Quote pattern was created using the design and functional specs

Screen Shot 2020-04-30 at 4 20 25 PM


packages/react/src/patterns/sub-patterns/Quote/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+> 💡 Don't forget to import the quote styles from
+> [@carbon/ibmdotcom-styles](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/styles).
+
+## Props
+
+| Name      | Required | Data Type | Default Value  | Description                                                                                                                                             |
+| --------- | -------- | --------- | -------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `copy`    | YES      | String    | null           | Main Quote                                                                                                                                              |
+| `types`   | NO       | String    | `doubleCurved` | See `Types` bellow.                                                                                                                                     |
```
jeffchew, line 62: minor thing, should it be `type` and not `types`?

RaphaelAmadeu: i've fixed that on the latest commit, good catch

packages/react/src/patterns/sub-patterns/Quote/__stories__/Quote.stories.js
```diff
@@ -0,0 +1,63 @@
...
+storiesOf('Patterns (Sub-Patterns)|Quote', module)
+  .addDecorator(withKnobs)
+  .addParameters({
+    readme: {
+      sidebar: readme,
+    },
+  })
+  .add('Default', () => {
+    const copy = text(
+      'Quote(Required): ',
```
jeffchew, line 24: a format I started following is to include the prop name in parenthesis, might be worth following (Carbon seems to do this too now):

```
const copy = text('quote text (copy)', 'Lorem Ipsum');
const markType = select('quotation mark type (markType):', types, types.doubleCurved);
...etc...
```

RaphaelAmadeu: I've updated it on the latest commit

packages/react/src/patterns/sub-patterns/Quote/Quote.js
```diff
@@ -0,0 +1,145 @@
...
+const { prefix } = settings;
+
+/**
+ * Quote Pattern
+ *
+ * @param {object} props props Object
+ * @param {string} props.markType type of the quote marks
+ * @param {string} props.copy quote copy
+ * @param {object} props.source source object
+ * @param {object} props.source.heading name heading for quote source
```
jeffchew, line 29: @RaphaelAmadeu can you check the types here? I don't think all are objects...

RaphaelAmadeu: @jeffchew I've updated this in the latest commit, Thanks!

PR #2278, opened 2020-05-03
fix(cta): refactor logic out of cta into class

annawen1:
### Related Ticket(s)

https://github.com/carbon-design-system/ibm-dotcom-library/issues/1842

### Description

refactor most of the logic out of the CTA component into `CTALogic` class

### Changelog

**New**

- `useVideoData` custom hook created
- `CTALogic` class created

**Changed**

- sub-components of CTA pulling from `CTALogic` instead of having to pass down from props


packages/react/src/components/CTA/CTA.js
```diff
@@ -48,7 +41,7 @@ const CTA = ({ style, type, customClassName, ...otherProps }) => {
    */
   const getVideoData = useCallback(async () => {
```
jeffchew, line 42: Is there a way to externalize this hook?

annawen1: yep done

packages/react/src/internal/hooks/useVideoData/useVideoData.js
```diff
@@ -0,0 +1,54 @@
...
+  /**
+   * retrieve duration and title information from the video if
+   * the type of the CTA is `video`
+   *
+   * sets the `videoTitle` state with an array of title objects
+   *
+   */
+  const getVideoData = useCallback(async () => {
+    if (type === 'video' || type.includes('video')) {
+      const videoId = CTALogic.getVideoId(style, otherProps);
```
jeffchew, line 36: Hmm having this tied to CTALogic seems off to me, there a way to pass in the videoId?

packages/react/src/components/CTA/CTA.js
```diff
@@ -5,21 +5,13 @@
...
+import React, { useState } from 'react';
 import ButtonCTA from './ButtonCTA';
 import CardCTA from './CardCTA';
 import FeatureCTA from './FeatureCTA';
-import { LightboxMediaViewer } from '../LightboxMediaViewer';
 import PropTypes from 'prop-types';
-import { smoothScroll } from '@carbon/ibmdotcom-utilities';
 import TextCTA from './TextCTA';
-import { VideoPlayerAPI } from '@carbon/ibmdotcom-services';
+import { useVideoData } from '../../internal/hooks/useVideoData';
```
jeffchew, line 14: As discussed, let's move this to `/src/hooks`

packages/react/src/components/CTA/CTA.js
```diff
@@ -32,188 +24,34 @@ import { VideoPlayerAPI } from '@carbon/ibmdotcom-services';
  */
 const CTA = ({ style, type, customClassName, ...otherProps }) => {
   const [renderLightBox, openLightBox] = useState(false);
-  const [videoTitle, setVideoTitle] = useState([{ title: '', key: 0 }]);
-  const [mediaData, setMediaData] = useState({});
 
-  useEffect(() => {
-    getVideoData();
-  }, [getVideoData, style, type]);
+  const videoTitle = useVideoData(otherProps, style, type);
```
jeffchew, line 28: We should just pass the videoId in here, and not have the hook reference CTALogic

packages/react/src/components/CTA/CTA.js
jeffchew: after more discussion, will keep it in `internal`

packages/react/src/internal/hooks/useVideoData/useVideoData.js
annawen1: ok I'll do that

Issue #2071, opened 2020-04-14
Increase unit test coverage for Masthead (React)


jeffchew:
#### User Story

> As a `[user role below]`:
IBM.com Library Developer

> I need to:
have minimum of 80% unit test coverage of the Masthead

> so that I can:
test the functionality of the Masthead via unit tests

#### Additional information

![Screen Shot 2020-04-14 at 9.59.26 AM.png](https://images.zenhubusercontent.com/5cf80f82b494896be7bc635b/7504f995-4aee-4e01-ba20-1d141f11afc3)

- Should decide what to do with the forked UIShell code, perhaps find a way to ignore that set of code in Jest?

#### Acceptance criteria
 - [ ] Minimum 80% unit test coverage

PR #2281, opened 2020-05-04
test(masthead): increase unit test coverage

guilhermelMoraes:
### Related Ticket(s)

Increase unit test coverage for Masthead (React) #2071 

### Description

Based uppon Jest's coverage report I'm working in all the components related to the Masthead itself. 

### Changelog

**New**

- `MastheadTopNav.test.js`, `MastheadLeftNav.test.js`;

**Changed**

 - `MastheadSearch.test.js`;


packages/react/src/components/Masthead/__tests__/MastheadSearch.test.js
```diff
@@ -55,33 +40,30 @@ xdescribe('MastheadSearch', () => {
...
-
     expect(SearchTypeaheadAPI.getResults).toHaveBeenCalled();
   });
 
-  it('should redirect to the results page when a user clicks a suggestion', async () => {
-    const masthead = mount(<MastheadSearch />);
-    const input = masthead.find(
-      `[data-autoid="${stablePrefix}--header__search--input"]`
-    );
+  // it('should redirect to the results page when a user clicks a suggestion', async () => {
```
kennylam, line 46: @guilhermelMoraes These comments should be removed.

packages/react/src/components/Masthead/__tests__/MastheadTopNav.test.js
```diff
@@ -0,0 +1,51 @@
...
+    const { links } = data.mastheadNav;
+    const wrapper = mount(<MastheadTopNav navigation={links} />);
+    const menuItems = links.map((_itens, index) => {
+      return wrapper.find(
+        `a[data-autoid="${stablePrefix}--masthead__l0-nav--nav-${index}"]`
+      );
+    });
+
+    expect(menuItems).toHaveLength(links.length);
+  }, 15000);
```
kennylam, line 34: @guilhermelMoraes What is this for?

guilhermelMoraes: @kennylam this third parameter is a timeout for the `it` function. Since we're fetching the api and not only mocking it, the timeout is an assurance that the test will not fail due to the api takes too long to respond

packages/react/src/components/Masthead/__tests__/MastheadSearch.test.js
guilhermelMoraes: okay, sure.

packages/react/src/components/Masthead/__tests__/MastheadTopNav.test.js
annawen1: hmm you should be able to mock like so:
```
jest.mock('@carbon/ibmdotcom-services', () => ({
  TranslationAPI: {
    getTranslation: jest.fn(() => Promise.resolve(MOCK_DATA)),
  },
}))
```
with `MOCK_DATA` being the example data you see (I pulled this out of Footer.test.js)

This way we don't have to make the actual call

packages/react/src/components/Masthead/__tests__/MastheadTopNav.test.js
```diff
@@ -0,0 +1,58 @@
...
+
+describe('MastheadTopNav', () => {
+  it('renders one MastheadTopNav', () => {
+    const wrapper = mount(<MastheadTopNav navigation={[]} />);
+
+    expect(wrapper).toHaveLength(1);
+  });
+
+  it('renders all the itens based in the `navigation` prop', async () => {
+    const data = await TranslationAPI.getTranslation();
```
annawen1, line 31: sorry looking at this again, I don't think we need to make this call in this test

you can probably just pass the mock data right into the `navigation` prop?

const wrapper = mount(<MastheadTopNav navigation={mockdata.links} />);

guilhermelMoraes: no worries! You're absolutely right! Thanks, Anna! 👍

packages/react/src/components/Masthead/__tests__/MastheadLeftNav.test.js
```diff
@@ -0,0 +1,47 @@
...
+    expect(wrapper).toHaveLength(1);
+  });
+
+  it('receives the `navigation` prop correctly', async () => {
+    const data = await TranslationAPI.getTranslation();
+    const wrapper = mount(
+      <MastheadLeftNav navigation={data.mastheadNav.links} />
+    );
+    expect(wrapper.prop('navigation')).not.toBeUndefined();
+  }, 10000);
```
annawen1, line 28: I think we can do the same here with passing mock data here

jeffchew:
@guilhermelMoraes @annawen1 @kennylam just checking on this, any outstanding items?

packages/react/src/components/Masthead/__tests__/MastheadLeftNav.test.js
jeffchew: @guilhermelMoraes can you take a look?

guilhermelMoraes: yes, I will

PR #2292, opened 2020-05-04
feat(contentblock): added inverse theme

andysherman2121:
### Related Ticket(s)

https://github.com/carbon-design-system/ibm-dotcom-library/issues/2212

Adds the inverse theme for ContentBlock. Select CTA as 'none' in the knobs as the CTA won't be supported in theme for this story


packages/react/src/patterns/sub-patterns/ContentBlock/ContentBlock.js
```diff
@@ -18,7 +18,16 @@ import { settings } from 'carbon-components';
 
 const { stablePrefix } = ddsSettings;
 const { prefix } = settings;
-
+/**
+ *
+ * @param {string} type returns inverse or default
+ * @returns {string} theme classname
+ */
+function themeClassname(type) {
```
jeffchew, line 26: minor thing, let's make this a private function (`_themeClassname`) and add the `@private` to the JSDocs.

packages/react/src/patterns/sub-patterns/ContentBlock/ContentBlock.js
```diff
@@ -18,7 +18,16 @@ import { settings } from 'carbon-components';
...
+  if (type === 'inverse') {
```
jeffchew, line 27: You can make this shorter/terse:
```
return type === 'inverse' ? `${prefix}--content-block-inverse` : `${prefix}--content-block`;
```

packages/styles/scss/patterns/sub-patterns/content-block/_content-block.scss
```diff
@@ -86,6 +72,70 @@
   }
 }
 
+@mixin content-block-inverse {
```
jeffchew, line 75: @andysherman2121 is this a modifier to the main `content-block`? I'm wondering if you should just target what is changing rather than duplicate the other mixin entirely?

andysherman2121:
> Some various comments, one overlying question is how we are handling the inverse class. Seems like we should treat as a modifier and only target specific things with the theme tokens affected.

Updated with requested changes

packages/styles/scss/patterns/sub-patterns/content-block/_content-block.scss
```diff
@@ -11,72 +11,71 @@
...
+    color: #ffffff;
+    padding-top: $carbon--layout-03;
+    padding-bottom: $carbon--layout-05;
+    @include carbon--breakpoint('md') {
+      padding-top: $carbon--layout-05;
+      padding-bottom: $carbon--layout-06;
+    }
+    @include carbon--breakpoint('max') {
+      padding-bottom: $carbon--layout-07;
+    }
```
jeffchew, line 25: @andysherman2121 doesn't seem like any of these are needed for the modifier?

packages/styles/scss/patterns/sub-patterns/content-block/_content-block.scss
```diff
@@ -11,72 +11,71 @@
 @import '../../blocks/feature-card/feature-card';
 
 @mixin content-block {
+  .#{$prefix}--content-block-inverse {
```
jeffchew, line 14: for modifiers, would usually be double dash: `#{$prefix}--content-block--inverse`.

packages/styles/scss/patterns/sub-patterns/content-block/_content-block.scss
andysherman2121: These spacings are all applied to the normal `.#{$prefix}--content-block` class so when you switch to the inverse modifier, it gets rid of necessary spacing

packages/styles/scss/patterns/sub-patterns/content-block/_content-block.scss
andysherman2121: will do

jeffchew: Just a sanity check, should it be `inverse`, or should we be naming based on the theme names? cc: @annawen1 @kennylam

kennylam: That's correct. It should be `g10`, `g90`, or `g100`.

packages/styles/scss/patterns/sub-patterns/content-block/_content-block.scss
andysherman2121: adjusted

packages/styles/scss/patterns/sub-patterns/content-block/_content-block.scss
andysherman2121: Yeah i believe it should be inverse

packages/react/src/patterns/sub-patterns/ContentBlock/ContentBlock.js
```diff
@@ -136,6 +141,7 @@ function _renderCTA(cta) {
 }
 
 ContentBlock.propTypes = {
+  type: PropTypes.oneOf(['__inverse', '']),
```
jeffchew, line 144: Minor thing, doesn't seem like the double underscore looks right here.

andysherman2121: oh yeah I meant to fix that

packages/react/src/patterns/sub-patterns/ContentBlock/__stories__/ContentBlock.stories.js
```diff
@@ -45,13 +47,19 @@ storiesOf('Patterns (Sub-Patterns)|ContentBlock', module)
       none: null,
     };
 
+    const inverse = boolean('inverse', false);
+
     return (
-      <div className="bx--grid">
+      <div
+        className={cx('bx-grid', {
```
asudoh, line 54: Weighing in; Noticed the class name change in the snapshot, just wondered if such change was intentional. Otherwise:

```suggestion
        className={cx('bx--grid', {
```

andysherman2121: good catch, updated

packages/styles/scss/patterns/sub-patterns/content-block/_content-block.scss
```diff
@@ -11,70 +11,61 @@
...
     @include carbon--breakpoint('md') {
       padding-top: $carbon--layout-05;
       padding-bottom: $carbon--layout-06;
     }
-
     @include carbon--breakpoint('max') {
       padding-bottom: $carbon--layout-07;
     }
-
+    //---
```
jeffchew, line 28: Looks good, but might not need this anymore...

packages/react/src/patterns/sub-patterns/ContentBlock/README.md
```diff
@@ -35,6 +35,7 @@ import { ContentBlock } from '@carbon/ibmdotcom-react';
 import 'yourapplication.scss';
 
 function App() {
+  const inverse = boolean('inverse', false);
```
jeffchew, line 38: @andysherman2121 we shouldn't have the `boolean('inverse', false)` here.

PR #2293, opened 2020-05-04
Feat/image with caption inverse

andysherman2121:
### Related Ticket(s)

https://github.com/carbon-design-system/ibm-dotcom-library/issues/2212

Adds the inverse theme for Image with Caption


packages/styles/scss/components/image-with-caption/image-with-caption.scss
```diff
@@ -30,6 +30,34 @@
   }
 }
 
+@mixin image-with-caption-inverse {
```
jeffchew, line 33: Similar to the other PR, I don't think we should duplicate the other mixin entirely, we should treat as a modifier and only target the theme tokens.

Issue #2253, opened 2020-04-30
Pattern: Change the Card section - simple and Card section with images patterns for the Cards Group


RobertaJHahn:
#### User Story

> As a `[user role below]`:
IBM.com Library developer

> I need to:
change the `Card section - simple` and `Card section - with images` patterns`

> so that I can:
provide the ibm.com adopter developers patterns/components they can use to build ibm.com web pages

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
 - Both the `Card section - simple` and `Card section - with images` patterns are used in the `Cards group`
- Functional specs issues (#2252 ) and (#2028)
 - [Card section - simple Box folder link](https://ibm.ent.box.com/folder/99476973715)
 - [Card section - with images box folder link](https://ibm.ent.box.com/folder/102045217811)

#### Acceptance criteria
- [ ] Include README for the react pattern and corresponding styles
- [ ] Minimum 80% unit test coverage

RobertaJHahn:
@kennylam Jeff assigned this issue to you because it is related to the Cards group development that you volunteered to do.

Issue #2029, opened 2020-04-10
Pattern: Develop Cards group - Prod level version


RobertaJHahn:
#### User Story

> As a `[user role below]`:
IBM.com Library developer

> I need to:
create/change the `Cards group`

> so that I can:
provide the ibm.com adopter developers patterns/components they can use to build ibm.com web pages

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- Design specs issue (#2027  )
- Functional specs issue (#2028  )
- [functional specs Box link](https://ibm.ent.box.com/folder/111654108347)

#### Acceptance criteria
- [ ] Create new storybook story
- [ ] create new sandbox for the Cards group
- [ ] export the Cards group properly to as a component
- [ ] export the Cards group styles properly with the styles package
- [ ] Include README for the react pattern and corresponding styles
- [ ] Minimum 80% unit test coverage
- [ ] A comment is posted in the Design QA issue, tagging Wonil and Roberta, when development is finished
 - [ ] A comment is posted in the Prod QA issue, tagging Praveen and Chetan, when development is finished

Issue #1844, opened 2020-03-26
Pattern:  Change the Card section, including creating the new Cards group,  based on the Solution page designs and updated functions specs


RobertaJHahn:
#### User Story

> As a `[user role below]`:
IBM.com Library developer

> I need to:
create/change the `Card section` pattern

> so that I can:
provide the ibm.com adopter developers patterns/components they can use to build ibm.com web pages

![image](https://user-images.githubusercontent.com/3793636/77663222-d4ff8500-6f4a-11ea-9999-fa6340462d70.png)

![image](https://user-images.githubusercontent.com/3793636/77662982-881bae80-6f4a-11ea-9742-58fdc0467776.png)

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- Design specs issue (#2007  )
- Card section Functional specs issue (#1820 )
- Cards group functional specs issue (#1839)
- Design QA testing issue (#2132 )
 - Prod QA testing issue (#2273 )
 -  Changes summary:  Remove the Cards group in Card Section, create a new Cards group pattern, create new cards group story and corresponding documentation, include the logic for the image and simple cards into the Cards group
- Existing Card section is actively being used by another team.  Changes to the Card Section must be transparent to this adopter.

#### Acceptance criteria
- [ ] New Card group sub-pattern
 - [ ] Card section uses the new Card group subpattern
 - [ ] Card group has 2 options:  simple and with image
 - [ ] Ensure the Cards group is properly exported as a component
 - [ ] Ensure the Cards group styles are properly exported
 - [ ] Include a sandbox for the cards group pattern
- [ ] Include README for the react pattern and corresponding styles
- [ ] Minimum 80% unit test coverage
- [ ] A comment is posted in the Design QA issue, tagging Wonil and Roberta, when development is finished
 - [ ] A comment is posted in the Prod QA issue, tagging Praveen and Chetan, when development is finished

jeffchew:
Holding on this for now until we complete discussions around https://github.com/carbon-design-system/ibm-dotcom-library/issues/1855.

jeffchew:
@kennylam this is one of the followup issues after `CardGroup` is created.

PR #2310, opened 2020-05-05
Feat/pattern cards group

kennylam:
### Related Ticket(s)

#1844, #2029, #2253

### Description

This adds the `CardGroup` sub-pattern, which replaces `CardSection`.

### Changelog

**New**

- `CardGroup` sub-pattern and styles

**Changed**

- `CardSectionImages` and `CardSectionSimple` now use `CardGroup` instead of `CardSection`

**Removed**

- `CardSection` component and styles


kennylam:
> 
> 
> @kennylam looks good. One thing we might want to figure out though is when switching card group type under the `CardGroup` pattern to trigger the height detection code so it doesn't render weird when switching or when turning on or off the cta.

@jeffchew Yeah I noticed that as well. I think this has been an issue in a few stories, we should definitely try to get that sorted.

Issue #2212, opened 2020-04-27
Content Block Simple Inverse Theme Dependencies


photodow:
#### User Story
> As a `[user role below]`:
ibm.com library developerr

> I need to:
set inverse themes to the correct content block simple dependencies

> so that I can:
Finish the callout with media stories.

#### Additional information
 - [x] Callout — with media [[directory](https://ibm.ent.box.com/folder/110327422852)]
 - [x] Content block — simple [[code](https://github.com/carbon-design-system/ibm-dotcom-library/tree/master/packages/react/src/patterns/blocks/ContentBlockSimple)] [[directory](https://ibm.ent.box.com/folder/99725042603)] [[storybook](http://ibmdotcom-react.mybluemix.net/?path=/story/patterns-blocks-contentblocksimple--default)]
 - [x] Content block [[code](https://github.com/carbon-design-system/ibm-dotcom-library/tree/master/packages/react/src/patterns/sub-patterns/ContentBlock)] [[directory](https://ibm.ent.box.com/folder/99481428439)] [[storybook](http://ibmdotcom-react.mybluemix.net/?path=/story/patterns-sub-patterns-contentblock--default)]
 - [x] Video player [[code](https://github.com/carbon-design-system/ibm-dotcom-library/tree/master/packages/react/src/components/VideoPlayer)] [[directory](https://ibm.ent.box.com/folder/108837570522)] [[storybook](http://ibmdotcom-react-experimental.mybluemix.net/?path=/story/components-videoplayer--default)]
 - [x] Image with caption [[code](https://github.com/carbon-design-system/ibm-dotcom-library/tree/master/packages/react/src/components/ImageWithCaption)] [[directory](https://ibm.ent.box.com/folder/109643123498)] [[storybook](http://ibmdotcom-react.mybluemix.net/?path=/story/components-imagewithcaption--default)]
 - Feature Card block - medium, Button Group and Link with icon are out of scope for this work.  Separate issues for each are created and linked to this issue.

![image](https://user-images.githubusercontent.com/3793636/80396268-0507b400-887a-11ea-8bc0-3d75cfd750df.png)

So in order to use content block simple with the upcoming callout pattern we need to enable inverse theming for it. This issue handles that and all of its relevant dependencies necessary to theme it correctly.

> 👀Check out `Callout — with media`, and `Content block — simple` for design specs.

Reach out if you have any questions.

#### Acceptance criteria
 - [x] Call out with media themed correctly using inverse tokens.

PR #2311, opened 2020-05-05
Feat/videoplayer inverse

andysherman2121:
### Related Ticket(s)

#2212

Adds the inverse theme for Videoplayer


packages/react/src/components/VideoPlayer/VideoPlayer.js
```diff
@@ -17,10 +18,13 @@ const { prefix } = settings;
 /**
  * VideoPlayer component
  *
- * @param {string} videoId Kaltura video id
+ * @param {object} props props object
+ * @param {string} props.type theming options
```
jeffchew, line 22: General question, is `type` the right prop name for this or we want to rename to something else?

cc: @andysherman2121 @kennylam @reesedownes @annawen1 @RaphaelAmadeu @guilhermelMoraes

annawen1: should we just have it as a boolean called `inverse`?

packages/styles/scss/components/video-player/_video-player.scss
```diff
@@ -38,6 +38,12 @@
       }
     }
   }
+  .#{$prefix}--video-player--inverse {
+    background-color: $inverse-02;
+  }
+  .#{$prefix}--video-player__video-description--inverse {
+    color: #ffffff;
```
annawen1, line 45: can we use a token for this?

andysherman2121: updated with token

packages/react/src/components/VideoPlayer/VideoPlayer.js
andysherman2121: Updated to `inverse`

packages/react/src/components/VideoPlayer/VideoPlayer.js
```diff
@@ -32,10 +41,23 @@ const VideoPlayer = ({ videoId, showDescription }) => {
...
+    `${prefix}--video-player`,
+    { [`${prefix}--video-player--inverse`]: inverse },
+    customClassName
+  );
+  const textclass = cx(
+    `${prefix}--video-player__video-description`,
+    {
+      [`${prefix}--video-player__video-description--inverse`]: inverse,
+    },
+    customClassName
```
annawen1, line 54: I don't think we need to add inverse modifier for the text. We can just use the 

`.#{$prefix}--video-player--inverse .#${prefix}--video-player__video-description {}` selector instead in the css

andysherman2121: updating

Issue #2179, opened 2020-04-23
Fix circular dependency in Card / CTA


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

There is a circular dependency identified between the `Card` and `CTA`. We should extract out the `CTA` from `Card` and perhaps reference the `LinkWithIcon` directly in `Card` instead.

If this isn't causing the Storybook rendering issues in https://github.com/carbon-design-system/ibm-dotcom-library/issues/2178, would like to address this as a high priority item in Sprint 9.

> Is this issue related to a specific component?

`Card` and `CTA`

> What did you expect to happen? What happened instead? What would you like to
> see changed?

No circular dependencies

## Additional information

- Would also like to set the propType share for `CTA` to pull in the actual propType from the component. Right now, it is `cta: PropType.object`, but should instead be something like `cta: PropTypes.oneOfType(PropTypes.shape(CTA.propTypes)).isRequired`. This throws an error currently since there is a circular dependency, and thinks that CTA appears to not be defined because of this.

PR #2326, opened 2020-05-06
fix(pattern): fixed circular dependency on card pattern

RaphaelAmadeu:
### Related Ticket(s)

#2179 

### Description

In order to fix a circular dependency issue between `CTA` and `Card`, I've replaced the usage of `CTA` for `LinkWithIcon` in the `Card` pattern.


packages/react/src/patterns/sub-patterns/Card/Card.js
```diff
@@ -89,7 +89,9 @@ function renderFooter(cta, type) {
     cta && (
       <div className={`${prefix}--card__footer`}>
         {type !== 'link' ? (
-          <CTA style="text" {...cta} customClassName={`${prefix}--card__cta`} />
+          <LinkWithIcon href={cta.href}>
+            <span>{cta.copy}</span> <cta.icon.src />
+          </LinkWithIcon>
```
annawen1, line 94: I think what we also need to include here is the cta `type` (ie. local, external, jump) that gets lost when we replace with LinkWithIcon. We might need to build that functionality into LinkWithIcon and use the functions from `CTALogic` class (still waiting for that PR to be merged rn)

jeffchew: PR is merged in, I think it makes sense to do this approach.

RaphaelAmadeu: @jeffchew , @annawen1 
I've tried to implement the `CTALogic` methods on my latest commit

annawen1: hmm I was expecting the changes to be done on `LinkWithIcon` as well

RaphaelAmadeu: @annawen1 I've updated this PR, now placing the logic on `LinkWithIcon`, thanks!

jeffchew:
@RaphaelAmadeu can you update your snapshot file?

packages/react/src/components/LinkWithIcon/LinkWithIcon.js
```diff
@@ -20,14 +21,25 @@ const { prefix } = settings;
  * @param {object} props react proptypes
  * @param {object} props.children User content
  * @param {string} props.href Link url
+ * @param {string} props.type Link type
  * @returns {*} LinkWithIcon component
  */
-const LinkWithIcon = ({ children, href, ...props }) => {
+const LinkWithIcon = ({ children, href, type, ...props }) => {
```
jeffchew, line 27: @RaphaelAmadeu @annawen1 I'm not sure if adding to here makes sense or not? If we do decide to do it this way, we would also need to set the icon by type as well which doesn't seem to be the case here. Perhaps we should only keep this on the `Card` level for now?

packages/react/src/patterns/sub-patterns/Card/Card.js
```diff
@@ -85,15 +96,16 @@ function optionalContent(copy) {
  * @returns {object} JSX object
  */
 function renderFooter(cta, type) {
+  const Icon = CTALogic.iconSelector(cta.type);
   return (
     cta && (
       <div className={`${prefix}--card__footer`}>
         {type !== 'link' ? (
-          <CTA style="text" {...cta} customClassName={`${prefix}--card__cta`} />
+          <LinkWithIcon type={cta.type} href={cta.href}>
```
jeffchew, line 104: @RaphaelAmadeu see comment at https://github.com/carbon-design-system/ibm-dotcom-library/pull/2326/files#r425952465. If we're going to only keep this on the `Card` level, you can just pass in the `onClick` behavior here.

cc: @annawen1

packages/react/src/patterns/sub-patterns/Card/Card.js
jeffchew: Sorry if the above wasn't too clear, I've made a couple of comments above but I'm wondering if we should keep `LinkWithIcon` as-is, but we can pass in the CTA logic on this level?

Issue #1224, opened 2020-01-28
[Automation] Investigation of automatic visual testing


jeffchew:
#### User Story

> As a `[user role below]`:
IBM.com Library developer

> I need to:
enable automated visual testing of the IBMN.com Library

> so that I can:
be consistently aware of any visual impacts from changes in the codebase

#### Additional information

 - Storybook automated visual testing: https://storybook.js.org/docs/testing/automated-visual-testing/
 - Storybook screenshot testing: https://medium.com/bleeding-edge/screenshot-testing-with-react-and-storybook-19ab7e49ec92
 - Visual validation tools: https://testguild.com/top-21-free-visual-validation-tools-testers/
- Include any investigations for cross-browser/device testing to be included in this (would we need the API level version of Browserstack to support this?)

#### Acceptance criteria
 - [ ] Determine the best approach in implementation of automated visual testing
 - [ ] Ideally will have a rough implementation, priority would be the Carbon Expressive storybook
 - [ ] Documentation creation

jeffchew:
Followup meeting with Browserstack scheduled on March 12th.

jeffchew:
Browserstack meeting rescheduled to March 31st.

jeffchew:
Browserstack meeting rescheduled to April 21st at 9am ET.
cc: @RobertaJHahn @Praveen-sr @chetankan

chetankan:
@jeffchew : Thanks for update.

PR #2332, opened 2020-05-07
Feat/localemodal staticdata

jeffchew:
### Related Ticket(s)

#1224 

### Description

This updates the `LocaleModal` to pass in static data for the navigation. I'm actually thinking of having this be undocumented as we may not want to have adopters abuse this, but we need this for rendering for snapshot testing.

### Changelog

**New**

- static data fields for the locale modal


packages/react/src/components/LocaleModal/LocaleModal.js
```diff
@@ -41,9 +43,16 @@ const LocaleModal = ({ isOpen, setIsOpen }) => {
...
-      const list = locale && (await LocaleAPI.getList(locale));
-      const getLangDisplay = await LocaleAPI.getLangDisplay();
+      let list, getLangDisplay;
+      if (localeData && localeDisplay) {
+        list = Object.assign({}, localeData);
+        getLangDisplay = localeDisplay;
+      } else {
+        const locale = await LocaleAPI.getLocale();
+        list = locale && (await LocaleAPI.getList(locale));
+        getLangDisplay = await LocaleAPI.getLangDisplay();
```
asudoh, line 53: Weighing in; Just wondered if avoiding critical path with sequential promise refs would be desirable (if not my apologies):

```suggestion
        const pair = await Promise.all([LocaleAPI.getLocale(), LocaleAPI.getLangDisplay()]);
        const locale = pair[0];
        list = locale && (await LocaleAPI.getList(locale));
        getLangDisplay = pair[1];
```

jeffchew: @asudoh thank you for the suggestion! Just added it.

Issue #1852, opened 2020-03-26
Pattern: Develop CTA section - Prod level version


RobertaJHahn:
#### User Story

> As a `[user role below]`:
IBM.com Library developer

> I need to:
create/change the `CTA section`

> so that I can:
provide the ibm.com adopter developers patterns/components they can use to build ibm.com web pages

![image](https://user-images.githubusercontent.com/3793636/77662540-ff9d0e00-6f49-11ea-8fcb-a42f163ec28d.png)

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- Design specs issue (#2031  )
- Functional specs issue (#1829  )
- [Design and Functional specs box folder](https://ibm.ent.box.com/folder/109648616047)
- Design QA testing issue (#2113 )
 - Prod QA testing issue (#2114 )
- Summary:  uses CTA block, supports 4 themes

#### Acceptance criteria
- [x] Create new storybook story
- [ ] create new sandbox for the CTA section
- [x] export the CTA section properly to as a component
- [x] export the CTA section styles properly with the styles package
- [x] Include README for the react pattern and corresponding styles
- [x] Minimum 80% unit test coverage
- [x] A comment is posted in the Design QA issue, tagging Wonil and Roberta, when development is finished
 - [x] A comment is posted in the Prod QA issue, tagging Praveen and Chetan, when development is finished

PR #2336, opened 2020-05-08
Feat/pattern cta section

guilhermelMoraes:
### Related Ticket(s)

Pattern: Develop CTA section - Prod level version #1852

### Description

Created the new pattern, the `CTASection`

### Changelog

**New**

- Pattern > Section > CTASection;
- Styles > Patterns > Sections > CTASection;


packages/react/src/patterns/sections/CTASection/CTASection.js
```diff
@@ -0,0 +1,59 @@
+import { ContentBlock } from '../../sub-patterns/ContentBlock';
+import { ContentItem } from '../../sub-patterns/ContentItem';
```
annawen1, line 2: I think you missed exporting the component in `.../packages/react/src/patterns/sections/index.js`

guilhermelMoraes: Fixed, @annawen1

packages/react/src/patterns/sections/CTASection/__stories__/CTASection.stories.js
```diff
@@ -0,0 +1,69 @@
...
+    },
+  })
+  .add('Default', () => {
+    const types = ['local', 'external'];
+    const ctaProps = {
+      style: 'button',
+      type: types[0],
+      buttons: [
+        {
+          type: select('ContentBlock | CTA type', types, types[0]),
```
annawen1, line 22: one thing, I think we want to follow the same format for other stories where we write the actual prop name in parenthesizes like so:

'Content Block CTA (required(?)) (cta.type):'

PR #2342, opened 2020-05-11
Fix/storybook dupe css

kennylam:
### Related Ticket(s)

#409

### Description

This removes all story-specific css from Storybook stories in favor of a global import. This is necessary as importing in each story bypasses the `import-once` mechanism which results in large amount of duplication.

Changing to this method of imports dramatically increased css load/HMR times, so [fast-sass-loader](https://www.npmjs.com/package/fast-sass-loader) is used to optimize performance. Overall load/HMR times should be much lower

### Changelog

**New**

- `fast-sass-loader` for Storybook webpack

**Changed**

- `ibm-dotcom-styles.scss` now imported globally in Storybook

**Removed**

- story-specific style imports and scss files


packages/react/.storybook/webpack.config.js
```diff
@@ -124,12 +107,52 @@ module.exports = ({ config, mode }) => {
...
+  const sassLoader = {
+    loader: 'sass-loader',
+    options: {
+      includePaths: [
+        path.resolve(__dirname, '..', 'node_modules'),
+        path.resolve(__dirname, '../../../', 'node_modules'),
+      ],
+      data: `
+        $feature-flags: (
+          ui-shell: true,
```
asudoh, line 119: Just wanted to let you know that `ui-shell` feature flag is no longer needed.

kennylam: Thanks @asudoh! I'll get rid of that 👍

Issue #2313, opened 2020-05-05
[VideoPlayer]: Caption should pull from video title not description


photodow:
## Detailed description

As of right now the video player is pulling from the video's description to populate the caption. I believe it should be displaying the video's title instead. The functional specs should back this up.

If you look at the image below the title of the video is different from the caption below.

##### As-is
![image](https://user-images.githubusercontent.com/3793636/81104629-6d890d80-8ed8-11ea-8ade-843b80dd5e85.png)

##### Functional spec screenshot
![image](https://user-images.githubusercontent.com/3793636/81104754-9a3d2500-8ed8-11ea-8e5e-7c6a2f55ccb6.png)

![image](https://user-images.githubusercontent.com/3793636/81104730-91e4ea00-8ed8-11ea-9026-004f8dae5d43.png)

## Additional information

- [Anna's confirmation](https://github.com/carbon-design-system/ibm-dotcom-library/issues/2312#issuecomment-624237455)
- [functional specs](https://ibm.box.com/s/drvsm6ez6kr206csllb5z1zvoxl11i62)

PR #2363, opened 2020-05-12
fix(video): use video name for caption

kennylam:
### Related Ticket(s)

#2313 

### Description

Update `VideoPlayer` to use `name` for caption

### Changelog

**Changed**

- video caption now uses `name` instead of `description`


packages/react/src/components/VideoPlayer/__stories__/VideoPlayer.stories.js
```diff
@@ -38,7 +38,7 @@ storiesOf('Components|VideoPlayer', module)
             <VideoPlayer
               inverse={inverse}
               videoId="0_uka1msg4"
-              showDescription={boolean('Show description', true)}
+              showCaption={boolean('Show caption', true)}
```
jeffchew, line 41: Adding prop name in the description:

```suggestion
              showCaption={boolean('Show caption (showCaption)', true)}
```

PR #2368, opened 2020-05-13
chore(components): avoid reference to stale async results

asudoh:
### Related Ticket(s)

(Please let me know if I should enter a ticker for this kind of change)

### Description

This change uses `useEffect()`'s "clean-up" API to detect if another `useEffect()` call will come or the component is being unmounted, to see if a return value from an asynchronous call is stale.

This ensures only one asynchronous call is in-flight, and avoids React's `Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.` error.

### Changelog

**New**

- Code to detect stale state in `useEffect()`.


packages/react/src/components/Footer/Footer.js
```diff
@@ -59,24 +59,40 @@ const Footer = ({
...
     }
+    return () => {
+      stale = true;
+    };
   }, [navigation]);
 
   useEffect(() => {
+    let stale = false;
     (async () => {
       const response = await LocaleAPI.getLangDisplay(langCode);
```
jeffchew, line 80: We'll need to catch stale for this one too.

asudoh, line 80:
```suggestion
      const response = await LocaleAPI.getLangDisplay(langCode);
      if (stale) {
         return;
      }
```

Issue #2303, opened 2020-05-05
Video Component light box media viewer does not work for many Kaltura Videos


kexuibm:
The Lightbox Media Viewer for Video does not seem to be working for many Kaltura Video IDs.

For example the following videos are publically available on Media Center but does not work when loaded from the Media Viewer:

https://mediacenter.ibm.com/media/1_2cuh9fvy

annawen1:
we need to update the `partnerId` and `uiconfiId` to match the ones from `v18`. The media center videos seem to be authorized to play only on ibm.com domains as well.

ljcarot:
Can we make this high severity - like sev 2?

jeffchew:
> Can we make this high severity - like sev 2?

@ljcarot This is fine, though we can actually configure partnerID and confID through environment variables, but may need to update documentation on that. 

@kexuibm , here are the partner ID/confID's for IBM prod, but you won't be able to see the videos render in lower environments unless you are pointing to `*.ibm.com`:

```
KALTURA_PARTNER_ID=1773841
KALTURA_UICONF_ID=27941801
```

We will update the Kaltura player defaults to IBM prod, but since this is already configurable via environment variables we don't think there is a need to patch this.

cc: @annawen1

kexuibm:
@jeffchew  @RobertaJHahn  I think this defect needs to be reopened.  So after plugging in the production Kaltura Partner and UI Conf IDs, the video itself loads in the light box however the attending title, description, and duration metadata does not load. 

Upon closer inspection, I noticed that the following Kaltura API calls is failing with an Invalid Widget ID error.
https://cdnapisec.kaltura.com/api_v3/index.php?service=multirequest&apiVersion=3.1&expiry=86400&clientTag=kwidget%3Av2.82.1&format=1&ignoreNull=1&action=null&1:service=session&1:action=startWidgetSession&1:**widgetId=1773841**&2:ks=%7B1%3Aresult%3Aks%7D&2:service=media&2:action=get&2:entryId=1_r1eanmna&kalsig=2c9c034a8648a8ae777bdc564dc2574c

It seems like the widgetId parameter in this call needs to be formatted to be _1773841 rather than just 1773841.

Here is what the Link List and Video in Light Box looks like when this happens, note that title, description, and duration is missing.

[...]

PR #2374, opened 2020-05-13
fix(videoplayer): change partnerid format for metadata api call

annawen1:
### Related Ticket(s)

Video Component light box media viewer does not work for many Kaltura Videos #2303

### Description

Metadata call for video fails when using production kaltura video id. 

The partnerid needs to be passed in with a `_` in front. The description for the video comes back with html tags which we need to remove 

EXPECTED:
Screen Shot 2020-05-14 at 8 31 01 AM

Screen Shot 2020-05-14 at 8 35 15 AM

### Changelog

**New**

- add functionality in `markDowntoHtml` util to return just the string

**Changed**

- use `markDowntoHtml` in LightboxMediaViewer and VideoPlayer for the video description


packages/utilities/src/utilities/markdownToHtml/markdownToHtml.js
```diff
@@ -58,10 +58,13 @@ function markdownToHtml(
     bold = true,
     useCarbonClasses = true,
     allowHtml = false,
+    cleanString = false,
+    textOnly = false,
```
jeffchew, line 62: @annawen1 can you add JSDocs for this?

packages/utilities/src/utilities/markdownToHtml/markdownToHtml.js
```diff
@@ -87,7 +90,7 @@ function markdownToHtml(
       });
     }
 
-    paraList += `<p>${para}</p>`;
+    paraList += textOnly ? para : `<p>${para}</p>`;
```
jeffchew, line 93: @annawen1 `textOnly`, I'm wondering, if this is the right name for this. This is pretty much removing the paragraph functionality, right? Maybe we can call it `createParagraphs` and have it be default as true?

annawen1: ok!

packages/react/src/components/VideoPlayer/VideoPlayer.js
```diff
@@ -55,7 +58,12 @@ const VideoPlayer = ({ inverse, showCaption, videoId, customClassName }) => {
       </div>
       {showCaption && (
         <div className={`${prefix}--video-player__video-caption`}>
-          {videoData.name} {videoDuration}
+          {videoData.description &&
+            markdownToHtml(videoData.description, {
+              textOnly: true,
```
jeffchew, line 63: @annawen1 you will need to update this to `createParagraphs: false`

annawen1: ack right

packages/react/src/components/VideoPlayer/VideoPlayer.js
```diff
@@ -55,7 +61,7 @@ const VideoPlayer = ({ inverse, showCaption, videoId, customClassName }) => {
       </div>
       {showCaption && (
         <div className={`${prefix}--video-player__video-caption`}>
-          {videoData.name} {videoDuration}
+          {videoDesc} {videoDuration}
```
kennylam, line 64: @annawen1 The caption should actually use `videoData.name`. This was fixed in #2363

annawen1: oh whoops - didn't notice that when i was dealing with the merge conflicts, my bad!

Issue #2361, opened 2020-05-12
[Image with caption] Launching lightbox media viewer


photodow:
#### User Story
> As a `[user role below]`:
Drupal adopter

> I need to:
be able to launch a lightbox media viewer from image with caption

> so that I can:
meet the needs of our authors

#### Additional information
 - [Design specs](https://ibm.box.com/s/36x742wza48h9zl5p3jfkq55me0fbcij)
 - [Functional specs](https://ibm.box.com/s/d0zlq6e6cu6s5l7szv4rxvh6hx3iz1ru)

#### Acceptance criteria
 - [ ] Image with caption optionally launches lightbox media viewer on click
 - [ ] User interactions when clickable
 - [ ] Icon to indicate it launches
 - [ ] whole image is optionally clickable
 - [ ] Caption is same as lightbox media viewer's heading

PR #2413, opened 2020-05-15
feat(image-with-caption): launch lightbox for image with caption

annawen1:
### Related Ticket(s)

[Image with caption] Launching lightbox media viewer #2361

### Description

Add functionality to launch LightboxMediaViewer on click of ImageWithCaption

May-15-2020 12-57-31

### Changelog

**New**

- add `lightbox` boolean prop to determine whether to make image clickable
- add optional `copy` prop for the light box description


packages/react/src/components/ImageWithCaption/__stories__/ImageWithCaption.stories.js
```diff
@@ -52,7 +52,12 @@ export const Default = () => {
       })}>
       <div className="bx--row">
         <div className="bx--col-sm-4 bx--col-lg-8 bx--offset-lg-4">
-          <ImageWithCaption inverse={inverse} image={image} heading={heading} />
+          <ImageWithCaption
+            inverse={inverse}
+            image={image}
+            heading={heading}
+            lightbox={boolean('lightbox', true)}
```
jeffchew, line 59: @annawen1 I think you're missing the `copy` prop here

packages/react/src/components/ImageWithCaption/ImageWithCaption.js
```diff
@@ -51,8 +85,16 @@ const ImageWithCaption = ({ inverse, image, heading, customClassName }) => {
 ImageWithCaption.propTypes = {
   inverse: PropTypes.bool,
   image: PropTypes.shape(Image.propTypes).isRequired,
+  lightbox: PropTypes.bool,
   heading: PropTypes.string.isRequired,
+  copy: PropTypes.string,
```
jeffchew, line 90: @annawen1 can you add to the README?

packages/react/src/components/ImageWithCaption/__stories__/ImageWithCaption.stories.js
annawen1: ah yep, I actually don't know where the `copy` text is supposed to go visually

@wonilsuhibm @larahanlon2 I see in the [func specs](https://ibm.ent.box.com/file/640289462720) in the `Caption` section that in the @sm breakpoint we show both `heading` (the caption that we have currently showing) and `copy` however I don't see how the `copy` is supposed to appear in the [visual specs](https://ibm.ent.box.com/file/640317162531) (there aren't any mobile specs). Can you provide clarification on this?

packages/react/src/components/ImageWithCaption/ImageWithCaption.js
annawen1: done

packages/react/src/components/ImageWithCaption/__stories__/ImageWithCaption.stories.js
larahanlon2: Hey @annawen1 I took a look at the `image with caption` and `lightbox` specs. I'm a little confused about the `copy` prop. For `image with caption` it seems that it should be an option in addition to the `heading` but I'm not seeing any visual spec for that, like you. But I also wonder why the copy would be included in `image with caption` component. Is it different to `lightbox` description field or the same just a different style? I can check with Lila otherwise we may need to wait for Wonil.

jeffchew: Is the copy going into the lightbox?

annawen1: yep the copy goes into lightbox. BUT according to the func specs, the lightbox functionality is only available for md to larger breakpoints. In mobile breakpoint, the user doesn't have the option launch the lightbox and they instead see the image, heading, and copy.

Issue #2421, opened 2020-05-18
Pattern: Card Section Simple - User cannot identify the link arrow icon on theme g90 and g100. 


Praveen-sr:
## Detailed description

Issue:- On g90 and g100 theme, user is not able to identify the link arrow icon, as its color is matching with the background theme.

**This bug has been prioritized to be included in the May release and must be finished before the code freeze EOD Monday, May 18.**

Note :- This link icon is in black color on all the card sections, it must be in blue color.

## Additional information
Actual :-

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/f03cb4c1-6532-4cbc-8f87-773dbc844249)

Expected :-
![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/5ee85fff-5929-4391-9002-723f53bf1474)

PR #2428, opened 2020-05-18
fix(card): change icon color for card back to blue and use icon passed in from props

annawen1:
### Related Ticket(s)

Pattern: Card Section Simple - User cannot identify the link arrow icon on theme g90 and g100. #2421

### Description

Style class that sets the icon color needed to be applied to the icon. Also use the icon passed in from props

EXPECTED:
Screen Shot 2020-05-18 at 1 43 08 PM

### Changelog

**Changed**

- add style class back to icon component
- use `cta.icon.src` instead of the function from `CTALogic`

### DIFF
--- a/packages/react/.storybook/_container.scss
+++ b/packages/react/.storybook/_container.scss
@@ -35,5 +35,11 @@
 
 // hide the cookie button
 #teconsent {
-  display: none;
+  visibility: hidden;
+}
+
+@media only percy {
+  [data-autoid='dds--privacy-cp'] {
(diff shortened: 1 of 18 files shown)


jeffchew:
@annawen1 just doublechecking the alignment of the FeatureCard cta, spotted in visual regression: https://percy.io/IBM-Digital-Design/IBM.com-Library-React/builds/5299256/view/311366873/1280?mode=diff&browser=chrome&snapshot=311366873

Issue #2461, opened 2020-05-20
Build integration test


asudoh:
### The problem

Efforts around #2198 and #2436 revealed that there are some kind of issues that doesn't surface in our Storybook environment and surfaces only with running our build in application environment.

### The solution

#2198 keeps track of the effort that creates an environment that emulates certain kind of applications. We can catch the error automatically by integration tests that runs such emulated application environment.

We could use Selenium/Webdriver for this kind of tests, but it's barely suitable for CI integrated to PRs, because they tend to take long time and to sometimes pass and sometimes fail both due to WebDriver nature. ([Ref](https://github.com/w3c/webdriver/blob/master/webdriver-bidi/webdriver.md))

Given errors in build does not have to be multiple-browser tests, we can use Pupeteer for this. We can potentially use [Playwright](https://github.com/microsoft/playwright) anytime we see the need for multiple-browser tests. Pupeteer/Playwright APIs are almost identical.

jeffchew:
Lining up for Sprint 12.

cc: @RobertaJHahn

jeffchew:
@asudoh actually, is the scope of this story the PR we have already merged, or for full coverage of the package?

asudoh:
You are right @jeffchew that #2462 covers the initial work. So I think it's good to close this and queue up any follow-up works as separate issue as they arise.

Issue #2438, opened 2020-05-19
React: Optimize build


asudoh:
## Detailed description

> Describe in detail the issue you're having.

An attempt to create an CodeSandbox example that contains only the masthead (https://github.com/carbon-design-system/ibm-dotcom-library/compare/master...asudoh:codesandbox) revealed that the (minified) build size is more than 6MB and building it online (in CodeSandbox server) is extremely slow.

It can be observed by checking out above branch, going to `packages/react/examples/codesandbox/masthead` directory, and running `yarn install && yarn build`.

Some attempts to bi-sect for the culprit revealed that stopping relying on tree-shaking of `@carbon/icons-react` library and potentially `carbon-components-react` library reduces the size (and thus better time spent on build).

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

A general issue?

> Is this issue related to a specific component?

No.

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Smaller build size (<1M in above CodeSandbox example), faster build.

> What version of the IBM.com Library are you using?

Latest `master`.

PR #2462, opened 2020-05-20
Test integration build

asudoh:
### Related Ticket(s)

Refs #2461.

### Description

This change introduces build integration tests, that runs application examples with Pupeteer. It intends to catch errors in our build artifacts that may not surface in our Storybook environment.

### Changelog

**New**

- CodeSandbox example for masthead (may not work live in CodeSandbox due to #2438, but works with local environment or the test automation environment here).
- The test automation that tests above CodeSandbox example.


packages/react/examples/codesandbox/masthead/src/App.js
```diff
@@ -0,0 +1,26 @@
+/**
```
jeffchew, line 1: Hey @asudoh , looks good! For the examples, is the idea that this doubles as the example codesandbox for adopters?

cc: @kennylam @RaphaelAmadeu

asudoh: Thanks @jeffchew, you are right that the example CodeSandbox can serve for documentation purpose, too.

jeffchew: For codesandbox I think we'd want the async call in place rather than bypassing with static data, I guess maybe we can have one example with static data that's used for integration testing, and another default one that uses the service call?

asudoh: @jeffchew Updated the CSB example not to use the mock data. And I may have been wrong that CSB code hosted on GH cannot be configured with environment variable... I could do it for my fork with my GH account. Let's try with our account once we get this PR in master.

jeffchew:
LGTM! There are merge conflicts however!

asudoh:
Thanks @jeffchew! Resolved the conflict.

Issue #2203, opened 2020-04-27
Components: Masthead :- Navigation links are getting displayed below the platform name.


chetankan:
Issue description:- In masthead- default component on selecting 'Platform' in knobs section navigation links are getting displayed under platform name section. 

## Steps to reproduce the issue

1. Enter url in the browser.
2. Select 'Default' in masthead component.
3. Select 'Platform Name --> Platform' in knobs.
4. Check where navigation links are displayed.

OS:- Windows 10, Mac
Browsers:- Chrome, Firefox, Safari

URL:- https://ibmdotcom-react-staging.mybluemix.net/?path=/story/components-masthead--default

![Navigation links.png](https://images.zenhubusercontent.com/5e21a841639f517d456f7e43/0f90f86f-d9ae-43f0-bedb-f3df954fec26)

PR #2465, opened 2020-05-20
fix(css): fix masthead nav styles

kennylam:
### Related Ticket(s)

#2203 

### Description

This fixes `masthead` styles to address layout issues when platform name is included in the nav.

### Changelog

**Changed**

- fix masthead styles for when platform name is displayed
- update license text


kennylam:
> Looks good. I wonder if we should create platform name as a separate story in storybook so percy can pick this up in the future.

@jeffchew Sure, good idea.

Issue #2422, opened 2020-05-18
Investigate usage of React DocGen tool for documenting React component props


asudoh:
### The problem

We are using JSDoc for documenting React component properties. However, JSDoc is not aware of React `prop-types` syntax (i.e. `PropTypes.bool.isRequired`)

### The solution

Switch to `react-docgen`

PR #2480, opened 2020-05-22
docs(components): introduce Storybook `addon-docs`

asudoh:
### Related Ticket(s)

Refs #2422.

### Description

This change adds Storybook `addon-docs` to our React storybook environment, starting with `<CTA>` component.

`addon-docs` harvests React DocGen data from our components and allows us to put in our doc. To use that feature, this change converts the `README` to `.mdx` and replaces the properties table in the `README` with the auto-generated one (`<Props of={CTA} />`).

This change also disables `jsdoc/require-jsdoc` ESLint rule for our React codebase, and introduces a custom ESLint rule to check for React prop types comment. Given we haven't completed the transition, the new ESLint rule is set as a warning (instead of an error) for now.

image

### Changelog

**New**

- `addon-docs` to our React Storybook environment.
- A new custom ESLint rule to check for React prop type comments.

**Changed**

- `README` to `README.stories.mdx` for `<CTA>`.
- Replaces prop type documentation in `README.stories.mdx` with auto-generated one `<Props>`.

**Removed**

- `jsdoc/require-jsdoc` rule from our React codebase.


jeffchew:
@asudoh this looks great! I like the Docs tab better than the README tab, which I'm assuming we can completely switch over to once we migrate the rest of the components/patterns to that. Do you have a sense of how much effort it will be to migrate over?

jeffchew:
@asudoh also, I know we originally talked about this in huddle, but this would be good to have as a followup topic and walkthrough of the changes.

asudoh:
Thanks @jeffchew! For migration, I think it'll take 30 minutes per story. Let me queue up this topic to huddle.

Issue #2016, opened 2020-04-09
Button group | 320px fixed button width and responsive setup need adjustments


sethibm:
## Detailed description

> Describe in detail the issue you're having.

320px fixed button width is incorrect, per latest visual spec.  The buttons should flex per the internal content, with 16px padding to left of the text, and 64px spacer to right of the text in button, up to a max 320px button width.  

Also, a related bug, per the visual spec, needs to react to total width of the button group, as related to the container or available space, for when the buttons stack, rather than sit left to right.  The rule: when container or available space is less than the sum of all button widths + the padding between the buttons, all of the buttons will stack vertically.

Please see the visual spec for reference:

[Visual spec >](https://ibm.ent.box.com/file/651713722198)

[Storybook >](https://ibmdotcom-react.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/1699/index.html?path=/story/patterns-sub-patterns-buttongroup--default)

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

Button group

> What did you expect to happen? What happened instead? What would you like to
> see changed?

see above

> What browser are you working in?

Chrome Version 80.0.3987.149 (Official Build) (64-bit) on Mac.

> What version of the IBM.com Library are you using?

1.5.1

## Steps to reproduce the issue

1. Open [Button group Storybook](https://ibmdotcom-react.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/1699/index.html?path=/story/patterns-sub-patterns-buttongroup--default)
2. Use Inspect to button size and how they react responsively, whether ordered left to right or stacked vertically (see screenshot below)

## Additional information

![Screen Shot 2020-04-09 at 5.14.00 PM.png](https://images.zenhubusercontent.com/5e5e747dbf668377018d1b14/cefd5550-8652-4d1b-be57-6f8cb9dc9012)

![Screen Shot 2020-04-09 at 5.14.57 PM.png](https://images.zenhubusercontent.com/5e5e747dbf668377018d1b14/73dec13b-bc10-419e-b417-d33c6340fbf2)

![Screen Shot 2020-04-09 at 5.15.31 PM.png](https://images.zenhubusercontent.com/5e5e747dbf668377018d1b14/42e55acd-9664-4008-b0c4-dd20f68d663e)

[...]

PR #2490, opened 2020-05-22
fix(ButtonGroup): adding sameWidth utility to the pattern 

guilhermelMoraes:
### Related Ticket(s)

Button group | 320px fixed button width and responsive setup need adjustments #2016

### Description

These changes aim to keep the `ButtonGroup` pattern in pair with the [visual specs](https://ibm.ent.box.com/file/651716907513). 

### Changelog

**New**

- `_stackElementsVertically` function to programatically change the buttons group container between `row-reverse` or `column-reverse` according to the sum of all buttons inner sizing + margins
- `setSameWidth` function to set all the buttons to the same width using as reference the larger one;


asudoh:
Weighing in; While I need to learn much more about this library's requirements, I wonder if we can negotiate on the visual specs where we don't need JavaScript-based size calculations here. While I know that JavaScript-based size calculations is un-avoidable in some use cases, but it's hard to get it right and it has a [performance concern](https://gist.github.com/paulirish/5d52fb081b3570c81e3a).

guilhermelMoraes:
https://github.com/carbon-design-system/ibm-dotcom-library/pull/2490#issuecomment-632920169

Thanks, @asudoh for your concern. I reached Anna once ago in order to know if there's a better approach to solve this problem and she told me that the most likely way to follow with this issue, would be create a new utility, like the `sameHeight`. This, though, doesn't handle the component behavior in small breakpoints. 
So I agree with you, about negotiate the visual specs. 

fyi: @annawen1

asudoh:
Cool, thanks @guilhermelMoraes for your thoughts! Some other thoughts/choices I can think of are:

* Relegating "align width/height of all elements to the one of the biggest size" requirement to application code, instead of automating it. This is what Carbon does for tabs
* Using [`ResizeObserver`](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) to alleviate the performance concern

asudoh:
@guilhermelMoraes The new code replacing `resize` event with `ResizeObserver` looks nice! Wondering if we can put `getAllWidths()` calculation in (possibly another) `ResizeObserver`, too. Doing so may make such calculation run at the timing of browser is better comfortable with doing layout query.

packages/react/src/patterns/sub-patterns/ButtonGroup/ButtonGroup.js
```diff
@@ -16,26 +16,55 @@ const { prefix } = settings;
...
+      const elementsObserver = new ResizeObserver(entries => {
+        const getWidths = entries.map(entry => entry.contentRect.width);
+        const wilderButton = Math.max.apply(null, getWidths);
+        elements.forEach(
+          element => (element.style.width = `${wilderButton}px`)
+        );
+      });
+
+      elements.map(element => elementsObserver.observe(element));
+    };
```
jeffchew, line 41: Could this be a private function instead of inside of the `useLayoutEffect`?

jeffchew:
CI checks are failing because `ResizeObserver` is not defined. I'm wondering if we need a polyfill for this?

cc: @asudoh

asudoh:
@jeffchew Seems so - Many people seem to suggest that tests use a mock version of `ResizeObserver`, where callbacks are called immediately. Some others suggest making such mock no-op, but it may not be the case for us.

asudoh:
Great updates so far @guilhermelMoraes! It works well with initial content. We seem to have a problem with after-initiallization content update though (e.g. via knobs). One cause seems to be because we are setting explicit widths to the exact same elements where we inspect the size to see how big the inner-contents are.

Adding a separate inline element (e.g. `<span>`) that enlarges or shrinks depending on inner-content, and using it for size inspection seems to solve the problem. However, it seems still constrainted by `max-width` of the `<button>` which seems `320px` (we can see it by setting extremely long words via knobs). Attempts to add `word-break` to that `<span>` to alleviate that problem seems to deprive us from the chance of making the button larger once it's shrunk (saw too narrow buttons where it should be wider).

So somehow I got inclined to UX negotiation (again)... Curious to see what people think.

BTW another observation is that the resize observer has to be cleaned up upon DOM change and React unmount, to avoid zombie `ResizeObserver`s. To keep the record, here's what I got my hands dirty with to better understand the code and this problem space:

```diff
diff --git a/packages/react/src/patterns/sub-patterns/ButtonGroup/ButtonGroup.js b/packages/react/src/patterns/sub-patterns/ButtonGroup/ButtonGroup.js
index 818d8f8b..4d5000f4 100644
--- a/packages/react/src/patterns/sub-patterns/ButtonGroup/ButtonGroup.js
+++ b/packages/react/src/patterns/sub-patterns/ButtonGroup/ButtonGroup.js
@@ -4,7 +4,7 @@
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
  */
-import React, { useRef, useLayoutEffect } from 'react';
+import React, { useRef, useEffect } from 'react';
 import Button from '../../../internal/vendor/carbon-components-react/components/Button/Button';
 import { settings as ddsSettings } from '@carbon/ibmdotcom-utilities';
 import PropTypes from 'prop-types';
@@ -19,29 +19,41 @@ const { prefix } = settings;
 const ButtonGroup = ({ buttons }) => {
   const orderedList = useRef(null);
 
-  useLayoutEffect(() => {
-    const { current } = orderedList;
+  const { current: elementsObserver } = useRef(new ResizeObserver(entries => {
+    const { current: orderedListNode } = orderedList;
 
-    /**
-     * Utility to give all the elements the same width, using the wilder one as reference for the others
-     *
-     * @param {Node} parentNode - the element
-     */
-    const sameWidth = parentNode => {
-      const elements = Array.from(parentNode.childNodes);
-      const elementsObserver = new ResizeObserver(entries => {
-        const getWidths = entries.map(entry => entry.contentRect.width);
-        const wilderButton = Math.max.apply(null, getWidths);
-        elements.forEach(
-          element => (element.style.width = `${wilderButton}px`)
-        );
-      });
+    const widthsFromResizeObserver = new WeakMap();
+    entries.forEach(entry => {
+      widthsFromResizeObserver.set(entry.target, entry.contentRect.width);
+    });
+
+    const textContentNodes = orderedListNode.querySelectorAll('[data-buttongroup-text-content]');
+    const widths = Array.prototype.map.call(textContentNodes, node => {
+      const widthFromResizeObserver = widthsFromResizeObserver.get(node);
+      return widthFromResizeObserver ?? node.offsetWidth;
+    });
+    const widestButtonWidth = Math.max(...widths);
+
+    Array.prototype.forEach.call(orderedListNode.childNodes, node => {
+      node.style.width = `${widestButtonWidth + 68}px`;
+    });
+  }));
+
+  useEffect(() => {
+    const { current: orderedListNode } = orderedList;
+    const textContentNodes = orderedListNode.querySelectorAll('[data-buttongroup-text-content]');
+    Array.prototype.forEach.call(textContentNodes, node => { elementsObserver.observe(node); });
+    return () => {
+      Array.prototype.forEach.call(textContentNodes, node => { elementsObserver.unobserve(node); });
+    };
+  }, [buttons, elementsObserver]);
 
-      elements.map(element => elementsObserver.observe(element));
+  useEffect(() => {
+    return () => {
+      elementsObserver.disconnect();
     };
+  });
 
-    sameWidth(current);
-  }, []);
 
   return (
     <ol
@@ -57,7 +69,7 @@ const ButtonGroup = ({ buttons }) => {
               {...button}
               type="button"
               kind={key === 0 ? 'primary' : 'tertiary'}>
-              {button.copy}
+              <span data-buttongroup-text-content>{button.copy}</span>
             </Button>
           </li>
         );
```

jeffchew:
@guilhermelMoraes checking on this, looks like you need to stub out `ResizeObserver` in jest. You can set that up under `/tasks/jest/setup.js` and put a mock in there.

jeffchew:
@guilhermelMoraes @asudoh @wonilsuhibm rounding back on the discussion of this as this PR is getting a bit old. We're moving forward with the JS approach with `ResizeObserver`, with the ability for adopters to turn off this functionality, Correct?

asudoh:
Makes sense - Such switching logic can be something like [here](https://github.com/carbon-design-system/ibm-dotcom-library/blob/729f8ac8a8dce45ce248bab4669d7dd34398a2b4/packages/react/src/components/ButtonGroup/ButtonGroup.js#L22-L23). The link contains [code for granular lifecycle management](https://github.com/carbon-design-system/ibm-dotcom-library/blob/729f8ac8a8dce45ce248bab4669d7dd34398a2b4/packages/react/src/components/ButtonGroup/ButtonGroup.js#L51-L87) (to cope with props changes and un-mounting as well, in case it helps.

packages/react/src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -16,26 +16,115 @@ const { prefix } = settings;
...
-            {button.copy}
-          </Button>
-        </li>
+const ButtonGroup = ({ buttons, enableSizeByContent }) => {
+  const groupRef = useRef(null);
+  const observedPseudoButtonNodesRef = useRef(new Set());
+  const shouldUseResizeObserver =
+    enableSizeByContent && typeof ResizeObserver !== 'undefined';
+  const resizeObserverButtonsRef = useRef(
+    new ResizeObserver(entries => {
```
asudoh, line 25:
```suggestion
    !shouldUseResizeObserver  ? null : new ResizeObserver(entries => {
```

My apologies that I may have missed this in my prototype. This may solve CI error.

guilhermelMoraes:
@asudoh thanks for your input, Akira! In my local enviroment things are working as expected now, with no errors about the `ResizeObserver`, at least.

PR #2495, opened 2020-05-25
docs(components): migrate to React DocGen comments

asudoh:
### Related Ticket(s)

Refs #2422.

### Description

This change migrates JSDoc comments to React DocGen comments for components in `packages/react/src/components`.

Also:

* Tried to make some prop type definitions more detailed, e.g. usage of `PropTypes.oneOf()` instead of `PropTypes.string` and usage of `PropTypes.shape()` instead of `PropTypes.object()` so that the components can catch more of prop type errors.
* Fixed some stale prop type settings.

### Changelog

**Changed**

- Migrated JSDoc to React DocGen comments for components in `packages/react/src/components`.
- Tried to make some prop type definitions more detailed, e.g. usage of `PropTypes.oneOf()` instead of `PropTypes.string` and usage of `PropTypes.shape()` instead of `PropTypes.object()` so that the components can catch more of prop type errors.
- Fixed some stale prop type settings.


jeffchew:
@asudoh this looks great!

I'm wondering if we can remove the README tab entirely, doesn't seem like we'd need it anymore. I can probably approve and merge this PR and we can address in a separate PR.

asudoh:
Thanks @jeffchew for reviewing!

> I'm wondering if we can remove the README tab entirely, doesn't seem like we'd need it anymore.

Agreed, let's do it once the transition completes. So queued up transition items are:

* [ ] Remove Storybook README tab
* [ ] Bump up missing React DocGen comments ESLint rule from warning to error

PR #2498, opened 2020-05-26
docs(sub-patterns): migrate to React DocGen comments

asudoh:
### Related Ticket(s)

Refs #2422.

### Description

This change migrates JSDoc comments to React DocGen comments for components in `packages/react/src/patterns/sub-patterns`.

Also:

* Tried to make some prop type definitions more detailed, e.g. usage of `PropTypes.oneOf()` instead of `PropTypes.string` and usage of `PropTypes.shape()` instead of `PropTypes.object()` so that the components can catch more of prop type errors.
* Fixed some stale prop type settings.

### Changelog

**Changed**

- Migrated JSDoc to React DocGen comments for components in `packages/react/src/patterns/sub-patterns`.
- Tried to make some prop type definitions more detailed, e.g. usage of `PropTypes.oneOf()` instead of `PropTypes.string` and usage of `PropTypes.shape()` instead of `PropTypes.object()` so that the components can catch more of prop type errors.
- Fixed some stale prop type settings.


asudoh:
> One thing I think this work is surfacing is the lack of `defaultProps` in a bunch of our patterns/components. We'll probably need to go back and doublecheck the specs and set those for the ones that are missing.

Great point @jeffchew - Any ideas on how we can best move forward with this...? Probably gathering the specs, reviewing them, and compare with the status quo of the code...?

jeffchew:
> > One thing I think this work is surfacing is the lack of `defaultProps` in a bunch of our patterns/components. We'll probably need to go back and doublecheck the specs and set those for the ones that are missing.
> 
> Great point @jeffchew - Any ideas on how we can best move forward with this...? Probably gathering the specs, reviewing them, and compare with the status quo of the code...?

Most likely. The patterns especially should have corresponding functional specs we can refer to. For components though, we might need to make those determinations from the engineering side.

PR #2561, opened 2020-05-28
fix(pattern): buttongroup inverse theme

greeshmareddyg:
### Related Ticket(s)

https://github.com/carbon-design-system/ibm-dotcom-library/issues/2291

### Changelog

**Changed**

Enabled the Inverse theme for button group


packages/react/src/patterns/sub-patterns/ButtonGroup/README.stories.mdx
```diff
@@ -115,6 +117,11 @@ ReactDOM.render(<App />, document.querySelector('#app'));
 
 <Props of={ButtonGroup} />
 
+| Name      | Required | Data Type | Default Value | Description                                                 |
+| --------- | -------- | --------- | ------------- | ----------------------------------------------------------- |
+| `buttons` | NO       | Array     | null          | Array of button objects to render. See `Button Item` below. |
+| `inverse` | NO       | Boolean   | false         | Changes theme to inverse/default               |
+
```
kennylam, line 124: @greeshmareddyg I don't think this is needed, right? Since we're already pulling in the props from the component in the line just above?

greeshmareddyg: @kennylam yeah, removed the props description here.

jeffchew:
@greeshmareddyg inverse doesn't really look like inverse:

button-inverse

greeshmareddyg:
> @greeshmareddyg inverse doesn't really look like inverse:
> 
> button-inverse



> @greeshmareddyg inverse doesn't really look like inverse:
> 
> button-inverse

@jeffchew I have used the $inverse-02, can you suggest the right one?

PR #2562, opened 2020-05-28
fix(ContentGroupHorizontal): adding horizontal rule at the end

greeshmareddyg:
### Related Ticket(s)

https://github.com/carbon-design-system/ibm-dotcom-library/issues/2370

### Changelog

**Changed**

Added a horizontal rule.


packages/react/src/patterns/blocks/ContentGroupHorizontal/ContentGroupHorizontal.js
```diff
@@ -20,20 +21,22 @@ const { prefix } = settings;
...
+          <ContentItemHorizontal
+            eyebrow={item.eyebrow}
+            heading={item.heading}
+            copy={item.copy}
+            cta={item.cta}
+            key={index}
+          />
+        ))}
+      </ContentBlock>
+      <HorizontalRule />
```
photodow, line 38: I don't think we want to hard code the horizontal rule into this pattern. It still needs to be able to nest inside smaller Layouts without the line and optional at the least. I think adding it here would cause the wrong grid alignment. Content block also has this as an option I believe?

jeffchew:
@greeshmareddyg checking on this PR as it's getting a bit stale. I agree with @photodow that we shouldn't hardcode the HR, and I believe it is an option for aside items in the ContentBlock and not the ContentBlock itself. You might want to create a new prop to include or exclude the border/horizontalrule for this particular component, but could get some other opinions on this?

Adding @annawen1 here since she worked on the border with aside items in the ContentBlock specifically.

packages/react/src/patterns/blocks/ContentGroupHorizontal/ContentGroupHorizontal.js
annawen1: yeah we can pass in a the `aside` object prop to `ContentBlock`

jeffchew: Ok, but I guess I'm wondering if this is the correct approach or not?

photodow:
@jeffchew any particular reason we're trying to treat this pattern different from the others? The optional divider is something for every pattern. In the past, I think the adopter used the nested Layout option to add the rule, or recently now with the enablement of aside. Does this pattern itself need that optional rule when the adopter has these other two ways to make this happen?

jeffchew:
> @jeffchew any particular reason we're trying to treat this pattern different from the others? The optional divider is something for every pattern. In the past, I think the adopter used the nested Layout option to add the rule, or recently now with the enablement of aside. Does this pattern itself need that optional rule when the adopter has these other two ways to make this happen?

If we're just turning the border on within this component and keeping the noise out of the decorator pattern, then I'm fine with using the aside approach, but if we are leaving it up to adopters to turn on the HR via aside, it seems kind of a hokey way of doing it:

```
<ContentGroupHorizontal border={true} />
```

vs

```
<ContentGroupHorizontal aside={{ items: <div></div>, border: true }} />
```

But like I said, if we are not giving it as an option at all then I'm fine with the aside approach and just leveraging what we have.

packages/styles/scss/patterns/blocks/content-group-horizontal/_content-group-horizontal.scss
```diff
@@ -7,3 +7,15 @@
 
 @import '../../../internal/content-block/content-block';
 @import '../../../components/content-item-horizontal/content-item-horizontal';
+
+@mixin content-group-horizontal {
+  .#{$prefix}--content-group-horizontal {
+    &--border {
+      border-bottom: solid 1px $ui-04;
```
jeffchew, line 14: @greeshmareddyg I don't think this is what we were talking about for this. We were thinking of setting the border to true inside of `ContentBlock`:

```
<ContentBlock 
  heading={heading} 
  aside={{
    items: <div></div>
    border: true
  }}
>
    {items.map((item, index) => (
        <ContentItemHorizontal ... />
      ))}
</ContentBlock>
```

greeshmareddyg: @jeffchew addressed this, still gave it as optional with default border true, is it ok.

packages/react/src/patterns/blocks/ContentGroupHorizontal/ContentGroupHorizontal.js
```diff
@@ -20,20 +19,26 @@ const { prefix } = settings;
  */
 const ContentGroupHorizontal = ({ heading, items }) => {
   return (
-    <ContentBlock
-      heading={heading}
-      autoid={`${stablePrefix}--content-group-horizontal`}
+    <div
+      data-autoid={`${stablePrefix}--content-group-horizontal`}
       className={`${prefix}--content-group-horizontal`}>
```
jeffchew, line 24: @greeshmareddyg just checking, is there a reason why a wrapper div was added around the `ContentBlock`?

cc: @kennylam

greeshmareddyg: @jeffchew No, earlier we have applied the classname if to the content block, as we are using within the contenthorizontalgroup I just have wrapped and applied styles. shall I apply to contentblock as before?

kennylam: @jeffchew This is actually the correct implementation. The Drupal team requested this change a while back because targeting the auto-id within `ContentBlock` itself was causing issues for them and suggested we lift it a level.

jeffchew: Got it, thanks @greeshmareddyg @kennylam

Issue #2289, opened 2020-05-04
Pattern: Feature card block - medium: Enable inverse theming 


RobertaJHahn:
#### User Story

> As a `[user role below]`:
IBM.com Library developer

> I need to:
create the `Feature card block - medium`

> so that I can:
provide the ibm.com adopter developers patterns/components they can use to build ibm.com web pages

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- This pattern should have inverse theming enabled.  Background information is included in issue (#2212)
- Prod QA testing issue (#2432 )

#### Acceptance criteria
- [ ] Include README for the react pattern and corresponding styles
- [ ] Minimum 80% unit test coverage
- [ ] A comment is posted in the Design QA issue, tagging Wonil and Roberta, when development is finished
 - [ ] A comment is posted in the Prod QA issue, tagging Praveen and Chetan, when development is finished

oliviaflory:
I reviewed this issue with @photodow and @RaphaelAmadeu today, the inverse of this pattern is not needed immediately by Reboot, so we agreed to hold off on enabling inverse theming until we can look at how we are handling theming and inverse tokens across all of our patterns. 

I will create a story to look into this to review with designers and developers and add it as a dependency to this issue. 

cc: @RobertaJHahn @jeffchew

PR #2575, opened 2020-05-28
feat(pattern): added inverse theming for feature-card

RaphaelAmadeu:
### Related Ticket(s)

#2289 

### Description

Added inverse theming as a prop for `FeatureCard`.


photodow:
I think this is supposed to add inverse theming to feature card block medium, not feature card. I reached out via slack, but let me know if we need to sync up.

PR #2579, opened 2020-05-28
fix(featurecard): fixed feature card style type

greeshmareddyg:
### Related Ticket(s)

https://github.com/carbon-design-system/ibm-dotcom-library/issues/2547

### Changelog

**Changed**

- feature card heading type

### DIFF
--- a/packages/styles/scss/components/feature-card/_feature-card.scss
+++ b/packages/styles/scss/components/feature-card/_feature-card.scss
@@ -51,7 +51,6 @@
 
       &__heading {
         margin-bottom: $layout-03;
-        @include carbon--type-style(expressive-heading-03, true);
       }
 
       &__wrapper,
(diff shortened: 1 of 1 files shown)


packages/styles/scss/components/feature-card/_feature-card.scss
```diff
@@ -51,7 +51,7 @@
 
       &__heading {
         margin-bottom: $layout-03;
-        @include carbon--type-style(expressive-heading-02, true);
+        @include carbon--type-style(expressive-heading-03, true);
```
photodow, line 54: Could we just delete this line since `Card` already uses `heading-03`?

https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/styles/scss/components/card/index.scss#L31

PR #2581, opened 2020-05-28
chore(styles): update carbon imports

### LINKED ISSUE #2437: Potential duplicate imports of Carbon styles
## Detailed description

Found that `ibm-dotcom-library` has some Sass imports from `@carbon/type`,`@carbon/grid`, etc. libraries, in addition to `carbon-components`.

`carbon-components` [imports those libraries from its own copy](https://github.com/carbon-design-system/carbon/blob/v10.12.0/packages/components/src/globals/scss/_css--font-face.scss#L10-L11). It means that `import-once` check is not applied between ones in `@carbon/something` and ones in `carbon-components`, which increases the Sass build size.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

A general issue?

> see changed?

All `@carbon/something` styles are imported from `carbon-components` library.

Latest `master`.

kennylam:
### Related Ticket(s)

Fixes #2437 

### Description

Update `scss` to import consistently from the same location. This allows `import-once` to work as it should and eliminate css duplication.

### Changelog

**Changed**

- imports from `@carbon` change to `carbon-components/scss/globals/scss/vendor/@carbon`

### DIFF
--- a/packages/styles/scss/components/card/index.scss
+++ b/packages/styles/scss/components/card/index.scss
@@ -7,7 +7,7 @@
 
 @import '../../globals/imports';
 @import '../../temp-carbon-expressive/temp-carbon-expressive';
-@import '@carbon/motion/scss/motion.scss';
+@import 'carbon-components/scss/globals/scss/vendor/@carbon/elements/scss/motion/motion.scss';
 
 @import 'carbon-components/scss/components/link/link';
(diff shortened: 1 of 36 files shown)


asudoh:
Thanks @kennylam for doing this! 🎉

Issue #2554, opened 2020-05-27
Change the Content group component based on the recent Reboot Learn template designs


RobertaJHahn:
#### User Story

> As a `[user role below]`:
IBM.com Library developer

> I need to:
change the `Content group` component

> so that I can:
provide the ibm.com adopter developers patterns/components they can use to build ibm.com web pages

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- Once sized, the development work should be finished by the end of sprint 20-11 (June 5).  DDS has committed to Reboot that this pattern will be included in the June release.
- **See the Epic for the Design and Functional specs information**
- Design QA testing issue (#2556 )
- Prod QA testing issue (#2558 )

#### Acceptance criteria
- [x] Include README for the react pattern and corresponding styles
- [x] Minimum 80% unit test coverage
- [x] A comment is posted in the Design QA issue, tagging Lila Title, when development is finished
- [x] A comment is posted in the Prod QA issue, tagging Praveen and Chetan, when development is finished
- [ ] The Canary URL is added to the Prod QA issue for their testing
- [x] The Storybook link is added to the Design QA issue for their testing

PR #2620, opened 2020-06-01
feat(patterns): adding optional copy for ContentGroup patterns

guilhermelMoraes:
### Related Ticket(s)

Change the Content group component based on the recent Reboot Learn template designs #2554

### Description

Adding the optional `copy` prop to the `ContentGroup` and it's respective patterns making use of it. The `copy` prop has support to the Markdown utility. 

### Changelog

**Changed**

- Now the `ContentGroupSimple`, `ContentGroupWithCards`, `ContentGroupWithPictograms` support the optional `copy` prop.


packages/react/src/patterns/blocks/ContentGroupPictograms/__stories__/ContentGroupPictograms.stories.js
```diff
@@ -100,6 +100,11 @@ export default {
             'Lorem ipsum dolor sit amet, consectetur adipiscing elit.',
             groupId
           ),
+          copy: text(
+            'Copy (optional)',
```
jeffchew, line 104: @guilhermelMoraes since you're updating the story, can you update the knob labels for this story to include the actual prop name in parenthesis?

packages/react/src/patterns/blocks/ContentGroupSimple/__stories__/ContentGroupSimple.stories.js
```diff
@@ -28,6 +28,7 @@ export default {
               ? ContentGroupSimpleKnobs.mediaData.image
               : ContentGroupSimpleKnobs.mediaData.video,
           heading: text('Heading', ContentGroupSimpleKnobs.heading, groupId),
+          copy: text('Copy (optional):', ContentGroupSimpleKnobs.copy, groupId),
```
jeffchew, line 31: @guilhermelMoraes can you check the knob labels in this one as well to add the actual prop names in parenthesis?

packages/react/src/patterns/blocks/ContentGroupPictograms/__stories__/ContentGroupPictograms.stories.js
guilhermelMoraes: I'm sorry, I didn't get what you mean. Perhaps set it to: `Copy (props.copy)(optional)`?

jeffchew: Right, well something like: https://ibmdotcom-react-canary.mybluemix.net/?path=/story/components-masthead--default

If you can add for the other props in this story too would be great!

jeffchew:
@guilhermelMoraes @wonilsuhibm @kathysaraf is this the right amount of spacing under the copy? Who is the designer who worked on this?

https://percy.io/IBM-Digital-Design/IBM.com-Library-React/builds/5493056/view/321635406/1280?mode=diff&browser=chrome&snapshot=321635406

wonilsuhibm:
The spacing is definitely not right. There is this ticket that shows you all the right spacings for the different expressions of this pattern. https://github.com/carbon-design-system/ibm-dotcom-library/issues/2624

Issue #2514, opened 2020-05-27
Create workflow for publishing beta releases of IBM.com Library


jeffchew:
#### User Story

> As a `[user role below]`:
IBM.com Library developer

> I need to:
have an easy way to publish beta releases

> so that I can:
quickly deliver a beta release of the IBM.com Library at the end of every sprint

#### Additional information

 - To avoid disrupting any existing prerelease flows, we should consider these as a canary release off of the `master` branch with the `beta` preid/dist-tag
 - This may be a manual process, and may want to consider not publishing as an official release in github

#### Acceptance criteria
 - [ ] CI job for beta release creation
 - [ ] Beta release playlist
 - [ ] Automated 2 week change log

PR #2621, opened 2020-06-01
fix(beta-releases): minor fix to remove the delimiter

jeffchew:
### Related Ticket(s)

#2514 

### Description

This is a minor fix to remove the delimiter from appearing in the last changelog item in the script.

### Changelog

**Changed**

- Beta release script tweak

### DIFF
--- a/tasks/beta-releases.js
+++ b/tasks/beta-releases.js
@@ -64,6 +64,13 @@ const prevTag = args.prevtag;
 const newTag = args.newtag;
 
 /**
+ * Uses a delimiter for splitting the comments into an array
+ *
+ * @type {string}
+ */
+const delimeter = '----DELIMITER----';
+
+/**
  * Returns back the commits in an array
  *
  * @returns {string[]} Commits array of objects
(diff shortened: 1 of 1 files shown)


tasks/beta-releases.js
```diff
@@ -96,6 +103,7 @@ function getChangelog() {
   const commitsArray = getCommits();
 
   commitsArray.forEach(commit => {
+    commit = commit.replace(delimeter, '');
```
asudoh, line 106: Wondering if we still need this line, given we split the original string by the delimiter. I may be missing something though, and having this like for the safety sake sounds good tome.

jeffchew: Hi @asudoh ,

I added it because when the changelog was generated, the last item for some reason still had the delimiter text in there. I ended up manually removing it from the test beta release (https://github.com/carbon-design-system/ibm-dotcom-library/releases/tag/1.7.1-beta.4303)

asudoh: I see, makes sense, thank you for clarifying @jeffchew!

PR #2622, opened 2020-06-01
fix(pattern): icon size in FeatureCardBlockLarge corrected

reesedownes:
### Related Ticket(s)

https://app.zenhub.com/workspaces/ibmcom-library-5d449f3642eb1962336cbe52/issues/carbon-design-system/ibm-dotcom-library/2400

### Description

CTA Icon font size corrected

### Changelog

**Changed**

Font size of CTA Icon changed from 20px to 32px

### DIFF
--- a/packages/styles/scss/patterns/blocks/feature-card-block-large/_feature-card-block-large.scss
+++ b/packages/styles/scss/patterns/blocks/feature-card-block-large/_feature-card-block-large.scss
@@ -34,6 +34,9 @@ $fcb-large-custom-bp-nocopy: 752px;
     &__card.#{$prefix}--card--inverse .#{$prefix}--card__cta:visited {
       float: none;
       color: $inverse-link;
+      width: $layout-03;
+      height: $layout-03;
+
       @include carbon--type-style('productive-heading-05');
     }
 


packages/styles/scss/patterns/blocks/feature-card-block-large/_feature-card-block-large.scss
```diff
@@ -32,6 +32,9 @@ $fcb-large-custom-bp-nocopy: 752px;
     &__card.#{$prefix}--card--inverse .#{$prefix}--card__cta:visited {
       float: none;
       color: $inverse-link;
+      width: $layout-03;
```
jeffchew, line 35: Looks good, though we might want to change this in the future if we're introducing icon tokens in Carbon? cc: @asudoh

asudoh: Agreed - Once Carbon gets a fix where token value customization works, we can start using it.

reesedownes:
@jeffchew two questions i had:

1. i assume i titled this incorrectly, and i know a thread went around about this, so i just wanted to clarify. is the format now just `typeofcommit(component): subject`? don't mention the issue number?

2. is there any intrinsic difference between saying, for example, $layout-03 or $spacing-07? i know that one list of tokens goes to a higher scale, but functionally are they different at all?

Issue #2516, opened 2020-05-27
[ToC] Children and optional rule should scroll away, not sticky


shixiedesign:
Only main Table of Content stays on page after scrolling. Everything above the main Table of Content, aka Children and top rule should scroll away. 

![May-27-2020 11-46-23](https://user-images.githubusercontent.com/15144993/83042518-bc295380-a00f-11ea-844e-6e00b6eb897a.gif)

PR #2631, opened 2020-06-02
fix(pattern): table-of-contents aside child now gets scrolled

RaphaelAmadeu:
### Related Ticket(s)

#2516 

### Description

I've changed the `TableOfContents` component, so now the aside children gets scrolled by and only the navigation remains sticky.


packages/react/src/components/TableOfContents/TableOfContents.js
```diff
@@ -196,17 +196,25 @@ const TableOfContents = ({
...
+            </div>
+          ) : null}
           <div className={`${prefix}--tableofcontents__mobile-top`}></div>
-          <TOCDesktop
-            menuRule={menuRule}
-            headingContent={headingContent}
-            {...props}
-          />
-          <TOCMobile {...props} />
+          <div style={{ position: 'sticky', top: '0' }} data-sticky="true">
```
jeffchew, line 210: @RaphaelAmadeu haven't looked at `Layout` in awhile, but since this isn't a direct child of `Layout` the `data-sticky` attribute is no longer needed here I think?

Issue #2610, opened 2020-06-01
Component: LinkList - Accessibility -  links are not Keyboard operable with JAWS. Hence unable to activate these links.


Praveen-sr:
## Detailed description

Issue :- Learn more, Containerization A Complete Guide, Micro-services and containers, Folgers Coffee (1:00) - links are not Keyboard operable with JAWS. Hence unable to activate these links.

Expected :- User should be able to activate these links with jaws running

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/8cc6ffc3-2b87-41c5-a780-e6b86530b45c)

guilhermelMoraes:
@Praveen-sr @chetankan this may not be a problem with jaws. I was able to navigate through the `LinksList` in Chrome, but not in Firefox until I followed the steps here: https://stackoverflow.com/questions/11704828/how-to-allow-keyboard-focus-of-links-in-firefox

Also, using the [chromevox extension](https://chrome.google.com/webstore/detail/chromevox-classic-extensi/kgejglhpjiefppelpmljglcjbhoiplfn?hl=en) it worked as expected, reading all the focused items in the list.

Can you guys provide a little bit more of information about this bug, please?

Praveen-sr:
@guilhermelMoraes Comments from Ram :
The mentioned defect seems to appear after the recent update happened in jaws 2020. i.e. ( Version 2020.2004.66). To Cross-check, When I tested the same links in the older version of jaws i.e. 2019, I am able to activate the links. 
Though the link role is not still identified even in the older version which is a related defect.

Praveen-sr:
@guilhermelMoraes Re tested this on the fix ([ URL link](https://ibmdotcom-react-experimental.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/2634/index.html?path=/story/components-linklist--default)) you provided but the issue still exists.

Here is the code from URL you provided us to test:-
![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/3274e2d9-d85e-4aa2-bbf6-58b1ee01b8d6)

Here is the fix, we are expecting :-
![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/de4abcdc-720b-4ce9-a3f2-83d067918ad5)

CC: @RobertaJHahn

guilhermelMoraes:
[...]

PR #2634, opened 2020-06-02
fix(acessibility):  jaws cannot read `LinksList`

guilhermelMoraes:
### Related Ticket(s)

Component: LinkList - Accessibility - links are not Keyboard operable with JAWS. Hence unable to activate these links. #2610

### Description

Based on our meeting today, earlier, I set the `role=region` in  `Card` component and the `aria-label` on it's parent div in order to jaws recognize the link in the `LinkList`

This one needs the QA team attention  since we don't have access to the jaws.

### Changelog

**New**

- `role=region` in the `Card` component;
- `aria-label` in the Card parent div in the `LinksList`.


guilhermelMoraes:
@Praveen-sr @chetankan hi! Just tagging you guys, so you can validate if the changes were enough to jaws recognize the `LinksList` during the navigation

Issue #2483, opened 2020-05-22
Pattern: Enhance the Card pattern to support video


RobertaJHahn:
#### User Story

> As a `[user role below]`:
IBM.com Library developer

> I need to:
change the `Card` pattern

> so that I can:
provide the ibm.com adopter developers patterns/components they can use to build ibm.com web pages

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- **See the Epic for the Design and Functional specs information**
- Design QA testing issue (#2485)
- Prod QA testing issue (#2484)

#### Acceptance criteria
- [x] Include README for the react pattern and corresponding styles
- [x] Minimum 80% unit test coverage
- [x] A comment is posted in the Design QA issue, tagging Wonil and Roberta, when development is finished
- [x] A comment is posted in the Prod QA issue, tagging Praveen and Chetan, when development is finished
- [x] The Canary URL is added to the Design and Prod QA issues for their testing

PR #2652, opened 2020-06-03
feat(component): enhance card to support video

annawen1:
### Related Ticket(s)

Pattern: Enhance the Card pattern to support video #2483

### Description

Allow for Card to support video styling

Thumbnail is Kaltura generated, but adopters can override by passing their own image in

Screen Shot 2020-06-03 at 6 22 31 PM

### Changelog

**New**

- add play icon in `styles` package
- `getThumbnailUrl` method in `VideoPlayerAPI` service to form kaltura thumbnail api url

**Changed**

- have `Image` component handle center of icon overlaying image
- separate duration and video title for the CTA component in state


packages/react/src/components/CTA/ButtonCTA.js
```diff
@@ -67,7 +67,9 @@ const _renderButtons = ({
       let title = videoTitle.filter(name => {
         return name.key === key;
       });
-      button.copy = !title[0] ? button.copy : title[0].title;
+      button.copy = !title[0]
+        ? button.copy
+        : `${title[0].title} ${title[0].duration}`;
```
asudoh, line 72: g11n; I'd add another prop for this kind of composite string, given it's hard to assume order of contents when different languages are in consideration:


```suggestion
        : formatButtonCopyWithTitle({ title: title[0].title, duration: title[0].duration });
```

```javascript
ButtonCTA.defaultProps = {
  formatButtonCopyWithTitle: ({ title, duration }) => `${title[0].title} ${title[0].duration}`;
};
```

Another approach may be letting `useVideoData()` set `title` with duration (meaning, relegating formatting logic to the caller (parent) of `<ButtonCTA>`, etc.), though I'm not sure if it's safe to do so (do we have some code that requires distinction between title and duration)?

packages/react/src/components/CTA/FeatureCTA.js
```diff
@@ -33,7 +32,7 @@ const FeatureCTA = ({
           card={_renderFeatureCard({
             card: {
               ...otherProps.card,
-              heading: videoTitle[0].title,
+              heading: `${videoTitle[0].title} ${videoTitle[0].duration}`,
```
asudoh, line 35: Please see my comments to `<ButtonCTA>` change.

packages/react/src/components/CTA/TextCTA.js
```diff
@@ -33,7 +33,7 @@ const TextCTA = ({
         <LinkWithIcon
           href="#"
           onClick={e => CTALogic.setLightBox(e, openLightBox)}>
-          <span>{videoTitle[0].title}</span>
+          <span>{`${videoTitle[0].title} ${videoTitle[0].duration}`}</span>
```
asudoh, line 36: Please see my comments to `<ButtonCTA>` change.

packages/react/src/internal/hooks/useVideoData/useVideoData.js
```diff
@@ -16,8 +16,13 @@ import { VideoPlayerAPI } from '@carbon/ibmdotcom-services';
...
  */
 function useVideoData(type, videoId) {
-  const [videoTitle, setVideoTitle] = useState([{ title: '', key: 0 }]);
+  const [videoTitle, setVideoTitle] = useState([
+    { title: '', duration: '', key: 0 },
+  ]);
 
+  useEffect(() => {
+    getVideoData();
+  }, [getVideoData, type]);
```
annawen1, line 25: Hey @asudoh I noticed the order of the `useEffect` placement was causing issues whenever I changed the CTA type to `video`. There would be an infinite re-rendering of the CTA component. I saw that you made the change to move the `useEffect` to fix transpilation. I wonder if there is another way to get around the issue?

packages/react/src/components/CTA/ButtonCTA.js
annawen1: Originally I had it set that this logic of concatenating the `title` and `duration` in `useVideoData()` but for CardCTA specifically we need to separate the two based on the designs.

You brought up a good point, I'll make changes to taken in another prop

packages/react/src/internal/hooks/useVideoData/useVideoData.js
asudoh: @annawen1 My apologies for the breakage. Yes, ensuring `getVideoData` variable declared at the top fixes the problem and thank you for doing this!

Issue #2464, opened 2020-05-20
Investigate Percy integration with integration tests


asudoh:
### The problem

As we run integration tests, it's nice to have visual regression tests in there, too.

### The solution

Integrate Percy to integration tests (#2461).

**Acceptance criteria**

- [ ] Build integration test that was introduced in #2462 sends some screenshots to Percy

jeffchew:
Documentation on running percy snapshots in Puppeteer: https://docs.percy.io/docs/puppeteer

RobertaJHahn:
Work is done.  Closing

PR #2658, opened 2020-06-04
test(integration): add more tests

asudoh:
### Related Ticket(s)

Refs #2464.

### Description

This change introduces build integration tests to recently added CodeSandbox examples.

This change also adds Percy integration to build integration tests, but disabled for now due to long build time in CodeSandbox example that negatively affects our Jenkins environment.

### Changelog

**New**

- Build integration tests. to recently added CodeSandbox examples.
- Percy integration to build integration tests. (Disabled for now due to long build time in CodeSandbox example that negatively affects our Jenkins environment)
- Some code in CodeSandbox examples for better error tracking.
- `@babel/core` as dependencies of CodeSandbox so that Parcel won't re-install `node_modules` dependencies when Parcel server starts up.
- `NAVIGATION_TIMEOUT`, separated from `LAUNCH_TIMEOUT` in integration tests so that we can separate the timeout for server setup/launch vs. one for opening a web page.
- Code to bring in the latest `@carbon/ibmdotcom-services` code and `@carbon/ibmdotcom-utilities` for build integration testing our React code.

**Changed**

- Vanilla CodeSandbox example to use `.sassrc`. (To enable `node_modules` imports). To do that, added entry point Sass files (Parcel won't read `.sassrc` if a JavaScript directly imports Sass files in `node_modules`).

**Removed**

- Optional chaining usage from `process.env.ENVVAR` code because Parcel cannot replace them with actual environment variables (presumably a limitation in its AST code).


asudoh:
Finally ready for review - Had to move from Jenkins to GH actions due to long build time. Given that, had to disable vanilla tests due to a required secret (proxy), and disable Percy integration tests also due to a required secret (its token).

Will revisit once we upgrade all CodeSandbox example to latest Carbon.

Issue #2612, opened 2020-06-01
Pattern / Component - Leadspace Block & Link List - keyboard focus is not covering all four sides of the element.


Praveen-sr:
## Detailed description

Issue :- Visual focus is there but its not covering all four sides of the element.

## Additional information
Leadspace block:
![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/dffe441b-0b72-45f1-8821-24cf5151d96c)

Link list :
![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/c3728690-880f-4fb5-bc64-f4ed38810947)

jeffchew:
@guilhermelMoraes can you take this one as well since you're already looking at `LinkList`?

guilhermelMoraes:
@jeffchew yes, sure

Praveen-sr:
@RobertaJHahn Re checked this issues, keyboard focus on elements is working as expected.

PR #2660, opened 2020-06-04
style(component): `overflow: hidden` removed from LinksList selector

guilhermelMoraes:
### Related Ticket(s)

Pattern / Component - Leadspace Block & Link List - keyboard focus is not covering all four sides of the element. #2612

### Description

**Previous**
68747470733a2f2f696d616765732e7a656e68756275736572636f6e74656e742e636f6d2f3565323161376234373031313861343938343239613230652f64666665343431622d306237322d343566312d383832312d323463663531353164393663

**Current**
Captura de Tela 2020-06-04 às 12 30 06
Captura de Tela 2020-06-04 às 12 30 41

### Changelog

**Removed**

- `overflow: hidden` from `LinksList` `list` selector;

### DIFF
--- a/packages/styles/scss/components/link-list/_link-list.scss
+++ b/packages/styles/scss/components/link-list/_link-list.scss
@@ -47,6 +47,8 @@
 
           svg {
             align-self: start;
+            position: relative;
+            top: 1px;
             margin-left: 0;
             margin-right: $carbon--spacing-03;
           }
--- a/packages/styles/scss/components/link-list/_link-list.scss
+++ b/packages/styles/scss/components/link-list/_link-list.scss
@@ -21,8 +21,6 @@
       color: $text-01;
     }
     &__list {
-      overflow: hidden;
-
       &__CTA {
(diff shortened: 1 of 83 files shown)


packages/styles/scss/components/link-list/_link-list.scss
```diff
@@ -21,8 +21,6 @@
       color: $text-01;
     }
     &__list {
-      overflow: hidden;
```
jeffchew, line 24: @kennylam I was checking the changelog and this was added for the PR https://github.com/carbon-design-system/ibm-dotcom-library/pull/2452. Just making sure this wasn't added for a particular reason or we're good with this change.

cc: @guilhermelMoraes

guilhermelMoraes: I thought that too. Another workaround for this problem would be set 1px padding for the list, but I don't know if we can do that

kennylam: Yep, this looks fine to me

PR #2667, opened 2020-06-05
test(Selenium): adding selenium e2e tests

chetankan:
### Related Ticket(s)

{{https://app.zenhub.com/workspaces/ibmcom-library-5d449f3642eb1962336cbe52/issues/carbon-design-system/ibm-dotcom-library/2455}}

### Description

Added following test scripts
Linklist
Footer
Masthead
Contentblocksimple
Leadspceblock
Pictogram
CalloutQuote
Contentblocksimple
Imagewithcaption
TOC

Added one folder 'URL_List'


chetankan:
Removed chromedriver and init file.

jeffchew:
@chetankan @Praveen-sr fyi, I pushed an initial change in order to let it run at least in a linux environment. However, to get this to work on linux, we would have to get selenium (which does exist) in the Jenkins agent, but would have to include `dotenv` and `chromedriver`, so we will have to do a customization there.

One thought, I'm wondering if we should be using WebDriverJS instead? It aligns fairly nicely with the rest of our repo, and we can pull those in and run agnostically across any operating system? If we're ok with that, I can take some time to convert the existing tests for WebdriverJS.

asudoh:
So we now seem to have a different error than the original one, basically need to get the offline mirror back in sync. I have the missing file in my local, please let me know if `yarn install` without offline mode doesn't download it automatically somehow.

jeffchew:
> So we now seem to have a different error than the original one, basically need to get the offline mirror back in sync. I have the missing file in my local, please let me know if `yarn install` without offline mode doesn't download it automatically somehow.

You're probably right. I tried running a `yarn install` after `yarn clean`, still getting the error:

Screen Shot 2020-06-30 at 1 03 19 PM

jeffchew:
@asudoh I'm also debating about ditching this approach entirely in favor of programmatically creating e2e tests like how you have set up with the karma accessibility test runner. How easy would it be to adapt component level tests similar to what's in here and get harnessed into what you are running with karma?

asudoh:
@jeffchew Sounds interesting, does it mean automatic walk-through of each stories and running tests against them? If so, wondering how test cases for each stories are same/different. For `karma-accessibility-checker` all the test cases are the same because the only thing the test case does is letting DAP engine scan the DOM.

jeffchew:
> @jeffchew Sounds interesting, does it mean automatic walk-through of each stories and running tests against them? If so, wondering how test cases for each stories are same/different. For `karma-accessibility-checker` all the test cases are the same because the only thing the test case does is letting DAP engine scan the DOM.

Hey @asudoh , what I meant was using the same approach where for each component, you set the `ReactDOM.render` of the story component with the corresponding props for that particular e2e test. Each test could be set for whatever variations you need to test for each component. For this PR, currently what is being rendered is the actual URL via selenium then using WebdriverIO API to traverse the DOM and manipulating the page for each e2e test.

asudoh:
@jeffchew I see, thank you for clarifying. Given Selenium nature where test cases are defined in out-of-browser process (e.g. node.js or Java), we need to create a test harness UI (pseudo app) to do the React rendering for each test cases, but it'll definitely be a good way to do E2E testing given doing so allows us to isolate out the Storybook-specific stuffs.

asudoh:
> You're probably right. I tried running a yarn install after yarn clean, still getting the error:

Screen Shot 2020-06-30 at 1 03 19 PM

I have seen this several times in my other efforts. For all of those cases I solved it be editing `yarn.lock` to remove all `@babel/core` entries and re-running `yarn install`.

jeffchew:
Closing this PR out. I'm going to investigate another way of approaching this per discussion above, but will keep the branch alive for now.

PR #2674, opened 2020-06-05
fix(Footer): fixed footer focus

andysherman2121:
### Related Ticket(s)

https://github.com/carbon-design-system/ibm-dotcom-library/issues/2304

### Description

Active focus was being cut off on the first column footer items. It was moving slightly into the gutter and not showing up.


packages/styles/scss/components/footer/_footer-nav-group.scss
```diff
@@ -13,7 +13,7 @@
   $TEMP--footer-nav-group-type: 'heading-02';
 
   .#{$prefix}--footer-nav-group {
-    padding-left: 0;
+    padding-left: 1px;
```
annawen1, line 16: wondering if we can just remove `padding-left` entirely?

andysherman2121: Yeah you're right Anna, i'll update it

kennylam: @andysherman2121 Actually, removing those `padding-left: 0` are now throwing it off the grid. I think we need to put those back. You can also add `outline-offset: -1px` here to get similar results.

https://github.com/carbon-design-system/ibm-dotcom-library/blob/25696ae9861545c045e228c10e37bc7645a4308d/packages/styles/scss/components/footer/_footer.scss#L51-L53

andysherman2121: @kennylam Ah gotcha i'll. adjust it

Issue #2690, opened 2020-06-08
Component: Video Thumbnail displays when using Card with video type in LinkList


annawen1:
##Detailed description

> Describe in detail the issue you're having.

The Kaltura thumbnail is displayed when using the CTA card style, video type in LinkList. The thumbnail should not be showing.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

CardCTA

## Additional information

<img width="665" alt="Screen Shot 2020-06-08 at 1 34 01 PM" src="https://user-images.githubusercontent.com/54281166/84062535-719fc380-a98d-11ea-8657-12a4aba5b091.png">

PR #2691, opened 2020-06-08
fix(CTA): disable thumbnail image when using video card in linklist

annawen1:
### Related Ticket(s)

Component: Video Thumbnail displays when using Card with video type in LinkList #2690

### Description

When using the video card in `<LinkList />`, it's pulling in the generating the kaltura thumbnail by default. Need to pass in a prop, `disableImage` to prevent the thumbnail from being generated.

BEFORE:
Screen Shot 2020-06-08 at 1 34 01 PM

NOW:
Screen Shot 2020-06-08 at 3 51 06 PM

### Changelog

**Changed**

- pass `disableImage` prop to from LinkList to CTA

### DIFF
--- a/packages/react/src/components/LinkList/__stories__/LinkList.stories.js
+++ b/packages/react/src/components/LinkList/__stories__/LinkList.stories.js
@@ -70,6 +70,7 @@ Default.story = {
               src: '0_uka1msg4',
               type: 'video',
             },
+            disableImage: true,
           },
         ];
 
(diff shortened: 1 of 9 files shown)


packages/react/src/components/LinkList/__stories__/LinkList.stories.js
```diff
@@ -70,6 +70,7 @@ Default.story = {
               src: '0_uka1msg4',
               type: 'video',
             },
+            disableImage: true,
```
jeffchew, line 73: @annawen1 I wonder if LinkList should disable the video and not have it set in the story? we probably can't rely on the adopter to set this themselves.

annawen1: hmm good point, yeah I'll set the prop in the component itself

Issue #2398, opened 2020-05-14
Pattern: Callout Quote: Incorrect text styles


ltitle:
## Callout quote - Text styles

> Describe in detail the issue you're having.

Quote should be using type token $quotation-02 for max, XL, large, and medium breakpoints
Quote should be using type token $quotation-01 for small breakpoint
You will see that these are text tokens that use the IBM Plex Serif font. 

Currently the implementation is using an expressive heading style that does not follow the proper quote tokens and is IBM Plex Sans
Quote type style is also not changing for small breakpoint

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

Pattern: Callout Quote

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Expected type styles from visual specs:
![image](https://user-images.githubusercontent.com/57198117/81995302-f0684180-9617-11ea-9e2e-77d0326b6f81.png)

![image](https://user-images.githubusercontent.com/57198117/81995319-f9591300-9617-11ea-8182-2bb13f85b1cf.png)

As seen in Storybook (https://ibmdotcom-react.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/2336/index.html?path=/story/patterns-blocks-calloutquote--default):

![image](https://user-images.githubusercontent.com/57198117/81995515-91ef9300-9618-11ea-824e-e1a473098375.png)

![image](https://user-images.githubusercontent.com/57198117/81995532-99af3780-9618-11ea-982a-ab85fdc38598.png)

> What browser are you working in?

Chrome

https://ibmdotcom-react.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/2336/index.html?path=/story/patterns-blocks-calloutquote--default

PR #2695, opened 2020-06-09
style(Quote): serif font added; $quotation-01 token added

guilhermelMoraes:
### Related Ticket(s)

Pattern: Callout Quote: Incorrect text styles #2398

### Description

The `quote` component was missing the `quotation-01` token. Also, it looks like we found a bug in the carbon expressive theme. The `quotation-` tokens aren't making use of the IBM Plex Serif mixin.
Captura de Tela 2020-06-09 às 10 19 15

### Changelog

**New**

- `IBM Plex Serif` import;
- `quotation-01` token for `sm` breakpoint in the `Quote` component;

### DIFF
--- a/packages/styles/scss/components/quote/_quote.scss
+++ b/packages/styles/scss/components/quote/_quote.scss
@@ -29,12 +29,14 @@
 
     &__copy {
       @include carbon--make-col-ready;
-      @include carbon--type-style(quotation-02, true);
+      @include carbon--type-style('quotation-01', true);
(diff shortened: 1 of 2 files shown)


guilhermelMoraes:
Looks like the `font-weight` is thinner than in the [visual specs](https://ibm.ent.box.com/file/652630918604), despite we're using the correct tokens now. 

Reached @oliviaflory for that, to see what she think of it

Issue #2656, opened 2020-06-04
Footer IBM links violate react rules


kyjak:
1.7.0 version
the links in the footer should have a unique key. This is not crucial, but when
you do some work on footer, add a simple counter based key to list items to clear this.

![image](https://user-images.githubusercontent.com/15310250/83756367-422c4680-a66f-11ea-8497-b7af3e60f184.png)

PR #2703, opened 2020-06-09
fix(LegalNav): `key` prop added to the trustE <li/> item

guilhermelMoraes:
### Related Ticket(s)

Footer IBM links violate react rules #2656

### Description

Added a `key` property to the `<li />` item, so the dev console doesn't throw an error.

### Changelog

**New**

- `key` prop in the `<li/>` item;

### DIFF
--- a/packages/react/src/components/Footer/LegalNav.js
+++ b/packages/react/src/components/Footer/LegalNav.js
@@ -17,13 +17,15 @@ const { prefix } = settings;
 /**
  * Placeholder <li/> element for injection of the TrustE cookie preferences link
  *
+ * @param {number} key - the key for the JSX object
  * @returns {*} JSX object
  */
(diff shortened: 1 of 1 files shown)


packages/react/src/components/Footer/LegalNav.js
```diff
@@ -73,7 +75,8 @@ function renderListItems(links) {
     );
   });
 
-  renderedLinks.push(renderTrusteItem());
+  const key = parseFloat(renderedLinks[renderedLinks.length - 1].key) + 1;
```
asudoh, line 78: Given the `.map()` loop above has `key={index}` and we seem to be trying to do `lastIndex + 1`, I guess we want below. My apologies if I'm missing anything!

```suggestion
  const key = renderedLinks.length + 1;
```

PR #2707, opened 2020-06-09
fix(Masthead): css fixes for RTL

greeshmareddyg:
### Related Ticket(s)

https://github.com/carbon-design-system/ibm-dotcom-library/issues/1744

### Changelog

**Changed**

- master head dropdown menu - rest all look good.

image

### DIFF
--- a/packages/styles/scss/components/masthead/_masthead.scss
+++ b/packages/styles/scss/components/masthead/_masthead.scss
@@ -352,6 +352,12 @@ $search-transition-timing: 95ms;
   .#{$prefix}--header__action.#{$prefix}--overflow-menu {
     height: $spacing-09;
   }
+
+  html[dir='rtl'] {
+    .#{$prefix}--overflow-menu-options.#{$prefix}--overflow-menu-options--open {
+        right: 0 !important;
+    }
+  }
 }
 
 @include exports('masthead') {
--- a/packages/styles/scss/components/masthead/_masthead.scss
+++ b/packages/styles/scss/components/masthead/_masthead.scss
@@ -320,6 +320,7 @@ $search-transition-timing: 95ms;
     background-color: $ui-background;
     position: fixed;
     z-index: 99999;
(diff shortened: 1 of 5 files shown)


packages/styles/scss/components/masthead/_masthead.scss
```diff
@@ -352,6 +352,12 @@ $search-transition-timing: 95ms;
   .#{$prefix}--header__action.#{$prefix}--overflow-menu {
     height: $spacing-09;
   }
+
+  html[dir='rtl'] {
+    .#{$prefix}--overflow-menu-options.#{$prefix}--overflow-menu-options--open {
+        right: 0 !important;
```
jeffchew, line 358: would this work without `!important`?

greeshmareddyg: @jeffchew I tried to avoid it but I had to use it here, otherwise its taking default left as the full width of menu in opposite way.

jeffchew:
Screen Shot 2020-06-09 at 6 08 14 PM

@greeshmareddyg can you check prettier and see what is off?

packages/styles/scss/components/masthead/_masthead.scss
asudoh: Just FYI; Carbon core encourages the users to use a toolchain to replace `left` with `right`, etc.: https://github.com/carbon-design-system/carbon/blob/v10.13.0/packages/components/docs/g11n.md

Given many components in this library inherits Carbon core components, wondering our library would fit in the same category.

In post-IE11 world, we can use `inset-block-start`: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties

jeffchew:
Thank you @greeshmareddyg , sorry we had to go back and forth on this but looks like this patch will work for now without tripping the linter:

https://github.com/carbon-design-system/ibm-dotcom-library/pull/2743

Will close this out for now until we can find a better solution for this.

Issue #2614, opened 2020-06-01
Pattern: Leadspace block: text colors


ltitle:
## Leadspace block: text colors

> Describe in detail the issue you're having.

Some text fields are not using proper type tokens

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

Leadspace-block pattern (will also apply to all content block patterns)

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Applies to the following fields: Subheadline and intro paragraph

**As seen in storybook:**
text color: 0,0,0
![image](https://user-images.githubusercontent.com/57198117/83428460-68987a80-a400-11ea-9c7b-46b6f6fbed7d.png)
![image](https://user-images.githubusercontent.com/57198117/83428345-35ee8200-a400-11ea-8a6d-d645fd368ed5.png)

**Expected**:
color token: $text-01

![image](https://user-images.githubusercontent.com/57198117/83428374-43a40780-a400-11ea-8d36-2c403fc55539.png)
![image](https://user-images.githubusercontent.com/57198117/83428388-47d02500-a400-11ea-8255-59469a0cd680.png)

> What browser are you working in?

Chrome

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

IBM.com Reboot

## Steps to reproduce the issue

https://ibmdotcom-react-experimental.mybluemix.net/?path=/story/patterns-blocks-leadspaceblock--default

PR #2710, opened 2020-06-10
fix(LeadSpaceBlock): `$text-01` added to the LeadspaceBlock copy

guilhermelMoraes:
### Related Ticket(s)

Pattern: Leadspace block: text colors #2614

### Description

 - `$text-01` token applied for the `LeadspaceBlock` subheadline and intro paragraph;
 - Fixed type style for the intro paragraph as well.

Obs.: the token description in the [Carbon guidelines](https://www.carbondesignsystem.com/guidelines/themes) says the token color hex code is #171717, but as seen in the storybook the hex code color being applied is the #161616. 

### Changelog

**New**

- `color: $text-01` for both subheadline and intro paragraph.

**Changed**

- intro paragraph type style from `expressive-heading-01` to `expressive-heading-03`.


packages/styles/scss/patterns/blocks/leadspace-block/_leadspace-block.scss
```diff
@@ -30,14 +30,21 @@
...
+        color: $text-01;
+      }
+
+      &__copy {
+        color: $text-01;
+        p {
+          @include carbon--type-style('expressive-heading-03', true);
+        }
+      }
+
```
kennylam, line 47: @guilhermelMoraes Just confirming, but I think these changes have already been made in [this PR](https://github.com/carbon-design-system/ibm-dotcom-library/pull/2712/commits/8ee29fa833d588b39c4a5f59661fe360374e44f9#diff-6c2d7f4ee5a9324ff588e92bb9a91337R35)? It's the same token change to `ContentBlock`, which is what `LeadspaceBlock` is using.

jeffchew: If we're setting the color token already in `ContentBlock`, perhaps the only change needed is the expressive heading token?

guilhermelMoraes: Yeah, I think so. I've done this update before the #2712, but I'm going to change that since it's affecting only the `LeadspaceBlock`

PR #2717, opened 2020-06-10
fix(card): themes enabled

greeshmareddyg:
### Related Ticket(s)

https://github.com/carbon-design-system/ibm-dotcom-library/issues/2481

### Changelog

**Changed**

- card - component : enabled themes g90 and g100


packages/react/src/components/Card/Card.js
```diff
@@ -127,6 +132,11 @@ export const cardPropTypes = {
    */
   heading: PropTypes.string,
 
+  /**
+   * Theme name.
+   */
+  theme: PropTypes.oneOf(['white', 'g10', 'g90', 'g100']),
```
jeffchew, line 138: @greeshmareddyg can you also set a defaultProp?

jeffchew:
@greeshmareddyg your snapshots appear to be out of date too. Also check and see @asudoh 's latest on theme switching, we may not need to provide a knob in the story itself anymore.

packages/react/src/components/Card/Card.js
```diff
@@ -19,6 +19,7 @@ import { LinkWithIcon } from '../LinkWithIcon';
 import PropTypes from 'prop-types';
 import React from 'react';
 import settings from 'carbon-components/es/globals/js/settings';
+import TableOfContents from '../TableOfContents/TableOfContents';
```
jeffchew, line 22: @greeshmareddyg this might have been a mistake to add `TableofContents` here?

greeshmareddyg: yeah, updated Jeff.

asudoh:
> Also check and see @asudoh 's latest on theme switching, we may not need to provide a knob in the story itself anymore.

Good point, wondering if we still need a new React prop.

jeffchew:
@greeshmareddyg can you look at the merge conflicts?

greeshmareddyg:
@jeffchew @asudoh I think we don't need to merge this, because the recent theme changes taking care of the theme for the component.

jeffchew:
Thanks @greeshmareddyg ! Will close this then.

Issue #2725, opened 2020-06-11
Component : CTA - Keyboard -The Focus is going out of the model page. 


Praveen-sr:
## Detailed description

Issue :- The Focus is going out of the model page. 

Expected Result:-
When the model page is open focus should be trapped with the model page, it should not go to the browser or parent page.

diegomarcuz:
Hey @Praveen-sr, I need your help
I didn’t understand well what you mean here, could you please clarify?

Praveen-sr:
@diegomarcuz 
1. On CTA,  go to knobs section
2. In Type edit/ change it to VIdeo
3. Click on the link and a video modal opens 
4. Here when the user navigates using tab/shift tab key, the focus is expected to be trapped within the modal

Ram, re tested this scenario on react-staging server, issue is fixed.

diegomarcuz:
As Praveen commented above and on the QA review meeting issue was fixed.

Just a representation what happens (on the close button, focus needs to happen when tab key is clicked)

![Screen Shot 2020-06-29 at 10.32.37.png](https://images.zenhubusercontent.com/5eed25994ac8fb057241e3b2/cf5b5823-a519-4824-aafa-6c9344cd5313)

cc: @RobertaJHahn

PR #2729, opened 2020-06-11
chore(devenv): remove `.bx--visually-hidden` override

asudoh:
### Related Ticket(s)

Refs #2725.

### Description

This change removes our CSS override for `.bx--visually-hidden` class that is defined in our Storybook environment.

Setting `display:none` to that class makes the DOM element no longer keyboard focusable that is against its purpose, which is, "hidden visually, but navigable with screen readers".

Carbon modal's focus wrap behavior relies on elements with `.bx--visually-hidden` keyboard-focusable to detect changes in focused element that goes out of modal.

### Changelog

**Removed**

- Our CSS override in Storybook for `.bx--visually-hidden`.

### DIFF
--- a/packages/react/.storybook/_container.scss
+++ b/packages/react/.storybook/_container.scss
@@ -36,10 +36,6 @@
   width: 100%;
 }
 
-.bx--visually-hidden {
-  display: none;
-}
-
 // hide the cookie button
 #teconsent {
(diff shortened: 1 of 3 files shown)


jeffchew:
LGTM, thank you @asudoh !

PR #2732, opened 2020-06-11
fix(storybook): adding disabled theme fix

jeffchew:
### Related Ticket(s)

Refs #2733

### Description

This is a fix to unset the theme for the story if `carbon-theme` is disabled. Right now it is setting it in the `white` theme, or whatever theme was previously set if coming from another component, so this is especially visible for components like the `Masthead`, `Footer`, etc.

### Changelog

**Changed**

- unsetting theme setting for disabled stories


jeffchew:
Closing this in favor of https://github.com/carbon-design-system/ibm-dotcom-library/pull/2735.

PR #2739, opened 2020-06-12
feat(FeatureCard): added codesandbox example implementation

RaphaelAmadeu:
### Related Ticket(s)

#1903 

### Description

Added codesandbox example implementation on `README` file.


RaphaelAmadeu:
> Just a thought, but do we want to set `"private": true` in these example packages?
> 
> cc: @jeffchew @asudoh

I've

PR #2740, opened 2020-06-12
test(integration): support new dependencies in packages

asudoh:
### Description

This change fixes an issue in build integration tests where adding dependencies to our packages causes test failure. Fix is done by modifying `package.json` in the packages in temporary directory to cross-reference dependencies within temporary directory, instead of pointing to ones in NPM registory.

My apologies @kennylam for the CI failure.

### Changelog

**Changed**

- The scripts for build integration tests so that the right `package.json` in our build artifacts are used for tests, by changing how to bring in and refer to build artifacts in our tests.


asudoh:
Re-targetted to relese branch upon our discussion, CC @kennylam @jeffchew

Issue #2755, opened 2020-06-15
Component :- Table of Content :- Mismatch in 'Jump to' section on selecting title.(Mobile devices)


chetankan:
Issue Description:- In table of content component there is a mismatch in displaying title selected and title getting displayed in 'Jump to' section.

## Steps to reproduce the issue

1. Enter url in mobile browser.
2. Select TOC component
3. Click on 'Jump to' option
4. Select 3rd or 4th title from drop down.
5. Check the title displayed in jump to section after selecting from drop down.

OS:- iOS 12, 13  Android 8,9
Browsers:- Chrome, Safari, UC browser

URL:- https://ibmdotcom-react-staging.mybluemix.net/?path=/story/components-table-of-contents--manually-define-menu-items

![TOC in stage env.jpg](https://images.zenhubusercontent.com/5e21a841639f517d456f7e43/090fd245-ab9c-4a95-b425-55dd07ebec22)

PR #2760, opened 2020-06-15
fix(TableOfContents): added offset to fix scroll issues on mobile

RaphaelAmadeu:
### Related Ticket(s)

#2755 

### Description

Added  an offset to section detection  in order to avoid mismatching  section issue on mobile view.


packages/react/src/components/TableOfContents/TableOfContents.js
```diff
@@ -124,7 +124,7 @@ const TableOfContents = ({
       .filter((elem, index, arr) =>
         elem.height === null
           ? arr[index - 1].position < arr[index - 1].height
-          : elem.position > -elem.height
+          : elem.position - 50 > -elem.height
```
asudoh, line 127: Just wondered where the `50` number came from, but we can go forward as long as you see this fixes the problem (and I believe so).

RaphaelAmadeu: That's close to the size of our navigation bar, so it would help in cases when the nav is shown, so if the component is bellow it, the next version can be considered

jeffchew:
@RaphaelAmadeu can you target the release branch for this?

packages/react/src/components/TableOfContents/TableOfContents.js
annawen1: I wonder if it would be clearer to maybe create a `const navBarHeight = 50` and use that so that it's more obvious?

jeffchew:
@RaphaelAmadeu I'm going to repoint this back to master then let's cherry pick the change to the release branch.

packages/react/src/components/TableOfContents/TableOfContents.js
asudoh: Yeah I think `const navBarHeight = 50` helps - Thanks @RaphaelAmadeu for your clarification and @annawen1 for your suggestion!

PR #2765, opened 2020-06-15
chore(release): fixing tag release script to adapt to changelog script

jeffchew:
### Related Ticket(s)

Refs #2747

### Description

Because of the introduction of beta releases, the changelog harvesting
method was starting to fail. Used this opportunity to convert the
existing bash script to a node script, and also modified the tag release
script to use the new changelog fetching method.

### Changelog

**Changed**

- Updated tag-release script to use node instead
- Updated tag-release script to utilize `get-changelog` script
- Abstracted `get-changelog` from `beta-releases` script


jeffchew:
> I think we could replace raw `https.request` calls with `node-fetch` to make the code more idiomatic, but it's a future thing and the code looks good. Thanks @jeffchew!

Agreed, that would make it nicer looking for sure! We can do in a separate PR.

Issue #1900, opened 2020-03-27
Create codesandbox for ContentGroupCards


jeffchew:
#### User Story

> As a `[user role below]`:
IBM.com Library adopter

> I need to:
see the `ContentGroupCards` in an isolated codesandbox

> so that I can:
see how I can easily integrate into my application

#### Additional information

 - Should create under the `ibmdotcom-bot` github user under https://codesandbox.io
- 1Password access for `ibmdotcom-bot`: https://ibm.ent.box.com/notes/496014563545
 - Template available under codesandbox: https://codesandbox.io/dashboard/templates

#### Acceptance criteria
 - [ ] Codesandbox created for `ContentGroupCards`
 - [ ] Link added to README

PR #2773, opened 2020-06-16
feat(content-group-cards): added codesandbox example implementation

RaphaelAmadeu:
### Related Ticket(s)

#1900 

### Description

I've added a codesandbox example implementation to the `README`


packages/react/examples/codesandbox/patterns/blocks/ContentGroupCards/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "description": "",
+  "main": "index.html",
+  "scripts": {
+    "start": "parcel index.html --open",
+    "build": "parcel build index.html"
+  },
+  "dependencies": {
+    "@carbon/ibmdotcom-react": "1.7.0",
+    "@carbon/ibmdotcom-styles": "1.7.0",
+    "@carbon/icons-react": "10.11.0",
```
asudoh, line 13: Just wondered if we can use newer version of `@carbon/icons-react`, for the sake of build speed. I believe it's compatible even with older version of `carbon-components-react` library.

```suggestion
    "@carbon/icons-react": "10.12.0",
```

PR #2786, opened 2020-06-17
chore(revert): revert markdown to 1.7.0

kennylam:
### Description

This reverts changes made in https://github.com/carbon-design-system/ibm-dotcom-library/pull/2715 due to build errors. Will target for `1.9.0` release.

### Changelog

**Removed**

- `marked` and `dompurify` packages
- full markdown support
- Expressive `list` styles


kennylam:
> LGTM basically, just one question; Was the removal of `_list.scss` intentional? Thanks @kennylam!

@asudoh Actually you're right, I could have left that in for now with no issues. I removed it because it was for the `markdownToHTML` list examples, which had to be removed as well. But no worries, we can re-add later np. Thanks!

Issue #2606, opened 2020-06-01
Pattern: Make the code changes to the Table of Contents that improve the colors and hover states


RobertaJHahn:
#### User Story

> As a `[user role below]`:
IBM.com Library developer

> I need to:
change the `Table of Contents`

> so that I can:
provide the ibm.com adopter developers patterns/components they can use to build ibm.com web pages

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- **This issue is created to separate the scrolling improvements from the color and hover states changes.  The scrolling improvement issue is (#2424)**
- Updated functional specs (v 1.3) uploaded to box: https://ibm.box.com/s/yfuicifhfeaefp4rl0pp62dpeacr18ry
- Functional specs issue (#1783  )
- Prod QA testing issue (#2425 )

#### Functional spec changes and highlights
- In desktop experience, 'on hover' of Anchor Link in selected state will display Scroll Indicator in hover state. Currently, no hover interaction occurs.
- In desktop experience, 'on hover' of Anchor Link in unselected state will display Scroll Indicator in hover state and Anchor Link in hover state. Currently, no hover interaction occurs.
- In mobile experience, the "Jump to..." label will not be clickable/tappable. Currently on click/on tap, the dropdown closes and Anchor Link Section Name updates to "Jump to...".

#### Acceptance criteria
- [ ] Include README for the react pattern and corresponding styles
- [ ] Minimum 80% unit test coverage
- [ ] A comment is posted in the Design QA issue, tagging Wonil and Roberta, when development is finished
 - [ ] A comment is posted in the Prod QA issue, tagging Praveen and Chetan, when development is finished

PR #2800, opened 2020-06-17
fix(TableOfContents): fixed missing hover and hidden default label

RaphaelAmadeu:
### Related Ticket(s)

#2606 

### Description

Added hover states to the active item in desktop view, and removed the "Jump to..." label from the select list, but it will remain as a holder until the `TableOfContents` is activated


kennylam:
@RaphaelAmadeu Should we add `width: 100%` here so the entire `select` is clickable in mobile?

https://github.com/carbon-design-system/ibm-dotcom-library/blob/725ffe43f78b05c5df715b86be18b0d647ebc710/packages/styles/scss/components/tableofcontents/_tableofcontents.scss#L66-L75

jeffchew:
@RaphaelAmadeu can you update jest snapshots?

Issue #2664, opened 2020-06-04
DotcomShell:  Accessibility:  "skip to main content link" is not working


Palke:
Using basic TAB navigation within DotcomShell activates "skip to main content link" when in focus, but using key **enter** doesn't change focus to the main content - nothing happens.

There is also no element with id "#main-content" in DotcomShell.

Steps to replicate:
1) Open https://ibmdotcom-react.mybluemix.net/iframe.html?id=components-dotcom-shell--default&viewMode=story in new tab
2) Use TAB to navigate within the content until "skip to main content link" appears
3) Press enter

image
<img width="707" alt="image" src="https://user-images.githubusercontent.com/54019675/83799223-28f6ba80-a6ae-11ea-97d2-f3a448780dc1.png">

Expected:
Focus should be moved into the main content area and skip Masthead navigation.

v1.7.0

PR #2802, opened 2020-06-17
chore(DotcomShell): use current patterns as example content

kennylam:
### Related Ticket(s)
https://github.com/carbon-design-system/ibm-dotcom-library/issues/2664, https://github.com/carbon-design-system/ibm-dotcom-library/issues/2665

### Description

The current `DotcomShell` story uses example content directly from Carbon, causing some inconsistencies with spacing and grid layout. This replaces `DotcomShell` example content with the latest patterns from the IBM.com Library to make everything consistent with our library.

### Changelog

**New**

- `DotcomShell` story uses current pattern examples

**Changed**

- fix invalid HTML
- add `main-content` ID for accessibility Skip to Content (#2664)


packages/react/src/components/DotcomShell/__stories__/data/content.js
```diff
@@ -1,160 +1,430 @@
...
+  ContentBlockCards,
+  ContentBlockSegmented,
+  ContentGroupHorizontal,
+  CTASection,
+  Layout,
+  LeadSpaceBlock,
+  LogoGrid,
+  FeatureCardBlockLarge,
+  TableOfContents,
+} from '@carbon/ibmdotcom-react';
```
jeffchew, line 21: @kennylam we can probably pull in relatively than from the node module.

Issue #1626, opened 2020-03-09
Learn page: inconsistent spacing between body copy and CTA across patterns


oliviaflory:
## Detailed description

> Describe in detail the issue you're having.
There are varying amounts of spacing between the body copy and CTA card in the Content Block mixed and content block media patterns. (I think these are the right blocks...but not 100% positive). Seen in the pattern **below** the pictograms and the pattern **above** feature card at the bottom of the page.

![Screen Shot 2020-03-06 at 4.46.36 PM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/5410dd60-37a4-4fa7-81bd-1beb6a74b51b)

![Screen Shot 2020-03-06 at 4.46.14 PM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/218f7198-76bf-489d-835b-0467875ac609)

> Is this a feature request (new component, new icon), a bug, or a general
> issue?
bug

> Is this issue related to a specific component?
Content block mixed
Content block media

> What did you expect to happen? What happened instead? What would you like to
> see changed?

> What browser are you working in?
Chrome

> What version of the IBM.com Library are you using?

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?
Web simplification

## Steps to reproduce the issue

1. [Learn page](http://ibmdotcom-nextjs-test.mybluemix.net/learn)

## Additional information

- Screenshots or code
- Notes

photodow:
@oliviaflory is this really a bug? Is that first images' CTA coming from a content group or content block? From a hierarchy standpoint, the content group's CTA has tighter spacing than the content block's CTA.

oliviaflory:
@photodow sorry just seeing this, honestly, I don't know anymore. The Learn page demo link gives me a 404 and it looks like I wasn't sure which blocks or groups they were so I'm not positive. 

Considering I opened this on March 9, and we are now at the end of June, it's possible this doesn't even exist anymore or is intentional.

PR #2803, opened 2020-06-17
fix(ContentBlockMixed): styling to content-block-mixed

ariellalgilmore:
### Related Ticket(s)

#1626 

### Description

Added styling to content-block-mixed in order to have consitent spacing with content-block-media

Screen Shot 2020-06-17 at 12 13 57 PM

### Changelog

**New**

- added margin bottom to the last child of content-block-mixed


jeffchew:
Hey @ariellalgilmore , this should be `package: react`, the patterns one is a deprecated label we no longer use.

kennylam:
@ariellalgilmore These spacing fixes should probably be applied globally and at a block/group level. We can get to these once we're all on the same page with spacing fixes.

PR #2811, opened 2020-06-17
feat(LinkWithIcon): Link with icon component enhancements

greeshmareddyg:
### Related Ticket(s)

https://github.com/carbon-design-system/ibm-dotcom-library/issues/2416

### Changelog

**Changed**

- By default the icon should be on the right of the text 
- Visited state in the knobs
- Set icon on right or left of text
- Fixed the  linkwithicon usage in other components.


packages/react/src/components/CTA/TextCTA.js
```diff
@@ -75,23 +71,9 @@ TextCTA.propTypes = {
...
   type: PropTypes.oneOfType([
-    PropTypes.oneOf([
-      'jump',
-      'local',
-      'external',
-      'download',
-      'video',
-      'default',
-    ]),
+    PropTypes.oneOf(['jump', 'external', 'download', 'video']),
```
jeffchew, line 74: This looks like a reduced set of icons, plus we aren't allowing for no icon. Is this part of the specs?

packages/react/src/components/LinkWithIcon/LinkWithIcon.js
```diff
@@ -17,33 +23,120 @@ const { prefix } = settings;
...
+  console.log('visited state', visited);
   return (
     <div
-      className={`${prefix}--link-with-icon__container`}
+      className={cx(`${prefix}--link-with-icon__container`, {
+        [`${prefix}--link-with-icon__container--disabled`]: props.disabled,
+      })}
       data-autoid={`${stablePrefix}--link-with-icon`}>
-      <Link href={href} className={`${prefix}--link-with-icon`} {...props}>
+      <Link
```
jeffchew, line 44: This doesn't look right. `LinkWithIcon` is meant to allow any kind of icons here, and not restricting it to the 4 you have imported. This is a generic component.

packages/react/src/components/LinkWithIcon/LinkWithIcon.js
```diff
@@ -5,8 +5,14 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+import ArrowDown20 from '@carbon/icons-react/es/arrow--down/20';
```
jeffchew, line 8: Not sure if we should be importing these icons on this level. `LinkWithIcon` should allow for positioning of the icon to be left or right, but not restricting what icons are used exactly.

packages/react/src/components/LinkWithIcon/LinkWithIcon.js
```diff
@@ -17,33 +23,120 @@ const { prefix } = settings;
...
+          <RenderIcon className={`${prefix}--link-with-icon_right-icon`} />
+        ) : (
+          ''
+        )}
       </Link>
     </div>
   );
 };
 
+const _renderLinkIcon = type => {
```
jeffchew, line 67: Don't think this logic belongs here. I think we already have this kind of logic in CTALogic. Let's keep `LinkWithIcon` as generic as possible.

packages/react/src/components/LinkWithIcon/LinkWithIcon.js
```diff
@@ -17,33 +23,120 @@ const { prefix } = settings;
...
+  href,
+  copy,
+  type,
+  direction,
+  visited,
+  ...props
+}) => {
+  const RenderIcon = _renderLinkIcon(type);
+
+  console.log('visited state', visited);
```
asudoh, line 37:
```suggestion
```

jeffchew:
Hi @greeshmareddyg , going to close this out. @kennylam will be implementing on his end as he needs this for a story he's working on.

Issue #2385, opened 2020-05-14
Pattern: CalloutQuote - Mobile - Left side padding is not as per visual specs.


Praveen-sr:
## Detailed description
Device - iPhone 11, Samsung S10

Issue :- The spacing, left side padding on callout quote is not as per specs. It must be 32px.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/1bc6c30f-67ae-46c3-bad4-bf6fff00b890)

PR #2831, opened 2020-06-18
fix(Quote): quotation mark spacing adjusted 

guilhermelMoraes:
### Related Ticket(s)

Pattern: CalloutQuote - Mobile - Left side padding is not as per visual specs. #2385

### Description

**Previous**
Captura de Tela 2020-06-18 às 15 35 07

**Current**
Captura de Tela 2020-06-18 às 15 36 16

### Changelog

**New**

- new absolute position to the quotation mark at small breakpoint.

### DIFF
--- a/packages/styles/scss/components/quote/_quote.scss
+++ b/packages/styles/scss/components/quote/_quote.scss
@@ -45,6 +45,7 @@
       @include carbon--type-style('quotation-01', true);
       @include carbon--font-family('serif');
       @include carbon--breakpoint-down('md') {
+        // this value was chosen by visually stay the closest as possible from the quotation text in the small breakpoint
         left: 1.4rem;
       }
       @include carbon--breakpoint('md') {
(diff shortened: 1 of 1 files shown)


packages/styles/scss/components/quote/_quote.scss
```diff
@@ -44,6 +44,9 @@
     &__mark {
       @include carbon--type-style('quotation-01', true);
       @include carbon--font-family('serif');
+      @include carbon--breakpoint-down('md') {
+        left: 1.4rem;
```
asudoh, line 48: Wondered if we can put a comment here explain where the number came from.

guilhermelMoraes: actually, this value was chosen by visually stay the closest as possible from the quotation text. Do you think it's important to add a comment explaining this?

jeffchew: I think it is worth adding a note for other developers that view this file in the future since it doesn't follow our normal conventions of the spacing scale.

Issue #2605, opened 2020-06-01
Callout Quote Block Pattern missing a property for property


kexuibm:
Using Callout Quote Block pattern from release 1.7.0.

It seems like there is a missing property for the Callout Quote specifically the company that provided the quote.

![image](https://user-images.githubusercontent.com/55464272/83412270-d1272d80-a3e7-11ea-878f-70e34039db35.png)

For example in the Visual design, there is no property for the company field( NBCUniversal ).

ljcarot:
@RobertaJHahn do you know who did the func spec for Callout Quote?

RobertaJHahn:
The second link under the copy line should be `copy 2` and not company as suggested by Ke.  The field should also be optional.

kexuibm:
@RobertaJHahn  and @jeffchew I don't see an additional copy field (copy2)  in the  Callout Quote Block Pattern in release 1.8.0.    Can you please re-open this defect?

ljcarot:
@guilhermelMoraes Can you confirm additional copy field was added?

ljcarot:
@kexuibm This is slated for Release 1.9.0.  That's why you didn't see it in 1.8.0.  You will see it in Beta Release 13 that gets created on Friday July 3.

guilhermelMoraes:
@ljcarot it was added, but I didn't add the knob in the panel of the `CallOutQuote` story. My apologies for that. I'm wrapping up a PR right now to fix this anyway, so it will be in better shape when the release is ready.

PR #2835, opened 2020-06-18
fix(Quote): optional `copy2` prop added to the component

guilhermelMoraes:
### Related Ticket(s)

Callout Quote Block Pattern missing property for property #2605

### Description

Optional `copy2`  prop added to the `Quote` component

### DIFF
--- a/packages/react/examples/codesandbox/patterns/blocks/ContentBlockSegmented/index.html
+++ b/packages/react/examples/codesandbox/patterns/blocks/ContentBlockSegmented/index.html
@@ -1,0 +1,23 @@
+<!-- 
+  Copyright IBM Corp. 2016, 2020
+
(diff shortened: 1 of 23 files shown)


jeffchew:
@guilhermelMoraes I'm fine with `copy2`! However, just checking if the extra spacing is supposed to happen after `copy2`, looks like that was added after this change?

cc: @asudoh

guilhermelMoraes:
@asudoh that's because I'm following Roberta's comment on the issue (https://github.com/carbon-design-system/ibm-dotcom-library/issues/2605#issuecomment-638225633).

@jeffchew the space under the `copy2` prop follow the visual specs. Before this new prop, it was there as a `copy` padding.

asudoh:
@guilhermelMoraes Uh I missed that my apologies and thank you for clarifying!

PR #2849, opened 2020-06-22
fix(CardLink): add missing css import

kennylam:
### Description

Add missing css import to `CardLink`

### Changelog

**New**

- import `Card` styles


jeffchew:
@kennylam is this something we should patch?

PR #2850, opened 2020-06-22
fix(CardLink): add missing css

kennylam:
### Description

Add missing css import to `CardLink`

### Changelog

**New**

- import `Card` styles


jeffchew:
@kennylam I'm going to let the tests run through this one to make sure everything is ok, but I approved this already.

Issue #2624, opened 2020-06-01
Pattern: Content group: Spacing


ltitle:
## Content group: Spacing

> Describe in detail the issue you're having.

Different spacing is needed under the intro paragraph for each content group pattern. Therefore the item level needs to be able to hold the extra spacing and turn it on and off

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

Content group
Content group simple
Content group with pictograms
Content group with cards

> What did you expect to happen? What happened instead? What would you like to
> see changed?

### **Content group simple**

**As seen in Storybook:**
Using 64px spacing between intro text and media or content item
64px padding below intro text

Example with media

![image](https://user-images.githubusercontent.com/57198117/83454923-effae380-a42a-11ea-9d7a-457612e6039d.png)

Example with content item

![image](https://user-images.githubusercontent.com/57198117/83454875-df4a6d80-a42a-11ea-9eb7-1d5d7367af97.png)

**As expected:**
Using 32px spacing between intro text and media or content item
32px padding below intro text

Example with media:
![image](https://user-images.githubusercontent.com/57198117/83454634-795de600-a42a-11ea-8c6a-fd8d88d75ccd.png)

Example with content item
![image](https://user-images.githubusercontent.com/57198117/83454987-11f46600-a42b-11ea-9030-886fb32e37c1.png)

---------------

### **Content group with pictograms**

**As seen in Storybook:**
Using 96px spacing between intro text and pictogram item
64px padding below the intro text + 32px padding above the pictogram item

![image](https://user-images.githubusercontent.com/57198117/83455844-972c4a80-a42c-11ea-8a03-329f94115b05.png)

32px padding above and below all pictogram items
![image](https://user-images.githubusercontent.com/57198117/83456276-63055980-a42d-11ea-8c37-8bf875785941.png)

**As expected:**
Using 64px spacing between intro text and pictogram item
Can accomplish by using 32px padding below intro text + 32px padding above pictogram item (just an idea to keep consistent with the need for 32px padding above and the fact that pictogram item seems to already have top padding)

![image](https://user-images.githubusercontent.com/57198117/83455965-c93dac80-a42c-11ea-874e-1d6a096d0bc3.png)

The last pictogram item in the series needs to remove bottom padding to have consistent spacing in Content block mixed groups (otherwise it is adding 32px with the 96px to create 120px between sections that come after pictogram group)
[...]

PR #2851, opened 2020-06-22
fix(content-group-pictograms): fixed bottom spacing

RaphaelAmadeu:
### Related Ticket(s)

#2624 

### Description

Updated spacing on `ConentGroupPictograms` bottom.


packages/styles/scss/patterns/blocks/content-group-pictograms/_content-group-pictograms.scss
```diff
@@ -15,6 +15,10 @@
       }
     }
 
+    .#{$prefix}--content-group {
+      margin-bottom: $layout-05;
+    }
+
```
annawen1, line 21: hey @RaphaelAmadeu I'm looking at the ticket, and @ltitle also mentioned about pictogram bottom padding needing to be removed for the very last pictogram item

RaphaelAmadeu: @annawen1 these remove the padding from the content-group itself, in the end both would have the same result.

annawen1: I guess that works...I'm a little concerned about other components placed into ContentGroup, will they always have 32px padding bottom?

annawen1: ah I was thinking generically, this is specifically the `_content-group-pictograms` styles so this works! thanks :D

RaphaelAmadeu:
reopened

Issue #2619, opened 2020-06-01
Pattern: Leadspace Block: Link list rules


ltitle:
## Pattern: Leadspace Block: Link list rules

> Describe in detail the issue you're having.

Rules between list items are not overhanging 16px on each side
Applies to all breakpoints

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

Leadspace-block using Link List

> What did you expect to happen? What happened instead? What would you like to
> see changed?

**As seen in Storybook:**
Rules do not overhang the grid

![image](https://user-images.githubusercontent.com/57198117/83432158-76e99500-a406-11ea-8922-3c34b8a5a740.png)

**As expected:**
Rules should overhang the grid by 16px on either side
Text and icons stay aligned to the grid

![image](https://user-images.githubusercontent.com/57198117/83432236-95e82700-a406-11ea-8fd5-054cdd366131.png)

> What browser are you working in?

Chrome

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

IBM.com Reboot

## Steps to reproduce the issue

https://ibmdotcom-react-experimental.mybluemix.net/?path=/story/patterns-blocks-leadspaceblock--default

Issue #2618, opened 2020-06-01
Pattern: Leadspace Block: Grid alignment


ltitle:
## Detailed description

> Describe in detail the issue you're having.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

> Is this issue related to a specific component?

> What did you expect to happen? What happened instead? What would you like to
> see changed?

#### Additional Information
- Adjust the Storybook container to be a max of 8 columns

**As seen in Storybook:**
Pattern expanding outside of 8 column container at Large and XL breakpoints
Media is not locking to grid columns

![image](https://user-images.githubusercontent.com/57198117/83431214-00986300-a405-11ea-9530-be084f5e7450.png)
![image](https://user-images.githubusercontent.com/57198117/83431353-363d4c00-a405-11ea-97c5-ee6ce6c1020d.png)

**As expected:**
Pattern will span 90% of the 8 column container on Large and XL breakpoints
Media should span 100% of the 8 column container on Large and XL breakpoints

![image](https://user-images.githubusercontent.com/57198117/83431487-6e448f00-a405-11ea-8e6e-205dbed06c57.png)
![image](https://user-images.githubusercontent.com/57198117/83431649-aa77ef80-a405-11ea-8552-6856c5223121.png)

> What browser are you working in?

Chrome

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

IBM.com Reboot

## Steps to reproduce the issue

https://ibmdotcom-react-experimental.mybluemix.net/?path=/story/patterns-blocks-leadspaceblock--default

PR #2870, opened 2020-06-23
fix(LinkList): 

caiobozato:
### Related Ticket(s)

Pattern: Leadspace Block: Link list rules #2619

### Description

The `Link List` component in it's `Vertical End` version has incorrect margins for the text and icons on sides. As result, the rulers don't overlap the grid:

image

This fix implements the correct margin for the rulers:

image

### Changelog

**New**

- Added negative margin left and right to `Link List` component `Vertical End` in order to provide rulers overlap
-  Added `width` and `max-width` for `Link List` and `Horizontal Rule` in `Leadspace Block` pattern.

**Removed**

- `max-width` for `Link List` component `Vertical End` because it wasn't allowing the component to have 8-column width


packages/styles/scss/patterns/blocks/leadspace-block/_leadspace-block.scss
```diff
@@ -85,6 +87,8 @@
     .#{$prefix}--hr {
       border-color: $ui-04;
       margin-top: 0;
+      width: 100%;
+      max-width: carbon--mini-units(80);
```
annawen1, line 91: not sure if we should set a max-width for the horizontal rule. I'm looking at the visual specs ([https://ibm.ent.box.com/file/662154081174](https://ibm.ent.box.com/file/662154081174)) and it looks like the `hr` is longer than the `LinkList`

packages/styles/scss/components/link-list/_link-list.scss
```diff
@@ -62,14 +62,16 @@
 
         .#{$prefix}--link-list__list--vertical-end & {
           border-bottom: 1px solid $ui-03;
-          max-width: 640px;
+          margin-left: -$carbon--spacing-05;
+          margin-right: -$carbon--spacing-05;
```
kennylam, line 66: Thanks @caiobozato! According to the [visual specs](https://ibm.ent.box.com/file/662151736386), the `max-width` should be set to `640px`. Also, Carbon has a nifty mixin that will apply those negative margins for you 👍 

```suggestion
          max-width: 640px;
          
          @include hang;
```

packages/styles/scss/patterns/blocks/leadspace-block/_leadspace-block.scss
caiobozato: Hello, Anna. Thanks for bringing this up. I'm going to fix it.

caiobozato: @annawen1 as we talked on Slack, I deleted my code but the Horizontal Rule is still not touching the 12th column. This is happening because the `LeadspaceBlock` grid was changed from 12 to 8 in `lg` breakpoint in this issue: #2618.

To fix the HR width, we will be working in a separated PR.

Issue #2295, opened 2020-05-04
Pattern: Link with icon - Enable theming


RobertaJHahn:
#### User Story
As a user of the link with icon, I need a way to use this in the inverse theme so that I can build patterns that require inverse color theme.

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- This pattern should have inverse theming enabled.  Background information is included in issue (#2212)
- This pattern is used in the Callout Quote pattern included in the Reboot Services template.
- **Also see the Epic for the Design and Functional specs information**
- Design issue (#2597) 
- Prod QA testing issue (#2433 )
- [Design Box folder link](https://ibm.ent.box.com/folder/107613134184)
- [Functional spec box link](https://ibm.ent.box.com/file/682502793940)

#### Acceptance criteria
- [x] Include README for the react pattern and corresponding styles
- [x] Create codesandbox example under `/packages/react/examples/codesandbox` and include in README
- [x] Minimum 80% unit test coverage
- [x] If a design is provided, Olivia is included as a Reviewer in the Pull Request
- [x] Provide a direct link to the deploy preview for the designer in the Pull Request description
- [x] A comment is posted in the Design QA issue, tagging Olivia when development is finished
- [x] A comment is posted in the Prod QA issue, tagging Praveen and Chetan, when development is finished
- [x] The Canary URL is added to the Prod QA issue for their testing
- [x] The Storybook link is added to the Design QA issue for their testing

RobertaJHahn:
@greeshmareddyg The design and functional specs are finished and the correct links have been added to this issue.  I have added the `sprint must have` label to this story as the pattern is needed for the Drupal Services template.  If you would pick up this story as soon as you have time, that would be great.  Let me know if you have any questions.

cc:  @ljcarot

PR #2887, opened 2020-06-24
feat(LinkWithIcon): enabled inverse theming

RaphaelAmadeu:
### Related Ticket(s)

#2295 

### Description

The `inverse` prop was added to `LinkWithIcon`, so we can toggle the inverse theming.

### DIFF
--- a/packages/react/src/components/LinkWithIcon/LinkWithIcon.js
+++ b/packages/react/src/components/LinkWithIcon/LinkWithIcon.js
@@ -21,10 +21,9 @@ const { prefix } = settings;
 const LinkWithIcon = ({ children, href, inverse, ...props }) => {
   return (
     <div
-      className={classnames(
-        `${prefix}--link-with-icon__container`,
-        inverse ? `${prefix}--link-with-icon__container__inverse` : null
-      )}
+      className={classnames(`${prefix}--link-with-icon__container`, {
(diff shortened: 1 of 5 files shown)


packages/react/src/components/LinkWithIcon/LinkWithIcon.js
```diff
@@ -17,10 +18,13 @@ const { prefix } = settings;
...
  * LinkWithIcon component.
  */
-const LinkWithIcon = ({ children, href, ...props }) => {
+const LinkWithIcon = ({ children, href, inverse, ...props }) => {
   return (
     <div
-      className={`${prefix}--link-with-icon__container`}
+      className={classnames(
+        `${prefix}--link-with-icon__container`,
+        inverse ? `${prefix}--link-with-icon__container__inverse` : null
```
asudoh, line 26:
```suggestion
        { [`${prefix}--link-with-icon__container__inverse`]: inverse }
```

Just another way to achieve the same thing (I think the original code is good, too).

oliviaflory:
Looks good to me! 

Question for the group: Should we add something like "applied when link is on $inverse-01` background?" to the Prop Docs? Could be good to clarify when it should be used? Let me know what you think.

Screen Shot 2020-06-29 at 10 49 38 AM

Issue #2615, opened 2020-06-01
Pattern: Leadspace block: top and bottom spacing


ltitle:
## Leadspace block: top and bottom spacing

> Describe in detail the issue you're having.

Incorrect spacing at the top and bottom of the pattern

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

Leadspace-block (should be consistent with all content block patterns)

> What did you expect to happen? What happened instead? What would you like to
> see changed?

### TOP PADDING:

**As seen in Storybook:**
32px at top of pattern for all breakpoints

**Expected:**
64px at top of pattern for Large, XL and Max breakpoints

![image](https://user-images.githubusercontent.com/57198117/83428953-4b17e080-a401-11ea-9570-743928c6dc7f.png)

32px at top of pattern for Medium and Small breakpoints

![image](https://user-images.githubusercontent.com/57198117/83428961-5408b200-a401-11ea-941e-68c7be324a57.png)

### BOTTOM PADDING:

**As seen in Storybook:**
96px at bottom of pattern on Medium, Large, XL and Max breakpoints

![image](https://user-images.githubusercontent.com/57198117/83429481-17898600-a402-11ea-8641-dd41051a70b4.png)

**Expected:**
160px at bottom of pattern for Large, XL and Max breakpoints

![image](https://user-images.githubusercontent.com/57198117/83429770-8d8ded00-a402-11ea-86bd-c82d3baf69d8.png)

96px at bottom of pattern for Medium breakpoint

![image](https://user-images.githubusercontent.com/57198117/83429724-76e79600-a402-11ea-9e6d-27930bd02199.png)

> What browser are you working in?

Chrome

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

IBM.com Reboot

## Steps to reproduce the issue

https://ibmdotcom-react-experimental.mybluemix.net/?path=/story/patterns-blocks-leadspaceblock--default

PR #2892, opened 2020-06-24
fix(LeadspaceBlock): top & bottom spacing

ariellalgilmore:
### Related Ticket(s)

#2615 

### Description

fixed padding top and bottom to correct breakpoints

### Changelog

**New**

- {{new thing}}

**Changed**

- {{changed thing}}

**Removed**

- {{removed thing}}


packages/styles/scss/patterns/blocks/leadspace-block/_leadspace-block.scss
```diff
@@ -15,19 +15,11 @@
...
+      padding-top: $layout-03;
       @include carbon--type-style('expressive-heading-06', true);
 
-      @include carbon--breakpoint('lg') {
-        padding-top: $layout-05;
-      }
-
       @include carbon--breakpoint('md') {
-        padding-top: $layout-03;
-      }
```
kennylam, line 27: @ariellalgilmore Looks like this was correct, but the order needed to be changed (`md` first, then `lg`). The way you have it now, it's `$layout-05` for both md/lg breakpoints.

packages/styles/scss/patterns/blocks/leadspace-block/_leadspace-block.scss
```diff
@@ -47,10 +39,10 @@
       }
 
       @include carbon--breakpoint('sm') {
-        padding-bottom: $layout-01;
+        padding-bottom: $layout-06;
       }
       @include carbon--breakpoint('md') {
-        padding-bottom: $layout-06;
+        padding-bottom: $layout-07;
       }
```
kennylam, line 46: These breakpoints should be `md` and `lg`, respectively. There actually is no `sm`.

ariellalgilmore: Thank you @kennylam! I did not realize that they needed to be in order, but that makes a lot of sense now.

Issue #2879, opened 2020-06-24
React: Make modules tree-shakable


asudoh:
### The problem

Though we know that `import { namedImport } from 'package-name'` has negative effects on application bundle size and build time and we have migrated away from it in our internal code, it's easier to use for our adopters and still want to provide good support for this usage.

### The solution

* Add `/*#__PURE__*/` annotations (with wrapping contents in IIFE as necessary).
* Upgrade to latest Babel for [better React component optimization](https://www.infoq.com/news/2020/05/babel-7-10-react-treeshaking/).

jeffchew:
@RobertaJHahn @asudoh moving this one out.

RobertaJHahn:
@asudoh Is this work closed?  Looks like the linked PR is closed.

asudoh:
Thank you for asking @RobertaJHahn - Followed up in team Slack.

RobertaJHahn:
Decided to move to `icebox` in the March 22, 2021 engineering issues refinement meeting.

PR #2895, opened 2020-06-25
chore(react): transform code for optimized bundle

asudoh:
### Related Ticket(s)

Refs #2879.

### Description

This change wraps variable assignments to a React component variable into IIFE marked as no side effect. Doing so avoids those React component module contents brought into application bundle unless such React component is in use.

This change also upgrades Babel to `7.10` that does similar performance optimizations for `React.forwardRef()`, etc.

### Changelog

**New**

- A new Babel plugin that wraps variable assignments to a React component variable into IIFE marked as no side effect.

**Changed**

- Upgrade Babel to `7.10`.


asudoh:
Build integration test seems to require updates upon this change. Will investigate.

asudoh:
Closing for now, will reopen when I can focus more on this area.

Issue #2693, opened 2020-06-08
Fix the console errors coming from React package


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

When running unit tests, there are a number of console errors appearing in the output. These should be fixed.

## Steps to reproduce the issue

1. Run `yarn test:unit` under `/packages/react`

## Additional information

<img width="1021" alt="Screen Shot 2020-06-08 at 3 21 08 PM" src="https://user-images.githubusercontent.com/1641214/84071437-b29ed480-a99b-11ea-90ac-f383e6184cb1.png">

caiobozato:
@jeffchew @asudoh @RobertaJHahn unassigned myself because next two weeks I'll be working on Support pages with @guilhermelMoraes

PR #2898, opened 2020-06-25
chore(Dotcomshell): avoid side effect in story module

asudoh:
### Related Ticket(s)

Refs #2693.

### Description

This change makes React component/element instances defined as module export (variable)s for some stories to functions that return React component/element, and thus they are React (function) components.

Doing so avoids React prop type warnings shown as orphaned React component/element tree, shown even in stories that don't use that React component/element tree.

This change also fixes some imports in `<Dotcomshell>` story referring to build artifacts. To see our changes live in storybook, we need to refer to the source.

### Changelog

**Changed**

- Made React component/element instances defined as module export (variable)s for some stories to functions that return React component/element, and thus they are React (function) components.
- Fixes some imports in `<Dotcomshell>` story referring to build artifacts. To see our changes live in storybook, we need to refer to the source.

### DIFF
--- a/packages/react/src/__tests__/__snapshots__/storyshots.test.js.snap
+++ b/packages/react/src/__tests__/__snapshots__/storyshots.test.js.snap
@@ -2172,716 +2172,683 @@ exports[`Storyshots Components|Dotcom Shell Default 1`] = `
                 }
               }
             >
-              <TableOfContents
-                menuItems={null}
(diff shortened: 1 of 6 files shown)


jeffchew:
@asudoh jest snapshots look out of date, can you update?

asudoh:
@jeffchew Sorry for the stale snapshot - Updated.

Issue #2383, opened 2020-05-14
Component: Footer - iPhone Voice over - "Expand /Collapsed" panels label on footer links is announced twice.


Praveen-sr:
## Detailed description

Device - iPhone 

Issue :- 
For the show/hide panels, along with Label,  the word "Expand /Collapsed "  added, and this , along with the actual ARIA Expand/Collapse  announcement creates dual  announcement .

Expected : 
Since already aria expand/collapsed is used to convey the toggle behavior there is no need for and additional expand/collapsed announcement hence it would be better to remove it.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/fa27794a-7ca2-4808-8b14-526a307309ad)

guilhermelMoraes:
Okay, so looks like this problem can be resolved just by updating the `carbon-components-react` package. In their latest version, they get rid of the title "Expand/Collapse" prop and a placeholder that appears when we don't pass the `title` prop to the `AccordionItem` - the component causing this issue.
I've opened a PR that is resolving this issue, though, it's a workaround. My question is: should we update the Carbon dependencies or go ahead if this workaround? 

@jeffchew @kennylam @asudoh

asudoh:
While the workaround itself looks good; Given we have just done a release it sounds a good timing to do something like core dependency update, but specific to `carbon-components-react` wondering if we have some routine we do for this kind of update.

kennylam:
@asudoh I agree we should try to stay on top of Carbon versions.

jeffchew:
@kennylam @asudoh @RaphaelAmadeu let's get the carbon dependencies updated, and let's run percy in the PR update and make sure nothing's broken!

RobertaJHahn:
Closing this issue as the Carbon upgrade fixes it.

PR #2907, opened 2020-06-25
fix(FooterNavGroup): adding null iconDescription to avoid repeated title

guilhermelMoraes:
### Related Ticket(s)

Component: Footer - iPhone Voice over - "Expand /Collapsed" panels label on footer links is announced twice. #2383

### Description

Added falsy value to the `iconDescription` prop in the Carbon component `AccordionItem` to avoid the title from being set to "Expand/Collapse" for all items. Since this problem cannot be fixed in our end, this is just a workaround. I'll keep this PR as a draft until I create and submit a PR to the Carbon team in order to fix this problem.

### Changelog

**New**

- set `iconDescription` prop to null at `AccordionItem`;


annawen1:
@guilhermelMoraes should we convert this PR to draft (read in your description that this PR would be kept in draft)

guilhermelMoraes:
@annawen1 thanks for letting me know! 😅  I clicked in the wrong button, I think

guilhermelMoraes:
I'm closing this PR since it's fix is just a workaround. The Carbon dependencies update should be enough. Follow the PR for the update: #2940

Issue #2348, opened 2020-05-12
Pattern: CalloutQuote - Mobile - Link arrow icon is smaller in size, not as per specs.


Praveen-sr:
## Detailed description
Device - iPhone 11 , Samsung S10

Issue:- When user give long text for the link, the arrow link gets smaller in size.

## Steps to reproduce the issue

1. Open the Storybook on mobile
2. Navigate to Callout Quote pattern
3. On knobs, edit/change the link text to large text
4. Verify the link arrow icon

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/a074e046-dab7-4bc8-ac55-66c1a4e94c16)

caiobozato:
@Praveen-sr Could you please test again and see if this issue still happen?

For me another issue is happening:

![mobile-issue.jpeg](https://images.zenhubusercontent.com/5eed1fec732800bd837801c2/5e7de533-ace6-43ff-a1c5-c45494e4facf)

I'm working to fix this above, btw.

Praveen-sr:
@caiobozato Re tested issue you mentioned, its because of no spacing in between the words. This very long word without spacing is not a valid scenario.

Issue :- When user gives long text (with spacing) for the link, the text actually wraps to next line and the arrow link gets smaller in size.

PR #2926, opened 2020-06-26
fix(CalloutQuote): fix icon size on mobile

caiobozato:
### Related Ticket(s)

Pattern: CalloutQuote - Mobile - Link arrow icon is smaller in size, not as per specs. #2348

### Description

The arrow icon was having it's size reduced when the link copy increases on `CalloutQuote`

I used CSS to make the Link Copy fit the space, respecting Icon's size (width + margin-left).

I'm going to open a Draft PR because I noticed this icon size reduction issue happens with all `LinkWithIcon` components, but each of them has different `children`, some have `span` and some don't. Any thoughts?

### Changelog

**New**

- Added a `<span>` element for CTA copy on Quote component file (JSX)
- Added a  `max-width: calc(100% - 1.75rem)` to make span with CTA copy fit the correct space


packages/styles/scss/components/quote/_quote.scss
```diff
@@ -16,6 +16,14 @@
       @include carbon--make-row;
     }
 
+    &__footer {
+      .#{$prefix}--link-with-icon {
+        span {
+          max-width: calc(100% - 1.75rem);
```
asudoh, line 22: Seems that the parent container is a flex box, so just a thought. The way as-is here is fine, too.

```suggestion
          flex: 1;
```

caiobozato: Thanks, Akira! I tested and worked well, will commit your suggestion.

asudoh: My apologies, my suggestion broke things... 🙇

Issue #2810, opened 2020-06-17
Create web components package in the monorepo


jeffchew:
#### User Story

> As a `[user role below]`:
IBM.com Library adopter

> I need to:
have an available IBM.com Library package using web components

> so that I can:
utilize the web components version of the IBM.com Library components in my application

#### Additional information

 - As discussed, we'd like to begin the creation of the IBM.com Library web components package in our monorepo
 - Should publish to package name `@carbon/ibmdotcom-web-components`
 - We probably want to start out with a pre-minor release version `0.0.0`?
 - devops flow needed for deploy previews as well as the test environment (https://ibmdotcom-webcomponents-canary.mybluemix.net)
 - @asudoh please include any additional details here

#### Acceptance criteria
 - [ ] `web-components` package added to the monorepo
 - [ ] devops flow setup for deploy previews based on label `package: web components`
 - [ ] devops flow setup for test environment upon updates to `master`

jeffchew:
@asudoh based on our discussion, I have made updates to the description of this story to reflect the name change. I also created a new epic for this work.

PR #2929, opened 2020-06-26
chore(dependencies): update internal dependencies

asudoh:
### Related Ticket(s)

Refs #2810.

### Description

This change removes `flatpickr` dependency as our library doesn't use Carbon date picker

### Changelog

**Changed**

- Removes `flatpickr` dependency as our library doesn't use Carbon date picker


packages/web-components/package.json
```diff
@@ -154,11 +158,8 @@
     "webpack": "^4.28.0"
   },
   "peerDependencies": {
-    "@carbon/ibmdotcom-services": "^1.8.0-rc.0",
-    "@carbon/ibmdotcom-utilities": "^1.8.0-rc.0",
     "carbon-components": "~10.11.0",
```
jeffchew, line 161: @asudoh thank you for this!

I think we should also lock down carbon dependencies so we can manage for adopters.

asudoh: Makes sense @jeffchew - Merged in your change and updated.

PR #2930, opened 2020-06-26
fix(webcomponents): updating package to public

jeffchew:
### Related Ticket(s)

Ref #2932 

### Description

This is a fix to properly publish the package as a public
package in NPM.

In addition, this includes a fix for the clean task, as it is
failing the global `yarn reset` since `gulp clean` requires a
prebuilt package through gulp, otherwise will fail. This is
now using a standard `rimraf` like the other packages.

Finally, the dependencies have been moved around so that the
library and carbon dependencies are required.

### Changelog

**Changed**

- various package.json changes


jeffchew:
@asudoh I was actually getting this PR in place in parallel to you, but includes some additional fixes. I made the versions explicit for the dependencies, but I haven't touched `flatpickr` as we discussed.

Issue #1574, opened 2020-03-04
Pattern: Leadspace - Large text on button wraps to 2, 3 lines and makes only one of the buttons taller.


Praveen-sr:
## Detailed description

Issue :- Large text on button warps to 2, 3 lines and makes either of the primary/ secondary buttons look taller.

## Steps to reproduce the issue

1. Open the ibm.com storybook url on Mobile
2. Navigate to Leadspace with image
3. Go to knobs, edit/add large text for secondary button
4. Go back to canvas and verify the button view

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/92ffd797-0036-4871-adb4-1278e54af672)

caiobozato:
@Praveen-sr I was reading you request and checking the Storybook. But isn't this the correct behavior for the CTA when you have larger texts?

I think it would be incorrect if the text overlaps the CTA **container** when it's size increases.

@ltitle @photodow Could you please check this and see if I'm saying something wrong?

ltitle:
The expectation would be that if the text is long and needs to wrap that the button height would increase (as shown in the visual above). However, if one button height increases both should increase even for the one with the shorter text string. That way they will always be the same size visually.

PR #2714, opened 2020-06-10
fix(ButtonGroup):  `sameHeight` utility added

guilhermelMoraes:
### Related Ticket(s)

I don't know if there is an issue ticket to put here, but we talked about it in today's QA stand up meeting.

### Description

Adding the `sameHeight` utility to the `ButtonGroup` component.

### Changelog

**Changed**

- the `buttons` in the `ButtonGroup` will now have the same height, using as reference the wider one.

PR #2939, opened 2020-06-29
fix(ButtonGroup): added sameHeight feature - Leadspace issue related

caiobozato:
### Related Ticket(s)

Pattern: Leadspace - Large text on button wraps to 2, 3 lines and makes only one of the buttons taller. #1574

### Description

The Buttons should have the same height, even if a text of one button is larger than the other. The issue mentioned above is related to `Leadspace` pattern, but the fix was applied to `ButtonGroup` since it's used in the Leadspace pattern.
 
### Changelog

**New**

- Added `sameHeight` utility to ButtonGroup


packages/react/src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -8,6 +8,8 @@ import React, { useEffect, useLayoutEffect, useRef } from 'react';
 import Button from '../../internal/vendor/carbon-components-react/components/Button/Button';
 import ddsSettings from '@carbon/ibmdotcom-utilities/es/utilities/settings/settings';
 import PropTypes from 'prop-types';
+import root from 'window-or-global';
+import { sameHeight } from '@carbon/ibmdotcom-utilities';
```
annawen1, line 12: I believe we are now importing exports like so

`import  sameHeight  from '@carbon/ibmdotcom-utilities/es/utilities/sameHeight/sameHeight';`

to optimize bundle size

packages/react/src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -97,6 +99,26 @@ const ButtonGroup = ({ buttons, enableSizeByContent }) => {
...
+  }, []);
+
+  /**
+   * Set the buttons to have the same height based on the tallest one
+   */
+  const setSameHeight = () => {
+    root.requestAnimationFrame(() => {
+      const { current: containerNode } = groupRef;
+      if (containerNode) {
+        sameHeight(
```
jeffchew, line 115: @caiobozato @annawen1 wonder if it's worth adding logic to only do this when there are 2 or more buttons?

annawen1: hmm good point we probably want to skip over all that logic if we're only using one button

asudoh:
Nice try @caiobozato! Similar to #2714, wondering if we need to support dynamic change of the content (e.g. application sets a different button content in certain context or at some point):

leadspace-resize

packages/react/src/components/ButtonGroup/ButtonGroup.js
caiobozato: @annawen1 @jeffchew Conditional logic added in latest commit

caiobozato:
> Nice try @caiobozato! Similar to #2714, wondering if we need to support dynamic change of the content (e.g. application sets a different button content in certain context or at some point):
> 
> leadspace-resize

@asudoh Like we talked on Slack, I saw your comments about `ResizeObserver` in #2714 , I think we can bring up this for discussion with DDS engineers team and see if wouldn’t be better to update the sameHeight utility to use the `ResizeObserver`

jeffchew:
> @asudoh As we talked on Slack, I saw your comments about `ResizeObserver` in #2714 , I think we can bring up this for discussion with DDS engineers team and see if wouldn’t be better to update the sameHeight utility to use the `ResizeObserver`, this way we can standardize the API use.

Agreed, we should centralize the change in the utility so all components using it will take advantage of this.

Issue #2861, opened 2020-06-23
Web components: Masthead


asudoh:
# The problem

Need a masthead that we can created in a sophisticated way, and works with all JavaScript frameworks.

### The solution

Create a masthead with Shadow DOM and Custom Elements standards.

#### Additional Information
- Prod QA testing issue #(#3033)

### Acceptance Criteria
- [x] Creation of the `masthead` component in web-components
- [x] Include README for adopters
- [x] Feature parity with the React version
- [x] Include story variations
- [x] Include `data-autoid` attributes, same as the React masthead and include in README
- [x] Post a message in the Prod QA testing issue tagging Praveen and Chetan when ready to test

Issue #2916, opened 2020-06-26
Update Web Components main README to be adopter focused


jeffchew:
#### User Story

> As a `[user role below]`:
IBM.com Library adopter

> I need to:
understand how to include IBM.com Library Web Components in my project

> so that I can:
begin to use the IBM.com Library web components for my applicatiojn

#### Additional information

 - Right now the README appears to be more developer focused. We should probably move some of that to developer specific docs. We should shape the README and cater more to consumers of the package.

#### Acceptance criteria
 - [ ] Updated README on root of `web-components` package
 - [ ] Add reference to README in storybook
 - [ ] Move developer docs to perhaps a docs folder

PR #2960, opened 2020-06-30
feat(masthead): introduce web components

asudoh:
### Related Ticket(s)

Refs #2861.

### Description

This change introduces Web Components version of masthead (L0).

This change also introduces:

* `lit-html` version of the icons defined in `@carbon/ibmdotcom-styles`. To do this, needed to hoist `carbon-components`/`carbon-custom-elements` dependencies given `lit-html` version of the icons that is create on-the-fly in development time isn't tied to `packages/web-components`
* Several mix-ins for masthead style rules. This required given they are under deeply nested selectors that are hard to define custom elements selectors alongside
* `lint-staged` support for TypeScript files

### Changelog

**New**

- Web Components version of masthead (L0)
- `lit-html` version of the icons defined in `@carbon/ibmdotcom-styles`
- `lint-staged` support for TypeScript files

**Changed**

- Moved several style rules of masthead to mix-ins


jeffchew:
@asudoh I was running into a similar problem on another branch, do you know what's causing this?

Screen Shot 2020-06-30 at 8 50 21 AM

asudoh:
@jeffchew Do you have the CircleCI log link? My finding is that it's likely specific to this PR (introduction of latest `fetch-mock`).

jeffchew:
> @jeffchew Do you have the CircleCI log link? My finding is that it's likely specific to this PR (introduction of latest `fetch-mock`).

This is the PR that's having issues. I actually can replicate locally as well, but haven't had a chance to isolate the issue.

https://github.com/carbon-design-system/ibm-dotcom-library/pull/2667

asudoh:
Thanks @jeffchew! Let's discuss further there.

jeffchew:
Looks great @asudoh!

Are there knobs that we can set for the story, like platform name, etc? It might also be worth creating as separate stories too and treat as a variation.

The masthead content below looks a little odd when you scroll down:

Screen Shot 2020-06-30 at 11 51 51 AM

jeffchew:
@asudoh noticed a couple of things with the masthead search. It is supposed to overlay the entire header when opened, but is showing as this:

Screen Shot 2020-06-30 at 11 55 02 AM

Also, when using the search typeahead, it seems to take over the whole storybook screen and can't find a way back:

masthead-search

asudoh:
Thanks @jeffchew for reviewing! Added knobs for brand name and authentication state, and fixed the issues.

jeffchew:
Thanks @asudoh ! Looks much better! I am still seeing the black banding when scrolling down though on the page, not sure if that's a storybook thing or the masthead causing it?

Also, I don't see the `data-autoid` attributes in the rendered markup.

asudoh:
Thanks @jeffchew!

> I am still seeing the black banding when scrolling down though on the page, not sure if that's a storybook thing or the masthead causing it?

One question here; How did you make the Storybook itself black? I think reproducing the problem requiring to do so. given what I'm seeing is:

image

asudoh:
Fixed the issue in operating system's dark mode (I guess newer Storybook supports [`@media (prefers-color-scheme: dark)`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme)). Thank you @kennylam and everyone for your help!

asudoh:
As discussed, added documentation for analytics on how tag names and custom events can be used in Web Components to play the role of `data-autoid`: https://github.com/carbon-design-system/ibm-dotcom-library/blob/26c948c30cb67c1cea9e8c8c56fb02045058787b/packages/web-components/README.md#web-components-way-for-analytics

Also added missing `dds-masthead-search-toggled` custom event.

jeffchew:
> As discussed, added documentation for analytics on how tag names and custom events can be used in Web Components to play the role of `data-autoid`: https://github.com/carbon-design-system/ibm-dotcom-library/blob/26c948c30cb67c1cea9e8c8c56fb02045058787b/packages/web-components/README.md#web-components-way-for-analytics
> 
> Also added missing `dds-masthead-search-toggled` custom event.

Thank you @asudoh ! FYI, I have a separate issue for updating the README to be more adopter focused (#2916), so we an address some of these thoughts there:
- Once we have the `dds-dotcomshell`, we should probably swap out the `dds-link-with-icon` with that. 
- I wonder if we should have a separate developer docs and link out the contents at the end to there?
- I'm wondering if we should have a quick explanation of what `lit-html` is and why we're using it, and secondary to that why it's a peer dependency vs regular dependency.
- Browser support: The url for building ibm.com experiences has the following browser support language (https://www.ibm.com/standards/web/browser-support), but it is also good to mention that IE isn't supported here.

packages/web-components/README.md
```diff
@@ -14,17 +14,96 @@ A Carbon Design System variant that's as easy to use as native HTML elements, wi
...
+
+View available web components at: http://ibmdotcom-web-components-canary.mybluemix.net/. You can see usage information in several ways:
+
+1. Going to Docs tab, where it shows the usage and available attributes, properties and custom events.
+2. Clicking the **KNOBS** tab at the bottom and changing values there. Most knobs are shown as something like `Button kind (kind)`, where `kind` is the attribute name
+3. Clicking the **ACTION LOGGER** tab at the bottom and interacting with the selected component. You may see something like `bx-modal-closed` which typically indicates that an event with such event type is fired.
+
+## Web Components way for analytics
+
+`@carbon/ibmdotcom-react`, etc. libraries have `data-autoid` attributes in several places, to convey better context than raw HTML `<div>`, etc. tags and their attributes when analytics libraries captures bubbling events e.g. `click` on `<body>` and inspects `event.target`.
```
jeffchew, line 76: At minimum, should we add the data-autoid for the parent level tags? e.g. `data-autoid="dds--masthead"`?

asudoh: We could do `<dds-masthead data-autoid="dds--masthead">`. That said, `<dds-masthead>` tag name seems to serve the purpose of `data-autoid`, but did you meant that we'd want to duplicate it...?

asudoh:
Thank you for another review @jeffchew! Good points:

> - Once we have the `dds-dotcomshell`, we should probably swap out the `dds-link-with-icon` with that. 

Makes sense.

> - I wonder if we should have a separate developer docs and link out the contents at the end to there?
> - I'm wondering if we should have a quick explanation of what `lit-html` is and why we're using it, and secondary to that why it's a peer dependency vs regular dependency.
> - Browser support: The url for building ibm.com experiences has the following browser support language (https://www.ibm.com/standards/web/browser-support), but it is also good to mention that IE isn't supported here.

All makes sense - Updated.

packages/web-components/README.md
jeffchew: I'm wondering if for analytics they are tracking anything across different applications within IBM.com, and if they need to have that consistent stable selector language when going from one part of the site to another? I might be overthinking it though.

asudoh: Yeah the thought overall makes sense. Given `data-autoid` inside shadow DOM doesn't make `event.target.dataset.audoId` work due to [event retargeting](https://developers.google.com/web/fundamentals/web-components/shadowdom#events) as discussed, I thought that providing a new set of guidelines makes more sense instead of providing `data-autoid` that makes it looks like "compatible" but not actually fully.

But if we want to go with "partial compatibility" route I'm all for it.

asudoh:
Thank you @jeffchew for reviewing! Let's queue up stable selector thing.

Issue #1161, opened 2020-01-23
Expressive theme – UI Shell


oliviaflory:
# WIP issue
This component is quite complex and we'd like to make as little updates as possible, it would be helpful to sit with the developer handling this component to review before making these updates.

### Header base
- looks good

### Header base w/ navigation
- [ ] navigation link text is 14px on desktop, is it not connected to a body-01 type token? possibly increase size.
- [x] update height of left side panel link container from 32 -> 40px tall
![Screen Shot 2020-01-23 at 10.46.19 AM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/d2249355-e536-449f-9ed4-1cb611e0481e)

### Header base w/ action
- [x] looks good

### Header base w/ navigation, actions and side nav
- [ ] side nav links don't seem to be connected to a type token? possibly increase size.
- possibly update icon sizes
- [x] update height of left side panel link container from 32 -> 40px tall
![Screen Shot 2020-01-23 at 12.51.12 PM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/60a85a45-a726-4490-bf06-06cbe32e6a73)

### Header base w/ side nav
- hopefully inherits any changes we make from previous example

### Header base w/ actions and switcher
- [x] update height of switcher panel link container from 32 -> 40px tall
![Screen Shot 2020-01-23 at 12.59.50 PM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/e00cb53d-472d-4efd-9e4c-d98e776e85b7)

### Fixed side nav
- hopefully inherits any changes we make from header base w/ actions and right panel

### Fixed side nav w/ icons
- hopefully inherits any changes we make from header base w/ actions and right panel

### Side nav rail
- hopefully inherits any changes we make from header base w/ actions and right panel

### Side nav rail w/ header
- hopefully inherits any changes we make from header base w/ actions and right panel
- link 4 in the header looks broken or :bug: 
![Jan-23-2020 13-08-59.gif](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/35a09d0b-3fc8-47fb-a37a-0073618f1231)

### Side nav w/ large side nav items
- ~this one is... crazy. need to review further before making changes~
- hopefully inherits container size changes from other examples.
- leave "large" heading as is.

annawen1:
### Header base w/ navigation
[...]

PR #2972, opened 2020-06-30
fix(expressive): ui-shell expressive container sizing

annawen1:
### Related Ticket(s)

Expressive theme – UI Shell #1161

### Description

Increase container sizes for the ui-shell

### Changelog

**Changed**

- set the `height` and `padding` for the ui-shell nav items


packages/styles/scss/themes/expressive/components/_ui-shell.scss
```diff
@@ -0,0 +1,40 @@
...
+/// Accordion styles (expressive)
+/// @access private
+/// @group accordion-expressive
+@mixin ui-shell-expressive {
+  $side-nav-link-padding: calc(
+    (#{$container-02} - #{map-get($body-long-01, 'line-height')}) / 2
+  );
+  $switcher-nav-link-padding: calc(
+    (#{$container-02} - #{map-get($productive-heading-01, 'line-height')}) / 2
+  );
```
asudoh, line 23: Seems that side nav item doesn't word wrap (I saw ellipsis instead). In such case, we can simply use flex box, etc. to make the content centered regardless of the container size. `padding-top: calc(...)` approach is an escape hatch for the cases where the height is variable (and thus centering does not yield to the right padding).

annawen1: yeah I initially had the padding set, but saw carbon sets the `height` so I had to override it. I can get rid of the `padding` since the height is set.

packages/styles/scss/themes/expressive/components/_ui-shell.scss
```diff
@@ -0,0 +1,43 @@
...
+/// Accordion styles (expressive)
+/// @access private
+/// @group accordion-expressive
+@mixin ui-shell-expressive {
+  //$side-nav-link-padding: calc(
+  //  (#{$container-02} - #{map-get($body-long-01, 'line-height')}) / 2
+  //);
+  //$switcher-nav-link-padding: calc(
+  //  (#{$container-02} - #{map-get($productive-heading-01, 'line-height')}) / 2
+  //);
```
asudoh, line 23:
```suggestion
```

I think these lines are meant to be removed? (If the flex box approach sure works)

jeffchew:
@annawen1 Can you check? Maybe one of the files might be missing the license text at the top?

Screen Shot 2020-07-02 at 10 25 44 AM

packages/styles/scss/themes/expressive/components/_ui-shell.scss
annawen1: yeah forgot to remove them 😅 thanks!

Issue #2910, opened 2020-06-26
Solutions Page : Lead space block & CTA section - on click, buttons are not operable


Praveen-sr:
## Detailed description

Keyboard and screen reader issue.

Issue :- 
1. With tab/Shift tab focus is not going to the mentioned links
2. Screen reader issue :- In arrow key navigation and swipe navigation ( left and right swipe) this control is available twice for the screen reader user.

Expected :-
1. Focus should go to each and every interactive element and it should be operable thru keyboard.
2. screen reader should identify the control only once.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/27b8a92c-9dd9-46fd-88f9-245d38d1fae7)

PR #2978, opened 2020-06-30
fix(ButtonGroup): removed row-reverse styling

ariellalgilmore:
### Related Ticket(s)

#2910 

### Description

`TabIndex` was going in reverse order, so removed `row-reverse` styling and now using `tabIndex` already defined in `Button`
### Changelog

**New**

- added to docs when Primary Button will be applied and adjusted copy in examples

**Changed**

- flex-direction to just `column`


ariellalgilmore:
@asudoh hahaha I talked with @kennylam and we decided to go with this fix because the order of the buttons should not matter just depends on the text the user adds. I added documentation of this in the markdown. Not sure if this fix will be possible without the order change?

asudoh:
@ariellalgilmore I see, thank you for the background and your thoughts. Another question I have is if we want to make sequential focus order logical (button1 then button2 in [the before-change UI](http://ibmdotcom-react-canary.mybluemix.net/?path=/story/components-buttongroup--default)), or visual (button2 then button1 in [the before-change UI](http://ibmdotcom-react-canary.mybluemix.net/?path=/story/components-buttongroup--default)). If it's the former, we can set `tabindex="0"` to all (visual) buttons. If it's the latter you are right that the logical order itself should be changed.

ariellalgilmore:
@asudoh i wondered that too! We discussed that the focus order should always be left to right as someone reads the page even if one button is the “primary” one

Issue #2416, opened 2020-05-15
Link with icon enhancements


oliviaflory:
### The problem
Users need the ability to choose if the icon appears on the right or the left of link with icon component.

### The solution
Add ability for user to set icon on the right or the left of text in link with icon component.

### Additional information
 - Any icon can be used 
 - By default, the icon should be on the right of the text but does not need to be labeled "default".
 - The design team would like to use hover color token for visited state by default like Carbon has [reference here](https://github.com/carbon-design-system/carbon/blob/master/packages/components/src/components/link/_link.scss), but also allow the user to enable the visited state if they wish [reference Allow visited styles](https://react.carbondesignsystem.com/?path=/story/link--default) in their storybook.
![Screen Shot 2020-05-15 at 4.45.59 PM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/2a0d195b-5f42-4dfa-a9e3-2333c0966eba)

**Specs**
Link to specs on [Box](https://ibm.ent.box.com/folder/107613134184)

![Primary_Link_Text-and-icon.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/32fe543d-159c-457c-bf27-bf4fba3e0173)

![Primary_Link_Text-and-icon-themes.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/7bff2d71-1417-4f65-b3c7-f0d41f6bb181)

### Acceptance criteria
- [ ] By default the icon should be on the right of the text (in Storybook)
- [ ] Visited state should use hover token (like Carbon)
- [ ] User can set icon on right or left of text

PR #2940, opened 2020-06-29
chore(Carbon): Carbon dependencies updated for all packages

guilhermelMoraes:
### Related Ticket(s)

Upgrade IBM.com Library to Carbon 10.14.0 #2909
Component: Footer - iPhone Voice over - "Expand /Collapsed" panels label on footer links is announced twice. #2383

### Description

Carbon dependencies updated for all the packages, including the most recent, `Web Components`. 

### Changelog

**Changed**

- Carbon dependencies across all packages.

PR #2980, opened 2020-06-30
feat(LinkWithIcon): add option for icon on left

kennylam:
### Related Ticket(s)

#2416 

### Description

Adds option to position the icon in `LinkWithIcon` to the left side.

### Changelog

**New**

- `LinkWithIcon` now supports icon positioning on the left side of link

**Changed**

- `TextCTA` now supports icon positioning on the left side of link
- `LinkList` now supports icon positioning on the left side of link for horizontal/vertical layouts


packages/styles/scss/components/link-with-icon/_link-with-icon.scss
```diff
@@ -47,15 +47,32 @@
       fill: currentColor;
     }
 
-    .#{$prefix}--link--disabled:visited {
-      // Carbon core's has ongoing discussion on disabled link support.
-      // Meanwhile, this CSS makes disabled and visited link makes the right color
-      color: $disabled-02;
+    &.#{$prefix}--link--disabled {
+      &:not([href]),
+      &:not([button]) {
```
asudoh, line 52: Question: Wondering what the change is here for. One guess I made is un-disabling link when `href` attribute is there, but `[button]` made me unsure about it. Any insights?

To put this topic in more context; Wondering if it affects CSS overrides scenario like what's described in: https://www.w3.org/TR/selectors-4/#zero-matches

packages/react/src/components/LinkWithIcon/LinkWithIcon.js
```diff
@@ -43,6 +48,11 @@ LinkWithIcon.propTypes = {
    */
   href: PropTypes.string,
 
+  /**
+   * Display icon to left of link.
+   */
+  iconOnLeft: PropTypes.bool,
```
asudoh, line 54:
```suggestion
  iconPlacement: PropTypes.oneOf(['left', 'right']),
```

suggestion(non-blocking): Carbon has had to make some breaking changes when we found a need for the 3rd option for where `bool` props are defined.

packages/styles/scss/components/link-with-icon/_link-with-icon.scss
kennylam: Thanks @asudoh, I agree. That's in the [current version of Carbon](https://github.com/carbon-design-system/carbon/blob/fd79e780e5c4c49a9a0720490d31cbfa641e8b6e/packages/components/src/components/link/_link.scss#L48-L53) we're using. I planned on removing once #2940 goes through.

asudoh: Makes sense, thank you very much for clarifying @kennylam! (My apologies that I didn’t see it came from Carbon)

packages/react/src/components/LinkWithIcon/LinkWithIcon.js
kennylam: Good call, thanks @asudoh!

PR #2990, opened 2020-07-02
chore(dotcomshell): fix prop type warnings in story

asudoh:
### Related Ticket(s)

Refs #2693.

### Description

This change fixes prop type warnings in `<DotcomShell>` story.

### Changelog

**Changed**

- Fixes for prop type warnings in `<DotcomShell>` story.


packages/react/src/components/ContentItemHorizontal/ContentItemHorizontal.js
```diff
@@ -91,13 +91,22 @@ ContentItemHorizontal.propTypes = {
...
-      type: PropTypes.oneOf(['local', 'external']).isRequired,
-      href: PropTypes.string.isRequired,
-      copy: PropTypes.string.isRequired,
-    })
-  ).isRequired,
+  cta: PropTypes.shape({
+    heading: PropTypes.string,
+    items: PropTypes.arrayOf(
+      PropTypes.shape({
+        style: PropTypes.oneOf(['text', 'card', 'button', 'feature']),
```
kennylam, line 98: Thanks @asudoh. I think at the moment the CTA in this component only allows for `text` style. [Func spec](https://ibm.ent.box.com/file/663532447262)
```suggestion
        style: PropTypes.oneOf(['text']),
```

packages/react/src/components/LinkList/LinkList.js
```diff
@@ -58,7 +58,7 @@ LinkList.propTypes = {
   /**
    * Describes heading of LinkList.
    */
-  heading: PropTypes.string.isRequired,
+  heading: PropTypes.string,
```
kennylam, line 61: According to the [func specs](https://ibm.ent.box.com/file/664463554207), `heading` is required.

```suggestion
  heading: PropTypes.string.isRequired,
```

packages/react/src/patterns/blocks/ContentGroupHorizontal/ContentGroupHorizontal.js
```diff
@@ -52,13 +52,22 @@ ContentGroupHorizontal.propTypes = {
...
-          type: PropTypes.oneOf(['local', 'external']).isRequired,
-          href: PropTypes.string.isRequired,
-          copy: PropTypes.string.isRequired,
-        })
-      ).isRequired,
+      cta: PropTypes.shape({
+        heading: PropTypes.string,
+        items: PropTypes.arrayOf(
+          PropTypes.shape({
+            style: PropTypes.oneOf(['text', 'card', 'button', 'feature']),
```
kennylam, line 59: @asudoh I believe only `text` and `card` are currently supported.

```suggestion
            style: PropTypes.oneOf(['text', 'card']),
```

packages/react/src/components/LinkList/LinkList.js
asudoh: Good point - Do you think we should remove the [existence check for `heading`](https://github.com/carbon-design-system/ibm-dotcom-library/blob/v1.8.0/packages/react/src/components/LinkList/LinkList.js#L31)?

packages/react/src/patterns/blocks/ContentGroupHorizontal/ContentGroupHorizontal.js
asudoh: Great catch - Found that [`style` there is set from another prop](https://github.com/carbon-design-system/ibm-dotcom-library/blob/v1.8.0/packages/react/src/components/LinkList/LinkList.js#L42), so removed this prop from the shape.

asudoh:
Thanks @kennylam for reviewing!

packages/react/src/components/LinkList/LinkList.js
kennylam: Yes please, good catch @asudoh thanks!

asudoh: Cool, thanks @kennylam for clarifying! - Updated.

PR #2995, opened 2020-07-02
fix(LocaleModal): region card focus state

kennylam:
### Description

Add missing border to focus state

image

Also updates various `tabIndex` values for keyboard navigation.

### Changelog

**New**

- focus state for region `CardLink`

**Changed**

- `tabIndex` values for keyboard navigation


packages/react/src/components/LocaleModal/LocaleModalCountries.js
```diff
@@ -140,6 +141,7 @@ const LocaleModalCountries = ({
             region.countries.map((country, index) => (
               <a
                 key={index}
+                tabIndex="0"
```
asudoh, line 144: question(non-blocking): Does this mean we should make this `<a>` sequential-focusable even if `country.href` is missing? Or does this have another intent?

kennylam: Thanks @asudoh, this isn't needed. Good call 👍

PR #3000, opened 2020-07-02
fix(leadspace-block): fixes leadspace block spacing

andysherman2121:
### Related Ticket(s)

https://github.com/carbon-design-system/ibm-dotcom-library/issues/2511

### Description

Fixes incorrect padding between the CTA and the horizontal bottom border


packages/styles/scss/patterns/blocks/leadspace-block/_leadspace-block.scss
```diff
@@ -45,7 +45,7 @@
         padding-bottom: $layout-06;
       }
       @include carbon--breakpoint('lg') {
-        padding-bottom: $layout-07;
+        padding-bottom: $layout-05;
```
guilhermelMoraes, line 48: Thanks for the changes, @andysherman2121! But looks like something is not right. In the `md` breakpoint, following the [visual specs](https://ibm.ent.box.com/file/662154081174), we should use `96px`, as it shows in the document: 
Captura de Tela 2020-07-02 às 15 03 11

In the breakpoints above (`lg`, `xlg` and `max`), it is `64px`, but visually this value is much bigger than the `96px` from above. Take a look at the proportion: 
Captura de Tela 2020-07-02 às 15 04 41

This red rule is showing the difference. I would say that visually the token replaced is much more accurate than the current. Should we ask for help from any of the designers?

ltitle:
@guilhermelMoraes Sorry! The specs seem to have the wrong value. On Lg, XL and Max break points it should say 160px (instead of 64px). Medium is correct at 96px

andysherman2121:
> @guilhermelMoraes Sorry! The specs seem to have the wrong value. On Lg, XL and Max break points it should say 160px (instead of 64px). Medium is correct at 96px

Awesome, will adjust. thanks @ltitle @guilhermelMoraes

andysherman2121:
The original token used was correct, specs were wrong. Closing PR

Issue #2718, opened 2020-06-10
CodeSandbox: Upgrade to Carbon 10.13


asudoh:
### The problem

Earlier version of `@carbon/icons-react` library has an issue where importing one icon, even with a specific path, bringing in all icons into build and build takes very long time. It caused build integration tests to take forever and clogging up our CI.

### The solution

Upgrade `@carbon/icons-react` in our CodeSandbox examples.

PR #3001, opened 2020-07-02
chore(examples): update dependencies

asudoh:
### Related Ticket(s)

Refs #2718.

### Description

This change upgrades `@carbon/icons-react` to boost up the build time in CodeSandbox examples.

This change also does:

* Use `node.js` CodeSandbox template for now, given it provides us with better stability.
* Run parcel without using HMR given it intermittently causes CodeSandbox to capture HMR port instead of web server port
* Upgrade other dependencies like `react`.
* Unified how package name and page title are defined
* Other misc fixes

### Changelog

**Changed**

* Upgrade `@carbon/icons-react` to boost up the build time in CodeSandbox examples.
* Use `node.js` CodeSandbox template for now, given it provides us with better stability.
* Run parcel without using HMR given it intermittently causes CodeSandbox to capture HMR port instead of web server port
* Upgrade other dependencies like `react`.
* Unified how package name and page title are defined
* Other misc fixes


packages/react/examples/codesandbox/components/ButtonGroup/package.json
```diff
@@ -4,20 +4,20 @@
...
   },
   "dependencies": {
     "@carbon/ibmdotcom-react": "latest",
     "@carbon/ibmdotcom-styles": "latest",
-    "@carbon/icons-react": "10.11.0",
-    "react": "16.9.0",
-    "react-dom": "16.9.0"
+    "@carbon/icons-react": "10.13.0",
+    "react": "16.13.0",
+    "react-dom": "16.13.0"
```
kennylam, line 15: @asudoh do you see any issues with us using a different version in our examples from our codebase?

asudoh: Good question @kennylam, I smoke-tested all examples and fixed issues I found (none of them seems to have been related to dependency update). The only remainder is #2999, where we need to wait for a new release.

kennylam: Great thanks!

Issue #3010, opened 2020-07-02
Suppress CalloutData from main Storybook


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

`CalloutData` should only appear under the https://ibmdotcom-react-experimental.mybluemix.net. It is appearing on canary as a blank canvas. We should suppress entirely if feature flags are not turned on.

![Screen Shot 2020-07-02 at 1.55.09 PM.png](https://images.zenhubusercontent.com/5cf80f82b494896be7bc635b/c46226a2-7ae7-4aa9-91f0-37ec99aa5394)

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

> Is this issue related to a specific component?

> What did you expect to happen? What happened instead? What would you like to
> see changed?

> What browser are you working in?

> What version of the IBM.com Library are you using?

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

- Screenshots or code
- Notes

PR #3013, opened 2020-07-02
fix(CalloutData): hide it from Storybook

caiobozato:
### Related Ticket(s)

Suppress CalloutData from main Storybook #3010

### Description

Since it's a feature flag, should not appear in Storybook if the env variable `DDS_CALLOUT_DATA` is not set or it's set to `false`

### Changelog

**New**

- Added condition to hide from Storybook based on `DDS_CALLOUT_DATA` variable


caiobozato:
@asudoh Could you please merge this one?

Issue #3018, opened 2020-07-02
Find a way to automate merging of approved pull requests


jeffchew:
#### User Story

> As a `[user role below]`:
IBM.com Library approver

> I need to:
automatically merge pull requests after it has been approved and all checks have passed

> so that I can:
not have to constantly monitor pull requests when they are ready to merge

#### Additional information

 - Possible solution: https://github.com/pascalgn/automerge-action

#### Acceptance criteria
 - [ ] GH action that runs as soon as a label is added to an approved PR
 - [ ] Auto rebases and auto merges pull requests after label is added

jeffchew:
Going to icebox this, the current method won't allow this to work with forked PRs. Carbon team is currently investigating Kodiak, but my thought is that it won't work there.

There is a developer preview of the pull request API that will allow merging of PRs, but will wait for that to be production ready before looking at this.

jeffchew:
Brought this back in, ended up using the same Github app that Carbon team is using which is Kodiak. This solution appears to work.

PR #3020, opened 2020-07-02
chore(automerge): tweak to automerge configuration

jeffchew:
### Related Ticket(s)

Refs #3018 

### Description

The merge retry needed to be tweaked since the CI checks take longer than the GH action can wait with the default settings.

### Changelog

**Changed**

- `MERGE_RETRIES` and `MERGE_RETRY_SLEEP` updated

### DIFF
--- a/.github/workflows/automerge-pr.yml
+++ b/.github/workflows/automerge-pr.yml
@@ -1,89 +1,28 @@
-# "version" is the configuration version, currently "1".
-version: 1
-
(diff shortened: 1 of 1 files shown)


jeffchew:
FYI, interesting tidbit. I looked up the error this GH action was throwing `Resource not accessible by integration`. Turns out that it will not allow this action to carry out since it is modifying the workflow file itself, this is for security reasons.

Issue #2976, opened 2020-06-30
Pattern: Callout with data: Source text width


ltitle:
## Callout with data: Source text width

> Describe in detail the issue you're having.

The source text width should be 90% of 8 columns with a 640px max width. The text is currently expanding beyond that width and spanning all 12 columns.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

bug

> Is this issue related to a specific component?

Callout with data pattern

> What did you expect to happen? What happened instead? What would you like to
> see changed?

**As seen in storybook:**

![image](https://user-images.githubusercontent.com/57198117/86171816-59821680-baeb-11ea-8e92-11435d770f22.png)

![image](https://user-images.githubusercontent.com/57198117/86171857-6a328c80-baeb-11ea-86eb-b7c16071c03c.png)

**As expected from visual specs:**

![image](https://user-images.githubusercontent.com/57198117/86171917-89311e80-baeb-11ea-81a7-c660ef1704b1.png)

![image](https://user-images.githubusercontent.com/57198117/86171940-977f3a80-baeb-11ea-9dfe-002dafd549af.png)

> What browser are you working in?

Chrome

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

IBM Reboot

## Steps to reproduce the issue

https://ibmdotcom-react-experimental.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/2873/index.html?path=/story/patterns-blocks-calloutdata--default

PR #3025, opened 2020-07-06
style(CalloutData): fixed source text width

caiobozato:
### Related Ticket(s)

Pattern: Callout with data: Source text width #2976

### Description

image

The source text should fit 90% of 640px width, this change fix that issue.

### Changelog

**New**

- Added `max-width: 640px` and set `width: 90%` for the source text in `Callout Data` pattern


packages/styles/scss/patterns/blocks/callout-data/_callout-data.scss
```diff
@@ -28,6 +28,8 @@
         &__source {
           @include carbon--type-style('body-short-01', true);
 
+          width: 90%;
+          max-width: 640px;
```
ariellalgilmore, line 32: should `carbon--mini-units` be used here?

caiobozato: @ariellalgilmore change applied: https://github.com/carbon-design-system/ibm-dotcom-library/pull/3025/commits/b6596afa9f813c3dc944e79e82a261afa119fcf8

Issue #2983, opened 2020-07-01
LeadSpace: LeadSpace content and gradient is appearing over the masthead


annawen1:
## Detailed description

> Describe in detail the issue you're having.

The LeadSpace title and gradient is appearing over the masthead

<img width="1406" alt="Screen Shot 2020-07-01 at 1 29 42 PM" src="https://user-images.githubusercontent.com/54281166/86274919-72003880-bba0-11ea-85e5-61e0f92c88c3.png">

Noticing this when used in the nextJs test service page PR: https://github.com/carbon-design-system/ibm-dotcom-library-nextjs-test/pull/28

PR #3031, opened 2020-07-06
fix(LeadSpace): lower z-index of overlay to prevent clash with masthead

annawen1:
### Related Ticket(s)

LeadSpace: LeadSpace content and gradient is appearing over the masthead #2983

### Description

Set `z-index` value to 1 to prevent it from overlapping the masthead on scroll

### Changelog

**Changed**

- set z-index to 1 for overlay


annawen1:
> LGTM 👍 - In future it may be a good discussion to see if it's worth introducing something akin to [Carbon z-index table](https://github.com/carbon-design-system/carbon/blob/v10.15.0/packages/components/src/globals/scss/_layout.scss#L13-L22). Thanks @annawen1!

@asudoh yeah that would be something useful to have!

Issue #3007, opened 2020-07-02
Pattern: Develop Content block - headlines based on the Reboot designs


RobertaJHahn:
#### User Story

> As a `[user role below]`:
IBM.com Library developer

> I need to:
create the `Content block - headlines`

> so that I can:
provide the ibm.com adopter developers patterns they can use to build ibm.com web pages

#### Additional information

- **This pattern will be available behind a feature flag.  Reboot will be doing user testing to decide on final design.**
- Story within Storybook with corresponding knobs
- Utilize Carbon
- **See the Epic for the Design specs information**
- Design QA testing issue (#3008 )
- [Visual Designs & Design Specs Box folder](https://ibm.ent.box.com/folder/117051778931)

#### Engineering notes
- We can probably leverage the `ContentBlock` for the outer component, and also the `ContentItem` for the individual items. We should probably create as its own sub-component, but we probably don't need to export it for now.
- Use the `sameHeight` utility to make sure that the links match each other in heights
- Interesting horizontalrule logic, we should make sure that we can capture all of that correctly.

#### Acceptance criteria
- [x] Include README for the react pattern and corresponding styles, and how to enable via feature flag
- [x] Designs were provided by Reboot, include Lila as a Reviewer in the Pull Request
- [x] Provide a direct link to the deploy preview for the designer in the Pull Request description
- [x] The Storybook link is added to the Design QA issue for their testing
- [x] Canary storybook should not show this story, it should only appear in the experimental storybook

kennylam:
@RobertaJHahn @Praveen-sr @chetankan This is ready for testing https://github.com/carbon-design-system/ibm-dotcom-library/pull/3040.

[Deploy preview](https://ibmdotcom-react-experimental.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/3040/index.html?path=/story/patterns-blocks-contentblockheadlines--default) link.

PR #3040, opened 2020-07-07
feat(ContentBlockHeadlines): add ContentBlockHeadlines pattern

kennylam:
### Related Ticket(s)

#3007 

### Description

Adds new `ContentBlockHeadlines` pattern

### Changelog

**New**

- `ContentBlockHeadlines` pattern
- `ContentBlockHeadlines` styles, story, and documentation


packages/react/src/patterns/blocks/ContentBlockHeadlines/__stories__/ContentBlockHeadlines.stories.js
```diff
@@ -0,0 +1,59 @@
...
+ * Copyright IBM Corp. 2016, 2020
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { object, text } from '@storybook/addon-knobs';
+import ContentBlockHeadlines from '../ContentBlockHeadlines';
+import { DDS_CONTENTBLOCK_HEADLINES } from '../../../../internal/FeatureFlags';
+import items from './data/items.json';
```
jeffchew, line 11: @kennylam we should probably get out of the habit of json blocks for knobs, any way to break this down?

ltitle:
@kennylam 3 small changes:

1. The end of section horizontal rule should span the full 12 columns not just the 8 columns like the in section rules above

2. Top of section padding on Medium breakpoint should be 32px. Is currently showing 64px:

image

3. On small breakpoint the spacing at the top of each item should remain 16px (same as all other breakpoints). It is currently increasing to 24px:

image

packages/react/src/patterns/blocks/ContentBlockHeadlines/__stories__/ContentBlockHeadlines.stories.js
kennylam: @jeffchew I've been of the opinion for a while that we shouldn't include those objects as knobs at all. Especially with our plans to create Codesandbox examples for everything, it's not needed. I'd say we just remove them entirely.

jeffchew: As discussed, we should either have dynamic item knobs like http://ibmdotcom-react.mybluemix.net/?path=/story/components-buttongroup--default, or remove for now and we address later.

kennylam: @jeffchew Removing for now. Will revisit the story later and look into adding dynamic items that can have randomized content to test `sameHeight` and CTA or not.

PR #3041, opened 2020-07-07
chore(masthead): support data-autoid

asudoh:
### Description

This change adds `data-autoid` attribute to several components in masthead, to provide some compatiblity with `@carbon/ibmdotcom-react` counterparts.

Due to Shadow DOM's nature that avoids direct access to the inner-DOM of components for the sake of abstraction, `data-autoid` support does _not_ intent to provide full compatibility. Added documentation on what elements support `data-autoid` and alatenate approaches to achieve the purpose of `data-autoid`, which are, tag names of custom elements and custom events.

This change also adds support for `title` attribute for some nav components, which serves the same purpose of text node as child nodes.

### Changelog

**New**

- `data-autoid` to several components in masthead.
- `title` attribute support for several nav elements.
- Documentation on `data-autoid` and its alternative.


jeffchew:
@asudoh can this be added as a huddle topic? Would be good for folks building new web components to know how we set stable selectors on our end for the corresponding custom elements.

PR #3044, opened 2020-07-07
fix(quote): fix color of optional copy

asudoh:
### Description

This fix changes the color of optional copy text to be the same as other text in quote.

### Changelog

**Removed**

- Explicit `$inverse-01` usage of text color for optional copy text.


jeffchew:
Screen Shot 2020-07-07 at 12 30 21 PM

Percy spotted this for darker themes now for quote.

asudoh:
Good catch @jeffchew - Fixed.

Issue #2884, opened 2020-06-24
Add percy snapshot testing for the web components package


jeffchew:
#### User Story

> As a `[user role below]`:
IBM.com Library tester

> I need to:
quickly see visual regression results of the web components package

> so that I can:
spot visual changes impacted by code changes

#### Additional information

 - _{{user research}}_
 - _{{user insights}}_
 - _{{user metrics}}_

#### Acceptance criteria
 - [ ] _{{State acceptance criteria}}_
 - [ ] _{{State another}}_
 - [ ] _{{And another}}_

PR #3049, opened 2020-07-07
chore(percy): adding percy automated visual regression to web-components

jeffchew:
### Related Ticket(s)

Refs #2884 

### Description

This adds percy snapshot testing for web-components pull requests. The additional tasks have already been added to Jenkins.

### Changelog

**New**

- Percy automated visual regression testing for web components
- Removed redundant/out-of-date devDependencies


jeffchew:
> @jeffchew Looks like it's complaining about missing `@carbon/ibmdotcom-styles` dependency
> image

Got it, checking it out now...

packages/web-components/src/components/masthead/__stories__/masthead.stories.ts
```diff
@@ -96,7 +97,7 @@ export default {
     props: {
       'dds-masthead-container': {
         // Lets `<dds-masthead-container>` load the nav links if `CORS_PROXY` is set
-        navLinks: process.env.CORS_PROXY ? undefined : links,
+        navLinks: process.env.CORS_PROXY && !inPercy() ? undefined : links,
```
asudoh, line 100: Does this mean we don't want to use mock version of masthead links in Percy...?

jeffchew: we actually do want to use the mocked version in percy. I think with this it means that we are indeed using the mocked version, `CORS_PROXY` is defined, and `inPercy()` returns true.

asudoh: Got it, my bad... Thanks @jeffchew for clarifying!

PR #3053, opened 2020-07-07
feat(MastheadL1): component enabled for dynamic data

guilhermelMoraes:
### Description

Now when we turn the `MastheadL1` on, the navigation data disappears from the `MastheadTopNav` and goes only to the `MastheadL1`. This behavior is needed at the Products Page. All these changes takes place only at the `lg` and above breakpoints

### Changelog

**New**

- MastheadL1 now accept dynamic data;

**Changed**

- When the `MastheadL1` is on, the `MastheadTopNav` disappears.


packages/react/src/components/Masthead/Masthead.js
```diff
@@ -189,7 +189,7 @@ const Masthead = ({
               <IbmLogo />
 
               <div className={`${prefix}--header__search ${hasPlatform}`}>
-                {navigation && (
+                {navigation && !DDS_MASTHEAD_L1 && (
```
jeffchew, line 192: Something a bit more nuanced. Instead of checking if the feature flag exists, I think we should have a prop for `navigationL1` and see if that's defined.

packages/react/src/components/Masthead/Masthead.js
```diff
@@ -245,6 +245,7 @@ const Masthead = ({
                 title={title}
                 eyebrowText={eyebrowText}
                 eyebrowLink={eyebrowLink}
+                navigation={mastheadData}
```
jeffchew, line 248: Make this pass in L1 specific navigation data instead, maybe `navigationL1`?

guilhermelMoraes: Okay, that makes sense.

packages/react/src/components/Masthead/Masthead.js
guilhermelMoraes: And the data from `navigationL1` should be the same of the `navigation`?

packages/react/src/components/Masthead/Masthead.js
jeffchew, line 248: These are all L1 data. I'm wondering if we should consolidate this prop that goes into `Masthead` as simply `mastheadL1`, and in that the props are:

```
{
   title: 'L1 Title',
   eyebrowText: 'Eyebrow Text'
   eyebrowLink: 'https://www.ibm.com/something',
  navigation: { ... nav data ... }
}
```

Thoughts @kennylam @guilhermelMoraes ?

guilhermelMoraes: @jeffchew @kennylam  thinking about it once again, I wonder if `navigation L1` is right. There are more validations to come if we choose that way. I think a better - and easier - approach would be something like: 
`<Masthead enableL1 />`. Since the Masthead will have only navigation data in one place (whether `TopNav` or `L1`), there is no need to distinguish it between `navigation` and `navigationL1`. As long as we define in the docs that this will happens, I believe this is a better approach.

jeffchew: We wouldn't want the L0 navigation data in the L1 space though, if the custom data isn't set there is a chance thatthat could happen. Also, we might want to leave room in case the L1 data structure ends up being different from L0, and also if design comes back and decides to have L0 and L1 nav data co-exist.

Issue #3047, opened 2020-07-07
Vanilla Footer – incorrect placement of legal links


jacobottesen-dgc:
## Vanilla Footer – incorrect placement legal links

-->

## Detailed description

> The vanilla footer doesn't have the correct placement of the legal links.

Correct placement:
![Screen Shot 2020-07-07 at 11.38.23 AM.png](https://images.zenhubusercontent.com/5da5ed9833f74500019c499f/d316fb4d-3a15-456a-a1f0-50e4b6648df6)

Incorrect placement:
![Screen Shot 2020-07-07 at 11.38.16 AM.png](https://images.zenhubusercontent.com/5da5ed9833f74500019c499f/0925de72-863f-4492-bf6f-70add81a225d)

> Is this a feature request (new component, new icon), a bug, or a general
> issue? Bug

> Is this issue related to a specific component?
Footer

> What browser are you working in?
Safari

## Steps to reproduce the issue

1. View the vanilla storybook https://ibmdotcom-vanilla.mybluemix.net/?path=/story/components-footer--default
2. View the legal links

PR #3061, opened 2020-07-08
fix(footer): divide legal nav to chunks

asudoh:
### Related Ticket(s)

Refs #3047.

### Description

This change divides legal nav contents to chunks of columns, as our React variant does.

### Changelog

**Changed**

- A change divides legal nav contents to chunks of columns.


asudoh:
Good point @kennylam - Probably [`lodash.chunk()`](https://lodash.com/docs/4.17.15#chunk)?

kennylam:
> Good point @kennylam - Probably [`lodash.chunk()`](https://lodash.com/docs/4.17.15#chunk)?

@asudoh Yes exactly. I didn't use `chunk` in mine because I forgot we recently added lodash. So I just basically did the same with `reduce`. But now that we have it, that would work great.

Issue #2768, opened 2020-06-16
Component: Image With Caption :- On mobile device in landscape mode image is not visible.


chetankan:
Issue Description:- On mobile landscape mode 'Image with Caption' component image is not visible .

## Steps to reproduce the issue

1.Enter url in mobile browser.
2. Select 'ImageWithCaption' component.
3. Turn to landscape mode.
4. Click on image.

OS:- iOS 12, 13  Android 8,9
Browsers:- Chrome, Safari, UC Browser.

URL :- https://ibmdotcom-react-staging.mybluemix.net/?path=/story/components-imagewithcaption--default

![imagewithcaption in landscpae mode.jpg](https://images.zenhubusercontent.com/5e21a841639f517d456f7e43/76bedd02-443c-43dc-90e9-32c0bd22d811)

RobertaJHahn:
@caiobozato Hi Caio, This issue has been moved into the Done pipeline, but the linked PR #3116 is still open.  Is this work done and the issue can be closed?

Issue #2445, opened 2020-05-19
Component:Image with Caption:- Caption text is overlapping on image in landscape mode.(Mobile)


chetankan:
Issue Description:- On mobile landscape mode in image with caption component , caption text is overlapping on image.

## Steps to reproduce the issue

1.Enter url int he mobile browser.
2. Select 'Image with Caption'
3. Open it in new page.
4. Turn mobile in landscape mode and notice caption text .

OS:- iOS 12, 13, Android 8,9

URL:- https://ibmdotcom-react-staging.mybluemix.net/?path=/story/components-imagewithcaption--default![Image with caption in landscape mode on mobile.jpg](https://images.zenhubusercontent.com/5e21a841639f517d456f7e43/97e3dd02-7fa6-4ef1-9b51-e526cf3f62d8)

PR #3063, opened 2020-07-08
fix(ImageWithCaption): image is hidden in sm & md

caiobozato:
### Related Ticket(s)

Component: Image With Caption :- On mobile device in landscape mode image is not visible. #2768

### Description

The image on modal was being occulted in `sm` and `md` breakpoints. I noticed a code was applying an `absolute` position below `lg` breakpoint was causing the issue. I removed the code, tested it in all resolutions (Modal and outside the modal) and the component is working well.

Note that I wasn't sure why that code was added there, but it's removal doesn't break the component since the code was only set for `img` element.

### Changelog

**Removed**

- In `LightBoxMediaViewer` SCSS file, the code related to `img` element - `carbon--breakpoint-down('lg')` and it's content.


packages/styles/scss/components/lightbox-media-viewer/_lightbox-media-viewer.scss
```diff
@@ -61,13 +61,6 @@
         max-width: 100%;
         height: auto;
         max-height: 100%;
-
-        @include carbon--breakpoint-down('lg') {
-          position: absolute;
-          margin: auto;
-          top: 0;
-          bottom: 0;
-        }
```
annawen1, line 70: hey @caiobozato I remember removing this part too a while back for another bug and I ended up putting it back in because I _think_ (if I remember correctly) the image would otherwise overlay/cover the text content. I tested the deploy preview on browserstack and see that the image is covering the "this is a caption" heading.

Screen Shot 2020-07-08 at 11 54 23 AM

caiobozato: @annawen1 thanks for sharing that. I noticed this issue is happening even if we don't remove the part of code I mentioned, as we can see in this issue: #2445 Maybe both issues can be fixed in this pull request at once

caiobozato: @annawen1 I fixed and tested the issue we mentioned above. But since we already have the #2445 card assigned for it, can we proceed with this PR as is and I will apply the fix in a separated PR connected to card #2445?

annawen1: ah ok, yeah that makes sense to me!

Issue #2948, opened 2020-06-29
Cleanup of Storybook stories for CTA


jeffchew:
#### User Story

> As a `[user role below]`:
IBM.com Library tester

> I need to:
see variations of the CTA component in storybook

> so that I can:
quickly see what available variations there are for the CTA component

#### Additional information

 - Storybook URL: http://ibmdotcom-react.mybluemix.net/?path=/story/components-cta--default
 - Should break down the types into their own stories
 - Break down any knobs with JSON

#### Acceptance criteria
 - [ ] Multiple storybook variations for CTA component
 - [ ] No knobs with JSON
 - [ ] Add actual prop names in knob description in parenthesis

PR #3086, opened 2020-07-08
chore(CTA): cleanup of storybook stories for CTA

annawen1:
### Related Ticket(s)

Cleanup of Storybook stories for CTA #2948

### Description

Breakdown CTA story into multiple stories based on style

### Changelog

**Changed**

- have each style (ex: button, card, etc) have its own story


annawen1:
> Nice work @annawen1! It'd be nice you have a chance to add a11y test parameters like: https://github.com/carbon-design-system/ibm-dotcom-library/blob/522f803/packages/react/src/components/ButtonGroup/__stories__/ButtonGroup.stories.js#L55-L70

@asudoh gotcha, I've added in the default props set to the stories, let me know if I need to add anything else

PR #3091, opened 2020-07-09
fix(MastheadL1): increased max-width, eyebrow validation and new prop

guilhermelMoraes:
### Description

 - Added validation to `eyebrownLink` and `eyebrowText`. Now If both props don't exist it shows nothing;
 - Increased `max-width` from `MastheadL1`. Now the `MastheadL1` has a more accurate sizing at the `max` breakpoint;
 - Added a `titleLink` prop. This prop is optional.

Now the "Stock Charts" can be clickable
Captura de Tela 2020-07-09 às 10 27 48

### DIFF
--- a/packages/react/src/components/Masthead/Masthead.js
+++ b/packages/react/src/components/Masthead/Masthead.js
@@ -322,6 +322,10 @@ Masthead.propTypes = {
     title: PropTypes.string,
 
     /**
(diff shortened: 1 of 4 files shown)


jeffchew:
There looks like a minor visual bug when the user scrolls down and the L1 shrinks:

Screen Shot 2020-07-09 at 9 57 11 AM

Issue #2196, opened 2020-04-24
[Card] Add link text


shixiedesign:
A feature request for component [Card](https://ibmdotcom-react.mybluemix.net/?path=/story/patterns-sub-patterns-card--link-clickable): currently only an icon is allowed for bottom left corner of card. We'd like to allow link text for potential future case cases. 

@photodow suggests to **simply implement a styled text and icon instead** of using existing [link with icon](https://ibmdotcom-react.mybluemix.net/?path=/story/components-link-with-icon--default) component. This is because: 

1. the added link text should be 14px from bottom, 16px from left of card (see spec below)
2. keep icon 16x16px from the bottom left corner when there's no link text
3. link text should have NO underline on hover

### Spacing

![image](https://user-images.githubusercontent.com/15144993/80248588-398f2c00-863e-11ea-81c8-e71ce9e30653.png)

### Visual spec

Selected screenshots only. For full visual spec of **Card**, please see Abstract collection: https://share.goabstract.com/f757e306-109b-43d6-a044-7154f87f7a7f

![image](https://user-images.githubusercontent.com/15144993/82916541-283c8680-9f40-11ea-9654-0cef821d5bc8.png)
![image](https://user-images.githubusercontent.com/15144993/82916591-3be7ed00-9f40-11ea-9d0d-65d167bf2934.png)

sethibm:
@shixiedesign @photodow Thanks for sharing this.  It seems like all elements in any pattern should at least be made up of other formalized patterns or components, rather than simply be styled within a given component or pattern?  

We otherwise may wind up with various bits and pieces in given patterns, groups, etc that lose connection to a cohesive system.  I've been trying to map and identify a few of these kind of things the past two days.

One of the current considerations is to set the default padding on lower level components like "Link with icon" to just safe / neutral padding in its standalone state, and then that padding is removed and changed to zero when it sits within another pattern like Cards, which allows that parent pattern to set particular padding like 16px left and 14px below.  

If we did the above, then I think the Link with icon could be used in this situation?  

For removing underline in a particular usage of Link with icon, it seems this could be of universal interest for this component.  @photodow, could this be added as an override / knob within the component?

RobertaJHahn:
Shixie will investigate the current state and update the issue to clearly describe the requirement.

shixiedesign:
Hi @RobertaJHahn I've updated the issue's title and description: 

- this is a feature request to **add Link text as an option to cards**
[...]

PR #3097, opened 2020-07-09
feat(Card): add link text option to clickable card

annawen1:
### Related Ticket(s)

[Card] Add link text #2196

### Description

Entire card should be clickable. The PR removes the "static" type option that existed previously which made only the link within the card clickable - so the `type` prop is no longer necessary. Logic that differentiates the "static" and "link" type has been removed as well.

Option to add text to the Card cta has been added.

### Changelog

**New**

- add option for icon placement as video card renders the play icon on the left of the cta text (cta.iconPlacement)

**Changed**

- remove "link" type being passed into Card component from parent components

**Removed**

- `type` prop from the Card component


packages/react/src/components/Card/__stories__/Card.stories.js
```diff
@@ -35,11 +35,11 @@ export default {
         cta: object(
```
jeffchew, line 35: @annawen1 since under the hood, can this be broken out as separate knobs?

jeffchew:
See this happening in percy, was this intentional?

Screen Shot 2020-07-09 at 2 56 40 PM

annawen1:
> See this happening in percy, was this intentional?
> 
> Screen Shot 2020-07-09 at 2 56 40 PM

oh no, let me get that fixed

packages/react/src/components/Card/__stories__/Card.stories.js
annawen1: yep done

jeffchew:
@annawen1 unfortunately more components got hit with the latest change:
https://percy.io/IBM-Digital-Design/IBM.com-Library-React/builds/6082707?utm_campaign=IBM-Digital-Design&utm_content=IBM.com-Library-React&utm_source=github_status_public

jeffchew:
So close! It looks like LocaleModal was still affected:

Screen Shot 2020-07-10 at 5 39 20 PM

packages/styles/scss/components/card/index.scss
```diff
@@ -23,7 +23,7 @@
       height: 100%;
       display: flex;
       flex-direction: column;
-      padding: $carbon--spacing-05 $carbon--spacing-05 0;
+      padding: $carbon--spacing-05;
```
diegomarcuz: I think what Jeff just mentioned It is happening because you are setting padding to all sides

annawen1: That code `padding: $carbon--spacing-05;` is already in the codebase, I edited it earlier to match the some specs but I reverted it and am going to speak to designer first

PR #3111, opened 2020-07-10
docs(README): update getting started

asudoh:
### Related Ticket(s)

Refs #2916.

### Changelog

**Changed**

- Updated getting started doc in `packages/web-components/README.md`, with modal example and its CodeSandbox link.


packages/web-components/README.md
```diff
@@ -40,29 +40,54 @@ yarn add @carbon/ibmdotcom-web-components lit-html lit-element
 The first thing you need is **setting up a module bundler** to resolve ECMAScript `import`s. Once you set up a module bundler, you can start importing our component modules, like:
 
 ```javascript
-import '@carbon/ibmdotcom-web-components/es/components/link-with-icon/link-with-icon';
+import '@carbon/ibmdotcom-web-components/es/components/button/button';
```
jeffchew, line 43: I'm wondering if we should hold off on this change, as ideally something like dotcomshell would we what we point them to. And also to not be repetitive, perhaps link to that doc instead of duplicating here?

asudoh: Thank you for your thoughts @jeffchew, holding the change is fine to me. Wrt linking, I think having to follow the link to get the idea of the basic usage is cumbersome - Good to see thoughts on how to solve such problem.

asudoh:
Closing for now upon the discussion - Will reopen once we have progress here.

asudoh:
Oops there is one remaining line that should still be made. Pushed an update and reopening.

packages/web-components/README.md
jeffchew: Right, well maybe we could make it something that can be included instead of duplicated? Just want to minimize places we need to make any updates if needed.

asudoh: I see, let's think through more. Agreed that having to change multiple places for one set of change makes the maintenance harder. Thank you for your thoughts @jeffchew!

Issue #3113, opened 2020-07-10
Investigate build failures in Jenkins


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

Builds are failing in Jenkins for test environments:

<img width="1190" alt="Screen Shot 2020-07-10 at 8 45 06 AM" src="https://user-images.githubusercontent.com/1641214/87163773-dfddeb80-c295-11ea-915d-c38b6020ecc8.png">

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

> Is this issue related to a specific component?

> What did you expect to happen? What happened instead? What would you like to
> see changed?

> What browser are you working in?

> What version of the IBM.com Library are you using?

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

- Screenshots or code
- Notes

PR #3119, opened 2020-07-10
fix(styles): optimization fixes for gulp-sass in styles package

jeffchew:
### Related Ticket(s)

Refs #3113 

### Description

The `gulp-sass` task was throwing out of memory errors. This is a step to mitigate the build times for the overall css artifact. To our knowledge, no adopters are using the main entry point and built artifact (nor should they), so for now this is removing sourcemaps ~as well as gulp-clean-css~ from the build pipeline until this can get sorted out.

### Changelog

**Changed**

- Reduced complexity of gulp-sass runner (removed sourcemaps ~and cleanCss~)

### DIFF
--- a/packages/styles/gulp/tasks/sass.js
+++ b/packages/styles/gulp/tasks/sass.js
@@ -13,7 +13,6 @@ const gulp = require('gulp'),
   rename = require('gulp-rename'),
   sass = require('gulp-sass'),
   path = require('path');
-// sourcemaps = require('gulp-sourcemaps');
 
 /**
  * @name _sass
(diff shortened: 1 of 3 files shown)


packages/styles/gulp/tasks/sass.js
```diff
@@ -43,11 +41,7 @@ function _sass() {
     )
     .pipe(rename(global.config.distCss))
     .pipe(gulp.dest('dist'))
-    .pipe(
-      cleanCSS({
-        level: 2,
-      })
-    )
+    .pipe(cleanCSS())
```
asudoh, line 44: Do we have raw CSS code? Otherwise `outputStyle: compressed` in Sass compilation may further improve the build time.

packages/styles/gulp/tasks/sass.js
```diff
@@ -29,7 +28,6 @@ function _sass() {
   // prettier-ignore
   return gulp
     .src(global.config.scssEntry)
-    .pipe(sourcemaps.init())
```
asudoh, line 32: Is this intentional...?

PR #3128, opened 2020-07-13
feat(cta): introduce text CTA for Web Components

asudoh:
### Related Ticket(s)

Refs #3129.
Refs #3130.

### Description

Introduces Web Components version of text CTA.

The most notable difference from its React counterpart is how CTA links are handled, notably with the video CTA type. `CTAMixin` automatically sets `href` to `#` for video CTA type and fires `dds-cta-run-action` custom event when the link is clicked. The event is with the original `href` value, which is the video ID, in `event.detail`.

This allows a separate component like `<dds-cta-container>` to create a media viewer dialog, instead of having to put in all media viewer logic into text CTA UI as a dependency.

### Changelog

**New**

- CTA component, starting with text style, consisting of:
  - `<dds-text-cta>`, a CTA variant with text style.
  - `<dds-cta-container>`, an example application-level component to render CTA from data and handles video request.
- Lightbox media viewer component, starting with video player, consisting of:
  - `<dds-lightbox-media-viewer-body>`, a lightbox media viewer variant of `<dds-modal-body>`.
  - `<dds-lightbox-video-player>`, the video player UI in lightbox media viewer.
  - `<dds-lightbox-video-player-container>`, an example application-level component to load video player data and put it into `<dds-lightbox-video-player>`.


packages/web-components/src/components/cta/__stories__/README.stories.mdx
```diff
@@ -0,0 +1,86 @@
...
+import 'carbon-custom-elements/es/components/modal/modal-close-button';
+import VideoPlayerAPI from '@carbon/ibmdotcom-services/es/services/VideoPlayer/VideoPlayer';
+import '@carbon/ibmdotcom-web-components/es/components/cta/text-cta';
+import '@carbon/ibmdotcom-web-components/es/components/modal/modal';
+import '@carbon/ibmdotcom-web-components/es/components/lightbox-media-viewer/lightbox-media-viewer-body';
+import '@carbon/ibmdotcom-web-components/es/components/lightbox-media-viewer/lightbox-video-player';
+
+...
+
+document.addEventListener('dds-cta-run-action', async event => {
```
jeffchew, line 52: This looks like a lot of lift for adopters to integrate a video player. I'm still not too far into the Typescript training, but wonder if there's something we could do to make it a bit easier for them to hook up a video?

asudoh: Good question @jeffchew - In story implementation there is `<dds-cta-container>` as an example implementation, but having our adopters refer to `<dds-cta-container>` implementation will make things `lit-element` specific. My thought is around adding Angular/React/Vue examples sometime later (like what `carbon-custom-elements` does) so we can provide examples tailored for the JavaScript framework of their choice (that our adopters can just copy/paste as the start).

jeffchew: What I meant was what if we had something like:
```
<dds-text-cta videoId="abcd12345"></dds-text-cta>
```

This would then hook up the video for them as well as populate the text link with the video title?

asudoh: I'd like `<dds-text-cta>` focus on its UI. All-in-one text CTA can be done by `<dds-cta-container cta-style="text" .item={ copy: 'Copy text', href=''abcd12345', type='video' }>` (`.item` here means `item` property instead of `item` attribute)

asudoh: Given OOTB solution seems more desirable here, added an example with that approach: https://github.com/carbon-design-system/ibm-dotcom-library/pull/3128/commits/e71c373d17d4b7736e974316a422037e3ffd7312

jeffchew: Thanks! I think this is probably what the majority of adopters would end up doing. Perhaps have the `<dds-cta-container>` approach first, then if you want to go into more details on having adopters customize the behavior, you can have something like `### Customized Actions Example` and then go into `dds-cta-run-action`?

asudoh: I see, updated the doc as such: https://github.com/carbon-design-system/ibm-dotcom-library/pull/3128/commits/90ecd0bb37871983b3177339117724182dafede4

jeffchew: thank you! looks good!

RobertaJHahn:
@asudoh I have created the dev and prod QA issues Jeff asked for.  text - CTA, Lightbox media viewer and Lightbox video player.  They are assigned to you and added to the `Doing` pipeline.  Once the code is merged you can send a message to Praveen and Chetan to begin their testing.  Mark all of the AC's complete in the dev issues and move them to the `Done` pipeline.  Let me know if you have any questions.

asudoh:
Cool thanks @RobertaJHahn!

Issue #2319, opened 2020-05-06
Experimental Language selector does not display current language label


kyjak:
Trying out the new language selector, if you hit the X button to enter a langauge
name by keyboard, and do not pick a language, rather leave the component, so it looses focus,
the language selector stays empty. no language label displayed.
fix: language selector should reset the label if it looses focus when cleared

jeffchew:
Unassigning this from myself in case other devs want to pick this up.

PR #3131, opened 2020-07-13
fix(LanguageSelector): refresh value on focus lost

caiobozato:
### Related Ticket(s)

Experimental Language selector does not display current language label #2319

### Description

The language selector was not resetting the value after cleaning all the field by clicking on `X` and clicking outside of it. I've created the useEffect based on code found on `Masthead` component.

Also, I've created a state to store the last selected item because clicking on `X` makes the state value to be `null`:

image

### Changelog

**New**

- Created state to store the value of the last selected item
- Created useEffect method and event listeners for combobox in order to refresh the value when clicking outside of it


packages/react/src/components/Footer/LanguageSelector.js
```diff
@@ -31,10 +38,32 @@ const LanguageSelector = ({ items, initialSelectedItem, callback }) => {
   function _setSelectedItem(selectedItem) {
     setSelectedItem(selectedItem);
     callback(selectedItem);
+    if (selectedItem !== null) {
+      setLastSelectedItem(selectedItem);
+    }
+  }
+
+  function useClickOutside() {
```
jeffchew, line 46: Can you add JSDocs for this?

caiobozato: Added the JSDocs

Issue #2875, opened 2020-06-24
Learn & Solution Page :- TOC jump to menu is not sticky on Mobile


chetankan:
Issue Description :-  In 'Learn & Solution' page on mobile device when user selects any title from 'TOC--> Jump to' section user is navigated to the selected title but 'jump to' section is getting disappeared.

Scope:  Assume that L0 exists for now.

## Steps to reproduce the issue

1. Enter url in mobile browser.
2. Click on 'Jump To' section.
3. Select a title from drop down.
4. After navigating to selected title check if 'jump to' section is displayed.

OS;- iOS 12, 13  Android 8,9
Browsers:- Safari, Chrome, UC Browser.

URL:- https://ibmdotcom-nextjs-staging.mybluemix.net/solutions.html

URL:- https://ibmdotcom-nextjs-staging.mybluemix.net/solutions.html

RobertaJHahn:
@sethibm Seth will investigate and document the design decision for the spacing between the TOC and the Masthead.  The gap between TOC and MH is not defined.  Then we can assign to Dev.

sethibm:
- Starter note here: @RobertaJHahn, the issue, at least from how it's described, is not spacing, rather that menu is not sticky I believe? Maybe I missed where this changed during QA review meeting.  However, please see below video about a potential padding / spacing issue. 

- On stickiness, please see attached export from visual spec, showing TOC should be sticky and remain attached below Masthead, when user either scrolls down, or clicks / taps any of the dropdown items in the TOC.  It currently disappears and this is an error.  It should be persistent and stick below masthead.

![tablet-mobile_scroll-behavior](https://user-images.githubusercontent.com/58601040/85769167-bd6d9f00-b6e7-11ea-909b-397561ac812c.jpg)

- @wonilsuhibm @larahanlon2 One question to clarify here. This visual spec shows medium breakpoint with 16px left and right padding on TOC, in above image.  It goes to 100% width for small breakpoint.  Is this a bug that needs to be fixed, or was a decision made to make it 100% width on both small and medium breakpoint? 

- Another potential issue to log:  For max, xlg, lg breakpoints, visual spec says TOC should maintain 32px top padding when user scrolls down.  It does not do so in this example link, rather touches the masthead with 0px padding.  See this video:

[TOC-32px-padding.zip](https://github.com/carbon-design-system/ibm-dotcom-library/files/4832986/TOC-32px-padding.zip)

Viz spec guidance:
[...]

PR #3133, opened 2020-07-13
fix(masthead): fix to profile menu scroll away issue

ariellalgilmore:
### Related Ticket(s)

#2875 

### Description

Masthead was covering TOC, but Masthead should scroll away

### Changelog

**Changed**

-changed position styling to absolute, so scroll away happens


packages/styles/scss/components/masthead/_masthead.scss
```diff
@@ -127,7 +127,7 @@ $search-transition-timing: 95ms;
 @mixin masthead {
   .#{$prefix}--masthead,
   :host(#{$dds-prefix}-masthead) {
-    position: fixed;
+    position: absolute;
```
jeffchew, line 130: Just doublechecking this actually, we want the masthead to be fixed right?
cc: @kennylam

kennylam: @jeffchew It should be for the initial 25% of scrolling, then it would begin scrolling. If L1 is present, then that would then stick. I think for now this is more of a fix to allow the TOC to stick properly to the edge of the viewport, since at the moment it will go behind the masthead and not visible.

jeffchew: Got it, so I'll approve this then!

jeffchew: Do we have another issue to track updating the 25% scroll logic?
cc: @RobertaJHahn @kennylam @ariellalgilmore @larahanlon2

kennylam: > Do we have another issue to track updating the 25% scroll logic?
> cc: @RobertaJHahn @kennylam @ariellalgilmore @larahanlon2

@jeffchew I don't think so, I'll create one.

Issue #3087, opened 2020-07-08
1:1 aspect ratio Video Player engineering investigation and development


RobertaJHahn:
#### User Story

> As a `[user role below]`:
Business Analyst

> I need to:
understand the engineering implementation capabilities

> so that I can:
correctly document the functional specifications

#### Additional information

 - [Current video functional specifications](https://ibm.ent.box.com/folder/108837570522)
- Engineering will the technical work required to provide the 1:1 aspect ratio requested.  
- Engineering will determine whether func spec changes are required.  If required, Engineering will post a message tagging @photodow and @RobertaJHahn 
 - Eng will reach out to Jerry and Felipe to get sample test videos
- This analysis is required to move the development work forward.  Additional issues will be created based on the results of this review.

#### Acceptance criteria
 - [ ] The current functional specs have been reviewed
 - [ ] Technical implementation issues have been worked out with Felipe and Jerry
 - [ ] Message has been sent to James and Roberta if functional spec changes are needed.

RobertaJHahn:
@kennylam Did you finish developing this component?  I am wondering whether two of the tasks listed in the Additional Information section are finished.  

- Engineering will determine whether func spec changes are required.  If required, Engineering will post a message tagging @photodow and @RobertaJHahn 
 - Eng will reach out to Jerry and Felipe to get sample test videos

Are func specs required?  The next step would be for Chetan or Praveen to perform the QA testing.  They will need test videos to do this.   Do you have test videos that you can attach to this issue or even guidance in how they might test.  

Thanks.

kennylam:
@RobertaJHahn I don't believe functional specs are required other than updating the documentation to list the available aspect ratios: `16x9`, `9x16`, `2x1`, `1x2`, `4x3`, `3x4`, `1x1`

Issue #2834, opened 2020-06-18
1:1 Aspect Ratio Video Player for George Hammer 


quincymass:
### The problem
_{{  The standard for video production has evolved and produced a new format that enables video to better integrate with social solutions and new web standards.  In addition to existing standards, IBM is now creating video contend in 1:1 aspect ratio (square format) but the existing player is unable to present the content in the new dimensions.  It is critical to IBM image that the player is able to display/play 1:1 video as it does all other formats.  Submitted by Jerry Ringuette on behalf of Cory Calhoun/George Hammer with Q3 delivery requested.)}}_

### The solution
_{{IBM MediaCenter is able to transcode the video in the proper format, but the Carbon player will need to be updated to recognize the new aspect ratio.}}_

### Additional information
 _{{Here is a link to a page that demonstrates the new format and how the Northstar player reacts: https://pages.github.ibm.com/mediacenter/docs/aspect-ratio/1x1/}}_
 - Engineering to review the technical work to be done.  Determine whether func spec changes are required.  Tag @photodow and @RobertaJHahn if func spec changes are needed.
 - Eng will reach out to Jerry and Felipe to get sample test videos

quincymass:
Submitted - by Jerry Ringuette - jerryr@us.ibm.com on behalf of Cory Calhoun Cory.Calhoun@ibm.com

tweenn:
If I may add a helping hand, here's a little code example on how to get a video information on kaltura's player api, with mediacenter's public credentials/information and calculate the aspect-ratio without embedding the video.
```js

const PID = '1773841'; // Mediacenter Kaltura's Production Partner ID
const uiConfID = '27941801' // Northstar Video Player ID
const scriptSRC = `https://cdnapisec.kaltura.com/p/${PID}/sp/${PID}00/embedIframeJs/uiconf_id/${uiConfID}/partner_id/${PID}`;

let kAPI = {};

/**
 * @comment There is not much error handling on getting
 * kaltura's player api script as this is not the focus
 * of the example, so don't mind the resolve(false) and
 * no rejection statement
 */
const getMediacentersKalturaAPI = async () => {
	const getScript = await new Promise((resolve, reject) => {
		let scriptObject = document.createElement('script');
		const scriptParentNode = document.getElementsByTagName('script')[0];

		scriptObject.async = 1;

		scriptObject.onload = scriptObject.onreadystatechange = function( _, isAbort ) {
[...]

PR #3139, opened 2020-07-14
feat(VideoPlayer): add aspect ratio overrides

kennylam:
### Related Ticket(s)

#3087 

In regards to my concern about videos that are wrapped in other components such as `Card`, I think this is not as big a deal as I first thought. We can control this through the containing component itself and not allow the aspect ratio to be set for those videos.

### Description

This change adds several aspect ratios which can be used to override the default of 16:9.

### Changelog

**New**

- aspect ratios: `16x9`, `9x16`, `2x1`, `1x2`, `4x3`, `3x4`, `1x1`


jeffchew:
@kennylam what was the whole deal about felipe's code and aspect ratios (https://github.com/carbon-design-system/ibm-dotcom-library/issues/2834)? It looks like the source itself will be coming back in a different width/height, so we might not be setting the aspect ratio based on prop, but actually from the data coming from Kaltura API...

jeffchew:
> @kennylam what was the whole deal about felipe's code and aspect ratios (#2834)? It looks like the source itself will be coming back in a different width/height, so we might not be setting the aspect ratio based on prop, but actually from the data coming from Kaltura API...

After speaking on this, let's confirm that this will resolve the ask from media center team.

jeffchew:
Sounds like we're good with this implementation after speaking with media center, thank you @kennylam for following up! @photodow do we need a functional spec update? This is a new field that allows adopters to set the aspect ratio.

Issue #2987, opened 2020-07-01
Web Component: CardLink


jeffchew:
# The problem

Need a CardLink that we can created in a sophisticated way, and works with all JavaScript frameworks.

### The solution

Create a CardLink with Shadow DOM and Custom Elements standards.

#### Additional Information
- Prod QA testing issue (#3036)

### Acceptance Criteria
- [ ] Creation of the `CardLink` component in web-components
- [ ] Include README for adopters
- [ ] Feature parity with the React version
- [ ] Include story variations
- [ ] Include `data-autoid` attributes, same as the React CardLink and include in README
- [ ] Post a message in the Prod QA testing issue tagging Praveen and Chetan when ready to test

PR #3143, opened 2020-07-14
feat(CardLink): adding CardLink to web components

RaphaelAmadeu:
### Related Ticket(s)

#2987 
### Description

Initial version of `CardLink` web component, still working on styling, feedback is appreciated as this is the first time i'm working on this,

I've took the code for `Card` from Akira's private repo


packages/web-components/src/components/card/card.ts
```diff
@@ -0,0 +1,133 @@
...
+      .assignedNodes()
+      .some(node => node.nodeType !== Node.TEXT_NODE || node!.textContent!.trim());
+    this[slotExistencePropertyNames[name] || '_hasCopy'] = hasContent;
+    this.requestUpdate();
+  }
+
+  /**
+   * @returns The inner content.
+   */
+  // eslint-disable-next-line class-methods-use-this
```
asudoh, line 65: This method uses `this`, and thus below ESLint directive is not needed (my apologies if my prototype had that line).

```suggestion
```

packages/web-components/src/components/card/card.scss
```diff
@@ -0,0 +1,88 @@
...
+// This source code is licensed under the Apache-2.0 license found in the
+// LICENSE file in the root directory of this source tree.
+//
+
+@import '@carbon/ibmdotcom-styles/scss/components/card/index';
+@import '@carbon/ibmdotcom-styles/scss/components/feature-card/feature-card';
+@import '@carbon/ibmdotcom-styles/scss/components/link-with-icon/link-with-icon';
+@import '../../globals/scss/vars';
+
+:host(#{$dds-prefix}-card) {
```
jeffchew, line 13: Just curious, should we be including :host / ::slotted in the styles package, and we keep the scss styling in the web-components package to a minimum?

jeffchew: @asudoh @RaphaelAmadeu this is more of a question, if this is something we should think about or it isn't possible? I still am not that far into the training just wanted to get an understanding.

packages/web-components/src/components/card/card.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  /**
+   * `true` if there is copy content.
+   */
+  private _hasCopy;
+
+  /**
+   * Handles `slotchange` event.
+   */
+  private _handleSlotChange({ target }: Event) {
```
jeffchew, line 53: In the spirit of the slack conversation we just had, wondering if we should consider moving any of the inner methods here out?

packages/web-components/src/components/card/card.scss
asudoh: @jeffchew It's in my personal roadmap. The blocker is how our styles (in `ibmdotcom-styles` package) are structured, that disallows us to reuse most of the style rules. One idea I have is limiting Sass nesting to pseudo elements only, and make stuffs work with less parent-child (ancestor/descendant) selectors.

packages/web-components/src/components/card/card.ts
asudoh: @jeffchew Do you have more details, e.g. what the code will be like and the motivation to do so?

packages/web-components/src/components/card-link/__stories__/README.stories.mdx
```diff
@@ -0,0 +1,41 @@
...
+import '@carbon/ibmdotcom-web-components/es/components/card-link/card-link';
+import ArrowRight20 from 'carbon-custom-elements/es/icons/arrow--right/20';
+```
+
+##### HTML
+
+```html
+<dds-card-link href="https://example.com">
+  <p>Copy text</p>
+  ${ArrowRight20({ slot: 'footer' }}
```
asudoh, line 31: `${ArrowRight20({ slot: 'footer' })}` is `lit-html` specific, probably we should put raw `<svg slot="footer">`?

packages/web-components/src/components/card/card.ts
jeffchew: more about being able to test internal methods as isolated functions. So this means either making some methods non-private so that unit tests can do isolated tests around that method, or taking out the method from the class entirely and export/create unit tests around that.

packages/web-components/src/components/card/card.scss
jeffchew: @asudoh sounds like a bunch of rework, but certainly we should think about it in order for us to find more reusability between the react and web-components package...

packages/web-components/src/components/card/card.ts
asudoh: Given the `private` runs in trasnspiler level (rather than [browser run-time level like `#handleSlotChange`](https://github.com/tc39/proposal-private-fields/blob/master/OLD_README.md)), we can still access to the private field by `(cardInstance as any)._handleSlotChange`.

RaphaelAmadeu: @jeffchew as web components uses classes, so it's not that of a problem, in opposite to our react functional components

packages/web-components/src/components/card-link/__stories__/README.stories.mdx
RaphaelAmadeu: Sounds good

packages/web-components/src/components/card/card.ts
jeffchew: makes sense!

PR #3164, opened 2020-07-15
fix(CardGroup): allow CardGroup to set Card with cta copy

annawen1:
### Related Ticket(s)

### Description

CardSectionSimple needs to have Cards that have text in the cta. Allowing CardGroup to set Card passing in any cta props.

Screen Shot 2020-07-15 at 1 41 22 PM

### Changelog

**Changed**

- pass in cta props to Card from CardGroup


jeffchew:
@annawen1 looks like the snapshot file needs updating.

Issue #2984, opened 2020-07-01
Create e2e test harness for directed storybook component testing


jeffchew:
#### User Story

> As a `[user role below]`:
IBM.com Library tester

> I need to:
automate the e2e tests in the IBM.com Library repository

> so that I can:
reliably test the library as part of the CI/CD workflow

#### Additional information

 - The previous effort was captured here: https://github.com/carbon-design-system/ibm-dotcom-library/issues/2697. However, after seeing the difficulties of having selenium driver navigate storybook in a reliable way, the effort has been pivoted more to the way that karma accessibility testing has been introduced in https://github.com/carbon-design-system/ibm-dotcom-library/pull/2630. This method introduces a way to render a storybook component and passing props for modifications in order to properly expose variations.
 - QA team has created test scripts in python, which were in the middle of being converted to jasmine scripts in this PR: https://github.com/carbon-design-system/ibm-dotcom-library/pull/2667. We will need to evaluate these tests and if they translate to this new methodology, but the idea of having per-component e2e tests still holds true, and would like to see something like:

- `ComponentName`
  - `__stories__`
  - `__tests__`
    - `ComponentName.test.js` <-- functional unit tests via Jest
    - `ComponentName.e2e.js` <-- e2e test

- Would like to see if part of the e2e tests can also send percy snapshots to capture variations in automated visual regression tests

#### Acceptance criteria
 - [ ] Create ability to render per-component e2e tests that can modify storybook props
 - [ ] Create github action to run on push/pr

PR #3166, opened 2020-07-15
test(e2e): adding WebdriverIO e2e test structure

jeffchew:
### Related Ticket(s)

#2984

### Description

This adds the core WebDriverIO test harness, Github action, and first initial tests that were converted from the Python script versions. The strategy taken is to use direct urls for the full screen view of the canvas. This includes passing in query params if necessary, rather than adjusting knobs.

There are still remaining .py files, which need to be converted to e2e.js files and in WebdriverIO format. I think it will be ok to tackle this in separate PRs.

### Changelog

**New**

- WebdriverIO e2e testing architecture
- e2e tests for `CalloutQuote`, `ContentBlockMixed`, `CardSectionImages`, `CardSectionSimple`


jeffchew:
One thing I might need to change, the github actions are currently targeting the canary environment. I'll need to hook up the harness for this to run on the PR level and isolated deploy preview, which will probably need to happen on jenkins. This action can otherwise run on master whenever new changes happen, which can continue to happen on GH actions.

asudoh:
Looks a progress @jeffchew, the only feedback I have is striving to capture events where the UI becomes in the state we are testing against, instead of running wait calls that assumes the UI becomes such state in certain seconds. Such approach makes the test less flaky; Flakiness is the biggest thing that drives people away from E2E testing.

jeffchew:
> Looks a good progress @jeffchew, the only feedback I have is it's good if we strive to capture events where the UI becomes in the state we are testing against, instead of running wait calls that assumes the UI becomes such state in certain seconds. Such approach makes the test faster and less flaky; Slowness and flakiness is the biggest thing that drives people away from E2E testing.

Good call, any thoughts on modifying the tests to check for that? We could also possibly do this in another PR too, and just get the initial architecture in place first.

packages/react/src/patterns/blocks/CalloutQuote/__tests__/CalloutQuote.e2e.js
```diff
@@ -0,0 +1,113 @@
...
+ * @private
+ */
+const _path = '/iframe.html?id=patterns-blocks-calloutquote--default';
+
+describe('CalloutQuote', () => {
+  beforeAll(() => {
+    browser.url(_url + _path);
+    browser.setWindowSize(1200, 800);
+    browser.pause(1000);
+  });
```
kennylam, line 30: @jeffchew @asudoh Regarding testing UI, perhaps something like this? 

```suggestion
  beforeAll(() => {
    document.addEventListener("DOMContentLoaded", function(event) {
      browser.url(_url + _path);
      browser.setWindowSize(1200, 800);
      browser.pause(1000);
      done();
    });
  });
```

asudoh: @kennylam Exactly, the notion is tracking for events that lets us proceed to the next step. It'll depend on each wait calls seen in this PR. And agreed with @jeffchew that we can do it in separate PRs.

Issue #2883, opened 2020-06-24
Create the Web Component for the Locale Modal


jeffchew:
#### User Story

> As a `[user role below]`:
IBM.com Library adopter

> I need to:
have the locale modal available in the web components package

> so that I can:
Integrate the locale modal experience in my application

#### Additional information

 - Prod QA testing issue (#3039)

#### Acceptance Criteria
- [ ] Creation of the `locale modal` component in web-components
- [ ] Include README for adopters
- [ ] Feature parity with the React version
- [ ] Include story variations
- [ ] Include `data-autoid` attributes, same as the React LocaleModal and include in README
- [ ] Post a message in the Prod QA testing issue tagging Praveen and Chetan when ready to test

PR #3185, opened 2020-07-16
feat(locale-modal): introduction

asudoh:
### Related Ticket(s)

Refs #2883.

### Changelog

**New**

- `<dds-locale-modal>`, the shell for locale modal.
- `<dds-regions>`, the shell for the regions UI in locale modal.
- `<dds-region-item>`, the region card in locale modal.
- `<dds-locale-search>` the locale search UI in locale modal.
- `<dds-locale-item>`, the locale selection link in locale modal.


packages/web-components/src/components/locale-modal/__stories__/README.stories.mdx
```diff
@@ -0,0 +1,111 @@
...
+import '@carbon/ibmdotcom-web-components/es/components/locale-modal/locale-search';
+import '@carbon/ibmdotcom-web-components/es/components/locale-modal/locale-item';
+```
+
+##### HTML
+
+```html
+<dds-locale-modal lang-display="United States — English" open>
+  <dds-regions>
+    <dds-region-item name="Americas"></dds-region-item>
```
jeffchew, line 35: Would adopters have to manage this themselves? Right now this is all driven by the alternative links in the `<head/>`, since each page authored can vary on available locale/langauges. This is also existing logic from v18 so that it continues to be managed that way on the page level.

asudoh: @jeffchew No, the plan is creating a Redux store that allows our users (or our code) to auto-generate `<dds-region-item>` and `<dds-locale-item>`. The pre-redux version of the code can be seen at: https://github.com/carbon-design-system/ibm-dotcom-library/blob/b91fb76d76a126557ba9f26c93e7ade72d0a82c0/packages/web-components/src/components/locale-modal/locale-modal-container.ts

jeffchew: Ok, in this case we might need a `simplified` getting started documentation on what the user is actually implementing, and how the data is getting loaded in. For at least the React version, there is just the addition of `<LocaleModal />` into where it needs to go, and behind the scenes it makes the API call and gets hydrated accordingly. We want to make sure that adopting this (and web components in general) is as easy as possible for adopters. We can certainly have extended documentation on the inner workings of this and how it can be modified for usage, but more for `advanced` users.

asudoh: OK thank you for your insights, let me update the doc once I get the updated version of `<dds-locale-modal-container>` with #3197 incorporated.

asudoh:
As discussed in last week's Huddle, wired `<dds-locale-modal-container>` to a Redux store so we can reduce the code in the component. Updated the document as such.

packages/web-components/src/components/locale-modal/__stories__/locale-modal.stories.ts
```diff
@@ -0,0 +1,46 @@
...
+    ...readme.parameters,
+    knobs: {
+      'dds-locale-modal-container': ({ groupId }) => ({
+        langDisplay: textNullable('Display language (lang-display)', 'United States — English', groupId),
+      }),
+    },
+    props: {
+      'dds-locale-modal-container': {
+        // Lets `<dds-locale-modal-container>` load the nav links if `CORS_PROXY` is set
+        localeList: process.env.CORS_PROXY && !inPercy() ? undefined : localeData,
```
jeffchew, line 42: Just wondering if in practice, we should be setting both the localeList and lang-display for the user, and not have this be props? The data is typically driven by the LocaleAPI, and the display name is coming from the JSON data via `getLangDisplay` method in the services package: https://github.com/carbon-design-system/ibm-dotcom-library/blob/a3c7463c140a01fa5ef2066c91f257da5a0c7d59/packages/services/src/services/Locale/Locale.js#L259

asudoh: You are right @jeffchew, in the most typical use cases we set it from our services. The property is for mocking (like `inPercy()` case) and more manual usage.

jeffchew:
@asudoh just a thought, perhaps we can inject static data for percy so we can do the visual testing. I believe we do that for React.

Screen Shot 2020-07-29 at 5 33 58 PM

asudoh:
@jeffchew Thank you for your suggestion, I thought we are doing this already (assuming you are referring to something like [this](https://github.com/carbon-design-system/ibm-dotcom-library/blob/v1.9.0/packages/react/src/components/LocaleModal/__stories__/LocaleModal.stories.js#L26-L27)), but let's keep investigating.

PR #3204, opened 2020-07-20
test(integration): introduce UI integration test suite

asudoh:
### Description

Intrudocing UI integration tests to Web Components package.

UI integration tests are for testing UI visuals and interactions, that runs faster and more deterministic way than full E2E tests, suitable for pull request checks, etc. to catch the breakages quicker.

### Changelog

**New**

- UI integration tests for Web Components.
- [CSS Shadow Parts](https://developer.mozilla.org/en-US/docs/Web/CSS/::part) to several inner elements, in `<bx-masthead-search>`.


packages/web-components/tests/integration/ui/jest.config.js
```diff
@@ -0,0 +1,26 @@
...
+
+module.exports = {
+  globalSetup: '<rootDir>/tests/integration/ui/setup',
+  globalTeardown: '<rootDir>/tests/integration/ui/teardown',
+  setupFilesAfterEnv: ['expect-playwright', '<rootDir>/tests/integration/ui/setup-env'],
+  rootDir: '../../../',
+  roots: ['<rootDir>/src/components'],
+  testEnvironment: 'jest-playwright-preset',
+  testRunner: 'jest-circus/runner',
+  testRegex: '.*_steps\\.js$',
```
jeffchew, line 22: Looks great Akira! One question, for the suffix `_steps.js` would these technically be considered e2e, or maybe `integration`? Just not sure if `steps` is as clear. We are following `ComponentName.e2e.js` as the filename structure for React, perhaps we can follow something similar?

asudoh: Thank you @jeffchew for your feedback. While UI integration tests are for partial integration rather than full integration as in E2E, you are right that there are some similarities.

`_steps` wording is from Cucumber, so I can see it sounds odd. If we use `.e2e.js` Jest finds WebDriverIO tests that would fail the test suite. That said, any other naming ideas...?

jeffchew: Got it, maybe `.steps.js` instead of `_steps.js`? At least we are following convention with `*.test.js` and `*.e2e.js`?

asudoh: I see, makes sense - Thanks @jeffchew for the idea!

asudoh: Updated: https://github.com/carbon-design-system/ibm-dotcom-library/pull/3204/commits/996b5e7732d55f0403939c84fb8722a24ad94c5c

packages/web-components/package.json
```diff
@@ -39,8 +39,10 @@
     "lint:src": "yarn lint:scripts",
     "lint:scripts": "eslint --ext .js,.ts,.tsx .",
     "typecheck": "tsc --noEmit -p tsconfig.json",
-    "test": "yarn test:unit",
+    "test": "yarn test:unit && yarn test:integration",
     "test:unit": "gulp test:unit",
+    "test:integration": "yarn yarn test:integration:ui",
```
jeffchew, line 44: Is yarn supposed to be repeated?

asudoh: Good catch @jeffchew thanks - Updated: https://github.com/carbon-design-system/ibm-dotcom-library/pull/3204/commits/8ba83ea27058c4392c0d8437afef952aea2d5bad

Issue #3078, opened 2020-07-08
Cleanup of Storybook stories for LeadSpaceBlock


jeffchew:
#### User Story

> As a `[user role below]`:
IBM.com Library tester

> I need to:
see variations of the LeadSpaceBlock component in storybook

> so that I can:
quickly see what available variations there are for the LeadSpaceBlock component

#### Additional information

 - Should break down the types into their own stories
 - Break down any knobs with JSON

#### Acceptance criteria
 - [ ] Multiple storybook variations for LeadSpaceBlock component
 - [ ] No knobs with JSON
 - [ ] Add actual prop names in knob description in parenthesis

PR #3211, opened 2020-07-20
chore(LeadspaceBlock): update story

kennylam:
### Related Ticket(s)

#3078 

### Description

Separate component variations into individual stories

### Changelog

**Changed**

- add stories for different variations (image, video)


jeffchew:
@kennylam for videos, we might need to add a percy-only class to suppress videos to remove any false positives: https://github.com/carbon-design-system/ibm-dotcom-library/blob/2df0086c99e41db06b55fe9ed42f6e130b4c1973/packages/react/.storybook/preview-head.html#L145

asudoh:
@jeffchew I see, do you have more details? Is it worth a new issue...? Thanks!

jeffchew:
> @jeffchew I see, do you have more details? Is it worth a new issue...? Thanks!

Yes, this is creating regions that are essentially ignored in percy snapshot tests. For instance, waiting for a video player to load in which could cause some false positive snapshot tests. We are already doing that in some places in React. More documentation here:

https://docs.percy.io/docs/percy-specific-css#hiding-regions-with-percy-specific-css

We could create a separate issue for this for web components.

asudoh:
Thanks @jeffchew! Created: #3242

kennylam:
> @kennylam for videos, we might need to add a percy-only class to suppress videos to remove any false positives

@jeffchew I see, done!

PR #3222, opened 2020-07-20
chore(LocaleModal): Raised test coverage

RaphaelAmadeu:
### Related Ticket(s)

2077

### Description

I've created tests for `LocaleModal` so the component's test coverage fulfils our 80% target.
Screen Shot 2020-07-20 at 18 04 35


jeffchew:
> LGTM 👍 - Thanks @RaphaelAmadeu! Given this PR seems to involve change in code-under-test, it'll be good if we can have a deploy preview and image diff so make sure nothing is broken functionally.

Good call, and as discussed in standup we will have QA team do a deeper regression of the component as there is dynamic functionality that percy can't capture. We should try to get localemodal e2e tests in place for this, perhaps set up a priority component list for them to add tests to.

Issue #3161, opened 2020-07-15
Masthead:- Accessibility :- Close icon (issues related to hamburger menu once it is expanded)


chetankan:
Issue Description :- There is no label for the mentioned control.

Recommended Fix:- Give appropriate label.

URL : http://ibmdotcom-web-components-canary.mybluemix.net/?

Happening on :- Voiceover ,iPhone

ljcarot:
@chetankan Is this when you tested the Web Components code or on React code too?  Please clarify when you open bug reports so we know where to address.

chetankan:
@ljcarot : In web component, added url now.

RaphaelAmadeu:
As the `carbon-custom-elements`  were updated, their close button now  have a proper aria-label

PR #3230, opened 2020-07-21
fix(masthead): added title and aria-label to masthead menu button

RaphaelAmadeu:
### Related Ticket(s)

#3161 

### Description

I've added aria-label and title attributes to the masthead close button, so it matches our react package version aria attributes


asudoh:
Thank you for jumping in @RaphaelAmadeu! For the sake of flexibility, we need to make `<dds-masthead-menu-button>` work for this a11y problem stand-alone. That said, can we make a fix to `<dds-masthead-menu-button>` itself? We can just reuse [a fix in upstream `<bx-header-menu-button>`](https://github.com/carbon-design-system/carbon-custom-elements/pull/450), too.

packages/web-components/src/components/masthead/masthead-container.ts
```diff
@@ -611,6 +611,8 @@ _  */
         <dds-masthead-menu-button
           button-label-active="${ifNonNull(menuButtonLabelActive)}"
           button-label-inactive="${ifNonNull(menuButtonLabelInactive)}"
+          button-label="Open menu"
```
jeffchew, line 614: @RaphaelAmadeu hey Raph, we probably should have this as a field that eventually gets passed in from the navigation data (I think we have that in there, do you recall @kennylam ?
We shouldn't hard code the english label for this, maybe have those as default value if nothing is passed in.

RaphaelAmadeu:
Closing this as this was fixed at Carbon's upstream.
cc: @asudoh @jeffchew

PR #3233, opened 2020-07-21
Fix(ctasection): Bottom spacing

greeshmareddyg:
### Related Ticket(s)

https://github.com/carbon-design-system/ibm-dotcom-library/issues/3017

### Changelog

**Changed**

- cta section bottom spacing :
        Lg and above -160px
        Sm - 64px

### DIFF
--- a/packages/styles/scss/components/ctasection/_ctasection.scss
+++ b/packages/styles/scss/components/ctasection/_ctasection.scss
@@ -30,8 +30,13 @@
 @mixin cta-section--pattern {
   .#{$prefix}--cta-section {
     &__cta {
+      padding-bottom: $carbon--layout-05;
+
       @include carbon--breakpoint('md') {
         padding-bottom: $carbon--layout-06;
+      }
+      @include carbon--breakpoint('lg') {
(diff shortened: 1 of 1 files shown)


jeffchew:
> LGTM 👍 - Would you want to remove the change in the snapshot, as this seems style-only change? Thanks @greeshmareddyg!

Right, @greeshmareddyg you probably want to revert the snapshot file as the change in the scss has no impact on the unit test snapshot.

greeshmareddyg:
> > LGTM 👍 - Would you want to remove the change in the snapshot, as this seems style-only change? Thanks @greeshmareddyg!
> 
> Right, @greeshmareddyg you probably want to revert the snapshot file as the change in the scss has no impact on the unit test snapshot. Right now the ci-check is failing because of this.

I will revert it Jeff.

asudoh:
@greeshmareddyg Looks that there still is a snapshot change. Would you want to rebase your branch to `master` and then checkout the snapshot file from `master` and checking it into your branch? It'll ensure there is no snapshot change in your branch.

Issue #3027, opened 2020-07-06
Pattern(CalloutData): iframe - white space not displaying properly and text overlapping


caiobozato:
## Detailed description

> Describe in detail the issue you're having.

When you add some white space on `Data` field and opens the component in the iframe, the space is displayed as urlencoded text `%20` also, a large text overlaps the pattern container.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

Yes, `CalloutData` pattern

> What browser are you working in?

Google Chrome

## Steps to reproduce the issue

1. Step one
Go to https://ibmdotcom-react-experimental.mybluemix.net/
2. Step two
Select `CalloutData` pattern
3. Step three
Click on knobs, select `Data` and add a larger text with white spaces
4. Step four
Click to open into iframe

## Additional information

Outside the iframe:
<img width="1027" alt="callout-data-issue-1" src="https://user-images.githubusercontent.com/26546511/86616488-48725300-bf8c-11ea-95db-be1a0fb40b10.png">

Displaying the component into iframe:
<img width="1171" alt="callout-data-issue-2" src="https://user-images.githubusercontent.com/26546511/86616499-4c05da00-bf8c-11ea-8e0f-e8206228bd23.png">

asudoh:
Filed: storybookjs/storybook#11641.

PR #3241, opened 2020-07-21
fix(callout-data): fixed wrong story display on iframe mode

RaphaelAmadeu:
### Related Ticket(s)

#3027 

### Description

I've fixed a bug on `CalloutData` story in iframe mode, where spaces on `data` prop, would be turned into unicode strings.


packages/react/src/components/CalloutData/__stories__/CalloutData.stories.js
```diff
@@ -40,11 +40,24 @@ export default !DDS_CALLOUT_DATA
...
+      let data;
+      if (parameters.props.CalloutData.props.data.includes('%20')) {
+        data = parameters.props.CalloutData.props.data
+          .split('%20')
+          .map(elem => `${elem} `)
+          .join('');
+      } else {
+        data = parameters.props.CalloutData.props.data;
+      }
+      console.log();
```
jeffchew, line 52: Make sure to remove the `console.log()`

packages/react/src/components/CalloutData/__stories__/CalloutData.stories.js
```diff
@@ -40,11 +40,24 @@ export default !DDS_CALLOUT_DATA
 export const Default = !DDS_CALLOUT_DATA
   ? undefined
   : ({ parameters }) => {
+      let data;
+      if (parameters.props.CalloutData.props.data.includes('%20')) {
+        data = parameters.props.CalloutData.props.data
+          .split('%20')
+          .map(elem => `${elem} `)
+          .join('');
```
asudoh, line 48: Do we need to care only about space character, or do we need full `decodeURIComponent()`?

RaphaelAmadeu: This problem is actually being caused but the percentage mark on the `data`, that if used with a space after it generates a double percentage mark on the url, that causes the bug.

I think that the main problem is the space character, but `decodeURIComponent` seems like a good solution as well

packages/react/src/components/CalloutData/__stories__/CalloutData.stories.js
RaphaelAmadeu: Removed that in my latest commit, thanks!

Issue #2979, opened 2020-06-30
TOC: Persistent 32px top padding needed 


sethibm:
## Detailed description

For max, xlg, lg breakpoints, visual spec says TOC should maintain 32px top padding when user scrolls down.  It does not do so in this example link, rather touches the masthead with 0px padding.  

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

Yes, TOC.

> What did you expect to happen? What happened instead? What would you like to
> see changed?

See this video:

I expect it to follow visual specs and have 32px top padding between TOC and masthead, when scrolling downward, like this:

Viz spec guidance:
<img width="1064" alt="32pxtoppadding" src="https://user-images.githubusercontent.com/58601040/85771796-41c12180-b6ea-11ea-9051-3a19055dbf7a.png">

Instead, there is 0 padding, see this video:

[TOC-32px-padding.zip](https://github.com/carbon-design-system/ibm-dotcom-library/files/4832986/TOC-32px-padding.zip)

> What browser are you working in?

Chrome

## Steps to reproduce the issue

1. Go to TOC on this test page: https://ibmdotcom-nextjs-staging.mybluemix.net/solutions.html
2. Scroll down
3. View how the TOC behaves with top padding

sethibm:
@larahanlon2 @wonilsuhibm I separated out this bug, so I can move the original TOC bug issue forward to dev.  Please lmk if it this is a bug we want to fix.

CC @RobertaJHahn, if we can get this moving at same time as the original TOC bug, the dev can do them both at once and likely save time.

RaphaelAmadeu:
Currently `TableOfContents` do have this padding at the top, this bug is happening only on our Next Js page, as the padding, is smaller than the Nav Height.

PR #3255, opened 2020-07-22
fix(TableOfContents): Sticky offset not working

RaphaelAmadeu:
### Related Ticket(s)

#2979 

### Description

Fixed a bug where the sticky offset prop would not work, impacting our next js pages


jeffchew:
@RaphaelAmadeu can you resolve merge conflicts with the jest snapshot file? Kodiak won't accept this otherwise and will remove the label.

Issue #3158, opened 2020-07-15
Masthead :- Accessibility :- Search edit combo list box


chetankan:
Issue Description :- No label for this control. 

Recommended Fix:- Give label as "search".

Happening on :- Voiceover iPhone JAWS FF

URL : 
http://ibmdotcom-web-components-canary.mybluemix.net/?

PR #3256, opened 2020-07-22
fix(Masthead): missing aria-label on masthead search

RaphaelAmadeu:
### Related Ticket(s)

#3158 

### Description

Added aria-label for search input to be read by screen readers.


packages/web-components/src/components/masthead/masthead-search.ts
```diff
@@ -7,6 +7,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+import { ifDefined } from 'lit-html/directives/if-defined';
```
asudoh, line 10:
```suggestion
import ifNonNull from 'carbon-custom-elements/es/globals/directives/if-non-null';
```

packages/web-components/src/components/masthead/masthead-search.ts
```diff
@@ -179,6 +186,7 @@ class DDSMastheadSearch extends BXDropdown {
         autocomplete="off"
         aria-controls="result-list"
         aria-autocomplete="list"
+        aria-label="${ifDefined(searchLabel)}"
```
asudoh, line 189:
```suggestion
        aria-label="${ifNonNull(searchLabel)}"
```

asudoh:
Once the update is done, could you please [update the snapshot by `gulp test:unit --update-snapshot`](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/web-components/docs/developer.md#running-unit-test)? Thanks!

PR #3273, opened 2020-07-24
chore(masthead): separating L1 stories for Masthead

jeffchew:
### Related Ticket(s)

No related issue

### Description

For experimental, the stories for L1 was directly integrated.
This separates out into their own stories.

One other change that had to be made, storyshots gets confused when it runs
snapshots over sub stories that are behind a feature flag. The only way I could find
to fix this issue is to enable feature flags for snapshot test. This seems to address the 
issue.

### Changelog

**New**

- Separated Masthead L1 stories in `Masthead` and `DotcomShell`

**Changed**

- Enabled feature flags for Jest snapshot testing

### DIFF
--- a/packages/react/src/__tests__/storyshots.test.js
+++ b/packages/react/src/__tests__/storyshots.test.js
@@ -9,6 +9,8 @@ import initStoryshots from '@storybook/addon-storyshots';
 import { mount } from 'enzyme';
 import path from 'path';
 
+process.env = Object.assign(process.env, { DDS_FLAGS_ALL: true });
+
 /**
  * This will initialize storyshots for snapshot testing.
  * NOTE: If a molecule/organism requires deeper testing, a __tests__ folder must be created in the corresponding folder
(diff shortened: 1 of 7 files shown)


jeffchew:
Putting in draft for now. The jest snapshots aren't agreeing for some reason.

packages/react/src/__tests__/storyshots.test.js
```diff
@@ -9,6 +9,8 @@ import initStoryshots from '@storybook/addon-storyshots';
 import { mount } from 'enzyme';
 import path from 'path';
 
+process.env = Object.assign(process.env, { DDS_FLAGS_ALL: true });
```
asudoh, line 12: Is there other tests that relies on feature flags? Wondering about `--runInBand` scenario (where the process of all test cases are shared).

jeffchew:
Going to close this for now, can't seem to find a reliable way to separate this out along with feature flags, and there are other fish to fry for now.

Issue #3126, opened 2020-07-10
Enable the expressive theme in the IBM.com Library


jeffchew:
#### User Story

> As a `[user role below]`:
IBM.com Library user

> I need to:
enable the expressive theme for Carbon

> so that I can:
create my IBM.com application with carbon components in expressive moments

#### Additional information

 - Current separate environment with the theme enabled: http://ibmdotcom-react-expressive.mybluemix.net/
 - Should create any issues if necessary to fix style issues (utilize percy if necessary)
 - should think about what can be decommissioned after this is enabled

#### Acceptance criteria
 - [ ] Enable the expressive theme through either the footer (the legal nav at minimum is required)
 - [ ] Decommission CI/CD workflow and environments in IBM Cloud for the IBM.com Library React with expressive (this will become the default state going forward)
- [ ] Remove "temp expressive" classes (https://github.com/carbon-design-system/ibm-dotcom-library/tree/master/packages/styles/scss/temp-carbon-expressive), and delete the folder
- [ ] Update environment dashboard
- [ ] Any necessary README updates

PR #3299, opened 2020-07-28
chore(styles): remove temp expressive

asudoh:
### Related Ticket(s)

Refs #3126.

### Description

This change removes classic expressive theme proxy code that was used for several components.

### Changelog

**Changed**

- Replaced usage of classic expressive theme proxy code with the latest one.

**Removed**

- Class expressive theme proxy code.


jeffchew:
Looks good @asudoh ! I noticed in percy the button icon is moved over slightly, but according to https://carbon-expressive.mybluemix.net that appears to be the correct spacing. Might need a designer to take a closer look though? @oliviaflory

asudoh:
> Also in web components, link with icon appears to be productive.

PR #3330, opened 2020-07-29
fix(CTASection): add sameHeight to content items

kennylam:
### Description

`CTASection` content items should be set to the same height.

image

### Changelog

**Changed**

- set `sameHeight` for content items


jeffchew:
LGTM!

Issue #3160, opened 2020-07-15
Masthead :- Accessibility :- After expanding the Hamburger menu


chetankan:
Issue Description :- Once we  expand the hamburger menu  visually there is close button and the options but screen reader is reading content which is not visually available.
For example IBM logo link, user profile and search all of IBM button in left and tight swipe.

Recommended Fix:- Screen reader should not read visually hidden components.

Happening on :-  Voiceover ,iPhone

URL : 
http://ibmdotcom-web-components-canary.mybluemix.net/?

PR #3348, opened 2020-07-30
feat(left-nav): support focus wrapping

asudoh:
### Related Ticket(s)

Refs #3160.

### Description

This change implements focus-wrapping in left nav so that the keyboard focus won't go out of left nav while it's open.

It's done in the following way:

* When `<dds-left-nav>` detects that the focus goes out of the element, it focuses on `<dds-masthead-menu-button>`.
* When `<dds-masthead-menu-button>` detects that the focus goes out of the element, it fires `dds-masthead-menu-button-request-focus-wrap` event with the information of if the user gesture was back-tab or forward-tab. `<dds-left-nav>` listens to the event to set the keyboard focus on the correct nav item.

This change also ensures that non-expanded top/left nav items are non-keyboard-focusable, partly upgrading `carbon-custom-elements` to `v1.2.0`.

### Changelog

**New**

- Code in `<dds-left-nav>` to focus on `<dds-masthead-menu-button>` when the keyboard focus goes out of `<dds-left-nav>`.
- `dds-masthead-menu-button-request-focus-wrap` event on `<dds-masthead-menu-button>` that is fined when the focus was lost,

**Changed**

- Fix to sub-nav items in top/left navs in masthead so it's not keyboard focusable unless it's expanded. It includes upgrading `carbon-custom-elements` to `v1.2.0`.


asudoh:
Moved focus detection code to `@carbon/ibmdotcom-utilities` so it's reusable for React. The latest code (portion not related to such new focus detection code) requires carbon-design-system/carbon-custom-elements#458 to work. CC @kennylam

Issue #3157, opened 2020-07-15
Masthead:- Accessibility :- The options with in the   Hamburger menu 


chetankan:
Issue Description :- The options are not available immediately after the  triggering element.
visually it is appearing after the triggering element but for the screen reader users it is available after the user profile button.

Recommended Fix:- screen reader should identify the options immediately after the triggering element.

URL : 
http://ibmdotcom-web-components-canary.mybluemix.net/?

Happening on :- Voiceover , iPhone

RaphaelAmadeu:
This was fixed on this PR in `carbon-custom-components`
https://github.com/carbon-design-system/carbon-web-components/pull/460
Thanks!

PR #3391, opened 2020-07-31
fix(Masthead): focus goes to the first element on the left nav once open

RaphaelAmadeu:
### Related Ticket(s)

#3157 

### Description

This update makes the first item of the left-nav focused once it's expanded.

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,6 +80,7 @@
     "jest-junit": "^6.4.0",
     "lerna": "^3.20.2",
     "lint-staged": "^8.1.5",
+    "node-fetch": "^2.6.0",
     "prettier": "^1.17.0",
     "react": "16.9.0",
     "rimraf": "^2.6.3",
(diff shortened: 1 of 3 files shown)


asudoh:
Thank you for jumping in @RaphaelAmadeu! For this kind of case, could you try changing [upstream `<bx-side-nav>`](https://github.com/carbon-design-system/carbon-custom-elements/blob/v1.2.0/src/components/ui-shell/side-nav.ts) with the following?

1. Make a change at [toggle button's event](https://github.com/carbon-design-system/carbon-custom-elements/blob/v1.2.0/src/components/ui-shell/side-nav.ts#L83) instead of property change, so we make sure such focus change happens only on user gesture (the codebase ensures focus change happens only on user gesture, the only exception to that rule is modal).
2. Define a selector (as a static property) that selects focusable item in side nav (`bx-side-nav-menu,bx-side-nav-menu-item,bx-side-nav-link` for `<bx-side-nav>` and `<dds-left-nav>` should override it to `dds-left-nav-menu,dds-left-nav-menu-item,dds-left-nav-item`) and use it to determine the element to focus on when left nav gets open.

Please don't hesitate to ping me if you have any questions. Thanks!

Issue #3383, opened 2020-07-31
CalloutWithMedia: theming isn't being set properly 


diegomarcuz:
## Detailed description

> Describe in detail the issue you're having.

Theming isn't being applied correctly on the Solutions and Services page for CalloutWithMedia component

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

CalloutWithMedia

> What browser are you working in?

Google Chrome

> What version of the IBM.com Library are you using?

1.10.0-alpha.4668

## Additional information

<img width="1063" alt="Screen Shot 2020-07-31 at 12 42 18" src="https://user-images.githubusercontent.com/37422384/89052248-d04d4200-d32b-11ea-8125-69e73cb27c70.png">

PR #3392, opened 2020-07-31
fix(CalloutWithMedia): g90 theme rendering well

diegomarcuz:
### Related Ticket(s)

CalloutWithMedia: theming isn't being set properly #3383

### Description

As inverse theme was removed, this PR remove the inverse class from Content Block component and add color $text-04 to the heading and copy

### Changelog

**New**

- color: $text-04

**Changed**

- moved the content-block whitin the callout__content

**Removed**

- class inverse from Content Block component 

Screen Shot 2020-07-31 at 18 36 26


diegomarcuz:
Percy is requesting verifications, if anyone during the code review could check these ones!! Thanks

Issue #3295, opened 2020-07-28
Component :- LightBoxMediaViewer(Stage):- On closing modal audio still plays.


chetankan:
Issue Description:- On closing embedded video modal, still audio keeps playing.

## Steps to reproduce the issue

1. Enter url in the browser.
2. Select 'Lighbox media viewer' component.
3. Click on 'Embedded Video' 
4. Open it in new tab, once video starts playing click on 'close X' icon on modal.

OS:- Wins 10, Mac
Browsers:- Chrome, Safari, Firefox, Edge.

URL:- https://ibmdotcom-react-staging.mybluemix.net/?path=/story/components-lightboxmediaviewer--default

![Embedded video.jpg](https://images.zenhubusercontent.com/5e21a841639f517d456f7e43/f653424d-094d-4fd8-ae3f-7db6b52acc57)

PR #3427, opened 2020-08-04
fix(VideoPlayerAPI): stop video when modal closes

ariellalgilmore:
### Related Ticket(s)

#3295 

### Description

Add listener to check if modal closes and stop video from continuing to play

### Changelog

**New**

- add listener to modal-close when clicked


packages/services/src/services/VideoPlayer/VideoPlayer.js
```diff
@@ -201,6 +203,13 @@ class VideoPlayerAPI {
         readyCallback: function(playerId) {
           var kdp = document.getElementById(playerId);
 
+          var container = document.querySelector(
+            `.${prefix}--modal-close, ${prefix}-modal-close-button`
+          );
+          container.addEventListener('click', function() {
+            kdp.sendNotification('doStop');
+          });
+
```
asudoh, line 212: Instead of baking in modal close handler here (given the code here is a service API library), we can make this API return `kdp` so that the caller can control the video play. In this way, our React code and our Web Components code can handle modal close event by themselves. So we can start from something like below?:

```diff
diff --git a/packages/services/src/services/VideoPlayer/VideoPlayer.js b/packages/services/src/services/VideoPlayer/VideoPlayer.js
index 0f8b3581..036d3a74 100644
--- a/packages/services/src/services/VideoPlayer/VideoPlayer.js
+++ b/packages/services/src/services/VideoPlayer/VideoPlayer.js
@@ -184,6 +184,7 @@ class VideoPlayerAPI {
   static async embedVideo(videoId, targetId, autoPlay) {
     const fireEvent = this.fireEvent;
     return await this.checkScript().then(() => {
+      const promiseKWidget = new Promise(resolve => {
         root.kWidget.embed({
           targetId: targetId,
           wid: '_' + _partnerId,
@@ -210,9 +211,17 @@ class VideoPlayerAPI {
             kdp.addJsListener('playerPlayEnd', function() {
               fireEvent({ playerState: 3, kdp, videoId });
             });
+
+            resolve(kdp);
           },
         });
       });
+      return {
+        kWidget() {
+          return promiseKWidget;
+        },
+      };
+    });
   }
 
   /**
```

packages/react/src/components/ExpressiveModal/ExpressiveModal.js
```diff
@@ -50,6 +50,12 @@ const ExpressiveModal = ({
    */
   function closeModal() {
     if (onClose?.() !== false) {
+      if (window.kWidget) {
+        window.kWidget.addReadyCallback(function(playerId) {
+          var kdp = document.getElementById(playerId);
+          kdp.sendNotification('doStop');
+        });
+      }
```
asudoh, line 58: Any way we can do this in light box media viewer code instead of backing video player-specific feature in expressive modal?

ariellalgilmore: Good point! Sorry I must have misunderstood. Thanks!

ariellalgilmore:
> LGTM 👍 - Future thing to think about is if we should restart the video player when the modal is re-opened, and if so, what should happen if user repeats opening/closing in a short amount of time (should we do some protection against race condition?). Thanks @ariellalgilmore!

Thank you @asudoh!! Yeah interesting point, I was checking in `contentBlockCards`  with video and the video will restart when you open it again. Maybe design would have some input on this? Also, I think doing protection against user repeating opening/closing makes a lot of sense. Possibly something interesting to talk about in a huddle?

asudoh:
> Maybe design would have some input on this?

Yeah great point!

> Possibly something interesting to talk about in a huddle?

Definitely!

Issue #3396, opened 2020-08-03
1:1 Aspect Ratio video player testing support


RobertaJHahn:
#### User Story

> As a `[user role below]`:
DDS Prod QA tester

> I need to:
have the video test files in the correct aspect ratios available in Storybook

> so that I can:
properly perform the Prod QA testing for all of the aspect ratios included in the design

#### Additional information

 - Repoint the partner ID's from public Kaltura to MediaCenter's test environment
 - Swap video ID's depending on the video ID aspect ratio
 - Partner ID will be provided in an email
- 16:9 - Test Video -> https://mediacenter.ibm.com/media/1_mq9h9c34      ID: 1_mq9h9c34
- 4:3 - Test Video -> https://mediacenter.ibm.com/media/1_p2osmd1z     ID: 1_p2osmd1z
- 1:1 - Test Video -> https://mediacenter.ibm.com/media/1_9h94wo6b       ID: 1_9h94wo6b

#### Acceptance criteria
 - [x] Updated to include the various videos in Storybook

PR #3434, opened 2020-08-04
feat(VideoPlayer): adding stories for different aspect ratios 

guilhermelMoraes:
### Related Ticket(s)

1:1 Aspect Ratio video player testing support #3396

### Description

adding a story for each aspect ratio

### Changelog

**New**

- the story for 1x1 aspect ratio;
- the story for 4x3 aspect ratio.


packages/react/src/components/VideoPlayer/__stories__/VideoPlayer.stories.js
```diff
@@ -23,26 +20,23 @@ export default {
...
-        aspectRatio: select(
-          'Aspect ratio (aspectRatio):',
-          ratios,
-          ratios[0],
-          groupId
-        ),
+        aspectRatio: 'default',
+        videoId: '0_uka1msg4',
       }),
     },
```
asudoh, line 26: Given different story seems to have different set of parameters now, would you move this block to the `parameters` of `Default` story?

packages/react/src/components/VideoPlayer/__stories__/VideoPlayer.stories.js
```diff
@@ -51,3 +45,41 @@ export const Default = ({ parameters }) => {
...
 };
+
+export const aspectRatio1x1 = ({ parameters }) => {
+  return <Default parameters={parameters} />;
+};
+
+aspectRatio1x1.story = {
+  name: 'Aspect ratio 1:1',
+  parameters: {
+    knobs: {
```
asudoh, line 56: Given there is no parameter controlled by a knob here:

```suggestion
    props: {
```

packages/react/src/components/VideoPlayer/__stories__/VideoPlayer.stories.js
```diff
@@ -51,3 +45,41 @@ export const Default = ({ parameters }) => {
...
+};
+
+export const aspectRatio4x3 = ({ parameters }) => (
+  <Default parameters={parameters} />
+);
+
+aspectRatio4x3.story = {
+  name: 'Aspect ratio 4:3',
+  parameters: {
+    knobs: {
```
asudoh, line 75: Given there is no parameter controlled by a knob here:

```suggestion
    props: {
```

packages/react/src/components/VideoPlayer/__stories__/VideoPlayer.stories.js
guilhermelMoraes: I'm sorry, @asudoh, I didn't understand quite well your question. 😭

asudoh: No problem @guilhermelMoraes, I meant:

```diff
diff --git a/packages/react/src/components/VideoPlayer/__stories__/VideoPlayer.stories.js b/packages/react/src/components/VideoPlayer/__stories__/VideoPlayer.stories.js
index d7589859..50aa4dd2 100644
--- a/packages/react/src/components/VideoPlayer/__stories__/VideoPlayer.stories.js
+++ b/packages/react/src/components/VideoPlayer/__stories__/VideoPlayer.stories.js
@@ -17,13 +17,6 @@ export default {
     percy: {
       skip: true,
     },
-    knobs: {
-      VideoPlayer: ({ groupId }) => ({
-        showCaption: boolean('Show caption (showCaption):', true, groupId),
-        aspectRatio: 'default',
-        videoId: '0_uka1msg4',
-      }),
-    },
   },
 };
 
@@ -46,6 +39,18 @@ export const Default = ({ parameters }) => {
   );
 };
 
+Default.story = {
+  parameters: {
+    knobs: {
+      VideoPlayer: ({ groupId }) => ({
+        showCaption: boolean('Show caption (showCaption):', true, groupId),
+        aspectRatio: 'default',
+        videoId: '0_uka1msg4',
+      }),
+    },
+  }
+}
+
 export const aspectRatio1x1 = ({ parameters }) => {
   return <Default parameters={parameters} />;
 };
```

diegomarcuz: @guilhermelMoraes, I think Akira is saying that you need to do for Default story the same thing you did for aspectRatio1x1 and 4x3 story.

e.g.
~~~js
aspectRatio1x1.story = {
  name: 'Aspect ratio 1:1',
  parameters: {
    props: {
      VideoPlayer: () => {
        return {
          showCaption: true,
          aspectRatio: '1x1',
          videoId: '1_9h94wo6b',
        };
      },
    },
  },
};
~~~

cc: @asudoh

asudoh: Let me explain more on this; The parameters defined in the default export affects all stories. Given I think we don't want that, we can move it to the `Default` story. Note that the default export and the default story are different animals; The former is for metadata for all stories, the latter is for a story with "default" name.

diegomarcuz: Yeah, got it!! Thanks @asudoh

jeffchew:
I realize we will need to probably also time the change for the environment variables as well. Let me know if you want me to do that @guilhermelMoraes 

cc: @asudoh

packages/react/src/components/VideoPlayer/__stories__/VideoPlayer.stories.js
diegomarcuz: Isn't necessary to follow same ideia on props/knobs for Default story as aspectRatio1x1 and 4x3 ? @asudoh

guilhermelMoraes: thanks for the clarification, @asudoh, and @diegomarcuz. ;)

packages/react/src/components/VideoPlayer/__stories__/VideoPlayer.stories.js
```diff
@@ -5,16 +5,13 @@
...
 import VideoPlayer from '../VideoPlayer';
 
-const ratios = ['default', '4x3', '2x1', '1x1'];
-
 export default {
   title: 'Components|VideoPlayer',
-
   parameters: {
     ...readme.parameters,
     percy: {
```
guilhermelMoraes, line 17: @asudoh @jeffchew should we get rid of this?

guilhermelMoraes:
@asudoh @jeffchew should we get rid of this?

https://github.com/carbon-design-system/ibm-dotcom-library/blob/20238579ef00bfcb281806bdd9018720480b52b0/packages/react/src/components/VideoPlayer/__stories__/VideoPlayer.stories.js#L17

jeffchew:
> @asudoh @jeffchew should we get rid of this?
> 
> https://github.com/carbon-design-system/ibm-dotcom-library/blob/20238579ef00bfcb281806bdd9018720480b52b0/packages/react/src/components/VideoPlayer/__stories__/VideoPlayer.stories.js#L17

No, we are still skipping this in percy since the video player itself causes false positives in visual regression.

packages/react/src/components/VideoPlayer/__stories__/VideoPlayer.stories.js
```diff
@@ -51,3 +38,53 @@ export const Default = ({ parameters }) => {
...
   );
 };
+
+Default.story = {
+  parameters: {
+    knobs: {
+      VideoPlayer: ({ groupId }) => ({
+        showCaption: boolean('Show caption (showCaption):', true, groupId),
+        aspectRatio: 'default',
+        videoId: '0_uka1msg4',
```
jeffchew, line 48: @guilhermelMoraes for default we should also use a media center video and not the kaltura public video (folgers)

jeffchew:
@guilhermelMoraes fyi, I disabled the kaltura environment variables in the CI, so it should use the default IDs now.

guilhermelMoraes:
@jeffchew can you please confirm if these are the right ids? The 4x3 and 1x1 videos aren't playing anymore

16:9 - Test Video -> https://mediacenter.ibm.com/edit/1_mq9h9c34 ID: 1_mq9h9c34
4:3 - Test Video -> https://mediacenter.ibm.com/media/1_p2osmd1z ID: 1_p2osmd1z
1:1 - Test Video -> https://mediacenter.ibm.com/edit/1_9h94wo6b ID: 1_9h94wo6b

jeffchew:
> @jeffchew can you please confirm if these are the right ids? The 4x3 and 1x1 videos aren't playing anymore
> 
> 16:9 - Test Video -> https://mediacenter.ibm.com/edit/1_mq9h9c34 ID: 1_mq9h9c34
> 4:3 - Test Video -> https://mediacenter.ibm.com/media/1_p2osmd1z ID: 1_p2osmd1z
> 1:1 - Test Video -> https://mediacenter.ibm.com/edit/1_9h94wo6b ID: 1_9h94wo6b

Default appears to be working, are these the ones we received from media center? We should check with them if these aren't working or if they are tied to the correct kaltura partner/uiconf IDs...

cc: @kennylam @RobertaJHahn

jeffchew:
@guilhermelMoraes I'm noticing for the 4x3 and 1x1 stories, they are showing up as undefined, it's possible that those stories aren't picking up the prop values.

Issue #3432, opened 2020-08-04
Web Components: Integrate ProfileAPI into Masthead


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

After a closer look, it looks like the ProfileAPI is not getting called in order to determine the authenticated state of the user.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

Masthead

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Replace the `authenticated` property and have that completely driven by the ProfileAPI call.

Also need to doublecheck, but the profile menu items should be driven by the translation file.

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

- Screenshots or code
- Notes

jeffchew:
@asudoh would you be able to look at this one? If not can see if someone else can pick up.

asudoh:
@jeffchew Sure I can take a look.

PR #3463, opened 2020-08-06
feat(masthead): support profile API

asudoh:
### Related Ticket(s)

Refs #3432.

### Description

This change adds support for `ProfileAPI` to our Redux store, and connects it to `<dds-masthead-container>`, so the component can monitor up-to-date user authentication status.

### Changelog

**New**

- Redux action/reducer/state for `ProfileAPI`.

**Changed**

- `<dds-masthead-container>` to connect to the new Redux store feature.


packages/web-components/src/components/masthead/masthead-container.ts
```diff
@@ -152,17 +168,36 @@ const defaultUnauthenticateProfileItems: MastheadProfileItem[] = [
   },
 ];
 
+/**
+ * @param props A key/value pair.
+ * @returns The modified version of the given `props` with all properties with `undefined` value removed.
+ */
+function cleanProps(props: { [key: string]: unknown }) {
```
jeffchew, line 175: Should this eventually be a utility?

asudoh: It's a possibility. Wanted to see if this approach (clean afterwards) is the best, or there is a better way to prevent redundant props defined here.

PR #3464, opened 2020-08-06
feat(Masthead): add megamenu to mobile nav

kennylam:
### Related Ticket(s)

#3300 

### Description

This updates the mobile `Masthead` component to use full megamenu data from IBM.com

### Changelog

**New**

- megamenu data used in `Masthead` mobile nav


packages/react/src/components/Masthead/MastheadLeftNav.js
```diff
@@ -86,23 +87,75 @@ const MastheadLeftNav = ({ navigation, isSideNavExpanded, platform }) => {
...
+              key={i}>
+              {item.title}
+              <div
+                className={`${prefix}--masthead__side-nav--submemu-section-title__icon`}>
+                <ArrowRight20 />
+              </div>
+            </SideNavLink>
+
+            {renderNavItem(item.megapanelContent.quickLinks.links)}
+          </SideNavMenu>
```
asudoh, line 123: This `<SideNavMenu>` and `<SideNavMenuItem>`/`<SideNavLink>` in it seems to work as a set of component, like "`<SideNavMenu>` with back/forward feature". If so, to better align to the Web Component version we'll develop in future, and better separate the render-from-data logic vs. render-UI logic, I suggest we create one React component, so the markup here will be like `<SideNavMenuWithBackForward>{renderNavItem(item.megapanelContent.quickLinks.links)}</SideNavMenuWithBackForward>`, if it makes sense.

packages/react/src/components/Masthead/MastheadLeftNav.js
```diff
@@ -86,23 +87,75 @@ const MastheadLeftNav = ({ navigation, isSideNavExpanded, platform }) => {
...
+        sectionItems.push(
+          <SideNavMenu title={item.title} key={j}>
+            <SideNavMenuItem
+              onClick={event => event.preventDefault()}
+              className={`${prefix}--masthead__side-nav--submemu-back`}
+              data-autoid={`${stablePrefix}--masthead__l0-sidenav--subnav-back-${j}`}
+              isbackbutton="true"
+              key={i}>
+              <ChevronLeft20 />
+              Back
```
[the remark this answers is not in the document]
kennylam: @asudoh Yes, you're correct. This needs to be added to the main translation file.

packages/styles/scss/components/masthead/_masthead-leftnav.scss
```diff
@@ -53,6 +53,13 @@
     }
   }
 
+  .#{$prefix}--side-nav {
+    &__header-navigation {
```
asudoh, line 57: Another thing I'm wondering is if there are any ways that we can avoid BEM concatenation by `&`. Though doing so goes out of our existing pattern of our Sass code, it helps sharing React style and Web Components style. Thanks!

kennylam:
> Looks great!!! I noticed with with tab accessibility it skips over the back button. This is happening in the current version as well though, so unsure if this is a bug.

@ariellalgilmore Yes, you are right. Good catch!

packages/styles/scss/components/masthead/_masthead-leftnav.scss
```diff
@@ -137,11 +142,32 @@
...
         .#{$prefix}--masthead__side-nav--submemu-title {
-          @include carbon--type-style(expressive-heading-03, true);
+          @include use-carbon-productive-tokens();
+          @include carbon--type-style(productive-heading-02);
 
+          display: flex;
           color: $text-01;
-          padding: $carbon--spacing-05;
+          padding: 0 $carbon--spacing-05;
+          height: 3rem;
```
jeffchew, line 151: Minor thing, I wonder if we should use `$container-03` here?

cc: @annawen1 @asudoh

kennylam: @jeffchew Nice, I didn't know about this token, thanks!

packages/react/src/components/carbon-components-react/UIShell/SideNavMenu.js
```diff
@@ -104,10 +104,16 @@ export class SideNavMenu extends React.Component {
     };
   }
 
-  handleToggleExpand = () => {
+  handleToggleExpand = event => {
     this.setState(state => ({ isExpanded: !state.isExpanded }));
   };
 
+  handleKeyToggleExpand = event => {
+    if (event.charCode === 13 || event.charCode === 32) {
```
asudoh, line 112: Just for the sake of a modern way here...:

```suggestion
    if (event.key === 'Enter' || event.key === ' ') {
```

kennylam: @asudoh Haha I actually had it that way at first, but changed it because it seemed it a bit too semantic. I can change that np 👍

Issue #3350, opened 2020-07-30
Component: Locale Modal (React) - locations and languages does not reset to start of the list, on selection of different geographic area.


Praveen-sr:
## Detailed description

Issue :- When user scroll on one geographic area and then goes back and selects another geographic area,  locations and languages does not reset to start of the list

This bug also exists in the web components version.  That fix is managed under issue (#3476)

## Steps to reproduce the issue

1. Open Locale modal (react-canary) 
2. Click on Americas and scroll thro the list 
3. Now, click on "Select geographic area" 
4. Select / click on Asia
5. The locations and languages list is not reset and remains at the scroll position at Americas.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/4fe686e9-4119-4983-9a37-2d6c19af0b56)

PR #3465, opened 2020-08-06
fix(LocalModal): reset scroll to top

ariellalgilmore:
### Related Ticket(s)

#3350 

### Description

reset scroll when different geographic area is selected

### Changelog

**New**

- added scrollTop = 0


packages/react/src/components/LocaleModal/LocaleModalCountries.js
```diff
@@ -29,6 +29,7 @@ const LocaleModalCountries = ({
   ...modalLabels
 }) => {
   useEffect(() => {
+    const localeList = document.querySelector(`.${prefix}--locale-modal__list`);
```
asudoh, line 32: Probably we can grab this element via React ref?

ariellalgilmore: Thanks Akira! Just changed. Let me know if that is better!

packages/react/src/components/LocaleModal/LocaleModalCountries.js
```diff
@@ -28,6 +28,11 @@ const LocaleModalCountries = ({
   setClearResults,
   ...modalLabels
 }) => {
+  const localList = useRef(null);
+  useEffect(() => {
+    localList.current.scrollTop = 0;
+  });
```
asudoh, line 34: `useEffect()` with no "dependency arg" causes running the callback upon every render. Looks that we need to define `open` prop in `<LocaleModalCountries>` and have `<LocaleModal>` pass along the open state:

```suggestion
  useEffect(() => {
    if (open) {
      localList.current.scrollTop = 0;
    }
  }, [open]);
```

Issue #3406, opened 2020-08-04
Web Component:- Locale Modal:- There is no space between geo cards on locale modal


chetankan:
Issue Description:- In locale modal components there is no spacing between geo cards.

## Steps to reproduce the issue

1. Enter url in the browser.
2. Select 'Locale Modal' component.
3. Observe if there is spacing between cards.

OS:- Win 10, Mac, iOS 12, 13 Android 8,9
Browsers:- Chrome, Safari, Firefox, Edge, UC

URL:- http://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-locale-modal--default

![Locale Modal without spacing.jpg](https://images.zenhubusercontent.com/5e21a841639f517d456f7e43/98211f58-6617-4d01-a3d7-ee1580c57874)

jeffchew:
Added blocker https://github.com/carbon-design-system/ibm-dotcom-library/issues/3482 to this issue, as that should resolve the issue with the styling.

@RobertaJHahn Moving this to the next release.

cc: @ariellalgilmore

RobertaJHahn:
airtable updated.

PR #3411, opened 2020-08-04
feat(footer): introduction

asudoh:
### Related Ticket(s)

Refs #2862.

### Description

This change adds a Web Components version of footer.

### Changelog

**New**

- `<dds-footer-composite>`, for CSR the remainder of components from data.
- `<dds-footer>`, the top-level element in footer.
- `<dds-footer-logo>`, the logo in footer.
- `<dds-footer-nav>`, the primary nav in footer.
- `<dds-footer-nav-group>`, a group of links of `<dds-footer-nav>`.
- `<dds-footer-nav-item>`, a link in `<dds-footer-nav>`.
- `<dds-legal-nav>`, a secondary nav in footer.
- `<dds-legal-nav-item>`, a link in `<dds-legal-nav>`.
- `<dds-locale-button>`, a launcher for `<dds-locale-modal>`.

**Changed**

- Footer styles so it can be reused for Web Components.

PR #3470, opened 2020-08-06
fix(localeModal): add outline

ariellalgilmore:
### Related Ticket(s)

#3406

### Description

add outline to cards
Screen Shot 2020-08-06 at 1 43 53 PM

### Changelog

**New**

- outline styles to link

### DIFF
--- a/packages/web-components/src/components/locale-modal/locale-modal.scss
+++ b/packages/web-components/src/components/locale-modal/locale-modal.scss
@@ -163,6 +163,8 @@
     @extend .#{$prefix}--tile--clickable;
     @extend .#{$prefix}--card;
 
+    outline: 1px solid $ui-02;
+    outline-offset: -1px;
     flex: 1;
 
     &:focus {


jeffchew:
@ariellalgilmore this looks good, though I'm wondering if we should actually take this opportunity to modify the styles in the styles package instead, so that we can reduce the separate styles code in `web-components`? 

cc: @asudoh @kennylam @annawen1 @RaphaelAmadeu @andysherman2121

asudoh:
@jeffchew Good question, to do that, we need to make sure a CSS ruleset with `.#{$prefix}--locale-modal-container .#{$prefix}--locale-modal__regions .#{$prefix}--card` selector is applied to `:host(dds-region-item) .#{$prefix}--link`.

jeffchew:
> @jeffchew Good question, to do that, we need to make sure a CSS ruleset with `.#{$prefix}--locale-modal-container .#{$prefix}--locale-modal__regions .#{$prefix}--card` selector is applied to `:host(dds-region-item) .#{$prefix}--link`.
> 
> Changes to `styles` packages in #3411 will be a good reference to see how we can make such change to locale modal.

@ariellalgilmore @asudoh I created issues like this to be lined up: https://github.com/carbon-design-system/ibm-dotcom-library/issues/3471

Should I make one for Locale Modal (actually I think I will...)? I think doing that will solve this bug.

asudoh:
@jeffchew Yes makes sense.

ariellalgilmore:
@jeffchew just confirming this can be closed and should be fixed here: #3482

PR #3475, opened 2020-08-07
feat(web-components): add React wrapper

asudoh:
### Description

Adds React wrapper generator (`createReactCustomElementType`), based on one in `carbon-web-components` to `@carbon/ibmdotcom-web-components` codebase.

It helps our adoptors using React to use our components, with the following support that React itself lacks for non-intrinsic elements:

* Element attributes mapping to non-string element property values (e.g. Boolean attributes)
* React props mapping to element properties instead of element attributes
* Event handlers

Also adds a Babel plugin that automatically creates React wrappers of our components from our components' `@property()` annotations, as well as the corresponding build process that generates the React wrappers to `es/components-react` directory. So in the most typical scenario, our users can use components in `es/components-react` directory in the same manner as regular React components.

Also adds React stories so we can demonstrate the React integration.

### Changelog

**New**

- React wrapper generator (`createReactCustomElementType`)
- A Babel plugin that automatically creates React wrappers of our components from our components' `@property()`


jeffchew:
@asudoh going to add this as a huddle topic so we can get a walkthrough of this, thanks!

jeffchew:
@asudoh as discussed in huddle, this looks great! The missing pieces, whether if we want to include in this PR or not, are:

- documentation on the web component side on how to use the React wrapper instead of the direct web component way (for each component)
- proxy component on the `@carbon/ibmdotcom-react` side that would provide the react wrapper for the corresponding component. (these may be the separate stories already captured actually for each corresponding component).

asudoh:
Thank you for the discussion @jeffchew - Wrt documentation will address it in a separate PR. Wrt proxy, one question; how would we handle components with native React versions available (e.g. masthead) where I naïvely think we need to avoid name conflicts?

jeffchew:
> Thank you for the discussion @jeffchew - Wrt documentation will address it in a separate PR. Wrt proxy, one question; how would we handle components with native React versions available (e.g. masthead) where I naïvely think we need to avoid name conflicts?

We'll need to take on a by case basis. For the dotcomshell, I think we agreed we will be supporting the pure react version for the foreseeable future. For others, we need to think about what makes sense, perhaps for existing we add a prop `useWebComponentVersion` where if it's true, we would proxy to the web component wrapper instead of the pure react version, and have as opt-in for now since we don't want to affect existing adopters?

asudoh:
@jeffchew Implemented the React-to-WC proxy mechanism. It's enabled by turning on `DDS_USE_WEB_COMPONENTS_REACT` feature flag, and in such environment, the React stories will be replaced with WC-React-wrapper ones where WC version is available.

jeffchew:
@asudoh LGTM! It looks like there is an integration test failure though if you can look into that. Also, I'm thinking holding onto this PR til after code freeze so we don't impact the upcoming release.
cc: @kennylam @annawen1 @RaphaelAmadeu @andysherman2121 (please provide any additional feedback until then!)

asudoh:
@jeffchew After my investigation on `ci-check/check-packages` failure, I determined that we need to withdraw React-to-WC proxy, at least for where native React version is available. The specific failure is [`packages/react/ssr-tests`](https://github.com/carbon-design-system/ibm-dotcom-library/tree/v1.9.0/packages/react/ssr-tests), where we test if our React codebase can be loaded in node.js environment.

Here's the reason why it's failing; React `<DotcomShell>` code loads its Web Components counterpart, `<dds-dotcom-shell-container>`. `<dds-dotcom-shell-container>` loads `lit-element`, which is not available in CJS and thus node.js fails to parse  it.

So, here's my thought on mitigating the problem:

* React source codebase auto-proxying to WC for new components only (e.g. link list)
* For existing React components (e.g. `<DotcomShell>`), proxy React-to-WC in Storybook level, where we won't have SSR issue. Though we don't have a way for`<DotcomShell>` to auto-proxy to WC, having React storybook load WC version e.g. with a feature flag will be a reasonable demonstration of how people can migrate from pure React to WC.

Thoughts? Thanks!

jeffchew:
> @jeffchew After my investigation on `ci-check/check-packages` failure, I determined that we need to withdraw React-to-WC proxy, at least for where native React version is available. The specific failure is [`packages/react/ssr-tests`](https://github.com/carbon-design-system/ibm-dotcom-library/tree/v1.9.0/packages/react/ssr-tests), where we test if our React codebase can be loaded in node.js environment.
> 
> Here's the reason why it's failing; React `<DotcomShell>` code loads its Web Components counterpart, `<dds-dotcom-shell-container>`. `<dds-dotcom-shell-container>` loads `lit-element`, which is not available in CJS and thus node.js fails to parse it.
> 
> So, here's my thought on mitigating the problem:
> 
> * React source codebase auto-proxying to WC for new components only (e.g. link list)
> * For existing React components (e.g. `<DotcomShell>`), proxy React-to-WC in Storybook level, where we won't have SSR issue. Though we don't have a way for`<DotcomShell>` to auto-proxy to WC, having React storybook load WC version e.g. with a feature flag will be a reasonable demonstration of how people can migrate from pure React to WC.
> 
> Thoughts? Thanks!

I like the idea of the feature flag to turn on holistically, but also wondering if we should maybe have a per-component prop that will turn it on so it's more opt-in on a per component basis? That way adopters can even pick and choose which ones they want to switch to web components especially if they want to slowly adopt it rather than all at once. In this regard, this would be a knob in storybook, so that tests should theoretically pass in this case?

asudoh:
Here's the conversation I had with @jeffchew offline; The problem lies on calling WC code from React code itself. Given `import` is never conditional, if our React code uses WC code, even conditionally from React prop or feature flag, the code ends up with loading WC code for all configurations. Then React SSR environment not being able to load ESM happens.

asudoh:
Holding the effort for now. We see if we have urgent requirement of new React components by end of this year when Node 14, with native ESM support, becomes LTS.

PR #3490, opened 2020-08-07
fix(jest): remove console errors

caiobozato:
### Related Ticket(s)

Fix the console errors coming from React package #2693

### Description

Fixing console errors displayed when running `yarn test:unit`

PS: This PR fixes only some of the errors. More fixes need to be done.

### Changelog

**Changed**

- Minimum changes and fixes con components and test files in order to fix console error issues

### DIFF
--- a/packages/react/src/components/CTASection/__tests__/CTASection.test.js
+++ b/packages/react/src/components/CTASection/__tests__/CTASection.test.js
@@ -10,33 +10,45 @@ import CTASection from '../CTASection';
 import { mount } from 'enzyme';
 import React from 'react';
 
-const contentItemsProps = [
-  {
-    heading: 'Get connected',
-    copy: `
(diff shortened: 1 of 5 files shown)


packages/react/src/components/CTASection/__tests__/CTASection.test.js
```diff
@@ -10,33 +10,45 @@ import CTASection from '../CTASection';
...
-    copy: `
-      IBM DevOps partners have a wide range of expertise.
-      Find one to build the right solution for you.
-      `,
-    cta: {
-      copy: 'Find a partner',
-      type: 'external',
-      href: 'https://example.com/',
+const contentItemsProps = {
+  heading: 'Content item heading',
```
guilhermelMoraes, line 14: not a big deal, but, actually, both `heading` and `copy` belongs to `ContentBlock` and not `ContentItem`.

PR #3513, opened 2020-08-11
fix(LocaleModal): html semantic, filter by region, scrollTop by region 

diegomarcuz:
### Related Ticket(s)

Component:- Locale Modal :- Select region modal dialog #3409

### Changelog

**New**

- current region as a prop on LocaleModalCountries

**Changed**

- Changed the HTML semantic to ul/li on the `LocaleModalCountries` so `list` and `listitem` can be applied properly
- Filter countries list by region name.
- This commit: dbe5fcd won't necessary part of this PR, but without handling it now, scrollTop feature won't be accessible so I talked to Ari, because of this PR: https://github.com/carbon-design-system/ibm-dotcom-library/pull/3465 and we agree that check if the modal is `open` won't be a good approach so we changed the dependencies on useEffect.

**Removed**

- removed `display: none` on styles, because of the filter by region name
- removed the filtering on LocaleModalRegion, because of the filter by region name


packages/react/src/components/LocaleModal/LocaleModalCountries.js
```diff
@@ -21,19 +21,19 @@ const { prefix } = settings;
...
   ...modalLabels
 }) => {
   const localList = useRef(null);
   useEffect(() => {
-    if (open) {
-      localList.current.scrollTop = 0;
-    }
-  }, [open]);
+    localList.current.scrollTop = 0;
+  }, [currentRegion, regionList]);
```
asudoh, line 36: I think it still make sense to reset the scroll position only when the modal is open:

```suggestion
    if (open) {
      localList.current.scrollTop = 0;
    }
  }, [currentRegion, regionList, open]);
```

diegomarcuz: This props is being set on the LocaleModal component which contains LocaleModalCountries and LocaleModalRegions, so I need to send the open props to LocaleModalCountries and every time LocalModal is loaded/opened LocaleModalCountries is too, however LocaleModal always keeps loaded/opened so LocaleModalCountries too so the open dependency will never change, that's why we put currentRegion and regionList as dependency, because every time region is selected reset the scroll position. @asudoh

asudoh: Great point, thank you for elaborating. And my apologies I missed that `open` ref there seems to have never worked before-your-change.

One way we can address this is making `localeList` React ref in `<LocaleModalCountries>` a React prop, that is passed along from `<LocaleModal>`. In this way, `<LocaleModal>` can grab the ref and change the `.scrollTop` when needed. I imagine `<LocalModal>` has better context of when to reset the scroll position and thus I think "resetting scroll position" logic more robust.

jeffchew: @diegomarcuz @asudoh This is targeted for `v1.10.0`, think we can merge as-is then get in another change in a separate PR post code-freeze?

asudoh: @jeffchew Sounds good to me.

Issue #3159, opened 2020-07-15
Masthead:- Accessibility :- After activating open ibm search field.


chetankan:
Issue Description :- Visually only Search  edit combo , search icon, close icon and the  main content is present.
, but for screen reader users  screen reader is reading  controls which are not available visually.
for example hamburger menu, IBM logo link,  user profile button etc."

Recommended Fix:- screen reader should not read visually hidden components

Happening on :- Voiceover iPhone

URL : 
http://ibmdotcom-web-components-canary.mybluemix.net/?

PR #3542, opened 2020-08-12
fix(Masthead): hide nav items from screen readers when not displayed

ccwilson-us-ibm-com:
### Related Ticket(s)

#3159

### Description

Screen readers were picking up collapsed left nav menu items even when
collapsed or responsively hidden. And also the top nav items when hidden
by the search input.  Updated styles and rendering to prevent screen readers
from finding visually hidden navs.

### Changelog

**Changed**

- Fixed web-component masthead styles
- Hide web-component top nav when hidden by the search input


ccwilson-us-ibm-com:
Animation behavior is maintained when opening/closing left nav.

ccwilson-us-ibm-com:
1. This PR is actually doing some hiding in 2 areas.  One for the left nav, which is handled by the css.  And one by the top nav which is handled by the component.

Left Nav: Screen readers were picking this up even when not shown.  Both when closed and also responsively hidden for large screens.  So adding visibility to the CSS handles that.

Top Nav: When the search input is open, screen readers were picking up the now hidden top nav.  Since search open/closed events are handled at the component, the fix to hide the nav when the search is open is also in the component.

2. I'll check this scenario out.  Thanks.

packages/web-components/src/components/masthead/masthead.scss
```diff
@@ -152,18 +152,24 @@
...
   &[collapse-mode][usage-mode='header-nav'] {
+    visibility: hidden;
     width: 0;
+    transition: visibility 0.11s step-end 0s, width 0.11s cubic-bezier(0.2, 0, 1, 0.9) 0s;
   }
 
   &[expanded][usage-mode='header-nav'],
   &[collapse-mode][expanded][usage-mode='header-nav'] {
     @include carbon--breakpoint-down('lg') {
+      visibility: visible;
```
asudoh, line 163:
```suggestion
      visibility: inherit;
```

Ref: https://jakearchibald.com/2014/visible-undoes-hidden/

packages/web-components/src/components/masthead/masthead.scss
```diff
@@ -152,18 +152,24 @@
...
       width: mini-units(32);
+      transition: visibility 0.11s step-start 0s, width 0.11s cubic-bezier(0.2, 0, 1, 0.9) 0s;
     }
 
     @include carbon--breakpoint-down('md') {
+      visibility: visible;
```
asudoh, line 169:
```suggestion
      visibility: inherit;
```

Ref: https://jakearchibald.com/2014/visible-undoes-hidden/

jeffchew:
@ccwilson-us-ibm-com @asudoh just checking on this one, will this make Monday's code freeze?

cc: @RobertaJHahn

asudoh:
@ccwilson-us-ibm-com Please let me know I can be of any help on this, thanks!

ccwilson-us-ibm-com:
Ok.  Updated.

Each item that needs to be hidden handles its own listeners.  Not using the container now.  The logo will be hidden when search is opened on the large breakpoint.   And the menu button and profile button will be hidden when search is opened on the medium breakpoint.

packages/web-components/src/components/masthead/masthead-global-bar.ts
```diff
@@ -21,16 +22,70 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
...
+    this.openedSearch = (event as CustomEvent).detail.active;
+  }
+
+  /**
+   * Handles media query event.
+   *
+   * @param event The event.
+   */
+  private _handleMediaQuery(event: Event) {
+    console.log(event);
```
asudoh, line 63:
```suggestion
```

packages/web-components/src/components/masthead/masthead-global-bar.ts
```diff
@@ -21,16 +22,70 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
  */
 @customElement(`${ddsPrefix}-masthead-global-bar`)
 class DDSMastheadGlobalBar extends FocusMixin(LitElement) {
+  /**
+   * Media query on medium breakpoint.
+   */
+  private mqMedium;
```
asudoh, line 28: The codebase has a coding convention to prefix private properties with `_`.

```suggestion
  private _mqMedium;
```

packages/web-components/src/components/masthead/masthead-global-bar.ts
```diff
@@ -21,16 +22,70 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
...
    * The shadow slot this action bar should be in.
    */
   @property({ reflect: true })
   slot = 'profile';
 
+  /**
+   * Search bar opened flag.
+   */
+  @internalProperty()
+  private openedSearch = false;
```
asudoh, line 40: The codebase has a coding convention to prefix private properties with `_`.

```suggestion
  private _openedSearch = false;
```

packages/web-components/src/components/masthead/masthead-global-bar.ts
```diff
@@ -21,16 +22,70 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
...
+
+  /**
+   * Medium breakpoint flag.
+   */
+  @internalProperty()
+  private responsiveMd = false;
```
asudoh, line 46: The codebase has a coding convention to prefix private properties with `_`.

```suggestion
  private _responsiveMd = false;
```

packages/web-components/src/components/masthead/masthead-global-bar.ts
```diff
@@ -21,16 +22,70 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
...
+    this.getRootNode().addEventListener('dds-masthead-search-toggled', this._handleSearchToggle.bind(this));
+
+    this.mqMedium = window.matchMedia(`(max-width: ${breakpoints.md.width})`);
+    this.responsiveMd = this.mqMedium.matches;
+    this.mqMedium.addEventListener('change', this._handleMediaQuery.bind(this));
+  }
+
+  disconnectedCallback() {
+    this.getRootNode().removeEventListener('dds-masthead-search-toggled', this._handleSearchToggle.bind(this));
+    this.mqMedium.removeEventListener('change', this._handleMediaQuery.bind(this));
```
asudoh, line 78: `.bind()` creates new function instance upon every call, and thus the function refs here will be different from ones in `connectedCallback()`, which causes event listeners not removed.

To solve the problem, we use [Carbon core `on()` API](https://github.com/carbon-design-system/carbon/blob/v10.17.0/packages/components/src/globals/js/misc/on.js). Here's an example: https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/web-components/src/components/footer/footer-composite.ts#L139-L144

packages/web-components/src/components/masthead/masthead-global-bar.ts
```diff
@@ -21,16 +22,70 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
...
+   * @param event The event.
+   */
+  private _handleMediaQuery(event: Event) {
+    console.log(event);
+    this.responsiveMd = (event as MediaQueryListEvent).matches;
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+    this.getRootNode().addEventListener('dds-masthead-search-toggled', this._handleSearchToggle.bind(this));
```
asudoh, line 69: We make these custom event names configurable. Please see https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/web-components/src/components/footer/footer-composite.ts#L139-L144 for more details.

asudoh (commented):
Great progress - Thanks @ccwilson-us-ibm-com! Added some comments to `masthead-global-bar.ts`, which applies to a few other files, too. Please feel free to ping me fo any questions. Thanks!

packages/web-components/src/components/masthead/masthead-global-bar.ts
```diff
@@ -21,16 +22,70 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
...
+
+    this.mqMedium = window.matchMedia(`(max-width: ${breakpoints.md.width})`);
```
asudoh, line 71: The codebase avoids reference to global `window`. `this.ownerDocument.defaultView` (which is, the window object associated with the custom element) is used instead.

packages/web-components/src/components/masthead/masthead-global-bar.ts
ccwilson-us-ibm-com: Ok... This one is interesting.  Check out the `addEventListener` line again:
```
this.getRootNode().addEventListener('dds-masthead-search-toggled', this._handleSearchToggle.bind(this));
```

And checkout the same function call in the `on` utility:
```
element.addEventListener.apply(element, args);
```

Using this `apply` function there with `element` as the `addEventListener` context and the first argument in the `apply` function means that `element` becomes the `this` object in the callback used as the event listener.

But in my `addEventListener` call, the context of the `addEventListener` call is `this.getRootNode()` and the bound `this` object on the callback in the current class `DDSMastheadGlobalBar` object.  So they are different, so I can't use the `on` utility as it currently exists.

ccwilson-us-ibm-com: But..... trying the properly remove the listener made me run into another curiosity.

I add the listener to `this.getRootNode()` in the `connectedCallback` and I remove the listener from `this.getRootNode()` in the `disconnectedCallback`.  But no matter how I juggled the listener callback function and it's binding, I couldn't get the listener properly removed.

Until I realized that `this.getRootNode()` returns different objects in `connectedCallback` vs. `disconnectedCallback`.  `this.getRootNode()` return the document object in `connectedCallback`.  And it returns `<div id="root-inner"></div>` in the `disconnectedCallback`.

Is this a behavior you are familiar with?  It feels unexpected to me.  I haven't been able to find any concrete references to this in my web searches.

asudoh: I meant:

```javascript
// Static `eventSearchToggle` property must be set pointing to 'dds-masthead-search-toggled'
const { eventSearchToggle } from this.constructor as typeof DDSMastheadGlobalBar;
this._hSearchToggle = on(this.getRootNode(), eventSearchToggle, this._handleSearchToggle.bind(this));
```

`this._hSearchToggle` is a tiny object with `.release()` method that you can use it in `.disconnectedCallback()`.

```javascript
disconnectedCallback() {
  if (this._hSearchToggle) {
    this._hSearchToggle = this._hSearchToggle.release(); // Removes the event listener
  }
  (Do other stuffs)
  super.disconnectedCallback();
}
```

This pattern is similar to [what React `useEffect()` returns](https://reactjs.org/docs/hooks-effect.html#example-using-hooks-1). The difference is object plus `.release()` method vs. plain function that does the release - Carbon `on()` API takes the former so that a class can implement the same interface, and is seen in [Dojo `on()` API](https://dojotoolkit.org/reference-guide/1.7/dojo/on.html).

BTW [recent commit to `master`](https://github.com/carbon-design-system/ibm-dotcom-library/pull/3626) allows an alternate approach:

```javascript
// Static `eventSearchToggle` property must be set pointing to 'dds-masthead-search-toggled'
@HostListener('parentRoot:eventSearchToggle')
private _handleSearchToggle = (event: Event) => { // Arrow function for `this` binding
  this.openedSearch = (event as CustomEvent).detail.active;
};
```

Above code attaches (and automatically detaches) the event listener with a [decorator](https://github.com/tc39/proposal-decorators).

packages/web-components/src/components/masthead/top-nav.ts
```diff
@@ -21,18 +24,40 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
...
+    return this._hideNav ? html`` : super.render();
+  }
+
+  /**
+   * Handles toggle event from the search component.
+   *
+   * @param event The event.
+   */
+  // private _handleSearchToggle: (event: Event) => void;
+  @HostListener(`parentRoot:${DDSMastheadSearch.eventToggle}`)
```
asudoh, line 56: To support alternate (typically) component composed in masthead, this codebase makes event name configurable in receiving component side ([an example](https://github.com/carbon-design-system/carbon-web-components/blob/v1.5.0/src/components/ui-shell/side-nav.ts#L198-L200)). So please define `eventToggle` static property to each classes receiving the event:

```suggestion
  @HostListener(`parentRoot:eventToggle`)
```

packages/web-components/src/components/masthead/masthead-global-bar.ts
```diff
@@ -20,17 +25,73 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
...
+   */
+  @internalProperty()
+  private _responsiveMd = false;
+
+  /**
+   * Handles toggle event from the search component.
+   *
+   * @param event The event.
+   */
+  @HostListener(`parentRoot:${DDSMastheadSearch.eventToggle}`)
```
asudoh, line 57: To support alternate (typically) component composed in masthead, this codebase makes event name configurable in receiving component side ([an example](https://github.com/carbon-design-system/carbon-web-components/blob/v1.5.0/src/components/ui-shell/side-nav.ts#L198-L200)). So please define `eventToggle` static property to each classes receiving the event:

```suggestion
  @HostListener(`parentRoot:eventToggle`)
```

packages/web-components/src/components/masthead/masthead-logo.ts
```diff
@@ -35,15 +45,66 @@ class DDSMastheadLogo extends FocusMixin(LitElement) {
...
+   */
+  @internalProperty()
+  private _responsiveLg = false;
+
+  /**
+   * Handles toggle event from the search component.
+   *
+   * @param event The event.
+   */
+  @HostListener(`parentRoot:${DDSMastheadSearch.eventToggle}`)
```
asudoh, line 65: To support alternate (typically) component composed in masthead, this codebase makes event name configurable in receiving component side ([an example](https://github.com/carbon-design-system/carbon-web-components/blob/v1.5.0/src/components/ui-shell/side-nav.ts#L198-L200)). So please define `eventToggle` static property to each classes receiving the event:

```suggestion
  @HostListener(`parentRoot:eventToggle`)
```

packages/web-components/src/components/masthead/masthead-menu-button.ts
```diff
@@ -24,12 +29,49 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
...
+   */
+  @internalProperty()
+  private _responsiveMd = false;
+
+  /**
+   * Handles toggle event from the search component.
+   *
+   * @param event The event.
+   */
+  @HostListener(`parentRoot:${DDSMastheadSearch.eventToggle}`)
```
asudoh, line 60: To support alternate (typically) component composed in masthead, this codebase makes event name configurable in receiving component side ([an example](https://github.com/carbon-design-system/carbon-web-components/blob/v1.5.0/src/components/ui-shell/side-nav.ts#L198-L200)). So please define `eventToggle` static property to each classes receiving the event:

```suggestion
  @HostListener(`parentRoot:eventToggle`)
```

asudoh (commented):
Thank you @ccwilson-us-ibm-com for the updates!

packages/web-components/src/components/masthead/masthead-global-bar.ts
```diff
@@ -20,17 +24,80 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
...
+
+    const _mqMedium = this.ownerDocument?.defaultView?.matchMedia(`(max-width: ${breakpoints.md.width})`);
+    this._responsiveMd = _mqMedium ? _mqMedium.matches : false;
+
+    this._mqHandler = on(_mqMedium, 'change', this._handleMediaQuery.bind(this));
+  }
+
+  disconnectedCallback() {
+    if (this._mqHandler) {
+      this._mqHandler.release();
```
asudoh, line 89:
```suggestion
      this._mqHandler = this._mqHandler.release();
```

This codebase nulls out handle properties when they are no longer in use.

packages/web-components/src/components/masthead/masthead-global-bar.ts
```diff
@@ -20,17 +24,80 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
  * @element dds-masthead-global-bar
  */
 @customElement(`${ddsPrefix}-masthead-global-bar`)
-class DDSMastheadGlobalBar extends FocusMixin(LitElement) {
+class DDSMastheadGlobalBar extends FocusMixin(HostListenerMixin(LitElement)) {
+  /**
+   * Media query listener handler.
+   */
+  private _mqHandler;
```
asudoh, line 31:
```suggestion
  private _mqHandler: Handle | null = null;
```

Also please add `import Handle from '../../globals/internal/handle';`.

packages/web-components/src/components/masthead/masthead-logo.ts
```diff
@@ -22,7 +26,19 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
  * @element dds-masthead-logo
  */
 @customElement(`${ddsPrefix}-masthead-logo`)
-class DDSMastheadLogo extends FocusMixin(LitElement) {
+class DDSMastheadLogo extends FocusMixin(HostListenerMixin(LitElement)) {
+  /**
+   * Media query listener handler.
+   */
+  private _mqHandler;
```
asudoh, line 33:
```suggestion
  private _mqHandler: Handle | null = null;
```

Also please add `import Handle from '../../globals/internal/handle';`.

packages/web-components/src/components/masthead/masthead-menu-button.ts
```diff
@@ -24,12 +28,56 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
...
+
   /**
    * The handle for focus wrapping.
    */
   private _hFocusWrap: Handle | null = null;
 
+  /**
+   * Media query listener handler.
+   */
+  private _mqHandler;
```
asudoh, line 47:
```suggestion
  private _mqHandler: Handle | null = null;
```

Also please add `import Handle from '../../globals/internal/handle';`.

packages/web-components/src/components/masthead/masthead-logo.ts
```diff
@@ -35,15 +51,66 @@ class DDSMastheadLogo extends FocusMixin(LitElement) {
...
+
+    const mqLarge = this.ownerDocument?.defaultView?.matchMedia(`(max-width: ${breakpoints.lg.width})`);
+    this._responsiveLg = mqLarge ? mqLarge.matches : false;
+
+    this._mqHandler = on(mqLarge, 'change', this._handleMediaQuery.bind(this));
+  }
+
+  disconnectedCallback() {
+    if (this._mqHandler) {
+      this._mqHandler.release();
```
asudoh, line 101:
```suggestion
      this._mqHandler = this._mqHandler.release();
```

This codebase nulls out handle properties when they are no longer in use.

packages/web-components/src/components/masthead/masthead-menu-button.ts
```diff
@@ -54,6 +102,23 @@ class DDSMastheadMenuButton extends BXHeaderMenuButton {
...
+
+    const mqMedium = this.ownerDocument?.defaultView?.matchMedia(`(max-width: ${breakpoints.md.width})`);
+    this._responsiveMd = mqMedium ? mqMedium.matches : false;
+
+    this._mqHandler = on(mqMedium, 'change', this._handleMediaQuery.bind(this));
+  }
+
+  disconnectedCallback() {
+    if (this._mqHandler) {
+      this._mqHandler.release();
```
asudoh, line 116:
```suggestion
      this._mqHandler = this._mqHandler.release();
```

This codebase nulls out handle properties when they are no longer in use.

asudoh (commented):
Almost there - Thanks @ccwilson-us-ibm-com!

asudoh (approved):
Thank you for all the updates @ccwilson-us-ibm-com! 🎉

packages/styles/scss/components/masthead/_masthead-search.scss
```diff
@@ -122,7 +122,7 @@
 
       @include carbon--breakpoint-down(md) {
         width: 100vw;
-        right: (-$carbon--spacing-09);
```
asudoh, line 125: Percy seems to show a visual difference that may have been caused by this change. Need investigation.

jeffchew: screencapture-localhost-3000-testpages-v18-base-rtl-html-2020-08-28-00_16_45

Yes, this looks bad, why was this removed?

ccwilson-us-ibm-com: Hmm....

Here's what was going on.  That `right` property was originally there to strategically place the search bar in the right spot over that user button while is actually lived along side it.  When I was originally developing this feature, when I removed the user button when the search bar is open on the medium breakpoint, the search bar would be placed too far to the right.  So the `right` property was removed to get the search bar back in the correct spot.

The question here is on the right side of the comparison, why is the user button not removed while the search bar is open.  Taking a look.

ccwilson-us-ibm-com: I haven't been able to replicate this Percy view locally.  Is there a way I can get more info about how Percy generated this view?  Does Percy do a `click` event to open the search?  Or is there a way to directly load a page with the search already open?  ...maybe a page that Percy opens that has the `activate-search` property on the masthead container set to `true`?  Then the user button wouldn't get the click event that the search opened and set itself to be hidden.

asudoh: Probably a more important aspect of this is; The screenshot is of React. Meaning that the broken UI is not of Web Components but is of React.

ccwilson-us-ibm-com: Ah.  I understand. The style is shared across the packages. Ok.

PR #3564, opened 2020-08-12
feat(LinkList): Component created

RaphaelAmadeu:
### Related Ticket(s)

#3372 

### Description

Initial version of the LinkList component on the Web-Components package,
I still need some help on getting the CardLink styled properly.

Documentation will still be added too


packages/web-components/src/components/link-list/link-list-item.ts
```diff
@@ -0,0 +1,40 @@
...
+ * Link list item.
+ *
+ * @element dds-link-list-item
+ */
+@customElement(`${ddsPrefix}-link-list-item`)
+class DDSLinkListItem extends LitElement {
+  // eslint-disable-next-line class-methods-use-this
+  protected render() {
+    return html`
+      <li class="${prefix}--link-list__list__CTA">
```
asudoh, line 27: To reduce extra DOM element created by `<dds-link-list-item>`, would you set `listitem` to the `role` of this element and add the `:host(dds-link-list-item)` to the selector that styles `.bx--link-list__list__CTA`? Please see https://github.com/carbon-design-system/ibm-dotcom-library/blob/93db39f/packages/web-components/src/components/footer/footer-nav-item.ts#L40-L42 and https://github.com/carbon-design-system/ibm-dotcom-library/blob/93db39f3a3c304b0b064b37c3df6efe12fc1e429/packages/styles/scss/components/footer/_footer-nav-group.scss#L63-L90 as the reference. In this way, `.render()` has to render only the `<slot></slot>`. Thanks!

packages/web-components/src/components/link-list/link-list.scss
```diff
@@ -0,0 +1,73 @@
...
+  @extend .#{$prefix}--link-list__list__CTA;
+
+  ::slotted(#{$dds-prefix}-card-link) {
+    @extend .#{$prefix}--card;
+
+    border-top: 1px solid $ui-03;
+    border-bottom: 1px solid $ui-03;
+    margin-bottom: -1px;
+  }
+}
```
asudoh, line 73: Would you go `styles` packages and make the change there? I personally am fine with having styles defined here, but there is a push for reducing WC-specific styles.

packages/web-components/src/components/link-list/link-list.ts
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Link list.
+ *
+ * @element dds-link-list
+ */
+@customElement(`${ddsPrefix}-link-list`)
+class DDSLinkList extends LitElement {
+  @property({ type: String })
+  type = 'default';
```
asudoh, line 25: Could you define `default` here as an enum, like [one for footer size](https://github.com/carbon-design-system/ibm-dotcom-library/blob/93db39f/packages/web-components/src/components/footer/footer.ts#L41-L45)?

packages/web-components/src/components/link-list/link-list.ts
```diff
@@ -0,0 +1,59 @@
...
+const { stablePrefix: ddsPrefix } = ddsSettings;
+
+/**
+ * Link list.
+ *
+ * @element dds-link-list
+ */
+@customElement(`${ddsPrefix}-link-list`)
+class DDSLinkList extends LitElement {
+  @property({ type: String })
```
asudoh, line 24: `String` is the default `type`:

```suggestion
  @property()
```

packages/web-components/src/components/link-list/link-list.ts
```diff
@@ -0,0 +1,59 @@
...
+  type = 'default';
+
+  // eslint-disable-next-line class-methods-use-this
+  protected render() {
+    return html`
+      <div class="${prefix}--link-list" data-autoid="${(this.constructor as typeof DDSLinkList).stableSelector}">
```
asudoh, line 30: To reduce extra DOM element created by `<dds-link-list-item>`, would you add the `:host(dds-link-list-item)` to the selector that styles `.bx--link-list`? Please see https://github.com/carbon-design-system/ibm-dotcom-library/blob/93db39f/packages/web-components/src/components/footer/footer-nav-item.ts#L40-L42 and https://github.com/carbon-design-system/ibm-dotcom-library/blob/93db39f3a3c304b0b064b37c3df6efe12fc1e429/packages/styles/scss/components/footer/_footer-nav-group.scss#L63-L90 as the reference. In this way, `.render()` has to render only the `<slot></slot>`. Thanks!

packages/web-components/src/components/link-list/link-list-item.ts
```diff
@@ -0,0 +1,41 @@
...
+ * Link list item.
+ *
+ * @element dds-link-list-item
+ */
+@customElement(`${ddsPrefix}-link-list-item`)
+class DDSLinkListItem extends LitElement {
+  // eslint-disable-next-line class-methods-use-this
+  protected render() {
+    return html`
+      <slot @slotchange=${this.handleSlotChange} />
```
asudoh, line 25:
```suggestion
      <slot @slotchange="${this._handleSlotChange}" />
```

And could you implement `._handleSlotchange()` private method to change the `style-scheme` attribute? It should replace `if (!this.querySelector('*')?.hasAttribute('style-scheme')) { this.querySelector('*')?.setAttribute('style-scheme', 'link-list'); }`.

asudoh (commented):
Great progress, very close to finish! In addition to my comment I'm adding here, could you please check https://github.com/carbon-design-system/ibm-dotcom-library/pull/3564/files#r469598372 and https://github.com/carbon-design-system/ibm-dotcom-library/pull/3564/files#r469598245? They were marked as resolved, but from the code-wise it seems not. Thanks @RaphaelAmadeu!

packages/web-components/src/components/link-list/link-list.ts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+/**
+ * Link list.
+ *
+ * @element dds-link-list
+ */
+@customElement(`${ddsPrefix}-link-list`)
+class DDSLinkList extends LitElement {
+  @property()
```
asudoh, line 49: Assuming we have an attribute selector for `type`. If not, please disregard:

```suggestion
  @property({ reflect: true })
```

packages/web-components/src/components/link-list/link-list-item.ts
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * Default style
+   */
+  DEFAULT = '',
+
+  /**
+   * Link List variant
+   */
+  LINK_LIST = 'link-list',
+}
```
asudoh, line 28: I should have mentioned earlier (my apologies), can we define this enum in `link-with-icon.ts` and/or `card-link.ts`? We need to define the corresponding properties to one or both of those, too, like:

```typescript
@property({ reflect: true, attribute: 'style-scheme' })
styleScheme = STYLE_SCHEME.DEFAULT;
```

packages/web-components/src/components/link-list/link-list-item.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  connectedCallback() {
+    if (!this.hasAttribute('role')) {
+      this.setAttribute('role', 'listitem');
+    }
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  private _handleSlotChange(e) {
+    e.target?.assignedNodes?.setAttribute('style-scheme', STYLE_SCHEME.LINK_LIST);
```
asudoh, line 52:
```suggestion
    e.target?.assignedNodes?.forEach(node => { node.styleScheme = STYLE_SCHEME.LINK_LIST; });
```

Please also see my comment to `STYLE_SCHEME` enum.

asudoh (commented):
Thank you for another update, getting close - Thanks @RaphaelAmadeu!

packages/web-components/src/components/link-list/link-list-item.ts
```diff
@@ -0,0 +1,56 @@
...
+  protected render() {
+    return html`
+      <slot @slotchange=${this._handleSlotChange} />
+    `;
+  }
+
+  private _handleSlotChange(e) {
+    if (!this.hasAttribute('role')) {
+      this.setAttribute('role', 'listitem');
+    }
```
asudoh, line 47: Given this changes the attribute of host element, this part needs to be moved to `.connectedCallback()`.

asudoh (commented):
Thank you for the updates @RaphaelAmadeu!

packages/web-components/src/components/link-list/link-list.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  // eslint-disable-next-line class-methods-use-this
+  protected render() {
+    return html`
+      <h4 class="${prefix}--link-list__heading"><slot name="heading" /></h4>
+      <ul name="list" class="${prefix}--link-list__list ${this.ulClasses()}">
+        <slot
+          @slotchange=${() => {
+            this._handleSlotChange();
+          }}
```
asudoh, line 75: Creating a function in every render call `render()` tend to call memory pressure (more garbage collections):

```suggestion
          @slotchange=${this._handleSlotChange}
```

asudoh (commented):
Almost there @RaphaelAmadeu!

1. Could you add comments to properties/methods, etc.? [This document](https://github.com/carbon-design-system/ibm-dotcom-library/blob/v1.10.0-rc.2/packages/web-components/docs/coding-conventions.md#tsdoc-comments) explains how the codebase puts comments.
2. Could make the grids in the stories Storybook decorators, like [the one for card link story](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/web-components/src/components/card-link/__stories__/card-link.stories.ts#L32-L40)? Please note that you can define story-specific decorator by defining e.g. `EndOfSectionSplit.story.decorators`.

Thanks!

asudoh:
Thank you for the update @RaphaelAmadeu! I believe some of the decorators are the same, right? If so, probably we can make some decorators simply refer to other?

asudoh (approved):
LGTM 👍🎉 - Thanks @RaphaelAmadeu for all the updates!

packages/web-components/src/components/link-list/link-list.ts
asudoh: Reminder: Let's not forget this part:

```suggestion
          @slotchange=${this._handleSlotChange}
```

packages/web-components/src/components/link-list/link-list.ts
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  /**
+   * Handler for @slotChange, toggles the split layout class and set the children link-list-item to the same height
+   *
+   * @private
+   */
+  private _handleSlotChange() {
+    const childItems = this.shadowRoot
+      ?.querySelector('ul slot')
```
asudoh, line 106: I think below change along with `@slotchange` optimization (see my earlier comments) will resolve the TS typecheck error:

```suggestion
  private _handleSlotChange(event: Event) {
    const childItems = (event.target as HTMLSlotElement)
```

RaphaelAmadeu: @asudoh this change breaks the functionality, as we can't access elements from the shadow dom from the slot

asudoh: I thought `this.shadowRoot.querySelector('ul slot')` and `event.target` here yields to the same element `<slot>`. Did you see it differently?

RaphaelAmadeu: @asudoh I've addressed this issue

packages/web-components/src/components/link-list/link-list-item.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  /**
+   * handler, toggles link list style scheme on child components.
+   *
+   * @param e onSlotChange Event
+   * @private
+   */
+  // eslint-disable-next-line class-methods-use-this
+  private _handleSlotChange(e) {
+    const nodes = e.target.assignedNodes();
```
asudoh, line 59:
```suggestion
  private _handleSlotChange(event: Event) {
    const nodes = (event.target as HTMLSlotElement).assignedNodes();
```

asudoh (commented):
Just a bit of change suggestion to better align to the rest of the codebase. Also, could you update the snapshot by following the instruction at: https://github.com/carbon-design-system/ibm-dotcom-library/blob/v1.10.0/packages/web-components/docs/developer.md#running-unit-test

Thanks @RaphaelAmadeu!

packages/web-components/src/components/link-list/link-list-item.ts
jeffchew: @RaphaelAmadeu fyi we also created a yarn script for updating the snapshots: `yarn test:unit:updateSnapshot`

asudoh:
To continue on @RaphaelAmadeu's great effort, made an update with the following changes:

* Merged `<dds-link-with-icon>` feature to `<dds-link-list-item>` and introduced `<dds-link-list-item-card>` so we don't have to propagate `style-scheme` attribute.
* Reduced Web Components-specific style rules. The remainder seems absolutely necessary.
* Introduced a story decorator so the grid part can be reused across stories

jeffchew:
@asudoh thank you for making the updates! Super minor thing, was a bottom margin or padding added to link list? Percy is showing a little bit of extra spacing in some snapshots.

jeffchew (approved):
Approving this, thank you for the huge effort here @RaphaelAmadeu and for the assistance @asudoh ! I only have a small note but I don't think it's blocking this PR.

Issue #3184, opened 2020-07-16
[Masthead]: fix L0/L1 scroll behavior


kennylam:
## Detailed description

> Describe in detail the issue you're having.

L0 should begin scrolling out of viewport when 25% of the document height is reached. If L1 is present it will remain pinned to top of viewport.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

> What did you expect to happen? What happened instead? What would you like to
> see changed?

> What browser are you working in?

> What version of the IBM.com Library are you using?

`latest`

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

- The PR 3133 should be reverted before the EOD 7/17.  **Once reverted, this issue should not be closed, but moved into sprint 20-15.**
- This fix will be included in the August release.

PR #3577, opened 2020-08-14
fix(Masthead): scroll behavior

diegomarcuz:
### Related Ticket(s)

[Masthead]: fix L0/L1 scroll behavior #3184

### Description

if both L0 and L1 are present, L1 will be sticky when 25% of page is reached. If there is only the L0, it will be scroll out when 25% of page is reached.

### Changelog

**New**

- `not` selector to "remove" L0 when there is only it.

**Changed**

- CSS specificity when there is both L0 and L1 so only L1 will be sticky and L0 will be scroll out.

**Removed**

- condition to apply 25% of page rule only when there is the L1


jeffchew:
@diegomarcuz adding `feature flag` so we can test the L1 as well.

diegomarcuz:
> @diegomarcuz adding `feature flag` so we can test the L1 as well.

Thanks @jeffchew

jeffchew:
@diegomarcuz I'm testing this out and the masthead is disappearing way before 25% from what I can tell, @wonilsuhibm @oliviaflory @sethibm @JennySanchez can you take a look at this behavior?

https://ibmdotcom-react.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/3577/index.html?path=/story/components-dotcom-shell--default

diegomarcuz:
> I'm testing this out and the masthead is disappearing way before 25% from what I can tell

Really ? I made some `console.log()` to see when the `.bx--masthead--sticky` is added and seems to be right. Look here @jeffchew 

This class only appears when scroll position is greater than 25% of the page height.

Screen Shot 2020-08-14 at 12 24 38

Screen Shot 2020-08-14 at 12 24 31

kennylam:
@diegomarcuz I'm also seeing the masthead disappear before 25%. The L0 disappears almost immediately whether the L1 is present or not.

packages/react/src/components/Masthead/Masthead.js
```diff
@@ -145,11 +145,9 @@ const Masthead = ({
        * L0 will hide/show only in the top 25% of the viewport.
        *
        */
-      if (mastheadL1Ref.current != null) {
-        setIsMastheadSticky(
-          root.pageYOffset > root.innerHeight * hideTopnavThreshold
-        );
-      }
+      setIsMastheadSticky(
+        root.pageYOffset > root.innerHeight * hideTopnavThreshold
```
annawen1, line 149: I wonder if the calculation needs to be tweaked here, this statement becomes true too soon after user starts scrolling

diegomarcuz: I think the calculation is fine, because it is getting 25% of the page and comparing with current page position. if it is too soon, I think that need to change the threshold (25%).

jeffchew:
Will re-open a new PR as this was never completed with a working solution.

PR #3579, opened 2020-08-14
fix(masthead): added entry point to self-init redux store

jeffchew:
### Related Ticket(s)

No related issues

### Description

One of the discoveries made is that the store has to be initialized in order for
the Masthead to render properly. Rather than adding that to the documentation
for the `quick start` version, this entry point was created so that adopters 
do not need to do this on their own. Most adopters most likely won't need to 
bootstrap the redux store manually. Documentation has been add for application 
teams that do want to do that though.

Some other things added were the `dotcom-shell` code sandbox, as that appears
to be missing. 

The `getLang` attribute was also getting hardcoded into the storybook stories. 
This adds logic to only apply if using a mock.

### Changelog

**New**

- `dotcom-shell` codesandbox example
- `footer` codesandbox example
- new entry points for `dotcom-shell` and `masthead` to auto-bootstrap redux store

### DIFF
--- a/packages/web-components/examples/codesandbox/components/dotcom-shell/.babelrc
+++ b/packages/web-components/examples/codesandbox/components/dotcom-shell/.babelrc
@@ -1,0 +1,22 @@
+{
+  "presets": [
+    [
+      "@babel/preset-env",
+      {
+        "modules": false,
+        "targets": [
+          "last 1 version",
+          "Firefox ESR",
+          "not opera > 0",
+          "not op_mini > 0",
+          "not op_mob > 0",
+          "not android > 0",
+          "not edge > 0",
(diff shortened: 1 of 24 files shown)


asudoh:
@jeffchew Thank you for jumping in and creating this PR! Sent a message offline for further discussion.

asudoh:
Upon a conversation offline with @jeffchew, we decided to go with the PR to get some portion of them, and @asudoh to create a new PR with the discussed API, where `<dds-masthead-container>` uses the default store without having to reducer manually configured, and creates `<dds-masthead-composite>` that is "`<dds-masthead-container>` minus store integration" and allows our users to manually configure the store. Thanks @jeffchew for the discussion!

Issue #3589, opened 2020-08-17
Web Components: Footer not rendering properly in Storybook


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

The footer links are not rendering properly in storybook:

<img width="1678" alt="Screen Shot 2020-08-17 at 11 40 48 AM" src="https://user-images.githubusercontent.com/1641214/90415151-ac556480-e07e-11ea-863b-f8bdf499ba88.png">

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

> Is this issue related to a specific component?

> What did you expect to happen? What happened instead? What would you like to
> see changed?

> What browser are you working in?

> What version of the IBM.com Library are you using?

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. This is happening on both canary and staging URLs:
- https://ibmdotcom-web-components-canary.mybluemix.net
- https://ibmdotcom-web-components-staging.mybluemix.net

## Additional information

- This might be a storybook configuration, but we should hotfix this.

PR #3590, opened 2020-08-17
fix(connect): look at store in initialization

asudoh:
### Related Ticket(s)

Refs #3589.

### Description

This change ensures that `<dds-footer-container>`, etc. properties are populated from Redux store when the component is initialized, in addotion to reacting to changes in Redux store.

Lack of such mechanism caused empty footer data when a new instance of `<dds-footer-container>` is created after Redux store has already loaded all data from the service.

### Changelog

**New**

- "Store to component props" logic at component initialization in `<dds-footer-container>`, etc.


jeffchew:
@asudoh fyi, I opened up PR https://github.com/carbon-design-system/ibm-dotcom-library/pull/3595 that has update snapshot files to fix the CI issues.

asudoh:
Superseded by: #3595

Issue #3407, opened 2020-08-04
Web component: Card Link - card width and top alignment is not as per specs.


Praveen-sr:
## Detailed description

Issue :- On web component Card Link - 
1. Card width is 420px but on 'react' its 320px
2. Card is aligned 48px from top, but on 'react' no spacing from top.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/985f5edd-0f95-4ea4-a316-edefdd653e13)

RaphaelAmadeu:
The spacing you see on top is not from card-link, but from web-components storybook.

PR #3591, opened 2020-08-17
fix(card-link): witdth in web-component story now matches react

RaphaelAmadeu:
### Related Ticket(s)

#3407 
### Description

Added a grid wrapper, so the width of the card-link matches it's react story version.


RaphaelAmadeu:
> Good catch @RaphaelAmadeu! Would you be able to add a story decorator like: https://github.com/carbon-design-system/ibm-dotcom-library/blob/v1.9.0/packages/web-components/src/components/cta/__stories__/cta.stories.ts#L56-L67
> 
> This helps potential need to add a story in future (given the decorator can be reused among stories). Thanks!

@asudoh I've updated the PR, thanks for the review.

asudoh (approved):
LGTM 👍 - Thanks @RaphaelAmadeu for the update!

PR #3599, opened 2020-08-18
chore(devenv): fix footer size

asudoh:
### Description

This change ensures that the default footer story uses regular size.

Lack of this change caused the default footer story use short size when going to short story and then default story.

### Changelog

**New**

-  Code to ensure that the default footer story uses regular size.


asudoh:
@jeffchew Yes I meant to that, sorry - Changed.

Issue #1644, opened 2020-03-10
Pattern:- Table of Content:- In landscape mode title are not getting displayed in bold(iOS).


chetankan:
Issue:- On iOS mobile phone in landscape mode 'Table of Content' titles are not getting displayed in bold.

## Steps to reproduce the issue

1. Enter the url in thebrowser
2. Click on 'Table of Contents'
3. Turn mobile in landscape mode.

## Additional information

OS:- iOS 12, 13
Device :- iPhone 11 pro, 11, Xs, Xs max, X, 8![Table of contents in landscape mode.png](https://images.zenhubusercontent.com/5e21a841639f517d456f7e43/0c14bc2d-4d99-46e4-9747-91a889e8160a)
Browser:- Chrome, Safari

jeffchew:
Moving this out to sprint 7.

RaphaelAmadeu:
This is caused by IOS and how it handles the font-sizes on landscape, as this is a low priority and none of the found solutions are suitable, we are putting this on hold.

jeffchew:
@RobertaJHahn I don't think this is a `Sprint Must Have`, right? I think we should push this out.

cc: @ljcarot @RaphaelAmadeu

ljcarot:
I agree we can push this out if sev 4.

jeffchew:
@RobertaJHahn moving this out to the next sprint.

diegomarcuz:
Hey @chetankan, I was checking it and seems to be bold, could you check again ?

This screenshot is from Iphone 11 Pro IOS 13

![Screen Shot 2020-08-12 at 11.18.03.png](https://images.zenhubusercontent.com/5eed25994ac8fb057241e3b2/3cfb29f4-f00c-464c-b644-a192de1a5161)

diegomarcuz:
@chetankan checking again...

diegomarcuz:
Hey @chetankan, could you check again ?? PR: [#3614](https://app.zenhub.com/workspace/o/carbon-design-system/ibm-dotcom-library/issues/3614)

PR #3614, opened 2020-08-18
fix(TOC): font-size on ios landscape mode

diegomarcuz:
### Related Ticket(s)

Pattern:- Table of Content:- In landscape mode title are not getting displayed in bold(iOS). #1644

### Description

Fix the display font-size on landscape mode for ios.

### Changelog

**New**

- `text-size-adjust` property

Screen Shot 2020-08-18 at 14 30 46


packages/styles/scss/components/tableofcontents/_tableofcontents.scss
```diff
@@ -46,6 +46,11 @@
 
 @mixin tableofcontents {
   .#{$prefix}--tableofcontents {
+    text-size-adjust: 100%;
+    -webkit-text-size-adjust: 100%;
+    -ms-text-size-adjust: 100%;
+    -moz-text-size-adjust: 100%;
```
kennylam, line 52: @diegomarcuz Are these prefixes needed?

asudoh: Yeah Carbon (and I believe our library) requires `autoprefixer` in place for Sass/CSS build chain.

diegomarcuz: ohhh got it, I thank you @asudoh, does it occurs even on dev mode ?

asudoh: Not certain, but it's a requirement of Carbon.

diegomarcuz: Because on my dev mode, It doesn't occur. When I remove the `webkit` prefix the property doesn't work. @asudoh

asudoh: Would you try looking at the Storybook build config and make sure it's configured with the latest `autoprefixer`?

diegomarcuz: it is what I found, is it right ?

Screen Shot 2020-08-24 at 09 25 44

asudoh: Does the browserslist setting cover the browsers you are dealing with? Also ref: https://github.com/postcss/autoprefixer/issues/264

diegomarcuz: Checking, because I am not so familiar with `autoPrefixer`.

asudoh:
```suggestion
```

With #3706, those style rules will be generated automatically.

diegomarcuz: @asudoh, Thanks so much for your help, I removed the prefixes.

asudoh (approved):
Thanks @diegomarcuz for the update!

PR #3625, opened 2020-08-19
chore(services-store): add store for search API

asudoh:
### Description

This change introduces Redux store for search API, eliminating all API-call logic from our Custom Elements implementations.

### Changelog

**New**

- Redux store for search API.

**Changed**

- `<dds-masthead-container>` and `<dds-dotcom-shell-container>` to the new Redux store feature.

**Removed**

- Temporary mix-in component for search that used to interact with search API (`DDSMastheadSearchContainerMixin`).


packages/web-components/src/globals/services-store/actions/searchAPI.ts
```diff
@@ -0,0 +1,119 @@
...
+import { SEARCH_API_ACTION, SearchAPIState } from '../types/searchAPI';
+
+/**
+ * @param language The locale.
+ * @param searchQueryString The search query string.
+ * @returns The endpoint URL for the given locale and search query string.
+ */
+function getSearchEndpoint(language: string, searchQueryString: string) {
+  const [primary, country] = language!.split('-');
+  return `https://www-api.ibm.com/search/typeahead/v1?lang=${primary}&cc=${country}&query=${searchQueryString}`;
```
jeffchew, line 21: The host is configurable via environment variable, wondering if we shouldn't be hardcoding it here and instead have a default value if not configured? https://github.com/carbon-design-system/ibm-dotcom-library/blob/14acb43fdc01937dc465f8ae2f768499465980d6/packages/services/src/services/SearchTypeahead/SearchTypeahead.js#L16

asudoh: Makes sense @jeffchew - Updating:

```suggestion
  return `${(process && process.env.SEARCH_TYPEAHEAD_HOST) || 'https://www-api.ibm.com'}/search/typeahead/v1?lang=${primary}&cc=${country}&query=${searchQueryString}`;
```

PR #3645, opened 2020-08-19
feat(Footer): add micro variation

kennylam:
### Related Ticket(s)

#3431 

### Description

Adds `micro` variation to Footer component

### Changelog

**New**

- `micro` version added to Footer component

### DIFF
--- a/packages/react/src/components/Masthead/Masthead.js
+++ b/packages/react/src/components/Masthead/Masthead.js
@@ -259,7 +259,6 @@ const Masthead = ({
               {(mastheadL1Data || navigation) && (
                 <HeaderMenuButton
                   aria-label={isSideNavExpanded ? 'Close menu' : 'Open menu'}
-                  id={`${stablePrefix}--masthead__hamburger`}
                   data-autoid={`${stablePrefix}--masthead-${navType}-sidenav__l0-menu`}
                   onClick={onClickSideNavExpand}
                   isActive={isSideNavExpanded}
--- a/packages/react/src/components/Masthead/Masthead.js
+++ b/packages/react/src/components/Masthead/Masthead.js
@@ -259,6 +259,7 @@ const Masthead = ({
               {(mastheadL1Data || navigation) && (
                 <HeaderMenuButton
                   aria-label={isSideNavExpanded ? 'Close menu' : 'Open menu'}
+                  id={`${stablePrefix}--masthead__hamburger`}
                   data-autoid={`${stablePrefix}--masthead-${navType}-sidenav__l0-menu`}
                   onClick={onClickSideNavExpand}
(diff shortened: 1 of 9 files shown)


packages/styles/scss/components/footer/_legal-nav.scss
```diff
@@ -33,60 +33,91 @@
...
+          }
+        }
       }
-      &:first-child {
-        @include carbon--make-col-offset(1, 4);
 
-        padding-left: $carbon--spacing-03;
+      .#{$prefix}--locale-btn__container {
+        margin: 0;
+        text-align: right;
```
ariellalgilmore, line 89: @kennylam looks like this text-align is causing the locale modal text to be adjust as well

kennylam: @ariellalgilmore Thanks, good catch! That doesn't even need to be there any more.

jeffchew:
@kennylam can you enable the language only selector knobs for the micro footer stories? This will be needed to test as that is how knowledge center will be adopting it.

packages/react/src/components/Masthead/Masthead.js
```diff
@@ -259,6 +259,7 @@ const Masthead = ({
               {(mastheadL1Data || navigation) && (
                 <HeaderMenuButton
                   aria-label={isSideNavExpanded ? 'Close menu' : 'Open menu'}
+                  id={`${stablePrefix}--masthead__hamburger`}
```
asudoh, line 262: @kennylam Wondering this is an attempt to solve a similar problem to #3660?

kennylam: @asudoh Yes, was just about to leave a comment in your PR. The masthead `autoid` is now variable, having `default`, `eco`, etc, types. I think it will be more reliable to target this way.

jeffchew:
Hey @kennylam , I'm getting this when I click the language selector...

Screen Shot 2020-08-20 at 9 55 11 AM

kennylam:
> Hey @kennylam , I'm getting this when I click the language selector...

@jeffchew Looks like this is an existing issue since it appears on [experimental](http://ibmdotcom-react-experimental.mybluemix.net/?path=/story/components-dotcom-shell--default) as well. Suggest we open a separate issue for this.

jeffchew:
> > Hey @kennylam , I'm getting this when I click the language selector...
> 
> @jeffchew Looks like this is an existing issue since it appears on [experimental](http://ibmdotcom-react-experimental.mybluemix.net/?path=/story/components-dotcom-shell--default) as well. Suggest we open a separate issue for this.

Thanks for checking! Issue created here: https://github.com/carbon-design-system/ibm-dotcom-library/issues/3665

PR #3654, opened 2020-08-19
feat(card-with-pictogram): adding new card variation

guilhermelMoraes:
### Description

This card has a pictogram at the bottom left side, as per the visual specs for the [Support Page](https://app.abstract.com/projects/7f83c114-7627-4169-8af7-6a85e7147ea6/branches/master/commits/latest/files/01ee3bd8-c63d-4b3b-bddb-aabaddb5c31d/layers/1FA6F2AC-E027-419F-8CEF-F58B3AE2E495?collectionId=c963cb53-73c2-42a5-af39-f5538e24c454&collectionLayerId=a278337e-62ec-464c-8ba6-b368d4b9ed64).

This variation was approved by Wonil, James, and Jeff and will be under a feature flag. Here's the [link to the Slack conversation](https://cognitive-app.slack.com/archives/G018JGL83QX/p1597692491002100).

Captura de Tela 2020-08-19 às 17 52 10

### DIFF
--- a/packages/react/src/components/Card/Card.js
+++ b/packages/react/src/components/Card/Card.js
@@ -31,6 +31,7 @@ export const Card = ({
   customClassName,
   copy,
   cta,
+  pictogram,
   ...props
 }) => {
   const TileType = props.disabled ? Tile : ClickableTile;
@@ -57,7 +58,7 @@ export const Card = ({
         {eyebrow && <p className={`${prefix}--card__eyebrow`}>{eyebrow}</p>}
         {heading && <h3 className={`${prefix}--card__heading`}>{heading}</h3>}
         {optionalContent(copy)}
-        {renderFooter(cta)}
+        {renderFooter(cta, pictogram)}
       </div>
     </TileType>
   );
@@ -85,7 +86,7 @@ function optionalContent(copy) {
  * @param {object} cta cta object
  * @returns {object} JSX object
  */
-function renderFooter(cta) {
+function renderFooter(cta, pictogram) {
   return (
     cta && (
       <div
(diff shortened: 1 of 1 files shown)


packages/react/src/components/Card/__stories__/Card.stories.js
```diff
@@ -32,19 +34,22 @@ export default {
...
-            src: ArrowRight20,
-          },
-          iconPlacement: select(
-            'Cta icon placement (cta.iconPlacement)',
-            ['left', 'right'],
-            'right',
-            groupId
-          ),
-        },
+        cta: !DDS_CARD_WITH_PICTOGRAM
```
jeffchew, line 37: @guilhermelMoraes one thought, not sure if we should be doing this logic based on the setting of the flag. Instead, we should probably check if the pictogram is defined or not. An adopter might want to have a mix of cards with pictograms and cards with CTAs on the same application.

packages/react/src/components/Card/__stories__/Card.stories.js
```diff
@@ -32,19 +34,22 @@ export default {
...
+              },
+              iconPlacement: select(
+                'Cta icon placement (cta.iconPlacement)',
+                ['left', 'right'],
+                'right',
+                groupId
+              ),
+            }
+          : null,
+        pictogram: DDS_CARD_WITH_PICTOGRAM ? <Bee /> : null,
```
jeffchew, line 52: I wonder if we should show as a knob to use pictogram instead of cta...

guilhermelMoraes: Makes sense. What about a boolean knob, like: "Show pictogram (hides CTA):"?

Issue #3641, opened 2020-08-19
[MM] When I scroll down and then up a menu, clicking on a mega menu doesn't take me to the destination - Safari only


lily-kenney:
## Detailed description

> Describe in detail the issue you're having.

When I open a mega menu and then scroll down and then back up and then attempt to click on a mega menu item, it doesn't bring me to the destination page but acts more like a jump up the mega menu.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

Carbon mega menu integration with the Carbon US HP

> What did you expect to happen? What happened instead? What would you like to
> see changed?

I expect that when I click on a menu item, it'll bring me to the corresponding destination page instead of jump up the mega menu.

> What browser are you working in?

Mac, Version 13.1.2 (15609.3.5.1.3)

> What version of the IBM.com Library are you using?

The one that is integrated with https://wwwstage.ibm.com/thought-leadership/ryan-test/index.html

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

8/31 carbon mega menu + carbon us hp launch date

## Steps to reproduce the issue

1. Go to https://wwwstage.ibm.com/thought-leadership/ryan-test/index.html
2. Click on Products & Solutions
3. Scroll down the menu to the bottom
4. Click on an item (e.g. Manufacturing)
5. Instead of taking you to the manufacturing page, it brings you up the menu

## Additional information

- Video at: https://ibm.box.com/s/qd8pvoqlyo9efjarj3p7y66vaf2s1bws

PR #3673, opened 2020-08-20
fix(MegaMenu): add tabindex attr for safari not detecting event

annawen1:
### Related Ticket(s)

[MM] When I scroll down and then up a menu, clicking on a mega menu doesn't take me to the destination - Safari only #3641

### Description

When clicking on a link in the megamenu, it triggers the `onBlur` function which checks for the `event.relatedTarget`. This is coming back null on safari. Adding `tabIndex={0}` makes the element appear in the `event.relatedTarget`

https://stackoverflow.com/questions/42764494/blur-event-relatedtarget-returns-null

### Changelog

**Changed**

- add `tabIndex={0}` to the megamenu links


annawen1:
@lily-kenney tested the preview link and it's working for her on safari

Issue #3633, opened 2020-08-19
[MM] Attempting to scroll through the mega menu and the home page scrolls instead


lily-kenney:
## Detailed description

> Describe in detail the issue you're having.

I click open one of the L0 labels and the corresponding mega menu opens. When I'm attempting to scroll down the mega menu on the trackpad on my Mac, the home page underneath scrolls instead. This is not expected as since the mega menu is in focus, I would expect the mega menu to be scrolling.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

The carbon mega menu on top of the carbon home page.

> What did you expect to happen? What happened instead? What would you like to
> see changed?

I expect that when the mega menu is in focus, I would be scrolling through the mega menu and not the page itself.

> What browser are you working in?

Mac, Safari Version 13.1.2 (15609.3.5.1.3)

> What version of the IBM.com Library are you using?

The one pulled onto this site: https://wwwstage.ibm.com/thought-leadership/ryan-test/index.html

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

The 8/31 launch date for the mega menu + US HP

## Steps to reproduce the issue

1. Go to https://wwwstage.ibm.com/thought-leadership/ryan-test/index.html
2. Click on one of the L0 labels, the mega menu will open
3. Use the track pad to scroll, the page underneath scrolls intead

## Additional information
See video at: https://ibm.box.com/s/3sb8veyu29tpf8en6dgmnyjrhw0f968e

lily-kenney:
@RobertaJHahn , @jeffchew , @ljcarot - FYI, Elliott identified this as a showstopper for launch. I think we can keep the severity the same but make sure it's pinned as a hot fix?

RobertaJHahn:
@lily-kenney The issue is assigned to the August release.  We should be good.

PR #3676, opened 2020-08-20
fix(MegaMenu): disable body scroll when megamenu is open

annawen1:
### Related Ticket(s)

[MM] Attempting to scroll through the mega menu and the home page scrolls instead #3633

### Description

Disable the body scroll when user has megamenu open on desktop

### Changelog

**Changed**

- when megamenu is expanded, set body overflow-y style to `hidden`

### DIFF
--- a/packages/react/src/components/carbon-components-react/UIShell/HeaderMenu.js
+++ b/packages/react/src/components/carbon-components-react/UIShell/HeaderMenu.js
@@ -19,6 +19,7 @@ import {
   matches,
 } from '../../../internal/vendor/carbon-components-react/internal/keyboard';
 import { AriaLabelPropType } from '../../../internal/vendor/carbon-components-react/prop-types/AriaPropTypes';
+import root from 'window-or-global';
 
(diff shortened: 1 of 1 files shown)


packages/react/src/components/carbon-components-react/UIShell/HeaderMenu.js
```diff
@@ -87,8 +88,13 @@ class HeaderMenu extends React.Component {
    */
   handleOnClick = index => {
     this.setState(prevState => {
-      if (prevState.expanded) this.props.setOverlay(false);
-      else this.props.setOverlay(true);
+      if (prevState.expanded) {
+        this.props.setOverlay(false);
+        root.document.body.style.overflowY = 'auto';
```
asudoh, line 93: Would we be able to define a CSS class for `<body>`? I tend to avoid inline style for the sake of customizability (though I know it's rare for this case).

asudoh (commented):
Nice effort for fixing this @annawen1!

Issue #3454, opened 2020-08-05
Content block aside elements not aligning correctly


larahanlon2:
## Detailed description

**Describe in detail the issue you're having.**
The aside elements for both `content block simple`, `content block media` `content block mixed` and `content block segmented` are not top-aligning as intended. 

Current story:
<img width="1448" alt="Screen Shot 2020-08-05 at 1 35 53 PM" src="https://user-images.githubusercontent.com/53093550/89445605-65cc4580-d721-11ea-905b-a28249a898e5.png">

**Is this a feature request (new component, new icon), a bug, or a general issue?**
Design defect

**Is this issue related to a specific component?**
`content block simple` and `content block segmented` and `content block mixed` and `content block media`

**What did you expect to happen? What happened instead? What would you like to see changed?**
The heading of the `aside elements` title should be top-aligned to the intro copy of the `content block`.
As per design spec:
<img width="997" alt="Screen Shot 2020-08-05 at 1 37 57 PM" src="https://user-images.githubusercontent.com/53093550/89445686-7d0b3300-d721-11ea-95dc-a67fdda89f3b.png">

> What browser are you working in?
Chrome

> What version of the IBM.com Library are you using?
The latest

## Steps to reproduce the issue

https://ibmdotcom-react.mybluemix.net/?path=/story/components-contentblocksegmented--with-aside-elements
## Additional information

https://ibmdotcom-react.mybluemix.net/?path=/story/components-contentblockmixed--with-aside-elements

https://ibmdotcom-react.mybluemix.net/?path=/story/components-contentblockmedia--with-aside-elements

https://ibmdotcom-react.mybluemix.net/?path=/story/components-contentblocksimple--with-aside-elements

- Screenshots or code
- Notes

diegomarcuz:
Waiting a Lara's help, because I have some design questions

diegomarcuz:
@larahanlon2, Could you check again ?? This is the deploy preview: https://ibmdotcom-react.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/3680/index.html

larahanlon2:
Looks great! I don't see any issues now. Thanks @diegomarcuz

diegomarcuz:
Thanks @larahanlon2

PR #3680, opened 2020-08-21
fix(AsideElement): top alignment

diegomarcuz:
### Related Ticket(s)

Content block aside elements not aligning correctly #3454

### Description

Changed a lot of spacing so I can achieve the top position of aside element (Link List component), in other words, it can be aligned with the text right before it and changed other spacing to be according what designers would.

I talked to Lara offline and she said me: 

- When there isn't the stacked Link List, in other words, when Link List isn't below everything should have 160px padding after the card. But on `medium` it should be 96px and smaller resolutions should be 64px. However LinkList starts to be stacked on `large` breakpoint that's why on the code 96px of padding is applied on `large` breakpoint and 64px of padding is applied on `medium` breakpoint, because the `carbon--breakpoint-down` mixin didn't handle `small` breakpoint.

- When there are links that stack at the bottom on smaller breakpoints, it should have the same top and bottom padding as the other designs. 64px top, 96px bottom.

### DIFF
--- a/packages/styles/scss/internal/content-block/_content-block.scss
+++ b/packages/styles/scss/internal/content-block/_content-block.scss
@@ -12,15 +12,17 @@
 
 @mixin content-block {
   .#{$prefix}--content-block {
-    padding-top: $carbon--layout-03;
-    padding-bottom: $carbon--layout-05;
-    @include carbon--breakpoint('md') {
-      padding-top: $carbon--layout-05;
-      padding-bottom: $carbon--layout-06;
+    padding-top: $layout-05;
+    padding-bottom: $layout-07;
+
+    @include carbon--breakpoint-down('lg') {
+      padding-bottom: $layout-03;
+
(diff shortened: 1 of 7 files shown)


packages/styles/scss/internal/content-block/_content-block.scss
```diff
@@ -12,15 +12,25 @@
...
-    @include carbon--breakpoint('md') {
-      padding-top: $carbon--layout-05;
-      padding-bottom: $carbon--layout-06;
+    padding-top: $layout-05;
+    padding-bottom: $layout-07;
+
+    @include carbon--breakpoint-down('lg') {
+      padding-bottom: $layout-03;
+
+      &:not(.#{$prefix}--content-block_has-linkList) {
```
jeffchew, line 21: `has-linkList` might not be the best name as the `aside` can have any type of content.

kennylam:
Updated PR https://github.com/carbon-design-system/ibm-dotcom-library/pull/3845

PR #3690, opened 2020-08-22
test(integration): make page navigation test stable

asudoh:
### Description

This change attempts to stabilize UI integration tests, in the area of testing page navigation, by explicitly watching for the event of completing page navigation.

This change also ensures the latest Playwright version (`1.3.x`) is in use.

### Changelog

**Changed**

- A change that attempts to stabilize UI integration tests, in the area of testing page navigation, by explicitly watching for the event of completing page navigation.
- A change to ensure the latest Playwright version (`1.3.x`) is in use.


.github/workflows/ci-check.yml
```diff
@@ -108,18 +108,22 @@ jobs:
         run: xvfb-run --auto-servernum yarn lerna run --stream --prefix --scope=@carbon/ibmdotcom-react test:integration:ui
         env:
           DDS_UI_INTEGRATION_TEST_BROWSER: ${{ matrix.browser }}
+          TEST_TIMEOUT: 60000
```
jeffchew, line 111: Hey @asudoh , more for my knowledge, where is this environment variable being used?

asudoh: @jeffchew It's [here](https://github.com/carbon-design-system/ibm-dotcom-library/blob/v1.10.0-rc.0/packages/web-components/tests/integration/ui/setup-env.js#L12-L13), meaning the timeout for each tests. Tests esp. ones involving page navigation tends to take longer time to finish.

jeffchew: Thanks @asudoh ! Wondering if we should rename it to be clear that it's for web components integration testing specifically? Probably can do that in another PR though.

asudoh: Actually it's used for React UI integration test, too. That said, I'm open for `DDS_UI_INTEGRATION_TEST_TIMEOUT` (in a separate PR).

jeffchew: Thanks! Yes, I like that name more!

Issue #3698, opened 2020-08-24
Expressive theme mixin "use-carbon-productive-tokens" does not work without CSS custom properties enabled


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

With CSS static builds, non-CSS custom properties tokens appear to be overwritten by the expressive token values when using `use-carbon-productive-tokens`. The way around this is to have `enable-css-custom-properties` set as `true`, but we would hopefully find a way for the theme to work without enabling CSS custom properties.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

Expressive Theme

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Expressive theme mixin `use-carbon-productive-tokens` works in production builds without the use of CSS Custom Properties in Carbon

## Steps to reproduce the issue

1. Go to https://github.com/carbon-design-system/ibm-dotcom-library-nextjs-test
2. Run `yarn upgrade` then `yarn dev`
3. See that the local deployment works by viewing the mega menu (rendering with the productive tokens)
4. Run `yarn build-export`
5. Deploy `./out/index.html` on a local server
6. See that the local static build has expressive tokens applied to the mega menu

## Additional information

Local dev (`yarn dev`):

<img width="1253" alt="Screen Shot 2020-08-24 at 3 07 05 PM" src="https://user-images.githubusercontent.com/1641214/91085506-8ea67300-e61b-11ea-9d61-5e908c13d243.png">

Static deploy (`yarn build-static`):

<img width="1251" alt="Screen Shot 2020-08-24 at 3 05 44 PM" src="https://user-images.githubusercontent.com/1641214/91085540-9a923500-e61b-11ea-9e9c-ade8c530ac2c.png">

- Find out from Carbon team when they plan to release this capability from behind a feature flag.
- This issue is about doing the leg work and then documenting our guidance.  
- Engineering and web site documentation is needed.

RobertaJHahn:
Added to Aug 31 engineering meeting agenda.

jeffchew:
@RobertaJHahn Pushing this one out further.

RobertaJHahn:
Closing.  There have not been any complaints from Adopter teams.  Carbon will be adopting custom properties by default.  No need for this feature.

PR #3699, opened 2020-08-24
fix(expressivetheme): enable css custom properties with the theme

jeffchew:
### Related Ticket(s)

Refs #3698, #3643

### Description

Currently the only way for the theme to work is to enable CSS custom
properties in Carbon.

~This is a test to enable CSS Custom properties by default in the expressive theme.~
This approach does not work, so adding documentation on how to enable by default.

### Changelog

**Changed**

- enable `enable-css-custom-properties` by default

### DIFF
--- a/packages/react/.storybook/webpack.config.js
+++ b/packages/react/.storybook/webpack.config.js
@@ -14,14 +14,6 @@ const rtlcss = require('rtlcss');
 const NODE_ENV = 'development';
 
 /**
- * Flag to enable the expressive theme
- *
- * @type {boolean}
- */
-const useCarbonExpressive =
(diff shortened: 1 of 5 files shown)


jeffchew:
If we're good with this, this is more of a documentation update on instructing adopters on how to implement the expressive theme. The part that we discovered is that CSS custom properties needs to be enabled in order for it to work. We might need to track a different issue (maybe #3698) to figure out how to make the theme work without the use of CSS custom properties. I recall in the past how I was running into roadblocks on this when trying to do this without it enabled, and might be the same issue.

asudoh (approved):
LGTM 👍 - Thanks @jeffchew!

Issue #3661, opened 2020-08-20
Component:- Masthead/Dotcom Shell(Stage)  :- Scroll is not working on mega menu 


chetankan:
**Issue Description :- 1)** On masthead/dotcom shell component in portrait mode, If we open component  in new tab in that case  'View all products' link under 'Products & Solutions'  is not visible, and also scroll is not working.

**2)** On masthead/dotcom shell component in landscape mode, sub menu scrolling is not working, when tries to scroll , back page is getting scrolled.

OS:- iOS 13, 12
Devices :- iPhone 11pro, XS, XR, 8

URL:- https://ibmdotcom-react-staging.mybluemix.net/?path=/story/components-dotcom-shell--default  (Dotcom Shell)
https://ibmdotcom-react-staging.mybluemix.net/?path=/story/components-masthead--default  (Masthead)

#### The following information is consolidated from Olivia's bug(#3646)

[Screen recording](https://ibm.ent.box.com/file/709883372853)
When browsing the mobile mega menu, I cannot see the last menu item `WebSphere` in the Top Products & Platforms category.  (See the video in the link above. )

Testing in the latest canary, seen on Homepage staging environment

Working in the following browsers.  Safari, iOS 13.6, iPhone 7

PR #3704, opened 2020-08-24
fix(masthead): ssr fix for scroll check

jeffchew:
### Related Ticket(s)

#3661 

### Description

An SSR bug was picked up after a fix for #3661, this is adding a check.

### Changelog

**Changed**

- Adding check for `document.body`

### DIFF
--- a/packages/react/src/components/carbon-components-react/UIShell/HeaderMenu.js
+++ b/packages/react/src/components/carbon-components-react/UIShell/HeaderMenu.js
@@ -140,14 +140,8 @@ class HeaderMenu extends React.Component {
   handleOnBlur = event => {
     if (!event.currentTarget.contains(event.relatedTarget)) {
       this.setState({ expanded: false, selectedIndex: null });
-      root.document.body.classList.remove(`${prefix}--body__lock-scroll`);
+      root.document?.body?.classList.remove(`${prefix}--body__lock-scroll`);
     }
-
-    const megamenuItems = [
-      `${prefix}--masthead__megamenu__category-headline`,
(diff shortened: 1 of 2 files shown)


packages/react/src/components/carbon-components-react/UIShell/HeaderMenu.js
```diff
@@ -140,15 +140,9 @@ class HeaderMenu extends React.Component {
   handleOnBlur = event => {
     if (!event.currentTarget.contains(event.relatedTarget)) {
       this.setState({ expanded: false, selectedIndex: null });
-      root.document.body.classList.remove(`${prefix}--body__lock-scroll`);
+      root.document?.body?.classList.remove(`${prefix}--body__lock-scroll`);
     }
 
-    const megamenuItems = [
```
jeffchew, line 146: FYI, I noticed that this was not being used at all so I removed this @annawen1

asudoh: I see, a side note @jeffchew what do you think of enabling [ESLint `no-unused-vars` rule](https://eslint.org/docs/rules/no-unused-vars)?

asudoh (approved):
LGTM 👍 - Thanks @jeffchew!

packages/react/src/components/carbon-components-react/UIShell/HeaderMenu.js
annawen1: yep, thanks for catching that!

jeffchew:
@asudoh agreed on the ESLint rule update, I'm actually surprised it's not enabled (thought it was)...

Issue #3663, opened 2020-08-20
Component: Masthead - On iPhone voice over,  first level of options within the Hamburger are read on second level.


Praveen-sr:
## Detailed description

Device - On iPhone Voice over 

URL - https://ibmdotcom-react-staging.mybluemix.net/?path=/story/components-masthead--default

Issue :- After expanding the hamburger menu - from within the Hamburger, when Second level menu options are available, Screen reader still reads the First level options. This would confuse the screen reader user.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/98a766e5-9837-407c-a88f-5e0158d50c06)

Praveen-sr:
@RaphaelAmadeu  Ram re tested this on staging, issue is not fixed.
After expanding the level 0 menus, the Level One options are displayed. Along with that the Level 0 options are also read by the screen reader but those are not visually present.

RobertaJHahn:
This was checked by Ram.  It is still not working.

PR #3705, opened 2020-08-24
fix(masthead): locked focus while left menu active

RaphaelAmadeu:
### Related Ticket(s)

#3663 

### Description

I've added a focus sentinel so it prevent focus going out of the left menu while expanded.

### DIFF
--- a/packages/react/src/components/Masthead/MastheadLeftNav.js
+++ b/packages/react/src/components/Masthead/MastheadLeftNav.js
@@ -5,10 +5,10 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+import React, { useRef } from 'react';
 import ddsSettings from '@carbon/ibmdotcom-utilities/es/utilities/settings/settings';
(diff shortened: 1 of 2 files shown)


jeffchew:
@RaphaelAmadeu I see a small button in the mobile/tablet menu:

Screen Shot 2020-08-24 at 7 23 46 PM

RaphaelAmadeu:
@jeffchew I've missed a file in my last commit, it should be ok now

jeffchew:
@RaphaelAmadeu this looks good! One thing I noticed though is when going forward or back in the menu levels, it doesn't move focus to the newly selected menu list. If you tab, it goes into there though. But the original bug doesn't appear to be there as far as I can tell!

asudoh:
Hi @jeffchew I think I had a discussion with @kennylam on this one. The consensus is that we apply a similar fix to #3348. I think from the discussion #3348 includes a change to `@carbon/ibmdotcom-utilities` package to extract out non-framework-specific part of #3348.

However as long as it's functional and doesn't cause any regression, I'm fine going with this change.

jeffchew:
@asudoh let's move forward with this then and cut the release.

PR #3706, opened 2020-08-25
chore(dependencies): upgrade autoprefixer

asudoh:
### Related Ticket(s)

Refs #1644.
Refs #3614.

### Description

This change adds `autoprefixer` to `devDependenvies`, with its latest version, as our WebPack configurations explicitly imports it.

Such change seems to ensure that `autoprefixer` covers iOS Safari for `text-size-adjust`.

### Changelog

**Changed**

- Adds `autoprefixer` to `devDependenvies`, with its latest version, as our WebPack configurations explicitly imports it.


jeffchew:
@asudoh thanks for this fix! It looks like the offline mirror is missing the file for `autoprefixer`.

asudoh:
Good catch @jeffchew - Fixed.

asudoh:
Thanks @jeffchew - I think the other PR needs to be rectified (removing prefixed style rules), but once it's done it'll be good to go.

PR #3723, opened 2020-08-25
fix(Masthead): keep focus in sub menus

RaphaelAmadeu:
### Related Ticket(s)

#3663

### Description

Added a sentinel in order to keep focus on the currently selected sub menu


annawen1:
@RaphaelAmadeu it works nicely for the second level menu items! Nice! One thing I noticed is we get the issue happening again if we go to the third level of menu items

Issue #3731, opened 2020-08-26
[HP] [MM] multiple menus open at once


oliviaflory:
## Detailed description

> Describe in detail the issue you're having.

Click My IBM (don’t close)
Click any mega menu link (don’t close)
Then click any other mega menu link

My IBM icon maintains focus (known issue: #3617 ) if user clicks and doesn't select anything in the dropdown menu, then clicks on any mega menu item.

If the user clicks a second, or third mega menu they will see multiple menus overlaid on top of each other

**My IBM focus, single menu**
<img width="1572" alt="Screen Shot 2020-08-26 at 9 58 41 AM" src="https://user-images.githubusercontent.com/45692622/91313662-9f71f880-e783-11ea-8f32-7c0fc0521353.png">

**Double menu**
<img width="1572" alt="Screen Shot 2020-08-26 at 9 42 55 AM" src="https://user-images.githubusercontent.com/45692622/91313693-a7ca3380-e783-11ea-93e0-4f2d18a1f17e.png">

**Triple menu**
<img width="1573" alt="Screen Shot 2020-08-26 at 9 56 38 AM" src="https://user-images.githubusercontent.com/45692622/91313710-ae58ab00-e783-11ea-8bfe-6a736955a816.png">

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

bug

> Is this issue related to a specific component?

mega menu

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Only one menu should be open at a time

> What browser are you working in?

Chrome
Version 85.0.4183.83 (Official Build) (64-bit)

Safari
Version 13.1.1 (15609.2.9.1.2)

> What version of the IBM.com Library are you using?

Latest release that was integrated with Homepage (Aug 26)

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. Click My IBM (don’t close)
2. Click any mega menu link (don’t close)
3. Then click any other mega menu link
4. Click another mega menu

## Additional information

- Screenshots or code
- Notes

RobertaJHahn:
@annawen1 This is the new bug created by Olivia.  I think she made you aware of it.  I don't have a severity yet, but it seems important so I am moving into the current sprint.

RobertaJHahn:
We will identify the root cause and then patch with dates TBD.

RobertaJHahn:
[...]

Issue #3617, opened 2020-08-18
[MM] My IBM maintaining focus when switching to mega menu


oliviaflory:
## Detailed description

> Describe in detail the issue you're having.

If I open the My IBM dropdown, then open one of the mega menus, the My IBM icon does not lose it's focus state.

![MM My IBM focus Aug-18-2020 17-16-37](https://user-images.githubusercontent.com/45692622/90566544-ab9ff980-e176-11ea-901d-08e07b36daa7.gif)

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

Masthead + mega menu

> What did you expect to happen? What happened instead? What would you like to
> see changed?

I would have assumed clicking on one of the mega menu labels would deselect the My IBM dropdown and it would no longer have an activated focus state

> What browser are you working in?

Chrome
Version 84.0.4147.125 (Official Build) (64-bit)
MacBook pro, osCatalina 10.15.5

> What version of the IBM.com Library are you using?

Latest canary, seen on homepage staging and in storybook

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. Click on My IBM icon in masthead (do not select anything else)
2. Click on mega menu
3. See focus on My IBM and mega menu is open
4. etc.

## Additional information

![Screen Shot 2020-08-18 at 5.03.39 PM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/18c9a6b4-9c10-448d-845e-fb55cc3e64be)

RobertaJHahn:
Praveen and Chetan tested this bug fix.  Looks good.  Can be closed.

PR #3751, opened 2020-08-26
fix(MegaMenu): explicity set focus to HeaderMenu L0 item when clicked

annawen1:
### Related Ticket(s)

[MM] My IBM maintaining focus when switching to mega menu #3617
[HP] [MM] multiple menus open at once #3731

### Description

The `<OverflowMenu />` component that we are using for the profile menu has logic to trap the focus when user clicks outside of the component. This affects the megamenu when user opens the profile menu, then clicks on the L0 nav item. The focus is trapped in the `<OverflowMenu />` which affects the `onBlur` event triggered used when opening another L0 nav item.

This PR forces focus on the L0 nav item when it is clicked.

**BEFORE:**
Aug-26-2020 16-17-05

**AFTER:**
Aug-26-2020 16-19-33

### Changelog

**Changed**

- force focus on the `HeaderMenu` L0 item


packages/react/src/components/carbon-components-react/UIShell/HeaderMenu.js
```diff
@@ -125,6 +126,7 @@ class HeaderMenu extends React.Component {
       `${prefix}--masthead__megamenu__category-headline`,
       `${prefix}--masthead__megamenu__category-group`,
       `${prefix}--masthead__megamenu__view-all-cta`,
+      `${prefix}--masthead__megamenu__l0-nav`,
```
annawen1, line 129: adding this here so when user clicks on another L0 menu item, the overlay doesn't flash

jeffchew:
@annawen1 Looks good! 

I'm thinking through this, and we are already set to go as-is with what we have currently tonight, so I'm thinking of having this retargeted to a release branch for a patch release instead, so let's hold on merging this for now.

jeffchew:
@annawen1 one other thought I have, perhaps we can target `master` for now so we can test around this on `canary`, and if we do a patch we can always cherry pick this commit. Thoughts?

annawen1:
> @annawen1 one other thought I have, perhaps we can target `master` for now so we can test around this on `canary`, and if we do a patch we can always cherry pick this commit. Thoughts?

@jeffchew yeah sure that works

PR #3755, opened 2020-08-26
fix(lang): fix to lang/country fetch from DDO

jeffchew:
### Related Ticket(s)

https://github.com/carbon-design-system/ibm-dotcom-library/issues/3754

### Description

This is a patch to the way that the lang/locale is fetched, as currently it is always returning as false.

### Changelog

**Changed**

- Updated logic for fetching lang/locale in LocaleAPI

### DIFF
--- a/packages/react/.storybook/preview-head.html
+++ b/packages/react/.storybook/preview-head.html
@@ -133,6 +133,29 @@
 <link rel="alternate" hreflang="en-zm" href="https://www.ibm.com/zm-en">
 <link rel="alternate" hreflang="en-zw" href="https://www.ibm.com/zw-en">
 
+<script>
+  digitalData = {
+    "page": {
+      "category": [],
+      "pageInfo": {
+        "language": "en-US",
+        "publisher": "IBM Corporation",
+        "version": "IBM.com Library",
+        "ibm": {
+          "country": "US",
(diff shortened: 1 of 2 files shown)


asudoh (commented):
Looks basically good, but looks like we are changing our core logic dealing with DDO. Should we have some new test cases to cover the change? Thanks @jeffchew!

jeffchew:
> Looks basically good, but looks like we are changing our core logic dealing with DDO. Should we have some new test cases to cover the change? Thanks @jeffchew!

Good call! I did notice that a bunch of other parts of the code are not covered either (and isn't merged with `master`) so I'm thinking of perhaps either covering for what was changed, or wait to add these tests after merging down to master.

jeffchew:
FYI, going to hold on merging this to `v1.10.0` for now, will see if we want to line up a `v1.10.1` patch release.

asudoh:
> I'm thinking of perhaps either covering for what was changed

I think this is reasonable at this point.

And yeah I'd be hesitant to merge this in the release given the potential risk I described earlier.

jeffchew:
I'm going to actually close this PR for now, as configuring the DDO on the application side seemed to have worked. I might do a separate PR for elaborating on how to configure the DDO to set the language/locale correctly.

PR #3768, opened 2020-08-28
fix(locale-search): reset query on page navigation

asudoh:
Refs #3769 

### Description

This change resets the search query in locale selector upon being navigated from region selector.

### Changelog

**Changed**

- A change to reset the search query in locale selector upon being navigated from region selector.


jeffchew:
@asudoh created https://github.com/carbon-design-system/ibm-dotcom-library/issues/3769 for this bug so we can track it.

PR #3830, opened 2020-09-01
feat(Icon): icon web component

annawen1:
### Related Ticket(s)

Web Component: Develop Icon (internal) of the React version #3538

### Description

Web component version of `<Icon />` or `<IBMLogo />`

We don't have a storybook story for `<Icon />`, but it is used within the `<Masthead />`

### Changelog

**New**

- `icon.ts` file
- `masthead-logo.ts` extends the icon


annawen1:
I believe this is an internal component, so we aren't set up any storybook stories for `<Icon />`

asudoh (approved):
LGTM basically, thanks @annawen1! One note for future enhancement, especially given we are making the masthead logo logic general-purpose, is sharing features with other components with link features, e.g. `<dds-link-with-icon>`, `<dds-card>`. They inherit [an abstract component](https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/web-components/src/globals/internal/link.ts). Thanks!

Issue #3473, opened 2020-08-06
Remove web-component specific styles for Card Link


jeffchew:
#### User Story

> As a `[user role below]`:
IBM.com Library developer

> I need to:
ensure all styles for Card Link live in the styles package

> so that I can:
manage styles for the Card Link regardless of framework

#### Additional information

 - Current card link styles in `web-components`: https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/web-components/src/components/card-link/card-link.scss
 - The way card link styles are structured (with nesting) it is preventing a cleaner inclusion of the styles package styles into web components

#### Acceptance criteria
 - [ ] No more custom styles for `dds-card-link` in `web-components`

PR #3835, opened 2020-09-01
fix(CardLink): remove specific styles

ariellalgilmore:
### Related Ticket(s)

#3473 

### Description

remove specific styles in web components

### Changelog

**Removed**

- web component styles


asudoh (approved):

Update: Great if you can take a look at Percy diff for React, just in case it may be a true positive. Thanks!

packages/styles/scss/components/card/index.scss
```diff
@@ -44,6 +44,7 @@
       @include carbon--type-style('body-short-02');
 
       margin-bottom: $carbon--spacing-07;
+      flex: 1;
```
kennylam, line 47: @ariellalgilmore Looks like this is causing the issue reported by @jeffchew

ariellalgilmore: @kennylam @jeffchew fixed!

asudoh:
@jeffchew Seems that the visual regression has been fixed. Does this look good to go? Thanks!

jeffchew (approved):
LGTM, thank you for the fixes!

Issue #3838, opened 2020-09-01
404 redirect when switching locale preference


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

When switching the locale/language preference from us-en to another combination, then switching back to us-en, a 404 error occurs when clicking the IBM logo:

![ipcinfo-bug](https://user-images.githubusercontent.com/1641214/91913475-b7fe7900-ec83-11ea-9ddf-fc78721459b8.gif)

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

bug

> Is this issue related to a specific component?

locale modal

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Do not trigger the 404 redirect when clicking the IBM logo

> What browser are you working in?

Firefox

> What version of the IBM.com Library are you using?

v1.10.1

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

US Homepage

## Steps to reproduce the issue

1. Go to www.ibm.com
2. Scroll down to the locale modal and select another region/country
3. Scroll down to the v18 locale selector and switch back to us-en
4. Click the IBM logo at the top

## Additional information

- This appears to be related to how the ipcinfo cookie is getting set, as there is a very minor difference between the two on how they are being set. v18 sets the domain as `.ibm.com`, and the IBM.com library sets it as `www.ibm.com`.

PR #3839, opened 2020-09-01
fix(ipcinfo): setting domain for the ipcinfo cookie

jeffchew:
### Related Ticket(s)

Refs #3838

### Description

There was a discrepency identified where the cookie was being set under
`www.ibm.com` domain on the IBM.com homepage, whereas v18 was setting it
 under `.ibm.com` on all other pages. When returning to the homepage and
  clicking the logo, this caused the server-side level redirect rule to
  kick in and create a garbled homepage url and causing a 404 to occur.

### Changelog

**Changed**

- Set the explicit domain when setting `ipcinfo` cookie


asudoh (commented):
LGTM 👍, but wanted to have another eyes on this given I don't have enough background on the requirements in this area. Thanks @jeffchew!

Issue #3833, opened 2020-09-01
Search results in locale modal does not load for Web Components in bare project


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

The locale modal loads in the `dotcom-shell-container`, but no results are loading in:

![Screen Shot 2020-09-01 at 2.53.04 PM.png](https://images.zenhubusercontent.com/5cf80f82b494896be7bc635b/e16a0c59-54f3-41cf-a49d-828aaf9ab3c8)

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

bug

> Is this issue related to a specific component?

locale modal (web components)

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Load the search results

> What version of the IBM.com Library are you using?

Canary

## Steps to reproduce the issue

1. Load the webpack/handlebars sample application: https://github.com/jeffchew/ibmdotcom-web-components-test
2. `yarn install` and `yarn start`
3. Click the locale button in the footer

jeffchew:
@asudoh would you be able to look at this?

cc: @RobertaJHahn

PR #3841, opened 2020-09-02
chore(locale-modal): warn missing links for alternate language

asudoh:
### Related Ticket(s)

Refs #3833.

### Description

Adds a logic to emit a console warning when `<link rel="alternate">` for alternate language is missing, that yeilds to an empty result in `<dds-locale-search>`.

### Changelog

**New**

- A logic to emit a console warning when `<link rel="alternate">` for alternate language is missing.


jeffchew:
Good call, thank you @asudoh !

Issue #3851, opened 2020-09-02
Masthead is not dynamically updating `state` param in myibm login url


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

The login url includes a `state` param that temporarily sets it as `{{window.location}}`, which assumes the client-side data will replace with actual `window.location` before setting the URL. This is not currently happening with the IBM.com Library. The resulting page is:

![image (7)](https://user-images.githubusercontent.com/1641214/92043148-8c45c680-ed49-11ea-902c-d75b1e7a3101.png)

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

Masthead (React) (assume we need to patch this for web components as well)

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Replace `state=https%3A%2F%2Fwww.ibm.com` with `state=[actual window.location value]`

> What version of the IBM.com Library are you using?

v1.10.1

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

US Homepage

## Steps to reproduce the issue

1. Go to https://www.ibm.com/us-en
2. Click the profile icon and choose "Login"
3. Log in via w3id
4. See error page

## Additional information

- The value before was set as `state={{window.location}}`. A patch was made to hardcode to `state=https%3A%2F%2Fwww.ibm.com%2Fus-en`
- We are deciding to keep the patched data as-is, but now looking for that specific hardcoded string instead to replace with actual `window.location`

PR #3852, opened 2020-09-03
fix(translation): adding string/replace of login url state param

jeffchew:
### Related Ticket(s)

Refs #3851 

### Description

The original implementation of the signin url for ibm included a string
replace of `{{window.location}}`, which was not accounted for in the IBM
.com Library. This adds a transformation logic now to replace the
default value in the data (state=https%3A%2F%2Fwww.ibm.com) and replace
with `state=[the actual window.location value]` in the service before
reaching the presentation layer.

### Changelog

**Changed**

- transformation logic in TranslationAPI to dynamically update the login url

### DIFF
--- a/packages/services/src/services/Translation/Translation.js
+++ b/packages/services/src/services/Translation/Translation.js
@@ -202,6 +202,18 @@ class TranslationAPI {
    * @returns {object} Translation data
    */
   static transformData(data) {
+    const signedout = data.profileMenu?.signedout;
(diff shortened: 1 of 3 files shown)


asudoh (commented):
I think it looks good, wanted to have another eyes before we go forward. Thanks @jeffchew!

kennylam (approved):
Looks good!

PR #3863, opened 2020-09-04
chore(components): add extensions to dependency imports

asudoh:
### Description

Adds `.js` extension to `import` statement from dependencies. This is another step to make Web Components work bettern in an environment without a toolstack using CDN that supports ESM.

### Changelog

**Changed**

- A series of changes to add `.js` extension to `import` statement from dependencies.


jeffchew (approved):
LGTM, thank you @asudoh !

PR #3864, opened 2020-09-04
test(integration-ui): add browser log

asudoh:
### Description

Temporarily adds browser's debug log in UI integration test, given we still have noticable failure rate. Though the failures won't block our CI, the high failure rate has made us ignore true positives that the test has captured. Good to have the test report only the true positives so it's obvious that we should act on.

### Changelog

**Changed**

- A change to emit browser debug log in UI integration test.


asudoh:
Withdrawing - Unfortunately the debug setting itself causes test failures... 😢

PR #3870, opened 2020-09-08
feat(content-block): introduction

asudoh:
### Related Ticket(s)

Refs #3640.
Refs #3734.
Refs #3792.

### Description

Introduces the following components:

* `<dds-content-block-simple>`: The simple version of content block
* `<dds-content-item>`: An item in content block
* `<dds-content-block-heading>`: The heading portion in content block
* `<dds-content-block-complementary>`: The complementary (aside) portion in content block
* `<dds-content-item-heading>`: The heading portion in content item

Also:

* Introduces `@mixin dds--make-col($span, $columns)` so the layout style can be reused in Sass level. This is done because of the following reasons:
  * `<dds-content-block-simple>` needs only simple grid layout for its complementary portion, but our layout component has much more than that
  * To drift away from using CSS class names indicating its style, which tend to require changing CSS class name in markup when we only want to change the style while we'd want to be able to do it with CSS change only
* A change to drift away from BEM concatenation in `_content-block.scss`, `_content-item.scss` so they can be reused between React and Web Components
* Locked `snowpack` version to `2.10.1`, given `2.10.2` seems to have introduced a regression in Sass `@import`s (Looked like treating Sass `@import` as if it's a JavaScript import)
* Adds non-autoplay support in `<dds-video-player>` and `<dds-video-player-composite>`
* Style fixes when `<dds-video-player>` is used stand-alone

### Changelog

**New**

* `<dds-content-block-simple>`: The simple version of content block
* `<dds-content-item>`: An item in content block
* `<dds-content-block-heading>`: The heading portion in content block
* `<dds-content-block-complementary>`: The complementary (aside) portion in content block
* `<dds-content-item-heading>`: The heading portion in content item
* `@mixin dds--make-col($span, $columns)` so the layout style can be reused in Sass level.
* Non-autoplay support in `<dds-video-player>` and `<dds-video-player-composite>`

**Changed**

* A change to drift away from BEM concatenation in `_content-block.scss`, `_content-item.scss` so they can be reused between React and Web Components
* Locked `snowpack` version to `2.10.1`
* Style fixes when `<dds-video-player>` is used stand-alone


jeffchew:
Thank you @asudoh for the PR!

I noticed that `<dds-content-block/>` was also introduced which wasn't in the notes above. 

One thing I noticed is that the video when embedded looks a little odd:

Screen Shot 2020-09-08 at 3 27 15 PM

packages/web-components/src/components/content-block/content-block.ts
```diff
@@ -0,0 +1,158 @@
...
+  REGULAR = 'regular',
+
+  /**
+   * A style with border.
+   */
+  WITH_BORDER = 'with-border',
+}
+
+/**
+ * Content block.
```
jeffchew, line 45: Super minor thing (and also something to consider for the other ones that fall under this category) but wondering if we want to add any documentation that this and others are internal components, which is why we have not included storybook entries for them?

asudoh: Good point @jeffchew - I was thinking of marking this class as `abstract`, will try this. BTW there is no `<dds-content-block>`, because `<dds-content-block-simple>`, etc. inherits `DDSContentBlock` class instead of using `<dds-content-block>` by a composition.

asudoh:
And good catch on the aspect ratio of the video @jeffchew, fixed.

packages/web-components/src/components/content-block/content-block.ts
asudoh: Updated, by adding `@abstract` JSDoc comment.

jeffchew:
@asudoh looks like snapshots are out of date, can you check?

asudoh:
@jeffchew Fixed!

jeffchew (approved):
Thank you @asudoh !

PR #3880, opened 2020-09-09
chore(components): add extensions to local imports

asudoh:
### Description

Adds `.js` extension to `import` statement from local package, as another step to make Web Components work bettern in an environment without a toolstack using CDN that supports ESM.

This is done to build artifacts only given TypeScript does not allow adding `.ts` extension to `import`s.

Also, as part of the effort, introduces an ESLint plugin to ensure that there is no `import` statement pointing to a directory name to load `index.js` file.

### Changelog

**New**

- A ESLint plugin to ensure that there is no `import` statement pointing to a directory name to load `index.js` file.

**Changed**

- A series of changes to add `.js` extension to `import` statement from local package.


jeffchew (approved):
Thank you for this addition!

Issue #3527, opened 2020-08-12
Web Component: Footer - locale modal - keyboard Focus on the static text (heading) must be removed.


Praveen-sr:
## Detailed description

Issue :- 
On Web components :
Locale modal - Focus on the static text (heading) i.e "Americas" must be removed.
 
## Additional information

Also an issue on the React version.  This work is managed under issue (#3580)

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/8fd9456a-7b2b-4ef4-8706-e66646dd0453)

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/86160b71-67d1-4613-bd21-9593cf694073)

PR #3888, opened 2020-09-09
fix(LocaleModal): remove header focus

kennylam:
### Related Ticket(s)

#3527

### Description

Removes focus from `LocaleModal` header in React and web component packages.

### Changelog

**New**

- `focus` styles for `LocaleModal` header

**Changed**

- break out concat scss into their own classes


jeffchew (approved):
LGTM!

Issue #3711, opened 2020-08-25
Component: Callout Quote - The link is not in single line, its wrapped.


Praveen-sr:
## Detailed description

URL: https://ibmdotcom-react-staging.mybluemix.net/

On iPhone, iPad --  Safari & chrome browsers

Issue: The link is not in single line, its wrapped.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/a2a7f9bd-ca2d-4205-8e34-89daa2b88db1)

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/6e1dc386-da70-4c34-bdf6-e589e54026e3)

PR #3891, opened 2020-09-09
fix(calloutQuote): remove footer styles

ariellalgilmore:
### Related Ticket(s)

#3711 

### Description

remove styles so footer link doesn't wrap

### Changelog

**Removed**

- remove footer styles


jeffchew (approved):
LGTM!

PR #3892, opened 2020-09-09
fix(HorizontalRule): change style prop name to type

annawen1:
### Related Ticket(s)

### Description

Changing the `style` prop name to `type` to align with the web components' horizontal-rule component.

When using `style` in the horizontal-rule web component, an error was thrown as `style` is a reserved attribute name.

### Changelog

**Changed**

- Change the `style` prop name to `type` instead
- Adjust the grid in the horizontal rule react storybook


asudoh (commented):
Nice change! BTW this change requires exiting applications `<HorizontalRule>` change the props. Would it be OK for us? I'm not fully sure what our rule of breaking API change policy is for minor release, so just wanted to double-check. Thanks @annawen1!

annawen1:
> Nice change! BTW this change requires exiting applications `<HorizontalRule>` change the props. Would it be OK for us? I'm not fully sure what our rule of breaking API change policy is for minor release, so just wanted to double-check. Thanks @annawen1!

Yeah I thought about that, I'm not sure how many folks are using the `<HorizontalRule />`. @jeffchew would it be enough to announce it in the release notes or should we keep it as is for now and add a deprecate note?

jeffchew:
We should add in a note in our release notes.

jeffchew (approved):

Issue #3894, opened 2020-09-10
Web component: Masthead - search bar does not show autocomplete suggestions.


Praveen-sr:
## Detailed description

Issue: Masthead - On search bar, when user enters a keyword it does not show autocomplete suggestions.

Note: This scenario is working fine on react version.

## Additional information

Actual :
![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/7d0e2fad-b514-4488-9943-50395148618e)

Expected:
![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/c4dfc6e1-8b17-4167-8e72-6aab272e26a0)

jeffchew:
@asudoh would you be able to take a look?

PR #3900, opened 2020-09-10
fix(masthead-search): fire custom event

asudoh:
### Related Ticket(s)

Refs #3894.

### Description

This change ensures that a custom event is fired when user types in search box as well as when user clicks on "clear search" button.

The event receiver code expects such custom event to be fired, and search API call won't happen without it.

### Changelog

**Changed**

- A change to ensure that a custom event is fired when user types in search box as well as when user clicks on "clear search" button.


jeffchew (approved):
Thank you @asudoh !

kyjak:
I could not find documentation on usage of custom events, can you point me please to an example/docs? thx

asudoh:
Good point @kyjak - Created: #4380

PR #3904, opened 2020-09-10
chore(package): add exports field

asudoh:
### Description

Adds `exports` fields to `services` and `utilities` packages as another step to make Web Components work bettern in an environment without a toolstack using CDN that supports ESM.

This is a similar change to:
carbon-design-system/carbon-web-components#443

### Changelog

**New**

- `exports` field to `services` and `utilities` packages.


jeffchew (approved):
Thank you @asudoh !

jeffchew:
@asudoh probably not for this PR, we should add in an ignore rule for percy snapshots around the video. I can take a look at this.

Screen Shot 2020-09-10 at 11 22 05 AM

jeffchew (approved):
Thank you @asudoh !

asudoh:
> we should add in an ignore rule for percy snapshots around the video. I can take a look at this.

Agreed - Thank you so much @jeffchew for taking this on!

PR #3905, opened 2020-09-10
chore(percy): suppressing dds-video-player-container in percy

jeffchew:
### Related Ticket(s)

No related issue

### Description

The video player is causing false positives in Percy, as the timing of
the video loading is an uncertainty when uploading the snapshots to
Percy servers for processing.

### Changelog

**Changed**

- Suppressing video player in percy snapshot testing


asudoh (approved):
LGTM 👍 - Thanks @jeffchew!

Issue #3907, opened 2020-09-10
Masthead L0 active link: hamburger menu active should be semibold


larahanlon2:
## Detailed description

**Describe in detail the issue you're having.**
When the hamburger menu is open, the active link in the menu list is the same font-weight as all other items but it should be semibold.

<img width="943" alt="Screen Shot 2020-09-10 at 1 34 53 PM" src="https://user-images.githubusercontent.com/53093550/92774319-5a48eb80-f36b-11ea-8235-9e47e8364764.png">

**Is this a feature request (new component, new icon), a bug, or a general issue?**
Design defect

**Is this issue related to a specific component?**
Masthead L0

**What did you expect to happen? What happened instead? What would you like to see changed?**

The blue sidebar/indicator is appearing correctly. Only need to change the font-weight. 

As per design specs:
<img width="396" alt="Screen Shot 2020-09-10 at 1 38 34 PM" src="https://user-images.githubusercontent.com/53093550/92774356-646aea00-f36b-11ea-9086-8d8fe341ae5c.png">

**What browser are you working in?**
Chrome

## Steps to reproduce the issue

1. Open canary version of masthead https://ibmdotcom-react-canary.mybluemix.net/?path=/story/components-masthead--default
2. Open in a new tab outside Storybook
3. Resize browser to activate the hamburger menu
4. Open menu

PR #3909, opened 2020-09-10
fix(Masthead): set selected menu item font weight to semibold

annawen1:
### Related Ticket(s)

Masthead L0 active link: hamburger menu active should be semibold #3907

### Description

When the hamburger menu is open, the active link in the menu list is the same font-weight as all other items but it should be semibold.
Screen Shot 2020-09-10 at 1 34 53 PM

As per design specs:
Screen Shot 2020-09-10 at 1 38 34 PM

### Changelog

**Changed**

- set selected menu item font-weight to semibold


larahanlon2 (approved):
Awesome, this looks perfect.

Issue #3899, opened 2020-09-10
dd-dotcom-shell-container not refecting the setAttribute("language", "ja-JP") 


tefiggins:
## Title:
**DEFECT:**  dd-dotcom-shell-container **not reflecting** the setAttribute("language", "ja-JP")  

## Detailed description
Web components are designed to be interacted with via their attributes, this way they can stay truly interoperable.
If you get your data after you have added the component to the DOM you can still interact with the web component via attributes like this:

Let's say you have your dialog component already added to the DOM:
```
<my-dialog id='mydialog1' heading="A Dialog">Lorem ipsum</my-dialog>
```
So you can easily update this DOM elements attributes at a later time with JavaScript:
```
document.querySelector('#mydialog1').setAttribute('init-data', JSON.stringify(obj));
```
**In my particular case**, I am using your  dotcom-shell-container  web component with footer-size set in my index.html file and desire to set the _language_ attribute later in the code after interrogating the IPCInfo cookie for locale set before entering my application form the referring application in the loosely coupled web services platform call IBM Marketplace. 

```
	<dds-dotcom-shell-container footer-size='short'>
		<div role="main">
			<div id="app"></div>
			<!-- built files will be auto injected -->
		</div>
	</dds-dotcom-shell-container>
```

So I attempt:
```	
const languageCode = lang ? lang.substring(0, 2) : "en";
const countryCode = lang ? lang.substr(3, 2) : "US";

console.log (languageCode + " and " + countryCode)
document.getElementsByTagName("html")[0].setAttribute("lang", lang);
digitalData.page.pageInfo.language = lang;
digitalData.page.pageInfo.ibm.country = countryCode;
				
document.getElementsByTagName('dds-dotcom-shell-container')[0].setAttribute("language", lang);
// NOT WORKING -- want to re-init container -- like changePageLocale did in v18
```
> The **setAttribute actually adds the language attribute** to the item in the DOM (can see with inspect browser tool)

> **However** dds-dotcom-shell-container REMAIN in the same instantiated language (**not the newly set language**); the component should be re-inited with new language 

> **This breaks the usage that is allowed with v18 that Web Components is replacing**:   Breaks the ability to update set header/footer language via setting from other sections of a loosely coupling web services architecture application on IBM Marketplace. Don't want Customer to have to reset the language as they move around the Marketplace.

[...]

PR #3915, opened 2020-09-11
feat(masthead): support for dynamic language switch

asudoh:
### Related Ticket(s)

Refs #3899.

### Description

Introduces code to load language display data as well as translation data with the new language when `language` property is changed in `<dds-masthead-container>`, `<dds-footer-container>` or `<dds-locale-modal-container>`.

To make it work, did the following changes:

* Add support specifying a specific language in `loadLangDisplay()`, `loadLocaleList()` and `loadTranslation()` Redux actions.
* Make Redux data for display language managed per language.
* Unsupport custom links data from `<dds-masthead-container>`, `<dds-footer-container>` and `<dds-locale-modal-container>`. For such use case, `-composite` counterparts should be used. Tests are changed to reflect such change.

Also, eliminated duplicate type definitions found during the effort.

dynamic-lang-switch

### Changelog

**New**

- Code to load language display data as well as translation data with the new language when `language` property is changed in `<dds-masthead-container>`, `<dds-footer-container>` or `<dds-locale-modal-container>`.
- Support specifying a specific language in `loadLangDisplay()`, `loadLocaleList()` and `loadTranslation()` Redux actions.

**Changed**

- A change to make Redux data for display language managed per language.

**Removed**

- Support for custom links data from `<dds-masthead-container>`, `<dds-footer-container>` and `<dds-locale-modal-container>`. For such use case, `-composite` counterparts should be used. Tests are changed to reflect such change.


jeffchew (approved):
Nice change, and thank you for getting this in right before code freeze!

jeffchew:
@asudoh Just had a thought, this might affect the codesandbox examples since links is no longer supported, right?

asudoh:
@jeffchew Here's what will happen as far as I know; In the environment where we explicitly set e.g. `.navLinks`, the API calls to get the corresponding data fails. And thus nothing will override the `.navLinks` we explicitly set and thus it still works.

jeffchew:
> @jeffchew Here's what will happen as far as I know; In the environment where we explicitly set e.g. `.navLinks`, the API calls to get the corresponding data fails. And thus nothing will override the `.navLinks` we explicitly set and thus it still works.

Sounds good! Still seeing unrelated issues around codesandbox so will ping you offline on that.

PR #3916, opened 2020-09-11
fix(card-footer): fix duplicate content

asudoh:
### Description

This change fixes duplicate content shown in `<dds-card-cta-footer>`, was noticed in its video mode, showing video duration as `1:001:00` where it should have been `1:00`.

The change removes `_renderContent()` from the `render()`. The super class (`<dds-link-with-icon>`)'s `_renderInner()` takes care of running `_renderContent()`.

### Changelog

**Changed**

- A fix for duplicate content in `<dds-card-cta-footer>`.


jeffchew (approved):
Good catch!

Issue #3917, opened 2020-09-11
Card icon placement


ariellalgilmore:
## Detailed description

> Describe in detail the issue you're having.

For web components the card icon placement only has the option to be on the left. There should be an option to have the icon be on the right. Also, the icon does not change positions in react version.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

Card

PR #3919, opened 2020-09-11
fix(Card): fix icon placement option and styles

ariellalgilmore:
### Related Ticket(s)

#3917

### Description

Add icon placement option for footer, so that it can be set to left or right. Fix styles so react version works too.

### Changelog

**New**

- add icon placement option

**Changed**

- fix styles


asudoh (approved):

annawen1 (approved):

Issue #3730, opened 2020-08-26
Component: Mega Menu - scrolling issue on second level menu options


Praveen-sr:
## Detailed description

On iPhone - Safari & chrome

URL : https://ibmdotcom-react-staging.mybluemix.net/?path=/story/components-masthead--default

Issue : when user clicks on level 1 option and scrolls down the options and goes to level 2, user is navigated to end of level 2 options. 

## Steps to reproduce the issue

1. Open Mega menu on Mobile and click on hamburger menu
2. Click on Product& Supply and scroll down  
3. Click on supply chain option 
4. scroll is at the bottom of the page.

## Additional information

![Megamenu2.gif](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/6cf4258c-dfd4-4aac-9d24-32958cf6718c)

RobertaJHahn:
@annawen1 This is the bug we talked about in the meeting with Praveen and Chetan earlier today.

cc:  @jeffchew @ljcarot

PR #3921, opened 2020-09-11
fix(MastheadLeftNav): when user goes through menu levels, set scrollTop to 0

annawen1:
### Related Ticket(s)

Component: Mega Menu - scrolling issue on second level menu options #3730

### Description

When user scrolls in the mobile masthead menu, clicks on an option, the scroll position remains the same when user should be brought to the top of the next menu level.

This PR resets the scroll position every time the user clicks on an menu option

### Changelog

**Changed**

- set `bx--side-nav__items` scrollTop to 0 on click


packages/react/src/components/carbon-components-react/UIShell/SideNavMenu.js
```diff
@@ -108,9 +108,21 @@ export class SideNavMenu extends React.Component {
     };
   }
 
+  scrollToTop = () => {
+    const navItemsElem = root.document.getElementsByClassName(
+      'bx--side-nav__items'
+    );
```
asudoh, line 114: It'll be nicer if we can better scope the selection of elements (e.g. only the menu items associated with this menu, or go up to the ancestor `.bx--side-nav` and start from there), but going as-is looks OK for now.

annawen1: hmm there's actually only one `bx--side-nav__items` element from `SideNavItems`, it wraps the entire menu. Would it be better to create a ref for the `SideNavMenu` and get the `bx--side-nav__items` from the ref instead?

jeffchew: It's certainly the "React" way of doing it...

annawen1: changed to target the `side-nav__items` element from the ref instead

jeffchew (approved):
LGTM!

Issue #3883, opened 2020-09-09
Web component: Masthead - On empty search bar, clicking on search icon,  page navigates to search results.


Praveen-sr:
## Detailed description

Issue: On empty search bar, user clicking on search icon, refreshes the page and navigates to search results.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/b3d12416-f182-43e9-add0-b6ac331aba71)

Landing page:
![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/d1232021-545e-48c6-8b5e-30680da20e3b)

PR #3932, opened 2020-09-14
fix(masthead-search): no redirect on empty value

asudoh:
### Related Ticket(s)

Refs #3883.

### Description

This change ensures redirect to dedicated search page upon clicking on
the loupe icon happens only when there is a search query in the input
box.

### Changelog

**Changed**

- A change to ensure redirect to dedicated search page upon clicking on
the loupe icon happens only when there is a search query in the input
box.


jeffchew (approved):
Thank you, good add!

Issue #3897, opened 2020-09-10
Web component: Masthead - Skip to main content link is missing from the page.


Praveen-sr:
## Detailed description

Issue: On Keyboard tab/shift tab navigation, Skip to main content link is missing from the page.

Expected: Skip to main content link should be available

## Additional information

Actual 
![Skip2MainCon.gif](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/5c98e70b-0611-43cf-8307-8d546463e023)

Expected:
![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/9ef1410e-bb4e-4e32-8ac7-154bc8c76abb)

PR #3937, opened 2020-09-14
chore(devenv): introduce skip-to-content link

asudoh:
### Related Ticket(s)

Refs #3897.

### Changelog

**New**

- Skip-to-content link in Web Components' Storybook environment.


jeffchew:
@asudoh thank you for this change!

@kennylam @asudoh just curious if the `skip to content` link is something built into the UI Shell, or this is something specifically configured in the React storybook as well?

kennylam:
> @kennylam @asudoh just curious if the `skip to content` link is something built into the UI Shell, or this is something specifically configured in the React storybook as well?

@jeffchew Yes, `skip to content` is already included in the React UI shell's `HeaderNav` component.

asudoh:
@jeffchew `carbon-components-react` has the corresponding component, but `carbon-web-components` doesn't, because skip-to-content link essentially is not shell and I was personally hesitant to implement it as a UI shell component. But if we want it in UI shell I can go ahead with it. Let me know.

jeffchew:
> @jeffchew `carbon-components-react` has the corresponding component, but `carbon-web-components` doesn't, because skip-to-content link essentially is not shell and I was personally hesitant to implement it as a UI shell component. But if we want it in UI shell I can go ahead with it. Let me know.

I don't have a strong opinion on this, any pros and cons either way?

cc: @kennylam

asudoh:
The pro is that less confusion on the responsibility of skip-to-content link, where it's essentially in application. The con is potentially some people cannot figure out how, but I think it's less likely.

jeffchew:
> The pro is that less confusion on the responsibility of skip-to-content link, where it's essentially in application. The con is potentially some people cannot figure out how, but I think it's less likely.

So the idea is that we leave it up to the adopters to apply it themselves, rather than have it be part of something like UI Shell. 

Want to get designer/FA opinion on this: @wonilsuhibm @larahanlon2 @oliviaflory @JennySanchez @jacobottesen-dgc @photodow

jeffchew:
@asudoh and others, one other thought is that we can offer `skip-to-content` as a separate component entirely?

photodow:
@jeffchew @asudoh for me I'm curious about how important is this to be applied consistently. If we leave it up to adopters it's most likely not going to get applied. If we're ok with, and disabled users don't really need it then I'm ok with introducing it separately. The way it's implemented today doesn't look great. It might also be worth bringing up to our accessibility team, and see how much value it is and how necessary it is. To me including it in the header provides users one less thing to think about when it comes to accessibility, and it's small on our part to implement.

asudoh:
@jeffchew @photodow Thank you for your thoughts/feedbacks. Given them I think defining the component as a non-UI-shell component makes the most sense.

asudoh:
Updated so we ship the skip-to-link as a component.

packages/web-components/src/components/skip-to-content/__stories__/README.stories.mdx
```diff
@@ -0,0 +1,26 @@
...
+### JS
+
+```javascript
+import '@carbon/ibmdotcom-web-components/es/components/skip-to-content/skip-to-content';
+```
+
+### HTML
+
+```html
+<dds-skip-to-content></dds-skip-to-content>
```
jeffchew, line 25: Thank you @asudoh ! For the README, can you add documentation on the `href` property?

asudoh: Thank you for the feedback @jeffchew! After thinking a while; Now trying to implement this in `carbon-web-components`: carbon-design-system/carbon-web-components#486 - The doc includes the prop description. Will update this PR once carbon-design-system/carbon-web-components#486 is merged and released.

asudoh: Made an update to use `<bx-skip-to-content>` from `carbon-web-components`.

jeffchew (approved):
LGTM! Just need to fix merge conflicts then this should be good to go.

asudoh:
Thanks @jeffchew - Solved the merge conflict.

PR #3939, opened 2020-09-14
fix(masthead-search): no redirect when search input empty

kennylam:
### Related Ticket(s)

React version fix of #3932


asudoh (approved):
LGTM 👍 - Thanks @kennylam!

PR #3955, opened 2020-09-14
fix(masthead-container): profile menu items from translate API

asudoh:
### Description

This change ensures profile menu items in masthead is brought in from translate API when `<dds-masthead-container>` is in use.

Also:

* Copes with missing profile menu items for logged out state, by adding a NPE guard in services code.
* Removes default profile menu items in `<dds-masthead-composite>`. User can bring them in from `<dds-masthead-container>` or set them manually.

### Changelog

**Changed**

- A change to ensure profile menu items in masthead is brought in from translate API when `<dds-masthead-container>` is in use.
- A NPE guard in services code to cope with missing profile menu items for logged out state

**Removed**

- The default profile menu items in `<dds-masthead-composite>`. User can bring them in from `<dds-masthead-container>` or set them manually.


packages/services/src/services/Translation/Translation.js
```diff
@@ -161,15 +161,17 @@ class TranslationAPI {
...
-    const strReplace = 'state=https%3A%2F%2Fwww.ibm.com';
-    const loginIdx = signedout.findIndex(
-      elem => elem.url?.indexOf(strReplace) !== -1
-    );
-    if (loginIdx !== -1 && root.location) {
-      const location = encodeURIComponent(root.location.href);
-      data.profileMenu.signedout[loginIdx].url = signedout[
-        loginIdx
-      ].url.replace(strReplace, `state=${location}`);
+    if (signedout) {
```
jeffchew, line 164: Thank you for the add!

packages/web-components/src/components/dotcom-shell/__stories__/dotcom-shell.stories.ts
```diff
@@ -97,10 +98,12 @@ export const Default = ({ parameters }) => {
             lang-display="${ifNonNull(langDisplay)}"
             footer-size="${ifNonNull(footerSize)}"
             user-status="${ifNonNull(userStatus)}"
+            .authenticatedProfileItems="${ifNonNull(authenticatedProfileItems)}"
```
jeffchew, line 101: @asudoh just checking, is this a mock from profile-items?

asudoh: @jeffchew Yes this is a mock data.

jeffchew: Thanks for confirming! 

I think we will need to eventually set up a test environment that uses live data, and we know that currently the codesandbox examples don't really cut it. I'll see if I can set up something similar to what we have for the NextJS test environments.

Also, we don't use mock data for the React version, wondering if we should be considering always using live data for our storybook instances, and only use mock data for our percy visual regression snapshots? Unless I'm misunderstanding the implementation here.

asudoh: Oh unless there is `&mock` URL query arg in place the data comes from live data.

jeffchew: Thanks for confirming!

kennylam (approved):
Nice, thanks @asudoh!

Issue #3940, opened 2020-09-14
Web component: Masthead - On iPhone Voice over - menu options are not available immediately after the  triggering element.


Praveen-sr:
## Detailed description

**Jeff would like this fix included in the v1.11.0rc scheduled for Wednesday, Sept 16.**

Issue: 
After activating the Hamburger menu button
The options are not available immediately after the  triggering element.
visually it is appearing after the triggering element but for the screen reader users it is available after navigating through the following elements: 
the IBM  logo link, Open IBM search field button and User profile.

Expected: screen reader should identify the control immediately after the triggering element.

Note: Working fine in react Staging.

annawen1:
The fix for this issue in React was to simply move the `<MastheadLeftNav/>` above the Logo, search, and profile components. The `<HeaderContainer />` that wraps the Masthead provides a ` isSideNavExpanded` flag which we used to determine whether the hamburger menu is open or not.

```
              {(navigation || mastheadL1Data) && isSideNavExpanded && (
                  <MastheadLeftNav />
                )}

                <IbmLogo
                  autoid={`${stablePrefix}--masthead-${navType}__l0-logo`}
                />

                <div className={`${prefix}--header__search ${hasPlatform}`}>
                  {navigation && !mastheadL1Data && (
                    <MastheadTopNav />
                  )}
                  {hasSearch && (
                    <MastheadSearch />
                  )}
                </div>
```

asudoh:
Thank you for the info @annawen1!

PR #3956, opened 2020-09-15
fix(left-nav): set the list of nav items

asudoh:
### Related Ticket(s)

Refs #3940.

### Description

This change ensures correct list of nav items are selected when `<dds-left-nav>` searches for the nav item to focus on.

Also upgrades `carbon-web-components` to `v1.6.1` to make the focusing work.

### Changelog

**Changed**

- A change to ensure correct list of nav items are selected when `<dds-left-nav>` searches for the nav item to focus on.
- Upgrades `carbon-web-components` to `v1.6.1` to make the focusing work.


jeffchew (approved):
Thank you @asudoh !

Are there any gotchas we need to test around since we're upgrading from `1.5.0` to `1.6.1`? We should convey to Praveen if so to pay closer attention to any affected components in the next round of regression.

asudoh:
@jeffchew No basically the upgrade is only about Carbon core upgrade and the side nav fix, and the only caveat about The Carbon core upgrade was the mobile tabs.

Issue #3958, opened 2020-09-15
L0 Nav randomly disappearing upon click of 8 bar logo in subdirectories (e.g. Top Products and Solutions, Artificial Intelligence, About IBM)


elliotttrice:
## Detailed description

L0 Nav randomly disappearing upon click of 8 bar logo in subdirectories (e.g. Top Products and Solutions, Artificial Intelligence, About IBM)

> Is this a feature request (new component, new icon), a bug, or a general. Bug
> issue?

> Is this issue related to a specific component? masthead

> What did you expect to happen? What happened instead? What would you like to
> see changed?

It should take users back to homepage with complete L0 populated.  Just the 8 bar exists

> What browser are you working in? 85

> What version of the IBM.com Library are you using?

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?  We need to address this asap as this seems to be affecting multiple landing pages on ibm.com

## Steps to reproduce the issue

1. Step one. Navigate to a landing page like Top Products & Platforms, Artificial Intelligence, About IBM
2. Step two.  Click on 8 bar logo on the selected landing page to navigate back to home page
3. Step three.  In some cases, the home page masthead no longer has the full L0 navigation. It's just blank with hamburger menu on the left hand side.

## Additional information

- [Video of issue](https://ibm.ent.box.com/file/719457775489)

jeffchew:
After looking at this with @kennylam and @annawen1 , it appears to be caused by client-side logic that accounts for the menu size when the `COVID-19` menu items were in there. I think we should remove that logic for now since `COVID` isn't even in the menu anymore, and we can have the final menu overflow in the next release.

cc: @ljcarot @elliotttrice @wonilsuhibm @oliviaflory

PR #3960, opened 2020-09-15
fix(masthead): remove hide L0 logic

kennylam:
### Related Ticket(s)

#3958 

### Description

Remove breakpoint calculation for masthead L0. New implementation will be added when final designs of navigation menu overflow are complete.

### Changelog

**Removed**

- masthead L0 visibility logic


annawen1 (approved):
LGTM!

PR #3973, opened 2020-09-16
chore(content-block): drift away from BEM concatenation

asudoh:
### Description

This change to content block style drifts away from BEM concatenation, so the style can be reused with Web Components.

### Changelog

**Changed**

- A change to content block style drifts away from BEM concatenation, so the style can be reused with Web Components.


jeffchew (approved):
Thank you @asudoh !

Issue #3471, opened 2020-08-06
Remove web-component specific styles for Masthead


jeffchew:
#### User Story

> As a `[user role below]`:
IBM.com Library developer

> I need to:
ensure all styles for Masthead live in the styles package

> so that I can:
manage styles for the Masthead regardless of framework

#### Additional information

 - Current masthead styles in `web-components`: https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/web-components/src/components/masthead/masthead.scss
 - The way masthead styles are structured (with nesting) it is preventing a cleaner inclusion of the styles package styles into web components

#### Acceptance criteria
 - [ ] No more custom styles for `dds-masthead` in `web-components`

jeffchew:
Due to risk, we are moving this out to the next release.

cc: @RobertaJHahn

PR #3975, opened 2020-09-16
chore(masthead): drift away from BEM concatenation

asudoh:
### Related Ticket(s)

Refs #3471.

### Description

This change to masthead style drifts away from BEM concatenation, so the style can be reused with Web Components.

### Changelog

**Changed**

- A change to masthead style drifts away from BEM concatenation, so the style can be reused with Web Components.


jeffchew (approved):
This looks good offhand so approving! We might want Praveen to do a sanity check on the masthead for both `React` and `Web Components` to make sure everything is still fine.

PR #3977, opened 2020-09-16
chore(leadspace): drift away from BEM concatenation

asudoh:
### Description

This change to leadspace style drifts away from BEM concatenation, so the style can be reused with Web Components.

### Changelog

**Changed**

- A change to leadspace style drifts away from BEM concatenation, so the style can be reused with Web Components.


jeffchew:
@asudoh Lead space content has been bumped up in this change, can you look?

Screen Shot 2020-09-16 at 1 37 14 PM

asudoh:
Great catch @jeffchew - Fixed.

jeffchew:
Thanks @asudoh ! It definitely looks better, but there's still a diff and looks a bit squished:

Screen Shot 2020-09-17 at 4 07 06 PM

asudoh:
Another great catch @jeffchew thanks! Fixed.

jeffchew (approved):
Thank you @asudoh !

Issue #3993, opened 2020-09-17
missing image import for Feature Card Block Large


ariellalgilmore:
## Detailed description

> Describe in detail the issue you're having.

Feature Card block large is missing the image import

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

bug

> Is this issue related to a specific component?

feature card block large

PR #3994, opened 2020-09-17
fix(FeatureCardBlockLarge): missing imports

ariellalgilmore:
### Related Ticket(s)

#3993 

### Description

add missing image import to featureCardblockLarge in example and stories. Also, fix example for FeatureCardBlock to use DDSImage and import image

### Changelog

**New**

- add missing imports


asudoh (approved):

jeffchew:
@ariellalgilmore should we actually make image be a dependency of `feature-card-block-large`, so adopters don't need to import two different things?

ariellalgilmore:
@jeffchew Thanks! Let me know if the changes now make sense :)

PR #3996, opened 2020-09-17
fix(left-nav): change element order

asudoh:
### Related Ticket(s)

Refs #3940.

### Description

Changes the element order of `<dds-left-nav>` vs. the rest of the masthead so VoiceOver cursor won't immediately go out as user moves VoiceOver cursor after opening left nav.

Also, deactivates the focus sentinel of `<dds-masthead-menu-button>` given it's not needed unless left nav opens and using it while left nav is closed adds redundant tab-stop.

### Changelog

**Changed**

- The element order of `<dds-left-nav>`.
- A change to deactivate the focus sentinel of `<dds-masthead-menu-button>` while left nav is closed.


jeffchew (approved):
LGTM!

Issue #3895, opened 2020-09-10
Web component: Footer - locale selector, search result message displays wrongly.


Praveen-sr:
## Detailed description

Issue: Users search for country/language, even if country/language is unavailable in that region message shows as available.

Note: Working fine on react canary version.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/a368bff7-3260-4d21-9aee-f88f38775e28)

PR #3998, opened 2020-09-18
fix(locale-modal): add missing text for unvailable locale

asudoh:
### Related Ticket(s)

Refs #3895.

### Description

This change adds `unavailability-label-text` attribute to `<dds-locale-search>` so it can show a different text when it gets an empty search result.

Also made some changes to ensure translation text for locale modal is in use fully.

### Changelog

**New**

- `unavailability-label-text` attribute to `<dds-locale-search>` so it can show a different text when it gets an empty search result.

**Changed**

- Some changes to ensure translation text for locale modal is in use fully.


jeffchew (approved):
LGTM, thank you @asudoh !

Issue #3999, opened 2020-09-18
[Masthead]: when opening L1 dropdown, body scroll is disabled


annawen1:
## Detailed description

> Describe in detail the issue you're having.

When the L1 dropdown is opened/expanded, user is unable to scroll the page.

> Is this issue related to a specific component?

Masthead

> What did you expect to happen? What happened instead? What would you like to
> see changed?

User should be able to still scroll the page when the dropdown is opened/expanded.

## Steps to reproduce the issue

1. Go to http://support-page-test.mybluemix.net/
2. Open one of the L1 dropdowns
3. Try to scroll the page
4. User is unable to scroll page

PR #4002, opened 2020-09-18
fix(Masthead): nav dropdowns should allow body scroll

annawen1:
### Related Ticket(s)

[Masthead]: when opening L1 dropdown, body scroll is disabled #3999 

### Description

Dropdown menus on the Masthead should allow for body scrolling as opposed to the MegaMenu which disables body scroll.

This PR passes in a `disableScroll` prop to the `<HeaderMenu />` component based on whether the menu is a MegaMenu or a regular dropdown. 

### Changelog

**New**

- `disableScroll` prop to the `<HeaderMenu />`

**Changed**

- use `disableScroll` to determine whether to disable scrolling


jeffchew (approved):
LGTM!

Issue #3990, opened 2020-09-17
React: Masthead - On empty search bar, clicking on search icon, page navigates to search results.


RobertaJHahn:
## Detailed description

Issue: On empty search bar, user clicking on search icon, refreshes the page and navigates to search results.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/b3d12416-f182-43e9-add0-b6ac331aba71)

Landing page:
![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/d1232021-545e-48c6-8b5e-30680da20e3b)

IgnacioBecerra:
What should the expected functionality be for an empty search? We could make it so clicking the icon/pressing enter won't work until text is inputted, but searching for " " (space) seems to return all the IBM pages, which may be a better result for an empty search. Any thoughts?

RobertaJHahn:
@photodow See the comment from @IgnacioBecerra above.  Can you answer his question?

photodow:
@IgnacioBecerra I believe the expected behavior on the IBM.com Library side is the use clicks the search icon, and executes a search by redirecting them to the IBM.com search page. Our search component doesn't have any expectations beyond that, but from a personal standpoint I believe I would expect there to be maybe an unfiltered list of search results waiting for me.

@RobertaJHahn I don't believe this to be a bug on our end, unless the expectation is to change the functionality of the search component to not execute if the input value is empty.

IgnacioBecerra:
@kennylam addressed this issue in #3939. The changes are currently live in the React-canary environment, which will eventually incorporated into the regular React release. Closing this for now -- if we still have issues after release we can reopen or create a new issue.

PR #4008, opened 2020-09-18
fix(Masthead): empty search handling

IgnacioBecerra:
### Related Ticket(s)

#3990 

### Description

Before, when users clicked/pressed enter into an empty search bar, they would get redirected to a page with no results. While it is expected behavior (seeing as an empty query would return nothing), search conditions were changed so an empty search cannot happen at all in the first place.

### Changelog

**Changed**

- Added search condition where it only runs when there's text inputted in the search box.

### DIFF
--- a/packages/react/src/components/Masthead/MastheadSearch.js
+++ b/packages/react/src/components/Masthead/MastheadSearch.js
@@ -224,7 +224,7 @@ const MastheadSearch = ({
    *
    */
   function searchIconClick() {
-    if (state.isSearchOpen && state.val.length) {
+    if (state.isSearchOpen && state.val.length > 1) {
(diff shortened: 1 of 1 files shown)


kennylam:
@ignacio-becerra Should this behave differently from https://github.com/carbon-design-system/ibm-dotcom-library/pull/3939?

IgnacioBecerra:
> @ignacio-becerra Should this behave differently from #3939?
I originally believed it was to fix that same issue, since clicking the search in ibmdotcome-react Storybook would still redirect, but saw that in the react-canary version it was fixed already.

I think this issue is actually about Storybook React redirecting the whole tab to the search results page instead of within Storybook (like it does in web-components).

PR #4017, opened 2020-09-21
fix(footer): fixing the default footer type definition

jeffchew:
### Related Ticket(s)

No related issue

### Description

Using the `Footer` as standalone was causing an issue where the
navigation links were not rendering if not defined, and passing in
`tall` did not work. This happens specifically in NextJS when
building and exporting the statically generated pages (rendering in
dev mode appears to be unaffected).

The default value for some reason was set as `full`, which is incorrect
. This logic now checks if the `type` is set as `tall` or `undefined`,
and will render accordingly.

### Changelog

**Changed**

- Fixes to `Footer` type definition


asudoh (approved):
LGTM 👍 - Thanks @jeffchew!

annawen1 (approved):

PR #4019, opened 2020-09-21
docs(examples): lock browserlist version

asudoh:
### Description

This change locks `browserlist` and `caniuse-lite` versions to work around a bug that is recently introduced in either of the packages, that broke our build integration tests.

### Changelog

**Changed**

- A change to kock `browserlist` and `caniuse-lite`.


jeffchew (approved):
Thank you for the fix!

Issue #3297, opened 2020-07-28
Component:Leadspace (Stage):- Buttons display issue on 672px .


chetankan:
Issue Description:- In leadspace component if there are more than 3 buttons on 672px md breakpoint in that case , primary button is not visible completely.

## Steps to reproduce the issue

1. Enter url in the browser.
2. Select 'Leadspace' component.
3. In knobs section , enter number of buttons 4.
4. Set browser to 672px break point.
5. Check if primary button is visible.

OS:- Android 8,9
Device:- Samsung Galaxy tab S5,6

URL:- https://ibmdotcom-react-staging.mybluemix.net/?path=/story/components-leadspace--default-with-no-image

![Leadspace on adnroid tablet.jpg](https://images.zenhubusercontent.com/5e21a841639f517d456f7e43/77f2d0f1-721f-45cd-a428-52af5d169a6a)

PR #4025, opened 2020-09-22
fix(leadspace): buttons now display properly at 672px breakpoint

IgnacioBecerra:
### Related Ticket(s)
#3297 

### Description

At 672px (md) breakpoint, adding more than 3 buttons would make the primary button would be counted as "overflow", and thus wouldn't display properly. By restructuring the children elements of`leadspace__content` to `position: relative`, the outer `section` would take into account the additional buttons included and wrap around them properly.

Before:

After:

### Changelog

**Changes**

- Refactored layout rules to accommodate for md breakpoint


kennylam (dismissed):
@IgnacioBecerra It looks like removing the overflow is affecting the layout with image

image

kennylam:
@IgnacioBecerra It looks like removing the overflow is affecting the layout with image

image

IgnacioBecerra:
Fixed layout issue @kennylam mentioned but discovered another issue with the image variation, where the image doesn't cover the button layout at 672px.



Will revert the PR to a draft until I fix it.

kennylam (approved):
Looks good thanks @ignacio-becerra

PR #4041, opened 2020-09-24
fix(styles): adjusting carbon-components-react to dev dependency

jeffchew:
### Related Ticket(s)

No related issue

### Description

It was reported that the styles package was pulling in `carbon-components-react`, which is not necessary for the styles package itself but for the storybook generation for the expressive theme. This moves this dependency to `devDependencies` instead.

### Changelog

**Changed**

- Moved `carbon-components-react` to `devDependencies` for the styles package


asudoh (approved):
LGTM 👍 - Thanks @jeffchew!

Issue #3980, opened 2020-09-16
Web Components: Provide CDN version of pre-built bundle with dotcom-shell


asudoh:
### The problem

Many of our v18 users want to use our library not requiring heavy toolstack but just with loading `.js`/`.css`.

Web Components library tried to meet such requirement with ECMAScript module, but it's blocked by:

1. No browser has implemented [import maps spec](https://github.com/WICG/import-maps), that is required for browser to understand `import { LitElement } from 'lit-element'` code (to figure out where `lit-element` module `.js` file is)
2. Some CDNs (e.g. JSPM/UNPKG) tries to make above work by replacing NPM module name with actual CDN path on-the-fly, but it has performance problem by having to load many modules, and tends to hit many glitches

### The solution

Create pre-built bundle.

### Additional information

- We can make use of the EI server and upload to there (`https://1.s81c.com/common/ibm-dotcom-library/@{version}/**`)

### Acceptance criteria

- [ ] Create generated bundle that includes the `dotcom-shell`
- [ ] Publish up to EI server (`beta`, `next` and `latest`)

jeffchew:
Should we have one where it's just dotcom shell, and another one with the kitchen sink?

We can host the artifact on EI server so it's behind akamai cache, so uploading should be fairly trivial. There's already a CDN version of v18 on there as well (under `http://www.ibm.com/common/v18`), we could create a folder like `http://www.ibm.com/common/ibm-dotcom-library`.

asudoh:
> Should we have one where it's just dotcom shell, and another one with the kitchen sink?

I'm thinking of it and asked the requester of this feature if she's using only dotcom shell, but she responded that it's not the case. So we may need to do some sort of [code-splitting pattern](https://webpack.js.org/guides/code-splitting/).

jeffchew:
> I'm thinking of it and asked the requester of this feature if she's using only dotcom shell, but she responded that it's not the case. So we may need to do some sort of [code-splitting pattern](https://webpack.js.org/guides/code-splitting/).

Right, perhaps we could create a webpack config that traverses the component folder, and creates an artifact for each component? Then people can inject by using something like:

```
<script src="//1.www.s81c.com/common/ibm-dotcom-library/dotcom-shell.js" /> 
<script src="//1.www.s81c.com/common/ibm-dotcom-library/lead-space.js" />
<script src="//1.www.s81c.com/common/ibm-dotcom-library/feature-card.js" />
```

[...]

PR #4045, opened 2020-09-24
chore(build): introduce bundle in Web Components

asudoh:
### Related Ticket(s)

Refs #3980.

### Description

Introduces `ibmdotcom-web-components-dotcom-shell.js` and `ibmdotcom-web-components-dotcom-shell.min.js` Rollup bundles so that our users can quickly set up dotcom shell without setting up any toolstack. Also introduces an example that demonstrates how to use the bundle.

### Changelog

**New**

- `ibmdotcom-web-components-dotcom-shell.js` and `ibmdotcom-web-components-dotcom-shell.min.js` Rollup bundles.
- An example that demonstrates how to use the bundle.

**Changed**

- Changes so our components stop rendering core `carbon-web-components` directly:
  - Replaces `<bx-search>` with `<dds-search>`. The latter is effectively the alias of the former
  - Replaces `<bx-header>` with plain `<header>`
- Replaces `clean-css` with `cssnano` so the minifier is well integrated to other PostCSS toolchain.
- Adds `@babel/plugin-transform-runtime` back so the same Babel helpers won't be repeated in the bundle.
- Integrated `yarn clean` task into `gulp clean` task.
- Changed how `lodash-es` modules are imported for the sake of less reliance to tree-shaking.

**Removed**


packages/web-components/examples/static-files/index.html
```diff
@@ -0,0 +1,127 @@
...
+        margin-top: 0;
+      }
+    </style>
+    <script type="module">
+      // Copy from `dist` and put it to the same directory as this file
+      import './ibmdotcom-web-components-dotcom-shell.min.js';
+    </script>
+  </head>
+  <body>
+    <dds-masthead>
```
jeffchew, line 40: @asudoh does the bundle also include `dds-dotcom-shell-container`? That should probably be the `quick start` version of this, followed by the customized `masthead` for more advanced implementations.

asudoh: @jeffchew Yes. I started with manual approach due to `CORS_PROXY` thing, van update this once our fix to its reliance us in.

jeffchew: You can probably already get started on switching over, the access control policy is already live on EI server.

asudoh: 🎉 - Updated the examples.

jeffchew (approved):
Thank you @asudoh , looks good!

PR #4021, opened 2020-09-22
feat(MegaMenu): layout iteration in breakpoint between 800 - 1055px

annawen1:
### Related Ticket(s)

React: MegaMenu iteration 1 breakpoint and layout updates #3948

### Description

New layout for Megamenu at 800px - 1055px.

NOTE: The L0 nav items will appear squished in the lower breakpoints (~800px). We will need the new overflow feature to the Masthead (React: Add the overflow feature to the Masthead #3875) for this to be fixed.

EXPECTED:
Screen Shot 2020-09-22 at 9 22 42 AM

### Changelog

**Changed**

- added styling for breakpoints in between 800 - 1055px

**Removed**

- stale styling for code used to hide masthead

PR #4046, opened 2020-09-24
fix(masthead): fix breakpoint

asudoh:
### Description

Found that the latest breakpoint introduced for React masthead (#4021) does not fit well with non-megamenu environment (like what our Web Components codebase is as of today).

Reverted the breakpoint only for Web Components styles, to fix that problem.

### Changelog

**Changed**

- Change the breakpoint of narrow mode masthead back to `lg`.


IgnacioBecerra (approved):
LGTM!

PR #4059, opened 2020-09-25
chore(components): use internalProperty decorator

asudoh:
### Description

Starts using `@internalProperty()` decorator for all components so we don't have to manually call `requestUpdate()` to re-render the view upon change in `private`/`proptected` properties.

### Changelog

**Changed**

- A series of changes to use `@internalProperty()` decorator for all components.


kennylam (approved):
Looks good thanks @asudoh!

Issue #3843, opened 2020-09-02
Web Component: Feature card - arrow link appears below the copy text


Praveen-sr:
## Detailed description
On Mobile & Tab view

Issue:  Feature card - arrow link appears below the copy text

Expected: Arrow link must be at right corner of the card.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/73765b94-95df-49c9-aa41-7f23322ada77)

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/a48429c7-5434-454c-bb73-e40310654cea)

PR #4066, opened 2020-09-25
fix(feature-card-block-large): repositioned arrow to bottom right

IgnacioBecerra:
### Related Ticket(s)

#3843

### Description

Repositioned the arrow to the bottom right to ensure it doesn't move around when changing resolutions. 

The web-component version imports the styles from @carbon/ibmdotcom-styles, which come from the style file that was changed in this PR. 

Before:

After:

### Changelog

**New**

- Added flex rule to `bx--card__footer` in `feature-card-block-large.scss` for React
- Added slotted rule to `feature-card-block-large.scss` for Web Components


packages/styles/scss/components/feature-card-block-large/_feature-card-block-large.scss
```diff
@@ -134,6 +134,10 @@ $fcb-breakpoint-up--lg: map-get(
       @include carbon--type-style('productive-heading-05');
     }
 
+    .#{$prefix}--card__footer {
+      flex-direction: row-reverse;
+    }
```
asudoh, line 139: Thank you for trying to fix this @IgnacioBecerra! Seems that we need deeper look at why React is working vs. why Web Components doesn't. The style rule here is reused between React and Web Components, and the change here seems to have adverse effect to React (as Percy shows).

IgnacioBecerra: I pushed a different Web-Component specific change that should fix this. However, the same changes are seen through Percy. From what I can see though, isn't the only change the arrow moving to the right, as it should be?

asudoh: @IgnacioBecerra OK given we have web-component specific change the fixes the problem, sounds that we don't need anything here? Or am I missing something? If it the case, could you please elaborate more on what you mean by:

> However, the same changes are seen through Percy. From what I can see though, isn't the only change the arrow moving to the right, as it should be?

Thanks!

IgnacioBecerra: Yeah, I think the issue is fixed on both React and Web Components. As for the comment, I was a little confused as Percy didn't show any adverse effect to React to my knowledge in the original commit. Please let me know if there's a Percy change I'm not noticing, but I think we're good now! 

Thanks Akira!

jeffchew (approved):
LGTM!

packages/styles/scss/components/feature-card-block-large/_feature-card-block-large.scss
asudoh, line 139: Thank you for your thoughts @IgnacioBecerra! In case I was not clear enough (my apologies if it's the case), here's what I'm actually suggesting. Please let me know if it doesn't work. Thanks!

```suggestion
```

IgnacioBecerra: Hmm, it seems removing it has no effect, the arrow stays on the left unless `row-reverse` is used. It's also used in the other Feature Cards as well.

jeffchew: Let's move forward with the existing solution.

asudoh: @IgnacioBecerra @jeffchew This code seems to have caused a visual regression in React. Let's see how we can resolve the arrow icon position in another way.

jeffchew: it looks like it moved the arrow to the lower right instead of lower left, that's the intention right?
Screen Shot 2020-10-01 at 6 34 48 PM

asudoh: Oh totally missed that the visual change is intentional. Please disreegard my earlier messages. My sincere apologies on this.

Issue #4049, opened 2020-09-24
React: Change Language Selector from behind the feature flag


RobertaJHahn:
#### User Story

> As a `[user role below]`:
IBM.com Library developer

> I need to:
remove the feature flag from the `Language Selector` component

> so that I can:
provide the ibm.com adopter developers components they can use to build ibm.com web pages

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- Prod QA testing issue (#4050)
- Create a separate sub-story to Language Selector in storybook
- Feature flag documentation: https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/react/docs/feature-flags.md

#### Acceptance Criteria
- [ ] Updated version of the existing component is available for testing and included in v1.12.0 release
- [ ] A comment is posted in the Prod QA issue, tagging Praveen, when development is finished

Issue #3665, opened 2020-08-20
Error showing for the Footer language selector when in DotcomShell


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

When clicking the language selector (not locale modal button), an error shows up:

<img width="1522" alt="Screen Shot 2020-08-20 at 9 55 11 AM" src="https://user-images.githubusercontent.com/1641214/90784582-6cd18700-e2cf-11ea-948b-27f92fe38973.png">

> Is this issue related to a specific component?

Footer language selector

## Steps to reproduce the issue

1. Go to React experimental (https://ibmdotcom-react-experimental.mybluemix.net)
2. Go to Dotcomshell
3. Enable "languageOnly"
4. Click the language selector dropdown

## Additional Information

- The standalone footer appears to work

PR #4069, opened 2020-09-28
feat(footer): graduating language selector to production ready

kennylam:
### Related Ticket(s)

#3665, #4049 

### Description

Removes footer language-only dropdown from behind feature flag.

### Changelog

**New**

- footer stories with `languageOnly` variant.

**Changed**

- update README

**Removed**

- feature flag for language dropdown


packages/react/src/components/Footer/__stories__/Footer.stories.js
```diff
@@ -54,13 +53,50 @@ Default.story = {
...
+         * Language callback demo function
+         *
+         * @param {string} selectedItem Selected item
+         */
+        const languageCallback = selectedItem => {
+          console.log(
+            'footer (language selector) selected item:',
+            selectedItem
+          );
+        };
```
jeffchew, line 66: @kennylam caught a wild console log here

packages/react/src/components/Footer/__stories__/Footer.stories.js
```diff
@@ -6,7 +6,6 @@
  */
```
jeffchew, line 6: @kennylam can stories be added to dotcom shell as well?

packages/react/src/components/Footer/__stories__/Footer.stories.js
kennylam: @jeffchew I saw these – I assumed they were intentional to display the `lc-cc` when a selection is made and are part of the story. I can remove if we're not using anymore.

packages/react/src/components/Footer/__stories__/Footer.stories.js
kennylam: I was debating this but felt it was redundant so decided against it. But I guess it can't hurt.

jeffchew: I think it can help since we can't really see the language selector in the short and micro versions since the open upward and off the screen in the footer storybook story. Thinking about this too, I wonder if we want to vertically center the footer in the footer stories?

kennylam: @jeffchew Yep, I actually did that already as well.

packages/react/src/components/Footer/__stories__/Footer.stories.js
jeffchew: Yeah I think removing would be fine.

packages/react/src/components/Footer/__stories__/Footer.stories.js
jeffchew: hey @kennylam I don't see the change yet.

jeffchew (approved):
Thanks @kennylam ! It looks like footer still is top aligned in the stories so the short and micro footers are still cutting off the language selector at the top, but I think we can merge in for now since we have these represented in the dotcom shell stories.

Also probably not related to this story, it looks like there are missing CTA stories in web components not, not sure why they are missing now.

kennylam:
> Thanks @kennylam ! It looks like footer still is top aligned in the stories so the short and micro footers are still cutting off the language selector at the top, but I think we can merge in for now since we have these represented in the dotcom shell stories.

@jeffchew This has been updated.

PR #4070, opened 2020-09-28
feat(callout-data): web component created

RaphaelAmadeu:
### Related Ticket(s)

#3743 

### Description

Web components version of `callout-data` created.


packages/web-components/src/components/callout-data/__stories__/callout-data.stories.ts
```diff
@@ -0,0 +1,62 @@
...
+            <span slot="copy">
+              ${copy}
+            </span>
+            <span slot="source">
+              ${source}
+            </span>
+          </dds-callout-data>
+        </div>
+      </div>
+    </div>
```
asudoh, line 34: Could you extract the grid portion out to a Storybook decorator?

packages/web-components/src/components/callout-data/callout-data.ts
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Callout Data.
+ *
+ * @element dds-callout-data
+ */
+@customElement(`${ddsPrefix}-callout-data`)
+class DDSCalloutData extends LitElement {
+  render() {
+    return html`
+      <dds-callout-data-container class="test">
```
asudoh, line 28: Any reason to have `<dds-callout-data-container>` as a separate component, instead of `<div class="${prefix}--callout__column">`?

asudoh (commented):
Great progress - Thanks @RaphaelAmadeu!

packages/web-components/src/components/callout-data/callout-data.ts
RaphaelAmadeu: @asudoh I had to style callout deeper than a slotted callout would permit, so creating a new component lilke this allow me to have more control over css.

asudoh: OK thank you for explaining @RaphaelAmadeu! I think the case here seems to be different from the case with `<dds-content-block-heading>`, etc. We have created `<dds-content-block-heading>`, etc. given the element is placed at light DOM where `<dds-content-block>`'s shadow DOM style does no control over. OTOH e.g. if we put `<div class="${prefix}--callout__column">` (or `<section class="${prefix}--callout__container">`) here in shadow DOM, the shadow DOM style of `<dds-callout-data>` has the control over the the `<div>` (or `<section>`). Does this make sense? Feel free to ping me otherwise - Thanks!

packages/web-components/src/components/callout-data/__stories__/callout-data.stories.ts
asudoh: Let's not forget about this.

asudoh (commented):
Getting there, just one remaining comment - Thanks @RaphaelAmadeu!

asudoh (approved):
🎉 - Thank you @RaphaelAmadeu!

Update: Can you fix the ESLint error? Self-closing tag for non-void elements is not supported by `lit-html` as it's an invalid syntax, and ESLint now detects it.

jeffchew (approved):

packages/web-components/src/components/callout-data/callout-data.ts
```diff
@@ -0,0 +1,42 @@
...
+ * @element dds-callout-data
+ */
+@customElement(`${ddsPrefix}-callout-data`)
+class DDSCalloutData extends LitElement {
+  render() {
+    return html`
+      <div class="${prefix}--callout__container">
+        <div class="${prefix}--callout__column">
+          <div class="${prefix}--callout__content">
+            <h4 class="${prefix}--callout-data__data"><slot name="data" /></h4>
```
asudoh, line 30:
```suggestion
            <h4 class="${prefix}--callout-data__data"><slot name="data"></slot></h4>
```

packages/web-components/src/components/callout-data/callout-data.ts
```diff
@@ -0,0 +1,42 @@
...
+            <p class="${prefix}--callout-data__copy"><slot name="copy" /></p>
```
asudoh, line 31:
```suggestion
            <p class="${prefix}--callout-data__copy"><slot name="copy"></slot></p>
```

packages/web-components/src/components/callout-data/callout-data.ts
```diff
@@ -0,0 +1,42 @@
...
+          </div>
+        </div>
+      </div>
+      <p class="${prefix}--callout-data__source"><slot name="source" /></p>
```
asudoh, line 35:
```suggestion
      <p class="${prefix}--callout-data__source"><slot name="source"></slot></p>
```

asudoh (commented):
Change suggestions to make the ESLint errors go away:

Issue #4063, opened 2020-09-25
search function stopped working in language selector


yellowdragonfly:
search function stopped working in language selector in v1.11.0

Adopter:  **Privacy Preference**

ljcarot:
@yellowdragonfly there is no search function in language selector, do you mean Locale Selector?

ljcarot:
@yellowdragonfly It would be helpful if you can provide either a url where you see this, a recording of your experience or a screenshot. Thank you.

yellowdragonfly:
https://proxy1.marketplace-test.ibmcloud.com/profile/dataprivacypreferences/welcome/us-en

jeffchew:
Hi @yellowdragonfly , this appears to be a style collision issue where the hidden class is not getting honored, and this only appears to be in the full bundled CSS (we aren't seeing this in our Storybook or NextJS test application).

For now, what you can do to get around this is adding this to your code:

```
.bx--locale-modal__locales-hidden {
  display: none !important;
}
```

We will take on this bug and get into the next release in the meantime.

cc: @kennylam @IgnacioBecerra @ljcarot @asudoh @RobertaJHahn

PR #4075, opened 2020-09-29
fix(css): update hidden styles for locale countries

kennylam:
### Related Ticket(s)

#4063 

### Description

Adds css specificity to avoid overrides

### Changelog

**Changed**

- add specificity to hidden locale items for filtering


jeffchew (approved):
LGTM!

Issue #3712, opened 2020-08-25
Component: Link with icon - On disabled, Keyboard focus is identified for the link


Praveen-sr:
## Detailed description

Issue :- On disabled mode, Keyboard focus is identified for the link

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/0ce61178-2633-4454-ab6a-446e30ef8c0c)

PR #4082, opened 2020-09-29
fix(link-with-icon): removed keyboard focus on disable

IgnacioBecerra:
### Related Ticket(s)

#3712 

### Description

The main issue that caused the element to still be focused even while disabled. Removed `tabIndex` as it should still respect the focus order while active and not focus during disabled state.

### Changelog

**Removed**

- Removed tabIndex from the `Link` element


asudoh (commented):
Hi @IgnacioBecerra I think the internal `<Link>` is keyboard-focusable even without `tabIndex` specified given it uses `<a>` internally. The internal `<Link>` automatically turns off such focusability if it's disabled, by [turning it to `<p>`](https://github.com/carbon-design-system/carbon/blob/v10.20.0/packages/react/src/components/Link/Link.js#L29). That said, would you try simply removing `tabIndex` from the internal `<Link>`? Thanks!

IgnacioBecerra:
@asudoh You're right, Link does turn it into `<p>`, I hadn't noticed that. Removed tabIndex as it's not necessary.

jeffchew:
@IgnacioBecerra looks like you need to update the snapshot file to reflect the change. You can do this with `yarn test:unit:updateSnapshot` in the web components package or react package.

packages/react/src/components/LinkWithIcon/LinkWithIcon.js
```diff
@@ -37,7 +38,7 @@ const LinkWithIcon = ({
       )}
       data-autoid={`${stablePrefix}--link-with-icon`}>
       <Link
-        tabIndex={0}
+        tabIndex={disabled ? -1 : 0}
```
asudoh, line 41:
```suggestion
```

packages/react/src/components/LinkWithIcon/LinkWithIcon.js
```diff
@@ -21,6 +21,7 @@ const { prefix } = settings;
 const LinkWithIcon = ({
   children,
   className: customClassName,
+  disabled,
```
asudoh, line 24:
```suggestion
```

packages/react/src/components/LinkWithIcon/LinkWithIcon.js
```diff
@@ -55,6 +56,11 @@ LinkWithIcon.propTypes = {
    */
   children: PropTypes.arrayOf(PropTypes.node),
 
+  /**
+   * Toggles keyboard tab focus for element
+   */
+  disabled: PropTypes.bool,
+
```
asudoh, line 63:
```suggestion
```

packages/react/src/components/LinkWithIcon/LinkWithIcon.js
```diff
@@ -78,6 +84,7 @@ LinkWithIcon.propTypes = {
 
 LinkWithIcon.defaultProps = {
   children: [],
+  disabled: false,
```
asudoh, line 87:
```suggestion
```

asudoh (commented):
Thank you for the update @IgnacioBecerra - Below is what I actually meant in case I wasn't clear enough (my apologies if it's the case). Please let me know if it doesn't work. Thanks!

IgnacioBecerra:
@asudoh Oh, my bad! I forgot to push the changes after writing that comment. No worries, you were clear enough 👍 

Also, I updated the React screenshots, thanks @jeffchew!

jeffchew:
> @asudoh Oh, my bad! I forgot to push the changes after writing that comment. No worries, you were clear enough 👍
> 
> Also, I updated the React screenshots, thanks @jeffchew!

No worries! It looks like you have merge conflicts now, can you resolve?

asudoh (approved):
Cool, LGTM 👍 once the snapshot is recreated to resolve the merge conflict. Thanks a lot @IgnacioBecerra!

jeffchew:
@IgnacioBecerra you may need to re-run the updateSnapshot task to fix this.

PR #4084, opened 2020-09-30
feat(leadspace): leadspace web component

annawen1:
### Related Ticket(s)

Web Component: Develop LeadSpace of the React version - Group 1 #3561

### Description

LeadSpace web component

Screen Shot 2020-10-13 at 9 57 39 AM

### Changelog

**New**

- leadspace web component

**Changed**

- react version of leadspace restructured in order to accommodate how the web components version needs to be built
- leadspace styles (the centered styles moved under the main `bx--leadspace` class


packages/web-components/src/components/leadspace/leadspace-body.ts
```diff
@@ -0,0 +1,182 @@
...
+  @property({ attribute: 'theme', reflect: true })
+  theme = LEADSPACE_THEME;
+
+  /**
+   * `true` to hide the divider.
+   */
+  @property({ type: Boolean, reflect: true, attribute: 'gradient' })
+  gradient = true;
+
+  _renderCopy() {
```
asudoh, line 69:
```suggestion
  protected _renderCopy() {
```

packages/web-components/src/components/leadspace/leadspace-body.ts
```diff
@@ -0,0 +1,182 @@
...
+    return html`
+      <div class="${prefix}--leadspace__row">
+        <p data-autoid="${ddsPrefix}--leadspace__desc" class="${prefix}--leadspace__desc">
+          <slot name="copy">${copy}</slot>
+        </p>
+      </div>
+    `;
+  }
+
+  _renderTitle() {
```
asudoh, line 80:
```suggestion
  protected _renderTitle() {
```

packages/web-components/src/components/leadspace/leadspace-body.ts
```diff
@@ -0,0 +1,182 @@
...
+    const { title } = this;
+    return html`
+      <slot name="title">${title}</slot>
+    `;
+  }
+
+  _renderImage() {
```
asudoh, line 87:
```suggestion
  protected _renderImage() {
```

packages/web-components/src/components/leadspace/leadspace-body.ts
```diff
@@ -0,0 +1,182 @@
...
+  /**
+   *  Carbon Gray 100 theme
+   */
+  G100 = 'g100',
+}
+
+/**
+ * LeadSpace body.
+ */
+abstract class DDSLeadSpaceBody extends LitElement {
```
asudoh, line 56: Naming nit, from the `abstract` annotation and how this class is used:

```suggestion
abstract class DDSLeadSpaceBase extends LitElement {
```

asudoh (commented):
Great progress, the comments for now in addition to the comments to the code are the following. Thanks @annawen1!

* Please make sure we add comments to all properties/methods, the only exceptions are inherited ones: https://github.com/carbon-design-system/ibm-dotcom-library/blob/v1.11.0/packages/web-components/docs/coding-conventions.md#tsdoc-comments
* Methods beginning with `get`/`set`: The codebase uses the naming convention where a function returns some computed value should begin with `get`, and a function that sets something should begin with `set`. From that, `gradientClass()`/`typeClass()`/`setBackgroundImage()` all seems to be in `get` category. BTW please make sure they also begin with `_`, given they are protected methods.

packages/web-components/src/components/leadspace/leadspace.ts
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * Copyright IBM Corp. 2020
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { customElement } from 'lit-element';
+import ddsSettings from '@carbon/ibmdotcom-utilities/es/utilities/settings/settings.js';
+import DDSLeadSpaceBase from './leadspace-body';
```
asudoh, line 12: Let's not forget about changing the file name here.

packages/web-components/src/components/leadspace/leadspace-body.ts
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  /**
+   * Sets background image for leadspace centered
+   *
+   * @protected
+   */
+  protected _getBackgroundImage() {
+    return this.defaultSrc && this.type === LEADSPACE_TYPE.CENTERED ? `background-image: url(${this.defaultSrc})` : '';
+  }
```
asudoh, line 192: Is it possible to move this code chunk above to where other `protected` methods are?

packages/web-components/src/components/leadspace/leadspace-body.ts
```diff
@@ -0,0 +1,197 @@
...
+  protected _getGradientClass() {
+    return this.gradient ? `${prefix}--leadspace--gradient` : '';
+  }
+
+  /**
+   * Returns a class-name based on the type parameter type
+   *
+   * @protected
+   */
+  protected getTypeClass() {
```
asudoh, line 173:
```suggestion
  protected _getTypeClass() {
```

packages/web-components/src/components/leadspace/leadspace-body.ts
```diff
@@ -0,0 +1,197 @@
...
+
+  /**
+   * The leadspace title.
+   */
+  @property()
+  title = '';
+
+  render() {
+    return html`
+      <div style="${this._getBackgroundImage()}" class="${prefix}--leadspace__container ${this.getTypeClass()}">
```
asudoh, line 142:
```suggestion
      <div style="${this._getBackgroundImage()}" class="${prefix}--leadspace__container ${this._getTypeClass()}">
```

packages/web-components/src/components/leadspace/leadspace-body.ts
```diff
@@ -0,0 +1,197 @@
...
+        </div>
+        ${this._renderImage()}
+      </div>
+    `;
+  }
+
+  /**
+   * Returns a class-name based on the gradient parameter type
+   *
+   * @protected
```
asudoh, line 162: We add `@protected` annotation only if we cannot do the annotation via TypeScript:

```suggestion
```

packages/web-components/src/components/leadspace/leadspace-body.ts
```diff
@@ -0,0 +1,197 @@
...
+   */
+  protected _getGradientClass() {
+    return this.gradient ? `${prefix}--leadspace--gradient` : '';
+  }
+
+  /**
+   * Returns a class-name based on the type parameter type
+   *
+   * @protected
```
asudoh, line 171: We add `@protected` annotation only if we cannot do the annotation via TypeScript:

```suggestion
```

packages/web-components/src/components/leadspace/leadspace-body.ts
```diff
@@ -0,0 +1,197 @@
...
+        return `${prefix}--leadspace--productive`;
+      default:
+        return '';
+    }
+  }
+
+  /**
+   * Sets background image for leadspace centered
+   *
+   * @protected
```
asudoh, line 188: We add `@protected` annotation only if we cannot do the annotation via TypeScript:

```suggestion
```

packages/web-components/src/components/leadspace/leadspace-body.ts
```diff
@@ -0,0 +1,197 @@
...
+   * Carbon color theme of the Leadspace
+   */
+  @property({ reflect: true })
+  theme = LEADSPACE_THEME;
+
+  /**
+   * `true` to hide the divider.
+   */
+  @property({ type: Boolean, reflect: true })
+  gradient = true;
```
asudoh, line 73: Is it possible to move this code chunk down to where other declared properties (ones with `@property()`) are?

asudoh (commented):
Thank you for the updates, great progress - BTW great to see that you updated the corresponding React code (probably to better align the codebase?). Thanks @annawen1!

packages/web-components/src/components/leadspace/leadspace.ts
annawen1: whoops, thanks for catching that!

packages/web-components/src/components/leadspace/leadspace-body.ts
```diff
@@ -0,0 +1,182 @@
...
+
+/**
+ * LeadSpace body.
+ */
+abstract class DDSLeadSpaceBody extends LitElement {
+  @property({ attribute: 'type', reflect: true })
+  type = LEADSPACE_TYPE.LEFT;
+
+  @property({ attribute: 'theme', reflect: true })
+  theme = LEADSPACE_THEME;
```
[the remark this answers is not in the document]
annawen1: yep working on this

asudoh:
Thank you for the updates, no additional comment but renaming `leadspace-body.ts` to `leadspace-base.ts`. Looking forward to seeing this in ready-for-review state. Thanks @annawen1!

packages/web-components/src/components/leadspace/leadspace.ts
```diff
@@ -0,0 +1,223 @@
...
+
+    if (this.theme) {
+      classname += `${prefix}--leadspace--${this.theme} `;
+    }
+
+    if (this.type === 'small') {
+      classname += `${prefix}--leadspace--productive `;
+    }
+    return classname;
+  }
```
asudoh, line 97: Is it possible to make this function use `lit-element`'s `classMap`? An example can be found at: https://github.com/carbon-design-system/carbon-web-components/blob/v1.7.0/src/components/button/button.ts#L207-L214

(Same for `_getGradientClass()`)

packages/web-components/src/components/leadspace/leadspace.ts
```diff
@@ -0,0 +1,223 @@
...
+   *  Leadspace type (small, left, or centered)
+   */
+  @property({ reflect: true })
+  type = LEADSPACE_TYPE.LEFT;
+
+  /**
+   * Carbon color theme of the Leadspace
+   */
+  @property({ reflect: true })
+  theme = LEADSPACE_THEME;
```
asudoh, line 191: Continuing on [the discussion on `theme` and CSS custom properties](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/4084#discussion_r497258273); Wondering if theme chooser introduced some weeks ago (#4106) is enough for showcase the leadspace with a particular theme, or `theme` attribute/property is still needed.

asudoh (commented):
Great progress - Thanks @annawen1!

packages/web-components/src/components/leadspace/leadspace.ts
annawen1: hm yeah I think you're right, I'll remove the `theme` property! Thanks!

packages/web-components/src/components/leadspace/leadspace.ts
annawen1: ah yes! this was something I was looking for! Thanks!

packages/web-components/src/components/leadspace/leadspace.ts
asudoh: Splendid - Thanks @annawen1!

annawen1: I ran into issues with the gradient styles, specifically with the `rgba()` method where it doesn't grab the value for `$ui-background` token in time so I had to hardcode the colors. So Im thinking of keeping the theming property for now

asudoh: @annawen1 Thank you for taking a look. Would you able to use [`getComputedStyle()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle) to query for `--ui-background` so that we can turn `theme` to an internal (private) property? If it sounds too much please let me know and I can take some look.

jeffchew:
Hey @annawen1 , just checking if this was an intentional change on the React side?

Screen Shot 2020-10-14 at 12 58 00 PM

annawen1:
> Hey @annawen1 , just checking if this was an intentional change on the React side?
> 
> Screen Shot 2020-10-14 at 12 58 00 PM

@jeffchew That occurred when I shifted the styles around, I looked at the specs and it matches - I believe the button sizes previously were too small

packages/web-components/src/components/leadspace/leadspace.ts
annawen1: I see, I'm not too sure how to implement that approach, should we merge this PR first and then take a look on how to do that?

asudoh: Makes sense @annawen1! Let's go this way.

asudoh (approved):

asudoh:
BTW do you want to update the React snapshots? Thanks!

asudoh:
Note for future reference, we could do auto-detection of effective theme by:

```typescript
updated() {
  requestAnimationFrame(() => {
    // TODO: Make our style automatically reflect the effective theme instead of having to manually set another CSS class
    if (this.ownerDocument.defaultView.getComputedStyle(this).getPropertyValue('--cds-ui-background') === '#161616') {
      this._theme = LEADSPACE_THEME.G100;
    }
  });
}
```

PR #4086, opened 2020-09-30
chore(footer): vertically center short footer stories

kennylam:
### Description

Vertically centers `short` and `micro` variations of footer. This is helpful for the `language only` versions of the footer where the language dropdown expands upwards.

### Changelog

**Changed**

- vertically align `short` and `micro` footer to the center of story viewport.


jeffchew (approved):
LGTM!

Issue #3964, opened 2020-09-16
React: LinkWithIcon - icon placement is wrong when disabled and inverse option is activated.


Praveen-sr:
## Detailed description

Issue: LinkWithIcon - icon placement is on wrong side when 'disabled' and 'inverse' option is activated.

## Steps to reproduce the issue

1. Open LinkWithIcon  on react staging link
2. On knobs section, change icon placement to "left"
3. Click on 'disabled' & 'inverse' checkbox 
4. Now, observe the link arrow icon is on right side, it must be on left. 
On hover, the arrow icon changes to left side.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/2f7c5e3d-8c77-4ede-a56b-e384c459122f)

PR #4095, opened 2020-09-30
fix(link-with-icon): fixed left icon placement w/options

IgnacioBecerra:
### Related Ticket(s)

#3964 

### Description

Icon repositioned to the right when both inverse and disable were active. It is a display issue where `display: flex` was only active during hover, while `display: inline` is active otherwise, causing the issue. Changed the `display` value to fit the hover behavior.

Before:

After:

### Changelog

**Changed**

- Changed `display: inline` to `display: flex` to account for both disabled and inverse options are active,


jeffchew (approved):
LGTM!

PR #4110, opened 2020-10-01
feat(pictogram-item): web-component created

RaphaelAmadeu:
### Related Ticket(s)

#3739 

### Description

Web component created, following the react version.


packages/web-components/src/components/pictogram-item/pictogram-item.ts
```diff
@@ -0,0 +1,64 @@
...
+  copy = '';
+
+  protected _handleSlotChange({ target }: Event) {
+    const { name } = target as HTMLSlotElement;
+    if (name === 'copy') {
+      this.copy = ((target as HTMLSlotElement).assignedNodes()[0] as HTMLElement).innerText
+        .replace(/[\n\r]+|[\s]{2,}/g, ' ')
+        .trim();
+    }
+  }
```
asudoh, line 40: Any reason mangling the text content of a slotted child instead of making `copy` a public attribute/prop and let user use it instead?

asudoh (commented):
Great to see you came up with this quick, just one comment - Thanks @RaphaelAmadeu!

packages/web-components/src/components/pictogram-item/pictogram-item.ts
RaphaelAmadeu: I thought it would be better to receive it as a `slot`. We are receiving content this way for most of our web-components.

asudoh: Good point, we have had a exception, though; Copy content containing markdown doesn't use `<slot>` as a similar reason as here.

If you feel very strongly about "every content should be `<slot>`", then one suggestion is using `textContent` instead of `innerText`. In this way, you need less mangling (or no mangling is needed).

packages/web-components/src/components/pictogram-item/__stories__/README.stories.mdx
```diff
@@ -0,0 +1,56 @@
...
+      d="M37,32 H11c-1.1,0-2-0.9-2-2V13c0-1.1,0.9-2,2-2h26c1.1,0,2,0.9,2,2v17C39,31.1,38.1,32,37,32z M17,37h14 M24,32v5 M9,27h30">
+    </path>
+  </svg>
+  <span slot="heading">Lorem ipsum dolor sit</span>
+  <span slot="copy">
+    Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna
+    aliqua. Ut enim ad minim veniam
+  </span>
+  <dds-link-with-icon slot="cta">
+    Lorem ipsum dolor ${ArrowRight20({ slot: 'icon' })}
```
jeffchew, line 51: @RaphaelAmadeu can you replace the `ArrowRight20` with an svg element?

packages/web-components/src/components/pictogram-item/__stories__/pictogram-item.stories.ts
```diff
@@ -0,0 +1,70 @@
...
+        stroke-linejoin="round"
+        stroke-miterlimit="10"
+        stroke-width=".72"
+        d="M37,32 H11c-1.1,0-2-0.9-2-2V13c0-1.1,0.9-2,2-2h26c1.1,
+        0,2,0.9,2,2v17C39,31.1,38.1,32,37,32z M17,37h14 M24,32v5 M9,27h30"
+      ></path>
+    </svg>
+    <span slot="heading">Lorem ipsum dolor sit</span>
+    <span slot="copy">
+      Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
```
jeffchew, line 44: @RaphaelAmadeu can you make these into knobs?

RaphaelAmadeu: @jeffchew This was added on my latest commit

packages/web-components/src/components/pictogram-item/__stories__/README.stories.mdx
RaphaelAmadeu: @jeffchew I've added this on my latest commit

packages/web-components/src/components/pictogram-item/pictogram-item.ts
RaphaelAmadeu: @asudoh this was addressed on my latest commit, but the replace and trim are still necessary, otherwise, the text looks off, similar when used within a `code` tag

asudoh (approved):
LGTM 👍 - Could you update the snapshot? Thanks @RaphaelAmadeu!

jeffchew (approved):
LGTM!

I know we didn't have the conversation yet in huddle about documenting slots yet, but I think we can do this in another PR once we have had that conversation as a team.

PR #4116, opened 2020-10-02
chore(format): introduce eslint-plugin-lit

asudoh:
### Description

Introduce `eslint-plugin-lit` to Web Components codebase so we can catch invalid `lit-html` template syntax, notably usage of self-closing tags for non-void elements.

Also removes the annotations to disable `class-methods-use-this` ESLint rule where it's not needed.

### Changelog

**New**

- `eslint-plugin-lit` to Web Components codebase.

**Removed**

- The annotations to disable `class-methods-use-this` ESLint rule where it's not needed.


jeffchew (approved):
Thanks @asudoh !

PR #4120, opened 2020-10-02
docs(svg): updating documentation for web components and svg imagery

jeffchew:
### Related Ticket(s)

No related issue

### Description

The exmaples shown originally were a bit misleading as embedding web
component markup on HTML would not yield an SVG when imported. This
shows how to use an SVG inlined and defining its slot when using in
conjunction with the IBM.com Library web components.

### Changelog

**Changed**

- Multiple READMEs for Web Components
- missing import for `feature-card-block-medium`


asudoh (approved):
Good catch and great fix and right that `lit-html` icons cannot be used in plain HTML. Thanks @jeffchew!

PR #4121, opened 2020-10-02
fix(video): changing video id across packages

jeffchew:
### Related Ticket(s)

No related issue

### Description

This PR fixes all of our test environments to use a Kaltura Video ID
that does not require the configuring of the partnerID or UI Conf ID.
This was a problem of inconsistency across all of our lower environments
 as some videos were not playing in certain isolated tests. This will
 also assist with new local development setups, as this is one less
 thing to configure locally.

### Changelog

**Changed**

- Changed the test video ID globally


asudoh (approved):
LGTM 👍 - Thanks @jeffchew!

PR #4126, opened 2020-10-05
chore(components): misc props update

asudoh:
### Changelog

**Changed**

Fixes the following, etc.:

- Missing `type: Boolean` for boolean attribute
- Camel-cased attribute names where they should be dash-cased
- Wrong position of code within clase


kennylam (approved):
Looks good thanks @asudoh!

Issue #4096, opened 2020-09-30
Web Components: Expose the standalone video player as a story in storybook


jeffchew:
#### User Story

> As a `[user role below]`:
IBM.com Library adopter

> I need to:
have a standalone video player web component

> so that I can:
implement the video player using web components in my application

#### Additional information

 - The video player already exists, we just need to expose as its own story in storybook so adopters know how to import and use
- Wondering if we should also consider exposing the kaltura partner ID/uiconfID as something that can be configured without environment variables

#### Acceptance criteria
 - [ ] Added video player story in storybook
 - [ ] Corresponding documentation

PR #4130, opened 2020-10-05
docs(video-player): add story

asudoh:
### Related Ticket(s)

Refs #4096.

### Changelog

**New**

- Storybook story for `<dds-video-player>`.
- CodeSandbox example for `<dds-video-player>`.

### DIFF
--- a/packages/web-components/.storybook/container.scss
+++ b/packages/web-components/.storybook/container.scss
@@ -70,6 +70,11 @@ body {
   align-items: stretch;
 }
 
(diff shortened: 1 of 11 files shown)


jeffchew:
Thank you @asudoh ! 

We may want to add a `percy: { skip: true }` to the video player story, so we don't introduce any false positives there. We do the same for the `VideoPlayer` for React.

asudoh:
Good point @jeffchew - Updated.

jeffchew (approved):
Thank you @asudoh !

Issue #4132, opened 2020-10-05
React - Masthead L1 - user can not see any menu options


Praveen-sr:
## Detailed description

On iPad - iOS 13
Issue: Masthead L1  - no menu options (hamburger menu or mega menu) are shown for the user.

Note: when we switch to landscape mode, menu is seen.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/536b3339-6a0d-4c0c-800d-9aad950bfb1d)

Landscape mode:
![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/596e8c99-e5c6-4978-ae68-61eec851d553)

PR #4135, opened 2020-10-05
fix(masthead): fix L1 breakpoint styles

kennylam:
### Related Ticket(s)

#4132 

### Description

This fixes responsive breakpoints for masthead that was setting `display: none` on the `L1` too early. This issue is not specific to tablet devices and can be reproduced in desktop browsers as well.

http://ibmdotcom-react-canary.mybluemix.net/iframe.html?id=components-masthead--with-l-1

### Changelog

**Changed**

- update responsive styles for `L1`


asudoh (approved):
LGTM 👍 - Thanks @kennylam!

PR #4137, opened 2020-10-05
chore(eslint): adding `dist` folder to .eslintignore

jeffchew:
### Related Ticket(s)

#3980 

### Description

The `dist` folder is generated when the `dotcom-shell` CDN artifact is
created locally. This is an additional rule to not run eslint under this
 folder.

### Changelog

**Changed**

- Added rule to `.eslintignore`


asudoh (approved):
Good catch - Thanks @jeffchew!

Issue #3881, opened 2020-09-09
Web component: Feature card block large - Focus is not covering entire card.


Praveen-sr:
## Detailed description

Issue: Feature card block large - On keyboard navigation, focus is not covering entire card.

Note: This is needs to be fixed on Web component & react canary versions.

## Additional information

Actual :
![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/5314ce8e-8312-4f34-a6ac-ceb41ea4e902)

Expected:
![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/afa79402-02f1-4336-b2c9-fc8367311dd5)

PR #4139, opened 2020-10-05
fix(feature-card-block-large): focus outline now wraps around card properly

IgnacioBecerra:
### Related Ticket(s)

#3881 

### Description

The previous FeatureCardBlockLarge wasn't focusing correctly due to reusing the `outline` values meant for the `Default` and `Medium` CardBlocks. Inserting custom values to the Large `:focus` state makes it fit more adequately. 

Before (notice bottom right):
Screen Shot 2020-10-05 at 3 59 55 PM

After: 
Screen Shot 2020-10-05 at 3 58 13 PM

### Changelog

**New**

- Added new `outline` values to `.bx--feature-card-block-large .bx--card:focus`


packages/styles/scss/components/feature-card-block-large/_feature-card-block-large.scss
```diff
@@ -34,6 +34,8 @@ $fcb-breakpoint-up--lg: map-get(
     .#{$prefix}--card {
       background: none;
       height: 100%;
+      outline-width: 3px;
+      outline-offset: 1px;
```
kennylam, line 38: Looks good @IgnacioBecerra. Small thing, can we put these styles into `:focus` to remain consistent with other component focus states.

IgnacioBecerra: Updated!

kennylam (approved):
Looks good thanks @IgnacioBecerra!

Issue #4140, opened 2020-10-06
React: Masthead - On tab, 'Platform' is not visible on portrait mode. 


Praveen-sr:
## Detailed description

On iPad - iOS 13
Issue: Masthead - 'Platform' is not visible on portrait mode. 

Note: On Landscape mode its seen.
## Additional information

Actual (portrait mode):
![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/b61cd034-2193-4a77-85f3-d5b2375c8600)

Landscape mode:
![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/800b21cd-13ee-47f1-b701-75d3f8ae755f)

PR #4165, opened 2020-10-07
fix(masthead): update platform name responsive css

kennylam:
### Related Ticket(s)

#4140 

### Description

Fix masthead platform name to remain visible until hamburger menu appears. Note this issue can be reproduced in all browsers, not only tablet/mobile devices.

Oct-07-2020 11-58-03

### Changelog

**Changed**

- update responsive css for masthead platform name


ariellalgilmore (approved):
LGTM :) thanks @kennylam

Issue #3669, opened 2020-08-20
[Tags] update padding to have 32px height tags with 14px texts to align with Carbon mobile native 


shixiedesign:
A minor visual update so tags in Expressive theme is aligned with Carbon iOS native library. Team agreed the 32px tall tags are preferred for: 

- visually more comfortable
- the selectable variant (with x for clear tag) has to be larger for touch target requirement on mobile

### The As-is

Current expressive theme tags: https://carbon-expressive.mybluemix.net/?path=/story/tag--default

![image](https://user-images.githubusercontent.com/15144993/90799593-719f3680-e2e1-11ea-9957-50cb14691dd5.png)

### The To-be

- [x] Increase padding top & bottom from 4 to 6px
- [x] Increase padding left & right from 4 to 8px
- [ ] Darken text color from 70 grade to 80 grade 
- [x] Increase text line height from 18 to line height: 20px 
- [ ] Allow for custom icon added to start of tag

![image](https://user-images.githubusercontent.com/15144993/90799339-11a89000-e2e1-11ea-9b97-4c28213f39ba.png)

#### Usage example: 

![image](https://user-images.githubusercontent.com/15144993/90797272-6e567b80-e2de-11ea-94a8-193ce7a88deb.png)

PR #4166, opened 2020-10-07
feat(tag): expressive theme update

IgnacioBecerra:
### Related Ticket(s)

#3669

### Description

Changing the Tag styles used for Expressive Theme. Bigger font, darker colors, more padding is included. 

image

(Allow icon will be included in a separate PR to the carbon-components repository)

### Changelog

**Changed**

- Expressive theme styles added
- Increased overall padding
- Darkened text color
- Increased line height and font


packages/styles/scss/themes/expressive/components/_tag.scss
```diff
@@ -16,6 +16,71 @@
...
+  }
+
+  .#{$prefix}--tag--filter {
+    padding-right: $spacing-04;
+  }
+
+  .#{$prefix}--tag--disabled,
+  .#{$prefix}--tag--filter.#{$prefix}--tag--disabled {
+    color: $disabled-02;
+  }
```
asudoh, line 72: Wondering the color styles in here were actually needed, as they are [implemented in Carbon core](https://github.com/carbon-design-system/carbon/blob/v10.21.0/packages/components/src/components/tag/_tag.scss#L48-L139).

IgnacioBecerra: The main difference between these and those is that the colors here are 80 grade while the others are 70 grade. One of the new things in the Expressive theme is the darkening of the colors. But if there's another way to get 80 grade colors from the Carbon core classes that I'm missing, please let me know!

asudoh: > One of the new things in the Expressive theme is the darkening of the colors.

Thank you for pointing it out @IgnacioBecerra! Sounds that we need more system-wide change, similar to what we do for type tokens. I think we can start with changing color tokens at https://github.com/carbon-design-system/ibm-dotcom-library/blob/v1.11.0/packages/styles/scss/themes/expressive/_tokens.scss#L371-L479.

@jeffchew Is this something worthwhile as a engineering huddle topic?

jeffchew: Sorry, I missed this comment. Let's bring up as a huddle topic @IgnacioBecerra @asudoh .

asudoh: Cool, done!

packages/styles/scss/themes/expressive/components/_tag.scss
```diff
@@ -16,6 +16,74 @@
 /// @group tag-expressive
 @mixin tag-expressive {
   .#{$prefix}--tag {
+    @include carbon--type-style('body-long-01');
```
oliviaflory, line 19: I don't think we want to change the type token to `body-long-01` here because it's increasing the text size to 16px?

I thought the expressive theme token updates should have updated the original token `label-01` to 14px per the design specs, unless I am mistaken.

jeffchew:
Recapping some of the bullets from conversations:
- update the height in the expressive theme
- contribute color changes through Carbon repo so the system stays aligned between Carbon productive/expressive
- add the variation for icon before the tag / contribute back

Based on this, we will not be handling the color changes in this PR. @IgnacioBecerra can you update?

@oliviaflory @wonilsuhibm you mentioned also the variation for icon you still want in but we eventually want to contribute back, right?

cc: @asudoh

packages/styles/scss/themes/expressive/components/_tag.scss
```diff
@@ -16,6 +16,36 @@
 /// @group tag-expressive
 @mixin tag-expressive {
   .#{$prefix}--tag {
+    @include carbon--type-style('label-01');
+
+    line-height: 20;
```
asudoh, line 21:
```suggestion
    line-height: carbon--rem(20px);
```

packages/styles/scss/themes/expressive/components/_tag.scss
```diff
@@ -16,6 +16,36 @@
...
+  .#{$prefix}--tag--filter.#{$prefix}--tag--disabled {
+    color: $disabled-02;
+  }
+
+  .#{$prefix}--tag__close-icon {
+    position: relative;
+    left: 6px;
+
+    width: 1.25rem;
+    height: 1.25rem;
```
asudoh, line 41:
```suggestion
    width: $icon-size-01;
    height: $icon-size-01;
```

asudoh (commented):
Thanks @IgnacioBecerra for the updates!

oliviaflory:
@IgnacioBecerra This might be nit picky, but is there a way to change the default label in storybook? It's a bit weird that it says "This is not a tag" when it is a tag..

Screen Shot 2020-10-22 at 12 13 46 PM

Ideally it would say "This is a tag" 

Realize this might not be in scope, so if it's a difficult thing I can open another issue to address this.

IgnacioBecerra:
@asudoh I had used `carbon--rem(20px)` before, but eslint doesn't expect the `line-height` value to have `px` (weird), so left it at 20. I changed it to `carbon--rem(20)` since it still appears to work.

@oliviaflory I always thought that was a weird label to have when I was testing, and even the carbon-components Storybook has the regular `This is a tag`. Changed it to that in the latest commit!

asudoh:
@IgnacioBecerra What specific StyleLint error did you get?

IgnacioBecerra:
@asudoh 
The error I got was: `unexpected unit px for property line-height` 

Screen Shot 2020-10-22 at 9 37 01 AM

When resolving this issue I found [the docs](https://stylelint.io/user-guide/rules/unit-whitelist), which said that having `line-height` without `px` wasn't considered a violation.

asudoh:
I see, it's hostile to how Carbon core defines one-height`. Also unitless is a different animal from pixel unit. Given those, we have to change the StyleLint rule. I can take a look at it tomorrow.

IgnacioBecerra:
I see, that's good to know unitless is not the same as pixel unit. Thanks Akira!

asudoh:
https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.12.0/packages/stylelint-config-ibmdotcom/rules/limit-language-features.js#L80 seems the culprit. A couple of alternatives are:

* Remove the rule
* Disable the rule for the particular line (Given `carbon--rem()` makes sure we use the `rem` unit)

IgnacioBecerra:
@asudoh Ended up disabling the rule for that specific line. 👍

packages/styles/scss/themes/expressive/components/_tag.scss
```diff
@@ -16,6 +16,37 @@
 /// @group tag-expressive
 @mixin tag-expressive {
   .#{$prefix}--tag {
+    @include carbon--type-style('label-01');
+
+    /* stylelint-disable-next-line */
```
asudoh, line 21:
```suggestion
    /* stylelint-disable-next-line declaration-property-unit-whitelist */
```

asudoh (commented):
Cool, just a suggestion to make the ignore thing more specific.

asudoh (approved):
Thanks @IgnacioBecerra for all the updates!

Issue #4006, opened 2020-09-18
Adding a Hotjar heatmap triggers to each of mega menu tabs -- ibm.com Reboot project


olchowo:
### The problem
Since August 31st ibm.com/us-en users have an opportunity to use new Carbon mega menu. We would like to use Hotjar heatmaps to learn, how the users are interacting with this new design. 
Those insights are important for further iterations of the mega menu and for work on other elements of overall IBM web navigational system, conducted by design team. 
Heatmaps are part of ongoing user research & Hotjar is the easiest to use tool that can generate them. Those heatmaps will be very valuable for us, so thank you in advance for your help on this matter.

### The solution
US homepage already have a Hotjar widget that allows us to map the user behaviour. Dynamic page elements though, like overlays, in this case: mega menu tabs, require a special JS trigger to be recognized and mapped by the tool. 
Please add following piece of code to JS function opening each of tabs:

for Products&Solutions tab: hj('trigger', 'MM_products-solutions');
for Services&Consulting tab: hj('trigger', 'MM_services-consulting');
for Learn&Support tab: hj('trigger', 'MM_learn-support');
for Explore more tab: hj('trigger', 'MM_explore-more');

When the code is deployed I will be able to set separate Hotjar heatmaps using each of those triggers. 

We've discussed this request with Anna Wen and Jeff Chew & they found a way to convert this into piece of code reusable for future purposes. Please check the chat transcript attached to this ticket.

### Additional information
 - Hotjar documentation page related to that request: https://help.hotjar.com/hc/en-us/articles/115011867948-How-to-Use-JavaScript-to-Trigger-Heatmaps
 - chat transcript:
[reboot-MM_hotjar__chat-transcript-18sep2020.txt](https://github.com/carbon-design-system/ibm-dotcom-library/files/5247293/reboot-MM_hotjar__chat-transcript-18sep2020.txt)
 - Thank you :)

RobertaJHahn:
@annawen1 @jeffchew Can you add additional details to this issue?  Thanks.

PR #4169, opened 2020-10-07
feat(masthead): add custom events to megamenu

kennylam:
### Related Ticket(s)

#4006 

### Description

This adds custom events to megamenu/sidenav, emitted on click.

### Changelog

**New**

- megamenu top-level `data-title` attribute
- emits `onMegaMenuToggle` event on nav link click

**Changed**

- update megamenu mock data to latest
- fix sideNav overlay responsive css


asudoh (commented):
Thank you for your quick change @kennylam! I know it's hard (esp. with React) and don't want to block this development upon my thoughts here, but I think it's better if we can avoid baking in the analytics feature here. Wondering if we can introduce `onSelectMenuItem` prop to `<Masthead>` so we can let our user call `hj()` with letting our user translate `data-autoid` to the ID analytics wants.

kennylam:
> Thank you for your quick change @kennylam! I know it's hard (esp. with React) and don't want to block this development upon my thoughts here, but I think it's better if we can avoid baking in the analytics feature here. Wondering if we can introduce `onSelectMenuItem` prop to `<Masthead>` so we can let our user call `hj()` with letting our user translate `data-autoid` to the ID analytics wants.

Thanks for the feedback @asudoh! Yes, we spoke separating this logic from the Masthead. This would be the ideal solution, but the tricky part is that the adopters in this case are not using the `autoid` as the identifier. But still I think you're right, let me chew on this and see if there's a better way to implement. Plus, you just reminded me that I left out the same triggers for `MastheadLeftNav`.

jeffchew:
Hey @asudoh ! thank you for the thoughts, when @annawen1 and I were discussing this, we didn't want to have specific hotjar methods baked into here. But instead, create a generic dispatch event that they can create an event listener for. This way, this implementation can be used for other purposes other than hotjar.

jeffchew (approved):
Looks much better, thank you @kennylam !

packages/react/src/components/carbon-components-react/UIShell/SideNavMenu.js
```diff
@@ -109,6 +109,14 @@ export class SideNavMenu extends React.Component {
 
   handleToggleExpand = event => {
     const { onToggle } = this.props;
+    let onMegaMenuOpen = new Event('megaMenuOpen', { bubbles: true });
+    let onMegaMenuClose = new Event('megaMenuClose', { bubbles: true });
```
asudoh, line 113:
```suggestion
    const onMegaMenuToggle = new CustomEvent('megaMenuToggle', { bubbles: true, detail: { expanded: this.state.isExpanded } });
```

packages/react/src/components/Masthead/MastheadLeftNav.js
```diff
@@ -48,7 +54,8 @@ const MastheadLeftNav = ({
               key={i}
               autoid={autoid}
               selected={rest.selectedMenuItem === link.titleEnglish}
-              navType={rest.navType}>
+              navType={rest.navType}
+              id={`${linkId}-leftnav`}>
```
asudoh, line 58: Wondering if we can use an attribute (probably `data-` attribute) rather than `id`, as a precaution to avoid duplicate IDs.

asudoh (commented):
Nice updates - Thanks @kennylam!

packages/react/src/components/carbon-components-react/UIShell/SideNavMenu.js
kennylam: Good call, thanks @asudoh!

packages/react/src/components/Masthead/MastheadLeftNav.js
kennylam: @asudoh. I was on the fence about using a `data` attribute, so thanks for suggesting!

asudoh (approved):
LGTM 👍 - Thanks @kennylam!

annawen1 (approved):
LGTM!

Issue #4171, opened 2020-10-08
Web component: CTA - feature card, copy text and links are not visible


Praveen-sr:
## Detailed description
On iPhone - iOS 13 & 14
Issue: On feature card, 
1. copy text on the card is not visible. 
2. When user changes knobs settings to "local", "Jump", "External" or "Video" the links are not visible.

Note: On Jaws (screen reader ) Copy text and links are read.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/687a5b1c-eec9-40f6-8a19-bef421a69e8b)

Issue #4128, opened 2020-10-05
Web component: CTA - Feature card sizing is small compared to react


Praveen-sr:
## Detailed description

Issue:   CTA - Feature card sizing is small compared to react

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/367c1eca-56f0-44b3-a009-4a60697efad5)

react version
![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/ffd89d19-531c-4b3d-8146-e50b9fe07b14)

PR #4176, opened 2020-10-08
fix(feature-card): remove redundant flex-wrap

asudoh:
### Related Ticket(s)

Refs #4171.

### Description

Fixes a bug that is observed where `<dds-feature-card>` is placed in a grid whose width is notably smaller than browser viewport.

Also made changes to our stories using grid to unify similar other stories, using common story parameter and CSS classes.

### Changelog

**Changed**

- Fixes a bug that is observed where `<dds-feature-card>` is placed in a grid whose width is notably smaller than browser viewport.
- Changes to our stories using grid to unify similar other stories.


asudoh:
@ariellalgilmore @RaphaelAmadeu Please let me know if you have any concerns with the changes to the stories. Thanks!

ariellalgilmore (commented):
Thanks @asudoh! This makes a lot of sense and will help for future development. Seems like it's causing extra padding around components in react though

asudoh:
Thanks @ariellalgilmore for pointing it out - Adjusted the padding, wondering if it works better. Please don't hesitate to speak up otherwise, thanks!

asudoh:
May address: #4128

jeffchew (approved):
Nice add, thank you @asudoh !

ariellalgilmore:
@asudoh looks pretty good! Sorry just saw you updated the padding. I see now though the padding in web components is different than in react. For example, cardgroup and featurecard are all positioned directly at the top in react, but in web components there is an extra space between the top and the component. Also, card seems to be centered aligned instead of left aligned. I think in an application this won't make a difference, but when testing it might throw the designers or QA team off

kennylam (approved):
Looks good thanks @asudoh! Looks like Ari's comments refer to the storybook containers and how our components are presented. We probably need to standardize the layouts between our react and web component stories. But that is separate from this issue.

asudoh:
@ariellalgilmore @kennylam Thank you for reviewing!

> Looks like Ari's comments refer to the storybook containers and how our components are presented. We probably need to standardize the layouts between our react and web component stories.

Exactly, was trying to create the standard layout for those stories and tried to make it sensible for Web Components. Once it makes sense to everybody it should be applied to React, too. That said, please definitely don't hesitate to point out if any stories looks odd (besides the difference from React).

Issue #4170, opened 2020-10-08
Callout: Consider removing base callout or making it as a mix-in


asudoh:
### The problem

`<dds-callout>` does not seem to have a use case of stand-along usage and there is no corresponding story. It essentially renders the markup for layout for the callout pattern, and callout pattern implementation tends to inherit other components (e.g. `<dds-callout-with-media>` inherits `<dds-content-block-simple>`).

### The solution

Let e.g. `<dds-content-block-simple>` simply render its layout by itself, or make `<dds-callout>` code a mix-in.

asudoh:
Fixed by: #4228

Issue #4122, opened 2020-10-04
Quote: Consider removing inverse attribute


asudoh:
### The problem

`<dds-quote>` has `inverse` attribute as its React counterpart has the corresponding property. However, in Carbon core, usage of `inverse` is limited to components that may be used in a components in inverse color scheme (e.g. Carbon core top navigation bar's color scheme) and there may be a case where React `<Quote>` one is merely for theme zoning.

### The solution

Clarify above question. If React `<Quote>` one is merely for theme zoning, remove `inverse` attribute from `<dds-quote>`. If `<dds-quote>` may really be used with components with inverse color scheme, rename `inverse` boolean attribute to `color-scheme` enum attribute so it's forward-compatible with color scheme addition.

RaphaelAmadeu:
I've confirmed with @photodow and `inverse` should remais as is right now, I'll update the PR so it just fixes the inverse funcionality. 

CC: @asudoh , @RobertaJHahn

PR #4179, opened 2020-10-08
feat(callout-quote): web-component created

RaphaelAmadeu:
### Related Ticket(s)

#3745 

### Description

Created web-components version of `callout-quote`


packages/web-components/src/components/callout-quote/callout-quote.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Callout Data.
+ *
+ * @element dds-callout-data
+ */
+@customElement(`${ddsPrefix}-callout-quote`)
+class DDSCalloutQuote extends LitElement {
+  render() {
+    return html`
+      <dds-callout>
+        <dds-quote inverse>
```
asudoh, line 28: As discussed in #4122, would we be able to do theme zoning with CSS custom properties instead of relying on `inverse` attribute here?

packages/web-components/src/components/callout-quote/__stories__/callout-quote.stories.ts
```diff
@@ -0,0 +1,71 @@
...
+        copy: textNullable(
+          'Quote (copy):',
+          'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vivamus est purus, posuere at est vitae, ' +
+            'ornare rhoncus sem. Suspendisse vitae tellus fermentum, hendrerit augue eu, placerat magna.',
+          groupId
+        ),
+        sourceHeading: textNullable('Source Heading(sourceHeading)', 'Lorem ipsum dolor sit amet', groupId),
+        sourceCopy: textNullable('Source Copy(sourceCopy)', 'consectetur adipiscing elit', groupId),
+        sourceBottomCopy: textNullable('Source Copy(sourceBottomCopy)', 'IBM Cloud', groupId),
+        inverse: boolean('Inverse (inverse)', false, groupId),
```
asudoh, line 56: This doesn't seem to be used, right?

asudoh (commented):
Thank you for coming up with this component quickly @RaphaelAmadeu!

packages/web-components/src/components/callout-quote/callout-quote.ts
asudoh: FYI, this can be done by:

```scss
:host(#{$dds-prefix}-callout-quote) #{$dds-prefix}-quote {
  @include carbon--theme($carbon--theme--g100, true);

  // The original quote text color inherits from body.
  // Re-defines the color to apply the theme zoning.
  color: $text-01;
}
```

An alternate approach is making `<dds-callout-quote>` inherit `<dds-quote>`. It can be done along with #4170.

asudoh (commented):
Thank you for several updates @RaphaelAmadeu! Have you had a chance to look at https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/4179#discussion_r501835432? If not no problem just wanted to make sure this is not buried in our conversation pile. Also please don't hesitate to ping me for any questions. Thanks!

Issue #4177, opened 2020-10-08
Web components: Masthead - menu options are not visible for the user.


Praveen-sr:
## Detailed description

Issue: When user clicks on the menu option, menus are not visible. They are hidden under the masthead.

## Additional information

![Web_Masthead.gif](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/2004f963-73fd-48b5-9ecb-d3e824d5b38f)

PR #4180, opened 2020-10-08
fix(masthead): fix menu

asudoh:
### Related Ticket(s)

Refs #4177.

### Description

The masthead nav overflow feature made tweaks to overflow style so it can keep track of overflowed content. It works with mega menu that "floats", but Web Components masthead menu with classic menu style still relies on the overflow area being visible. Introduces Web Components-specific style to fix that.

### Changelog

**Changed**

- Fix to masthead style so the overflow area is visible in Web Components, as masthead nav menu relies on that style.


jeffchew (approved):
LGTM!

PR #4194, opened 2020-10-09
chore(dotcom-shell): rename composite SCSS

asudoh:
### Description

Renames `dotcom-shell-container.scss` to `dotcom-shell-composite.scss` so it aligns to other `.scss` files applied to `<*-composite>` and `<*-container>` components.

### Changelog

**Changed**

- Renames `dotcom-shell-container.scss` to `dotcom-shell-composite.scss`.


jeffchew (approved):
Thanks @asudoh !

Issue #4190, opened 2020-10-08
React: Masthead - search suggestions showing underneath bar


IgnacioBecerra:
## Detailed description

In React-Canary, the search results aren't visible to the user. What's happening under the hood is that the suggestions cause an overflow which somehow becomes hidden.

> Is this issue related to a specific component?

Masthead

> What browser are you working in?

Chrome

## Steps to reproduce the issue

1. Search for anything on the Masthead Search bar.

## Additional information

![Screen Shot 2020-10-08 at 2 54 44 PM](https://user-images.githubusercontent.com/24970122/95519539-25d04b80-097a-11eb-9453-0f5678dd3356.png)

- Notes
Removing `overflow-x: auto` fixes this issue, however, it affects the overflow feature, removing the ">" icon and thus its functionality.

![Screen Shot 2020-10-08 at 3 02 25 PM](https://user-images.githubusercontent.com/24970122/95519720-8f505a00-097a-11eb-9e94-e126c7465047.png)

Related: #4119

Issue #4181, opened 2020-10-08
Masthead overflow – gradient rendering as gray(?) in Safari


oliviaflory:
## Detailed description

> Describe in detail the issue you're having.

The gradient within the masthead when the overflow is activated looks gray(?), it should be a white to transparent gradient.

<img width="902" alt="Screen Shot 2020-10-08 at 12 03 13 PM" src="https://user-images.githubusercontent.com/45692622/95484515-8f426100-095e-11eb-80d8-ae0825327d3e.png">

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

Masthead

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Should match specs in [spec document](https://ibm.ent.box.com/file/716827406287?s=00ieobl9wsxjsxb3rhs7y6o7mccurx5r), see page 32 

> What browser are you working in?

Safari
Version 13.1.2 (15609.3.5.1.3)

> What version of the IBM.com Library are you using?

Canary

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

October release

## Steps to reproduce the issue

1. View in Safari, make browser small to activate overflow
2. Step two
3. Step three
4. etc.

## Additional information

- Screenshots or code
- Notes

PR #4205, opened 2020-10-09
fix(masthead): search options visible

ariellalgilmore:
### Related Ticket(s)

#4190 #4181 

### Description

add search state to know if search is open which will then turn overflow off, so that the search options are still visible. Also, change transparency token to rgba values for Safari

### Changelog

**New**

- state for open/closed search bar

**Changed**

- styles for overflow and transparency

search options
Screen Shot 2020-10-09 at 12 17 57 PM

safari button transparency
Screen Shot 2020-10-09 at 12 19 54 PM


kennylam (approved):
Looks good thanks @ariellalgilmore!

ariellalgilmore:
thank you @kennylam!!!!!

Issue #3922, opened 2020-09-11
Enable feature flag ability in web components


jeffchew:
#### User Story

> As a `[user role below]`:
IBM.com Library adopter

> I need to:
have the ability to use experimental features in IBM.com Library web components

> so that I can:
test out features quickly before they are released for general use

#### Additional information

 - Current feature flag functionality in React: https://github.com/carbon-design-system/ibm-dotcom-library/blob/master/packages/react/docs/feature-flags.md
- Should probably use a similar approach where they are enabled via environment variables
- Should include a `DDS_FLAGS_ALL` as well
- Should create an experimental environment that is deployed with the `feature flag` label in deploy previews

#### Acceptance criteria
 - [x] Feature flag mechanism in web components package
 - [x] Added corresponding documentation on how to use
 - [x] Add link to feature flag documentation in web components storybook (similar to React: https://ibmdotcom-react.mybluemix.net/?path=/docs/overview-feature-flags--page)
 - [x] Update the environment dashboard
 - [x] Update Github actions to deploy to test environment on change to `master`
 - [x] Update to Jenkins to deploy feature flag version of web components in deploy previews with `feature flag` label in PR

PR #4220, opened 2020-10-13
feat(feature-flag): adding feature flag capability in web components

jeffchew:
### Related Ticket(s)

Refs #3922 

### Description

This PR includes the ability to add and utilize feature flags within
Web Components. As part of this support, the following are included:

- Github Action for deploying experimental environment
- Feature Flag documentation added for Web Components and reference in
Storybook
- `.env.example` file added
- `DDS_FLAGS_ALL` flag added
- Added `doctoc` support in web components package
- Added feature flag `DDS_CALLOUT_DATA` and added to the callout-data component

In addition, deploy previews can be created now for both React and Web components when adding the `feature flag` label to PRs.

### Changelog

**New**

- `feature-flag.ts` and corresponding documentation
- `DDS_CALLOUT_DATA` for web component version

**Changed**

- Updated feature flag documentation for React to include storybook usage
- Put `callout-data` behind a feature flag


asudoh (approved):
LGTM 👍 - Thanks @jeffchew!

jeffchew:
@asudoh thank you for the approval!

It looks like unit tests are failing since I wrapped callout-data with a feature flag. Any thoughts on where the best place to enable `process.env.DDS_CALLOUT_DATA=true` when unit tests are run? I think we should have a general place for this in case we have multiple active experiments.

asudoh:
> Any thoughts on where the best place to enable process.env.DDS_CALLOUT_DATA=true when unit tests are run?

A couple of ways I can think of:

* Dependency injection of feature flag file (Ref: https://webpack.js.org/plugins/normal-module-replacement-plugin/)
* Force setting the environment variable: (Ref: https://webpack.js.org/plugins/define-plugin/)

We have a WebPack configuration used in unit tests: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.11.0/packages/web-components/tests/karma.conf.js#L63-L157

BTW an additional observation: Seems that the custom element is defined even if the feature flag is not set. To avoid that, we can avoid using `@customElement` decorator and use raw [`customElements.define()`](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/define) instead.

jeffchew:
> > Any thoughts on where the best place to enable process.env.DDS_CALLOUT_DATA=true when unit tests are run?
> 
> A couple of ways I can think of:
> 
> * Dependency injection of feature flag file (Ref: https://webpack.js.org/plugins/normal-module-replacement-plugin/)
> * Force setting the environment variable: (Ref: https://webpack.js.org/plugins/define-plugin/)
> 
> We have a WebPack configuration used in unit tests: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.11.0/packages/web-components/tests/karma.conf.js#L63-L157
> 
> BTW an additional observation: Seems that the custom element is defined even if the feature flag is not set. To avoid that, we can avoid using `@customElement` decorator and use raw [`customElements.define()`](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/define) instead.

Thank you for the feedback! I went with `DefinePlugin`, and also added it to the documentation so anyone adding will know to add it there. I also switched over to the raw `customElements.define()`.

Issue #4074, opened 2020-09-29
React: On Masthead - search bar, enter keyword and delete it by mouse select, search bar closes


Praveen-sr:
## Detailed description

Issue: On Masthead - search bar, enter keyword and delete it by mouse select, search bar closes

## Additional information

![Search_del.gif](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/c4bb143e-0cb9-464c-9f36-19b6d1066224)

mcastre:
@RobertaJHahn Hi! Id like to work on this issue.

jeffchew:
@mcastre thank you for participating in `Hacktoberfest`! Looking forward to seeing your PR and feel free to reach out in slack with any questions!

mcastre:
@jeffchew PR has been opened. https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/4221

jeffchew:
> @jeffchew PR has been opened. #4221

Thank you, our team will review this! Want to also quickly check, are you an IBMer?

mcastre:
@jeffchew not exactly sure what you mean by an IBMer

jeffchew:
> @jeffchew not exactly sure what you mean by an IBMer

Sounds like you're not then! Would you be able to fill out the following CLA (contributor license agreement):

[Carbon for IBM.com CLA individual.pdf](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/files/5378291/Carbon.for.IBM.com.CLA.individual.pdf)

Can you also e-mail to digdes@us.ibm.com in additional to mailing to us?

We will start reviewing this PR, thank you again!

mcastre:
Sure! Do you want me to fill out the pdf and email it to that email address?

jeffchew:
Yes please, thank you!

mcastre:
@jeffchew ok i sent the agreement to that email

jeffchew:
@mcastre thank you, we received it! 

Our team reviewed your PR and there was some feedback, can you take a look?

PR #4221, opened 2020-10-13
chore(React Masthead): Keep Search Bar Active if User Selects Text

mcastre:
### Related Ticket(s)

Resolves [#4074](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/4074)

### Description

Check if state.val has a length in handleClickOutside() before dispatching 'setSearchClosed'

### Changelog

**New**

- {{new thing}}

**Changed**

- {{changed thing}}

**Removed**

- {{removed thing}}


kennylam (commented):
@mcastre Thanks for contributing! One issue is that clicking outside the search will close the autosuggestions and keep the input open.

image

mcastre:
@kennylam so im not sure what fix you are looking for. In the original issue description, the problem is that the user is moving their mouse outside of the search box to select text, and once the user clicks off, the search bar disappears. However, this was how the search bar was originally intended to behave, as outlined in the comments on line 170 in react/src/components/Masthead/MastheadSearch.js. 

With my PR, the fix i implemented keeps the search bar open if there is text entered in search. If the user didnt want to search anymore, they have multiple ways to close it, the X icon, clicking outside, or clearing the text, which is how most search bars work out in the world. Im not exactly sure what the expected behavior is?

kennylam:
> If the user didnt want to search anymore, they have multiple ways to close it, the X icon, **clicking outside**,

This is what I meant. Clicking outside no longer closes the search if there is an input value, it only clears the autosuggestions.

@wonilsuhibm @oliviaflory Are you ok with this behavior?

wonilsuhibm:
@kennylam Product Discovery team confirmed that this is the desired behavior. Unless @oliviaflory has other concerns, I agree with this request.

kennylam (approved):
LGTM thanks @mcastre!

Issue #4224, opened 2020-10-14
Footer: IBM logo cannot be seen in smaller breakpoint


asudoh:
## Detailed description

> Describe in detail the issue you're having.

At [our footer CodeSandbox example](https://codesandbox.io/s/github/carbon-design-system/carbon-for-ibm-dotcom/tree/master/packages/web-components/examples/codesandbox/components/footer), the IBM logo cannot been seen in smaller breakpoint (<1056px).

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

A bug.

> Is this issue related to a specific component?

Yes, footer.

> What did you expect to happen? What happened instead? What would you like to
> see changed?

The IBM logo can be seen with all breakpoints.

> What browser are you working in?

Chrome 86.

> What version of Carbon for IBM.com are you using?

Latest `master`.

## Additional information

The cause seems to be wrong `width`/`height` attributes in the `<svg>` tag for IBM logo.

PR #4226, opened 2020-10-14
fix(footer): ensure IBM logo icon is visible

asudoh:
### Related Ticket(s)

Refs #4224.

### Description

Fixes the rendering issue of IBM logo icon in footer by fixing wrong `width` and `height` attributes.

Also fixes wrong content in the `<title>` attribute in corresponding SVG, by avoiding to bake it in the `lit-html` SVG template for the IBM logo icon. The approach is taken because the `<title>` content is not the design asset, and may have to be translated in future. Due to such reason, omitted some other attributes from the `lit-html` SVG template for the IBM logo icon, too.

### Changelog

**Changed**

- Fixes to IBM logo icon:
  - Add right `width` and `height` attributes
  - Omit `<title>` and some attributes from `lit-html` SVG template


kennylam (approved):
Looks good thanks @asudoh!

kennylam:
@asudoh Looks like the only thing is the snapshot is outdated.

PR #3933, opened 2020-09-14
fix(masthead-search): activate/deactivate on focus/blur

asudoh:
### Related Ticket(s)

Refs #3882.

### Description

This change activates the search box when it gets focus, and deactivates the search box when it loses focus.

### Changelog

**New**

- A mechanism to activate the search box when it gets focus, and deactivate the search box when it loses focus.

**Changed**

- A change to consolidate reactive properties in the file to a chunk.
- Fix for missing custom events in the doc.

PR #4227, opened 2020-10-14
fix(masthead-search): re-apply #3932

asudoh:
### Related Ticket(s)

Refs #3883.

### Description

Re-applies #3932 as #3933 ended up with erasing #3932.

### Changelog

**Changed**

- Re-applies #3932 as #3933 ended up with erasing #3932.


kennylam (approved):
Looks good thanks @asudoh!

PR #4231, opened 2020-10-15
chore(ci): upgrade action-ibmcloud-cf

asudoh:
### Description

This change upgrades `action-ibmcloud-cf` GitHub action so that we can use CF `7.x` and its rolling app deployment feature.

### Changelog

**Changed**

- Upgrades `action-ibmcloud-cf` to `v1.3.0`.


jeffchew (approved):
Thanks @asudoh !

Issue #3848, opened 2020-09-02
Component: Lightbox Media Viewer - Allow title & description to be customized when using video


rsebade:
### The problem
The current component loads the title and description directly from media center via the API. As an adopter of the library we would like the ability to tailor these content sections to our users, rather than the default title and description provided.

### The solution
Allow the user to pass props in to override the title and description provided by the API.

RobertaJHahn:
@rsebade would you be interested in making this change and contributing it back?

RobertaJHahn:
Roberta will schedule a meeting to talk to Ryan about contribution model.  Ryan, Jeff invitees.

RobertaJHahn:
Meeting scheduled Monday, Sep 28.

RobertaJHahn:
Ryan has volunteered to make these changes.  Assigned the issue to him.  He will work on it in sprint 20-20.

RobertaJHahn:
@rsebade I am moving this to the next sprint.  Have you worked on this feature request?  We link the PR's to the issues so that we know when dev has finished their work.

rsebade:
Hey @RobertaJHahn , sorry about the delay, I just submitted a pull request!

PR #4238, opened 2020-10-15
feat(LightboxMediaViewer): allow user override vid title/desc

rsebade:
### Related Ticket(s)

[#3848](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/3848)

### Description

Added title and description props to the media object used by the lightbox media viewer. In the image option these props were available, so it brings the video type in line with that. Title and description provided in props will take precedence over API provided title and description.


kennylam (approved):
LGTM, thanks @rsebade!

packages/react/src/components/LightboxMediaViewer/__stories__/LightboxMediaViewer.stories.js
```diff
@@ -89,6 +89,16 @@ EmbeddedVideoPlayer.story = {
         media: {
           src: '1_9h94wo6b',
           type: 'video',
+          title: text(
+            'title (required)',
+            'Curabitur malesuada varius mi eu posuere',
+            groupId
+          ),
+          description: text(
+            'description (required)',
```
jeffchew, line 98: Great work @rsebade ! One thing, the `title` and `description` I believe are `optional` and not `required`, since if this isn't defined it will just pull from the Kaltura API data.

rsebade: @jeffchew Good catch! Just updated this.

Issue #3959, opened 2020-09-15
Add cache busting logic for TranslationAPI and LocaleAPI


jeffchew:
#### User Story

> As a `[user role below]`:
IBM.com User

> I need to:
be sure I have the latest version of the IBM.com Masthead/Footer navigation information

> so that I can:
always have the latest navigation data during my IBM.com browsing experience

#### Additional information

 - The TranslationAPI stashes the data into sessionStorage and serves up from there. Currently, there is no cache clearing policy in place, and assumes it will be refreshed when the browser is restarted.
 - A timestamp can probably be set and added to the saved object in sessionStorage, then compare that to current time
 - Decision is to expire the sessionStorage after 2 hours

#### Acceptance criteria
 - [ ] Adding cache clearing logic to the TranslationAPI and LocaleAPI

IgnacioBecerra:
I'll be splitting this into two PRs, one for Translation and another for Locale.

PR #4240, opened 2020-10-15
feat(translation): introduce cache busting logic

IgnacioBecerra:
### Related Ticket(s)
#3959

### Description

Added a new `timestamp` property within the response json to compare with the current time. If the timestamp is older than two hours, a new response will be fetched with a new, current timestamp. 

Also implemented tests for the new methods.

### Changelog

**New**

- `timestamp` property added to the response json for cache busting purposes


asudoh:
Interesting change @IgnacioBecerra - Seems that the code adds timestamp for each cache data item so you wanted to manage cache TTL per-item? Note that `clearCache()` removes all cache data items.

IgnacioBecerra:
That's the idea, add a timestamp to the cached item, and before using it again, check if it's been longer than two hours. If so, it expires and we retrieve the newest response + update the timestamp. I'll keep that `clearcache()` in mind, thanks!

packages/services/src/services/Translation/__tests__/Translation.test.js
```diff
@@ -76,7 +100,43 @@ describe('TranslationAPI', () => {
...
+    sessionStorageMock.setItem(
+      'dds-translation-us-en',
+      JSON.stringify(oldSession)
+    );
+    const previousSession = JSON.parse(
+      sessionStorageMock.getItem('dds-translation-us-en')
+    );
+
+    // reinitializing import
+    const TranslationAPI = (await import('../Translation')).default;
```
asudoh, line 117: I won't block proceeding this PR with my comment here, but I'd rather see resetting things done via mocks. The behavior of dynamic import tends to be dependent on build toolchain and we may have high chance to get an undefined behavior if it's used as a mean for "resetting".

packages/services/src/services/Translation/__tests__/Translation.test.js
```diff
@@ -76,7 +100,43 @@ describe('TranslationAPI', () => {
...
+
+    // should contain timestamp
+    expect(response).toHaveProperty('timestamp');
+
+    // should not equal old timestamp
+    expect(previousSession.timestamp).not.toEqual(response.timestamp);
+
+    // timestamps should have at least a two hour difference
+    const timeDiff = response.timestamp - previousSession.timestamp,
+      _twoHours = 60 * 60 * 2000;
```
asudoh, line 139: Would you explain what makes the cache elapsed time here greater than two hours?

asudoh (commented):
Great to see this is ready for review @IgnacioBecerra!

packages/services/src/services/Translation/__tests__/Translation.test.js
IgnacioBecerra: Sure! As an example, the cached time in oldResponse is `1337` (a very old timestamp), while `response` would have the current time as `1603206344`. By doing `1603206344 - 1337` we receive `160319297`, which is greater than `7200000` (two hours). This is to ensure that the cache busting code runs for this test, as it is required for the time difference to be greater than two hours. 

If the previous session timestamp was much closer in time to the current one, the difference would be less than two hours, and the `response` received would remain unchanged from the previous session.

packages/services/src/services/Translation/__tests__/Translation.test.js
IgnacioBecerra: Ah I see, I'll be applying the same logic for Locale API, so I'll use mocks there and also add it here. Thanks!

packages/services/src/services/Translation/__tests__/Translation.test.js
asudoh: Thank you for your answer @IgnacioBecerra! I was asking where in the code or in test that makes `timeDiff` greater than two hours, given the test says "`timeDiff` should be greater than two hours". What code sets `response.timestamp` and what code sets `previousSession.timestamp`? Did you make the timestamp deterministic by e.g. mocks (rather than relying on system time that is non-deterministic)?

IgnacioBecerra: Ah I see what you mean! I set `previousSession` in line 9 using `import oldSession from './data/timestamp_response.json';`. 

`timestamp_response.json` uses the same data in `./data/response.json`, with the only difference is that it includes `timestamp: 1337`. This would be the only deterministic timestamp. On the other hand, `response` always contains the current timestamp from the latest response received from `TranslationAPI`.

asudoh: Cool thanks @IgnacioBecerra! Given your response, given we know the timestamps come from mock, why do we need to test the two timestamps between the two mocks? Or was it simply for ensuring we are picking up the right (fresh) data?

IgnacioBecerra: It's the latter, I wanted to make extra sure the logic was working. Since we tested the fresh data in `sessionStorage` was the same as in the `response`, I used the mock to compare times, and had to check if `timeDiff` was greater than two hours.

asudoh: I see, thank you for clarifying @IgnacioBecerra! I suggest more direct way for testing if we pick up the right data. One approach can be object ref. Another approach may be unique key/ID.

packages/services/src/services/Translation/__tests__/Translation.test.js
```diff
@@ -104,18 +109,25 @@ describe('TranslationAPI', () => {
   });
 
   it('should return a json with a recent timestamp', async () => {
+    const mockDate = new Date('2019');
+    const _Date = Date;
+    global.Date = jest.fn(() => mockDate);
+    global.Date.now = _Date.now;
```
asudoh, line 115: Would you try: https://jestjs.io/docs/en/snapshot-testing#2-tests-should-be-deterministic

packages/services/src/services/Translation/__tests__/Translation.test.js
```diff
@@ -131,12 +143,38 @@ describe('TranslationAPI', () => {
...
+    // reinitializing import
+    const TranslationAPI = (await import('../Translation')).default;
+    const response = await TranslationAPI.getTranslation({
+      lc: 'en',
+      cc: 'us',
+    });
+
+    // timestamp should remain unchanged
+    expect(response.timestamp).toEqual(previousSession.timestamp);
+  });
```
asudoh, line 179: Do you have some details on what this case tests?

asudoh (commented):
Some good improvements @IgnacioBecerra!

packages/services/src/services/Translation/__tests__/Translation.test.js
IgnacioBecerra: I tried using that and many other variations before using this approach, but unfortunately this one was the only one that actually had any effect on `response`. Not sure what is causing it though, the implementations don't differ much from another.

packages/services/src/services/Translation/__tests__/Translation.test.js
IgnacioBecerra: We use `responseSuccess` that was just recently refreshed as the cached data. Since it's very recent (just ran), then running the API again shouldn't change the cached data again -- it simply retrieves it. Here we just check if the timestamp remains the same from the previous session.

packages/services/src/services/Translation/Translation.js
```diff
@@ -124,10 +132,12 @@ class TranslationAPI {
           })
           .then(response => this.transformData(response.data))
           .then(data => {
-            sessionStorage.setItem(
-              `${_sessionTranslationKey}-${country}-${lang}`,
-              JSON.stringify(data)
-            );
+            data['timestamp'] = new Date().getTime();
```
asudoh, line 135:
```suggestion
            data['timestamp'] = Date.now();
```

packages/services/src/services/Translation/__tests__/Translation.test.js
```diff
@@ -76,7 +105,76 @@ describe('TranslationAPI', () => {
...
     });
-
     expect(response).toEqual(responseSuccess);
   });
+
+  it('should return a json with a recent timestamp', async () => {
+    const mockDate = new Date('2019');
+    const _Date = Date;
+    global.Date = jest.fn(() => mockDate);
+    global.Date.now = _Date.now;
```
asudoh, line 115: Going away from `new Date()` for generating timestamp seems to allow us to test with less mock like below:

```suggestion
    const mockDate = 1546300800000; // Epoch time of January 1, 2019 midnight UTC
    global.Date.now = jest.fn(() => mockDate);
```

packages/services/src/services/Translation/__tests__/Translation.test.js
```diff
@@ -76,7 +105,76 @@ describe('TranslationAPI', () => {
...
+    );
+
+    // newest response and storage data should match
+    expect(response).toEqual(newSession);
+
+    // should contain timestamp
+    expect(response).toHaveProperty('timestamp');
+
+    // should equal mock timestamp
+    expect(response.timestamp).toEqual(mockDate.valueOf());
```
asudoh, line 147:
```suggestion
    expect(response.timestamp).toEqual(mockDate);
```

packages/services/src/services/Translation/__tests__/Translation.test.js
asudoh: I see, given we don't (and won't) change the timestamp of existing cache, probably we wanted to simply see if the unique ID of `response` equals to one of `previousSession`?

asudoh (commented):
Thank you for your response to my earlier question @IgnacioBecerra!

IgnacioBecerra:
@asudoh The unique ID for `responseSuccess` is `TRANSLATION_FRESH`, given that it was just created. If it's been requested again within two hours, it changes to `TRANSLATION_UNCHANGED` (changed it from `TRANSLATION_OLD`). 

Since it's the only thing that differs between these two responses, I checked the ID change, deleted the IDs from their jsons, and compared the rest in the latest commit. Also applied your suggestions, thanks!

jeffchew:
@IgnacioBecerra did something happen to this branch? There's now 94 changed files.

IgnacioBecerra:
@jeffchew I merged master into this branch to resolved the merge conflict, so the changed files should be from all the recent commits from master.

jeffchew:
> @jeffchew I merged master into this branch to resolved the merge conflict, so the changed files should be from all the recent commits from master. I hadn't noticed the 94 files, but I think there's probably a better way to do this

It shouldn't be this many files, just the ones you changed so I think the merge might not have happened correctly.

IgnacioBecerra:
@jeffchew Resolved the issue! Back to only three files changed.

asudoh:
> @asudoh The unique ID for `responseSuccess` is `TRANSLATION_FRESH`, given that it was just created. If it's been requested again within two hours, it changes to `TRANSLATION_UNCHANGED` (changed it from `TRANSLATION_OLD`). 

Could you fix this? Given we set the unique ID solely for testing purpose and we use it for the cache record identify, the ID being changed won't make sense. Thanks!

packages/services/src/services/Translation/__tests__/Translation.test.js
asudoh: Any update here...? Thanks!

packages/services/src/services/Translation/__tests__/Translation.test.js
```diff
@@ -76,7 +100,71 @@ describe('TranslationAPI', () => {
...
+    expect(response).toHaveProperty('timestamp');
+
+    // should equal mock timestamp
+    expect(response.timestamp).toEqual(mockDate);
+    expect(response.id).toEqual('TRANSLATION_FRESH');
+
+    // timestamps should have at least a two hour difference
+    const timeDiff = response.timestamp - previousSession.timestamp,
+      _twoHours = 60 * 60 * 2000;
+    expect(timeDiff).toBeGreaterThan(_twoHours);
```
asudoh, line 146: Per discussion, these lines are no longer needed:

```suggestion
```

asudoh (commented):

packages/services/src/services/Translation/__tests__/Translation.test.js
```diff
@@ -23,6 +25,31 @@ jest.mock('axios', () => {
...
+const sessionStorageMock = (() => {
+  let cache = {};
+
+  return {
+    getItem(key) {
+      return cache[key] || null;
+    },
+    setItem(key, value) {
+      value = JSON.parse(value);
+      value['id'] = 'TRANSLATION_FRESH';
```
asudoh, line 37: The ID is better set from test cases:

```suggestion
```

packages/services/src/services/Translation/__tests__/Translation.test.js
```diff
@@ -76,7 +98,69 @@ describe('TranslationAPI', () => {
...
+    sessionStorageMock.setItem(
+      'dds-translation-us-en',
+      JSON.stringify(oldSession)
+    );
+
+    const previousSession = JSON.parse(
+      sessionStorageMock.getItem('dds-translation-us-en')
+    );
+
+    expect(previousSession.id).toEqual('TRANSLATION_FRESH');
```
asudoh, line 119: Given we are not testing `sessionStorageMock` here, I don't think this is needed:

```suggestion
```

packages/services/src/services/Translation/__tests__/Translation.test.js
```diff
@@ -76,7 +98,69 @@ describe('TranslationAPI', () => {
...
   });
+
+  it('should return a json with a recent timestamp', async () => {
+    const mockDate = 1546300800000; // Epoch time of January 1, 2019 midnight UTC
+    global.Date.now = jest.fn(() => mockDate);
+
+    // using very old cached session
+    sessionStorageMock.setItem(
+      'dds-translation-us-en',
+      JSON.stringify(oldSession)
```
asudoh, line 112:
```suggestion
      JSON.stringify(Object.assign(oldSession, { id: uniqueId }))
```

packages/services/src/services/Translation/__tests__/Translation.test.js
```diff
@@ -76,7 +98,69 @@ describe('TranslationAPI', () => {
         origin: 'https://ibm.com',
       },
     });
-
+    delete response['id'];
     expect(response).toEqual(responseSuccess);
   });
+
+  it('should return a json with a recent timestamp', async () => {
+    const mockDate = 1546300800000; // Epoch time of January 1, 2019 midnight UTC
```
asudoh, line 106: Also please refer to how the ID is set via `Object.assign()`:

```suggestion
    const uniqueId = `__translation_cache__${Math.random().toString(36).slice(2)}`;
    const mockDate = 1546300800000; // Epoch time of January 1, 2019 midnight UTC
```

packages/services/src/services/Translation/__tests__/Translation.test.js
```diff
@@ -76,7 +98,69 @@ describe('TranslationAPI', () => {
...
+    const response = await TranslationAPI.getTranslation({
+      lc: 'en',
+      cc: 'us',
+    });
+
+    const newSession = JSON.parse(
+      sessionStorageMock.getItem('dds-translation-us-en')
+    );
+
+    response['id'] = newSession.id;
```
asudoh, line 130: It seems that we are testing if the data comes from the cache. If so, the cache should have its ID assigned already and thus we don't set one to `response`:

```suggestion
```

packages/services/src/services/Translation/__tests__/Translation.test.js
```diff
@@ -76,7 +98,69 @@ describe('TranslationAPI', () => {
...
+
+    // newest response and storage data should match
+    expect(response).toEqual(newSession);
+
+    // should contain timestamp
+    expect(response).toHaveProperty('timestamp');
+
+    // should equal mock timestamp
+    expect(response.timestamp).toEqual(mockDate);
+    expect(response.id).toEqual('TRANSLATION_FRESH');
```
asudoh, line 140: It seems that we are testing if the data comes from the cache. If so, ID check is the only assertion we need:

```suggestion
    expect(response.id).toEqual(uniqueId);
```

packages/services/src/services/Translation/__tests__/Translation.test.js
```diff
@@ -76,7 +98,69 @@ describe('TranslationAPI', () => {
...
+    );
+
+    const response = await TranslationAPI.getTranslation({
+      lc: 'en',
+      cc: 'us',
+    });
+
+    // body and timestamp should remain unchanged
+    expect(response).toEqual(previousSession);
+  });
```
asudoh, line 161: Given we don't have any code to change the timestamp of cache, this test case seems not necessary:

```suggestion
```

asudoh (commented):

asudoh (approved):
LGTM 👍 - Thanks @IgnacioBecerra for all the updates! 🎉

Issue #2723, opened 2020-06-11
Component: CTA - Keyboard - After the model page opens, focus not set on close icon.


Praveen-sr:
## Detailed description

Issue :- After the model page opens, focus not set on close icon.

Expected :- As soon as model page opens, focus should set on the close button of the model page.

RobertaJHahn:
@kennylam Moving this work to the next sprint as it is a sev 4.

PR #4242, opened 2020-10-16
fix(modal): add primary focus on open

kennylam:
### Related Ticket(s)

#2723 

### Description

This focuses the close button in `LightboxMediaViewer` when the modal first opens.

### Changelog

**New**

- add `selectorPrimaryFocus` to `ExpressiveModal`


asudoh:
Thanks @kennylam for getting into this! I'm not familiar enough on how `<ExpressiveModal>` should work, but just thought it'd be nice if we can limit the change in behavior to light box only; Given for many kind of modals the primary focus tends to be on the primary button.

kennylam:
@asudoh Yes, you're correct. Currently `ExpressiveModal` simply imports Carbon's `ComposedModal` and the only component of ours that uses it is `LightboxMediaViewer`. If it becomes problematic we can make close the default prop but open it up to be user-defined.

The larger issue I'm wondering about is whether we should be using Carbon's `ComposedModal` vs `ModalWrapper`? It seems `ModalWrapper` also includes passing focus back to the triggering element onClose, which `ComposedModal` does not have.

asudoh:
> @asudoh Yes, you're correct. Currently `ExpressiveModal` simply imports Carbon's `ComposedModal` and the only component of ours that uses it is `LightboxMediaViewer`. If it becomes problematic we can make close the default prop but open it up to be user-defined.

Thank you for clarifying @kennylam! And we don't expect our users to use `<ExpressiveModal>` directly?

> The larger issue I'm wondering about is whether we should be using Carbon's `ComposedModal` vs `ModalWrapper`? It seems `ModalWrapper` also includes passing focus back to the triggering element onClose, which `ComposedModal` does not have.

`<ModalWrapper>` is a demonstration-only component that was shipped by mistake. So our choice is between `<Modal>` and `<ComposedModal>`. `<Modal>` is a simple choice, `<ComposedModal>` provides more flexibility in content.

The reason `<ModalWrapper>` being demonstration-only component is that interaction with launcher button has many application-only use cases which does not make the problem area abstracted. That said, it's a purview of application code to handle "focusing back to the launcher button on close".

kennylam:
Ah, I see, thanks for the explanation. It can get a little confusing with all the different Carbon modal versions.

> Thank you for clarifying @kennylam! And we don't expect our users to use `<ExpressiveModal>` directly?
I don't think so, no. I'm not very familiar with `ExpressiveModal` and how it even came about. But from what I see so far, it seems more like it should be an internal wrapper for Carbon modals and not exposed for direct use (but currently it is).

asudoh (approved):
Per discussion, the change seems safe enough - Thanks @kennylam!

PR #4248, opened 2020-10-18
chore(devenv): add mising @storybook/addon-info

asudoh:
### Description

This change adds `@storybook/addon-info` dependency to `packages/styles`. `packages/styles` seems to require this, but was not explicitly declared as `devDependencies`.

### Changelog

**New**

- `@storybook/addon-info` dependency to `packages/styles`.


kennylam (approved):
Looks good thanks @asudoh!

Issue #3258, opened 2020-07-22
Put Redux store in a separate package


asudoh:
#### User Story

> As a `[user role below]`:

`ibmdotcom-react`, `ibmdotcom-web-components` or `ibmdotcom-vanilla` maintainer (as well as the user),

> I need to:

Have a non-framework-dependent way of getting application state of loading data from our locale/translation services (e.g. loading in progress, loading failed),

> so that I can:

* Project such state to skeleton, spinner, and/or notifications, regardless of what JavaScript framework I use.
* Avoid having to implement such logic in React Component, Web Components or vanilla component.

#### Additional information

* Pending implementation of Web Components version of locale modal (#3185) takes up [large portion of its code](https://github.com/carbon-design-system/ibm-dotcom-library/blob/b91fb76d76a126557ba9f26c93e7ade72d0a82c0/packages/web-components/src/components/locale-modal/locale-modal-container.ts#L100-L281) to deal with such logic.
* React version of locale modal had a problem dealing with service API calls considering lifecycle management. It was patched (#2368), but still does not have solid way to project loading/failure state to the UI.
* Good to avoid having to duplicate above logic for each JavaScript frameworks.

#### Acceptance criteria
 - [ ] A way to manage application state for service calls
 - [ ] Above logic is framework-independent (and separated from framework repositories)
 - [ ] Above logic is separated from UI component code

jeffchew:
@RobertaJHahn can we add this to Monday's agenda to talk about? Want to think about sizing, priority, and timing on the work.

jeffchew:
@RobertaJHahn moving this further out, we have other pressing things to look at at the moment.

asudoh:
In tne engineering huddle in October 7, 2020, we decided go with an approach of:

1. Making the API internal (either in a hard way or in a soft way like naming)
2. Ensure it can be used across our offerings for different JavaScript frameworks, i.e. Web Components and React

One approach that is possible is creating an internal package in our monorepo and a build script that copies the content over to Web Components/React repositories, in a similar manner as how we use `carbon-components-react` packages in our React code.

PR #4252, opened 2020-10-19
chore(services-store): move to private separate package

asudoh:
### Related Ticket(s)

Refs #3258.

### Description

Moves `packages/web-components/src/globals/services-store` to a separate internal package so:

1. We don't have to maintain `services-store` code as public API
2. It can be reused for React

Given the new package is an internal one that's never published to NPM, the build process copies the new package content to
`src/internal/vendor/@carbon/ibmdotcom-services-store` directory and Web Components code refers to such directory in the `import`s, instead of `imports` referring to a NPM package.

Also, removed custom Redux store examples from `README`s for now. Those examples uses our Redux code that we don't let our users use, for now.

### Changelog

**New**

- `@carbon/ibmdotcom-services-store` internal package, the code is moved from `packages/web-components/src/globals/services-store`.

**Removed**

- Examples using our Redux store from `README`.


kennylam (approved):
Looks great thanks @asudoh!

PR #4267, opened 2020-10-20
chore(services-store): added null check of services-store vendor folder

jeffchew:
### Related Ticket(s)

#3258 

### Description

The motivation of this PR is that a `yarn reset` would fail if the
vendor folder for the service store doesn't exist, primarily during a
`gulp clean`. In order to allow the `gulp clean` to continue, it cannot
have an exit code 1 if the folder doesn't exist, and just continue with
the clean.

The full `yarn install`/`yarn build` works afterwards, as the link works
after that.

### Changelog

**Changed**

- added null check for `servicesStoreESSrcDir` in gulp `config.js`

### DIFF
--- a/packages/web-components/gulp-tasks/config.js
+++ b/packages/web-components/gulp-tasks/config.js
@@ -28,13 +28,6 @@ const { browser: browsers, spec: specs, ...rest } = commander
   .parse(process.argv);
 const cloptions = { browsers: Array.from(browsers), specs: Array.from(specs), ...rest };
 
-let servicesStoreLink;
-try {
-  servicesStoreLink = path.resolve(path.dirname(require.resolve('@carbon/ibmdotcom-services-store/package.json')), 'es');
-} catch (err) {
-  servicesStoreLink = null;
-}
-
 module.exports = {
   ENV_PRODUCTION: 'production',
   cloptions,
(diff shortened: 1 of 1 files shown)


packages/web-components/gulp-tasks/config.js
```diff
@@ -28,6 +28,13 @@ const { browser: browsers, spec: specs, ...rest } = commander
   .parse(process.argv);
 const cloptions = { browsers: Array.from(browsers), specs: Array.from(specs), ...rest };
 
+let servicesStoreLink;
+try {
+  servicesStoreLink = path.resolve(path.dirname(require.resolve('@carbon/ibmdotcom-services-store/package.json')), 'es');
+} catch (err) {
+  servicesStoreLink = null;
+}
+
```
asudoh, line 37:
```suggestion
```

packages/web-components/gulp-tasks/config.js
```diff
@@ -38,7 +45,7 @@ module.exports = {
   sassDestDir: 'scss',
   tasksDir: 'gulp-tasks',
   testsDir: 'tests',
-  servicesStoreESSrcDir: path.resolve(path.dirname(require.resolve('@carbon/ibmdotcom-services-store/package.json')), 'es'),
+  servicesStoreESSrcDir: servicesStoreLink,
```
asudoh, line 48:
```suggestion
  servicesStoreESSrcDir: path.resolve(__dirname, '../../servies-store/es'),
```

asudoh (commented):
Thank you to trying to fix the build problem @jeffchew! Unfortunately we cannot leave `serviesStoreESSrcDIr` un-defined, it'll essentially defer the problem later on. I suggest getting the configuration problem solved (by establishing the symlink as I talked offline), but if you want to get out of the error before setting that up, using different mean from using `require.resolve()` will be the way to go.

kennylam (approved):
Looks good to me thanks @jeffchew

Issue #4244, opened 2020-10-16
Canary releases not publishing when there is an active conventional prerelease


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

When `Carbon for IBM.com` is in code freeze (release candidates), canary releases are not producing in the correct semver or refCount, causing publishing to fail.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

Lerna

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Publish latest canaries with the proper semver when there is an existing prerelease cycle.

> What browser are you working in?

N/A

> What version of Carbon for IBM.com are you using?

`v1.12.0-rc.0`

## Steps to reproduce the issue

1. See logs for canary releases: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/actions?query=workflow%3A%22publish-canary+%28Publish+Canary+release+to+NPM%29%22

## Additional information

- Issue also opened on the lerna repo: https://github.com/lerna/lerna/issues/2741

PR #4273, opened 2020-10-20
fix(canary): adding github run_id to the publish preid

jeffchew:
### Related Ticket(s)

Refs #4244 

### Description

We were running into issues of incorrect canary version bumps and
refCounts, so adding the unique github run_id should solve for this when
 adding to the `preid`

### Changelog

**Changed**

- `publish-canary.yml` and `publish-nightly.yml` updated to include `github.run_id`.


kennylam (approved):
LGTM!

PR #4278, opened 2020-10-21
docs(examples): force 16 columns Carbon grid

asudoh:
### Description

This change enables 16 columns Carbon grid in Web Components examples as well as Web Components storybook. Accordion to @annawen1, Carbon for IBM.com prefers 16 columns Carbon grid.

`carbon-components` library sets 12 columns grid the default presumably for historical compatibility reasons as opposed `@carbon/grid` library's default 16 columns grid.

Also removes an unused CSS ruleset for `.bx--card-group__cards` selector. The original commit that introduced such ruleset (originally `.bx--card-section__cards`) had no element matching to that selector. Such element does not exist in latest `master`, either, and applying such ruleset to `<dds-card-group>` tag caused adverse effect to the styling.

### Changelog

**Changed**

- Enables 16 columns Carbon grid in Web Components examples as well as Web Components storybook.

**Removed**

- Unused CSS ruleset for `.bx--card-group__cards` selector.


annawen1 (commented):
Thanks @asudoh! One thing is some components like the `CardGroup` look off with this new grid change, probably just need to adjust the grid classes in the stories. Should we include that in this PR or open a separate one?

asudoh:
OK good point @annawen1 - Will take a look at it tomorrow.

asudoh:
@annawen1 Found that it was caused by a redundant CSS ruleset for `.bx--card-group__cards` (and `<dds-card-group>` tag) that is never used by React. Fixed.

annawen1 (approved):
Thanks @asudoh!

Issue #4264, opened 2020-10-20
Web component: Table of content - Navigate using Jump to,  heading of the section is not seen.


Praveen-sr:
## Detailed description
On Mobile and Tab

Issue: When user navigates using Jump to,  heading of the section is not seen.

Note: Working fine on react 

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/998980de-1b75-4832-93aa-5d6a553ac422)

PR #4297, opened 2020-10-26
fix(table-of-contents): ensure anchor is visible upon jump-to-content

asudoh:
### Related Ticket(s)

Refs #4264.

### Description

Introduces `bx--tableofcontents__contents` CSS class that ensures `<a name="anchorname">` is visible upon getting scroll to by selection of an item in `<dds-table-of-contents>`.

### Changelog

**New**

- `bx--tableofcontents__contents` CSS class.

**Removed**

- Code setting the margin offset of `InsersectionObserver` for scroll hijacking code. It's effectively replaced with the new `bx--tableofcontents__contents` CSS class.


packages/web-components/src/components/table-of-contents/__stories__/README.stories.mdx
```diff
@@ -17,25 +17,36 @@ import contributing from '../../../../../../docs/contributing-license.md';
...
-  <a name="1" data-title="Section - 1"></a>
-  <h3>Section - 1</h3>
-  <p>Section 1 content</p>
-  <a name="2" data-title="Section - 2"></a>
-  <h3>Section - 2</h3>
-  <p>Section 2 content</p>
-  <a name="3" data-title="Section - 3"></a>
-  <h3>Section - 3</h3>
-  <p>Section 3 content</p>
+  <div class="bx--tableofcontents__contents">
```
jeffchew, line 39: @asudoh are we replying on adopters to add this wrapper class? Anyway this element can be transcluded instead?

asudoh: > @asudoh are we replying on adopters to add this wrapper class?

Yes. Seems that we need this kind of wrapper class for React, too, given navigating directly to hashbang URL fails there.

> Anyway this element can be transcluded instead?

Do you have any more details here...? Thanks!

jeffchew: What I mean is can we add the wrapper div for the user instead of asking them to add it?

asudoh: Thank you for clarifying @jeffchew! I thought of making it automatic, but the problem is that our style (shadow DOM style) doesn't have direct control over the content style, and we allow arbitrary content for `<dds-table-of-contents>`. Therefore it looks like it's more prudent for us to clearly document that user has to use our CSS class to adjust the anchor position. I used below wording as an explanation of what the example code does, but good to hear better wording, etc.:

> 💡 `bx--tableofcontents__contents` class in the content `<div>` ensures the `<h3>` is visible by navigating to a hashbang URL (e.g. `https://yoursite.com/yourpage#2`).

kennylam (approved):
Looks good thanks @asudoh!

Issue #4262, opened 2020-10-20
Web component: Table of content - top padding should be 64px but its more than 64px.


Praveen-sr:
## Detailed description

Issue: Table of content - top padding for the slide bar should be 64px according to the specs but its more than 64px.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/41aa2d09-054f-430c-a8ce-b4adc4ca08e0)

PR #4298, opened 2020-10-26
fix(table-of-contents): fix header space

asudoh:
### Related Ticket(s)

Refs #4262.

### Description

This change ensures the area for the heading content is hidden unless there is its content actually.

### Changelog

**New**

- UI integration test for below change.

**Changed**

- A change to ensure the area for the heading content is hidden unless there is its content actually.


kennylam (approved):
Looks good thanks @asudoh!

Issue #4030, opened 2020-09-23
Web component: Lightbox Media - Close icon and the scroll bar are overlapping @1050px


Praveen-sr:
## Detailed description

Issue: Lightbox Media - Close icon and the scroll bar are overlapping @1050px

Note: Working on react.

## Additional information

Actual :
![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/43df7148-3131-4d0f-8719-1c6367ec59b1)

Expected:
![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/ddc10188-0583-48b5-945f-71d79588d781)

PR #4299, opened 2020-10-26
fix(modal): remove close button offset for full-width mode

asudoh:
### Related Ticket(s)

Refs #4030.

### Changelog

**Removed**

- The offset of modal close button for full width mode.


jeffchew:
@asudoh This looks ok to me, but looks like there's a new option for `expressive size`, want to run this by design as well to see if having this option looks right, and also if this is the right name for the attribute.

@wonilsuhibm @oliviaflory you can go to:
https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/4299/index.html?path=/story/components-modal--default

Then there is a new knob called `expressiveSize`.

asudoh:
Thank you for your feedback @jeffchew! For a bit of background on `expressive-size` attribute; I was originally meant to be `size`, but it conflicts with Carbon core `<bx-modal>` with different semantics, which is T-shirt size of modal.

kennylam:
@asudoh Thanks for this. Only thing is that it seems the scrollbar and close button still overlap on both modal sizes. Is this PR not meant to address that?

oliviaflory:
Hmm, the name is a little odd now because Carbon core has 'modal' and we have another 'modal'.

Should ours be called `Modal expressive` with a `Default` and `Full width` size? 

What makes it "expressive" is that the padding is increased from Carbon core and uses regular buttons instead of the fluid buttons.

Just a thought. What do you think @wonilsuhibm @asudoh @jeffchew

asudoh:
> Should ours be called Modal expressive with a Default and Full width size?

Makes sense to me!

asudoh:
> @asudoh Thanks for this. Only thing is that it seems the scrollbar and close button still overlap on both modal sizes @1050px. Is this PR not meant to address that?

@kennylam My apologies for not getting back to this as soon as I should have. The overlap is removed only for "full width" variant (you can choose it via a knob), as the issue (#4030) is against lightbox media viewer and lightbox media viewer comes with full width mo

oliviaflory:
> > @asudoh Thanks for this. Only thing is that it seems the scrollbar and close button still overlap on both modal sizes @1050px. Is this PR not meant to address that?
> 
> @kennylam My apologies for not getting back to this as soon as I should have. The overlap is removed only for "full width" variant (you can choose it via a knob), as the issue (#4030) is against lightbox media viewer and lightbox media viewer comes with full width mode. @oliviaflory Please let us know the overlap should be fixed for normal mode, too.

@asudoh I'm looking at this on Browser stack, Windows 10, Edge and it looks like the top arrow of the scroll bar is still overlapping with the Close button, it's most apparent when you hover over the close button, the user cannot access the top arrow to scroll back up. Is this what we were intending to fix?

**Lightbox media viewer**
Screen Shot 2020-11-05 at 10 36 29 AM

Screen Shot 2020-11-05 at 10 36 32 AM


**Expressive modal – Expanded**
Screen Shot 2020-11-05 at 10 47 42 AM

Screen Shot 2020-11-05 at 10 47 45 AM


I do think it would be nice for the default expressive modal to fix the overlap as well, it's ok with me if we need to address this in a separate issue.

Screen Shot 2020-11-05 at 10 49 26 AM

cc: @wonilsuhibm

asudoh:
Thank you @oliviaflory for taking a deeper look! From your feedback and your screenshot involving our React version, moved the close button to the top right corner for all expressive modals React/Web Components and pushed down the contents. Given pushing down the contents may not be desirable from design-wise, please let us know if you and @wonilsuhibm have any better ideas wrt how to coordinate the positions of the content and the close button. Thanks!

kennylam:
> @kennylam My apologies for not getting back to this as soon as I should have. The overlap is removed only for "full width" variant (you can choose it via a knob), as the issue (#4030) is against lightbox media viewer and lightbox media viewer comes with full width mode. @oliviaflory Please let us know the overlap should be fixed for normal mode, too.

Got it, thanks for clearing that up @asudoh 👍 

> Thank you @oliviaflory for taking a deeper look! From your feedback and your screenshot involving our React version, moved the close button to the top right corner for all expressive modals React/Web Components and pushed down the contents. Given pushing down the contents may not be desirable from design-wise, please let us know if you and @wonilsuhibm have any better ideas wrt how to coordinate the positions of the content and the close button. Thanks!

@oliviaflory @wonilsuhibm We had the button push down the content at one point and design asked us to remove the extra spacing so the padding was even all around. The way the content/button is laid out i don't think it's possible to avoid that overlap unless we customize the scrollbar (don't recommend this).

wonilsuhibm:
The overlap is not ideal at all but also acceptable as-is for now since the user has other ways to scroll. @oliviaflory please approve this if you agree. If you don't, let me know. :)

oliviaflory (approved):
Looks good to me!

jeffchew:
@asudoh can you update snapshots?

asudoh:
Thanks @oliviaflory and @wonilsuhibm for reviewing, and @kennylam for your input! Updated the snapshot.

PR #4303, opened 2020-10-26
chore(percy): hiding the cookie preferences link in percy

jeffchew:
### Related Ticket(s)

No related issues

### Description

The cookie preferences link is showing up randomly which is causing
false positive reports in Percy. We are already suppressing this in
React, this is doing the same for Web Components.

### Changelog

**Changed**

- Hiding cookie preferences in storybook


ariellalgilmore (approved):
LGTM :) thanks @jeffchew

Issue #4272, opened 2020-10-20
Upgrade dependencies to latest Carbon


jeffchew:
#### User Story

> As a `[user role below]`:
Carbon for IBM.com adopter

> I need to:
load the latest Carbon dependencies in my application

> so that I can:
ensure I am using the latest version of Carbon with Carbon for IBM.com

#### Additional information

 - Enable the labels `package: styles`, `RTL`, and `feature flag` to ensure we are capturing testing across
- each package has the npm script `yarn upgrade-carbon` which can be used for this task

#### Acceptance criteria
 - [ ] All dependencies for react, web-components, and styles are updated to latest Carbon
 - [ ] Activate deploy previews for carbon expressive, web components and react experimental

PR #4316, opened 2020-10-27
chore(upgrade): upgrade carbon deps to v10.22.0

### LINKED ISSUE #4200: React: Masthead - With tab/shift tab key, focus is not going to the profile elements.
## Detailed description

Issue: 
With tab/shift tab key, focus is not going to the "My IBM, Log in" elements.

Expected :
With tab/shift tab key, Focus should go to each and every interactive element and it should be operable thru keyboard.

kennylam:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/4272, https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/4200

### Description

Upgrade Carbon dependencies to latest `v10.22.0`. This also fixes #4200 which was caused by Carbon `v10.21.0`.

### Changelog

**Changed**

- Carbon dependencies upgraded to `v10.22.0`


asudoh (approved):
Thank you for taking this on @kennylam! Could you update the snapshot and check the diff? Thanks!

kennylam:
> Thank you for taking this on @kennylam! Could you update the snapshot and check the diff? Thanks!

Doh! Good catch thanks @asudoh!

packages/styles/package.json
```diff
@@ -38,7 +38,7 @@
     "@carbon/import-once": "10.3.0",
     "@carbon/layout": "10.13.0",
     "@carbon/motion": "10.9.0",
-    "@carbon/themes": "10.20.0",
+    "@carbon/themes": "10.21.0",
     "@carbon/type": "10.16.0",
     "carbon-components": "10.21.0"
```
jeffchew, line 43: @kennylam it doesn't look like all components were updated, did you run the script `yarn upgrade-carbon` to get all of the carbon dependencies?

kennylam: @jeffchew I did. not sure why that one didn't get picked up

Issue #3772, opened 2020-08-28
Masthead | Web Components | RTL enabled has rendering issues


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

The web components masthead has rendering issues in RTL:

<img width="1433" alt="Screen Shot 2020-08-28 at 8 48 05 AM" src="https://user-images.githubusercontent.com/1641214/91562573-6795bd00-e90b-11ea-8969-6ada80f8cbdd.png">

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

Masthead

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Match rendering to React version of the Masthead: http://ibmdotcom-react-rtl.mybluemix.net/?path=/story/components-dotcom-shell--default

> What browser are you working in?

Chrome

> What version of the IBM.com Library are you using?

`v1.10.0`

## Steps to reproduce the issue

1. Go to http://ibmdotcom-web-components-canary.mybluemix.net/iframe.html?id=components-dotcom-shell--default
2. Inspect the page, and under the `<html>` element, add the attribute `<html dir="rtl">`

asudoh:
Hi @jeffchew thank you for reporting, we need to create another Storybook build and test there. For Storybook build the instruction is similar to React (using `STORYBOOK_IBMDOTCOM_WEB_COMPONENTS_USE_RTL=true` environment variable).

asudoh:
The remaining works:

* Create documentation to how to use RTL from our library users' perspective. The one for `carbon-web-components` can be found at: https://web-components.carbondesignsystem.com/?path=/docs/introduction-custom-styles--page#dependency-injection
* (Optional): Create RTL version of `lit-html` CSS build

RobertaJHahn:
@asudoh Are you working on this issue?  Just checking as you have not assigned yourself.

asudoh:
@RobertaJHahn Thank you for double-checking, not at this moment, but I can definitely look into this if this item should be prioritized.

RobertaJHahn:
[...]

PR #4317, opened 2020-10-27
docs(example): introduce RTL example

asudoh:
### Related Ticket(s)

Refs #3772.

### Description

Introduces a CodeSandbox example that demonstrates how our adaptors can create a RTL page with `@carbon/ibmdotcom-web-components`, as well as its corresponding documentation.

Also fixes an issue that the left nav overlay is not flipped for RTL.

### Changelog

**New**

- RTL CodeSandbox example.
- Documentation on how to use RTL.


jeffchew:
Thanks for this @asudoh ! And probably not part of this PR,. but I am thinking if we are continuing our CDN offering we may want to also provide a build for RTL for the dotcom shell as well, maybe something like:

`https://www.ibm.com/common/carbon-for-ibm-dotcom/latest/ibmdotcom-web-components-dotcom-shell.rtl.min.js`

jeffchew (approved):

PR #4325, opened 2020-10-28
feat(locale): introducing cache busting logic

IgnacioBecerra:
### Related Ticket(s)

#3959 

### Description

As done in [TranslationAPI](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/4240), this PR adds a new timestamp property within the response json to compare with the current time. If the timestamp is older than two hours, a new response will be fetched with a new, current timestamp.

Also implemented tests for the new methods.

### Changelog

**New**

- `timestamp` property added to the response json for cache busting purposes


asudoh (approved):

Issue #4331, opened 2020-10-28
React: Language selector - drop down is not operable.


Praveen-sr:
## Detailed description
On iPhone voiceover

Issue:
In iPhone, the dropdown is identified by Voiceover. But it is not operable 

Expected:
Screen reader should be able to Identify the dropdown and the user should be able to navigate through the options and make selection

kennylam:
@RobertaJHahn @Praveen-sr a PR has been opened for this issue https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/4335

Issue #4330, opened 2020-10-28
React: Language selector - dropdown not identified on JAWS shortcut keys


Praveen-sr:
## Detailed description
On JAWS FF

Issue: 
This dropdown is not identified by JAWS quick navigation keys but functions correctly via tab focus. Since it is not blocking functionality and still accessible, @Praveen-sr and Ram have agreed it can wait until the next release and lowering the severity. This only pertains to desktop breakpoints and browsers, since mobile device screen readers correctly identify the selector.

Excepted:
Screen reader should be able to Identify the dropdown with JAWS quick navigation keys.

kennylam:
@RobertaJHahn @Praveen-sr a PR has been opened for this issue https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/4335

Praveen-sr:
@kennylam Re tested on the new PR link, above issue is fixed, Working fine now.

But new issue is found,
Issue: There is no label for this dropdown

Expected: Provide label for the dropdown using html label tag or aria label

PR #4335, opened 2020-10-28
fix(language-selector): update language selector to use select

kennylam:
### Related Ticket(s)

#4330, #4331 

### Description

This change adds Carbon's `Select` component for the footer language selector in mobile/tablet breakpoints. Desktop views will continue to use `ComboBox`.

### Changelog

**Changed**

- `Footer` language selector now uses Carbon's `Select` instead of `ComboBox` for mobile/tablet breakpoints
- updated various styles for new component

### DIFF
--- a/packages/react/src/components/Footer/LanguageSelector.js
+++ b/packages/react/src/components/Footer/LanguageSelector.js
@@ -4,10 +4,11 @@
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
  */
-
-import React, { useState } from 'react';
+import React, { useState, useEffect, useRef } from 'react';
+import ComboBox from '../../internal/vendor/carbon-components-react/components/ComboBox/ComboBox';
 import ddsSettings from '@carbon/ibmdotcom-utilities/es/utilities/settings/settings';
 import PropTypes from 'prop-types';
+import root from 'window-or-global';
 import Select from '../../internal/vendor/carbon-components-react/components/Select/Select';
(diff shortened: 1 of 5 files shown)


packages/react/src/components/Footer/LanguageSelector.js
```diff
@@ -5,11 +5,11 @@
...
  */
 
-import React, { useState, useEffect, useRef } from 'react';
-import ComboBox from '../../internal/vendor/carbon-components-react/components/ComboBox/ComboBox';
+import React, { useState } from 'react';
 import ddsSettings from '@carbon/ibmdotcom-utilities/es/utilities/settings/settings';
 import PropTypes from 'prop-types';
-import root from 'window-or-global';
+import Select from '../../internal/vendor/carbon-components-react/components/Select/Select';
+import SelectItem from '../../internal/vendor/carbon-components-react/components/SelectItem/SelectItem';
```
jeffchew, line 12: I thought the idea was that we would only use select for mobile/tablet, and the combobox remains for desktop?

kennylam: @jeffchew Yep, you are right. Confirmed with @wonilsuhibm that it should use combobox for desktop.

Praveen-sr:
@kennylam Re tested on the new PR link, above issue is fixed, Working fine now.

But new issue is found, 
Issue: There is no label for this dropdown

Expected: Provide label for the dropdown using html label tag or aria label

jeffchew (approved):
LGTM!

Looks like Praveen has some feedback.

RaphaelAmadeu (approved):
lgtm!

Issue #4347, opened 2020-10-29
Web Component: Button Group focus outline extends past button


IgnacioBecerra:
## Detailed description

> Describe in detail the issue you're having.
Whenever a button in `ButtonGroup` is pressed/tabbed upon, the focused outline goes past the button and covers empty space to the right. 

> Is this a feature request (new component, new icon), a bug, or a general
> issue?
Bug

> Is this issue related to a specific component?
ButtonGroup/ButtonGroupItem

## Steps to reproduce the issue

1. Open https://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-button-group--default
2. Press tab until reaching either of the two buttons.

## Additional information

- Default view
![Screen Shot 2020-10-29 at 12 11 07 PM](https://user-images.githubusercontent.com/24970122/97621692-80405300-19e0-11eb-92f6-68a9a3a9e52f.png)

- On smaller screens / column mode
![Screen Shot 2020-10-29 at 12 11 18 PM](https://user-images.githubusercontent.com/24970122/97621693-80d8e980-19e0-11eb-873e-1abde5ea9e93.png)

PR #4348, opened 2020-10-29
fix(button-group): fixed focus outline going past the buttons on web components

IgnacioBecerra:
### Related Ticket(s)

#4347 

### Description

In ButtonGroup, whenever the user focuses on any button (by pressing or tabbing), the focus outline extends past the button as pictured below.

Before:
Screen Shot 2020-10-29 at 12 11 18 PM

After:

Screen Shot 2020-10-29 at 12 23 43 PM

### Changelog

**New**

- added new css properties for `button-group.scss` specific to the web components version

### DIFF
--- a/packages/web-components/src/components/button-group/button-group.scss
+++ b/packages/web-components/src/components/button-group/button-group.scss
@@ -6,3 +6,14 @@
 //
 
 @import '@carbon/ibmdotcom-styles/scss/components/buttongroup/buttongroup';
+
+#{$prefix}--buttongroup,
+:host(#{$dds-prefix}-button-group) {
+  align-items: flex-start;
+}
+
+.#{$prefix}--buttongroup-item,
+:host(#{$dds-prefix}-button-group-item) {
+  margin-right: 1rem;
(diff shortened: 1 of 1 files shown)


kennylam (approved):
LGTM thanks @IgnacioBecerra!

asudoh:
Hi @IgnacioBecerra thank you for jumping in! Some quick comments:

* With your change, the focus outline seems too thick. To fix that (and to fix the original focus outline issue), `:host(#{$dds-prefix}-button-group)` needs to have `outline: none`. The reason why the problem was happening is that `<dds-button-group>`'s shadow DOM has `delegatesFocus: true` option, which causes focus outline added to the host element as well as to the focusable element in shadow DOM.
* Seems that some of the change is hostile to our existing button group style, like https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.12.0/packages/styles/scss/components/buttongroup/_buttongroup.scss#L51-L53. I think with `outline: none` change, the flex box style change and the margin/padding change won't be needed.

Issue #4327, opened 2020-10-28
Markdown: Consider rendering to light DOM


asudoh:
### The problem

`<dds-markdown>` renders the transformed HTML content to shadow DOM as of today. It deprives application from styling the rendered HTML.

### The solution

Consider rendering the content in light DOM.

Issue #4129, opened 2020-10-05
Web component: Content Block simple - Visual focus is there but not having sufficient contrast


Praveen-sr:
## Detailed description

Issue: Visual focus is there but not having sufficient contrast

For more information please find the below link,
https://www.w3.org/WAI/GL/low-vision-a11y-tf/wiki/Contrast_(Minimum)

Expected Result:
Provide thick and strong visual focus.
1. If, Visual Focus Indicator with 3 CSS px stroke then provide a sufficient contrast that is equal to 3:1. 
2. Visual focus indicator for a link that is a 1 CSS px then provide a sufficient contrast greater than 4.5:1

Note: This defect does not exist in react staging

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/069da1cd-2d19-4d6d-80ea-05107e95d3c8)

annawen1:
@Praveen-sr can you check again? The visual focus looks pretty visible from what I'm seeing. May have been already resolved:

<img width="447" alt="Screen Shot 2020-10-19 at 5 10 49 PM" src="https://user-images.githubusercontent.com/54281166/96512198-209eb680-122e-11eb-8d65-4eca3bd3bd17.png">


https://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-content-block-simple--with-aside-elements

Praveen-sr:
@annawen1 Re checked this on Firefox and chrome browsers, 
1. In chrome it is working fine because that is the browser visual focus (inbuilt browser feature)
2. In Firefox, visual focus is there but not having sufficient contrast for "list item & consectetur adipiscing" link

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/1b958c31-6c71-43ca-b2af-ab2c4cddda92)

asudoh:
This is caused by #4327.

Issue #4104, opened 2020-10-01
Web component: Content Block simple - No spacing between all sections in the page.


Praveen-sr:
## Detailed description

Issue: 
1. No spacing between heading & copy text
2. No spacing between copy text and image
3. No spacing between list items in the page.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/489209cd-2ba8-43c1-a5bc-209d24226a4b)

Praveen-sr:
Note: 
1. No spacing between heading & copy text
2. No spacing between copy text and image

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/f36d2fbb-00e2-47cb-9260-110ccf9e6ed6)

PR #4351, opened 2020-10-30
chore(markdown): render to light DOM for styling

asudoh:
### Related Ticket(s)

Refs #4327.
Refs #4129.
Refs #4104.

### Description

Changes `<dds-markdown>` to:

1. Render its converted HTML to light DOM instead of in shadow DOM, to allow adaptors to apply their custom styles to the rendered markdown content.
2. Apply Carbon standard style automatically via `carbon-web-components`.
3. Removes attributes for markdown rendering options that seems not supported by `markdownToHtml()`.

For the second one above, enhances `markdownToHtml()` to allow custom renderer and custom tags.

From the same/similar purpose of the first two above, introduces `<dds-content-block-copy>` and `<dds-content-item-copy>`, that extends `<dds-markdown>` and replaces `copy` attribute of `<dds-content-block>` and `<dds-content-item>`, respectively. `<dds-content-block-copy>` and `<dds-content-item-copy>` renders new `<dds-content-block-paragraph>` and `<dds-content-item-paragraph>` tags, respectively, to achieve their specific styling needs.

Due to the third one above, `<dds-card-cta-markdown>` has become the same as `<dds-markdown>` and thus it's removed.

Also:

* Adds expressive theme to the Storybook global style, so `carbon-web-components` components in there get the expressive theme.
* Renames `_hasCta` private property to `_hasCTA` in `<dds-content-item>`, to align to other components.
* Makes `<dds-pictogram-item>` inherit `<dds-content-item>` so it can better represent its nature, and use `<dds-content-item-copy>` instead of rendering `<dds-markdown>` and `<dds-content-item-heading>` instead of rendering `<dds-content-item-heading>` by itself.
* A change to ensure `<dds-image-with-caption>` does not render modal unless it's `lightbox` mode.
* Applys a similar change to `<dds-content-group>`.

### Changelog

**New**

- Custom renderer and custom tag support for `markdownToHtml()`.
- `<dds-content-block-copy>`, the markdown copy content of `<dds-content-block>`.
- `<dds-content-item-copy>`, the markdown copy content of `<dds-content-item>`.
- `<dds-content-block-paragraph>`, the paragraph content of `<dds-content-block>`.
- `<dds-content-item-paragraph>`, the paragraph content of `<dds-content-item>`.

**Changed**

- Changes `<dds-markdown>` to:
  1. Render its converted HTML to light DOM instead of in shadow DOM, to allow adaptors to apply their custom styles to the rendered markdown content.
  2. Apply Carbon standard style automatically via `carbon-web-components`.
  3. Removes attributes for markdown rendering options that seems not supported by `markdownToHtml()`.
- Makes `<dds-pictogram-item>` inherit `<dds-content-item>`.
- A change to ensure `<dds-image-with-caption>` does not render modal unless it's `lightbox` mode.

**Removed**

- `<dds-card-cta-markdown>`


jeffchew:
Thank you @asudoh ! Can you also update the READMEs for the components using the new custom elements (like `dds-content-block-simple`), as it is still referencing the old way of setting copy and paragraph.

Also just for my knowledge, wondering if the copy should be a slot instead of directly referencing the sub-component custom elements and setting as attributes for those?

asudoh:
@jeffchew 

> Can you also update the READMEs for the components using the new custom elements (like dds-content-block-simple), as it is still referencing the old way of setting copy and paragraph.

It's done something like this, but please let me know if there are any missing CodeSandbox changes: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/4351/files#diff-ec211775d10ab010fc9500490855ad3a24ea089b67f6e590d7285b96940d2769

> Also just for my knowledge, wondering if the copy should be a slot instead of directly referencing the sub-component custom elements and setting as attributes for those?

If I understand what you said correctly, yes, above CodeSandbox example reflects such change. But please let me know if I'm missing anything. Thanks!

jeffchew:
> > Can you also update the READMEs for the components using the new custom elements (like dds-content-block-simple), as it is still referencing the old way of setting copy and paragraph.
> 
> It's done something like this, but please let me know if there are any missing CodeSandbox changes: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/4351/files#diff-ec211775d10ab010fc9500490855ad3a24ea089b67f6e590d7285b96940d2769

Yes, looks like you updated the codesandbox but not the README file: https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/4351/index.html?path=/docs/components-content-block-simple--default

asudoh:
Got it, thank you for clarifying @jeffchew! Updated.

jeffchew:
> > Also just for my knowledge, wondering if the copy should be a slot instead of directly referencing the sub-component custom elements and setting as attributes for those?
> 
> If I understand what you said correctly, yes, above CodeSandbox example reflects such change. But please let me know if I'm missing anything. Thanks!

I guess my question is why is it like:

```
<dds-content-item-copy content="Lorem ipsum *dolor* sit amet"></dds-content-item-copy>
```

And not:

```
<dds-content-item-copy>
   Lorem ipsum *dolor* sit amet
</dds-content-item-copy>
```

Or even:

```
<dds-content-block-simple>
  <slot slot="heading">Curabitur malesuada varius mi eu posuere</slot>
  <slot slot="copy">Lorem ipsum *dolor* sit amet</slot>
  ...
</dds-content-block-simple>
```

asudoh:
@jeffchew Good question - If we put the copy content as a child text node, the whitespace in the text node directly affects what's rendered as markdown. Given that, `copy` property (instead of attribute) is the primary way of `<dds-markdown>` (and its descendants like `<dds-content-item-copy>`) as its API to set the markdown content. `copy` attribute can be used, too, as long as the spacing is right.

BTW `<slot>` is for shadow DOM to _define_ the slot, instead of for component users to use the slot. We have some wrong usage of `<slot>`s in that regard (we can search for `<slot slot="...">`) that needs to be fixed.

jeffchew:
> @jeffchew Good question - If we put the copy content as a child text node, the whitespace in the text node directly affects what's rendered as markdown. Given that, `copy` property (instead of attribute) is the primary way of `<dds-markdown>` (and its descendants like `<dds-content-item-copy>`) as its API to set the markdown content. `copy` attribute can be used, too, as long as the spacing is right.
> 
> BTW `<slot>` is for shadow DOM to _define_ the slot, instead of for component users to use the slot. We have some wrong usage of `<slot>`s in that regard (we can search for `<slot slot="...">`) that needs to be fixed.

Ok, I'm wondering if we stick with `<dds-content-item-copy>`, if users have paragraphs of content then having that in an attribute might not work, especially if this is happening in something like a CMS if they configure the `content block simple` or `content group simple` as, say, a parsys element in AEM.

packages/web-components/src/components/pictogram-item/pictogram-item.ts
```diff
@@ -7,12 +7,11 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import { customElement, html, LitElement, property } from 'lit-element';
+import { customElement, html } from 'lit-element';
 import ddsSettings from '@carbon/ibmdotcom-utilities/es/utilities/settings/settings.js';
 import settings from 'carbon-components/es/globals/js/settings';
+import DDSContntItem from '../content-item/content-item';
```
jeffchew, line 13: @asudoh a misspelling here for `DDSContntItem`

packages/web-components/src/components/image-with-caption/image-with-caption.ts
```diff
@@ -12,10 +12,11 @@ import settings from 'carbon-components/es/globals/js/settings';
 import ddsSettings from '@carbon/ibmdotcom-utilities/es/utilities/settings/settings.js';
 import on from 'carbon-components/es/globals/js/misc/on';
 import '../image/image';
+import '../modal/modal';
+import '../modal/modal-close-button';
```
jeffchew, line 16: @asudoh should the expressive modal changes go in first?

packages/web-components/src/components/content-item/content-item-copy.ts
```diff
@@ -0,0 +1,45 @@
+/**
```
jeffchew, line 1: I know this is an internal component but wondering if we should start to document the things in here regardless if we're importing into a README or not, it would at least help other developers when maintaining the code.

packages/web-components/src/components/image-with-caption/image-with-caption.ts
asudoh: @jeffchew We can get one that's merged later in sync with another.

asudoh:
> Ok, I'm wondering if we stick with `<dds-content-item-copy>`, if users have paragraphs of content then having that in an attribute might not work, especially if this is happening in something like a CMS if they configure the `content block simple` or `content group simple` as, say, a parsys element in AEM.

@jeffchew Wondering if it's about having to put newline character in attribute, or something else. If it's the case, our adaptors can simply HTML-encode newline characters. If not, interested in hearing more details, e.g. I have't been able to find a website that serves as a good summary of AEM parsys.

packages/web-components/src/components/content-item/content-item-copy.ts
asudoh: @jeffchew Good question - [Web Component Analyzer](https://github.com/runem/web-component-analyzer#readme) automatically harvests the API from ancestor class, and thus the API doc in README reflects ancestor class's API. I think it suffices.

jeffchew:
> > Ok, I'm wondering if we stick with `<dds-content-item-copy>`, if users have paragraphs of content then having that in an attribute might not work, especially if this is happening in something like a CMS if they configure the `content block simple` or `content group simple` as, say, a parsys element in AEM.
> 
> @jeffchew Wondering if it's about having to put newline character in attribute, or something else. If it's the case, our adaptors can simply HTML-encode newline characters. If not, interested in hearing more details, e.g. I have't been able to find a website that serves as a good summary of AEM parsys.

I guess a simpler example would be handlebars, or any CMS template file (AEM uses HTL). The template markup would look something like this:

```
<dds-content-block-simple>
  <dds-content-block-heading>{{ heading }}</dds-content-block-heading>
  <dds-content-item-copy content="{{ copy }}"></dds-content-item-copy>
  <dds-image slot="media" alt="Image alt text" default-src="{{ image.rendition.large }}">
    <dds-image-item media="(min-width: 672px)" srcset="{{ image.rendition.large }}"></dds-image-item>
    <dds-image-item media="(min-width: 400px)" srcset="{{ image.rendition.medium }}"></dds-image-item>
    <dds-image-item media="(min-width: 320px)" srcset="{{ image.rendition.small }}"></dds-image-item>
  </dds-image>
  <dds-text-cta slot="cta" href="{{ urlLink }}" type="local">{{ urlLabel }}</dds-text-cta>
</dds-content-block-simple>
```

In this scenario, the CMS will put in whatever `copy` in the authoring field that would go into `<dds-content-item-copy>`, which may or may not include spaces or line breaks. We're saying it's up to the CMS to escape those? I think it might be better if we did it as `<dds-content-item-copy>{{ copy }}</dds-content-item-copy>` then let our markdowndown utility handle whatever gets put in there.

asudoh:
@jeffchew You are right, we basically ask our adaptors to escape LFs in attribute for the particular case. The reason is that putting the contents in as the child node (text node) will cause redundant whitespace that affects how the markup is rendered. If it's possible with the CMS, a better approach is rendering the markup server-side, because by doing so we can avoid all the problems like whitespace and the extra code in client-side to sanitize the rendered markdown.

packages/web-components/src/components/pictogram-item/pictogram-item.ts
asudoh: Good catch @jeffchew thanks! Fixed.

jeffchew:
> @jeffchew You are right, we basically ask our adaptors to escape LFs in attribute for the particular case. The reason is that putting the contents in as the child node (text node) will cause redundant whitespace that affects how the markup is rendered. If it's possible with the CMS, a better approach is rendering the markup server-side, because by doing so we can avoid all the problems like whitespace and the extra code in client-side to sanitize the rendered markdown.

I still think that having as an attribute isn't a good idea. We should expect adopters to trim whitespace on their own if it's an issue. If anything, if they pass non-trimmed/non-escaped content into an attribute that's even worse. I think having as content of the element makes more sense, and providing in the copy description that they should trim whitespace to not introduce unwanted rendering issues.

asudoh:
Thank you for your thought @jeffchew - Sent a message offline.

asudoh:
Per discussion with @jeffchew, unsupported `content` as an attribute and add support for child text node for providing markdown content, with a clear documentation that it's our adaptor's responsibility to take extra care not to put any extra whitespace in/around the markdown content. Thanks @jeffchew for your feedback!

kennylam (approved):
LGTM thanks @asudoh!

packages/web-components/docs/markdown-contents.md
```diff
@@ -0,0 +1,9 @@
+# Markdown contents
+
+`<dds-content-*-copy>` components, used as the copy text of our content pattern, takes markup as its content. It takes either `content` property (_not_ attribute) or child text node as the markdown content. The markdown content will be converted to HTML with sanitization.
+
+Given the nature of child text node, some extra caution is required to use child text node as the markdown content:
+
+* After-initialization change of markdown content via child text node is not supported.
+* Don't put any extra whitespace, e.g. line feeds, between the start/end tag and the markdown content.
+* HTML-espape the content as needed.
```
jeffchew, line 9:
```suggestion
* HTML-escape the content as needed.
```

PR #4380, opened 2020-11-02
docs(masthead-search): fix wrong event name in the doc

asudoh:
### Changelog

**Changes**

- Fix wrong doc for custom search box implementation, where `dds-masthead-search-input` event name instead of `input` should be used.


kennylam (approved):
Looks good thanks @asudoh!

PR #4386, opened 2020-11-03
style(button-group): fixing alignment for icons on iOS

guilhermelMoraes:
### Description

The explicit declaration of `top` and `bottom` positioning fixed the icon's alignment on iOS devices.

**Previous**
Screen Shot 2020-11-03 at 17 57 46

**Current**
Screen Shot 2020-11-03 at 17 57 01

### Changelog

**New**

- `bottom` and `top` set to `13px` at the button's icons.

### DIFF
--- a/packages/styles/scss/components/buttongroup/_buttongroup.scss
+++ b/packages/styles/scss/components/buttongroup/_buttongroup.scss
@@ -57,6 +57,8 @@
     ::slotted([slot='icon']) {
       position: absolute;
       right: 1rem;
+      top: 13px;
+      bottom: 13px;
       flex-shrink: 0;
     }
     .#{$prefix}--btn {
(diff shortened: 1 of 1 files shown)


packages/styles/scss/components/buttongroup/_buttongroup.scss
```diff
@@ -57,6 +57,8 @@
     ::slotted([slot='icon']) {
       position: absolute;
       right: 1rem;
+      top: 13px;
+      bottom: 13px;
```
asudoh, line 61: We'd want to avoid hard-coding the position, especially considering different size of button. That said, would you be able to try `top: 50%; transform: translateY(-50%)` trick?

asudoh (commented):
Thank you for jumping in @guilhermelMoraes!

asudoh (approved):
LGTM 👍 - Thanks @guilhermelMoraes for the update!

PR #4387, opened 2020-11-03
chore(ddo): removing redundant carbon in version text

jeffchew:
### Related Ticket(s)

No related issue

### Description

After renaming to `Carbon for IBM.com`, our version was renamed but
retained the `Carbon:` prefix as it was originally `Carbon:IBM.com
Library`. Now that we have Carbon in the name, it is a bit redundant to
have the prefix now.

### Changelog

**Changed**

- DDO version


asudoh (approved):
LGTM 👍 - Thanks @jeffchew!

Issue #4394, opened 2020-11-04
Video Player SCSS import is causing compile time error


vishals9711:
## Detailed description

- I am trying to implement Video Player in React written in TSX.

- Video Player scss import is causing compile-time error.

- This error is not present if we don't import the scss of video player.

- Please note all steps from ibmdotcom have been followed to implement the videoPlayer component.

## Additional information
Screenshot of error

![image (16)](https://user-images.githubusercontent.com/26935949/98102030-0f55db00-1eb9-11eb-89eb-c0b33bdd4ef9.png)

PR #4396, opened 2020-11-04
fix(styles): ensure dds-prefix variable is included

asudoh:
### Related Ticket(s)

Refs #4394.

### Description

Ensures `.scss` file using `$dds-prefix` variable or its corresponding `index.scss` imports `@carbon/ibmdotcom-styles/scss/globals/vars` as it defines `$dds-prefix` variable.

`@carbon/ibmdotcom-styles/scss/globals/vars` also subsumes the corresponding file in `carbon-web-components` codebase, as the other portions (expressive theme variables) are no longer needed.

### Changelog

**Changed**

- Ensures `.scss` file using `$dds-prefix` variable or its corresponding `index.scss` imports `@carbon/ibmdotcom-styles/scss/globals/vars`.

**Removed**

- `packages/web-components/src/globals/scss/_vars.scss` as it's merged to `@carbon/ibmdotcom-styles/scss/globals/vars`.


packages/web-components/src/components/content-block/content-block.scss
```diff
@@ -6,7 +6,7 @@
 //
 
 @import 'carbon-components/scss/globals/scss/vars';
-@import '../../globals/scss/vars';
+@import '@carbon/ibmdotcom-styles/scss/globals/vars';
 @import '../../globals/scss/box-sizing';
```
jeffchew, line 10: @asudoh just checking based on our slack conversation, do we need to reference `@carbon/ibmdotcom-styles` for `box-sizing` too?

asudoh: @jeffchew Good question - Not sure the issue happens with non-Web-Components codebase. The issues was that some elements in Web Components got `box-sizing: content-box` (default in UA style) for some of its element. If we see it happens in non-Web-Components codebase in any way, we should make the move.

jeffchew (approved):
Thank you @asudoh !

Issue #3906, opened 2020-09-10
React: Change the IBM logo so that it can be customized


RobertaJHahn:
#### User Story

> As a `[user role below]`:
IBM.com Library developer

> I need to:
change the `IBM Logo`

> so that I can:
provide the ibm.com adopter developers components they can use to build ibm.com web pages

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- **See the Epic for the Design and Functional specs information**
- **Added the `sprint must have` label as this feature is committed for the November release**
- Prod QA testing issue (#3908)

#### Acceptance criteria
- [ ] Built as a pure React component/variant
- [ ] Include README for the react component and corresponding styles
- [ ] Add any necessary stable selectors
- [ ] Create codesandbox example under `/packages/react/examples/codesandbox` and include in README
- [ ] Minimum 80% unit test coverage
- [ ] A comment is posted in the Prod QA issue, tagging Praveen, when development is finished

PR #4401, opened 2020-11-04
feat(Masthead): React masthead alternate logo and tooltip options

kennylam:
### Related Ticket(s)

#3906  

### Description

The masthead now accepts an optional logo options from IBM.com translation files which allows the setting of an alternate IBM logo to use, as well as an optional tooltip when the icon is hovered over. When an alternate logo is used, settings include an expiration date and allow/deny lists for controlling paths on which it can be displayed. The tooltip can be used with both default and alternate logos.

### Changelog

**New**

- `MastheadLogo` service to determine whether an alternate logo can be used based on values passed in the object
- React `Masthead` component now accepts options from IBM.com translations for an alternate logo and tooltip
- internal `ConditionalWrapper` component 

**Changed**

- add masthead logo svg dimension styles


packages/services/src/services/MastheadLogo/MastheadLogo.js
```diff
@@ -0,0 +1,46 @@
...
+   * Determines whether to return custom or default IBM logo
+   *
+   * @returns {boolean} Use alternate logo or not
+   *
+   * @example
+   * import { MastheadLogoAPI } from '@carbon/ibmdotcom-services';
+   *
+   * const useAlternateLogo = MastheadLogoAPI.setMastheadLogo(logoData);
+   */
+  static setMastheadLogo(logoData) {
```
jeffchew, line 19: I'm wondering actually if we should pass in data from an actual data fetch (maybe it lives in IBM.com Translations), so that the app doesn't need to set this manually? Wondering how this can be managed across multiple gateway pages and the Masthead then renders the IBM logo accordingly?

kennylam: @jeffchew I agree. I'm going to remove the prop and the data will optionally come from the translation files.

jeffchew (approved):
Looks good! Do we have a place where this can be demoed?

kennylam:
> Looks good! Do we have a place where this can be demoed?

Not yet, I'll set up a Codesandbox with a mock example.

Issue #4055, opened 2020-09-25
Web component: Feature card medium - arrow icon on the card is placed on wrong direction. 


Praveen-sr:
## Detailed description

On Mobile and Tab view

Issue: The arrow icon is placed on the wrong side of the card. 
No as per specs.

Note: working fine on react 

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/0cb30cbe-3b21-4469-9fce-d3dace887c75)

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/ea2888cf-60d5-455b-8331-e6d02835a2f4)

asudoh:
@Praveen-sr Is this still reproducible? Thanks!

![image](https://user-images.githubusercontent.com/1259051/97952244-093ffb80-1de0-11eb-92be-9c61748c725f.png)

Praveen-sr:
@asudoh Yes, the issue still exists on Mobile and Tab.

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/9b6f53cd-bafe-4297-bd29-fa6eb65bbe3c)

PR #4412, opened 2020-11-05
fix(feature-card): fix icon position

asudoh:
### Related Ticket(s)

Refs #4055.

### Description

Fixes feature card's icon position in Safari browser, by using `<dds-feature-card-footer>` to make sure its style (that uses flex box to adjust the icon position) is in effect.

Also removes the styles for the old markup structure.

### Changelog

**Changed**

- Feature card stories ensuring that `<dds-feature-card-footer>` is used.

**Removed**

- The styles for the old markup structure.


ariellalgilmore (commented):
LGTM! One weird thing is in feature-card-block-medium where the footer is now positioned after the card. I think this could be because of the margin being applied the component.

asudoh:
> One weird thing is in feature-card-block-medium where the footer is now positioned after the card. I think this could be because of the margin being applied the component.

Good catch @ariellalgilmore - Fixed.

ariellalgilmore (approved):
LGTM 👍 ! Thanks @asudoh!

PR #4413, opened 2020-11-05
chore(build): update node version

asudoh:
### Description

Fixes an issue where wrong node.js version is used in CI.

### Changelog

**Changed**

- Fixes an issue where wrong node.js version is used in CI.


kennylam (approved):
LGTM thanks @asudoh!

PR #4415, opened 2020-11-05
fix(Footer): fix micro footer button css

kennylam:
### Description

Currently there is no color theme being used in the `micro` footer variation. This adds the correct theme which fixes background and hover colors.

image

### Changelog

**Changed**

- add `g90` theme support for micro footer


asudoh (approved):
LGTM 👍 - Thanks @kennylam!

Issue #4423, opened 2020-11-05
Storybook: Use Plex in manager


asudoh:
### The problem

The manager UI in our Storybook uses the following font:

* React: Storybook default font, which is _not_ Plex
* Web Components: Uses Plex only in Chrome, Storybook default font for other browsers

### The solution

Use Plex across the board

jeffchew:
As mentioned, this is probably best configured in Storybook theming: https://storybook.js.org/docs/react/configure/theming

Wonder if we should add as part of `storybook-addon-theme` package?

jeffchew:
and another note, we might want to work with design to also see if they want to change things like color theme as well to match IBM colors.

asudoh:
@jeffchew Right, #4424 uses `@storybook/theming`. I have seen an attempt to make Storybook manager UI look more like Carbon, including color scheme, but it seems to have been unsuccessful. We may want to do long-term investigation to see how much Storybook has evolved since then. BTW `storybook-addon-theme` is for the UI setting Carbon color theme, instead of the theme content.

PR #4424, opened 2020-11-05
chore(devenv): use Plex in Storybook manager UI

asudoh:
### Related Ticket(s)

Refs #4423.

### Description

Makes a change to React Storybook environment to use Plex in manager UI. Also fixes an issue in Web Components Storybook where:

* Plex is used only in Chrome
* Plex is not used for code snippet

### Changelog

**Changed**

- A change to React Storybook environment to use Plex in manager UI.
- Fix for an issue in Web Components Storybook where:
  - Plex is used only in Chrome
  - Plex is not used for code snippet


packages/react/.storybook/theme.js
```diff
@@ -0,0 +1,20 @@
...
+
+export default create({
+  base: 'light',
+  brandTitle: 'Carbon for IBM.com React',
+  brandUrl:
+    'https://github.com/carbon-design-system/carbon-for-ibm-dotcom/tree/master/packages/react',
+  fontBase: '"IBM Plex Sans", "Helvetica Neue", Arial, sans-serif',
+  fontCode:
+    '"IBM Plex Mono", Menlo, "DejaVu Sans Mono", "Bitstream Vera Sans Mono", Courier, monospace',
+});
```
jeffchew, line 20: Sorry, just noticed this after I made a comment in the issue! Thank you for opening this so quickly!

As mentioned, I'm wondering if we should have this centralized in the `storybook-addon-theme` package, so the theme can be shared between react and web components? I think changing other things like colors might be something worth looking into, and would be nice to manage that in one place.

asudoh: @jeffchew `storybook-addon-theme` is for the UI to choose Carbon color theme, and it doesn't have (or deal with) theme content at all. Sharing Storybook theme content is an interesting idea, though, we can try it once we get more Storybook theme content.

jeffchew: Thanks, and I think this is probably a future conversation anyway, as discussed we will add this to the design team's ideation board for further research around this.

jeffchew (approved):
Thank you @asudoh !

PR #4426, opened 2020-11-06
chore(dependencies): upgrade carbon-web-components

asudoh:
### Related Ticket(s)

To `v1.8.0`.

### Changelog

**Changed**

- Upgrades `carbon-web-components` to `v1.8.0`.


packages/web-components/tests/snapshots/dds-locale-modal.md
```diff
@@ -64,10 +61,7 @@
 >
   <div class="bx--modal-content">
     <dds-expressive-modal-header>
-      <dds-expressive-modal-close-button
-        role="button"
-        tabindex="0"
-      >
+      <dds-expressive-modal-close-button>
```
jeffchew, line 64: @asudoh I see in the snapshot that the role/tabindex is gone in the snapshot, wondering if this is intentional?

asudoh: Thank you @jeffchew for bringing this up! The answer is yes, it's intentional. Both of those features are moved to `<button>` in the shadow DOM of `<dds-expressive-modal-close-button>`.

jeffchew: thanks, just confirming!

Side note on this, this means that the snapshots made are shallow? Wondering if we should do deeper shapshots?

jeffchew (approved):
LGTM!

packages/web-components/tests/snapshots/dds-locale-modal.md
asudoh: > Wondering if we should do deeper shapshots?

Only single DOM tree is supported by our toolchain, unfortunately. It means that we can generate only a single shadow DOM tree of a component, not nested (composed) tree.

PR #4430, opened 2020-11-09
chore(callout-with-media): simplify class hierarchy

asudoh:
### Description

Makes `<dds-callout-with-media-image>` directly inherit `<dds-image-with-caption>` because the custom element in the middle of class hierarchy is not actually used but as a intermediate class.

Makes a similar change to `<dds-callout-with-media-video>` as well from the same reason.

### Changelog

**Changed**

- Makes `<dds-callout-with-media-image>` directly inherit `<dds-image-with-caption>`
- Makes `<dds-callout-with-media-video>` directly inherit `<dds-video-player-container>`


asudoh:
Per discussion with @ariellalgilmore, ended up with removing intermediate classes as they are not actually used. Thank you @ariellalgilmore for sharing your insights!

ariellalgilmore (approved):
LGTM @asudoh!! Thanks for cleaning this up :D

Issue #4436, opened 2020-11-10
Web Components: Some CodeSandbox does not load


asudoh:
## Detailed description

> Describe in detail the issue you're having.

Running some CodeSandbox example, e.g. https://codesandbox.io/s/github/carbon-design-system/carbon-for-ibm-dotcom/tree/master/packages/web-components/examples/codesandbox/components/cta-section, shows an error like below in CodeSandbox console:

```
[snowpack] Package "@carbon/ibmdotcom-web-components" exists but package.json "exports" does not include entry for "./es/components/button-group/button-group".
```

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

A bug.

> Is this issue related to a specific component?

No.

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Not seeing above error.

> What browser are you working in?

Chrome 86.

> What version of Carbon for IBM.com are you using?

Latest `master`.

PR #4437, opened 2020-11-10
docs(examples): add JS extension

asudoh:
### Related Ticket(s)

Refs #4436.

### Description

Adds `.js` extensions to the `import`s in CodeSandbox example as well as our READMEs, because some module bundlers work with latest Node module resolution algorithm that disallows ESM imports without `.js`.

Also removes JSPM instructions from READMEs because we have seen issues with it (likely in JSPM side) and our adaptors have reported it even.

It's worth re-evaluating JSPM usage in future, though, when such issue will potentially be resolved.

### Changelog

**Changed**

- Adds `.js` extensions to the `import`s in CodeSandbox example as well as our READMEs

**Removed**

- JSPM instructions.

### DIFF
--- a/packages/web-components/examples/codesandbox/components/button-group/src/index.js
+++ b/packages/web-components/examples/codesandbox/components/button-group/src/index.js
@@ -7,5 +7,5 @@
  * LICENSE file in the root directory of this source tree.
  */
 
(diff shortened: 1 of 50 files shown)


jeffchew (approved):
Thank you @asudoh !

Issue #4023, opened 2020-09-22
Do not overlap the platform label


kyjak:
When having no menu in the masthead, but a simple project label only,
we would like the search box not to overlap the project link. 
![image](https://user-images.githubusercontent.com/15310250/93894680-48623380-fcef-11ea-8ab2-5c2079ed6c7c.png)
![image](https://user-images.githubusercontent.com/15310250/93894696-4d26e780-fcef-11ea-87db-43d62d25b617.png)
so basically the search box would stretch only up to the end of the "Documentation"

#### Additional Information
- **We committed to provide this fix in the December release to KC.**

RobertaJHahn:
Jenny will include this issue in the Oct 2, 2020 design meeting.

JennySanchez:
@RobertaJHahn Shared the issue detail with the designers, overall we recognize this is a bug but not a feature request. We suggest to covert this issue for the bug fix.

larahanlon2:
@RobertaJHahn @wonilsuhibm @JennySanchez confirming as per L0 design spec:

<img width="1340" alt="Screen Shot 2020-10-02 at 4 20 28 PM" src="https://user-images.githubusercontent.com/53093550/94966441-40627a80-04cb-11eb-9034-3a513732e372.png">

PR #4442, opened 2020-11-10
fix(masthead): search styles

ariellalgilmore:
### Related Ticket(s)

#4023 

### Description

fix styles so search doesn't overlap platform name

Screen Shot 2020-11-10 at 1 44 07 PM

### Changelog

**New**

- added nowrap to platform name

**Removed**

- removed position absolute from search when active


jeffchew:
Hey @ariellalgilmore , I am checking the deploy previews and doesn't seem to be working in React or Web Components.

ariellalgilmore:
@jeffchew hmm are you sure? I just tried checking the deploy previews again and I am seeing the expected results.

jeffchew:
> @jeffchew hmm are you sure? I just tried checking the deploy previews again and I am seeing the expected results.

Sorry, I tried again and it's working. I might have been looking at a deploy preview for another PR.

packages/styles/scss/components/masthead/_masthead-search.scss
```diff
@@ -13,9 +13,11 @@
   padding-left: $carbon--spacing-05;
   padding-right: $carbon--spacing-05;
   margin-left: $carbon--spacing-05;
+  margin-right: $carbon--spacing-05;
```
kennylam, line 16: @ariellalgilmore Do we want this here? When the search is active and platform link is hovered, it looks weird.

image

@wonilsuhibm @oliviaflory Thoughts?

ariellalgilmore: Screen Shot 2020-11-10 at 7 54 13 PM

This is in the designs. When the menu items are there though it causes an additional 16px of white space between the platform and the menu items too.

oliviaflory: @ariellalgilmore I don't think we want to introduce an extra 16px between the platform name and the menu items. In the specs we still have 16px spacing on either side of the platform name and menu items.
Screen Shot 2020-11-11 at 9 29 45 AM



However, I do see why the spec might have been confusing with the search. It looks like the original spec was trying to have the padding be even on either side of the platform, but that would mean stopping the Search at the first letter of the menu item, see below
Screen Shot 2020-11-11 at 9 30 20 AM

I'm guessing it would be easier to have the Search extend to the right of the first menu item like this so we don't have to add another 16px.
Screen Shot 2020-11-11 at 9 37 23 AM

ariellalgilmore: Thanks @oliviaflory for explaining!! Just updated without the margin-right. Should look like this now
Screen Shot 2020-11-11 at 6 45 37 AM

oliviaflory: Thanks @ariellalgilmore ! I will update the spec file to be more clear.

kennylam (approved):
Looks good thanks @ariellalgilmore!

packages/web-components/src/components/masthead/top-nav.ts
```diff
@@ -49,7 +51,13 @@ class DDSTopNav extends StableSelectorMixin(HostListenerMixin(BXHeaderNav)) {
   hideDivider = false;
 
   render() {
-    return this._hideNav ? html`` : super.render();
+    return this._hideNav
+      ? html``
+      : html`
+          <nav class="${prefix}--header__nav">
```
asudoh, line 57: `<dds-top-nav>` has `.bx--header__nav` style:
https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.12.0/packages/web-components/src/components/masthead/masthead.scss#L128
https://github.com/carbon-design-system/carbon-web-components/blob/v1.7.0/src/components/ui-shell/header.scss#L17

So adding `<nav class="${prefix}--header__nav">` will duplicate the element. Probably we just missed some specific style in `.bx--header__nav`? (e.g. one for specific state)

asudoh (commented):
Thank you for jumping in @ariellalgilmore!

packages/web-components/src/components/masthead/top-nav.ts
kennylam: Hm, no I don't think so. That class is missing from the markup.

image

The `ul` should have a parent `.bx--header__nav`.

asudoh: > The `ul` should have a parent `.bx--header__nav`. It looks like `<dds-header-nav>` is missing.

@kennylam Thank you for elaborating the thoughts! What I meant is that `<dds-top-nav>` itself should work as `.bx--header__nav`. Here's how the code intends to do that; [`<dds-top-nav>` inherits Carbon core `<bx-header-nav>`](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.12.0/packages/web-components/src/components/masthead/top-nav.ts#L26). [`<bx-header-nav>` has the style of `.bx--header-nav`](https://github.com/carbon-design-system/carbon-web-components/blob/v1.7.0/src/components/ui-shell/header.scss#L16-L17).

Therefore, I think `<dds-top-nav>` tag selector needs more styles defined.

asudoh: Update: From the conversation offline with @ariellalgilmore (thanks Ari!) I now see what’s going on; Seems that we want to prevent the entire `<dds-top-nav>` from being rendered when search box is activated. However, given `render()` only dictates the shadow DOM the host element remains, that causes an extra padding.

@ariellalgilmore You can use `display: contents` style similar to https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/e8af89a06fb41a649529fa441ff29916f4821655/packages/web-components/src/components/carousel/carousel.scss#L16-L19 along with a comment like: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/e8af89a06fb41a649529fa441ff29916f4821655/packages/web-components/src/components/carousel/carousel.ts#L206

ariellalgilmore: Thanks @asudoh! After trying this and discussing further with @kennylam we realized that even adding `display: contents` to `<dds-top-nav>` will not give the correct results. In this case the padding needs to be within `dds-top-nav`, so the search input can go 100% and cover up its own padding. `dds-top-nav` has the padding-left and the red shows that it’s preventing search input to open fully. If the padding was within this element the search would be able to open fully.

I will also change the `<nav>` to a `<div>` as it is not necessary for it to be `<nav>` in this case. 

image

asudoh: Thanks @ariellalgilmore for the update! One question; Does React version have such additional elements in addition to `<nav>`?

asudoh: Sent a DM to @ariellalgilmore for further analysis.

asudoh (approved):
LGTM 👍 - Thanks @ariellalgilmore for all the updates!

PR #4453, opened 2020-11-10
feat(leaving-ibm): Leaving IBM modal web component

annawen1:
### Related Ticket(s)

Web Components: Develop Leaving IBM component #3327

### Description

Screen Shot 2020-11-11 at 6 55 55 PM

### Changelog

**New**

- leaving-ibm modal web component


packages/web-components/src/components/leaving-ibm/leaving-ibm-composite.ts
```diff
@@ -0,0 +1,127 @@
...
+      <dds-leaving-ibm-modal open="${open}" size="sm">
+        <bx-modal-header>
+          <bx-modal-close-button></bx-modal-close-button>
+          <dds-leaving-ibm-modal-heading>${leavingIbmCopy?.LEAVING001}</dds-leaving-ibm-modal-heading>
+        </bx-modal-header>
+        <bx-modal-body>
+          <p>${leavingIbmCopy?.LEAVING002}</p>
+          <p class="${prefix}--leaving-ibm__text">${leavingIbmCopy?.LEAVING003}</p>
+          <a class="${prefix}--leaving-ibm__link" href="https://www.ibm.com">${_getHostname(url)}</a>
+        </bx-modal-body>
```
asudoh, line 116: Great if we can create an inheritance of `<bx-modal-body>` as a custom element that does something like below:

```javascript
@customElement(`${ddsPrefix}-leaving-ibm-modal-body`)
class DDSLeavingIBMModalBody extends LitElement {
  ...

  render() {
    const { href } = this;
    return html`
      <p><slot></slot></p>
      <p class="${prefix}--leaving-ibm__text"><slot name="supplemental"></slot></p>
      <a class="${prefix}--leaving-ibm__link" href="${ifNonNull(href)}"><slot name="link-description">${new URL(href).hostname}</slot></a>
    `;
  }

  ...
}
```

In this way:

* We can render the content of `<dds-leaving-ibm-composite>` to light DOM (instead of shadow DOM) by relegating the CSS definition of `.bx--leaving-ibm__text`, etc. to `<dds-leaving-ibm-modal-copy>`. We render content of `<dds-*-composite>` to light DOM so the rendered markup structure represents the semantics of the content. Also, for rendering modals, we ensure the content is rendered to the end of rendering tree using [`ModalRenderMixin`](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.12.0/packages/web-components/src/globals/mixins/modal-render.ts) so the modal content won't be hidden by the surrounding elements. Here's an example: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.12.0/packages/web-components/src/components/lightbox-media-viewer/lightbox-video-player-composite.ts#L96-L114
* `<dds-leaving-ibm-composite>` does not have to deal with the HTML structure of the modal body content.

asudoh (commented):
Nice addition @annawen1!

packages/web-components/src/components/leaving-ibm/leaving-ibm-composite.ts
```diff
@@ -101,21 +89,20 @@ class DDSLeavingIbmComposite extends LitElement {
...
+        <dds-leaving-ibm-modal-body href="${href}">
           <p>${leavingIbmCopy?.LEAVING002}</p>
-          <p class="${prefix}--leaving-ibm__text">${leavingIbmCopy?.LEAVING003}</p>
-          <a class="${prefix}--leaving-ibm__link" href="https://www.ibm.com">${_getHostname(url)}</a>
-        </bx-modal-body>
+          <span slot="supplemental">${leavingIbmCopy?.LEAVING003}</span>
+        </dds-leaving-ibm-modal-body>
         <bx-modal-footer>
-          <bx-btn kind="primary">${leavingIbmButtonLabel}</bx-btn>
+          <bx-btn part="cta" href="${href}" kind="primary">${leavingIbmButtonLabel}</bx-btn>
```
asudoh, line 105: Given we are rendering the content to light DOM, we can use `data-autoid` here.

packages/web-components/src/components/leaving-ibm/leaving-ibm-modal-heading.ts
```diff
@@ -15,9 +15,9 @@ import styles from './leaving-ibm.scss';
 const { stablePrefix: ddsPrefix } = ddsSettings;
 
 /**
- * Leaving Ibm modal body.
+ * Leaving Ibm modal body heading.
  *
- * @element dds-leaving-ibm-modal-body
+ * @element dds-leaving-ibm-modal-body-heading
```
asudoh, line 20:
```suggestion
 * @element dds-leaving-ibm-modal-heading
```

packages/web-components/src/components/leaving-ibm/leaving-ibm-modal-heading.ts
```diff
@@ -15,9 +15,9 @@ import styles from './leaving-ibm.scss';
 const { stablePrefix: ddsPrefix } = ddsSettings;
 
 /**
- * Leaving Ibm modal body.
+ * Leaving Ibm modal body heading.
```
asudoh, line 18:
```suggestion
 * Leaving Ibm modal heading.
```

asudoh (commented):
Great update @annawen1!

asudoh (approved):
LGTM 👍, great work 🎉 - Thanks @annawen1!

jeffchew:
Interesting thing I noticed, switching to the Docs tabs still has the modal up:

Screen Shot 2020-11-12 at 11 07 49 AM

annawen1:
> Interesting thing I noticed, switching to the Docs tabs still has the modal up:
> 
> Screen Shot 2020-11-12 at 11 07 49 AM

hmm I notice that with lightbox-media-viewer as well. I'm wondering if it has to do with the [ModalRenderMixin](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.12.0/packages/web-components/src/globals/mixins/modal-render.ts)

annawen1:
@asudoh I removed the `ModalRenderMixin` as the modal would still be visible when we switched to the documentation tab. I also noticed the default lightbox-media-viewer doesn't apply it either. Is this something that you recommend is necessary?

jeffchew:
> @asudoh I removed the `ModalRenderMixin` as the modal would still be visible when we switched to the documentation tab. I also noticed the default lightbox-media-viewer doesn't apply it either. Is this something that you recommend is necessary?

I'll doublecheck the change, but I think it should be fine to move forward with this as the current lightbox media viewer doesn't use this class/mixin.

jeffchew:
@annawen1 can you check the unit tests that are failing?

asudoh:
> @asudoh I removed the ModalRenderMixin as the modal would still be visible when we switched to the documentation tab. I also noticed the default lightbox-media-viewer doesn't apply it either. Is this something that you recommend is necessary?

@annawen1 Yes. This is for coping with something like:

```html
<dds-*modal-container></dds-*modal-container>
<div style="position:fixed;z-index:1000000;width:100vh;height:100vh;background-color:green"></div>
```

What problem did you hit by `ModalRenderMixin`?

PR #4461, opened 2020-11-11
fix(cta-section): aligning link list with other text elements

IgnacioBecerra:
### Related Ticket(s)

#3366

### Description

The horizontal rules from LinkList now overhang in order to make the links match with the text above.

### Changelog

**Removed**

- unnecessary left padding that made horizontal rules align with text instead of overhanging


kennylam (approved):
LGTM thanks @IgnacioBecerra!

PR #4480, opened 2020-11-12
chore(tests): moved integration ui ci-checks to its own workflow

jeffchew:
### Related Ticket(s)

No related issue

### Description

The integration Ui tests are still flakey and are causing many false
positives, so this PR moves them to its own workflow, where the triggers
 are only targeted for a specific branch pattern `test/*`. This way,
 regular ci-checks can continue and report out regularly while we
 sequester integration ui testing until it is more stable.

### Changelog

**Changed**

- moved out `integration-ui` ci checks


asudoh (approved):
LGTM 👍 - Thanks @jeffchew!

Issue #4182, opened 2020-10-08
Masthead overflow – keyboard should skip overflow buttons


oliviaflory:
## Detailed description

> Describe in detail the issue you're having.

- **Included the `sprint must have` label as this overflow feature is committed for the Nov release** 

When navigating through the masthead with the keyboard, the overflow buttons are not being skipped and are getting a focus state. 
The functional specs state the overflow buttons should skip focus on keyboard:

<img width="739" alt="Screen Shot 2020-10-08 at 12 14 08 PM" src="https://user-images.githubusercontent.com/45692622/95485602-fc0a2b00-095f-11eb-90ee-e787b8d24236.png">

![overflow button should skip focus Oct-08-2020 12-18-06](https://user-images.githubusercontent.com/45692622/95485957-68852a00-0960-11eb-85f8-08bb3056e4ce.gif)

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

bug

> Is this issue related to a specific component?

masthead

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Should match functional spec

> What browser are you working in?

Chrome
Version 85.0.4183.121 (Official Build) (64-bit)

> What version of the IBM.com Library are you using?

Canary

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

October release

## Steps to reproduce the issue

1. Open in Chrome
2. Use keyboard to navigate through masthead
3. Step three
4. etc.

## Additional information

- Screenshots or code
- Notes

jeffchew:
@oliviaflory I think I remember that Ram and @Praveen-sr opened a different approach where we would provide a different aria-label that tells what it is, but not hide it. We might need functional spec updates on this @photodow ?

wonilsuhibm:
This isn't just about the screen reader though—the mismatch between what's on the screen and the pagination and functionality will be confusing to users and this does feel like an error.

photodow:
@jeffchew @wonilsuhibm @oliviaflory do we need to set up a short meeting to discuss this? I feel like we're talking about 3 different things, or at least coming at it from 3 different angles and it might be nice to align before we make any changes.

oliviaflory:
That sounds like a good idea @photodow

RobertaJHahn:
Roberta will schedule a meeting next week.

RobertaJHahn:
[...]

PR #4489, opened 2020-11-13
fix(masthead): add aria-hidden to overflow nav carets

ariellalgilmore:
### Related Ticket(s)

#4182 

### Description

add aria-hidden to caret buttons for screen readers using up and down arrows

### Changelog

**New**

- add aria-hidden="true" to carets


annawen1 (approved):
LGTM thanks Ari!

Issue #4061, opened 2020-09-25
Web component: Link List -  In  iPhone VO there is no role for the mentioned links


Praveen-sr:
## Detailed description

On Voiceover iPhone 

Issue: Links found in Horizontal, Vertical, End Of Section links found at the first card in (Vertical With Cards): - in iPhone VO there is no role for the mentioned links
Note: in react staging screen reader is identifying the role link.

Expected - Screen reader should identify the role for the mentioned links.

RaphaelAmadeu:
@Praveen-sr 
Could you please give us a little more info about this? 
Doesn't VO mention the listitem role?

Praveen-sr:
@RaphaelAmadeu Here the mentioned links are announced only as text. The "link" role is not announced. 
Hence the user would not know that these are interactive elements.
Expected behavior: VO should announce these as links instead of only announcing as "text".

asudoh:
@Praveen-sr Could your team re-test this issue at: http://ibmdotcom-web-components-canary.mybluemix.net/iframe.html?id=components-link-list--vertical

Out investigation shows that the latest code has the right a11y tree and works well with Safari VoiceOver. The VoiceOver one in below screenshot is of desktop, but I see mobile Safari has the same result. Thanks!

![image](https://user-images.githubusercontent.com/1259051/99326458-b6744280-28bb-11eb-82b0-7490e578317a.png)
![link-role](https://user-images.githubusercontent.com/1259051/99326439-ae1c0780-28bb-11eb-95a2-91b9e774cdf7.gif)

PR #4498, opened 2020-11-16
fix(link-list): change role for link-list items to link

annawen1:
### Related Ticket(s)

Web component: Link List - In iPhone VO there is no role for the mentioned links #4061

### Description

Change the role to link from link-list item and card for screen readers. Previously it was set to `listitem` which does not let screen reader users know that it's actually a clickable link.

### Changelog

**Changed**

- change `role` for `link-list-item` and `link-list-card` to `link`


ariellalgilmore (approved):
LGTM thanks @annawen1!

Issue #4440, opened 2020-11-10
Masthead underline shape


wonilsuhibm:
## Detailed description

The blue underline that shows the currently active menu item has angled ends (see the attached image).

The ends should be flat (see the attached image)

## Steps to reproduce the issue

1. Open this [StoryBook page](https://ibmdotcom-react.mybluemix.net/?path=/story/components-masthead--default)
2. Check both ends of the underline.

## Additional information
- A screenshot has been added as a comment below.

wonilsuhibm:
<img width="1471" alt="Screen Shot 2020-11-10 at 8 23 19 AM" src="https://user-images.githubusercontent.com/30183447/98679720-6c74e500-232e-11eb-9f07-6f50c3d87f83.png">

oliviaflory:
Not sure if this is helpful, but it looks like the Carbon UI shell header is getting around the angled underline with 2px of space on either side of the blue highlight/selected state, example screen shot below. 

Note: it looks like there is a 1px border or something below the focus and highlight.

Link to [Carbon React storybook](https://react.carbondesignsystem.com/?path=/story/ui-shell--header-base-w-navigation)

<img width="760" alt="Screen Shot 2020-11-10 at 8 38 44 AM" src="https://user-images.githubusercontent.com/45692622/98686574-b6fa5f80-2336-11eb-89e3-4e5f533cf2e9.png">

<img width="708" alt="Screen Shot 2020-11-10 at 9 26 42 AM" src="https://user-images.githubusercontent.com/45692622/98686719-e315e080-2336-11eb-8855-fb8774285e00.png">

PR #4502, opened 2020-11-17
fix(Masthead): use aria-selected selector and outline styling

annawen1:
### Related Ticket(s)

Masthead underline shape #4440

### Description
The selected style (border-bottom) had angled edges due to the default transparent border around the menu item. Removing the default border styles and using `outline` when menu item is active or focused allows the selected border-bottom style to have the correct look.

Screen Shot 2020-11-17 at 7 51 37 AM

### Changelog

**Changed**

- use `[aria-selected]` attribute as selector instead to identify when to set border-bottom styling
- remove default border styling, use `outline` for `active` and `focus` states

**Removed**

- unnecessary `masthead__l0-nav--selected-item` classname


asudoh (approved):
LGTM 👍 - Thanks @annawen1!

Issue #4474, opened 2020-11-12
Web component: CTA section - with content items - mobile / tab - copy text and link are overlapping, don't have space btw them


Praveen-sr:
## Detailed description

Issue: CTA section - with content items - mobile / tab - copy text and link are overlapping, don't have space btw them

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/6fd8e710-01b0-4fee-92d8-e6f67731a475)

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/7331929d-3e85-4eec-89a3-3c359162b91f)

Issue #4472, opened 2020-11-12
Web component: CTA section - On mobile, button are overlapping no space between them.


Praveen-sr:
## Detailed description

On iPhone and Android phones.

Issue: 
1. CTA section - On mobile, button are overlapping no space between them.
2. On desktop - Size (width ) of secondary button is more compared to primary button.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/e8002662-eb50-4865-8c9d-57f23075edc6)

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/8027cd0d-783a-43be-afa3-7a4684bdb66d)

PR #4511, opened 2020-11-17
fix(cta-section): uneven copy text height

IgnacioBecerra:
### Related Ticket(s)

#4472 
#4474 

### Description

Introduced `ResizeObserver` to the `cta-section-item-copy` elements in order to keep their height consistent between the siblings and the `cta-text` elements.

### Changelog

**New**

- added `ResizeObserver` to children elements within `cta-section`


packages/web-components/src/components/cta-section/cta-section.ts
```diff
@@ -62,6 +64,37 @@ class DDSCTASection extends StableSelectorMixin(DDSContentItem) {
...
+      if (window.innerWidth >= parseFloat(breakpoints.md.width) * baseFontSize) {
+        entries.forEach(entry => {
+          entry.target.style.height = `${Math.max(
+            ...newArray.map(o => {
+              return o.clientHeight;
+            })
+          )}px`;
+        });
+      }
+    });
```
asudoh, line 91: The `ResizeObserver` has to be lifecycle-managed (should be gone when the element goes out of the render tree). Please see https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/176b82ad97cc9cbb7302ce4f8ebc281eef6fca93/packages/web-components/src/components/carousel/carousel.ts#L200-L212 for more details.

asudoh (commented):
Thank you for trying to address this problem space @IgnacioBecerra!

packages/web-components/src/components/cta-section/cta-section.ts
IgnacioBecerra: Thanks for the example! I think I managed to implement the lifecycle management correctly, let me know how it looks!

asudoh (commented):

packages/web-components/src/components/cta-section/cta-section.ts
```diff
@@ -35,20 +37,44 @@ const slotExistencePropertyNames = {
  */
 @customElement(`${ddsPrefix}-cta-section`)
 class DDSCTASection extends StableSelectorMixin(DDSContentItem) {
+  /**
+   * Content Items nodes.
+   */
+  @query(`slot[name='content-item']`)
```
asudoh, line 43:
```suggestion
  @query(`.${prefix}--content-item-wrapper`)
```

packages/web-components/src/components/cta-section/cta-section.ts
```diff
@@ -35,20 +37,44 @@ const slotExistencePropertyNames = {
...
+    if (contentsNode?.assignedNodes().length) {
+      if (this._observerResizeContainer) {
+        this._observerResizeContainer.disconnect();
+        this._observerResizeContainer = null;
+      }
+
+      if (create) {
+        setTimeout(() => {
+          this._observeResizeContainer();
+        });
```
asudoh, line 75: `this._observeResizeContainer()` shouldn't be called directly, because `ResizeObserver` should determine the most optimal timing to call `this._observeResizeContainer()`. If you don't see `this._observeResizeContainer()` invoked from `ResizeObserver` in below way please let me know.

```suggestion
        // TODO: Wait for `.d.ts` update to support `ResizeObserver`
        // @ts-ignore
        this._observerResizeContainer = new ResizeObserver(this._observeResizeContainer);
        this._observerResizeContainer.observe(this._contentsNode);
```

IgnacioBecerra: That's good to know, in the latest commit I added this in, as well as refactored the actual height-changing code.

packages/web-components/src/components/cta-section/cta-section.ts
```diff
@@ -64,6 +91,55 @@ class DDSCTASection extends StableSelectorMixin(DDSContentItem) {
     this[slotExistencePropertyNames[name] || '_hasFooter'] = hasContent;
   }
 
+  /**
+   * The observer for the resize of the content item slot
+   */
+  private _observeResizeContainer = () => {
+    const { _contentsNode: contentsNode } = this;
+
+    const childItems = contentsNode?.children[0].assignedNodes();
```
asudoh, line 100: Probably we should define another private property for `<slot>`.

packages/web-components/src/components/cta-section/cta-section.ts
```diff
@@ -64,6 +91,55 @@ class DDSCTASection extends StableSelectorMixin(DDSContentItem) {
...
+    });
+
+    if (window.innerWidth >= parseFloat(breakpoints.md.width) * baseFontSize) {
+      newArray.forEach(entry => {
+        entry.style.height = `${Math.max(
+          ...newArray.map(o => {
+            return o.clientHeight;
+          })
+        )}px`;
+      });
```
asudoh, line 119: Seems that the `clientHeight` calculation makes it <code>O(n<sup>2</sup>)</code> - Probably we should calculate the max height before setting `entry.style.height`?

asudoh (commented):

packages/web-components/src/components/cta-section/cta-section.ts
```diff
@@ -35,20 +37,51 @@ const slotExistencePropertyNames = {
...
+  /**
+   * Cleans-up and creats the resize observer for cta-section
+   *
+   * @param [options] The options.
+   * @param [options.create] `true` to create the new resize observer.
+   */
+  private _cleanAndCreateObserverResize({ create }: { create?: boolean } = {}) {
+    const { _contentsNode: contentsNode } = this;
+
+    if (contentsNode?.children[0]?.assignedNodes().length) {
```
asudoh, line 72:
```suggestion
    const { _slotNode: slotNode } = this;

    if (slotNode?.assignedNodes().length) {
```

Also, given we are checking if the `<slot>` has any assigned nodes, seems that we need to run this method upon `slotchange` event.

IgnacioBecerra: I see, seems that we won't need `_contentsNode` anymore, but could you explain why we changed the query to [this](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/4511#discussion_r526504287) earlier? Given that the wrapper is a parent element of the actual slot element we're focusing on, it seemed slightly redundant to target it over the child slot itself no?

asudoh: @IgnacioBecerra Good question, it was for the container `ResizeObserver` observes for its size. Seems that you changed it to `this`. So my question is; Does `contentsNode` resize?

IgnacioBecerra: Hmm, `contentsNode` resizes when the window also does, which I believe also affects the child elements in this case. 

I think I understand -- we were only watching `contentsNode` for resize before, but now that `ResizeObserver` observes `this` (the greater component itself), once the window resize, all the elements within also resize, so there's no need for `contentsNode` now that it's not being observed directly., I see.

packages/web-components/src/components/cta-section/cta-section.ts
```diff
@@ -108,7 +104,7 @@ class DDSCTASection extends StableSelectorMixin(DDSContentItem) {
 
     childItems?.forEach((elem, index) => {
       /* eslint-disable-next-line prefer-destructuring */
-      newArray[index] = elem.children[1];
+      newArray[index] = (elem as any).children[1];
```
asudoh, line 107: Wondering what `.children[1]` is meant for...? BTW TypeScript did a good catch here because `.assignedNodes()` returns any types of nodes but text node, etc. does not have `.children`.

packages/web-components/src/components/cta-section/cta-section.ts
asudoh: Thank you for explaining @IgnacioBecerra! If `contentsNode` resizes, we can observe the size of `contentsNode` directly. e.g. What if somebody creates a draggable splitter UI and put our CTA section in either side of the split pane? (Where CTA section resizes without window resizing)

packages/web-components/src/components/cta-section/cta-section.ts
IgnacioBecerra: `.children[1]` returns a `dds-cta-section-item-copy`, which is where we get the `clientHeight` values that we need.

asudoh: Thanks @IgnacioBecerra for explaining! In such case, we should define `selectorCopy` static property and do something like below so future change in DOM structure won't break the code:

```javascript
const { selectorCopy } = this.constructor as typeof DDSCTASection;
const copyNodes = slotNode?.assignedNodes().reduce((acc, node) => {
  if (node.nodeType === Node.ELEMENT_NODE) {
    acc.push(...(node as Element).querySelectorAll(selectorCopy));
  }
  return acc;
}, []);
```

Search for `selectorSomething` static properties in other components to see the details.

packages/web-components/src/components/cta-section/cta-section.ts
IgnacioBecerra: Ohh, I see, I get it now. I did not write the rest of your earlier suggestion of using `observe(this._contentsNode)` and left it at `observe(this)` instead. Even though using `this` still worked, having that realization + your splitter UI example made everything click. Thanks @asudoh!

packages/web-components/src/components/cta-section/cta-section.ts
IgnacioBecerra: That is much cleaner, I like it, thanks @asudoh!

IgnacioBecerra:
Closing due to having found a muuuuuch easier approach (#4716).

Issue #4266, opened 2020-10-20
Web component: Image with caption - issues related to modal


Praveen-sr:
## Detailed description
On Voiceover iPhone
Issue: 
1. after activating the image alt text button, for the screen reader users the focus  remains on the triggering control itself
2. After activating close button,  the dialog got closed visually and the focus was taken to the parent page.
But screen reader users was not able to navigate using left and right swipe screen reader was only announcing as "close button"

Expected: 
1. Set the focus to the close button.
2. after activation screen reader users should be able to use left and right swipe.

PR #4512, opened 2020-11-18
fix(expressive-modal): set list of tabbable elements

asudoh:
### Related Ticket(s)

Refs #4266.

### Description

Updates `selectorTabbable` static property of `<dds-expressive-modal>` so it sets the focus to the close button when the modal gets open and no other focusable element is there in the modal.

Also changes `<dds-image-with-caption>` so focusing on `<dds-image-with-caption>` is delegated to the `<button>` in it.

### Changelog

**Changed**

- Updates `selectorTabbable` static property of `<dds-expressive-modal>`, adding `dds-modal-close-buttton`, etc. selector.
- Changes `<dds-image-with-caption>` so focusing on `<dds-image-with-caption>` is delegated to the `<button>` in it.


annawen1 (approved):
LGTM!

Issue #4199, opened 2020-10-09
React: Masthead - screen reader is reading  controls which are not available visually.


Praveen-sr:
## Detailed description
 
On Voiceover iPhone
Issue: 
1. visually only Search  edit combo , search icon and  close icon is present.
 but for screen reader users screen reader is reading controls which are not available visually.
for example hamburger menu, IBM logo link,  user profile button etc.
2. after entering the search term visually the options are appearing but   screen reader users are not able to navigate through the options within the combo box

Expected:
1. screen reader should not read visually hidden components
2. the user should be able to navigate through the options  using left and right swipe and make the selection.

PR #4529, opened 2020-11-19
fix(Masthead): fix screen reader when search active

kennylam:
### Related Ticket(s)

#4199 

### Description

This fixes an issue where elements in the masthead were being announced by screen readers when search is active, despite being under the search and not visible.

### Changelog

**Changed**

- add CSS to hide IBM logo and profile menu in mobile views when search is active to prevent screen readers announcing them
- do not render hamburger when search is active


ariellalgilmore (approved):
LGTM! thanks @kennylam!

Issue #4526, opened 2020-11-19
React: Masthead - On iPhone Voice over, sub menu options are not identified


Praveen-sr:
## Detailed description

Issue: 
1. Voiceover iPhone Safari identifies the mentioned menus as tabs instead of menus. 
2. When we activate these menus visually the sub-menus are displayed but screen reader focus goes only to the parent menu not the 
sub-menus within it.

Expected:
Either these menus should be made in adherence to ARIA standards. Or an accessible alternate should be given. 
Below is the reference link for aria 1.1 examples. Note: We recommend the second approach. 
1. Navigation Menu bar Example | WAI-ARIA Authoring Practices 1.1
https://www.w3.org/TR/wai-aria-practices/examples/menubar/menubar-1/menubar-1.html
2. Example Disclosure for Navigation Menus | WAI-ARIA Authoring Practices 1.1
https://www.w3.org/TR/wai-aria-practices/examples/disclosure/disclosure-navigation.html

PR #4530, opened 2020-11-19
fix(Masthead): change side nav menu item role to menuitem

annawen1:
### Related Ticket(s)

React: Masthead - On iPhone Voice over, sub menu options are not identified #4526

### Description

Change the `role` from tab to `menuitem` and also change `aria-selected` from `false` to empty string so the screenreader doesn't read "not selected" for every menu item.

### Changelog

**Changed**

- change role to menu item
- return empty string for `aria-selected`


kennylam (approved):
LGTM!

PR #4536, opened 2020-11-20
feat(content-block-segmented): create web component

ariellalgilmore:
### Related Ticket(s)

#3790 

### Description

create content-block-segmented. 

*still need to match storybook knobs in react

### Changelog

**New**

- create component, stories, and tests


packages/web-components/src/components/content-block-segmented/__tests__/content-block-segmented.test.ts
```diff
@@ -0,0 +1,208 @@
...
+              >
+              ${image}
+              <dds-text-cta slot="footer" cta-type="local" icon-placement="right" href="https://example.com"
+                >Lorem Ipsum dolor sit</dds-text-cta
+              >
+            </dds-content-block-segmented-item>
+            <dds-card-cta slot="cta" cta-type="local" href="https://example.com">
+              Lorem ipsum dolor
+              <dds-card-cta-footer></dds-card-cta-footer>
+            </dds-card-cta>
```
asudoh, line 86: Wondering if this test case and similar test cases can be simpler. Note that the snapshot test in Web Components codebase generates snapshots of only Shadow DOM, not light DOM. That said, changing `children` for snapshot testing is worthwhile only if `children` should change Shadow DOM content, and thus `children` should be minimum that ensures triggering change in Shadow DOM content.

asudoh (commented):
LGTM basically, just one thing - Thanks @ariellalgilmore!

packages/web-components/src/components/content-block-segmented/__tests__/content-block-segmented.test.ts
ariellalgilmore: Thanks @asudoh! I was wondering the same thing. It definitely makes sense to only render Shadow DOM. I will update :)

asudoh (approved):
LGTM 👍 - Thanks @ariellalgilmore for the update!

PR #4538, opened 2020-11-20
fix(SideNavMenu): remove role from li wrapper on menu item buttons

annawen1:
### Related Ticket(s)

React: Masthead - On iPhone Voice over, sub menu options are not identified #4526

### Description

On iphone VO, with swipe, the focus goes to the `<li>` element instead of the button where the label to be read out is. Removing the `role` attribute from the `<li>` and moving the `aria-selected` attribute to the button instead allows for the focus to be on the button on swipe.

### Changelog

**Changed**

- set `aria-selected` attribute to the menu item button

**Removed**

- `role` attribute from wrapper `<li>` element


asudoh (approved):
LGTM 👍 - Thanks @annawen1!

kennylam (approved):
LGTM!

Issue #3270, opened 2020-07-24
[Cards] Revisit card aspect ratio


shixiedesign:
During Product landing page's implementation phase, we noticed the cards in DDS library do not observe any **aspect ratio**. Original design intent is to always follow one of the aspect ratios https://www.ibm.com/design/language/2x-grid/#aspect-relationships. 

**Current:** 
![image](https://user-images.githubusercontent.com/15144993/88405458-4df9d680-cd9d-11ea-9b37-9e4453a31980.png)

**Expected (wireframe):** 
![image](https://user-images.githubusercontent.com/15144993/88405611-84cfec80-cd9d-11ea-99ee-9b03589bf23d.png)

Slack discussion on this issue: 
FYI @photodow 

![image](https://user-images.githubusercontent.com/15144993/88404396-d4adb400-cd9b-11ea-81dd-3465597e4b27.png)

![image](https://user-images.githubusercontent.com/15144993/88404162-800a3900-cd9b-11ea-90c7-c8817e1fb573.png)

RobertaJHahn:
Next step:  This issue will be used for the Engineering exploration.  If a solution is available, then will be turned into an Epic and related issues will be created.  Adding to the Engineering meeting agenda.

PR #4542, opened 2020-11-20
Fix/card aspect ratio

kennylam:
### Related Ticket(s)

#3270 

### Description

This adds a default `2:1` aspect ratio to `Card` and all components using it. The `fixed` option is set to false, which allows the card to grow with the content. This can be changed on a component-level when the aspect ratio should be preserved regardless of content height.

### Changelog

**Changed**

- add `2:1` aspect ratio to `Card` and components using it as a dependency

**Removed**

- {{removed thing}}


asudoh (approved):
LGTM 👍 - Thanks @kennylam!

Issue #4550, opened 2020-11-23
React: Footer - Language only - On click of "Clear selected item" icon, id null error pops up.


Praveen-sr:
## Detailed description

Issue: Footer - Language only - On click of "Clear selected item" icon, id null error pops up.

## Additional information

![Footer1.gif](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/1a66ead5-f487-46dd-a7cc-6ae9fe0867ca)

PR #4557, opened 2020-11-23
fix(Footer): fixes language selector clear

kennylam:
### Related Ticket(s)

#4550 

### Description

Fixes an error when clearing the current language selector value

Nov-23-2020 15-55-26

### Changelog

**New**

- add `placeholder` value to language selector text input

**Changed**

- check for `null` value when clearing language selector text input


annawen1 (approved):
LGTM!

Issue #4492, opened 2020-11-13
Web components: mega menu arrow icons not vertically centered with text


oliviaflory:
## Detailed description

> Describe in detail the issue you're having.

The arrow icons on all of the Category heading are not vertically centered with the text, they are too low

The arrow icons on the CTAs at the bottom of the Products & Services and Services & Solutions menu are not vertically centered with the text, they are too high.

**As-is (Safari)**
![Screen Shot 2020-11-13 at 4.00.39 PM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/3a32bcdb-46d1-4582-82ed-7fd23a495935)

**What it should look like (Chrome)**
![Screen Shot 2020-11-13 at 4.05.44 PM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/80361976-2f6e-4cae-9d6a-bdbe95819d0e)

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

bug

> Is this issue related to a specific component?

Mega menu 

> What did you expect to happen? What happened instead? What would you like to
> see changed?

> What browser are you working in?

Safari Version 14.0 (15610.1.28.1.9, 15610)

> What version of Carbon for IBM.com are you using?

[Canary](https://ibmdotcom-web-components-canary.mybluemix.net/iframe.html?id=components-masthead--default)

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

- Screenshots or code
- Notes

IgnacioBecerra:
Seems like this issue has been fixed already?

annawen1:
@IgnacioBecerra I'm seeing the issue on safari on my end - I'm on version Version 14.0.1

IgnacioBecerra:
@annawen1 Ah, my bad, I completely overlooked it was a Safari bug. Thanks for letting me know!

PR #4563, opened 2020-11-24
fix(link-with-icon): fixed uncentered arrow icons on Safari

IgnacioBecerra:
### Related Ticket(s)

#4492

### Description

Aligned uncentered arrow icons that only happen in Safari. Originally an issue with the specific component below, but upon closure inspection all of the components using `LinkWithIcon` in Safari have an uncentered arrow. They can also be seen below -- all the arrow icons are slightly shifted down a few pixels. 

Compared to Chrome:
Screen Shot 2020-11-24 at 2 56 44 PM

### Changelog

**New**

- added in-line style of `display: flex` to the icon `slot` of `link-with-icon.ts` to ensure the arrows are positioned properly on Safari


asudoh (commented):
Thank you for jumping in @IgnacioBecerra! A couple of comments:

1. We avoid using inline styles
2. React does not have `<slot>` and I suspect React may be broken. The same effect of styling `<slot>` tends to be achievable with styling its parent element

IgnacioBecerra:
Thanks for the comments @asudoh! Fortunately, the React version shows up normally on Safari, this is only a Web Component issue.

As for the styles, I had first added a property within `_masthead-megamenu.scss` but after noticing it's an issue within `LinkWithIcon`, I applied the inline style so it affected throughout the other components extending it. Since it's actually only used in Megamenu and Card Footer, I went back with the original fix, now adding it to `link_with_icon.scss` as well to fix the other slightly uncentered arrows.

asudoh:
Thank you for elaborating what you tried @IgnacioBecerra! Would you try the following?

1. Adding `display: block` to https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.12.0/packages/styles/scss/components/link-with-icon/_link-with-icon.scss#L25-L31 - `<svg>`'s default `display: inline` adds space to accommodate one for 'y', 'g', etc. characters. `vertical-align: middle` is there presumably to alleviate that, but `display: block` will be a simpler fix given the icon is placed in flex container.
2. Adding `:host(#{$dds-prefix}-megamenu-link-with-icon) { @extend .#{$prefix}--link-with-icon; }` to [`packages/web-components/src/components/masthead/masthead.scss`](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.13.0-rc.0/packages/web-components/src/components/masthead/masthead.scss) - Without that the style of link with icon is not applied to mega menu one.

I believe doing above eliminates the need for  `* { display: flex; }` and `a * { display: flex; }`. Please ping me for any questions. Thanks!

asudoh (approved):

Issue #3628, opened 2020-08-19
Component:- Image with Caption(Stage) :- On entering large text in copy text field image is not visible fully.


chetankan:
Issue Description:- In image with caption component when large amount of text is entered in copy text filed in that case image is not fully visible.

## Steps to reproduce the issue

1. Enter url in the browser.
2. Select 'Image with Caption'.
3. In knobs , enter 8 to 10 lines of text in copy field.
4. Open component in new tab and click on image.

OS;- Wins 10, Mac
Browsers;- Chrome, Safari, Firefox , Edge.

URL:- https://ibmdotcom-react-staging.mybluemix.net/?path=/story/components-imagewithcaption--default

![Image with Caption with large text.jpg](https://images.zenhubusercontent.com/5e21a841639f517d456f7e43/6b792c32-1ebe-4708-a90c-68d8d7601aef)

PR #4564, opened 2020-11-24
fix(lightbox-media-viewer): add overflow to row 

ariellalgilmore:
### Related Ticket(s)

#3628 

### Description

Image is not visible with large text in copy for `ImageWithCaption`
Screen Shot 2020-11-24 at 1 22 06 PM

### Changelog

**New**

- `overflow-y: auto` to row `div`


asudoh (approved):

Issue #2224, opened 2020-04-28
DBCS languages breaking in the middle of a word


wonilsuhibm:
## Detailed description

> Describe in detail the issue you're having.

When using a DBCS language such as Korean or Japanese, lines wrap in the middle of words.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

This is a bug.

> Is this issue related to a specific component?

It seems to be happening globally in all patterns and components.

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Text lines should break at spaces instead of in the middle of a word.

> What browser are you working in?

- FireFox 68.3.0esr
- Chrome 81.0.4044.122
on Mac OS X 10.15.4

> What version of the IBM.com Library are you using?

The latest through the storybook here (https://ibmdotcom-react.mybluemix.net/?path=/story/components-imagewithcaption--default)

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

N/A

## Steps to reproduce the issue

1. Open the image with caption component on the [Storybook](https://ibmdotcom-react.mybluemix.net/?path=/story/components-imagewithcaption--default)
2. Type in the following Korean text into the heading field:

`한글 테스트 중입니다. 한글 테스트 중입니다. 한글 테스트 중입니다. 한글 테스트 중입니다. 한글 테스트 중입니다. 한글 테스트 중입니다. `

## Additional information

- Screenshots or code
- Notes

wonilsuhibm:
<img width="1135" alt="Screen Shot 2020-04-28 at 5 58 17 PM" src="https://user-images.githubusercontent.com/30183447/80541827-dbce4d00-8979-11ea-9f50-f9ef644eed78.png">

<img width="1091" alt="Screen Shot 2020-04-28 at 5 59 34 PM" src="https://user-images.githubusercontent.com/30183447/80541908-0fa97280-897a-11ea-9dee-ae0681bb3576.png">

ljcarot:
@wonilsuhibm doesn't this get handled by the CMS where they would allow editor to flag 'do not break' word?

asudoh:
[...]

PR #4566, opened 2020-11-25
feat(storybook): language word-break

ariellalgilmore:
### Related Ticket(s)

#2224 

### Description

In storybook using a DBCS language such as Korean or Japanese, lines wrap in the middle of words.
Screen Shot 2020-11-25 at 7 31 06 AM

### Changelog

**New**

- `word-break: keep-all` styles

### DIFF
--- a/packages/react/.storybook/_container.scss
+++ b/packages/react/.storybook/_container.scss
@@ -44,6 +44,7 @@ body {
   background-color: $ui-background;
   color: $text-01;
   height: 100%;
+  word-break: keep-all;
 }
 
 #root,
--- a/packages/web-components/.storybook/container.scss
+++ b/packages/web-components/.storybook/container.scss
@@ -47,6 +47,7 @@ body {
   color: $text-01;
   background-color: $ui-background;
   line-height: 1;
+  word-break: keep-all;
 }
 
 // hide the cookie button


packages/react/.storybook/_container.scss
```diff
@@ -44,6 +44,7 @@ body {
   background-color: $ui-background;
   color: $text-01;
   height: 100%;
+  word-break: keep-all;
```
asudoh, line 47: Can we make it Korean language-only?

asudoh (commented):
Thank you for jumping in @ariellalgilmore!

packages/react/.storybook/_container.scss
ariellalgilmore: When testing I believe it can't detect the language unless the specific tag is specifically set like `<p lang="ko">한</p>`. This then will not change if the a storybook user just tries typing within the knobs.

asudoh: @ariellalgilmore Would you try [`:lang(ko)`](https://developer.mozilla.org/en-US/docs/Web/CSS/:lang? )?

ariellalgilmore: Yeah I tried `  lang(ko) {
    word-break: keep-all;
  }`, but nothing changes in storybook when then typing Korean into the storybook knobs

asudoh: Sorry I think I made a typo in the issue comment. I meant one beginning with `:`. Would you try?

ariellalgilmore: I tried both options

asudoh: Would you want to try to take a deeper look at https://developer.mozilla.org/en-US/docs/Web/CSS/:lang and see how the selector supposed to work? Or would you try to investigate how to make the style change Korean-only? We need to take more time for investigation. Feel free to ping me for any questions. Thanks!

jeffchew:
Hey @ariellalgilmore , just checking if we want to keep this open or close for now and do further investigation? Not sure if there's an easy fix around this or not?

ariellalgilmore:
> Hey @ariellalgilmore , just checking if we want to keep this open or close for now and do further investigation? Not sure if there's an easy fix around this or not?

Hey @jeffchew! I think it makes sense to close for now. @asudoh and I discussed and believe it actually makes sense to be application specific.

PR #4567, opened 2020-11-25
fix(leadspace-block): use styles from styles package and not relative path

annawen1:
### Description

Accidentally left relative path to leadspace-block styles, should import from styles package instead


ariellalgilmore (approved):
LGTM

Issue #3652, opened 2020-08-19
React: Change Local Modal - Add link with icon 


RobertaJHahn:
_**This work is a follow-on to the development work done to create / change the same Web Components version. The Web Components development was managed under epic (#3648)**_

#### User Story

> As a `[user role below]`:
IBM.com Library developer

> I need to:
create/change the `Local Modal`

> so that I can:
provide the ibm.com adopter developers components they can use to build ibm.com web pages

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- **See the Epic for the Design and Functional specs information**
- Design QA testing issue (#3653)
- Prod QA testing issue (#3655)

#### Acceptance criteria
- [ ] Built as a pure React component/variant
- [ ] Include README for the react component and corresponding styles
- [ ] Add any necessary stable selectors
- [ ] Create codesandbox example under `/packages/react/examples/codesandbox` and include in README
- [ ] Minimum 80% unit test coverage
- [ ] The Storybook link is added to the Design QA issue for their testing
- [ ] A comment is posted in the Prod QA issue, tagging Praveen and Chetan, when development is finished

Issue #3648, opened 2020-08-19
Web component: Locale Modal - Add Link with icon


RobertaJHahn:
This epic contains functional and design links that define the scope of the deliverable.  This epic also includes links to all of the work required to deliver the component. 

This was a feature request from Lara.  

#### Additional Information

-  Corresponding React epic (#3651)
- [Functional spec Box link](https://ibm.ent.box.com/folder/82255256134)
- Story within Storybook with corresponding knobs
- Utilize Carbon
- Create with Shadow DOM and Custom Elements standards
- **See the Epic for the Design and Functional specs information**
- Design QA testing issue (#3649)
- Prod QA testing issue (#3650)

#### Acceptance criteria
- [ ] Include README for the web component and corresponding styles
- [ ] Create Web Components styles in styles package
- [ ] Do not create knobs in Storybook that include JSON objects
- [ ] Break out Storybook stories into multiple variation stories, if applicable
- [ ] Create codesandbox example under `/packages/web-components/examples/codesandbox` and include in README
- [ ] Minimum 80% unit test coverage
- [ ] If a design is provided, the Designer is included as a Reviewer in the Pull Request
- [ ] Provide a direct link to the deploy preview for the designer in the Pull Request description
- [ ] A comment is posted in the Design QA issue, tagging Designer, when development is finished
- [ ] The Storybook link is added to the Design QA issue for their testing
- [ ] A comment is posted in the Prod QA issue, tagging Praveen and Chetan, when development is finished

RobertaJHahn:
@IgnacioBecerra The PR should be linked to the development issues, not the epic.  The epics remain in the Prioritized Epics pipeline until the link issues are closed.  The dev issue is (#3446).  Gui referenced your PR in that issue.  I am moving this epic back to the Prioritized Epics pipeline and assigning Ignacio to the dev issue as he did the work.

PR #4568, opened 2020-11-26
feat(locale-modal): link with icon replacement

IgnacioBecerra:
### Related Ticket(s)

#3648 
#3652 

### Description

Replaced the button with a `LinkWithIcon` instead.

Before:
Screen Shot 2020-11-25 at 6 07 17 PM

After:
Screen Shot 2020-11-25 at 6 06 51 PM

### Changelog

**Changed**

- replaced button with `dds-link-with-icon` in Web Components
- replaced icon and text with `<LinkWithIcon>` in React


packages/web-components/src/components/locale-modal/locale-modal.scss
```diff
@@ -18,6 +18,10 @@
   #{$dds-prefix}-expressive-modal-header {
     padding: 0 20% 0 $spacing-05;
     margin-bottom: $carbon--spacing-03;
+
+    > * {
+      width: fit-content;
+    }
```
asudoh, line 24: Could you explain what this style rule is for?

asudoh (commented):

packages/web-components/src/components/locale-modal/locale-modal.scss
IgnacioBecerra: When using `LinkWithIcon`, the component takes the whole width of the Locale Modal Header, so when tabbing into focus there's a long rectangle focusing on empty space. Clicking on the empty space is also considered as if the content is being clicked. Using this style rule makes it so it only wraps around the actual content and avoid these problems.

Btw, I tried using `display: inline-block`. While it also works, it shifts the component down an unspecified amount of pixels and I wasn't able to replicate the original positioning.

asudoh: I see, thanks @IgnacioBecerra for explaining! Looks that you wanted `:host(#{$dds-prefix}-locale-modal) #{$dds-prefix}-expressive-modal-heading { width: fit-content }`? We should try to avoid `*` selector unless it's absolutely necessary to make it more specific and avoid selecting elements that we don't intend.

IgnacioBecerra: Oh I see, that's what I wanted, thanks! 

Before using the `*` selector, I tried many selector combinations (including a similar one to the one above, except I used `:host` instead for the second one). The nuances about Web Component styling vs regular are interesting, thanks for always helping out with them @asudoh !

asudoh (approved):

jeffchew:
@IgnacioBecerra looks like snapshots are out of date, can you update?

PR #4573, opened 2020-11-30
fix(translation): support for running in node.js environment

asudoh:
### Description

Another fix to support our translation service to run in node.js environment, by using session storage only if it's available.

### Changelog

**Changed**

- A change to our translation service code to use session storage only if it's available.


jeffchew (approved):
Thank you @asudoh !

PR #4574, opened 2020-11-30
fix(masthead): introduce separate L1 nav variant

asudoh:
### Description

introduces `<dds-top-nav-l1>` that disables the feature to hide the nav when search box is open.

### Changelog

**New**

- `<dds-top-nav-l1>` that disables the feature to hide the nav when search box is open.


asudoh:
From the original code, I have an impression that L1 should hide the divider between the product name and the nav items, but anybody please correctly if I'm wrong here. Thanks!

jeffchew:
Thank you @asudoh ! This looks like is based on the issue that was brought up in the QA call this morning, correct? Let's tie this PR to that once the issue is opened by @Praveen-sr .

jeffchew (approved):
Looks good! Let's address the divider in a separate PR in case we need to change it, but at least we knock out the higher severity issue here.

kennylam (approved):
LGTM thanks @asudoh!

PR #4578, opened 2020-11-30
chore(percy): skipping card stories that use sameHeight

jeffchew:
### Related Ticket(s)

No related issue

### Description

When Percy runs its diff, it is unknown when the snapshot is created in
relation to when the sameHeight utility kicks off, thus creating random
false positives. The team has opted to skip these tests in Percy.

### Changelog

**Changed**

- skipping `content block cards` and `card section simple with CTA` in Percy tests


IgnacioBecerra (approved):
LGTM! Thanks @jeffchew, those constant Percy diffs were driving me crazy haha

Issue #3627, opened 2020-08-19
Component: Locale Modal(Stage) :- Message is not getting reset on search by country or language name


chetankan:
Issue Description:-  On locale modal component "This page is unavailable in your preferred location or language" is not getting rest when user searches for country / language on different geo.

## Steps to reproduce the issue

1. Enter url in browser.
2. Select locale modal.
3. Click on 'Americas'
4. In search box enter 'German'.
5. Message will be displayed saying "This page is unavailable in your preferred location or language".
6.Now click on 'Select geographic area', and select 'Europe'
7. Now observe the message below search box.

OS:- Wins 10, Mac, iOS 12, 13 and Android 8,9
Browsers:-  Chrome, Safari, Firefox, Edge, Samsung Internet

URL:- https://ibmdotcom-react-staging.mybluemix.net/?path=/story/components-locale-modal--default

PR #4587, opened 2020-11-30
fix(locale-modal): reset message upon unavailability

IgnacioBecerra:
### Related Ticket(s)

#3627 

### Description

Availability message now resets after backing out of the Country/Language search function. 

Previously, after not getting any results back, the unavailability message persisted even after clicking on a new Region, even if there was nothing being queried yet.

Before:
Screen Shot 2020-11-30 at 3 51 35 PM

### Changelog

**New**

- reset availability message on `LocalModalCountries.js` after backing out from search function


kennylam (approved):
Looks good thanks @IgnacioBecerra! Looks like you just need to update the snapshots.

Issue #4591, opened 2020-12-01
Web components: Masthead with L1 - On Tab view both hamburger menu and L1 menu options are displayed.


Praveen-sr:
## Detailed description

On iPad & Tab S7
Issue: Masthead with L1 - @800px to 1000px both hamburger menu and L1 menu options are displayed.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/a1bb7bcb-0402-4678-b58b-2d07539e204e)

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/8b900cc1-9e23-43e9-81f1-59b2e265786d)

PR #4593, opened 2020-12-01
fix(masthead): adjust breakpoints

asudoh:
### Related Ticket(s)

Refs #4591.

### Description

Aligns all the style code of `<dds-top-nav>` to the intended structure, by using `<nav class="bx--header__nav">` instead of its host element (`<dds-top-nav>`) to dictate the style including its responsive behavior. Doing so ensures L1 top navigation bar in masthead won't be shown unless left nav (hamberger menu) is hidden.

### Changelog

**Changed**

- Aligns all the style code of `<dds-top-nav>` to the intended structure, by using `<nav class="bx--header__nav">` instead of its host element (`<dds-top-nav>`) to dictate the style including its responsive behavior.


kennylam (approved):
LGTM thanks @asudoh!

Issue #4234, opened 2020-10-15
NextJS page: Content Group Horizontal section, links are misaligned. 


Praveen-sr:
## Detailed description

Issue: On nextJS page, user scrolls to Content Group Horizontal section, links are misaligned. 

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/ecc82898-0872-4b6d-95a0-982876218c1e)

PR #4601, opened 2020-12-01
fix(ContentGroupHorizontal): row to 100% width for consistent alignment

annawen1:
### Related Ticket(s)

NextJS page: Content Group Horizontal section, links are misaligned. #4234

### Description

The Content Item Horizontal can become misaligned when the content text is of different lengths as there is no set width for the wrapper div. Setting the `row` div wrapper to take up `width: 100%` solves this issue.

BEFORE:
Screen Shot 2020-12-01 at 6 40 43 PM

AFTER:
Screen Shot 2020-12-01 at 6 23 27 PM

### Changelog

**Changed**

- set content-item-horizontal row div to 100% width


asudoh (approved):
LGTM assuming the Percy diff is intentional - Thanks @annawen1!

Issue #4493, opened 2020-11-13
Web components: mega menu Safari lag in responsive desktop resizing


oliviaflory:
## Detailed description

> Describe in detail the issue you're having.

The mega menu has a significant lag when resizing the browser on a desktop in Safari.

[Screen recording](https://ibm.box.com/s/fmpagp3d6em0nahg2jymjje1q8mzb6kd)

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug (?)

> Is this issue related to a specific component?

web component mega menu

> What did you expect to happen? What happened instead? What would you like to
> see changed?

> What browser are you working in?

Version 14.0 (15610.1.28.1.9, 15610)

> What version of Carbon for IBM.com are you using?

[Canary](https://ibmdotcom-web-components-canary.mybluemix.net/iframe.html?id=components-masthead--default)

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. Open web component mega menu in safari
2. Scale browser width (while mega menu is open)
3. Step three
4. etc.

## Additional information

- Screenshots or code
- Notes

wonilsuhibm:
I'm also seeing layout issue between large (1056px) and max (1584px) breakpoints—the forth column starts disappearing

### As-is
<img width="1304" alt="Screen Shot 2020-11-19 at 8 48 51 AM" src="https://user-images.githubusercontent.com/30183447/99674600-48a75280-2a44-11eb-9664-c0b91a399dc2.png">

### To-be (as seen in the React version)
<img width="1198" alt="Screen Shot 2020-11-19 at 8 51 11 AM" src="https://user-images.githubusercontent.com/30183447/99674738-72f91000-2a44-11eb-9e8d-e56a24ef86bd.png">

PR #4602, opened 2020-12-02
fix(masthead): workaround Safari bug in Shadow DOM

asudoh:
### Related Ticket(s)

Refs #4493.

### Description

Adds a code to `<dds-megamenu-top-nav-menu>` to manually set viewport width via `ResizeObserver` and use it in our CSS. This works around Safari's bug where `100vw` in Shadow DOM causes delayed rendering.

### Changelog

**New**

- A code to `<dds-megamenu-top-nav-menu>` to manually set viewport width via `ResizeObserver` and use it in our CSS.


annawen1 (approved):
😯 Looks good to me! Thanks for working on this @asudoh!

Issue #4519, opened 2020-11-18
Callout quote component link color is not accessible


wonilsuhibm:
## Detailed description

The link at the bottom of this component (see the attached image) currently does not pass AA/AAA. It needs to be blue 40.

## Steps to reproduce the issue

1. Open this storybook page: https://ibmdotcom-web-components-staging.mybluemix.net/?path=/story/components-callout-quote--default
2. Check the color of the link

## Additional information

<img width="1303" alt="Screen Shot 2020-11-18 at 6 40 27 PM" src="https://user-images.githubusercontent.com/30183447/99601668-db58da80-29cd-11eb-8417-fef315c62ab7.png">

wonilsuhibm:
<img width="1303" alt="Screen Shot 2020-11-18 at 6 40 27 PM" src="https://user-images.githubusercontent.com/30183447/99601668-db58da80-29cd-11eb-8417-fef315c62ab7.png">

PR #4603, opened 2020-12-02
fix(callout): apply theme zoning

asudoh:
### Related Ticket(s)

Refs #4519.

### Description

Applies `g90` theme to callout so `<dds-link-with-icon>` put in callout automatically changes the link color.

### Changelog

**Changed**

- Applies `g90` theme to callout.


kennylam (approved):
LGTM thanks @asudoh!

Issue #4596, opened 2020-12-01
Web component:  Leaving IBM - No label for the dialog ( both jaws and iPhone)


Praveen-sr:
## Detailed description

Issue: No label for the dialog ( both jaws and iPhone)

Expected:
Provide label for the dialog by associating the Heading using aria-labelledby
4.1.2 Name, Role, Value

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/5b24dca3-10a3-40c8-a80d-0864fed0b60c)

PR #4607, opened 2020-12-02
fix(leaving-ibm): add dialog role and associating with heading by aria-labelledby attribute

annawen1:
### Related Ticket(s)

Web component: Leaving IBM - No label for the dialog ( both jaws and iPhone) #4596

### Description

Add dialog `role` attribute to the `<dds-leaving-ibm-modal>` and associate it with the `<dds-leaving-ibm-heading>` according to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role

### Changelog

**New**

- dialog `role` attribute to the `<dds-leaving-ibm-modal>`
- add `id` to the `<dds-leaving-ibm-heading>` to connect the modal


packages/web-components/src/components/leaving-ibm/leaving-ibm-modal.ts
```diff
@@ -22,6 +22,16 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
...
 class DDSLeavingIbmModal extends StableSelectorMixin(BXModal) {
+  connectedCallback() {
+    if (!this.hasAttribute('role')) {
+      this.setAttribute('role', 'dialog');
+    }
+    if (!this.hasAttribute('aria-labelledby')) {
+      this.setAttribute('aria-labelledby', 'leaving-ibm-heading');
+    }
+    super.connectedCallback();
+  }
```
asudoh, line 33: To avoid (theoretical) ID duplication (though it's less likely for an app to have more than one leaving IBM modal), we can auto-generate the ID, like below.`@HostListener` should be from `import HostListener from 'carbon-web-components/es/globals/decorators/host-listener.js'`:

```suggestion
  /**
   * The unique ID for ID ref.
   */
  private _uniqueId = `__${ddsPrefix}-leaving-ibm-modal__${Math.random()
    .toString(36)
    .slice(2)}`;

  /**
   * Handles `slotchange` event.
   *
   * @param event The event.
   */
  @HostListener('shadowRoot:slotchange')
  // @ts-ignore: The decorator refers to this method but TS thinks this method is not referred to
  private _handleSlotChange = (event: Event) => {
    const { selectorHeading } = this.constructor as typeof DDSLeavingIbmModal;
    if (!this.hasAttribute('aria-labelledby')) {
      const headingNode = (event.target as HTMLSlotElement).assignedNodes().reduce((acc, node) => {
        if ((node as Element).matches?.(selectorHeading)) {
          acc.push(node);
        } else {
          acc.push(...((node as Element).querySelectorAll?.(selectorHeading) ?? []));
        }
        return acc;
      }, [] as Node[])[0] as Element | void;
      if (headingNode) {
        if (!headingNode.id) {
          const { id, _uniqueId: uniqueId } = this;
          headingNode.id = id || uniqueId;
        }
        if (headingNode.id) {
          this.setAttribute('aria-labelledby', headingNode.id);
        }
      }
    }
  };

  connectedCallback() {
    if (!this.hasAttribute('role')) {
      this.setAttribute('role', 'dialog');
    }
    super.connectedCallback();
  }

  /**
   * A selector selecting the heading.
   */
  static get selectorHeading() {
    return `${ddsPrefix}-leaving-ibm-modal-heading`;
  }
```

packages/web-components/src/components/leaving-ibm/leaving-ibm-modal-heading.ts
```diff
@@ -21,6 +21,13 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
  */
 @customElement(`${ddsPrefix}-leaving-ibm-modal-heading`)
 class DDSLeavingIbmModalHeading extends BXModalHeading {
+  connectedCallback() {
+    if (!this.hasAttribute('id')) {
+      this.setAttribute('id', 'leaving-ibm-heading');
+    }
+    super.connectedCallback();
+  }
+
```
asudoh, line 30: To avoid (theoretical) ID duplication (though it's less likely for an app to have more than one leaving IBM modal), we can let `<dds-leaving-ibm-modal>` auto-generate the ID:

```suggestion
```

asudoh (commented):
Thank you for jumping in @annawen1!

packages/web-components/src/components/leaving-ibm/leaving-ibm-modal.ts
annawen1: oh good point! Thanks @asudoh!

asudoh (approved):
Thanks @annawen1 for the update!

Issue #4281, opened 2020-10-21
Utility:  Create the MVP animation / fade utility


RobertaJHahn:
#### User Story

> As a `[user role below]`:
IBM.com Library developer

> I need to:
create the `animation / fade utility`

> so that I can:
provide the ibm.com adopter developers with a utility that will enable them to include animation in the ibm.com web pages they build

#### Additional information

- **See the Epic for the Design and Functional specs information**
- [design specs for the scroll into view utility](https://ibm.box.com/s/raw94fq6dcfuey786zks0c6mx24kf0cd)
- Prod QA testing issue (#4433)

#### Acceptance criteria
- [ ] Include README for the web component and corresponding styles
- [ ] Minimum 80% unit test coverage
- [ ] Add this feature to the React NextJS test page for testing
- [ ] A comment is posted in the Prod QA issue, tagging Praveen when development is finished

jeffchew:
Look into using Intersection Observer API for this.

PR #4608, opened 2020-12-03
feat(fade-in-out): introducing new components

IgnacioBecerra:
### Related Ticket(s)
#4281

### Description

Utility made for adopters that will enable them to include fade in/out animations for the ibm.com web pages they build.

`fadeInOut` also includes parameter of 
`keepAnimation`: to define whether the animation persists after the first activation (defaults to `true`)

`fadeInOut` also supports custom delay duration by setting `--#{$dds-prefix}--scroll-into-view-delay` to the desired delay inside the targeted class declaration.

https://user-images.githubusercontent.com/24970122/104062914-2c0bd900-51b0-11eb-92df-b81717d1ccc4.mov

``` 
import '@carbon/ibmdotcom-styles/scss/internal/scroll-into-view/_scroll-into-view.scss';

const selectorTargets = 'bx--content-block, bx--content-group'

// React
import { FadeInOut } from '@carbon/ibmdotcom-react';
<FadeInOut selectorTargets={selectorTargets} keepAnimations={false} />

// Web Component
import '@carbon/ibmdotcom-web-components/es/components/fade-in-out/fade-in-out.js';
<dds-fade-in-out .selectorTargets="${selectorTargets}"></dds-fade-in-out>


// custom-app.scss
.bx--content-block {
    --#{$dds-prefix}--fade-in-out-delay: 3s; // custom delay
}

```

### Storybook Variation
Added a new Storybook variation for `Dotcom Shell`, `With Fade Animations` for testing purposes.
Under `Other` knobs, one can adjust the delay in seconds to give an idea of the custom delay that can be used. 

### Changelog

**New**

- `fade-in-out` React and Web Component introduced
- `With Fade Animations` Storybook variations for `Dotcom Shell` for testing purposes


packages/utilities/src/utilities/scrollIntoView/scrollIntoView.js
```diff
@@ -0,0 +1,75 @@
...
+            if (!iterations) {
+              observer.unobserve(entry.target);
+            }
+          } else {
+            entry.target.style.transitionTimingFunction = motion(
+              'exit',
+              'expressive'
+            );
+            entry.target.style.transitionDuration = delay;
+            entry.target.style.opacity = 0;
```
asudoh, line 61: Would you be able to add/remove a CSS class, instead of applying inline style?

asudoh (commented):
Good to see the progress @IgnacioBecerra!

packages/utilities/src/utilities/scrollIntoView/scrollIntoView.js
```diff
@@ -25,40 +33,32 @@ const viewportMargin =
...
     () => {
       const elements = document.querySelectorAll(selector);
-
-      console.log(elements);
-
-      const options = {
-        rootMargin: viewportMargin,
-        threshold: 0,
-      };
+      const root = document.querySelector(':root');
```
asudoh, line 46:
```suggestion
      const root = document.documentElement;
```

packages/utilities/src/utilities/scrollIntoView/scrollIntoView.js
```diff
@@ -25,40 +33,32 @@ const viewportMargin =
...
 
       const observer = new IntersectionObserver(function handleIntersect(
         entries
       ) {
+        root.style.setProperty('--delay', delay);
```
asudoh, line 51: Probably we can simply let application set the CSS custom property in the application CSS?

packages/utilities/src/utilities/scrollIntoView/scroll-into-view.scss
```diff
@@ -0,0 +1,24 @@
...
+ * Copyright IBM Corp. 2016, 2020
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+@import '@carbon/motion/scss/motion.scss';
+
+:root {
+  --delay: 400ms;
```
asudoh, line 11: To avoid variable name conflict with other CSS libraries and application code:

```suggestion
  --#{$dds-prefix}--delay: 400ms;
```

packages/utilities/src/utilities/scrollIntoView/scroll-into-view.scss
```diff
@@ -0,0 +1,24 @@
...
+}
+
+.bx--fade-in {
+  transition-timing-function: carbon--motion(entrance, expressive);
+  transition-duration: var(--delay);
```
asudoh, line 16:
```suggestion
  transition-duration: var(--#{$dds-prefix}--delay);
```

packages/utilities/src/utilities/scrollIntoView/scroll-into-view.scss
```diff
@@ -0,0 +1,24 @@
...
+  opacity: 1;
+}
+
+.bx--fade-out {
+  transition-timing-function: carbon--motion(exit, expressive);
+  transition-duration: var(--delay);
```
asudoh, line 22:
```suggestion
  transition-duration: var(--#{$dds-prefix}--delay);
```

packages/utilities/src/utilities/scrollIntoView/scrollIntoView.js
IgnacioBecerra: The func specs detailed an option to allow adopters to provide their custom delay, this is in case the adopter wants to use different delays for different components (e.g. target `boxA` for `3s` and `boxB` for `1s` or so). This way the duration can differ between components over setting a "global" delay.

packages/utilities/src/utilities/scrollIntoView/scrollIntoView.js
```diff
@@ -0,0 +1,78 @@
...
+    Math.round(document.documentElement.clientHeight / (100 / 12.5)) + 32
+  ).toString() +
+  'px 0px';
+
+const options = {
+  rootMargin: viewportMargin,
+  threshold: 0,
+};
+
+const scrollIntoView = (selector, delay = '400ms', iterations = true) => {
```
asudoh, line 44:
```suggestion
const scrollIntoView = (selector, iterations = true) => {
```

packages/utilities/src/utilities/scrollIntoView/scrollIntoView.js
```diff
@@ -0,0 +1,78 @@
...
+  window.addEventListener(
+    'load',
+    () => {
+      const elements = document.querySelectorAll(selector);
+      const root = document.documentElement;
+
+      const observer = new IntersectionObserver(function handleIntersect(
+        entries
+      ) {
+        root.style.setProperty(`--${stablePrefix}--delay`, delay);
```
asudoh, line 54: I suggest we let our users set the CSS custom property in the application's CSS code:

```suggestion
```

packages/utilities/src/utilities/scrollIntoView/scrollIntoView.js
```diff
@@ -0,0 +1,78 @@
+/**
+ * Copyright IBM Corp. 2016, 2020
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import '@carbon/ibmdotcom-styles/scss/internal/scroll-into-view/_scroll-into-view.scss';
```
asudoh, line 8: Wondering if we should import our users use this Sass file by themselves, instead of putting the styles in the utilities package, which means:

```suggestion
```

asudoh (commented):

packages/utilities/src/utilities/scrollIntoView/scrollIntoView.js
asudoh: I see and the custom delay can be done by:

```
.app--some-area-containing-scroll-into-view {
  --dds-delay: 3s
}
```

packages/styles/scss/internal/scroll-into-view/_scroll-into-view.scss
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+@import '../../globals/vars';
+@import '../../globals/imports';
+
+:root {
+  --#{$dds-prefix}--delay: 400ms;
```
asudoh, line 12: My apologies for another name change request, but I figured that the variable should be more specific of what it does (instead of using name implying a general purpose):

```suggestion
  --#{$dds-prefix}--scroll-into-view-delay: 400ms;
```

IgnacioBecerra: You're right, the more specific the better to avoid any potential conflicts 👍

packages/utilities/src/utilities/scrollIntoView/scrollIntoView.js
IgnacioBecerra: Ah I see! That's definitely a better way to allow the adopter to have their own delay within their own CSS file.

asudoh (approved):

jeffchew (changes requested):
Adding notes here from what we discussed offline:

- Create a story in the Dotcom Shell with the utility enabled, so we can see it in action
- The JSDocs generation isn't quite right, from my notes:
  -  it doesn't look like `scrollIntoView` is showing up in the JSDoc generation, but `viewportMargin` is: https://ibmdotcom-utilities.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/4608/global.html#viewportMargin. you might need to make `viewportMargin` and `options` into private variables (add `@private` to the JSDoc description of those), and move the JSDoc declaration to on top of `scrollIntoView`. Also, rename `viewportMargin` and `options` to `_viewportMargin` and `_options`

jeffchew:
Thanks @IgnacioBecerra !

Can you update the snapshot file too?

jeffchew:
@IgnacioBecerra would it also be too much effort to also add the story to dotcom shell in web components too?

jeffchew:
@IgnacioBecerra thank you for adding! And sorry if this is unraveling other feedback, I noticed that when opening in web components in a new window, the title disappears:

Screen Shot 2020-12-04 at 3 17 33 PM

Also, perhaps add additional mock content so that the page scrolls a bit more?

packages/utilities/src/utilities/scrollIntoView/scrollIntoView.js
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * .bx--content-block {
+ *   --#{$dds-prefix}--scroll-into-view-delay: 3s;
+ * }
+ *
+ * @param {*} selector menu item selector id
+ * @param {boolean} iterations to define whether its continuous or not
+ */
+
+const scrollIntoView = (selector, iterations = true) => {
```
photodow, line 55: Mere question, but will this only allow for a string selector, or can it pass in a Node object?

photodow (changes requested):
A couple of small details...

1. So right now it's coming in and exiting is a little off. You can visibly see it fade out, and that's not supposed to happen. It's supposed to stay visible until it leaves the visible viewport.  


2. There appears to be a bug in certain situations where it scrolls into view ok, but then when it fades out and you try to go back it is completely gone. It looks like it's a problem if it's still fading in and you quickly change course and it gets confused? See the gif below.  


3. There is a scenario i'm seeing where it fades in. You scroll down, and it goes out of the page. When you go back up it doesn't fade out (my expectation is it would fade in at this point), and yet if you go up until it's out of view and scroll down it fades back in. It's either that or it's fading in too early. Not sure the giphy helps much here... Maybe it's coming in too early because you're not accounting for the masthead as a grid influencer maybe?  


4. Minor storybook knob clarification. The delay isn't clear what you're delaying. Maybe be more specific? scroll animation delay maybe? Does that delay them all globally?


5. Can we add an option to switch iterations on or off to storybook?

6. Sometime soon we'll be adding other types of animations beyond just fade. Has there been any thought into how this scales to allow different types of animations, and maybe combining them?

Looking good! Let me know if you have any questions, or need to meet to hash somethings out. I can take another look when some of these get addressed. Thanks!

packages/utilities/src/utilities/scrollIntoView/scrollIntoView.js
```diff
@@ -0,0 +1,86 @@
...
+ * For default values of 400ms and continuous play:
+ * scrollIntoView('.bx--content-block')
+ *
+ * With 'one and done' option:
+ * scrollIntoView('.bx--content-block', false)
+ *
+ * For custom delay time, set within targeted class in the application's CSS code as such:
+ *
+ * .bx--content-block {
+ *   --#{$dds-prefix}--scroll-into-view-delay: 3s;
```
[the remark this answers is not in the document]
IgnacioBecerra: Hm yeah, I originally wrote it like this, but since we're trying to move away from inline styling, I moved forward with the other approach by having the adopters set the delay within their own classes. I think both approaches have their pros and cons, but I do agree it'd be easier for the adopter to set it through the API. 

@asudoh I'm wondering if you have some thoughts about this?

packages/utilities/src/utilities/scrollIntoView/scrollIntoView.js
IgnacioBecerra: Currently it's set to receive strings, mainly since it's wrapped around a listener that waits until the whole page is loaded before querying the element. If the adopters can make sure the element exists before calling the function we could make it so they can pass node objects instead.

packages/web-components/src/components/dotcom-shell/__stories__/dotcom-shell.stories.ts
```diff
@@ -126,6 +129,79 @@ export const Default = ({ parameters }) => {
...
+
+withFadeAnimations.story = {
+  parameters: {
+    knobs: {
+      DotcomShell: () => {
+        scrollIntoView('h2');
+        scrollIntoView('p');
+
+        const delayOptions = {
+          Default: '400ms',
```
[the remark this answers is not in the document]
IgnacioBecerra: Ah, it seems that I misread the func specs, I thought the `delay` mentioned as an option was the same as the "duration" of the effect. I'll make sure to separate them.

packages/utilities/src/utilities/scrollIntoView/scrollIntoView.js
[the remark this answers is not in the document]
asudoh: @IgnacioBecerra is right, we are avoiding defining style by JavaScript unless it's absolutely necessary. If we use inline style there is no way for application style to override it.

packages/react/src/components/DotcomShell/__stories__/DotcomShell.stories.js
```diff
@@ -301,3 +303,52 @@ WithL1.story = {
...
+    'carbon-theme': { disabled: true },
+    knobs: {
+      escapeHTML: false,
+      DotcomShell: () => {
+        const {
+          Masthead: mastheadKnobs,
+        } = mastheadStory.story.parameters.knobs;
+        const { Footer: footerKnobs } = footerStory.story.parameters.knobs;
+
+        scrollIntoView('.bx--content-block');
```
asudoh, line 325: Can we move this to the story implementation rather than knob definition?

packages/react/src/components/DotcomShell/__stories__/DotcomShell.stories.js
```diff
@@ -301,3 +303,52 @@ WithL1.story = {
...
+          1: '1s',
+          3: '3s',
+          5: '5s',
+        };
+        const defaultDelay = '400s';
+        let delay = select('Delay (in seconds)', delayOptions, defaultDelay);
+
+        document.querySelectorAll('.bx--content-block').forEach(e => {
+          e.style.setProperty('--dds--scroll-into-view-delay', delay);
+        });
```
asudoh, line 338: I think we should better implement this in story-specific Sass file. The API confusion from @photodow may have come from this.

asudoh (commented):
Some additional comments - Thanks @IgnacioBecerra!

packages/utilities/src/utilities/scrollIntoView/scrollIntoView.js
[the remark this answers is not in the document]
photodow: @asudoh not sure I'm fully following that logic. What's a scenario that is "absolutely necessary"? The adopters would be the ones setting this, wouldn't they? This isn't stopping them from resetting it, does it? It seems odd to me they'd have to touch two separate files to get this working, and it uses two different sets of APIs. It also feels laborious to implement a proper delay that builds on top of itself using `SCSS`, and the logic that might be needed for the delay might be built out separately or duplicated. I don't know... Keep doing it if y'all are sure this is the right direction, but this doesn't jive in my brain.

packages/react/src/components/DotcomShell/__stories__/DotcomShell.stories.js
photodow: No, this is not where the confusion is stemming from, but If you expect users to set this in the `SCSS` then I would mimic that behavior. "Lead by example" lol. The problem with that, that I imagine @IgnacioBecerra was going for is he wanted to set the delay programmatically via knobs.

IgnacioBecerra:
So, couple things that I've noted while working on this a bit more.

I've been looking for a way to make it so the element disappears once it reaches the top using a single `IntersectionObserver`, but since the API stops firing after crossing the threshold (as pictured by the purple box below), one solution would be to have two `IntersectionObserver` objects: One with the current rootMargin, and one with `0px`. 

The former will take care of the `fade-in` effect, while the latter will make it disappear.

Screen Shot 2020-12-11 at 1 45 57 PM

As for the resizing issue, `IntersectionObserver` currently does not support the dynamic setting of the `rootMargin`, only leaving us with the option to replace the current `observer` object with a new one with the updated `rootMargin` every time the window gets resized.

I'm not sure if implementing these two approaches are efficient compared to adding an event listener for `scroll` and having the functionality within it. Do you guys have any thoughts about this?

packages/utilities/src/utilities/scrollIntoView/scrollIntoView.js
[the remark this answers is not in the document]
asudoh: @photodow Application may want to define the rule of scrolling delay enforced to everywhere in the application, or everywhere in a specific portion. However, if a JavaScript code defines inline style, such inline style wins all the time and thus there is no way (that is natural in CSS world) for setting it application wide.

packages/react/src/components/DotcomShell/__stories__/DotcomShell.stories.js
asudoh: @photodow Yeah I can imagine that. This can be achieved by defining CSS rules like below and set the class name via knob:

```css
.dds-devenv--scroll-into-view--delay-default {
  --dds--scroll-into-view-delay: 400ms;
}

.dds-devenv--scroll-into-view--delay-slight-slow {
  --dds--scroll-into-view-delay: 1s;
}

.dds-devenv--scroll-into-view--delay-slow {
  --dds--scroll-into-view-delay: 3s;
}

.dds-devenv--scroll-into-view--delay-very-slow {
  --dds--scroll-into-view-delay: 5s;
}
```

asudoh:
@IgnacioBecerra One key merit of `IntersectionObserver` is that you don't need to ask browser for layout (e.g. with `getBoundingClientRect()`) to see if an element is "scrolled into view". Another key merit of `IntersectionObserver` is that the callback runs only when there is a change in intersection state, instead of running it every time user scrolls (`onscroll` tends to fire dozens of times per user's scroll gesture that catches the middle of scrolling state). Before `IntersectionObserver` was introduced, web apps had to do hard work to throttle `onscroll` so such expensive calculation of intersection state happens less often.

packages/utilities/src/utilities/scrollIntoView/scrollIntoView.js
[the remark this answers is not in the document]
photodow: @asudoh I'm not bought into the user experience you're creating here at the moment. I also don't know if we want other parts of the CSS overriding this value. I personally feel like this would be a scenario where inline would make sense, but for the sake of progress I'll step aside, and we can see where this takes us. Thanks as always @asudoh!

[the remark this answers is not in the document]
asudoh: Thanks @photodow for your thoughts! Another option is using an enum (that corresponds to a set of CSS classes) rather than a direct number.

packages/utilities/src/utilities/scrollIntoView/scrollIntoView.js
```diff
@@ -0,0 +1,163 @@
...
+ *
+ * .bx--content-block {
+ *   --#{$dds-prefix}--scroll-into-view-delay: 250ms;
+ * }
+ *
+ * @param {*} selector menu item selector id
+ * @param {boolean} iterations to define whether its continuous or not
+ */
+
+const scrollIntoView = (selector, iterations = false) => {
```
asudoh, line 99: As mentioned offline:

```suggestion
const scrollIntoView = (selector, { iterations = false = {}) => {
```

asudoh (commented):

packages/react/src/components/DotcomShell/__stories__/DotcomShell.stories.js
```diff
@@ -301,3 +313,46 @@ WithL1.story = {
     },
   },
 };
+
+export const WithFadeAnimations = ({ parameters }) => {
+  let { delay, iterations } = parameters?.props?.DotcomShell ?? {};
+
+  scrollIntoView('.bx--content-block', iterations);
```
asudoh, line 320:
```suggestion
  scrollIntoView('.bx--content-block', { iterations });
```

packages/utilities/src/utilities/scrollIntoView/scrollIntoView.js
```diff
@@ -0,0 +1,163 @@
...
+ *
+ * .bx--content-block {
+ *   --#{$dds-prefix}--scroll-into-view-delay: 250ms;
+ * }
+ *
+ * @param {*} selector menu item selector id
+ * @param {boolean} iterations to define whether its continuous or not
+ */
+
+const scrollIntoView = (selector, { iterations = false = {}) => {
```
asudoh, line 99: Sorry I seem to have made a typo:

```suggestion
const scrollIntoView = (selector, { iterations = false } = {}) => {
```

packages/web-components/src/components/dotcom-shell/__stories__/dotcom-shell.stories.ts
```diff
@@ -344,6 +358,71 @@ export const Default = ({ parameters }) => {
   `;
 };
 
+export const withFadeAnimations = ({ parameters }) => {
+  const { brandName, userStatus, navLinks } = parameters?.props?.MastheadComposite ?? {};
+  const { langDisplay, language, size: footerSize, legalLinks, links: footerLinks, localeList } =
+    parameters?.props?.FooterComposite ?? {};
+  const { useMock, delay, iterations } = parameters?.props.FadeOptions ?? {};
+
+  scrollIntoView('dds-table-of-contents > *', iterations);
```
asudoh, line 367:
```suggestion
  scrollIntoView('dds-table-of-contents > *', { iterations });
```

packages/utilities/src/utilities/scrollIntoView/scrollIntoView.js
```diff
@@ -0,0 +1,163 @@
...
+ * import '@carbon/ibmdotcom-styles/scss/internal/scroll-into-view/_scroll-into-view.scss';
+ * import { scrollIntoView } from '@carbon/ibmdotcom-utilities';
+ *
+ * As an example, the function can be called to target '.bx--content-block' as such:
+ *
+ * For default values of 400ms and 'one and done' play:
+ * scrollIntoView('.bx--content-block')
+ *
+ * With 'continous play' option:
+ * scrollIntoView('.bx--content-block', true)
```
asudoh, line 87:
```suggestion
 * scrollIntoView('.bx--content-block', { iterations: true })
```

asudoh (commented):
Thank you for all the updates @IgnacioBecerra!

packages/utilities/src/utilities/scrollIntoView/scrollIntoView.js
IgnacioBecerra: seems like I forgot that one, thanks!

packages/utilities/src/utilities/scrollIntoView/scrollIntoView.js
```diff
@@ -0,0 +1,176 @@
...
+        elements.forEach(e => {
+          _rootObserver.observe(e);
+          _innerObserver.observe(e);
+        });
+      });
+      _resizeObserver.observe(document.documentElement);
+    },
+    false
+  );
+};
```
asudoh, line 126: I now see one thing that slipped my mind... My apologies that I didn't notice thing earlier. We need to detach event listeners, `IntersectionObserver` and `ResizeObserver` when the target DOM nodes are removed. Such use case often happens with "routers" (e.g. React routers) when navigating to different "route" swaps DOM and thus the event listeners, etc. will be orphaned (and causes memory leaks, JavaScript errors, etc).

A quick way to solve this is adding something like below:

```javascript
return {
  release() {
    _resizeObserver.disconnect();
    _innerObserver.disconnect();
    _rootObserver.disconnect();
    window.removeEventListener('load', handleOnLoad);
  },
};
```

Another way (that I'd prefer given scroll into view thing now effectively has a lifecycle) is making it a component rather than a function. We can do it by one of the following:

* Create React component as well as Web Component
* Use Carbon core vanilla API if we still want to make framework neutral: https://github.com/carbon-design-system/carbon/tree/v10.25.0/packages/components/src/globals/js/mixins#component-lifecycle-create-componentjs

CC @jeffchew

jeffchew: For performance purposes a component wrapper might make more sense for this. Perhaps we can build web component wrapper first, and would we be able to create the react wrapper of that or a pure react component would make more sense @asudoh @IgnacioBecerra ?

asudoh: @jeffchew Web Component and a React wrapper on top of it makes sense to me - @IgnacioBecerra what do you think?

IgnacioBecerra: @jeffchew @asudoh Hm, just so I understand, would creating a Web Component for this would be something like:
```
// fade-component.ts
class DDSFadeComponent {
   // scroll-into-view code here + the disconnect thing
}

// dotcom-shell.stories.ts
let elementsToFade = "dds-content-block, dds-content-item, etc..."
let _iterations = true;
.
.
.
return html`
  <dds-fade-component elements="${elementstoFade}" iterations="${_iterations}"/>
  <dds-table-of-contents>
     // etc etc 
`
```

After typing it out like this I think I understand it better, but let me know if my understanding is wrong.

asudoh: Great question @IgnacioBecerra! We have two choices:

1. Wrap each element (or groups of elements) for fade in/out effect with `<dds-fadeinout>`
    * Pro
        * Idiomatic wrt what set of elements need fade in/out effect
        * Strict control of the lifecycle
    * Con:
        * Extra HTML elements
2. Make each scrolling element `<dds-fadeinout>`
    * Pro
        * Less extra HTML elements
    * Con
        * Still need to specify target elements (via CSS selector)
        * Still have possibility of memory leaks (if the target elements have different lifecycle from `<dds-fadeinput>`)

Basically, if we tend to (or can) add fade in/out effect to "region"s or components rather than some indivisual (specific) elements, 1. works the best (e.g. are they only heading/copy elements that need fade in/out effect in content block?). Otherwise 2. works better.

packages/web-components/src/components/fade-in-out/fade-in-out.ts
```diff
@@ -0,0 +1,208 @@
...
+      if (entry.intersectionRatio > 0) {
+        entry.target.classList.remove('bx--fade-out');
+        entry.target.classList.add('bx--fade-in');
+        if (!this._iterations) {
+          this._rootObserver.unobserve(entry.target);
+          this._innerObserver.unobserve(entry.target);
+        }
+      }
+    });
+  }
```
asudoh, line 173: * Make sure private methods are prefixed with `_`.
* Make sure we have type annotations to method args.
* Usage of arrow function avoids the need for `.bind()`.
* Style nit: `<dds-table-of-contents>` uses `records` arg for `IntersectionObserver` callback. I think it's good to align the naming strategy.

```suggestion
  private _handleEntrance = (records: IntersectionObserverEntry[]) => {
    records.forEach({ intersectionRatio, target } => {
      if (intersectionRatio > 0) {
        target.classList.remove('bx--fade-out');
        target.classList.add('bx--fade-in');
        if (!this._iterations) {
          this._rootObserver.unobserve(target);
          this._innerObserver.unobserve(target);
        }
      }
    });
  };
```

asudoh (commented):
Thank you for the update @IgnacioBecerra! Made comments only to Web Components code, but several of them apply to React code, too.

packages/web-components/src/components/fade-in-out/fade-in-out.ts
```diff
@@ -0,0 +1,208 @@
...
+   */
+  @property({ attribute: false })
+  elementList?: string[];
+
+  /**
+   * Iteration boolean for continuous play option.
+   *
+   */
+  @property({ type: Boolean, attribute: 'iterations' })
+  _iterations = false;
```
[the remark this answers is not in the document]
IgnacioBecerra: If `iterations=false` it should unobserve all objects upon effect being displayed. Otherwise, `iterations=true` will keep making the effect happen.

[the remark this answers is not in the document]
asudoh: @IgnacioBecerra No, seems that you didn't mean to be private/internal. In this case, we should remove `_` prefix.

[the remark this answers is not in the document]
IgnacioBecerra: I see, seems like it was leftover from the original `scrollIntoView` code. Will remove.

packages/web-components/src/components/fade-in-out/fade-in-out.ts
```diff
@@ -0,0 +1,208 @@
...
+  };
+
+  /**
+   * The inner viewport calculation for the root margins.
+   *
+   * @private
+   */
+
+  private _getViewportMargin() {
+    return `-${((document.documentElement.clientHeight * this._colSpan) / breakpoints.max.columns).toString()}px 0px`;
```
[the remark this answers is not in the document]
IgnacioBecerra: The math was provided by @photodow, it's to calculate the proper margin so the animation displays at a certain threshold. It should run once at render and update when the screen gets resized within the `ResizeObserver` handler.

packages/web-components/src/components/fade-in-out/fade-in-out.ts
```diff
@@ -0,0 +1,256 @@
...
+ * }
+ *
+ * @element dds-fade-in-out
+ */
+@customElement(`${ddsPrefix}-fade-in-out`)
+class DDSFadeInOut extends StableSelectorMixin(LitElement) {
+  /**
+   * The CSS selector that selects all fade in-out target elements.
+   */
+  @property({ attribute: 'selectorTargets' })
```
asudoh, line 49:
```suggestion
  @property({ attribute: 'selector-targets' })
```

packages/web-components/src/components/dotcom-shell/__stories__/dotcom-shell.stories.ts
```diff
@@ -344,6 +374,99 @@ export const Default = ({ parameters }) => {
...
+            user-status="${ifNonNull(userStatus)}"
+            .legalLinks="${ifNonNull(legalLinks)}"
+            .localeList="${ifNonNull(localeList)}"
+            .footerLinks="${ifNonNull(footerLinks)}"
+            .navLinks="${navLinks}"
+          >
+            ${StoryContent()}
+          </dds-dotcom-shell-container>
+        `}
+    <dds-fade-in-out .selectorTargets="${selectorTargets}" keep-animation="true"></dds-fade-in-out>
```
asudoh, line 419:
```suggestion
    <dds-fade-in-out selector-targets="${selectorTargets}" keep-animation="true"></dds-fade-in-out>
```

packages/web-components/src/components/dotcom-shell/__stories__/dotcom-shell.stories.ts
```diff
@@ -344,6 +374,99 @@ export const Default = ({ parameters }) => {
...
+            user-status="${ifNonNull(userStatus)}"
+            .legalLinks="${ifNonNull(legalLinks)}"
+            .localeList="${ifNonNull(localeList)}"
+            .footerLinks="${ifNonNull(footerLinks)}"
+            .navLinks="${navLinks}"
+          >
+            ${StoryContent()}
+          </dds-dotcom-shell-container>
+        `}
+    <dds-fade-in-out .selectorTargets="${selectorTargets}"></dds-fade-in-out>
```
asudoh, line 466:
```suggestion
    <dds-fade-in-out selector-targets="${selectorTargets}"></dds-fade-in-out>
```

packages/web-components/src/components/fade-in-out/fade-in-out.ts
```diff
@@ -0,0 +1,256 @@
...
+
+  /**
+   * Intersection Observer options
+   *
+   * @private
+   */
+  private _options = {
+    rootMargin: '0px',
+    threshold: 0,
+  };
```
asudoh, line 107: Instead of making `_options` placeholder, it's make `rootMargin` and `threshold` themselves private properties (don't forget adding `_` prefix). In this way, we can track changes separately in future.

packages/web-components/src/components/fade-in-out/fade-in-out.ts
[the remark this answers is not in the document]
asudoh: Per discussion with @IgnacioBecerra offline, we are doing this math (and update new `_viewportMargin` private property) in `ResizeObserver` callback. We are also adding comments to the code to better explain where the math here comes from. Thanks @IgnacioBecerra for your input!

packages/web-components/src/components/fade-in-out/fade-in-out.ts
```diff
@@ -0,0 +1,256 @@
...
+   * @param [options] The options.
+   * @param [options.create] `true` to create the new intersection observer.
+   */
+  private _cleanAndCreateInnerObserver({ create }: { create?: boolean } = {}) {
+    if (this._innerObserver) {
+      this._innerObserver.disconnect();
+      this._innerObserver = null;
+    }
+
+    if (create && this._elements) {
```
asudoh, line 172: Instead of creating `this._elements` in another function, we can run `qSA` with `selectorTarget` here.

packages/web-components/src/components/fade-in-out/fade-in-out.ts
```diff
@@ -0,0 +1,256 @@
...
+  @property({ type: Boolean, attribute: 'keep-animation' })
+  keepAnimation = false;
+
+  /**
+   * Amount of columns used for calculation.
+   *
+   * @private
+   */
+
+  private _colSpan = 3;
```
asudoh, line 68: Let's hoist this outside the class, given no code in this class changes this property.

packages/web-components/src/components/fade-in-out/fade-in-out.ts
```diff
@@ -0,0 +1,256 @@
...
+    this._cleanAndCreateInnerObserver();
+    this._cleanAndCreateResizeObserver();
+
+    if (create) {
+      this._cleanAndCreateRootObserver({ create: true });
+      this._cleanAndCreateResizeObserver({ create: true });
+
+      const { selectorTargets } = this;
+      if (selectorTargets) {
+        forEach(document.querySelectorAll(selectorTargets), item => {
```
asudoh, line 134: The `qSA` should be scoped in the host element of this custom element:

```suggestion
        forEach(this.querySelectorAll(selectorTargets), item => {
```

packages/web-components/src/components/fade-in-out/fade-in-out.ts
```diff
@@ -0,0 +1,256 @@
...
+          this._rootObserver?.observe(item);
+          this._elements.push(item as HTMLElement);
+        });
+      }
+      this._resizeObserver.observe(document.documentElement);
```
asudoh, line 139:
```suggestion
      this._resizeObserver.observe(this.ownerDocument!.documentElement);
```

asudoh (commented):
Thanks @IgnacioBecerra for the updates!

packages/web-components/src/components/fade-in-out/fade-in-out.ts
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  /**
+   * Handler to add recalculated rootMargin to observer.
+   *
+   * @private
+   */
+  private handleResize() {
+    this._rootMargin = this._getViewportMargin();
+    this._cleanAndCreateInnerObserver({ create: true });
```
asudoh, line 190: Can we remove `_getViewportMargin()` private method and send the calculated viewport margin to `._cleanAndCreateInnerObserver()` like `._cleanAndCreateInnerObserver({ create: true, viewportMargin: `-${((this.ownerDocument!.documentElement.clientHeight * _colSpan) / breakpoints.max.columns).toString()}px 0px` })`? Wanted to ensure viewport margin calculation won't happen except in `_handleResize()`.

packages/web-components/src/components/fade-in-out/fade-in-out.ts
```diff
@@ -0,0 +1,256 @@
...
+    this._cleanAndCreateInnerObserver();
+    this._cleanAndCreateResizeObserver();
+
+    if (create) {
+      this._cleanAndCreateRootObserver({ create: true });
+      this._cleanAndCreateResizeObserver({ create: true });
+
+      const { selectorTargets } = this;
+      if (selectorTargets) {
+        forEach(this.ownerDocument!.querySelectorAll(selectorTargets), item => {
```
asudoh, line 116: I may have made a wrong suggestion here, my apologies if it's the case. The `selectorTargets` search should be scoped within `<dds-fade-in-out>`:

```suggestion
        forEach(this.querySelectorAll(selectorTargets), item => {
```

packages/react/src/components/FadeInOut/FadeInOut.js
```diff
@@ -0,0 +1,202 @@
...
+   */
+  useEffect(() => {
+    _rootObserver.current = new IntersectionObserver(handleExit);
+    _resizeObserver.current = new ResizeObserver(handleResize);
+
+    if (selectorTargets) {
+      document.querySelectorAll(selectorTargets).forEach(item => {
+        _rootObserver?.current.observe(item);
+      });
+    }
```
asudoh, line 111: `selectorTargets` search should be scoped in `<FadeInOut>`. Given React component is not automatically associated with a DOM element that's a big hard... One way to achieve this is rendering `<div>` in addition to `children`.

asudoh (commented):
Almost there - Thanks @IgnacioBecerra!

packages/web-components/src/components/fade-in-out/fade-in-out.ts
```diff
@@ -0,0 +1,253 @@
...
+   * grabbed by the animation without restricting their view and perception of
+   * the adopting website. The displayed elements will keep the user's attention
+   * for a longer time as they scroll down the website.
+   *
+   * @private
+   */
+  private _handleResize() {
+    this._cleanAndCreateInnerObserver({
+      create: true,
+      viewportMargin: `-${((document.documentElement.clientHeight * _colSpan) / breakpoints.max.columns).toString()}px 0px`,
```
asudoh, line 179:
```suggestion
      viewportMargin: `-${((this.ownerDocument!.documentElement.clientHeight * _colSpan) / breakpoints.max.columns).toString()}px 0px`,
```

asudoh (commented):

packages/web-components/src/components/fade-in-out/fade-in-out.ts
IgnacioBecerra: ah this totally slipped through me, thanks for noticing!

packages/web-components/src/components/fade-in-out/fade-in-out.ts
```diff
@@ -0,0 +1,256 @@
...
+import StableSelectorMixin from '../../globals/mixins/stable-selector';
+import { forEach } from '../../globals/internal/collection-helpers';
+
+const { stablePrefix: ddsPrefix } = ddsSettings;
+
+/**
+ * Amount of columns used for calculation.
+ */
+
+const _colSpan = 3;
```
asudoh, line 22:
```suggestion
const colSpan = 3;
```

packages/react/src/components/FadeInOut/FadeInOut.js
```diff
@@ -0,0 +1,200 @@
...
+import PropTypes from 'prop-types';
+import settings from 'carbon-components/es/globals/js/settings';
+const { prefix } = settings;
+
+/**
+ * Amount of columns used for calculation.
+ *
+ * @private
+ */
+const _colSpan = 3;
```
asudoh, line 19:
```suggestion
const colSpan = 3;
```

packages/react/src/components/FadeInOut/FadeInOut.js
```diff
@@ -0,0 +1,200 @@
...
+ * }
+ *
+ */
+const FadeInOut = ({ children, selectorTargets, keepAnimations }) => {
+  /**
+   * Outer div component ref for using with query selector.
+   *
+   * @private
+   */
+  const _componentRef = useRef(null);
```
asudoh, line 56:
```suggestion
  const componentRef = useRef(null);
```

packages/react/src/components/FadeInOut/FadeInOut.js
```diff
@@ -0,0 +1,200 @@
...
+
+  /**
+   * Intersection Observer that watches outer viewport.
+   *
+   * @private
+   */
+  const _rootObserver = useRef(null);
```
asudoh, line 63:
```suggestion
  const rootObserver = useRef(null);
```

packages/react/src/components/FadeInOut/FadeInOut.js
```diff
@@ -0,0 +1,200 @@
...
+
+  /**
+   * Intersection observer that watches the inner viewport.
+   *
+   * @private
+   */
+  const _innerObserver = useRef(null);
```
asudoh, line 70:
```suggestion
  const innerObserver = useRef(null);
```

packages/react/src/components/FadeInOut/FadeInOut.js
```diff
@@ -0,0 +1,200 @@
...
+
+  /**
+   * Resize observer to trigger rootMargin recalculations
+   *
+   * @private
+   */
+  const _resizeObserver = useRef(null);
```
asudoh, line 77:
```suggestion
  const resizeObserver = useRef(null);
```

asudoh (commented):
Thanks @IgnacioBecerra! The last naming nit - While we use `_` prefix for private/protected properties/methods, we don't use it for closure variables even if we meant it as private. And please make sure the CI passes. Almost there!

asudoh (approved):
LGTM 👍 - Let's see our designers' input - Thanks @IgnacioBecerra!

jeffchew (changes requested):
Thank you for all of the updates @IgnacioBecerra ! As mentioned in slack, now that `FadeInOut` is its own component now, it makes sense to have it as its own storybook story, along with Docs tab on how adopters can use.

jeffchew:
@IgnacioBecerra looking good so far! For React, I don't think the knobs are necessary since they are related to the dotcom shell and not fade specifically. We can probably remove those for the fade storybook stories.

packages/styles/scss/internal/scroll-into-view/_scroll-into-view.scss
```diff
@@ -0,0 +1,24 @@
+/**
```
jeffchew, line 1: Not sure if this file technically belongs in `internal`, maybe move to `components` since it is part of a component now. Be sure to update corresponding documentation too.

photodow (approved):
Looking good! 👍

jeffchew:
@IgnacioBecerra looks good! before I approve can you check on the CI checks? Probably need to update snapshot files.

IgnacioBecerra:
@jeffchew Yup, ran into a merge conflict with the new image update, just pushed a fix for that. Let's hope the CI passes

jeffchew:
@IgnacioBecerra not critical, but looks like the dotcom shell default in web components has a missing headline now:

Screen Shot 2021-01-14 at 1 51 56 PM

jeffchew (approved):

Issue #4544, opened 2020-11-20
[Masthead]: React - gradient appearing on first nav item on ipad and desktop safari of older version


annawen1:
## Detailed description

> Describe in detail the issue you're having.

On iOS 13 Ipad and Version 13.0.5 of Safari desktop, there is a vertical gradient that appears on top of the first nav item.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

> Is this issue related to a specific component?

The gradient seems to be coming from the overflow left caret `::after` pseudo element styles. It seems that even with the left caret button hidden, the pseudo styles still appear.

If user clicks on nav item, the gradient disappears, but will reappear if user clicks on right overflow caret and then left overflow caret

<img width="512" alt="Screen Shot 2020-11-20 at 10 10 39 AM" src="https://user-images.githubusercontent.com/54281166/99826500-71077d80-2b26-11eb-9a2c-1f9993ceed5d.png">

PR #4619, opened 2020-12-03
fix(Masthead): remove pseudo elements and replace with div elements for overflow gradient

annawen1:
### Related Ticket(s)

[Masthead]: React - gradient appearing on first nav item on ipad and desktop safari of older version #4544

### Description

Problem with the gradient styling applied to the caret pseudo elements not disappearing seems to be an issue with Safari version 13 and ios 13. Workaround is the remove the pseudo elements and use empty div element to apply the gradient styles instead. 

BEFORE:
Screen Shot 2020-12-03 at 11 59 42 AM

AFTER:
Screen Shot 2020-12-03 at 11 59 17 AM

### Changelog

**New**

- Add wrapper parent container div around the caret buttons and gradient div
- new empty div will handle gradient styling instead of the button's pseudo elements


jeffchew (approved):
LGTM!

Issue #4369, opened 2020-10-30
Web Components | Add Stable Selectors to Video Player


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

Stable selectors are missing for `dds-video-player`

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

`dds-video-player`

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Full documentation on stable selectors, and add stable selectors for any applicable sub components.

asudoh:
React `<VideoPlayer>` has one `data-autoid` in its inner content `<div class="bx--video-player__video-container">`. Web Components versions renders it in Shadow DOM and thus analytics, etc. cannot access the element via `data-autoid`. So the suggestion is simply using `<dds-video-player>` tag to find the video player.

PR #4631, opened 2020-12-07
docs(video-player): add documentation of stable selector

asudoh:
### Related Ticket(s)

Refs #4369.

### Description

Adds the documentation of stable selector for our video player component. React `<VideoPlayer>` has one `data-autoid` in its inner content `<div class="bx--video-player__video-container">`. Web Components versions renders the corresponding element in Shadow DOM and thus analytics, etc. cannot access the element via `data-autoid`. So the instruction is using `<dds-video-player>` tag to find the video player.

### Changelog

**New**

- The documentation of stable selector for our video player component


jeffchew (approved):
Thank you @asudoh !

Issue #4366, opened 2020-10-30
Web Components | Add Stable Selectors to Link List


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

Stable selectors are missing for `dds-link-list`

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

`dds-link-list`

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Full documentation on stable selectors, and add stable selectors for any applicable sub components.

PR #4632, opened 2020-12-07
fix(link-list): add missing mixin

asudoh:
### Related Ticket(s)

Refs #4366.

### Description

Ensures `<dds-link-list>` extends `StableSelectorMixin` so the static property that defines that stable selector works.

### Changelog

**Changed**

- A change to `<dds-link-list>` to ensure that it extends `StableSelectorMixin`.


RaphaelAmadeu (approved):
Looks good to me!

Issue #4357, opened 2020-10-30
Web Components | Add Stable Selectors to Card


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

Stable selectors and prop table are missing for `dds-card`

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

`dds-card`

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Full documentation on stable selectors.

PR #4633, opened 2020-12-07
fix(card): add missing mixin

asudoh:
### Related Ticket(s)

Refs #4357.
Refs #4361.

### Description

Ensures that `<dds-card>` extends `StableSelectorMixin` so the static property that defines the stable selector works.

### Changelog

**Changed**

- A change to ensure that `<dds-card>` extends `StableSelectorMixin`.


jeffchew (approved):
Good catch!

PR #4634, opened 2020-12-07
chore(feature-card): use theme zoning

asudoh:
### Description

Applies theme zoning to feature card instead of inverse variant of card style. This allows us to make feature card style simpler, especially for Web Components.

### Changelog

**Changed**

- A change to apply theme zoning to feature card instead of inverse variant of card style.


jeffchew (approved):
Thanks @asudoh !

Issue #4112, opened 2020-10-01
Consider changing cta slots to footer


asudoh:
## Detailed description

Some components in Web Components codebase use `<slot name="cta">` given their React counterparts use `cta` property, but they effectively are the footer, and some other components in Web Components codebase use `<slot name="footer">`. This issue is for considering aligning them to `<slot name="footer">`.

Issue #4079, opened 2020-09-29
Carbon for ibm.com Jan 2021 Release Plan (v1.15.0)


RobertaJHahn:
# Timeline

- **January 15, 2021**: Code freeze
- **January 18 - 25, 2021**: Release Candidate(s) and testing
- **January 27, 2021**: Full Release

## Global
- **Upgrade Carbon to v10.26.0** (#4857)

### Announcement

With the release of `Carbon for IBM.com v1.15.0`, our team will be graduating our web components package to `v1.0.0`! This release is filled with final arrangements, feature parity with React, various fixes, as well as numerous API reconciliation to ensure that we can provide stable support for the future of this package. We thank all of our early adopters of web components, and look forward to providing continued support for your application needs!

### Breaking Changes

* The slot names for multiple components have changes, namely the reconciliation of the naming of the `footer` slots, as well as `action` slots. Be sure to check updated documentation when updating to this version release of web components. PR reference: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/4635
* The `mark-type` attribute of `<dds-quote>` now takes dash-cased values (e.g. `double-curved`) instead of camel-cased values (`doubleCurved`). PR reference: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/4658
* The shadow slot names of `<dds-quote>` are now dash-cased (e.g. `source-heading`) instead of camel-cased (e.g. `sourceHeading`). PR reference: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/4756
* Card slots `eyebrow` and `heading` are now custom elements `dds-card-eyebrow` and `dds-card-heading` respectively. PR reference: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/4783
* `copy` slot in `<dds-quote>` now is the default slot. Also other slots in `<dds-quote>` no longer has styling elements around it for better flexibility. The default styles are implemented by `<dds-quote-source-copy>`/`<dds-quote-source-heading>`/`<dds-quote-source-bottom-heading>`. PR reference: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/4840
[...]

PR #4635, opened 2020-12-07
chore(components): unify slot names

asudoh:
### Related Ticket(s)

Refs #4112.

### Description

Changes all `<slot name="cta">` occurrences to `<slot name="footer">` to neutralize the purpose of the `<slot>`.

The only exception is one in `<dds-cta-section>` where there is a CTA in the middle and one in the footer. Used `<slot name="action">` for the former.

### Changelog

**Changed**

- All `<slot name="cta">` occurrences to `<slot name="footer">` to neutralize the purpose of the `<slot>`.


jeffchew (approved):
Thanks @asudoh !

jeffchew:
BTW we might want to make a note in our release notes in the next release about the slot name changes since it's an API change.

asudoh:
👍 Makes sense @jeffchew!

jeffchew:
Added a note to our pinned issue: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/4079

PR #4647, opened 2020-12-07
fix(locale-modal): replaced href target to avoid storybook refresh

IgnacioBecerra:
### Description

Fixed the bug where Storybook refreshes entire page upon clicking the `LinkWithIcon` component to go back to the Selection geographic area menu.

It was caused by using `href='#`, which made the page redirect to the `iframe` address url over Storybook's. With this change, the component will stop the `href` from activating and fall back to the original functionality.

### Changelog

**New**
-- added `e.preventDefault()` to the event handlers

### DIFF
--- a/packages/react/src/components/LocaleModal/LocaleModal.js
+++ b/packages/react/src/components/LocaleModal/LocaleModal.js
@@ -114,7 +114,7 @@ const LocaleModal = ({ isOpen, setIsOpen, localeData, localeDisplay }) => {
         <ModalHeader
           data-autoid={`${stablePrefix}--locale-modal__region-back`}
           label={[
-            <LinkWithIcon href="javascript:void(null);" iconPlacement={'left'}>
+            <LinkWithIcon href="#" iconPlacement={'left'}>
               <span>{modalLabels.headerTitle}</span>
               {
                 <ArrowLeft20
--- a/packages/react/src/components/LocaleModal/LocaleModal.js
+++ b/packages/react/src/components/LocaleModal/LocaleModal.js
@@ -114,7 +114,7 @@ const LocaleModal = ({ isOpen, setIsOpen, localeData, localeDisplay }) => {
         <ModalHeader
           data-autoid={`${stablePrefix}--locale-modal__region-back`}
           label={[
-            <LinkWithIcon href="#" iconPlacement={'left'}>
+            <LinkWithIcon href="javascript:void(null);" iconPlacement={'left'}>
               <span>{modalLabels.headerTitle}</span>
               {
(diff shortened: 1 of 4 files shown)


packages/react/src/components/LocaleModal/LocaleModal.js
```diff
@@ -114,7 +114,7 @@ const LocaleModal = ({ isOpen, setIsOpen, localeData, localeDisplay }) => {
         <ModalHeader
           data-autoid={`${stablePrefix}--locale-modal__region-back`}
           label={[
-            <LinkWithIcon href="#" iconPlacement={'left'}>
+            <LinkWithIcon href="javascript:void(null);" iconPlacement={'left'}>
```
annawen1, line 117: Hey @IgnacioBecerra I think React throws console errors for `javascript:void(null)`

> Warning: A future version of React will block javascript: URLs as a security precaution. Use event handlers instead if you can. If you need to generate unsafe HTML try using dangerouslySetInnerHTML instead. React was passed "javascript:;".

You might want to switch to event handler instead for the link and use `e.preventDefault()`?

kennylam: > You might want to switch to event handler instead for the link and use `e.preventDefault()`?

+1 we favor `preventDefault()` over inline javascript.
Looks like adding it to these to listeners will take care of it.
https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/5eb2353f5e8bed96e3f7fa7acc3b77ba5901353e/packages/react/src/components/LocaleModal/LocaleModalRegions.js#L153-L162

jeffchew (commented):
I think this was one of the items we wanted to get into the release. Can you change the target branch to the release branch? Otherwise this looks good!

packages/web-components/src/components/link-list/link-list.ts
```diff
@@ -10,6 +10,7 @@ import { customElement, html, LitElement, property } from 'lit-element';
 import settings from 'carbon-components/es/globals/js/settings';
 import ddsSettings from '@carbon/ibmdotcom-utilities/es/utilities/settings/settings';
 import sameHeight from '@carbon/ibmdotcom-utilities/es/utilities/sameHeight/sameHeight';
+import StableSelectorMixin from '../../globals/mixins/stable-selector';
```
jeffchew, line 13: @IgnacioBecerra I think you cut from the master branch for this change, you might need to cut a branch off the release branch instead, because we don't want to have changes from the master branch for the release.

jeffchew (approved):
Thank you @IgnacioBecerra !

Issue #4535, opened 2020-11-19
React layout sticky nav overlap


andysherman2121:
## Detailed description

On medium breakpoint the sticky nav is overlapping with the rest of the text on the page when scrolling.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

Layout

> What did you expect to happen? What happened instead? What would you like to
> see changed?

> What browser are you working in?

Chrome

## Steps to reproduce the issue

1. Go to the Layout component and change the screen size to medium breakpoint
2. You will see the sticky nav overlapping with the other columns

## Additional information

![Screen Shot 2020-11-19 at 4 41 42 PM](https://user-images.githubusercontent.com/56688604/99727564-2e8e6400-2a86-11eb-9c85-0ff9256161ec.png)

kennylam:
@andysherman2121 I can't reproduce this on any of the environments. Where were you seeing this?

andysherman2121:
@kennylam I just reproduced it on canary. If you go to the layout react story, and then make the screen go to medium breakpoint. Once you start scrolling, the sticky column will overlap with the rest of the page

PR #4649, opened 2020-12-07
style(layout): setting sticky property only for lg and above breakpoints

guilhermelMoraes:
### Description

The `position: sticky` property set at the 'md' and below was causing the columns to overlap each other.


jeffchew (approved):
Thanks @guilhermelMoraes !

kennylam:
Fixes #4535

Issue #4215, opened 2020-10-13
React: Masthead -  search open by default - when we tries to activate the user profile button the button is not operable


Praveen-sr:
## Detailed description

Issue: When the search popup is displayed,  when we tries to activate the user profile button the button is not operable. instead the search popup is getting closed when we tries to activate it. Note: user profile button is not operable only when the search popup is  displayed.

Expected : User profile button should be operable.

## Additional information

![Masthead2.gif](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/0e6189ae-7210-4ead-9d50-670e2672d9f6)

RobertaJHahn:
The only team that has search open by default is the IBM Search team.  Their configuration does not have the profile icon active.  No teams are using this configuration option currently.

ljcarot:
@RobertaJHahn Let's move to Jan release.

PR #4652, opened 2020-12-07
fix(masthead): allow search to remain open when profile button is clicked

IgnacioBecerra:
### Related Ticket(s)
#4215

### Description

The previous code only allowed clicks within the `MastheadSearch` component. With this update it now allows clicks anywhere on the `Masthead` component (which includes `MastheadProfile`) and have the search box remain open.

### Changelog

**Changed**

- conditional to allow `MastheadSearch` to be open if user clicks anywhere within parent component


packages/react/src/components/Masthead/MastheadSearch.js
```diff
@@ -169,12 +169,14 @@ const MastheadSearch = ({
...
-     * elements other than the close button are clicked.
+     * elements other than the close button and the
+     * profile button are clicked.
      *
-     * @param {*} event Click event outside search component
+     * @param {*} event Click event outside masthead component
      */
     const handleClickOutside = event => {
-      if (ref.current && !ref.current.contains(event.target)) {
+      let mastheadRef = ref.current.parentNode.parentNode;
```
asudoh, line 178: Would you try `ref.current?.closest('.some--css-class')` to avoid reliance to the specific DOM structure?

asudoh (commented):

packages/react/src/components/Masthead/MastheadSearch.js
```diff
@@ -169,12 +169,14 @@ const MastheadSearch = ({
...
+     * elements other than the close button and the
+     * profile button are clicked.
      *
-     * @param {*} event Click event outside search component
+     * @param {*} event Click event outside masthead component
      */
     const handleClickOutside = event => {
-      if (ref.current && !ref.current.contains(event.target)) {
+      let mastheadRef = ref.current?.closest('.bx--masthead');
+      if (ref.current && !mastheadRef.contains(event.target)) {
```
asudoh, line 179:
```suggestion
      if (mastheadRef && !mastheadRef.contains(event.target)) {
```

asudoh (commented):
Just one other thing - Thanks @IgnacioBecerra!

asudoh (approved):

Issue #4375, opened 2020-11-01
Button group: Guide how to make all items have equal width/height


asudoh:
### The problem

The React version of button group, etc. makes each buttons have same width/height. OTOH it has a performance concern with doing it out of the box.

### The solution

Determine what to do with our Web Components library. The minimum requirement is documenting how to do it in adaptor side.

PR #4657, opened 2020-12-08
fix(card-group): same height by CSS grid

asudoh:
### Related Ticket(s)

Refs #4375.

### Description

Switches flex-box-bassed Carbon grid layout system to native CSS grid so we can apply same height to all cards upon card contents.

### Changelog

**Changed**

- Switches flex-box-bassed Carbon grid layout for `<dds-card-group>` to native CSS grid.


kennylam (approved):
LGTM!

Issue #4160, opened 2020-10-07
Quote: Consider making enums dash-cased


asudoh:
### The problem

`<dds-quote>` code has a property that takes an enum. The values of such enums currently is camel-cased, where as most of enums in the codebase are dash-cased.

### The solution

See if camel-casing those enum values are absolutely necessary (e.g. for styling purpose). Otherwise change them to dash-cased.

PR #4658, opened 2020-12-08
chore(quote): make enums dash-cased

asudoh:
### Related Ticket(s)

Refs #4160.

### Description

Changes the enums for quote choices dash-cased to align it to how the corresponding attribute is named.

### Changelog

**Changed**

- The enums for quote choices to dash-cased.


kennylam (approved):
LGTM!

PR #4659, opened 2020-12-08
chore(link-list): use CSS for same height

asudoh:
### Related Ticket(s)

Refs #4375.

### Description

Switches the "same height" feature to one based on CSS grid for better performance and fix FOUC.

### Changelog

**Changed**

- Switches the "same height" feature to one based on CSS grid.


kennylam (approved):
Looks good thanks @asudoh!

PR #4662, opened 2020-12-08
fix(card-group): removes margin adjustment

asudoh:
### Description

Removes margin adjustments from some components that uses `<dds-card-group>` inside.

`<dds-card-group>` removed dependency to Carbon grid in #4657, and such (removed) dependency adjusted margin set by Carbon grid.

Also adjusts some stories.

### Changelog

**Changed**

- Adjusts some stories, especially one of `<dds-content-group-cards>` fixing the wrong icon position.

**Removed**

- Margin adjustments from some components that uses `<dds-card-group>` inside.


andysherman2121 (approved):
Looks good to me!

Issue #4667, opened 2020-12-08
Web component: Footer - Provide label as "Select geographic area" button which triggers the modal 


Praveen-sr:
## Detailed description
On JAWS FF

Issue:
The name of the currently selected option is given as the label for this button instead of an appropriate label. i.e. "United States — English"

Expected:
provide label as Select geographic area 

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/7972d4cb-156a-47d0-93da-45ef18a88110)

PR #4669, opened 2020-12-08
fix(footer): added label to locale button for better description

IgnacioBecerra:
### Related Ticket(s)
#4667

### Description

Added `aria-label` to the locale button in order to provide an accessible description of its purpose in Web Components.

### Changelog

**New**

- added `aria-label="Select geographic area"` to `locale-button`


jeffchew (commented):
thanks @IgnacioBecerra !

We might want to have that as a property with default value for english, in case any teams want to customize it to other languages.

jeffchew (approved):
Thank you, looks good to me!

Issue #4668, opened 2020-12-08
Web component: Locale selector - list of options (locations and languages) are not identified by the screen reader. Only a "List box" is identified


Praveen-sr:
## Detailed description

Issue:
The list of options are not identified by the screen reader. Only a "List box" is identified instead. Here within the list box, when we navigate with up/down arrow keys, Screen reader does not announce anything. The options are announced only in tab/shift tab navigation. This behavior would be confusing. Hence the user would not be able to interact with the elements efficiently.

Expected:
Remove the role "list box" from the options.

PR #4670, opened 2020-12-08
fix(locale-modal): remove listbox role for accessibility

annawen1:
### Related Ticket(s)

Web component: Locale selector - list of options (locations and languages) are not identified by the screen reader. Only a "List box" is identified #4668

### Description

Remove the "listbox" role from the locale modal options as screen reader user will think it is a dropdown and try to navigate using up/down arrow keys.

### Changelog

**Removed**

- remove "listbox" role from locale modal language options list


jeffchew (approved):
Thanks @annawen1 !

jeffchew:
@annawen1 can you check the web components ci-checks? Probably snapshots out of date.

Issue #4056, opened 2020-09-25
Web component: Card & Card group - Visual focus is not covering the entire card. Focus is only covering lower half of the card.


Praveen-sr:
## Detailed description

Issue: On card and Card group - Visual focus is not covering the entire card. Focus is only covering lower half of the card.

Expected - Focus should cover the entire element/card.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/9a9d3b8a-8e99-434c-a993-210b4b94f82a)

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/b3ebd294-0644-4791-a747-70a15567e686)

PR #4671, opened 2020-12-08
fix(feature-card): removed bugged double focus

IgnacioBecerra:
### Related Ticket(s)
#4056

### Description

When focusing on `dds-feature-card`, both the component itself and the `a` element within were selected, causing double focus states.

Before:
Screen Shot 2020-12-08 at 9 12 35 AM

After:
Screen Shot 2020-12-08 at 9 12 46 AM

### Changelog

**Changed**
-- changed focus outline for the component so it only comes from the `a` element itself


packages/web-components/src/components/feature-card/feature-card.scss
```diff
@@ -14,6 +14,10 @@
 :host(#{$dds-prefix}-feature-card),
 :host(#{$dds-prefix}-feature-cta) {
   display: block;
+
+  &:focus {
+    outline: none;
+  }
```
asudoh, line 20:
```suggestion
  outline: none;
```

asudoh (commented):
Thanks @IgnacioBecerra for jumping in!

asudoh (approved):

Issue #4251, opened 2020-10-19
Components: Ensure forward-compatible with React wrappers in Node 14


asudoh:
### The problem

#3475 tried to introduce automatic React wrapper, but abandoned because SSR with it doesn't work in Node 12 environment.

### The solution

Test it again with Node 14 and ensure our component code is forward-compatible for such React wrapper.

PR #4678, opened 2020-12-09
feat(web-components): add React wrapper

asudoh:
### Related Ticket(s)

Refs #4251.

### Description

Adds React wrapper generator (`createReactCustomElementType`), based on one in `carbon-web-components` to `@carbon/ibmdotcom-web-components` codebase.

It helps our adaptors using React to use our components, with the following support that React itself lacks for non-intrinsic elements:

* Element attributes mapping to non-string element property values (e.g. Boolean attributes)
* React props mapping to element properties instead of element attributes
* Event handlers

Also adds a Babel plugin that automatically creates React wrappers of our components from our components' `@property()` annotations, as well as the corresponding build process that generates the React wrappers to `es/components-react` directory. So in the most typical scenario, our users can use components in `es/components-react` directory in the same
manner as regular React components.

Also adds React stories so we can demonstrate the React integration.

### Changelog

**New**

- React wrapper generator (`createReactCustomElementType`).
- Build logics to generate React wrapper, one for running in browser and one for running in Node.

**Changed**

- Moves enums to `defs.ts` to make above build steps simpler.


kennylam (approved):
Looks good thanks @asudoh!

jeffchew:
Thank you @asudoh ! Can you resolve the conflicts?

asudoh:
Thanks @kennylam for reviewing! @jeffchew Merged.

Issue #4675, opened 2020-12-08
Web Components | Locale Modal | When a region has no available country/language, should not be clickable


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

When loading the locale modal in Web Components, if one region has no available country/languages, the option should not be clickable:

![language-unavailable](https://user-images.githubusercontent.com/1641214/101555295-5f2b3480-3986-11eb-8a8d-0b3bb3d3cffb.gif)

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

bug

> Is this issue related to a specific component?

locale modal

> What did you expect to happen? What happened instead? What would you like to
> see changed?

The region should be grayed out with an unavailable icon, and non-clickable

> What browser are you working in?

Chrome

> What version of Carbon for IBM.com are you using?

Latest canary

## Steps to reproduce the issue

1. Spin up dotcom-shell and reduce `<link rel="alternate">` values to not include a region
2. Open the locale modal
3. Click the region with no available options

PR #4682, opened 2020-12-09
fix(locale-modal): disable invalid region

asudoh:
### Related Ticket(s)

Refs #4675.

### Changelog

**Changed**

- A change to invalid the card for invalid region (region with no associated countries).


packages/web-components/tests/snapshots/dds-region-item.md
```diff
@@ -28,23 +28,24 @@
...
-      </div>
+</p>
+<div class="bx--card__wrapper">
+  <div class="bx--card__content">
+    <h3 class="bx--card__heading">
+      <slot>
+        name-foo
+      </slot>
+    </h3>
+    <div class="bx--card__footer">
```
jeffchew, line 43: @asudoh thank you for the quick turnaround on this! Is there a way to visually see this? I see that the snapshot isn't wrapping the card with the `<p>` tag, wondering what that looks like...

asudoh: @jeffchew I somehow edited https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.12.0/packages/web-components/src/components/locale-modal/locale-modal-composite.ts#L151 to replace `countryList.length === 0` with `true`, where the country list comes from totally slipped from my mind... My bad.

asudoh: The snapshot looks strange... There are definitely not two `<p>` tags rendered...

asudoh: Actually it was a good catch, found that `<p>` cannot contain `<div>` according to "Permitted content" section at: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/p

Pushed up a fix. Thank you for bringing this up @jeffchew!

jeffchew (approved):
Much better, thank you @asudoh !

Issue #4677, opened 2020-12-09
Web component: Masthead - "Undefined menubar" Between IBM logo link and "Open IBM search field" button.


Praveen-sr:
## Detailed description

Issue:
Screen reader is identifying an menu as "Undefined menubar" Between IBM logo link and "Open IBM search field" button. It is not visible and is available only in Up/Down Arrow key navigation. When we try to activate it with "enter" or "spacebar", no action takes place.

Expected:
If the user is not expected to interact with this element, this should be hidden from the Screen reader user as well.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/48426eb9-a4b0-4fbf-bf42-0c0078a7ee8f)

PR #4690, opened 2020-12-10
fix(masthead): avoid rendering L0 if L1 is there

asudoh:
### Related Ticket(s)

Refs #4677.

### Changelog

**Changed**

- A change to avoid rendering L0 if L1 is there.


jeffchew (approved):
LGTM!

Issue #4684, opened 2020-12-09
Web Components | footer and dotcom-shell are missing property to disable the locale button


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

React currently has the prop `disableLocaleButton`, which is an option to suppress the rendering of the locale button in the footer. This is available for `Footer` and `DotcomShell`. However, this isn't available in web components. This should be added.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug (need to be feature parity with React)

> Is this issue related to a specific component?

Footer and Dotcom Shell

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Property available to disable the locale button

> What browser are you working in?

All browsers

> What version of Carbon for IBM.com are you using?

Latest

## Steps to reproduce the issue

1. Compare http://ibmdotcom-react.mybluemix.net/?path=/docs/components-footer--default to http://ibmdotcom-web-components.mybluemix.net/?path=/docs/components-footer--default and notice that `disableLocaleButton` is missing in web components

PR #4691, opened 2020-12-10
feat(footer): support forcing omitting locale button

asudoh:
### Related Ticket(s)

Refs #4684.

### Changelog

**New**

- `disable-locale-button` attribute to `<dds-footer-composite>` and `<dds-dotcom-shell-composite>`.


jeffchew (approved):
thank you @asudoh !

Minor thing, I don't see the prop table appearing for the dotcom shell Docs tab. We can probably resolve that in a different PR though if we can't tackle that here.

asudoh:
Makes sense @jeffchew - Updated.

PR #4716, opened 2020-12-11
fix(cta-section): same height by CSS for copy text

IgnacioBecerra:
### Related Ticket(s)
#4472
#4474

### Description

Different copy text caused uneven height between the two components. This PR uses CSS `grid` feature to make both elements be the same height without any complicated or time-consuming code.

### Changelog

**New**

- added `display: grid` for `cta-section-item-paragraph` to match sibling element's height

**Removed**

- unused styles for `slotted(p)` within `cta-section-item-copy`


asudoh (approved):

PR #4717, opened 2020-12-11
feat(masthead): add horizontal scroller

asudoh:
### Related Ticket(s)

Refs #4457.

### Description

TBD

### Changelog

**New**

- The horizontal scroller in `<dds-top-nav>`.

**Changed**

- The breakpoint of left/top nav switch (from `md` to `800px`).


kennylam (commented):
@asudoh Small thing it looks like the L0 is appearing behind the profile button because it has no background color set. Giving `dds-masthead-profile` a background color of white should fix.

asudoh:
Good catch @kennylam! - Updated.

jeffchew (approved):
Thank you @asudoh , looks good!

PR #4718, opened 2020-12-11
chore(button-group): same height effect by CSS grid

asudoh:
### Related Ticket(s)

Refs #4375.

### Description

Uses CSS grid for the same height effect for better performance and fix FOUC.

### Changelog

**Changed**

- Uses CSS grid for the same height effect for better performance and fix FOUC.


jeffchew:
Thank you for the change @asudoh !

I think buttons are supposed to be full width in mobile breakpoints:

Screen Shot 2020-12-11 at 2 24 17 PM

asudoh:
Thanks @jeffchew (and @kennylam as well) for catching this - Fixed.

jeffchew (approved):
thank you @asudoh !

PR #4720, opened 2020-12-11
docs(cta): update documentation link to proper dependency

photodow:
### Description

Noticed this was pointing to `Card` when the code was using `CTA`. Figured I would update it to help out. Without it, adopters can't find the types allowed to pass in.

> Side question: Is there a reason Card is still a dependency in here, and we're not using CTA for that?

### Changelog

**Changed**

- Card to CTA link reference.


jeffchew (approved):
Thanks for the update @photodow !

Issue #4688, opened 2020-12-09
[leaving-ibm modal] missing 20% padding-right


photodow:
Looks like the modal in leaving ibm is missing 20% padding-right.

———
Here is what I'm seeing...

<img src="https://user-images.githubusercontent.com/3793636/101693828-cc4ad280-3a37-11eb-9b61-febe9c90c716.png" width="400px" />

http://ibmdotcom-web-components.mybluemix.net/iframe.html?id=components-leaving-ibm--default

———
Here it is over at Carbon...

<img src="https://user-images.githubusercontent.com/3793636/101693424-34e57f80-3a37-11eb-97c2-42a8fda86e22.png" width="400px" />

https://www.carbondesignsystem.com/components/modal/code

## Additional information

- Leaving ibm [[box](https://ibm.box.com/s/g79wdwdj0syjcfkdlgu5rpwe3zp9s37w)]

PR #4724, opened 2020-12-14
fix(leaving-ibm): add 20% padding right to modal

annawen1:
### Related Ticket(s)

[leaving-ibm modal] missing 20% padding-right #4688

### Description

Add padding-right: 20% to modal body to match specs

Screen Shot 2020-12-14 at 3 12 06 PM

### Changelog

**New**

- match expressive carbon modal by adding `padding-right: 20%` to modal body


asudoh (approved):
LGTM basically - Thanks @annawen1! Just FYI, (assuming leaving IBM modal is small size variant) Carbon core uses `$spacing-05` for breakpoints smaller than `xlg`: https://github.com/carbon-design-system/carbon/blob/v10.25.0/packages/components/src/components/modal/_modal.scss#L155-L189 - Just in case, CC @wonilsuhibm

Issue #3524, opened 2020-08-11
Visual Regression | Carbon Expressive | UI Shell Sidenav item heights are smaller


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

After upgrading to Carbon v10.17.0, the UI Shell sidenav items appear to be shorter in height.

<img width="1402" alt="Screen Shot 2020-08-11 at 5 38 44 PM" src="https://user-images.githubusercontent.com/1641214/89951606-ba783080-dbf9-11ea-9480-fc29938a713b.png">

Percy build:
https://percy.io/538fc19a/Carbon-Expressive-Theme/builds/6460290

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

bug

> Is this issue related to a specific component?

UI Shell sidenav

Creatoon:
@jeffchew  hello sir , i want to work on this bug , could you please guide me how can i reach that?

RobertaJHahn:
@jeffchew How do I know if this volunteer ever finished this work?

PR #4735, opened 2020-12-16
fix(ui-shell): adjust ui shell for expressive from carbon update

annawen1:
### Related Ticket(s)

Visual Regression | Carbon Expressive | UI Shell Sidenav item heights are smaller #3524

### Description

Height for UI Shell sidenav changed when upgrading carbon, this pr adjusts the heights back to the expected expressive height.


asudoh (approved):
LGTM 👍 - Thanks @annawen1!

PR #4749, opened 2020-12-18
feat(link): added :part pseudo-element to component

guilhermelMoraes:
### Description

Adding a `::part` pseudo-element to the `feature card`, so that we can update it at the application level. 

### Changelog

**New**

- [::part pseudo-element](https://developer.mozilla.org/en-US/docs/Web/CSS/::part)


asudoh (approved):
LGTM 👍 - Thanks @guilhermelMoraes for doing this!

PR #4759, opened 2020-12-21
style(feature-card): fixing img aspect ratio

guilhermelMoraes:
### Description

There are some cases where the `feature-card` image aspect ratio mixin is bugging the image height. Define an explicit `height: 100%` would solve the problem.

Screen Shot 2020-12-21 at 10 42 03

Screen Shot 2020-12-21 at 10 45 00

### Changelog

**Changed**

- Explicit `height: 100%` to `feature-card` image.


asudoh (approved):

PR #4752, opened 2020-12-18
feat(leadspace): added ::part pseudo-element to component

guilhermelMoraes:
### Description

Added `::part` pseudo-element to `leadspace`

### Changelog

**New**

- `::part` pseudo element in `leadspace.ts`

PR #4763, opened 2020-12-21
style(leadspace): added explicit height declaration for sections

guilhermelMoraes:
### Description

After the `::part` pseudo-element introduction (#4752) to change the leadspace height according to the Partner world landing page requirements, I've missed adding an explicit `height: 100%` declaration for the wrappers below it. This is causing the leadspace to be visually broken for the `md` and `sm` breakpoints.


asudoh (approved):

Issue #4768, opened 2020-12-22
Web Components: Missing type definition in the package


asudoh:
## Detailed description

> Describe in detail the issue you're having.

An adopter reported that they saw the following compile error in their project:

```
ERROR in [at-loader] ./node_modules/@carbon/ibmdotcom-web-components/es/components/video-player/video-player-container.d.ts:51:63
    TS2304: Cannot find name 'Constructor'.
ℹ ｢wdm｣: Failed to compile.
```

That revealed that `@carbon/ibmdotcom-web-components` package doesn't have the definition of [`Constructor`](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.13.0/packages/web-components/src/typings/constructor.d.ts#L13), though there are several modules in the package that refers to `Constructor`.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

A bug.

> Is this issue related to a specific component?

No, it spreads across many mix-in definitions in our Web Components package.

> What did you expect to happen? What happened instead? What would you like to
> see changed?

The definition of [`Constructor`](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.13.0/packages/web-components/src/typings/constructor.d.ts#L13) is in `@carbon/ibmdotcom-web-components` package.

> What browser are you working in?

N/A

> What version of Carbon for IBM.com are you using?

`v1.13.0`

PR #4769, opened 2020-12-22
fix(web-components): add `Constructor` to the package

asudoh:
### Related Ticket(s)

Refs #4768.

### Description

Attempts to fix `TS2304: Cannot find name 'Constructor'` error reported by an adopter by making sure the definition of `Constructor` is in `@carbon/ibmdotcom-web-components` package.

### Changelog

**Changed**

- Moved the definition of [`Constructor`](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.13.0/packages/web-components/src/typings/constructor.d.ts#L13) from a file of internal type definition to a file that is part of `@carbon/ibmdotcom-web-components` package.


jeffchew (approved):
Thanks @asudoh !

PR #4779, opened 2020-12-24
feat(cta): introduce button variant

asudoh:
### Related Ticket(s)

Refs #3556.

### Description

Introudces button variant of the CTA, that inherits `<dds-button-group-item>`.

`<dds-btn>`, the parent class of `<dds-button-group-item>`, is written from scratch (instead of inheriting `<bx-btn>`) so it opens up the APIs that CTA requires.

### Changelog

**New**

- `<dds-button-cta>`, a button variant of CTA, inheriting `<dds-button-group-item>`.


guilhermelMoraes (approved):
LGTM! Thanks for the help, @asudoh! 👍🏻

PR #4782, opened 2020-12-28
fix(card): added grid to story

RaphaelAmadeu:
### Description

Added grid column sizes to `Card` story.

### DIFF
--- a/packages/web-components/src/components/card/__stories__/card.stories.ts
+++ b/packages/web-components/src/components/card/__stories__/card.stories.ts
@@ -98,7 +98,11 @@ export default {
   decorators: [
     story => html`
       <div class="dds-ce-demo-devenv--simple-grid dds-ce-demo-devenv--simple-grid--card">
-        ${story()}
+        <div class="bx--row dds-ce-demo-devenv--grid-row">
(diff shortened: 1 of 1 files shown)


jeffchew (approved):
Thank you @RaphaelAmadeu !

packages/web-components/src/components/card/__stories__/card.stories.ts
```diff
@@ -98,7 +98,11 @@ export default {
   decorators: [
     story => html`
       <div class="dds-ce-demo-devenv--simple-grid dds-ce-demo-devenv--simple-grid--card">
-        ${story()}
+        <div class="bx--row dds-ce-demo-devenv--grid-row">
+          <div class="bx--col-sm-4 bx--col-md-3 bx--col-lg-12 bx--col-xlg-8 bx--no-gutter">
+            ${story()}
+          </div>
+        </div>
```
asudoh, line 105: Hi @RaphaelAmadeu thank you for the adjustment; Would you be able to try doing this adjustment using `grid-column`, like: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/072e68f/packages/web-components/.storybook/container.scss#L77-L101

Issue #4622, opened 2020-12-03
Carbon v10.25.0 style bug: checkbox


wonilsuhibm:
## Detailed description
This bug is scheduled in the Jan 2021 release.

It seems that some of our expressive components are having positioning and spacing issues after applying Carbon v10.25.0. The problems are seen here:

https://percy.io/538fc19a/Carbon-Expressive-Theme/builds/7956640

For the checkbox component seen here:

<img width="1933" alt="Screen Shot 2020-12-03 at 2 13 58 PM" src="https://user-images.githubusercontent.com/30183447/101076975-1f251580-3572-11eb-8f18-72799bda32c7.png">

The label needs to be higher per the [design specs](https://ibm.box.com/s/g3hgag5womogmzxa35kd3zv3icbhnq6s)

wonilsuhibm:
<img width="1933" alt="Screen Shot 2020-12-03 at 2 13 58 PM" src="https://user-images.githubusercontent.com/30183447/101076975-1f251580-3572-11eb-8f18-72799bda32c7.png">

asudoh:
Seems to have been caused by carbon-design-system/carbon#7373, and seems intentional.

RobertaJHahn:
Fix must be made in the Expressive Theme

PR #4785, opened 2020-12-29
fix(checkbox-expressive): removed top padding so text aligns properly

RaphaelAmadeu:
### Related Ticket(s)

#4622 

### Description

An recent update on Carbon added top-padding to the checkbox component, changing the intended alignment for it on our expressive theme, this PR removes that padding so it aligns as supposed to in our expressive theme.


jeffchew:
@RaphaelAmadeu may need a design eye, but it looks top aligned now and not vertically centered. I think the top padding might need to be maybe a pixel?

jeffchew (approved):
LGTM!

PR #4788, opened 2020-12-29
chore(percy): suppressing callout media video player in percy

jeffchew:
### Related Ticket(s)

No related issue

### Description

This suppresses the callout media with video in percy, which is causing false positives in visual regression tests.

### Changelog

**Changed**

- hide `dds-callout-with-media-video` in percy/browserstack


asudoh (approved):
LGTM 👍 - Thanks @jeffchew!

PR #4790, opened 2020-12-29
chore(dependabot): upgrading jest to v25 and node-notifier to v8

jeffchew:
### Related Ticket(s)

No related issue

### Description

This is in response to a dependabot security risk identified with
node-notifier, where a minimum of v8.0.1 needs to be used. The repo is
currently on v24 of jest, which is using a much older version of
`node-notifier` (v5). We currently cannot upgrade to Jest 26
easily as Storybook v6 is needed for Jest 26 support. This patch will
however upgrade to Jest 25, then resolving `node-notifier` to v8.0.1 so
at least there is less of a version gap for that dependency.

Storybook 6 upgrade is being handled separately, but is a large upgrade
across the board.

### Changelog

**Changed**

- Upgrade to Jest v25
- Resolve `node-notifier` to `>=8.0.1`
- Adding `.prettierignore` configuration for `storybook-static-react` folder


asudoh (approved):
LGTM 👍 - Thanks @jeffchew!

Issue #4784, opened 2020-12-29
Web Components | Button CTA video type not behaving properly


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

The video type for Button CTA is not opening the lightbox media player. It opens to the iframe version of the story. The iframe version doesn't appear to do anything.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

CTA > Button > Video type

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Button click opens the lightbox media viewer with video

> What browser are you working in?

Chrome

> What version of Carbon for IBM.com are you using?

Canary release

## Steps to reproduce the issue

1. Go to http://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-cta--button
2. Change type to "video"
3. Click the button

PR #4793, opened 2020-12-30
fix(button-cta): supports video mode

asudoh:
### Related Ticket(s)

Refs #4784.

### Description

Adds inheritable `_handleClick()` method to `<dds-button>` so `<dds-button-cta>` can override it.

Also renames `_handleClickLink()` in internal link component to `_handleClick()` so CTA code can handle them in a single method.

### Changelog

**New**

- Inheritable `_handleClick()` method to `<dds-button>`

**Changed**

- Renames `_handleClickLink()` in internal link component to `_handleClick()` so CTA code can handle them in a single method.


jeffchew (approved):
Thank you for the fix, @asudoh !

PR #4794, opened 2020-12-30
docs(content-block): remove slot attribute from heading

asudoh:
### Description

Remove `slot` attribute from `<dds-content-block-heading>` examples as `<dds-content-block-heading>` has `slot` attribute set by default.

### Changelog

**Removed**

- `slot` attribute from `<dds-content-block-heading>` examples.


jeffchew (approved):
LGTM!

Issue #4326, opened 2020-10-28
Components: Introduce a component for heading/copy, etc.


asudoh:
### The problem

Most of the components with heading contents in Web Components codebase has a corresponding component for heading contents, for the styling purpose. OTOH `<dds-link-list>` doesn't.

Also, `<dds-image-with-caption>` supports `heading` attribute but it doesn't have a corresponding `<slot>`.

### The solution

Consider introducing a similar one for `<dds-link-list>`. Also scan the codebase to see if we need any more components for heading purpose, for other areas.

### Additional information

Similar thing is seen with `<dds-card>`, etc.

RobertaJHahn:
@asudoh Can this issue be closed?  I looks like all of the linked PRs are merged.

asudoh:
@RobertaJHahn Thank you for checking - Would like to keep this open if possible. Further code inspection is needed to ensure we have finished everything.

asudoh:
@RobertaJHahn This work is complete now. Thanks again for checking!

PR #4795, opened 2020-12-30
chore(link-list): introduce heading component

asudoh:
### Related Ticket(s)

Refs #4326.

### Description

Introduces `<dds-link-list-heading>` to make a choice of the approach of `<div>`/`<span>` with specifying `slot` attribute vs. dedicated components to the latter, to align to other components.

The dedicated components approach allows more control over styling.

### Changelog

**New**

- `<dds-link-list-heading>`, the heading for link list.


jeffchew:
Thank you @asudoh, can you update the documentation too? It still references as a slot.

asudoh:
@jeffchew Seems that the document of `<dds-link-list>` does not refer to heading at all, added it.

@RaphaelAmadeu Seems that the document refers only to card usage... Would you add text usage, too? Thanks!

jeffchew:
> @jeffchew Seems that the document of `<dds-link-list>` does not refer to heading at all, added it.
> 
> @RaphaelAmadeu Seems that the document refers only to card usage... Would you add text usage, too? Thanks!

@asudoh I will approve as-is, we can add more documentation in a separate PR.

jeffchew (approved):
Thank you @asudoh !

PR #4817, opened 2021-01-05
fix(content-block-patterns): fixing mobile margins 

guilhermelMoraes:
### Description

Added `64px` margin both to `content-block-segmented` and `content-block-simple` for `md` breakpoint and below. 

**Current**
Screen Shot 2021-01-05 at 16 48 24

**Updated**
Screen Shot 2021-01-05 at 16 50 00


annawen1 (approved):
LGTM!

Issue #4813, opened 2021-01-05
Web component: Masthead L1 - Menu options does not open on desktop


Praveen-sr:
## Detailed description

Issue: Masthead L1 - Menu options does not open on desktop

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/2d87a280-6fcb-43ad-ac7c-2792e41faac1)

PR #4822, opened 2021-01-06
fix(masthead): ensure L1 won't be crippled

asudoh:
### Related Ticket(s)

Refs #4813.

### Changelog

**Changed**

- A change to ensure L1 menu body won't be crippled.


kennylam (approved):
LGTM thanks @asudoh

Issue #4824, opened 2021-01-06
LinkList: incorrect padding in card variation


kennylam:
## Detailed description

Link list card variation has extra left/right padding in non-mobile breakpoints.

## Additional information

**Current**
![image](https://user-images.githubusercontent.com/909118/103774018-ec02f580-4ff9-11eb-8a58-76f536ef3e81.png)

**Expected**
![image](https://user-images.githubusercontent.com/909118/103774322-60d62f80-4ffa-11eb-9ad7-a75b04e25657.png)

PR #4825, opened 2021-01-06
fix(LinkList): fix linklist card padding

kennylam:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/4824

### Description

Fixes padding for Link list card variation in non-mobile breakpoints.

### Changelog

**Changed**

- remove extra padding


asudoh (commented):
Thank you for the fix @kennylam! Seems that the padding (or margin) is increased in small screen, wondering if the case/intentional.

kennylam:
> Thank you for the fix @kennylam! Seems that the padding (or margin) is increased in small screen, wondering if the case/intentional.

Oops, good catch thanks @asudoh.

jeffchew (approved):
LGTM!

Issue #4814, opened 2021-01-05
Web component: Masthead - Overflow icon remains on left side of menu's, even if no options to show


Praveen-sr:
## Detailed description

Issue: Overflow icon remains on left side of menu's, even if no options to show.
Overflow icon must be on the right side.

Steps to reproduce 
1. Open the masthead component on web component canary 
2. Click on overflow icon on the right side
3. Click on search icon and close the search 
4. Now, overflow icon is on left side of the menu's 
5. Click on it, but it still remains there.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/43cbddf1-48fc-498e-9af2-ba39f931ee17)

PR #4838, opened 2021-01-07
fix(masthead): recreate IntersectionObserver upon showing/hiding nav

asudoh:
### Related Ticket(s)

Refs #4814.

### Description

Ensures `IntersectionObserver` in `<dds-top-nav>` is re-created when the nav content is shown after it's hidden, to avoid stale DOM node references.

### Changelog

**Changed**

- A change to ensure `IntersectionObserver` in `<dds-top-nav>` is re-created when the nav content is shown after it's hidden.


jeffchew (approved):
Thanks @asudoh !

Issue #4737, opened 2020-12-17
Footers not displaying properly in FF incognito


Daynssjal:
## Detailed description

Footer components are not displaying properly in Firefox incognito window. 
Default and Short footers are only black banners with languages dropdown, but no legal links.
Micro footer is not visible at all.

Please open your storybook in FF incognito to see this issue:
https://ibmdotcom-react.mybluemix.net/?path=/story/components-footer--default

> Is this issue related to a specific component?
Footer 

> What browser are you working in?
Firefox - incognito window

> What version of Carbon for IBM.com are you using?
1.13.0
1.14.0

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?
IBM Knowledge Center - IBM Docs - deadline 15th February

![footer_issue_1](https://user-images.githubusercontent.com/33804471/102484181-27974900-4066-11eb-9d82-19ebe16deb14.jpg)

kennylam:
This occurs in non-incognito as well and seems to be caused by Firefox's tracking protections. All data loads correctly in `ibm.com` for both FF incognito and regular browsing. Possibly CORS or mixed content (http/https) causing the issue? Need to investigate further.

![image](https://user-images.githubusercontent.com/909118/103654538-02df1480-4f34-11eb-9f18-ebb3c009fd36.png)

Disabling enhanced tracking will correctly load the translation data.

![image](https://user-images.githubusercontent.com/909118/103654641-24d89700-4f34-11eb-9871-321d205cc6eb.png)

annawen1:
This is the list endpoints that are being blocked by FF's tracking protection
<img width="435" alt="Screen Shot 2021-01-07 at 9 02 47 AM" src="https://user-images.githubusercontent.com/54281166/103901507-3bfbbe00-50c7-11eb-80e8-28360359b920.png">

ljcarot:
@annawen1 We should involve Juan Leon.  These are all urls coming from loading the tags via tealium.  He should be aware that FF is blocking these.  I'm abit surprised to see cloud.ibm.com in the list though.  Do we know which of these is where we are loading the content for mh and footer?

annawen1:
These are the links being blocked by FF's tracking protection:

<img width="435" alt="Screen Shot 2021-01-07 at 9 02 47 AM" src="https://user-images.githubusercontent.com/54281166/103903529-01dfeb80-50ca-11eb-854d-8677ed14df12.png">

[...]

PR #4848, opened 2021-01-07
fix(Footer): swap out translation host default endpoint for FF

annawen1:
### Related Ticket(s)

Footers not displaying properly in FF incognito #4737

### Description

Firefox's private window has tracking protection enabled by default that blocks the ibm.com endpoint api calls so users don't see the footer or masthead populated when using the private window.

Solution is to swap the translation host from ibm.com to the akamai endpoint instead.

### Changelog

**Changed**

- set the default host for Locale and Translation api to akamai


jeffchew (approved):
LGTM!

Issue #4609, opened 2020-12-03
Web component: Cards - CTA type is not changing on the card [ Storybook issue]


Praveen-sr:
## Detailed description

Issue: CTA type "jump, external & download" is not changing on the card 

## Additional information

![CardCTA.gif](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/575d6503-731a-4839-a958-28a93380c098)

PR #4853, opened 2021-01-08
fix(ContentGroupSimple): fix stories card cta knobs

ariellalgilmore:
### Related Ticket(s)

#4609 

### Description

Fix Content Group Simple card cta story, so that knobs change card to local, jump, external, download, etc...

### Changelog

**Changed**

- stories to adjustable parameters


jeffchew (approved):
LGTM!

PR #4855, opened 2021-01-08
fix(quote): changed inverse attribute for color-scheme

RaphaelAmadeu:
### Related Ticket(s)

#4122

### Description

After a discussion about the inverse attribute we decided to remove it and use the color-scheme attribute instead, this pr adresses this change.


asudoh (commented):
Hi @RaphaelAmadeu thank you for jumping in! Has removing `inverse` (or `color-scheme`) entirely been considered? The best is quote style works accommodating the outer theme.

RaphaelAmadeu:
@asudoh Currently the `inverse` theming is part of the functional specs, so it's still intended to be used.

packages/styles/scss/components/quote/_quote.scss
```diff
@@ -22,7 +23,20 @@
     }
   }
 
-  .#{$prefix}--quote__container {
+  :host(#{$dds-prefix}-quote).#{$prefix}--quote__inverse {
```
asudoh, line 26: Instead of adding a CSS class that replaces `inverse` attribute, would you try simply making the quote work in dark theme? Please don't hesitate ping me offline if you have any questions. Thanks!

asudoh (commented):
Thank you for keeping trying @RaphaelAmadeu!

Issue #4843, opened 2021-01-07
Web component: Card component Pictogram - Placement bottom - On hover, Keyboard user cannot access this textual content.


Praveen-sr:
## Detailed description

Issue: When the user hovers the mouse on the pictogram, the pictogram disappears and a textual preview appears instead. This behavior is not Keyboard Accessible. The Keyboard user cannot access this textual content.

Expected:
Make the mouse hover content accessible for the Keyboard user as well.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/c9394f4c-6c4c-4141-8166-2962927a8824)

PR #4861, opened 2021-01-11
fix(card-pictogram): now copy is also visible on focus

guilhermelMoraes:
### Related tickets

Web component: Card component Pictogram - Placement bottom - On hover, Keyboard user cannot access this textual content. #4843

### Description

Now the copy is also visible on focus


jeffchew (approved):
Thank you @guilhermelMoraes !

Issue #4485, opened 2020-11-12
Web components: mega menu content is not aligned to grid


oliviaflory:
## Detailed description

> Describe in detail the issue you're having.

The mega menu content is not aligned to the grid per the design, on all menus.

The text should be aligned to the columns and the interactive area should overhang. In this example the 'highlighted' left column and View all CTA at the bottom are correct

**Web components**
![Screen Shot 2020-11-12 at 5.06.48 PM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/3bcf64be-3c32-4f76-b9bd-8fe16f58bbe1)

**React**
This is how the text should be aligned:

![Screen Shot 2020-11-12 at 5.28.25 PM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/0cd0bd47-f5a8-4ec7-abc2-dc46bd268530)

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

> Is this issue related to a specific component?

Web components: mega menu

> What did you expect to happen? What happened instead? What would you like to
> see changed?

> What browser are you working in?

Chrome

> What version of Carbon for IBM.com are you using?

[Canary](https://ibmdotcom-web-components-canary.mybluemix.net/iframe.html?id=components-masthead--default)

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. Open mega menu
2. Turn on grid overlay through Carbon dev tool
3. Step three
4. etc.

## Additional information

![Screen Shot 2020-11-12 at 5.07.16 PM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/873e27f1-9796-4ded-8ab8-6d2e9d732311)

PR #4867, opened 2021-01-11
fix(Megamenu): align megamenu categories with grid

annawen1:
### Related Ticket(s)

Web components: mega menu content is not aligned to grid #4485

### Description

Align the categories in the megamenu according to the grid

BEFORE:

AFTER:
Screen Shot 2021-01-11 at 5 50 56 PM
Screen Shot 2021-01-11 at 5 50 22 PM

### Changelog

**Changed**

- remove left padding for the right navigation section
- set column-gap to 0


ariellalgilmore (approved):
LGTM! Thanks @annawen1!

Issue #4851, opened 2021-01-08
Content block: Automatically define slot attribute for the copy component


asudoh:
### The problem

`<dds-content-block-copy>` requires explicitly specifying `slot` attribute, though the slot is `copy` for all use cases.

### The solution

Define `slot="copy"` by default. Or make `<slot name="copy">` in component `<dds-content-block-copy>` is used with the default slot (Ref: #4627)

Issue #4627, opened 2020-12-04
Content pattern: Determine what the default slot be used for


asudoh:
### The problem

Some content patterns in Web Components use the default slot for the copy content. OTOH, some have `<slot name="content">` that may be suitable for the default slot.

For example, `<dds-content-block-cards>` and `<dds-content-block-headlines>` have `<slot name="content">` for the cards, that may be eligible for the default slot. Some components have `<slot name="copy">` that may be another candidate.

### The solution

Determine what the default slot should be used for and make all content pattern align to the strategy.

PR #4869, opened 2021-01-12
chore(content-block): reorganize slots

asudoh:
### Related Ticket(s)

Refs #4627.
Refs #4851.

### Description

Unifies the slot name strategy, using the default slot for child contents, and `copy` slot for copy contents, for `<dds-content-block>`, `<dds-content-group>` and their descendants.

Such unification allows us to reduce the overrides, especially for slot names and layout, in the descendants on top of the ancestor components.

`<dds-content-block-copy>` now sets `copy` slot automatically. `<dds-content-block-headlines-item>` no longer sets the slot.

Also:

* Ensures that `<div>`s surrounding copy, media, etc. contents won't be rendered unless the corresponding contents are there.
* Introduces `size="sm"` option to `<dds-content-block-copy>`, so we can use it instead of `<dds-content-item-copy>` in
  `<dds-content-block>`, `<dds-content-group>` and their descendants.
* Introduces `<dds-content-group-copy>` so we can apply styling specific to content group.

### Changelog

**New**

- `size="sm"` option to `<dds-content-block-copy>`.
- `<dds-content-group-copy>`, the copy content of content group.

**Changed**

- Changes to unify the slot name strategy, using the default slot for child contents, and `copy` slot for copy contents, for `<dds-content-block>`, `<dds-content-group>` and their descendants.
- `<dds-content-block-copy>` to set `copy` slot automatically
- `<dds-content-block-headlines-item>` to no longer set the slot.
- Changes to ensure that `<div>`s surrounding copy, media, etc. contents won't be rendered unless the corresponding contents are there.


ariellalgilmore (approved):
LGTM!

Issue #4062, opened 2020-09-25
Web component: Card - Screen reader announces a lengthy text which makes it oververbose.


Praveen-sr:
## Detailed description
On JAWS FF & VO iPhone safari

Issue:  when focus reaches the link in the card, Screen reader announces a lengthy text which makes it oververbose. 
Note: Issue found in React Staging as well.

Expected : It may be possible if the entire content in the card is given in anchor tag. In this scenario, only the heading can be given in anchor tag. Note: the link text/label should be concise and convey only the relevant information. 

## Additional information

- Screenshots or code
- Notes

IgnacioBecerra:
@Praveen-sr Just to make sure, what should be the only text that needs to be read? Should it be the heading (Lorem ipsum... in Storybook), or the cta text?

Due to the way it's structured, the whole card is an anchor element and we can't tab into the inner elements specifically, thus the overly verbose text.

IgnacioBecerra:
From this morning's QA meeting, the heading text should be the **only** element to be read using the screen reader. Thanks for the clarification!
cc: @Praveen-sr @RobertaJHahn

PR #4870, opened 2021-01-12
fix(card): voiceover now only reads heading

IgnacioBecerra:
### Related Ticket(s)

#4062

### Description

This PR makes screen readers read **only** the heading content of the card, ignoring the eyebrow and CTA text.

### Changelog

**Changed**

- added `aria-hidden='true'` to eyebrow and CTA text of Card component


andysherman2121 (approved):
Looks good to me! nice work ignacio

PR #4874, opened 2021-01-12
chore(content-block-copy): setting slot="copy" for it automatically

guilhermelMoraes:
### Description

content-block-copy: setting slot="copy" for it automatically


packages/web-components/src/components/content-block/content-block-copy.ts
```diff
@@ -23,6 +23,14 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
...
 @customElement(`${ddsPrefix}-content-block-copy`)
 class DDSContentBlockCopy extends StableSelectorMixin(DDSMarkdown) {
+  connectedCallback() {
+    super.connectedCallback();
+
+    if (!this.hasAttribute('slot')) {
+      this.setAttribute('slot', 'copy');
+    }
+  }
+
```
asudoh, line 33: Would you simply define `slot` property with `@property({ reflect: true })` and `copy` as the default value? The definition should come after private/protected properties/methods. Thanks!

asudoh (approved):
LGTM 👍 - Thank you for the update @guilhermelMoraes!

Issue #4778, opened 2020-12-23
Replace image placeholder with internal image placeholder for Storybooks


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

We are getting a lot of false positives in Percy image regression as a lot of times, the images fail to load in snapshot tests. For example:

https://percy.io/538fc19a/Carbon-for-IBM.com-React/builds-next/8187213/changed/465859488?browser=chrome&browser_ids=9&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1280&widths=320%2C1280

Ideally, we replace this with an internal, home grown image placeholder so we can avoid these false positives.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

All Storybook components with placeholder images

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Consistent placeholder image loading

> What browser are you working in?

Percy/Browserstack

jeffchew:
I was playing around a bit on this one, one solution I looked into is `holder.js` (https://github.com/imsky/holder). Immediately ran into an issue with dynamically regenerating the placeholder images when changing knobs, as the library is not designed to access any elements in the shadow root for re-initializing imagery.

Will have to see if there are other ways around this.

wonilsuhibm:
Can we use images that we produced? I see Helvetica or Arial being used in the current images.

jeffchew:
@wonilsuhibm yes, we can use our own imagery, would the design team be able to provide?

wonilsuhibm:
@jeffchew I have these updated designs going:
![Artboard Copy 43](https://user-images.githubusercontent.com/30183447/103693717-bca3a880-4f67-11eb-9494-aa249206e39e.png)
![Artboard Copy 69](https://user-images.githubusercontent.com/30183447/103693718-bca3a880-4f67-11eb-8b73-446ef982c5e2.png)
![Artboard Copy 64](https://user-images.githubusercontent.com/30183447/103693719-bca3a880-4f67-11eb-9c9e-dbe49bf38ae9.png)
![Artboard Copy 59](https://user-images.githubusercontent.com/30183447/103693720-bca3a880-4f67-11eb-8561-0fdfb8ef59aa.png)
![Artboard Copy 54](https://user-images.githubusercontent.com/30183447/103693721-bca3a880-4f67-11eb-9917-8c7cae18f663.png)
[...]

PR #4879, opened 2021-01-12
chore(images): swapping FPO images with custom imagery

jeffchew:
### Related Ticket(s)

Refs #4778 

### Description

This swaps out the FPO imagery used from fpoimg.com and replaces with
our own custom imagery. Part of this move is to avoid false positives
our automated visual regression tests are seeing when the FPO imagery
loading fails, since it is an externally hosted set of assets.

Using real imagery would also surface any issues our team would
otherwise would not have picked up.

### Changelog

**New**

- Storybook imagery


kennylam:
@jeffchew New images look great. Looks like you forgot `ContentBlockCards - with images`.

jeffchew:
> @jeffchew New images look great. Looks like you forgot `ContentBlockCards - with images`.

Thanks @kennylam ! I think I got it now, can you double check?

kennylam (approved):
LGTM thanks @jeffchew!

PR #4882, opened 2021-01-12
fix(quote): inverse style not working

RaphaelAmadeu:
### Related Ticket(s)

#4122

### Description

After checking with @photodow if `inverse` attribute should remain, I've fixed an issue where inverse theme was not working.


asudoh:
Thank you for your update @RaphaelAmadeu! Sent a note offline.

asudoh (approved):
LGTM 👍 - Thanks @RaphaelAmadeu for the update!

Issue #4811, opened 2021-01-05
Web component: Masthead - On Mobile and Tab view, Hamburger menu and menu options both are displayed 


Praveen-sr:
## Detailed description

Issue: On Mobile and Tab view, Hamburger menu and menu options both are displayed 

Note: On react it is fine.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/1b03a13e-96f8-49a5-8392-32de9a4bf5cd)

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/8f600c2d-d497-4ac6-9898-30e314d68638)

asudoh:
@Praveen-sr Tried iPhone 12, but couldn't reproduce this:

![image](https://user-images.githubusercontent.com/1259051/103745048-25f7ea00-5042-11eb-9dd1-5c5818788fc6.png)

Would you be able to let us know what specific phone you reproduced this issue with? Thanks!

Praveen-sr:
@asudoh I tested on iPhone 11 with iOS 14, am able to re produce the issue 

![Mast1.gif](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/f616c798-7354-486d-961c-b4bcb1a2ebfb)

asudoh:
Thanks @Praveen-sr! I could reproduce this _only_ if I rotate the iPhone after opening left nav. If I open the left nav _after_ rotating the iPhone, the problem doesn't seem to reproduce.

@Praveen-sr Is it same your end, too? Also, could you share the specific iPad device and iOS version where you reproduced the issue with iPad? Thanks!

Praveen-sr:
@asudoh 
1. Yes, we need to open the hamburger menu and then rotate the device. 
2. iPad - I have tested on iPad 7, 8 and iPad Air 4 with iOS 13, 14

![Mast2.gif](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/b1ec4f88-8963-4c4a-b0b7-aeef0fe37f74)

PR #4886, opened 2021-01-13
fix(left-nav): ensure hidden for large screen

asudoh:
### Related Ticket(s)

Refs #4811.

### Description

Fixes `<dds-left-nav>` and `<dds-left-nav-overlay>` to ensure that they are hidden, even in its active state, in screen bigger than or equal to 800 pixels.

Doing so fixes a bug in a condition where the screen is enlarged after opening left nav.

### Changelog

**Changed**

- Fix to `<dds-left-nav>` and `<dds-left-nav-overlay>` to ensure that they are hidden, even in its active state, in screen bigger than or equal to 800 pixels.


jeffchew (approved):
Thanks @asudoh !

Issue #3818, opened 2020-09-01
Component: Micro Footer - On Select by location or language, search edit field, the results are getting updated dynamically, but screen reader user would not be aware


Praveen-sr:
## Detailed description
On Firefox JAWS & iPhone VO

Issue : As we provide the input in the search edit field, the results are getting updated dynamically. But, Screen reader user would not be aware of this.

Expected:
1. Screen reader should announce the number of results as they dynamically update. 
2. Provide the following instruction and Programmatically Associate using aria-described by
"Note: Location/Language will be dynamically updated as you provide the input".

PR #4893, opened 2021-01-13
feat(locale-modal): screenreader reads out number of results

IgnacioBecerra:
### Related Ticket(s)
#3818

### Description

Added a label within the input element to inform users of the dynamic search results.
Also introduced feature where upon input, the screenreader will read out the number of results available. 

### Changelog

**New**

- added invisible `<p>` element that updates itself with the number of results available for accessibility purposes


packages/react/src/components/LocaleModal/LocaleModalCountries.js
```diff
@@ -88,7 +88,7 @@ const LocaleModalCountries = ({
         <Search
           data-autoid={`${stablePrefix}--locale-modal__filter`}
           placeHolderText={modalLabels.searchPlaceholder}
-          labelText={modalLabels.searchLabel}
+          labelText={`${modalLabels.searchLabel}. Note: Location/Language will be dynamically updated as you provide the input`}
```
jeffchew, line 91: We probably need to set this as something configurable, in case this is translated in the other 15 languages we support. We can make english the default value.

packages/styles/scss/components/locale-modal/_locale-modal.scss
```diff
@@ -267,6 +267,17 @@
     flex: 1;
     overflow-y: auto;
   }
+
+  .#{$prefix}--sr-only {
```
jeffchew, line 271: This looks like the standard hide for screen readers. @asudoh @kennylam @annawen1 do you know if we have something like this already available either in our repo or Carbon?

asudoh: @jeffchew Yes, there is `.#{$prefix}--assistive-text`.

IgnacioBecerra: Ah didn't know about that class, will replace it with that one in the next push. Thanks for the heads up!

packages/react/src/components/LocaleModal/LocaleModalCountries.js
IgnacioBecerra: Opened a [PR in ibm-dotcom-translations](https://github.ibm.com/webstandards/ibm-dotcom-translations/pull/251) to address this. At the moment I've only modified the `us-en`/default value for `searchLabel`, once approved the change will be reflected in the live versions.

For now, this current PR still reads out the number of results as intended.

cc: @Praveen-sr

packages/react/src/components/LocaleModal/LocaleModalCountries.js
```diff
@@ -208,6 +212,18 @@ export const filterLocale = (
...
       : modalLabels.availabilityText;
+
+  /**
+   *  Reflect number of results in the accessibility readout
+   */
+  const resultsElement = document.querySelector(`.${prefix}--assistive-text`);
+  const resultsCount = localeItems.length - localeItemsHidden.length;
+
+  if (resultsElement) {
+    resultsElement.innerHTML = `${resultsCount} ${
```
asudoh, line 223: `innerHTML` causes HTML parser to run and thus using it for text content will be redundant:

```suggestion
    resultsElement.textContent = `${resultsCount} ${
```

packages/react/src/components/LocaleModal/LocaleModalCountries.js
```diff
@@ -208,6 +212,18 @@ export const filterLocale = (
     localeItems.length === localeItemsHidden.length
       ? modalLabels.unavailabilityText
       : modalLabels.availabilityText;
+
+  /**
+   *  Reflect number of results in the accessibility readout
+   */
+  const resultsElement = document.querySelector(`.${prefix}--assistive-text`);
```
asudoh, line 219: This `querySelector()` call has to be closely scoped. Note that `bx--assistive-text` is general-purpose CSS class for "hidden visually but visible from screen reader" elements.

asudoh (commented):
Thank you for eliminating the dup of `.bx--assistive-text` @IgnacioBecerra!

packages/react/src/components/LocaleModal/LocaleModalCountries.js
```diff
@@ -208,6 +212,21 @@ export const filterLocale = (
     localeItems.length === localeItemsHidden.length
       ? modalLabels.unavailabilityText
       : modalLabels.availabilityText;
+
+  /**
+   *  Reflect number of results in the accessibility readout
+   */
+  const resultsElement = document.querySelector(
+    `.${prefix}--locale-modal__list .${prefix}--assistive-text`
+  );
```
asudoh, line 221: Thank you for the update @IgnacioBecerra, another step; Would you try getting `ref` of the top-level `<div>` and run `.querySelector()` from there?

asudoh (approved):
Looks good for now, it'll be nice if you can discuss with the team to see if the screen reader content should be translatable. If it's the case, we need a React prop for the translatable content with plural support, which would have a default value like:

```javascript
({ count }) => `${count} ${count == 1 ? 'result' : 'results'} found.`
```

Thanks @IgnacioBecerra!

packages/react/src/components/LocaleModal/LocaleModalCountries.js
jeffchew: Thanks @IgnacioBecerra , I approved the translation change so that part should be good. Can you check the CI checks and see what's failing?

Issue #4802, opened 2021-01-04
Web component and React: Local modal - Placeholder (Search by location or language)


Praveen-sr:
## Detailed description

Actual Result:
Foreground:#AFADAD
Background:#F4F4F4

The contrast ratio is: 2.0:1
Text failed at Level AA

Expected Result:
Provide below contrast ratio.
Foreground:#7F7E7E
Background:#F4F4F4

The contrast ratio is:3.7:1
Text passed at Level AA

PR #4896, opened 2021-01-13
fix(locale-modal): replaced placeholder color for a bolder one

IgnacioBecerra:
### Related Ticket(s)
#4802

### Description

The previous placeholder color for the Web Components version was failing contrast ratio testing. Changed the placeholder text to match the `$text-02` token used in the React version, where the contrast ratio has been fixed.

### Changelog

**Changed**

- `<dds-search>` placeholder color changed to `$text-02` to match React version


asudoh (commented):
LGTM basically, can we add a comment referring to https://github.com/carbon-design-system/carbon/blob/v10.25.0/packages/components/src/globals/scss/_helper-mixins.scss#L44-L51 and say that we are deliberately diverge from that, and the reason why we do so? (Background; People tend to wonder why we are defining one-off placeholder color here) Thanks @IgnacioBecerra!

asudoh (approved):

oliviaflory:
@IgnacioBecerra Is this something we can propose to the Carbon repo? IMO, we should be avoiding making overrides to core components that we have to maintain. 

I did a little digging in the Carbon repo and found this issue where they updated the placeholder mixins to address accessibility issues with the place holder text. https://github.com/carbon-design-system/carbon/pull/6914
Do we have this mixin applied in our locale selector? 
Their fix of #6f6f6f ($text-05) passes by 4.57 on #f4f4f4 ($field-01).

asudoh:
@oliviaflory Interesting point, thank you for bringing this up - I believe the original code used `$text-05` (@IgnacioBecerra please correct me if I'm wrong here) where QA team is complained against. Any thoughts @IgnacioBecerra? Thanks!

IgnacioBecerra:
@asudoh @oliviaflory Yeah I think that's correct, the placeholder was `$text-05`. I consulted with the QA team and was told that the React version had been updated, I copied the value over from there.

asudoh:
@IgnacioBecerra Would you please check the color contrast of the original code to see if it really misses our color contrast standard? I think latest Chrome DevTools has a feature that shows the color contrast.

IgnacioBecerra:
@asudoh I tried using the color contrast tool but it appears that it doesn't appear to be working.
Screen Shot 2021-01-14 at 9 27 47 AM

However, I found [this website](https://webaim.org/resources/contrastchecker/) and got these results:
(`$text-05` and `$field-01` respectively)
Screen Shot 2021-01-14 at 9 37 37 AM

Looks like it passes the WCAG AA test, but fails on AAA.

However, using `$text-02`, it passes both:
Screen Shot 2021-01-14 at 9 39 44 AM

cc: @oliviaflory

oliviaflory:
Thank you @IgnacioBecerra, I didn't realize we were having to pass WCAG AAA, I was under the impression we only needed AA. Might be good to confirm with @Praveen-sr if this is the new standard we should be aware of.

asudoh:
@IgnacioBecerra The associated issue (#4802) seems to say that the pseudo element's color in before-change code was `#AFADAD` whereas you saw it actually was `#6F6F6F`. Is it true...?

IgnacioBecerra:
@asudoh Yeah that's correct, this is the original placeholder using `#6F6F6F` (`$text-05`):
Screen Shot 2021-01-14 at 4 24 09 PM

This is with`#AFADAD` (no token associated with this)
Screen Shot 2021-01-14 at 4 24 25 PM

Just as an extra, this is with the current color `#525252` (`$text-02`):
Screen Shot 2021-01-14 at 4 27 36 PM

I thought they had copied `#AFADAD` from another element when writing the issue, given that they still mentioned the placeholder in the title + the correct background color.

asudoh:
I see, does it mean that the issue was never reproducible? Is it worth checking with QA team?

PR #4897, opened 2021-01-14
refactor(cta-section): reorganize slots

asudoh:
### Related Ticket(s)

Refs #4627.

### Description

Repurposes the default `<slot>` in `<dds-cta-section>` from the copy content to the child content items. Also makes `<dds-cta-section>` inherit `<dds-content-block>` instead of `<dds-content-item>` in its TypeScript code (not style code). These are because `<dds-cta-section>` doesn't work as a (child) content item; Instead, it _contains_ child items.

### Changelog

**Changed**

- Repurposes the default `<slot>` in `<dds-cta-section>` from the copy content to the child content items.
- Makes `<dds-cta-section>` inherit `<dds-content-block>` instead of `<dds-content-item>` in its TypeScript code (not style code).


jeffchew (approved):
Thank you @asudoh ! Can you update the pinned issue?

asudoh:
Thanks @jeffchew - Updated #4079.

Issue #4701, opened 2020-12-10
Change profile endpoint on Carbon Masthead for signed in state


ljcarot:
**This work must be done by the end of Jan 2021**

IDaaS is being replaced with new ISV.  This change will impact the endpoint that is currently part of the Profile Icon on the masthead which displays if a user is signed out or signed in.

**Describe the solution you'd like** 
Rick McCarty has provided the solution here: 
With respect to the status API calls made by the masthead, these will need some changes  for ISV.

Currently, if the current page is under www.ibm.com, the masthead makes a request to the following endpoint:

https://idaas.iam.ibm.com/v1/mgmt/idaas/user/status/

If the page is under another domain, for example myibm.ibm.com, etc, the call is a bit different - it adds a callback parameter for JSONP to avoid CORS issues:

https://idaas.iam.ibm.com/v1/mgmt/idaas/user/status/?callback=__jp1

With the change to ISV, the new endpoint will be:

https://login.ibm.com/v1/mgmt/idaas/user/status/

Also, ISV will no longer support the "callback=" parameter, meaning that JSONP can no longer be used to get around CORS issues. Instead, we will have to include a more comprehensive list of domains in our ISV configuration to handle CORS.

Original request is here: https://jsw.ibm.com/browse/IRFC-8

RobertaJHahn:
@ljcarot Can you let us know when this ISV change is being made and when our change is required?  Thanks.

ljcarot:
This change is needed now.  The Carbon change should be in January release and v18 can be any time before end of January.  It needs to be done in advance of ISV cut over tentatively set for late January.

jeffchew:
@RobertaJHahn can you add to the next engineering refinement meeting to size? I also will need to get back to Rick McCarty on a whitelist for lower environment testing.

RobertaJHahn:
@jeffchew Done

annawen1:
@jeffchew has provided Ricky with a list of domains we need whitelisted. Waiting for those domains to be whitelisted so we can test the new endpoint locally.

PR #4909, opened 2021-01-14
feat(profile): change profile api endpoint

annawen1:
### Related Ticket(s)

Change profile endpoint on Carbon Masthead for signed in state #4701

### Description

Change profile endpoint which no longer allows for jsonp to get around CORS errors. Set the new preprod profile endpoint as default. 

### Changelog

**Changed**

- use `axios` for the profile calls
- set default host endpoint to preprod url

**Removed**

- `jsonp` package - no longer needed


packages/services/src/services/Profile/Profile.js
```diff
@@ -1,18 +1,17 @@
...
-import jsonp from 'jsonp';
+import axios from 'axios';
 
 /**
  * @constant {string | string} Host for the profile status API call
  * @private
  */
 const _host =
-  (process && process.env.PROFILE_HOST) || 'https://idaas.iam.ibm.com';
+  (process && process.env.PROFILE_HOST) || 'https://prepiam.ice.ibmcloud.com';
```
jeffchew, line 14: we should actually set this as prod. We can test the preprod environment with PROFILE_HOST.

jeffchew (changes requested):

packages/services/src/services/Profile/Profile.js
jeffchew: `https://login.ibm.com` I believe

packages/services-store/src/actions/profileAPI.ts
```diff
@@ -40,8 +40,8 @@ export type ProfileAPIActions = ReturnType<typeof setMonitorUserStatusError> | R
  * @returns A Redux action that monitors user authentication status.
  */
 export function monitorUserStatus(): ThunkAction<void, { profileAPI: ProfileAPIState }, void, ProfileAPIActions> {
-  return dispatch => {
-    ProfileAPI.monitorUserStatus((error: Error, status: UserStatus) => {
+  return async dispatch => {
+    await ProfileAPI.monitorUserStatus((error: Error, status: UserStatus) => {
```
asudoh, line 44: Seems that this PR changes `ProfileAPI.monitorUserStatus()` from a callback-based function to an async function. And thus the call here should be changed accordingly?

asudoh: Oh sorry didn't realize we had `ProfileAPI.getUserStatus()` already... My apologies. In this case, we can rename `monitorUserStatus()` Redux action to `getUserStatus()`, change this implementation to use `ProfileAPI.getUserStatus()` (instead of `ProfileAPI.monitorUserStatus()`) and change `masthead-container.ts` and `masthead-composite.ts` accordingly.

jeffchew (approved):
LGTM!

packages/services-store/src/actions/profileAPI.ts
```diff
@@ -16,13 +16,25 @@ import { UserStatus, PROFILE_API_ACTION, ProfileAPIState } from '../types/profil
...
+ * @param status The promise of the REST call for  user status that is in progress.
+ * @returns A Redux action to set the state that the REST call for user status is in progress.
+ * @private
+ */
+export function setRequestUserStatusInProgress(status: Promise<UserStatus>) {
+  return {
+    type: PROFILE_API_ACTION.SET_REQUEST_USER_STATUS_IN_PROGRESS,
+    status,
+  };
+}
```
asudoh, line 36: As seen at https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.14.0/packages/services-store/src/actions/localeAPI.ts#L14-L24, we use `request` variable/property names (instead of `status`) in this kind of Redux action. There are three places to make the change, one is JSDoc comment, one is function arg, one is property name.

packages/services-store/src/actions/__tests__/profileAPI.test.ts
```diff
@@ -25,45 +25,42 @@ const mockStore = configureMockStore<
...
-      {
-        type: PROFILE_API_ACTION.SET_USER_STATUS,
-        status: { user: USER_AUTHENTICATION_STATUS.UNAUTHENTICATED },
-      },
-    ]);
-  });
+  // it('dispatches the action to get user authentication status', async () => {
+  //   ProfileAPI.getUserStatus.mockResolvedValue({ user: 'Unauthenticated' });
+  //   const store = mockStore();
+  //   await store.dispatch(ProfileAPI.getUserStatus());
```
asudoh, line 40: Instead of `ProfileAPI.getUserStatus()` (that makes a call to `@carbon/ibmdotcom-services` module), we need to call a Redux action here. Here is a test in this category for reference: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.14.0/packages/services-store/src/actions/__tests__/localeAPI.test.ts#L84

packages/services-store/src/actions/__tests__/profileAPI.test.ts
```diff
@@ -25,45 +25,42 @@ const mockStore = configureMockStore<
...
+  //   });
+  //   const store = mockStore();
+  //   store.dispatch(getUserStatus());
+  //   expect(convertValue(store.getActions())).toEqual([
+  //     {
+  //       type: PROFILE_API_ACTION.SET_ERROR_REQUEST_USER_STATUS,
+  //       error: 'error-getuserstatus',
+  //     },
+  //   ]);
+  // });
```
asudoh, line 65: A reference of test in this category: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.14.0/packages/services-store/src/actions/__tests__/localeAPI.test.ts#L107-L127

packages/web-components/src/components/masthead/masthead-composite.ts
```diff
@@ -529,7 +528,7 @@ class DDSMastheadComposite extends LitElement {
    * The user authentication status.
    */
   @property({ attribute: 'user-status' })
-  userStatus?: USER_AUTHENTICATION_STATUS;
+  userStatus = 'Unauthenticated';
```
asudoh, line 531: Wondering `@carbon/ibmdotcom-services` still returns `Unauthenticated` string whereas it seems to return an user ID (email address) for authenticated state. Is it the case? If so, I suggest we define a const in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.14.0/packages/services-store/src/types/profileAPI.ts so we don't need to define `'Unauthenticated'` string literals all over the place.

packages/web-components/src/components/masthead/masthead-composite.ts
```diff
@@ -575,7 +574,8 @@ class DDSMastheadComposite extends LitElement {
       l1Data,
       _loadSearchResults: loadSearchResults,
     } = this;
-    const authenticated = userStatus === USER_AUTHENTICATION_STATUS.AUTHENTICATED;
+    console.log('userStatus', userStatus);
```
asudoh, line 577: I believe we meant to remove this line?

packages/services-store/src/actions/profileAPI.ts
```diff
@@ -34,19 +46,24 @@ export function setUserStatus(status: UserStatus) {
...
-export function monitorUserStatus(): ThunkAction<void, { profileAPI: ProfileAPIState }, void, ProfileAPIActions> {
-  return dispatch => {
-    ProfileAPI.monitorUserStatus((error: Error, status: UserStatus) => {
-      if (error) {
-        dispatch(setMonitorUserStatusError(error));
-      } else {
-        dispatch(setUserStatus(status));
-      }
-    });
+export function getUserStatus(): ThunkAction<Promise<UserStatus>, { profileAPI: ProfileAPIState }, void, ProfileAPIActions> {
```
asudoh, line 57: I should have told you earlier (my apologies), Redux actions in this category takes `load*` name and we should align the naming strategy. Here's a reference: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.14.0/packages/services-store/src/actions/localeAPI.ts#L145-L164

asudoh (commented):
Thank you for the work @annawen1!

Issue #4599, opened 2020-12-01
Masthead Search click event enhancement


kyjak:
Currently the event fired when user activates the search from masthead is an info event only.
We need a way to prevent the default action to happen, something like normal events have
e.stopPropagation or e.preventDefault
or simply if the event hook function returns True, it would mean the search was handled and
dotcom would not go to the default ibm.com search.

kyjak:
Adding just to be sure, the same interception should be applied to the keyboard (hitting ENTER) an any other possible submit event that launches the search.

RobertaJHahn:
The KC team is OK with scheduling in the Jan 2021 release

RobertaJHahn:
Functional spec updates are needed.

ljcarot:
Feb 15 is their launch date

PR #4913, opened 2021-01-15
feat(masthead): search no redirect

kennylam:
### Related Ticket(s)

#4599 

### Description

This adds an optional `searchNoRedirect` prop to the `Masthead` component. When used, the masthead search will no longer redirect to the default search results page. Autosuggest items will still populate, but clicking them will not trigger a redirect. Any redirect actions must be provided in the `onSearchNoRedirect` custom event.

### Changelog

**New**

- `searchNoRedirect` prop to `Masthead`
- `onSearchNoRedirect` custom event. Emits when `searchNoRedirect` is true

**Changed**

- clean up `MastheadSearch` props


ariellalgilmore (approved):
LGTM! Thanks for the documentation updates too!

kyjak:
There is a styling bug, opening search panel hides the platform. Probably because we don't use menu, just the platform.
image

Issue #4889, opened 2021-01-13
Carbon Web Component: Footer - Language selector, issues on screen reader (JAWS)


Praveen-sr:
## Detailed description

Issue:
1. There is no label
2. Screen reader announces this as a list box. Hence the user does not know that this is an Edit combo.
3. When we navigate within the combo box using up/down arrow keys, Screen reader does not announce anything. Hence Unable to select the preferred option 

In React staging, The language selector control is not identified by jaws in arrow key navigation. An issue is already opened on the same and is being tracked.

Expected:
1. Provide appropriate label
2. If this is an editable combo box, Screen reader should convey the "edit combo" role to the user 
3. Screen reader Users should be able to navigate and make the selection within the combo box.

IgnacioBecerra:
@Praveen-sr I'm able to navigate through the combo box and the item list with VoiceOver properly. I know this is a JAWS issue, but I'm under the impression it would also work with it. Would you be able to re-test this? If the issue still persists, please let me know!

For the moment I opened a PR (#4914) that addresses the `edit combo` issue, as well as adding an aria label to the clear button.

cc: @RobertaJHahn

RobertaJHahn:
Reviewed with Ram in QA meeting

Praveen-sr:
@IgnacioBecerra Original defect fixed but few regression issues found.

Now JAWS identifies the list box Role in up/down arrow navigation. But the following issues are found.
1. In up/down arrow navigation, the list box role is announced without any label. Where as, in tab/shift tab navigation, the same is announced as "Edit combo box". This confuses the user.
2. There is a clear button present after this control which is not identified in up/down arrow navigation. It is identified only in tab/shift tab navigation. And there is no label for this button as well.
3. After providing an input. Or without providing any input, when we navigate through the dropdown options using up/down arrow keys, JAWS announces the options in an Incomprehensible way. Hence the user does not understand which option /Language it is.

IgnacioBecerra:
@Praveen-sr Thank you, there is an open PR addressing the first two issues.

Could you expand upon the “incomprehensible way” when reading the dropdown options? Though I'm thinking this may be a bug with JAWS itself, due to the way it handles input fields.

[...]

PR #4914, opened 2021-01-15
fix(language-selector): added clear label attribute

IgnacioBecerra:
### Related Ticket(s)
#4889

### Description

The component now has the role of `edit combo` to ensure the screen readers convey its purpose to the users. The clear button also lacked a label to convey functionality upon click/enter. 

### Changelog

**New**

- added clear button label property


asudoh (commented):
Hi @IgnacioBecerra thank you for trying to fix this - Would you want to try fixing this directly in `carbon-web-components` unless this needs to be included in `v1.15.0`?

IgnacioBecerra:
@asudoh Yeah, I think that's a good idea. I don't think we need to get this into the release right away, and besides this is a change that should be coming from ComboBox itself.

IgnacioBecerra:
There was a button label attribute for ComboBox, so that addition can be reflected on our side.

asudoh (approved):
LGTM 👍 - Thanks @IgnacioBecerra for the update!

PR #4915, opened 2021-01-16
chore(masthead): remove a duplicate rule set

asudoh:
### Changelog

**Removed**

- A duplicate CSS rule set, that was introduced to ensure Carbon core doesn't override masthead top nav item's hover color. Instead of introducing a new CSS rule set, the new code fixes the issue by bumping up the priority of our CSS rule defining masthead top nav color.


guilhermelMoraes (approved):
LGTM! 👍🏻

PR #4923, opened 2021-01-19
chore(github-workflow): add profile host env var to various testing envs

annawen1:
### Related Ticket(s)

Change profile endpoint on Carbon Masthead for signed in state #4701

### Description

Add the profile host env var to various testing envs
previously added the variable to only react testing envs

deploy-canary:
- react-experimental
- react-rtl
- web-components
- web-components-rtl
- web-components-experimental
- web-components-react
- services

deploy-staging:
- web-components


jeffchew (approved):
LGTM!

Later on we might want to clear out the env vars that aren't really necessary for various deployments, like services and utilities. They generate JSDocs so those aren't needed. We can address this on master branch but it's not urgent as they are pretty much ignored.

Issue #4399, opened 2020-11-04
Card link: Storybook: Weird left/right margin balance for single card, need top margin


asudoh:
## Detailed description

> Describe in detail the issue you're having.

The card link demo in our Storybook has weird balance between the left/right margin for single card and no margin at the top:

![image](https://user-images.githubusercontent.com/1259051/98129599-a5f7bb80-1efc-11eb-984d-2f115bfba19e.png)

> Is this issue related to a specific component?

Yes, card.

> What did you expect to happen? What happened instead? What would you like to
> see changed?

According to design team, we let adopters put arbitrary numbers of cards with arbitrary left/right margins. From that, I'd expect:

* Center the card
* `16px` top margin

#### Additional Information
- [Box Note React Storybook study](https://ibm.ent.box.com/notes/736875354683)
- [React Card canary](http://ibmdotcom-react-canary.mybluemix.net/iframe.html?id=components-card--default)
- [React Card link canary](http://ibmdotcom-react-canary.mybluemix.net/iframe.html?id=components-cardlink--default)

ariellalgilmore:
After discussing with designers this issue is actually a web component issue where the cards should align to the left and have a top spacing. 

> Got some initial feedback from the other designers..
> we agree consistency between the two storybooks is the most important, and being aligned to the grid in some way!
> We leaned towards the React storybook with the card aligned to the left most column because:
> we don’t center align most things in Carbon/IDL
> we can be more consistent aligning left because not all components can be easily centered on the grid

PR #4924, opened 2021-01-19
fix(card): change stories to align cards center

ariellalgilmore:
### Related Ticket(s)

#4399 

### Description

Card and Card Link stories now changed so that they align to center and have a top padding.

### Changelog

**Changed**

- align items to center and `top-padding: 1rem`


asudoh (commented):
Hi @ariellalgilmore thank you for jumping in! Instead of adding inline style I suggest adding a CSS class to `.storybook/_container.scss` or a story-specific `.scss` file. Also given you seem to have found that `bx--grid` class no long suits well for the story, I suggest we remove other Carbon grid-related CSS classes (`bx--col-sm-4`, `bx--col-md-4`, `bx--col-lg-4` and `bx--no-gutter`) altogether. If still want responsive card size, we can copy https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.15.0-rc.0/packages/web-components/.storybook/_simple-grid.scss over to React storybook and use it. Here's a usage example: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.15.0-rc.0/packages/web-components/src/components/card-link/__stories__/card-link.stories.ts#L35-L37 (Where the style comes from: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.15.0-rc.0/packages/web-components/.storybook/container.scss#L77-L101)

ariellalgilmore:
Closing this PR and will create another after talking with designers.

PR #4935, opened 2021-01-20
chore(build): remove unused tool

asudoh:
### Description

`svg-result-carbon-icon.js` was used for creating `lit-html` template from Carbon core icons, but `@carbon/ibmdotcom-web-components` codebase seems to no longer use it. Removing the file.

### Changelog

**Removed**

- `packages/web-components/tools/svg-result-carbon-icon.js`.


jeffchew (approved):
Thank you @asudoh !

PR #4939, opened 2021-01-20
chore(package): update Web Components version to 1.x

asudoh:
### Description

Updates `version` field of Web Components' `package.json` to `1.x` as we release Web Components package as GA as of `v1.15.0`.

### Changelog

**Changed**

- The `version` field of `packages/web-components/package.json`.

### DIFF
--- a/packages/web-components/package.json
+++ b/packages/web-components/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@carbon/ibmdotcom-web-components",
-  "version": "1.15.0-rc.0",
+  "version": "1.0.0-rc.0",
   "description": "Carbon for IBM.com Web Components",
   "license": "Apache-2.0",
   "exports": {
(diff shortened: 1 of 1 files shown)


asudoh:
@jeffchew Do we need to update any release script for this? Thanks!

packages/web-components/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@carbon/ibmdotcom-web-components",
-  "version": "0.8.0-rc.0",
+  "version": "1.15.0-rc.0",
```
jeffchew, line 3: Thank you for updating @asudoh!

We should actually bump to `1.0.0-rc.0`, right?

asudoh: Yeah wondering which was better, will update. Thanks @jeffchew for your feedback!

jeffchew (approved):
Thank you @asudoh !

PR #4954, opened 2021-01-20
chore(fade-in-out): added percy skips

IgnacioBecerra:
### Related Ticket(s)
none

### Description
The FadeInOut component story gets the content from the DotcomShell story, but Percy is detecting changes within the FadeInOut story that may have to do with the fade animation. Since these two stories are the same, we will now be skipping these.

### Changelog

**New**

- added percy skip to the `FadeInOut` component story


IgnacioBecerra:
Closing since #4955 also adds the percy skip to Web Components

Issue #4748, opened 2020-12-18
Remove productive mixin & apply expressive tokens to Carbon for IBM.com components


oliviaflory:
#### User Story

> As a `[user role below]`:

DDS developer
> I need to:

update all Carbon for IBM.com components that use the productive tokens to use expressive theme tokens and remove the productive mixin

> so that :

components use one type scale and are more consistent for our adopters

#### Additional information
Here is a list of the components that need updating and which token they should use with the expressive theme applied.

- [x] **Leaving IBM** (bx--leaving-ibm__text) --> `$label-01`

- [x] **call out data** (bx--callout-data__source) --> `$caption-01`
- [x] **callout with media** (bx--image__caption) - uses the productive zoning mixin --> `$caption-01`
- [x] **Card** (bx--card__eyebrow  - both use productive zoning mixin --> `$label-01`
- [x] **Card** bx--card__video bx--card__footer span --> remove styling that is not being applied: 

```
.#{$prefix}--card__video {
      .#{$prefix}--card__footer span {
        color: $text-02;
        @include use-carbon-productive-tokens();
        @include carbon--type-style('body-short-01');
      }
    } 
```

- [x] **image with caption** (bx--image__caption) - uses the productive zoning mixin --> `$caption-01`
- [x] **locale modal** (bx--modal-header__label) - uses the productive zoning mixin --> `$label-01`

- [x] **content item horizontal** (bx--content-item-horizontal__item--eyebrow) - uses the productive zoning mixin --> `$label-01`
- [x] **video player** (bx--video-player__video-caption) - uses the productive zoning mixin --> `$caption-01`

- [x] **locale modal** (bx--locale-modal__search-text) - uses the productive zoning mixin --> `$heading-02`

- [x] **logo grid** (bx--content-block__cta .bx--card__heading) - uses the productive zoning mixin --> check if there is outdated code

**Note**: we will not be removing the productive mixin from the mega menu, decided in #4341 

#### Acceptance criteria
 - [x] Productive mix-in is removed from all components listed above
 - [x] no breaking changes have been introduced 
 - [x] Designer (Olivia) reviewed all changes in preview

PR #4957, opened 2021-01-21
fix(expressive): remove productive mixin from component styles and replace with expressive tokens

annawen1:
### Related Ticket(s)

Remove productive mixin & apply expressive tokens to Carbon for IBM.com components #4748

### Description

Remove productive mixin from components (leave as is in the masthead) styles and replace with expressive tokens

### Changelog

**Changed**

- swap tokens with expressive tokens

**Removed**

- unused card video footer styles
- edit the LogoGrid card props to set card heading as intended from the LogoGrid design specs (https://ibm.ent.box.com/file/661851053906)
- productive token mixin use from components other than masthead


oliviaflory:
_Adding from slack just for record_

Looks great! The only one that’s weird is the Leaving IBM in the react wrapper storybook.
“The link you requested…” text is bigger than in the Web Component. That could be a separate thing though?

Screen Shot 2021-01-22 at 5 25 16 PM

oliviaflory (approved):
Looks good!

Issue #4971, opened 2021-01-21
menu item doesn't show active state in hamburg menu in masthead


yellowdragonfly:
talked with Anna Wen on this issue. She was able to reproduce it on current testing environment.

When setting the last menu item link in the masthead to be selected, the item does not have a selected state in the mobile nav, only desktop

DESKTOP:
<img width="1297" alt="Screen Shot 2021-01-21 at 3 26 30 PM" src="https://user-images.githubusercontent.com/54281166/105408578-25904f00-5bfd-11eb-8b16-c64209812e75.png">

MOBILE MENU: (You can see the selected menu item in desktop does not have selected state in mobile)
<img width="649" alt="Screen Shot 2021-01-21 at 3 26 38 PM" src="https://user-images.githubusercontent.com/54281166/105408551-1d381400-5bfd-11eb-91a7-cfbe4abd177c.png">

RobertaJHahn:
@annawen1 Issue is ready to finish the work.  Airtable updated, labels and release added.

PR #4973, opened 2021-01-21
fix(Masthead): make sure last selected menu item has selected styles in mobile nav

annawen1:
### Related Ticket(s)

menu item doesn't show active state in hamburg menu in masthead #4971

### Description

If selected menu item is a link (not a dropdown or has megamenu panel) and is the last menu item in the nav, it currently does not get the selected styles applied on the mobile nav.

Compare correctly the titleEnglish and selected prop from adopter and set classname.

### Changelog

**Changed**

- create a `selected` const so we don't have the same comparison statement in multiple places
- add `titleEnglish` field to last menu item in the custom nav data json file so we can test on storybook

### DIFF
--- a/packages/react/src/components/Masthead/MastheadLeftNav.js
+++ b/packages/react/src/components/Masthead/MastheadLeftNav.js
@@ -1,5 +1,5 @@
 /**
- * Copyright IBM Corp. 2016, 2020
+ * Copyright IBM Corp. 2016, 2021
  *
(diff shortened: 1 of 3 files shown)


ariellalgilmore (approved):
LGTM~ thanks for the updates @annawen1!

annawen1:
Was double checking - noticed the font isn't semibold for the non-dropdown selected menu item

Issue #4940, opened 2021-01-20
[Card section carousel] Discuss how to avoid adoptors setting too large number of cards per page


asudoh:
### The problem

While `<dds-carousel>` sets the default number of cards per page that best suits the available viewport width, `<dds-carousel>` also allows our adaptors to specify arbitrary number of cards per page, via `page-size` attribute. It caused weird UI caused by extreme `page-size`, as #4918 and #4919 reports.

### The solution

This issue is to discuss how to prevent such misconfiguration. A couple of options I can rattle off are:

1. Define the maximum number of cards per page for different breakpoints, and ignore any `page-size` attribute value that exceeds that maximum number
2. Define several sets of number of cards per page (see below) and have adaptors use such set instead of having them directly set the page size

(Just example of 2., the actual numbers should be determined by design team)

| Configuration | Number of cards per page for `sm` breakpoint | `md` | `lg` |
| ------------- | -------------------------------------------: | ---: | ---: |
| `SMALL`       |                                            2 |    4 |   6  |
| `DEFAULT`     |                                            1 |    2 |   3  |
| `LARGE`       |                                            1 |    1 |   2  |

CC @jeffchew @kennylam @oliviaflory

asudoh:
@RobertaJHahn We have short-term (band-aid) fix for #4918 and #4919 already (#4921) so that January release doesn't have to wait for the resolution of this issue, FYI.

oliviaflory:
My thoughts:

The maximum number of cards or images sort of depends on how many columns of the grid the adopter is using. Adding some examples of my thinking here... I don't know if this is too complicated to achieve in storybook? 

### `16` columns used on desktop view
| Configuration | Number of Cards for `sm` breakpoint | `Md` | `Lg`–`Max`|
| --- | --- | --- | --- |
| Default | 1 | 2 | 4 |
| Large | 1 | 1 | 2 |

<img width="547" alt="Screen Shot 2021-01-21 at 10 31 32 AM" src="https://user-images.githubusercontent.com/45692622/105372812-fdd8c100-5bd3-11eb-93bc-e8af986bfb82.png">
<img width="529" alt="Screen Shot 2021-01-21 at 10 31 39 AM" src="https://user-images.githubusercontent.com/45692622/105372823-016c4800-5bd4-11eb-81ab-a3373b6556a6.png">


### `12` columns used on desktop view
[...]

PR #4974, opened 2021-01-22
docs(carousel): stories for different grid cells usage

asudoh:
### Related Ticket(s)

Refs #4940.

### Description

Adds a couple of stories, one takes up 12 Carbon grid cells (of `lg` and beyond breakpoint) at the right, and another takes up 8 Carbon grid cells at the center.

Also adds knobs to select alternate number of cards per page, for those stories/breakpoints combinations.

### Changelog

**New**

- Stories for `<dds-carousel>`, one takes up 12 Carbon grid cells (of `lg` and beyond breakpoint) at the right, and another takes up 8 Carbon grid cells at the center.
- Knobs to select alternate number of cards per page.

### DIFF
--- a/packages/web-components/src/components/carousel/__stories__/README.stories.mdx
+++ b/packages/web-components/src/components/carousel/__stories__/README.stories.mdx
@@ -89,6 +89,27 @@ import '@carbon/ibmdotcom-web-components/es/components/content-section/content-s
 </dds-carousel>
 ```
 
+### SCSS
+
+While `<dds-carousel>` has `page-size` attribute, `<dds-carousel>` also looks at `--dds--carousel--page-size` CSS custom property to determine the number of cards per page.
+This is useful for setting different number of cards for different breakpoints.
+
+```scss
+@import 'carbon-components/scss/globals/scss/vendor/@carbon/elements/scss/layout/breakpoint';
+
+dds-carousel {
+  --dds--carousel--page-size: 1;
+
+  @include carbon--breakpoint(md) {
+    --dds--carousel--page-size: 2;
+  }
+
+  @include carbon--breakpoint(lg) {
+    --dds--carousel--page-size: 3;
+  }
+}
+```
+
 ## Props
 
 <Props of="dds-carousel" />
--- a/packages/web-components/src/components/carousel/__stories__/README.stories.react.mdx
+++ b/packages/web-components/src/components/carousel/__stories__/README.stories.react.mdx
@@ -48,6 +48,27 @@ function App() {
 }
 ```
 
+### SCSS
+
+While `<dds-carousel>` has `page-size` attribute, `<dds-carousel>` also looks at `--dds--carousel--page-size` CSS custom property to determine the number of cards per page.
+This is useful for setting different number of cards for different breakpoints.
+
+```scss
+@import 'carbon-components/scss/globals/scss/vendor/@carbon/elements/scss/layout/breakpoint';
+
+dds-carousel {
+  --dds--carousel--page-size: 1;
+
+  @include carbon--breakpoint(md) {
+    --dds--carousel--page-size: 2;
+  }
+
+  @include carbon--breakpoint(lg) {
+    --dds--carousel--page-size: 3;
+  }
+}
+```
+
 ## Props
 
 <Props of={PropTypesRef} />
--- a/packages/web-components/src/components/carousel/__stories__/carousel.stories.react.tsx
+++ b/packages/web-components/src/components/carousel/__stories__/carousel.stories.react.tsx
@@ -7,6 +7,8 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+import { select } from '@storybook/addon-knobs';
+import classnames from 'classnames';
 import React from 'react';
 import ArrowRight20 from '@carbon/icons-react/es/arrow--right/20.js';
 // Below path will be there when an application installs `@carbon/ibmdotcom-web-components` package.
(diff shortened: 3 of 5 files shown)


kennylam (approved):
Looks good thanks @asudoh!

oliviaflory:
@asudoh these are looking really great and align with what we discussed.

A few comments / questions for you below:

1. On the current implementation of the Carousel component the user can see a hint of the next card or page at the far right of the grid, but in the 16 and 12 column examples we don't see that anymore. 

_Current_
Screen Shot 2021-01-26 at 4 33 51 PM

_16 column_
Screen Shot 2021-01-26 at 4 33 01 PM

_12 column_
Screen Shot 2021-01-26 at 4 33 12 PM


2. This goes against the logic in the point above, but could we hide the hint of the next card / page in the 8 column layout? It looks pretty weird like the carousel is broken or something because there is space to show another card, but I think the goal when I suggested was more to align with the center 8 columns only. I realize this might be complicated so feel free to tell me that's not possible.

Screen Shot 2021-01-26 at 4 34 20 PM


3. On the small breakpoint, the controls push up against the browser edge, and are not contained within the grid anymore

_PR example_
Screen Shot 2021-01-26 at 4 35 15 PM


_Design intent_
Screen Shot 2021-01-26 at 4 43 48 PM

asudoh:
@oliviaflory Thank you for reviewing! Made updates upon your comments.

As I tried to align the card group on Carbon grid gutter line upon your comment, one question I have is if each card should also be aligned to Carbon grid gutter line. Here's what I have as of now (I deliberately add border to cards so we can easily see them):

image

Any thoughts...? Thanks!

oliviaflory:
@asudoh thanks for sharing the screen shot and adding borders, that is super helpful.

Referencing the original specs Jacob created for the card-section-carousel, he references what Carbon calls the ["Narrow grid" mode](https://www.carbondesignsystem.com/guidelines/2x-grid/implementation#grid-modes) (example below in the teal color)

Screen Shot 2021-01-27 at 9 13 39 AM


The car should overhang on the left side 16px, so the text aligns to the columns, then the right side of the card aligns to the 4th column with a 16px gap between the end of one card and the start of another. 

I tried to show this in sketch, adding borders to the cards so we can see them better, the solid red line is the Carbon grid gutter line, and the dotted red line shows the text aligned to the Carbon grid columns.

Screen Shot 2021-01-27 at 9 25 19 AM


I am not very familiar with how the grid works, and if there really are different "grid modes" as specified on the Carbon website. So if this is not the case, we may want to re-consider how the cards are aligned to the grid and whether we can achieve the hint of the 5th card on the far right side.

asudoh:
Thanks @oliviaflory for more details on the design intent! Changed carousel style to align to the desired grid line above.

I have a follow-up question for smaller breakpoint; Carbon grid removes left/right margins for smaller breakpoints, like this (the dark blue is grid gutter):

image

Wondering if it means we should do below for below breakpoint. where the left edge of card is aligned to one of browser viewport, whereas the right edge has some space (that's what the latest deploy preview here works):

image

Any thoughts...? Thanks!

wonilsuhibm:
@asudoh @oliviaflory That empty margin on the right side is a bit awkward. @asudoh is that because we are fitting the cards into grid? Design wise, I think it makes sense for the card to be full width.

One question for @oliviaflory when the user swipes, should there be a gap between cards or should they touch each other?

oliviaflory:
@wonilsuhibm I don't believe there is a "swiping" ability on the cards, the user has to navigate with the controls, at least that's how it seemed to work in inspect. 

The desktop cards have space between them, and I am unclear if we can switch to 'touching' cards at the small breakpoint, I thought that was a different "grid mode". I think it would work to have them touch in small breakpoint, or have an even gap.

asudoh:
> @asudoh is that because we are fitting the cards into grid?

@wonilsuhibm Thank you for asking, it's yes. Having an even gap in smaller breakpoint that @oliviaflory is one option I have had in my mind, though it effectively switches "Carbon grid mode" in smaller breakpoint, it looks more natural.

And @oliviaflory is right that we don't have support for swiping gesture, at least as of today. I naïvely felt that swiping gesture is more natural for touch-screen devices, regardless of screen size - Even with iPad. But I think it's a design call.

asudoh:
@wonilsuhibm @oliviaflory I just brought up the left padding _only_ in the smallest breakpoint. Good to hear what you think. Thanks!

wonilsuhibm:
@asudoh @oliviaflory this looks great. A few thoughts and suggestions:

1. Gutter
This looks for now, I'm hoping that the margin on either side will somewhat help the user understand that this is a carousel. Let's go with this for now? @oliviaflory 

2. Motion
@asudoh what are the motion specs for the carousels? It feels a bit slow. I got these parameters from the [motion generator](https://ibm.github.io/motion/)

```
# expressive version
Duration: 270ms
Easing: cubic-bezier(0.4, 0.14, 0.3, 1)
```

```
# productive version
Duration: 191ms
Easing: cubic-bezier(0.2, 0.2, 0.38, 0.9)
```

3. We should add the swipe support for a future iteration.

asudoh:
@wonilsuhibm Thank you for your feedback!

1. Thanks - @oliviaflory good to see your thoughts, too.
2. It's [`motion(standard, productive)` (`cubic-bezier(0.2, 0, 0.38, 0.9)`)](https://github.com/carbon-design-system/carbon/blob/v10.27.0/packages/motion/scss/motion.scss#L14) with [`$duration--slow-01` duration (`400ms`)](https://github.com/carbon-design-system/carbon/blob/v10.27.0/packages/components/src/globals/scss/_motion.scss#L70). Let me know if any other choice of tokens is desirable.
3. Agreed.

oliviaflory (approved):
Looks great, thanks Akira

oliviaflory:
I may have approved before seeing @wonilsuhibm motion request 🙈 

Do we need to add another issue to cover motion changes @asudoh ?

asudoh:
> Do we need to add another issue to cover motion changes @asudoh ?

@oliviaflory Yes, if the issue includes the specific motion token we want, it'll be great. Thank you!

Issue #4976, opened 2021-01-22
Web component: Leadspace - Open canvas in new tab, gives an error page for the user


Praveen-sr:
## Detailed description

Issue: 
Leadspace - On knobs, edit/change 'number of buttons' or 'Icon' and then Open canvas in new tab, gives an error page for the user

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/0e60914f-f718-4cf7-833e-c22ec5260d1b)

kennylam:
@Praveen-sr @RobertaJHahn This is ready for testing http://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-leadspace--default-with-image

PR #4986, opened 2021-01-22
chore(storybook): add file-loader for static images

kennylam:
### Related Ticket(s)

#4976 

### Description

This changes the `web-component` webpack to use `file-loader` instead of `url-loader`. The latter base64 encodes images, which interfere with Storybook when used in knobs.

### Changelog

**Changed**

- Web components webpack config now uses `file-loader`


jeffchew (approved):
Thank you for this fix!

PR #4997, opened 2021-01-22
chore(ci-check): extending ci-check to include lint and stylelint

jeffchew:
### Related Ticket(s)

No related issue

### Description

Not sure if this was a decision made awhile ago to not include, but I
think ci-checks should include full lint and stylelint so these can be
caught early on rather than later when these checks are made during
preflight.

### Changelog

**Changed**

- Expanded ci-check

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
   },
   "scripts": {
     "build": "lerna run build --stream --prefix --npm-client yarn",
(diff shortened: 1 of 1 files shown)


asudoh (approved):
LGTM 👍, a stretch-goal is setting up `lint` tasks in each package and having `lerna` run those from here. Thanks @jeffchew!

Issue #4993, opened 2021-01-22
Web Components: Document Optimizing layout query


RobertaJHahn:
### The problem

Querying for browser layout in `@carbon/ibmdotcom-web-components` has some new technical approaches in its implementation details.

### The solution

Document them so new developers can easily jump into the codebase.

PR #5013, opened 2021-01-27
docs(coding-convention): note on optimizing layout query

asudoh:
### Related Ticket(s)

Refs #4993.

### Changelog

**New**

- Implementation notes for optimizing layout query, so that new developers can do the same optimization and won't cause performance issues.


packages/web-components/docs/coding-conventions.md
```diff
@@ -568,3 +569,13 @@ But there are certain kind of components that have to work with complex data, wh
...
+
+`@carbon/ibmdotcom-web-components` codebase uses [`IntersectionObserver`](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) and [`ResizeObserver`](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) as new best practices. Those APIs leverages browser's layout information it calculates at the right timing so that application doesn't need to run another query that may not be at the right timing.
+
+Make sure not missing [lifecycle management](#lifecycle-management). For example, we don't want to keep observing intersection of DOM elements that are no longer there inside `<body>`, like ones that are removed as routing happens. An example of lifecycle management code can be found at [here](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.15.0-rc.1/packages/web-components/src/components/masthead/top-nav.ts#L109-L136), which is called from `connectedCallback()` and `disconnectedCallback()`.
```
jeffchew, line 579:
```suggestion
Be mindful of [lifecycle management](#lifecycle-management). For example, we don't want to keep observing intersection of DOM elements that are no longer there inside `<body>`, like ones that are removed as routing happens. An example of lifecycle management code can be found at [here](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.15.0/packages/web-components/src/components/masthead/top-nav.ts#L109-L136), which is called from `connectedCallback()` and `disconnectedCallback()`.
```

jeffchew: Debating if this should point to the lines of code in master? Or keep it on the version?

kennylam: Good catch @jeffchew. But eventually maybe we rewrite this so it doesn't point to a version at all, just latest.

kennylam (approved):
Looks good thanks @asudoh!

PR #5016, opened 2021-01-27
feat(megamenu): megamenu web component HC & AI updates

annawen1:
### Related Ticket(s)

Web Component: Change Mega Menu to accommodate HC + AI updates component #4641

### Description

*** NOTE: The React mobile version looks broken, another PR coming to address that so I would recommend waiting for that PR before we merge this one ****

Megamenu Web Components Hybrid Cloud and AI updates

Screen Shot 2021-01-29 at 11 46 45 AM

Screen Shot 2021-01-27 at 2 04 18 PM

Screen Shot 2021-01-27 at 2 04 25 PM

Also created "With Custom Data" separate story for masthead so we can view how the design updates look with the expected data.

### Changelog

**New**

- `dds-left-nav-menu-category-heading` for the heading text (ie. The essentials)
- `dds-left-nav-menu-highlighted` and `dds-left-nav-menu-item-highlighted` for highlighted mobile menu items so we can target when to add the darker bottom border
- `dds-megamenu-category-group-copy` for the heading and description text in the megamenu


annawen1:
@jyoon25 you can test the megamenu updates here: https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/5016/index.html?path=/story/components-masthead--with-custom-data

packages/web-components/src/components/masthead/left-nav-menu-highlighted.ts
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Masthead left nav submenu item.
+ *
+ * @element dds-left-nav-menu-item
+ */
+@customElement(`${ddsPrefix}-left-nav-menu-highlighted`)
+class DDSLeftNavMenuHighlighted extends DDSLeftNavMenu {
+  static styles = styles; // `styles` here is a `CSSResult` generated by custom WebPack loader
+}
```
asudoh, line 25: My knee-jerk reaction here was adding prop for highlighted state rather than a new component, but looking at `<dds-masthead-composite>` code, it seems something different from highlighted state for the nav? Do you have any details?

asudoh (commented):
Thank you for coming up with this @annawen1!

packages/web-components/src/components/masthead/left-nav-menu-highlighted.ts
annawen1: Yeah I tried adding a `highlighted` attribute to the `dds-left-nav-menu` component but styling wise for the mobile menu I need to target the last highlighted nav menu to set the darker border and 
`:host(#{dds-prefix}-left-nav-menu)[highlighted]:last-of-type` didn't work as the `:last-of-type` or `:last-child` selector does not take attributes into account

asudoh: Thank you for elaborating @annawen1! Just in case - Did you try `:host(#{dds-prefix}-left-nav-menu[highlighted]):last-of-type`? I think there are other CSS rules whose selector matches host element with certain attributes, and I suggest to look at those.

annawen1: @asudoh yes that works! I didn't think to wrap the attribute within the `:host()` 😅. This is great, I'll remove the unnecessary components. Thanks!

annawen1: @asudoh I might have spoken too quickly...I tested the above by just setting the attribute `highlighted` without any conditionals `<dds-left-nav-menu highlighted .... >` and it worked. But as soon as I add a conditional whether to render the attribute `<dds-left-nav-menu-item ?highlighted=${ifNonNull(item.highlighted)} ...>` then I don't see the styles at applied at all. I tried setting a class depending on whether it is highlighted or not but that doesn't seem to work either

asudoh: Thank you for the update @annawen1 - One quick comment is we don't need `ifNonNull()` for boolean attributes, because `lit-html` automatically unsets the attribute if the value is falsy.

For the issue you described, would you inspect the DOM and see what CSS ruleset isn't applied while it should? If you cannot figure it out, probably you can push the code you tried (but it doesn't work)?

annawen1: @asudoh sure, I can try pushing the code.

When I inspect, I am expecting to see this CSS applied to the last `left-nav-menu[highlighted]`element
```
  :host(#{$dds-prefix}-left-nav-menu[highlighted]):last-of-type {
    border-bottom: 1px solid $ui-04;
  }
```
Screen Shot 2021-01-29 at 8 53 48 AM

asudoh: @annawen1 Thank you for elaborating! Seems that the CSS rule works if I add `highlighted` attribute to the `<dds-left-nav-menu>` for "Explore more" at the top-level. One thing I imagined the menu you brought up is not really a `:last-of-type` - As long as what the DOM inspector shows.

annawen1: @asudoh Right it works when the attribute is added to the last `nav-menu` of the section, if I try to add it to "Learn & Support" it doesn't work. What I'm trying to do is target the last `dds-left-nav-menu` with the `highlighted` attribute.

asudoh: I see - BTW what does "highlighted" mean? I'm asking this because adding `highlighted` attribute doesn't actually seem to make the menu highlighted. Any context...?

annawen1: The `highlighted` is to note that the menu item is part of the "highlighted section" - the left nav section in desktop and with the new designs, the highlighted items have a distinct border to separate between the highlighted menu items). 

Screen Shot 2021-01-27 at 2 04 04 PM

Screen Shot 2021-01-27 at 2 04 25 PM

https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/5016/iframe.html?id=components-masthead--with-custom-data

asudoh: @annawen1 I see thanks for clarifying! Given you brought up "highlighted section" words, my suggestion is creating a component for such "highlighted section".

annawen1: @asudoh yeah that makes sense - I'll try that out, thanks!

annawen1: So I tried created a component (`dds-left-nav-highlighted`) to wrap the `dds-left-nav-menu` items into the "highlighted section" but ran into the same issue with the `:last-of-type` or `:last-child` selector. I still need to target the last `dds-left-nav-menu` item to remove its own border.

```
 :host(#{$dds-prefix}-left-nav-highlighted) {
    display: block;
    border-bottom: 1px solid $ui-04;

    ::slotted(#{$dds-prefix}-left-nav-menu:last-of-type) .#{$prefix}--side-nav__submenu::after {
      content: none;
    }
  }
```

Doesn't seem like I'm able to target the last `dds-left-nav-menu` within the hightlighted section here either. 

You can see the border of the last menu item is still showing here:
Screen Shot 2021-01-29 at 1 28 27 PM

annawen1: I ended up just setting the `last-highlighted` attribute to the last highlighted menu item of the section and styling with that instead. That avoids creating new components and doesn't require us to use the `:last-of-type` selector!

asudoh: @annawen1 Thank you for explaining what you stumbled at! My suggestion is adding `::after` pseudo element to `<dds-left-nav-highlighted>` with something like below, so the border there is placed on top of (in z-axis) the bottom border of the last `<dds-left-nav-menu>`:

```css
:host(#{$dds-prefix}-left-nav-highlighted) {
  position: relative;

  /* (Some other rules) */

  &::after {
    content: '';
    position: absolute;
    bottom: 0;
    border-bottom: 1px solid $ui-04;
    width: 100%;
  }
}
```

annawen1: @asudoh so I also attempted to create a wrapper container in React and another issue I found is that because the submenus structurally are within each other, if I add a border to the wrapper container and then click into "Hybrid Cloud" submenu, the border remains for the next menu section which is something we don't want. It seems to be more trouble to set when to hide the border in this case

asudoh: Thank you for taking deeper look @annawen1! I see that `position:relative` at the wrapper container changes what `min-height:100%` at `<ul class="bx--side-nav__menu">` in `<dds-left-nav>`'s shadow DOM refers to, that breaks its feature of "covering up parent menu items". So I think we'd go with your solution of adding attribute to the last item in the "highlighted" group.

ariellalgilmore (approved):
LGTM :)! Awesome jon @annawen1!

jyoon25:
@annawen1  Here's design QA for megamenu. Let me know if you have any comments. Thanks!
QA pdf link: https://ibm.box.com/s/dum8331jjkseqibworat5obgitb6u57p
Original design abstract link: https://share.goabstract.com/6f02a9ae-517f-45b8-b76f-b77273b9a4ad

image

Issue #4810, opened 2021-01-05
React:  Card section - images: image height of the last card


wonilsuhibm:
## Detailed description

> Describe in detail the issue you're having.

In both React and WC component storybook pages, the image on the last card is 1px taller. (see the image below)

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

It's a bug.

> Is this issue related to a specific component?

Yes—`CardSectionImages`

> What did you expect to happen? What happened instead? What would you like to
> see changed?

All images should have the same size.

> What browser are you working in?

Chrome Version 87.0.4280.88 (Official Build) (x86_64)

> What version of Carbon for IBM.com are you using?

The latest production Carbon for IBM.com

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

N/A

## Steps to reproduce the issue

1. Open these two pages:
https://ibmdotcom-web-components.mybluemix.net/?path=/story/components-card-section-images--default
https://ibmdotcom-react.mybluemix.net/?path=/story/components-cardsectionimages--default

2. Compare the height of the image on the last card to the one next 

## Additional information

- Screenshots or code
- Notes

PR #5017, opened 2021-01-27
fix(card-group): Image height of the last card

NSDrowned:
### Related Ticket(s)

React: Card section - images: image height of the last card #4810

### Description

The image on the last card is 1px taller. This happens because the right padding is removed on the last card to hide the dark border. The missing padding allows the image to be 1px taller.

### Changelog

**Changed**

- `_card-group.scss to fix the height


asudoh (commented):
Hi @NSDrowned congrats to (if I'm right) your first PR! Honestly I wish there is a way not having to tweak the card size, but I want to defer to other devs.

NSDrowned:
Thank you (yes, my first) and I agree @asudoh, it felt kind of hacky. All the other solutions I came up with required modifying the HTML structure, but let me try to figure out a different (CSS only) solution.

NSDrowned:
Pushed a new CSS only approach (No more % height calculation). I added the padding that was removed from the last card to another spot of the card so that there won't be any missing padding that will affect the image height. This will only apply to the last card.

asudoh (approved):
LGTM 👍 - Thanks @NSDrowned for the update!

kennylam (approved):
LGTM thanks @NSDrowned!

Issue #4740, opened 2020-12-17
Web component: Content block cards - Link in each card - Screen reader announces a lengthy text which makes it over verbose.


Praveen-sr:
## Detailed description
On JAWS FF/Voiceover iPhone

Issue:
In tab / shift tab navigation, when focus reaches the link in each card, Screen reader announces a lengthy text which makes it over verbose.

Expected: 
It may be possible if the entire content in the card is given in anchor tag. In this scenario, only the heading can be given in anchor tag. 
Note: the link text/label should be concise and convey only the relevant information.

PR #5019, opened 2021-01-28
fix(card): add aria-hide to card-copy

IgnacioBecerra:
### Related Ticket(s)
#4740 

### Description

Follow-up to #4870 to also hide the `card-copy` text from screen readers to avoid unnecessary long description.

### Changelog

**New**

- added `aria-hidden='true'` to `card-copy` div


kennylam (approved):
Looks good thanks @IgnacioBecerra!

Issue #4799, opened 2021-01-04
Web component: Carousel - User navigates with tab/shift tab keys, focus on additional cards makes the keyboard navigation disorients


Praveen-sr:
## Detailed description

Issue: 
When additional cards are displayed on focus while the user navigates with tab/shift tab keys, it makes the keyboard navigation tedious and also disorients the Blind and low-vision users because it is not expected to trigger any action on tab/shift tab navigation

Expected Result:
Tab Focus should go to all the visible cards since all these are interactive elements. And the next set of cards should be displayed upon activation of previous or next button

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/1d0efbd2-cd72-4205-9a96-899d1bdb758b)

PR #5035, opened 2021-01-28
fix(carousel): improved tab page navigation

IgnacioBecerra:
### Related Ticket(s)
#4799

### Description

Carousel now moves pages automatically when using tab navigation and focusing on the next card in a different page. Works going forward or backwards.

### Changelog

**New**

- pages now change dynamically depending on the focused card index


packages/web-components/src/components/carousel/carousel.ts
```diff
@@ -111,6 +114,39 @@ class DDSCarousel extends LitElement {
...
+      // going forwards, change page if focused card is the first on the next page
+      if (oldChildIndex < currentChildIndex && currentChildIndex !== 0 && targetPageIndex === 0) {
+        this._handleClickNextButton();
+
+        // going backwards, change page if focused card is the last on the previous page
+      } else if (currentChildIndex < oldChildIndex && targetPageIndex === this.pageSize - 1) {
+        this._handleClickPrevButton();
+      }
+    }
+  };
```
asudoh, line 148: I expect the logic here to be simply updating `start` if the newly focused card is out of the page range, instead of detecting "hit the boundary" and using the event handler for prev/next button. Would you be able to try?

asudoh (commented):

packages/web-components/src/components/carousel/carousel.ts
IgnacioBecerra: That's what it does; if the newly focused card is out of range, it goes to the next/prev page. The handlers are doing the `start` updating logic themselves, so I think it'd be redundant updating `start` on its own.

IgnacioBecerra: Wait nevermind, I think I see what you mean, let me give it a try

packages/web-components/src/components/carousel/carousel.ts
```diff
@@ -111,6 +114,39 @@ class DDSCarousel extends LitElement {
...
+    const targetPageIndex = currentChildIndex % this.pageSize;
+
+    // reset to first page if tabbing back into the carousel after previously moving pages
+    if (currentContains && currentChildIndex === 0) {
+      this.start = 0;
+    }
+
+    if (currentContains && oldContains) {
+      // going forwards, change page if focused card is the first on the next page
+      if (targetPageIndex === 0) {
```
asudoh, line 140: This seems still "on the edge" detection, given it seems to detect if "focus is on first card on the page", right? Can we detect if the index of the focused card of out of the page range (larger than equal to `start` and less than `start + pageSize`) instead?

asudoh (commented):
Thank you for trying to update @IgnacioBecerra!

packages/web-components/src/components/carousel/carousel.ts
```diff
@@ -111,6 +114,34 @@ class DDSCarousel extends LitElement {
...
+    }
+
+    if (currentContains) {
+      // going forwards, change page depending on card index
+      if (currentCardIndex >= this.start + this.pageSize) {
+        this.start = Math.min(currentCardIndex, this._total + 1 - this.pageSize);
+
+        // going backwards, change page depending on card index
+      } else if (currentCardIndex < this.start) {
+        this.start = currentCardIndex + 1 - this.pageSize;
```
asudoh, line 140:
```suggestion
        this.start = Math.max(currentCardIndex + 1 - this.pageSize, 0);
```

Also, this logic conflicts with your logic for "previous focus was outside carousel and now on carousel" scenario. I suggest using `event.relatedTarget` to detect that and do things appropriately here.

asudoh (commented):

packages/web-components/src/components/carousel/carousel.ts
```diff
@@ -111,6 +114,34 @@ class DDSCarousel extends LitElement {
...
+
+    // reset to first page if tabbing back into the carousel after previously moving pages
+    if (currentContains && currentCardIndex === 0) {
+      this.start = 0;
+    }
+
+    if (currentContains) {
+      // going forwards, change page depending on card index
+      if (currentCardIndex >= this.start + this.pageSize) {
+        this.start = Math.min(currentCardIndex, this._total + 1 - this.pageSize);
```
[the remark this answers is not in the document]
IgnacioBecerra: I think both of our approaches are off by one in different cases. Using the suggested calculation and focusing on the 5th (index `4`) would make the page start with that same card, instead of the 4th (index `3`) as it should. This is because we're only focusing on the last page scenario (subtracting from `this.total`).

Using `currentCardIndex - (currentCardIndex % this.pageSize)` will always ensure getting the right start index for the current card's page.

packages/web-components/src/components/carousel/carousel.ts
```diff
@@ -111,6 +114,35 @@ class DDSCarousel extends LitElement {
...
+  @HostListener('shadowRoot:focusin')
+  // @ts-ignore: The decorator refers to this method but TS thinks this method is not referred to
+  private _handleFocus = async ({ target, relatedTarget }: FocusEvent) => {
+    const currentContains = target !== this && this.contains(target as DDSCard);
+    const oldContains = target !== this && !this.contains(relatedTarget as DDSCard);
+    const currentCardIndex = Array.from(this.children).indexOf(target as HTMLElement);
+
+    // reset to first page if tabbing back into the carousel after previously moving pages
+    if (currentContains && oldContains) {
+      this.start = 0;
```
asudoh, line 131: 1. Did you really want to run the next `if ()` block in this condition, or did you want to bail out from `_handleFocus()` in this case?
2. What should happen in back-tab scenario? (e.g. When the focus was on the previous page button (pagination button with left caret icon) and user hits <kbd>Shift</kbd>+<kbd>Tab</kbd> key)?
3. If user was previously on the non-first page and then reaches to carousel in tab key, what should happen? (e.g. User goes to the 3rd page by clicking on next page button (pagination button with right caret icon) two times, clicks on the blank area above the carousel and then hit tab key)

packages/web-components/src/components/carousel/carousel.ts
```diff
@@ -111,6 +114,35 @@ class DDSCarousel extends LitElement {
...
+  /**
+   * Handles card focus throughout pages.
+   *
+   * @param event The event.
+   */
+  @HostListener('shadowRoot:focusin')
+  // @ts-ignore: The decorator refers to this method but TS thinks this method is not referred to
+  private _handleFocus = async ({ target, relatedTarget }: FocusEvent) => {
+    const currentContains = target !== this && this.contains(target as DDSCard);
+    const oldContains = target !== this && !this.contains(relatedTarget as DDSCard);
```
asudoh, line 126: Did you mean `currentNotContains`?

asudoh (commented):
Thank you for another update @IgnacioBecerra!

packages/web-components/src/components/carousel/carousel.ts
```diff
@@ -111,6 +114,36 @@ class DDSCarousel extends LitElement {
...
+    // keep current page if tabbing back into the carousel after previously moving pages
+    if (currentContains && oldNotContains && currentCardIndex === 0) {
+      (this.children[this.start] as HTMLElement).focus();
+      return;
+    }
+
+    if (currentContains) {
+      // going forwards, change page depending on card index
+      if (currentCardIndex >= this.start + this.pageSize) {
+        this.start = currentCardIndex - (currentCardIndex % this.pageSize);
```
asudoh, line 138: Considering the following condition:

* We have many cards in carousel
* `this.start` previously was `2`
* Screen size as large as putting 4 cards in a page

And user keeps hitting tab key to reach to the code here. What should happen?

asudoh (commented):

packages/web-components/src/components/carousel/carousel.ts
IgnacioBecerra: @asudo I see, wasn't aware the `this.start` could be set by the user. Added an additional calculation where it considers the offset to calculate the right index if `this.start` isn't a proper multiple of `this.pageSize`.

Now in the condition you provided `this.start` will become the index `6` instead of `4` as it was before i.e. the next page will start with the proper card.

packages/web-components/src/components/carousel/carousel.ts
```diff
@@ -111,6 +114,39 @@ class DDSCarousel extends LitElement {
...
+      return;
+    }
+
+    if (currentContains) {
+      // going forwards, change page depending on card index
+      if (currentCardIndex >= this.start + this.pageSize) {
+        const nextStart = currentCardIndex - (currentCardIndex % this.pageSize);
+        const pageOffset = this.start % this.pageSize;
+
+        this.start = nextStart + pageOffset;
```
asudoh, line 141: Probably you meant `this.start = currentCardIndex`? (Though there are some edge cases to consider)

asudoh (commented):
Thanks for the update @IgnacioBecerra!

packages/web-components/src/components/carousel/carousel.ts
IgnacioBecerra: `nextStart` calculates the correct `this.start` if the user focuses on a random card (and show its containing page)
`pageOffset` adds any remainder to the calculation if needed, usually it's just 0 (adds `2` in the previous scenario) 

`this.start = currentCardIndex` only works if we tab forward but doesn't address the "random card focus" case.

packages/web-components/src/components/carousel/carousel.ts
```diff
@@ -111,6 +114,39 @@ class DDSCarousel extends LitElement {
...
+    // keep current page if tabbing back into the carousel after previously moving pages
+    if (currentContains && oldNotContains && currentCardIndex === 0) {
+      (this.children[this.start] as HTMLElement).focus();
+      return;
+    }
+
+    if (currentContains) {
+      // going forwards, change page depending on card index
+      if (currentCardIndex >= this.start + this.pageSize) {
+        const nextStart = currentCardIndex - (currentCardIndex % this.pageSize);
```
asudoh, line 138:
```suggestion
        // The `currentIndex` floored by `pageSize`
        const nextStart = currentCardIndex - (currentCardIndex % this.pageSize);
```

Or you can do: `const nextStart = Math.floor(currentCardIndex / this.pageSize) * this.pageSize`

packages/web-components/src/components/carousel/carousel.ts
asudoh, line 141:
```suggestion
        // Ensures the page moves by `this.pageSize`
        this.start = nextStart + pageOffset;
```

packages/web-components/src/components/carousel/carousel.ts
```diff
@@ -111,6 +114,39 @@ class DDSCarousel extends LitElement {
...
+  private _handleFocus = async ({ target, relatedTarget }: FocusEvent) => {
+    const currentContains = target !== this && this.contains(target as DDSCard);
+    const oldNotContains = target !== this && !this.contains(relatedTarget as DDSCard);
+    const currentCardIndex = Array.from(this.children).indexOf(target as HTMLElement);
+
+    // keep current page if tabbing back into the carousel after previously moving pages
+    if (currentContains && oldNotContains && currentCardIndex === 0) {
+      (this.children[this.start] as HTMLElement).focus();
+      return;
+    }
```
asudoh, line 133: Do we still need this logic?

packages/web-components/src/components/carousel/carousel.ts
```diff
@@ -111,6 +114,39 @@ class DDSCarousel extends LitElement {
...
+      // going forwards, change page depending on card index
+      if (currentCardIndex >= this.start + this.pageSize) {
+        const nextStart = currentCardIndex - (currentCardIndex % this.pageSize);
+        const pageOffset = this.start % this.pageSize;
+
+        this.start = nextStart + pageOffset;
+
+        // going backwards, change page depending on card index
+      } else if (currentCardIndex < this.start) {
+        this.start = Math.max(currentCardIndex + 1 - this.pageSize, 0);
```
asudoh, line 145: Do you want to apply your improvement to "page forward" logic to the "page backward" logic here, too?

asudoh (commented):
Thank you for clarifying @IgnacioBecerra!

kennylam (commented):
@IgnacioBecerra This is looking really good. One thing I'm seeing in the `center 8 columns` story is when advancing, the pagination happens first – another tab is needed to move focus to the next item. Going back in pages works fine.

Feb-08-2021 08-53-07

packages/web-components/src/components/carousel/carousel.ts
IgnacioBecerra: @asudoh Yes, we still need that logic to focus on the current starting card after having tabbed outside the carrousel. Without it, the focus will always go to the first card automatically, out of view.

IgnacioBecerra:
@kennylam Thanks for the comment! After investigating, I found the issue is caused by `--dds--carousel--overflow-hint-size: 0rem;`, which places the next card out of the viewport. This causes the browser to make the element scroll into view when focusing on the next (hidden) element, which messes with the pagination view. I've messaged Akira to ask if he knows a workaround for this using `HostListener`.

packages/web-components/src/components/carousel/carousel.ts
IgnacioBecerra: Since we'd use the same code going in either direction, the previous logic runs in both directions whenever the focus is out of the current page.

packages/web-components/src/components/carousel/carousel.ts
asudoh: I see, thank you for clarifying - I suggest preserving the "focused card index" in `_handleFocus()`. It can be done by "old contains but current node contains" detection and preserving `event.relatedTarget`. In this way we can set `this.start` according to the index of such "previously focused card".

asudoh:
@kennylam Good catch!

@IgnacioBecerra This is caused by `<div class="bx--carousel__scroll-container">` scrolls upon the "scrolling into view" effect as you described. Though it's a bit dirty, looks that we can fix that behavior by handling `onscroll` event on that `<div>` and calling `event.target.scrollTo(0, 0)` there.

packages/web-components/src/components/carousel/carousel.ts
```diff
@@ -107,10 +119,57 @@ class DDSCarousel extends LitElement {
         // @ts-ignore
         this._observerResizeContainer = new ResizeObserver(this._observeResizeContainer);
         this._observerResizeContainer.observe(contentsNode);
+        containerNode?.addEventListener('scroll', this._handleScrollFocus);
```
asudoh, line 122: Could you attach this event listener with `lit-html` template syntax? In this way we don't need to explicitly remove the event listener.

packages/web-components/src/components/carousel/carousel.ts
```diff
@@ -107,10 +119,57 @@ class DDSCarousel extends LitElement {
...
+      // focus coming from the top
+      if (currentCardIndex === 0) {
+        (this.children[this.start] as HTMLElement).focus();
+
+        // focus coming from bottom
+      } else if (currentCardIndex === this._total - 1) {
+        (this.children[this.start + this.pageSize - 1] as HTMLElement).focus();
+      }
+      return;
+    }
```
asudoh, line 160: Could you elaborate why you think we still need this logic here? I thought the first card automatically gets focus in forward-tab scenario, and last card automatically gets on focus in backward-tab scenario. If it's not the case, could you point out what element gets focus instead in such scenarios?

asudoh (commented):
Thank you for the update @IgnacioBecerra and the discussion offline!

packages/web-components/src/components/carousel/carousel.ts
IgnacioBecerra: Sure, we still need this logic because the page will not automatically focus on the first/last page. The focus actually either goes to the very first card (index 0), or to the very last one, so even if the current page is the right one, the focus will be entirely out of view, and the user would need to keep tabbing/shift+tabbing until it reaches the current page if we don't have the `focus()` here.

asudoh: Thank you for clarifying. My thought is tabbing on the very first card, instead of the first card on the current page, in forward-tab scenario. Similar for backward-scenario. Could you work with the designer to double-check that? Thanks!

IgnacioBecerra: @asudoh Discussed with @photodow and confirmed that focusing back on the current page (as it is currently) works best.

asudoh: Thanks @IgnacioBecerra - Please add a comment as such.

asudoh (approved):
Thanks @IgnacioBecerra for all the updates!

PR #5010, opened 2021-01-26
fix(linkList): add aria attribute to item-card

ariellalgilmore:
### Related Ticket(s)

#4060 

### Description

To provide each link-list-item-card with a landmark region and appropriate label to convey beginning and ending boundary for the screen reader user. 

### Changelog

**New**

- `aria-label` attribute

**Changed**

- `role` to using `region`

PR #5039, opened 2021-01-29
chore(linkList): revert changes to aria roles

ariellalgilmore:
### Related Ticket(s)

This is removing changes from this [PR](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/5010). 

### Description

After discussing in more detail it seems that this issue needs to be brought up with QA team wondering why it should read "region" vs "link" as the link list item is passed a `href` value
### Changelog

**Changed**

- `role: listitem`


kennylam (approved):
LGTM thanks @ariellalgilmore!

Issue #4781, opened 2020-12-28
Web component: Card Section- Carousel - In 300% zoom, we are not able to see the elements and content of the entire page.


Praveen-sr:
## Detailed description

Issue: 
In 300% zoom, we are not able to see the elements and content of the entire page.

Expected: 
1.Page should be visible with all the elements present on that.
2.Content can be presented without loss of information or functionality, and without requiring scrolling in two dimensions.

PR #5042, opened 2021-01-29
fix(card-section-carousel): fixes horizontal scroll

andysherman2121:
### Related Ticket(s)

#4781 

### Description

For Card Section Carousel, on 300% zoom you need to scroll both horizontally and vertically to see the full page contents. This fix gets rid of the horizontal scroll, bringing accessibility back in check. This will also fix horizontal scroll on all other web components that use Content Section


kennylam (approved):
LGTM!

Issue #5040, opened 2021-01-29
AEM Link List: One column for 1-3 links


StephanieHoman:
## Detailed description

> Describe in detail the issue you're having.
Link list items should span 1 column when holding 1-3 links.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?
Bug

> Is this issue related to a specific component?
Link List

> What did you expect to happen? What happened instead? What would you like to
> see changed?
The link list should switch to one column with 1-3 links. Currently it is 2 columns.

> What browser are you working in?
Safari Version 14.0

> What version of Carbon for IBM.com are you using?

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Additional information
[Link to the DDS Design spec file](https://ibm.ent.box.com/file/698433449371?s=hgrc62y10kopxdk5z75fcxq48m6ad3vc) in Box Design is on pg 4.

PR #5053, opened 2021-02-01
fix(link-list): fixed wrong column count and styling on knob change

RaphaelAmadeu:
### Related Ticket(s)

#5040 

### Description

Solved a bug when `LinkList` at `end` variant would split in two columns with 3 elements, when it only should split after 4 items.

Corrected styling on `dds-link-list-item-cta` on `end` variant, so it has the same look as `dds-link-list-item`.

Corrected a Storybook bug that made the items on `three-columns` variant break if `cta-type` changed on knobs.


packages/web-components/src/components/link-list/link-list.ts
```diff
@@ -119,6 +122,7 @@ class DDSLinkList extends StableSelectorMixin(LitElement) {
   }
 
   updated() {
+    console.log(this.type);
```
asudoh, line 125:
```suggestion
    console.log(this.type);
```

asudoh (commented):
Just one thing, thanks @RaphaelAmadeu!

asudoh (approved):
Thanks @RaphaelAmadeu for the update!

kennylam (approved):
LGTM thanks @RaphaelAmadeu!

Issue #4744, opened 2020-12-17
Web component: Content block cards with Video - After closing the dialog visually it is disappeared. But, Screen reader focus is stuck on close button


Praveen-sr:
## Detailed description

On iPhone voiceover

Issue: 
After closing the dialog visually it is disappeared. But, Screen reader focus is stuck on close button. We are unable to navigate using Left/Right swipe. This issue is not found in react Staging.

Expected:
After closing the dialog, User should be able to navigate to the parent page. And the dialog should disappear for the screen reader user as well.

RaphaelAmadeu:
@Praveen-sr 

Can you please check if the issue is still occuring for you? It seems that the focus changes and it's working finelly with chromevox

Praveen-sr:
@RaphaelAmadeu Ram has retested this issue in IOS 13.7 & 14.3 in both Safari and Chrome 87and able to reproduce - Not fixed.

CC @RobertaJHahn

RaphaelAmadeu:
@Praveen-sr 

Could you please try it again using this url?
http://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-content-block-cards--with-videos

Thank you!

Praveen-sr:
@RaphaelAmadeu Ram re tested this on the above link, issue not fixed.

RaphaelAmadeu:
https://images.zenhubusercontent.com/193129061/10c1b950-6e1d-4347-92e5-fb24d9015297/video_2021_02_05_17_11_57.mp4

@Praveen-sr 
This is a capture from my Iphone using voice over, isn't this the supposed behavior?

Praveen-sr:
@RaphaelAmadeu Ram re tested the same scenario on iPhone, Screen reader focus is stuck on close button. We are unable to navigate using Left/Right swipe.

IgnacioBecerra:
@Praveen-sr I wasn't able to replicate this behavior using an iPhone 8 on the latest iOS, would you be able to retest, and if possible, provide a video? Thank you!

cc: @RobertaJHahn

Praveen-sr:
@IgnacioBecerra retested in Web component canary contain block card video in VO IOS,
After closing the dialog now the Screen reader focus is not stuck on "close" button. focus is set back on the triggering element. Above issues is working fine now. 

But following regression found, i.e. Even after closing the modal, VO still identifies the player controls at the end of the page which is not visible.

PR #5083, opened 2021-02-03
fix(card): added focus on click

RaphaelAmadeu:
### Related Ticket(s)

#4744 

### Description

Added focus on click for the card component, as Safari would faltly consider `body` as the active element when the card was clicked, this caused an accessibility bug where the focus would be trapped inside `ExpressiveModal` even after closed.


guilhermelMoraes (approved):
LGTM! 👍🏻

PR #32, opened 2019-07-30
Release/v0.1.0

jeffchew:
Merging release v0.1.0 and v0.1.0-rc.0 into master

#### Changelog

**New**

- packages version bump
- v0.1.0 changelogs
- v0.1.0-rc.0 changelogs

PR #5088, opened 2021-02-03
fix(profile): catch error for profile api and console log error

annawen1:
### Related Ticket(s)

[Unhandled Runtime Error #32](https://github.com/carbon-design-system/carbon-for-ibm-dotcom-nextjs-template/issues/32)

### Description

Catch the profile api call error and console log it in attempt to prevent error overlay from appearing in nextjs template dev mode.

### Changelog

**Changed**

- catch error and console log while returning `Unauthenticated`


jeffchew (approved):
LGTM!

PR #5091, opened 2021-02-04
chore(masthead): update codesandbox example

kennylam:
### Description

Fixes `scss` import errors from `carbon-components` that were preventing the example from rendering properly.

### Changelog

**New**

- `.sassrc` to fix import paths

**Changed**

- use `latest` version of `carbon-components`

**Removed**

- redundant `masthead-search` example


asudoh (approved):
LGTM 👍 - Thanks @kennylam for the fix!

PR #5095, opened 2021-02-04
fix(contentBlockMedia): align to non-gutter grid

ariellalgilmore:
### Related Ticket(s)

#4961 

### Description

Align to non-gutter grid line. Also updated stories to match react storybook options. 

### Changelog

**Changed**

- use CSS grid for layout for `<dds-content-block-media>`.


asudoh (approved):
LGTM 👍 - Thanks @ariellalgilmore for jumping in!

Issue #4142, opened 2020-10-06
CTA / Content Block Simple - downloads a file, open external link information is not conveyed for the screen reader user.


Praveen-sr:
## Detailed description
On JAWS FF &  iPhone 

Issue: There is an icon near these links which indicates it downloads a file, open external link  but, the same information is not conveyed  for the screen reader user.

Note: the same issue we found in react staging as well.

Expected: 
The information conveyed visually should be available for the screen reader user as well. Anyone of the following methods can be used to make it screen reader accessible.
1. provide the information in the link text. 
2. title attribute. or 
3. as an off screen text
a generic example: 
1. feed back form opens in new window
2. products opens external link

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/36b652e9-78ff-4980-b3ed-0713af7fb9b2)

Issue #4060, opened 2020-09-25
Web component: Link List - Beginning and Ending boundary for the cards


Praveen-sr:
## Detailed description
On  JAWS FF & VO iPhone safari

Issue: The beginning and ending boundary for each card is not conveyed for the screen reader user. 
Note: Working fine in React Staging

Expected : Provide each card within the landmark region and give appropriate label.

asudoh:
Here's what I see in React storybook, not sure if the behavior, especially with the wider width, is something we should mirror to Web Components. Any thoughts...? Thanks! CC @oliviaflory @JennySanchez @RaphaelAmadeu 

![linklist-responsive-padding](https://user-images.githubusercontent.com/1259051/94703767-2b2a0680-037a-11eb-891a-f15eada7ac9a.gif)

ariellalgilmore:
@Praveen-sr is it possible to get more info on why this should be labeled a `region`. The link list cards are passed in `href` values meaning this is a `link` item. Should the possible fix instead be changing the React version to a `listitem` role instead of a `region` role

cc: @asudoh

Issue #2728, opened 2020-06-11
Component: CTA - Screen reader - Play icon information on card is not read for screen reader user.


Praveen-sr:
## Detailed description

Issue :- Visually there is a play icon present near the link indicating this link opens a video. But the same info is not available for the screen reader user.

Expected :-
Convey the same info for the screen reader user as well. This can be done by adding the text "plays video" along with the link text /label.

kennylam:
@Praveen-sr Adding copy such as 'plays video' is challenging since this would need to be translated, and not enforceable by us since we can't control what copy is used for the link. This is the same issue for all links that play video or open the url in a new page/tab.

Maybe one solution would be to provide an `aria-label`/`title` for these types of links as props, using `us-en` as the default. Users would could then add copy in whatever language is needed and the responsibility of providing the translation is theirs.

@jeffchew @asudoh @annawen1 @ljcarot Thoughts?

Praveen-sr:
@kennylam Aria label can be used if there is an interactive element like a link or a button. Because Screen reader generally do not pick the aria label on non-interactive elements. If it is an image tag, providing alt text for the image would be a better solution.

kennylam:
> @kennylam Aria label can be used if there is an interactive element like a link or a button. Because Screen reader generally do not pick the aria label on non-interactive elements. If it is an image tag, providing alt text for the image would be a better solution.

@Praveen-sr I was talking more about putting the `aria-label`/`title` on the link itself and not the play icon. or even the CTA link. But yes, it would be on an interactive, tabbable element. As a prop we could default it to English, while still allowing users to override in the required language.

@jeffchew @ljcarot @andysherman2121 Thoughts?

andysherman2121:
@kennylam I agree, this is definitely the best solution I can think of. This would give the user the ability to change the prop to whatever language they want. So we don't have to handle any of the translation

jeffchew:
@RobertaJHahn moving this out to the next release.

cc: @andysherman2121

Issue #2384, opened 2020-05-14
Accessibility - External link opens in new tab is not conveyed to user.


Praveen-sr:
## Detailed description

Issue :- 
There is an icon near these links which indicates that these links open in new window but, the same information is not conveyed  for the screen reader user.

Expected :
The information conveyed visually should be available for the screen reader user as well. Anyone of the following methods can be used to make it screen reader accessible.
1. provide the information in the link text. 
2. title attribute. or 
3. as an off screen text

Example: 
1. feed back form opens in new window
2. products opens external link

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/eba1f3a0-229e-4216-8d9b-bbc93b36cbfa)

caiobozato:
I've talked with @RaphaelAmadeu and @kennylam and since this is something we don't have a native way to do, we need to discuss with all dev team and design as well.

My suggestion is to use something like `aria-label={"{link text} - This link opens in a new tab"}`

But again, we need to discuss that.

asudoh:
Weighing in; I think this is a good idea. [More resource](https://www.w3.org/TR/WCAG20-TECHS/G200.html)

RaphaelAmadeu:
I think that the only thing to consider is that as a Design System we'll have to support another languages. An idea would be an utility script for this purpose, maybe using the translation api

ljcarot:
So we shouldn't assume that External links open in a new tab, their purpose is just to indicate leaving IBM to an "external page".  See v18 standard https://www.ibm.com/standards/web/v18/design/external-links/ for description.  Have we changed this behavior in Carbon?

RaphaelAmadeu:
@ljcarot as far as I know, the previous QA guidelines pointed that all external links should be opened in another tab, we could confirm that

ljcarot:
We should open up an additional knob in storybook to allow for open in new window or open in existing.  Then if we open in new window, screenreader will be able to detect that.  Dev story to work on.  cc: @jeffchew

RaphaelAmadeu:
@ljcarot We just need to remember that opening external links in a new tab is set by code on `CTA` component itself, so this would require some editing on it, in order to be able to create a prop for target.

PR #5096, opened 2021-02-04
fix(cta): add default aria-label for different types

IgnacioBecerra:
### Related Ticket(s)
#2728
#2384
#4142
#4060 

### Description

Added an `aria-label` to inform screen reader users the CTA about the type of link they are focused on, such as `external`, `video`, or `download`. The default label is provided in English by default, but the message can be a custom translation that is passed in using the `ariaLabel` prop by adopters.

This PR also changes the `role` of LinkList to `listitem` with the use of a custom `ariaRole` prop to specify different roles depending on the use within components.

### Changelog

**New**

- added `ariaLabel` to `CTA` component for adopters to customize the message in a different language
- added `ariaProp` to `CTA` component to set a custom role depending on the use in other components

**Changed**

- `aria-label` message for CTA Text and Card
- set `role='listitem` in `LinkList`


ariellalgilmore (approved):
LGTM!

annawen1 (approved):
LGTM too!

PR #5110, opened 2021-02-05
fix(contentBlockMixed): align to non-gutter grid

ariellalgilmore:
### Related Ticket(s)

#4962 

### Description

Align to non-gutter grid line. Also updated stories to match react storybook options. 

### Changelog

**Changed**

- use CSS grid for layout for `<dds-content-block-mixed>`.


asudoh (commented):
LGTM basically, one question; Did we need any CodeSandbox updates? Thanks @ariellalgilmore!

ariellalgilmore:
> LGTM basically, one question; Did we need any CodeSandbox updates? Thanks @ariellalgilmore!

Good catch! looks like a codesandbox was never created for this component.

asudoh:
Thank you for clarifying @ariellalgilmore! Another question; Seems that there are changes to two other components. Do their stories and CodeSandbox need any update? Or are they not necessary...?

ariellalgilmore:
> Thank you for clarifying @ariellalgilmore! Another question; Seems that there are changes to two other components. Do their stories and CodeSandbox need any update? Or are they not necessary...?

The other components are not necessary. I will create a codesandbox example for content-block-mixed though and include in this PR.

asudoh (approved):
Thanks @ariellalgilmore for the update!

Issue #4985, opened 2021-01-22
Web Components: Document the automatic license banner update


RobertaJHahn:
### The problem

The mechanism of automatic license banner update of `@carbon/ibmdotcom-web-components` has some new technical approaches in its implementation details.

### The solution

Document them so new developers can easily jump into the codebase.

PR #5114, opened 2021-02-08
docs(license): notes on license header

asudoh:
### Related Ticket(s)

Refs #4985.

### Description

Adds implementation notes on license banner checker so that new developers can quickly understand how it works.

### Changelog

**New**

- Implementation notes on license banner checker so that new developers can quickly understand how it works.


kennylam (approved):
Thanks @asudoh!

Issue #4703, opened 2020-12-10
OrderedList - spacing between number and start of the paragraph is too crowded when the number is more above 9


Daynssjal:
## Detailed description

> Insufficient spacing when step numbers are double-digit.
The default Carbon font-size for list items is 0.875rem, and the spacing is adjusted to it - the ::before selector has property left: - 1.5rem
https://react.carbondesignsystem.com/?path=/story/orderedlist--default

> In the Dotcom expressive theme, the list items font-size is larger - 1rem, so also the spacing should be adjusted accordingly - ::before selector should also be larger, but it is still -1.5rem. It looks very crowded - there should definitely be some space between the number and the start of the paragraph. Please see the attached image.
http://carbon-expressive.mybluemix.net/?path=/story/orderedlist--default

> Is this a feature request (new component, new icon), a bug, or a general
> Bug

> What did you expect to happen? What happened instead? What would you like to
> see changed?
> There should be more space between the number and the start of the paragraph, if the number is double-digit

> What browser are you working in?
Chrome, FF, Edge, Safari

> What version of Carbon for IBM.com are you using?
1.13.0

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?
KC - IBM Docs

## Additional information
<img width="834" alt="crowded-numbers-ordered-list" src="https://user-images.githubusercontent.com/33804471/101793441-18fcdf00-3b06-11eb-97a2-a76b9a9dff92.PNG">

PR #5119, opened 2021-02-08
fix(ordered-list): spacing too crowded when number above 9

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/4703

### Description

Ordered list spacing between number and start of the paragraph is too crowded when the number is above 9. Modified to have the same position as nested list.

ordered-list-solution

### Changelog

Created 
- New scss file for lists (packages/styles/scss/themes/expressive/components/_list.scss) in expressive theme

**Changed**

- Added new scss imports to index (packages/styles/scss/themes/expressive/index.scss)

**Removed**

- Nothing


oliviaflory:
@NSDrowned is this update applied to the Carbon Expressive canary storybook? I don't see a deploy link above, and the lists are in that storybook. https://carbon-expressive.mybluemix.net/?path=/story/orderedlist--default

NSDrowned:
@oliviaflory I updated the story and the deploy preview has been generated

oliviaflory:
Thank you @NSDrowned for providing the link to the Expressive  storybook

For this fix, I think we need to maintain the same amount of padding that Carbon provides specs for, of 8px between the number and the list item text. Looks like there is only 4px of spacing in our solution.

Carbon
Screen Shot 2021-02-09 at 2 13 56 PM

Ours
Screen Shot 2021-02-09 at 2 20 55 PM


----

Another general question I have, Carbon has two different variations for Ordered list

[Default](https://react.carbondesignsystem.com/?path=/story/orderedlist--default), which keeps the numbers left aligned

Screen Shot 2021-02-09 at 2 22 56 PM


and

[Native list styles](https://react.carbondesignsystem.com/?path=/story/orderedlist--native-list-styles) which are right aligned as you shared

Screen Shot 2021-02-09 at 2 23 21 PM


Were you able to see if the upgrade to the latest release v10.28.0 fixed the overlapping problem? Ideally we would stay as aligned with the base system as possible. 
Have we considered bringing in these newer variants in order to be 1:1? cc: @jeffchew

NSDrowned:
Hello @oliviaflory, I have updated the margin to 8px ($carbon--spacing-03 instead of $carbon--spacing-02).

This update only affects the default list, not the nested one. I checked the latest version and they're still using position absolute left for the number. This means the larger numbers will never push the item text to the right, which causes the overlapping. My update keeps the absolute position but to the right, which lets the numbers grow to the left avoiding the overlap and makes it look exactly the same as the nested list.

NSDrowned:
Hello @oliviaflory, I have updated the margin to 8px ($carbon--spacing-03 instead of $carbon--spacing-02).

8px

https://carbon-expressive.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/5119/index.html

This update only affects the default list, not the nested one. I checked the latest version and they're still using position absolute left for the number. This means the larger numbers will never push the item text to the right, which causes the overlapping. My update keeps the absolute position but to the right, which lets the numbers grow to the left avoiding the overlap and making it look exactly the same as the nested list.

oliviaflory:
Ok, thanks for explaining @NSDrowned, I will approve for now!

In the future, for the expressive components we have usually submitted a bug or ideally a PR to fix the issue straight to Carbon, so that it's fixed at the source. That way, we maintain as much parody with the Design System as possible, and don't have to maintain a bunch of overrides.

In this instance I think fixing on our side was correct so we have a fix for our adopter's problem. Let's submit a fix to Carbon so they can address this issue as well, then we can remove the override down the road :)

oliviaflory (approved):
Approved, + need to submit a fix to Carbon so we can remove override in the future

NSDrowned:
@oliviaflory Sounds good and I agree, it should be fixed at the source. Thanks!

jeffchew:
@oliviaflory @NSDrowned the alignment caught in one of our snapshots looks to have the ordered list a bit to the left and alignment looks off: https://percy.io/538fc19a/Carbon-for-IBM.com-React/builds/8855555/changed/502371318?browser=chrome&browser_ids=9&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1280&widths=320%2C1280

oliviaflory:
@jeffchew @NSDrowned  hmm, yeah that's not good. I thought the fix would have shifted the entire list item over, not just the numbers

NSDrowned:
@jeffchew @oliviaflory changing the number alignment shouldn't have moved the list positioning (since the numbers are absolute positioned pseudo elements). Let me check what happened.

NSDrowned:
Hello @oliviaflory, I reviewed the test and it looks like the list items are not moving, just the numbers with the new margin (8px) and alignment. It looks like the previous snapshot still has the old margins (4px) and that's why the visual test is failing. Perhaps you can review the changes in Percy an approve them:

https://percy.io/538fc19a/Carbon-Expressive-Theme/builds/8876042?utm_campaign=538fc19a&utm_content=Carbon-Expressive-Theme&utm_source=github_status_public

https://percy.io/538fc19a/Carbon-for-IBM.com-React/builds/8875917?utm_campaign=538fc19a&utm_content=Carbon-for-IBM.com-React&utm_source=github_status_public

oliviaflory:
@NSDrowned Hmm, I'm not sure...

I cannot approve the example here because the alignment of the numbers is off in comparison to the other text and list items alignment: https://percy.io/538fc19a/Carbon-for-IBM.com-React/builds/8875917?utm_campaign=538fc19a&utm_content=Carbon-for-IBM.com-React&utm_source=github_status_public

Previously, the numbers are aligned with the other list items, and the text above. This is what we need to achieve, leaving the numbers in their position, and possibly adding some spacing to the right, therefor pushing the list items to the right as well to avoid overlap.
Screen Shot 2021-02-16 at 1 42 18 PM


In comparison, the updates push the numbers to the left and causes them to be misaligned with the other list items and text.
Screen Shot 2021-02-16 at 1 42 28 PM


It's possible I'm not understanding this fully, so let me know if it would be worth while to have a quick meeting to discuss.

---

I'm also still hesitant to change to right aligned numbers when Carbon includes the option under the Native Style variant. Can we pull that variation into the expressive theme?


I just want to make sure we're fixing this in the right way to not deviate from the system too much...

NSDrowned:
Hello @oliviaflory. You're right, I have reverted the list to a left alignment. I added a bit of extra margin so that the double digit numbers are 8px apart from the list item, that way we don't deviate that much just for this issue.  Let me know what you think.

ordered-list-02

oliviaflory:
Thank you @NSDrowned, looks great!

PR #5130, opened 2021-02-09
fix(callout-with-media): Align to non gutter grid line

andysherman2121:
### Related Ticket(s)

#4958 

### Description

Align to non gutter grid line

### Changelog

**Changed**

- use CSS grid for layout for `dds-callout-with-media`


asudoh (approved):
LGTM 👍 - Thanks @andysherman2121!

PR #5133, opened 2021-02-10
fix(fade-in-out): removed dotcom shell from stories

IgnacioBecerra:
### Related Ticket(s)
None.

### Description
Removed the Masthead and Footer from the FadeInOut component stories, to only bring focus to the fade effects of the content itself.

### Changelog

**Removed**

- The `DotcomShell` container that wrapped around the `FadeInOut` component


kennylam (approved):
Thanks @IgnacioBecerra!

Issue #5081, opened 2021-02-03
Web component: Mega Menu - "The essentials,  Hybrid Cloud" are read as menu options for screen reader users.


Praveen-sr:
## Detailed description

On JAWS FF
Issue:
Screen reader is not supposed to identify the mentioned texts (The essentials,  Hybrid Cloud) as "menus". And if they are headings visually, They have to be identified as "heading" instead.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/861e5d47-ab9f-4ef0-bbd4-b86239e5a85d)

annawen1:
@Praveen-sr can this be tested again? I'm tabbing through the menu and JAWS doesn't identify the "The essentials" or "Hybrid Cloud" text, should I be using some other keyboard command?

https://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-masthead--with-custom-data

Praveen-sr:
@annawen1 Re tested this on web component, issue still exists 
Steps to reproduce
1. After expanding "Product and solutions"
2. Use tab navigation to come to "what is hybrid cloud" link, then use up arrow keys to navigate through the text
"The essentials", "Hybrid Cloud".
Here "The essentials, Hybrid Cloud" are read as "menu" options for screen reader users.

annawen1:
@Praveen-sr Ah ok, thank you for the detailed steps - will try that

annawen1:
@Praveen-sr I tried the steps you mentioned, tabbing to the "what is hybrid cloud" link then using the arrows keys to navigate.

I am using a Virtual Machine and am trying to figure out how to enable audio but for now I have JAWS subtitles/captioning enabled and I do not see it mentioning "menu" when I navigate to the "The essentials" or "Hybrid cloud" text. I've posted a video. Please let me know if there is anything I'm missing or I should try.

https://user-images.githubusercontent.com/54281166/107439902-cbe6ba80-6b00-11eb-9e3d-65ed5205646c.mov

annawen1:
@Praveen-sr I tried with JAWS 2020 version on both Firefox and Chrome but still can't replicate the issue

https://user-images.githubusercontent.com/54281166/107566398-d8c3e680-6bb2-11eb-9901-1a710625a081.mov

Praveen-sr:
@annawen1 Hi, Ram was not able to screen record JAWS audio, instead he is provided the JAWS speech history, 

[...]

PR #4624, opened 2020-12-03
fix(MegaMenu): remove unnecessary role for dropdown menu/megamenu

annawen1:
### Related Ticket(s)

React: Issues Mega menu options navigation #4196

### Description

The dropdown menu/megamenu shouldn't have the `role="menu"` applied as listed here:
Example Disclosure for Navigation Menus | WAI-ARIA Authoring Practices 1.1
https://www.w3.org/TR/wai-aria-practices/examples/disclosure/disclosure-navigation.html

Screen Shot 2020-12-03 at 4 25 20 PM

### Changelog

**New**

- {{new thing}}

**Changed**

- {{changed thing}}

**Removed**

- `role` attribute from the `<HeaderMenu />` component

PR #5160, opened 2021-02-11
fix(megamenu): remove menu role for accessibility

annawen1:
### Related Ticket(s)

Web component: Mega Menu - "The essentials, Hybrid Cloud" are read as menu options for screen reader users. #5081

### Description

This PR removes the `role='menu'` from the `ul` as it is unnecessary and conflicts with the elements placed in the megamenu/dropdown.

PR here from carbon-for-ibm-dotcom where we removed the menu role from our own version of the header menu
https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/4624

The dropdown menu/megamenu shouldn't have the role="menu" applied as listed here:
Example Disclosure for Navigation Menus | WAI-ARIA Authoring Practices 1.1
https://www.w3.org/TR/wai-aria-practices/examples/disclosure/disclosure-navigation.html

Screen Shot 2020-12-03 at 4 25 20 PM

### Changelog

**Changed**

- grab the `bx--header-menu` element and remove the `role` attribute

### DIFF
--- a/packages/web-components/src/components/masthead/megamenu-top-nav-menu.ts
+++ b/packages/web-components/src/components/masthead/megamenu-top-nav-menu.ts
@@ -7,7 +7,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import { customElement } from 'lit-element';
+import { customElement, query } from 'lit-element';
 import settings from 'carbon-components/es/globals/js/settings';
 import ddsSettings from '@carbon/ibmdotcom-utilities/es/utilities/settings/settings.js';
 import { forEach } from '../../globals/internal/collection-helpers';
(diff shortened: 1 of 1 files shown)


packages/web-components/src/components/masthead/megamenu-top-nav-menu.ts
```diff
@@ -76,6 +82,7 @@ class DDSMegaMenuTopNavMenu extends DDSTopNavMenu {
 
   updated(changedProperties) {
     super.updated(changedProperties);
+    this._menuNode.removeAttribute('role');
```
asudoh, line 85: Would you do this in `firstUpdated()` given it doesn't have to do with a property update?

asudoh (commented):
Thank you for isolating the change @annawen1!

packages/web-components/src/components/masthead/megamenu-top-nav-menu.ts
annawen1: ah yes! will do

asudoh (approved):
Thanks @annawen1 for the update!

PR #5168, opened 2021-02-12
fix(masthead): customtypeahead api fix

kennylam:
### Description

Updates masthead search `customTypeAheadApi` to be fully on the application side. When using a custom API, it will no longer pass through `SearchTypeaheadAPI` at all. This allows users to pass a full API including any necessary query params that were not defined in `SearchTypeaheadAPI`. With this change the use of environment variables is no longer necessary.

### Changelog

**Changed**

- moves `customTypeAheadApi` out of `SearchTypeaheadAPI`
- update jest environment variables
- section titles no longer appear when there are no section items


annawen1 (approved):
LGTM!

Issue #4888, opened 2021-01-13
Web component: Footer - Language selector, the width of the selector is smaller compared to react version.


Praveen-sr:
## Detailed description

Issue: 
On desktop & Tab view
Language selector, the width of the selector is smaller compared to react version.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/cb58fa45-0585-448b-8532-7f2e866a91b8)

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/924ecc5f-22e7-4560-950e-c7368960f41a)

PR #5176, opened 2021-02-16
fix(language-selector): width of web component different than react version

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/4888

### Description

The default and short language only selectors where missing styles and the max breakpoint that the react version has. This affected how they looked in tablet and desktop. 

Affected files: packages\web-components\src\components\footer\footer.scss

### Changelog

**New**

- Added the 'max' breakpoint and it's values to both (short and default) selectors

**Changed**

- Updated the 'lg' breakpoints for both (short and default) selectors


packages/web-components/src/components/footer/footer.scss
```diff
@@ -254,7 +259,14 @@
 
   @include carbon--breakpoint('lg') {
     order: 1;
+    flex: 0 0 25%;
+    max-width: 20rem;
+    min-width: 13.5rem;
```
kennylam, line 264: Thanks @NSDrowned! We can avoid hard-coding the width values by using `carbon--mini-units()`.

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/0038432f7fa8b2596b403a534b2f42c527b3f122/packages/styles/scss/components/footer/_language-selector.scss#L51-L52

NSDrowned: Nice! I didn't know that :)

jeffchew: @NSDrowned I see you resolved this thread but I don't see the change, perhaps forgot to push?

NSDrowned: @jeffchew forgot to push 🤦

jeffchew:
Thanks @NSDrowned , there looks to be prettier issues now, can you take a look?

kennylam (approved):
LGTM thanks @NSDrowned!

Issue #4743, opened 2020-12-17
Web component: Content block cards with Video - The screen reader focus is not trapped within the dialog


Praveen-sr:
## Detailed description

Issue: 
1. In jaws FF, The screen reader focus is not trapped within the dialog. Screen reader focus is trapped in the modal in React staging 
2. In Vo iPhone, the screen reader focus is not trapped in both the Environments

Expected:
Screen reader focus should be trapped within the modal dialog

PR #5182, opened 2021-02-16
fix(expressive-modal): handle focus wrapping with video

IgnacioBecerra:
### Related Ticket(s)
#4743

### Description

When focusing on the Expressive Modal with Video, there are times when the focus event has a `relativeTarget` that is null, which comes from tabbing around the video `IFrame` element. This PR handles this special case.

Also, since the original logic used `focusout`, detecting and redirecting out of viewport focus will fail (since it's not possible to focus on elements if focus not within page), so this PR changes to `focusin` to properly wrap around the elements.

### Changelog

**New**

- added special case to handle focus wrapping when `relatedTarget` is null

**Changed**

- changed focus listener to `focusin`


asudoh:
Thank you for trying to fix the problem @IgnacioBecerra - Our modal uses "focus sentinel" (hidden focusable elements around modal body) to detect if the focus is about to go out of the viewport. It means that handling `focusout` event still should work. That said, do you want to bring up the very specific steps to reproduce the problem you found? Thanks!

asudoh:
Also wondering if #5189 helps.

IgnacioBecerra:
@asudoh You're right, retesting using `focusout` I was able to get it working, I must've had the wrong conditional check when previously testing. 

To reproduce the issue, open [Content Block Cards - With Video](https://ibmdotcom-web-components-canary.mybluemix.net/iframe.html?id=components-content-block-cards--with-videos), select any of the cards, and once the modal appears, continue tabbing until the focus goes outside the viewport to the address bar.

kennylam:
> To reproduce the issue, open [Content Block Cards - With Video](https://ibmdotcom-web-components-canary.mybluemix.net/iframe.html?id=components-content-block-cards--with-videos), select any of the cards, and once the modal appears, continue tabbing until the focus goes outside the viewport to the address bar.

@IgnacioBecerra Focus remains in the video for me?

IgnacioBecerra:
@kennylam Hm that's odd, focus doesn't stay within the video for me, it's always left the modal. Here's a video replicating the issue in Web Components Canary.

https://user-images.githubusercontent.com/24970122/108221210-612f1380-70ec-11eb-9526-f101ca8a8669.mov

After discussing with @asudoh offline, we decided to separate this edge case within its own `onfocusin` listener instead of placing it inside `onfocusout` with the original logic.

asudoh (approved):

Issue #5034, opened 2021-01-28
[card]: underline on active


photodow:
## Detailed description

The card component should not be getting an underline state when `:active`.

## Steps to reproduce the issue

1. inspect element
2. trigger active state

## Additional information

<img src="https://user-images.githubusercontent.com/3793636/106175180-6f2bed00-615b-11eb-9689-77185471c5d2.png" width="300px" />

PR #5183, opened 2021-02-16
fix(card): remove text decoration from active state

kennylam:
### Related Ticket(s)

#5034 

### Description

Removes underline from `Card` active state 

### Changelog

**Changed**

- update card styles to remove underline when active


jeffchew (approved):
LGTM!

Issue #5194, opened 2021-02-17
React & Web component: Carousel - Screen reader is not identifying the copy content for all cards.


Praveen-sr:
## Detailed description

On JAWS & iPhone voiceover
Issue:
Screen reader JAWS is not identifying the content present below the  h3 heading, for all the cards with in the carousel.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/79535e43-eb48-4f44-8ca8-e60772d0ad02)

PR #5196, opened 2021-02-17
fix(card): remove aria-hidden from copy

ariellalgilmore:
### Related Ticket(s)

#5194 

### Description

remove aria-hidden from copy. Reads copy text in carasoul

### Changelog

**Removed**

- `aria-hidden: true`


guilhermelMoraes (approved):
LGTM, Ari! Thanks! 👍🏻

Issue #4845, opened 2021-01-07
React:  Masthead L0 - Grouped custom typeahead results enhancement


kyjak:
We have several groups of typeahead (search, product, last searches), so we need to forward labels for the groups and typeahead results for them.

![image](https://user-images.githubusercontent.com/15310250/103895952-7da15e80-50f1-11eb-9dd0-88eaf309a62e.png)

#### Additional Information
- Design and functional spec updates are needed.  This work is included under the following issue:
      - [Functional spec change issue](https://app.zenhub.com/workspaces/carbon-for-ibmcom-5d449f3642eb1962336cbe52/issues/carbon-design-system/carbon-for-ibm-dotcom/4944)
      - [Design spec change issue](https://app.zenhub.com/workspaces/carbon-for-ibmcom-5d449f3642eb1962336cbe52/issues/carbon-design-system/carbon-for-ibm-dotcom/4945)
- [link to the KC adopter meeting notes](https://ibm.ent.box.com/notes/601531416485)
- Wonil will provide design input by EOD Friday, Jan 22.
- Design QA testing issue (#5006 )
- Prod QA testing issue (#5007 )

#### Acceptance criteria
- [ ] Include README for the react component and corresponding styles
- [ ] Do not create knobs in Storybook that include JSON objects
- [ ] Break out Storybook stories into multiple variation stories, if applicable
- [ ] Create matching stable selectors in React component from the corresponding Web Component
- [ ] Create codesandbox example under `/packages/react/examples/codesandbox` and include in README
- [ ] Minimum 80% unit test coverage
- [ ] The Storybook link is added to the Design QA issue for their testing
- [ ] A comment is posted in the Prod QA issue, tagging Praveen, when development is finished

RobertaJHahn:
Type ahead is available, but the grouping is not provided.   Next steps:

1)  Design is needed for the feature to group the keyword suggestions.  The DDS squad would have to investigate the impact to the ibm.com search results.  
2)  Kenny will investigate and begin building the solution.  
3)  Andree will work offline with Wonil to provide the design, sending the designs to Jenny and Wonil for review.

wonilsuhibm:
Here is a design based on standards set in the megamenu:
<img width="1587" alt="Screen Shot 2021-01-07 at 9 01 39 AM" src="https://user-images.githubusercontent.com/30183447/103904675-939c2880-50cb-11eb-8d88-a2535cc37e75.png">

Key points here:
[...]

PR #5198, opened 2021-02-17
feat(search): allow custom url with noSearchRedirect

kennylam:
### Related Ticket(s)

#4845 

### Description

This adds support for user-defined URLs when using the [`searchNoRedirect`](http://ibmdotcom-react.mybluemix.net/?path=/docs/components-masthead--default#searchnoredirect) option in the masthead search. If provided, a URL will be placed in a search element's `data-href` attribute. This is useful when a URL redirect is required, instead of using the search value as a query param.

**Note:** when using `searchNoRedirect`, all logic must be provided by the user.

### Changelog

**New**

- add custom URL support to search result elements


jeffchew (approved):

guilhermelMoraes (approved):
LGTM! Thanks, Kenny! 👍🏻

Issue #5033, opened 2021-01-28
[card]: focus state outline is off


photodow:
## Detailed description
The card's focus state is thicker than it should, rounded in some cases, and adds corners/bevels in random spots.

## Steps to reproduce the issue

1. Inspect element
2. Trigger focus state

## Additional information

<img src="https://user-images.githubusercontent.com/3793636/106175725-114bd500-615c-11eb-85da-e974f2add964.png" width="300px" />

PR #5207, opened 2021-02-18
fix(card): focus state outline is off

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5033

### Description

Dds-card and dds-card-link had 2 outlines showing at the same time. I removed the default outline behavior which caused the black rounded border.

### Changelog

**Added**

+ outline: none on /packages/styles/scss/components/card/index.scss
+ outline: none on /packages/web-components/src/components/card/card.scss

### DIFF
--- a/packages/styles/scss/components/card/index.scss
+++ b/packages/styles/scss/components/card/index.scss
@@ -102,6 +102,7 @@
     }
   }
 
+  :host(#{$dds-prefix}-card-cta),
   :host(#{$dds-prefix}-card-link) {
     outline: none;
   }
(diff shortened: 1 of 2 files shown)


ariellalgilmore (commented):
Looks great @NSDrowned! Thanks! Just wondering if we can also apply these changes to cta card version

NSDrowned:
@ariellalgilmore ahh, missed that one. Will do right away!

ariellalgilmore (approved):
LGTM! Thanks for the updates

Issue #5003, opened 2021-01-25
[Leaving-IBM]: React wrapper - text styles larger than expected


annawen1:
## Detailed description

> Describe in detail the issue you're having.

The "The link you requested or were directed to is" line in the leaving-ibm modal React wrapper is larger compared to the web-components version.

The text should have styles:
```
  display: block;
  margin-top: $spacing-07;
  color: $text-02;
  @include carbon--type-style('label-01');
``` 
as expected in the web-components

REACT WRAPPER:
<img width="630" alt="Screen Shot 2021-01-22 at 5 25 16 PM" src="https://user-images.githubusercontent.com/54281166/105723086-5ff73600-5ef4-11eb-88fb-a648bee94472.png">

WEB COMPONENTS:
<img width="709" alt="Screen Shot 2021-01-25 at 9 58 11 AM" src="https://user-images.githubusercontent.com/54281166/105723111-684f7100-5ef4-11eb-9851-4924a1c1dd4b.png">

PR #5243, opened 2021-02-19
fix(leaving-ibm): fixed text styles

andysherman2121:
### Related Ticket(s)

#5003 

### Description

Fixed the React Wrapper text styles appearing larger than expected


kennylam (commented):
@andysherman2121 I'm not seeing the change in the `react-wrapper` version?

image

packages/web-components/src/components/leaving-ibm/leaving-ibm.scss
```diff
@@ -34,6 +34,6 @@
 :host(#{$dds-prefix}-leaving-ibm-modal-supplemental) {
   display: block;
   margin-top: $spacing-07;
-  color: $text-02;
   @include carbon--type-style('label-01');
+  @include carbon--type-style(body-long-01, true);
```
kennylam, line 38: Is `label-01` still needed?

andysherman2121: I got rid of `label-01` for web-components but still needed for the react-wrapper fix

packages/web-components/src/components/leaving-ibm/leaving-ibm-modal-body.ts
```diff
@@ -33,7 +33,7 @@ class DDSLeavingIbmModalBody extends BXModalBody {
     const { href } = this;
     return html`
       <p><slot></slot></p>
-      <slot name="supplemental"></slot>
+      <span><slot name="supplemental"></slot></span>
```
kennylam, line 36: @andysherman2121 Instead of adding a new `span` element you can use `::slotted` selector in the css.

andysherman2121:
@kennylam Getting rid the `span` tag and adding the `slotted`  selector works for `web-components`.  There seems to be a `span` styling on the `react-wrapper` version that's overwriting some of the `slotted` styles

kennylam:
> There seems to be a `span` styling on the `react-wrapper` version that's overwriting some of the `slotted` styles

@andysherman2121 You need to make the change here
https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/50daa3234a48a8e6e83eee46ea8fbaa45e4bd9a6/packages/web-components/src/components-react/leaving-ibm/leaving-ibm-composite.tsx#L101

andysherman2121:
@kennylam When I got it working in `react-wrapper`, the web component version styling wasn't working.  So I have two slotted stylings in the `scss`

packages/web-components/src/components/leaving-ibm/leaving-ibm.scss
```diff
@@ -29,9 +29,16 @@
...
+  ::slotted(slot[name='supplemental']) {
+    display: block;
+    margin-top: $spacing-07;
+    color: $text-02;
+    @include carbon--type-style('label-01');
+  }
 }
 
-:host(#{$dds-prefix}-leaving-ibm-modal-supplemental) {
+::slotted([slot='supplemental']) {
```
annawen1, line 41: would this affect other slots named 'supplemental'?

andysherman2121: ah good point. I can change it to target the classname

andysherman2121: @annawen1 @kennylam updated

annawen1 (approved):

Issue #4820, opened 2021-01-05
Utility: Add the slide up & down to the MVP animation / fade utility 


RobertaJHahn:
#### User Story

> As a `[user role below]`:
IBM.com Library developer

> I need to:
create the `animation / fade utility`

> so that I can:
provide the ibm.com adopter developers with a utility that will enable them to include animation in the ibm.com web pages they build

#### Additional information

- **See the Epic for the Design and Functional specs information**
 - [Design Issue 182](https://app.zenhub.com/workspaces/carbon-for-ibmcom-5d449f3642eb1962336cbe52/issues/carbon-design-system/digital-design-ideation/182)
 - [Prototype link](https://codepen.io/photodow/full/vYKaaxK)
- [Functional spec updates Box link](https://ibm.box.com/s/jjz5j9gofyqam02lldrbx7el6ekfvttz)
- Prod QA testing issue (#4821)

#### Acceptance criteria
- [ ] Include README for the web component and corresponding styles
- [ ] Minimum 80% unit test coverage
- [ ] Add this feature to the React NextJS test page for testing
- [ ] A comment is posted in the Prod QA issue, tagging Praveen when development is finished

PR #5248, opened 2021-02-19
feat(scroll-animations): introducing slide animations

IgnacioBecerra:
### Related Ticket(s)
#4820

### Description

Refactored `FadeInOut` component into `ScrollAnimations` to allow future scroll animations to be added or changed with ease, starting with the addition of the sliding scroll animations. 

### Changelog

**New**

- added `animation` property to `ScrollAnimations` to define which animations to display
- added slide animation effects to `ScrollAnimations` component

**Changed**

- refactored `FadeInOut` into `ScrollAnimations`


packages/web-components/src/components/scroll-animations/scroll-animations.ts
```diff
@@ -74,13 +77,28 @@ class DDSFadeInOut extends StableSelectorMixin(LitElement) {
...
+  private _exitEffectClass?: string;
+
   /**
    * Cleans observers upon update or exit, and creates new instances if needed.
    *
    * @param [options] The options.
    * @param [options.create] `true` to create the new intersection observer.
    */
   private _cleanAndCreateObservers({ create }: { create?: boolean } = {}) {
+    this._setAnimationClasses();
```
asudoh, line 101: Seems that this should be called in `updated()`, given it depends on `animation` property. Another option is simply maintaining a map of `animation` property and the CSS class being used.

asudoh (commented):
Looks good additions - Thanks @IgnacioBecerra! Web Component tag name change is a breaking change and wanted to have @jeffchew chime in on that.

asudoh (approved):
LGTM 👍 from my perspective, let's see what @photodow thanks. Thanks @IgnacioBecerra!

photodow (approved):

Issue #4937, opened 2021-01-20
Web component:  Lightbox media viewer - Modal layout size is small and copy text is not displayed correctly. 


Praveen-sr:
## Detailed description

Issue:
Lightbox media viewer - Modal layout size is not small and copy text is not displayed correctly. 

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/cd678beb-070a-4854-8c81-810c9a791cba)

#### Solution
- when the copy is long, the media should be fixed at the top and only the text should be scrollable.

PR #5254, opened 2021-02-22
fix(lightbox-m-viewer): fixing the expressive-modal styling issues

guilhermelMoraes:
### Related Ticket(s)

Web component: Lightbox media viewer - Modal layout size is small and copy text is not displayed correctly. #4937

### Description

Updating the selectors in the `expressive-modal` that were not targeting anything. With that, the new selector should work for the `full-width` version of the `expressive-modal`


asudoh (approved):

PR #5269, opened 2021-02-23
fix(content-group-cards): removed same-height usage for grid

RaphaelAmadeu:
### Related Ticket(s)

#4695 

### Description

In order to improve our performance and reduce flickering on load, I've changed our implementation of `ContentGroupCards` so it uses grid instead of our same-height utility.

### Changelog

**New**

- Grid based layout

**Removed**

- Same Height utility usage


asudoh (approved):
LGTM 👍 - A stretch goal for future would be aligning the markup to Web Components (see the snapshot change in #5110) so we can streamline our Sass code. Thanks @RaphaelAmadeu!

jeffchew (changes requested):
Hey @RaphaelAmadeu , it looks like something happened to the mobile view, can you check? https://percy.io/538fc19a/Carbon-for-IBM.com-Web-Components/builds/9003990/changed/510506575?browser=chrome&browser_ids=16&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=320&widths=320%2C1280

jeffchew (approved):
Thank you for fixing @RaphaelAmadeu !

Issue #5134, opened 2021-02-10
Web component: Masthead-With custom data - On landscape mode, overflow previous button is not working


Praveen-sr:
## Detailed description

Issue:
Masthead - With custom data- On landscape mode, overflow previous button is not working

## Additional information

![Mast-Cust.gif](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/cfe05aeb-d78e-4e60-9477-2e8df3fbe44d)

annawen1:
@Praveen-sr The fix to this bug has been merged, you can test here: https://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-masthead--with-custom-data

Praveen-sr:
@annawen1 Re tested this on above link, On landscape mode, overflow previous button is working fine. Thanks

CC @RobertaJHahn

PR #5272, opened 2021-02-24
fix(top-nav): drift away from offsetLeft property

asudoh:
### Related Ticket(s)

Refs #5134.

### Description

Drifts away from using `.offsetLeft` to calculate the first menu item that should be visible as user scrolls in top nav, because Safari seems to be error-prone in this area.

Changes to use `.getBoundingClientRect()` to get the equivalent value.

### Changelog

**Changed**

- Repalces `.offsetLeft` with `.getBoundingClientRect()` to calculate the first menu item that should be visible as user scrolls in top nav.
- Changes in CSS so the scrolling content area changes by showing/hiding the scrolling button.


ariellalgilmore (approved):
LGTM!! Thanks Akira for the updates!

annawen1 (approved):

Issue #5022, opened 2021-01-28
Web component: Masthead - On knobs, Brand name must be changed to Platform.


Praveen-sr:
## Detailed description

Issue:
No platform is directly visible to the user. On knobs section Brand name must be changed to Platform.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/047b3a01-42ab-4f45-8f20-6b43dc6920c2)

PR #5291, opened 2021-02-25
fix(masthead): change brand-name to platform

ariellalgilmore:
### Related Ticket(s)

#5022 

### Description

Change `brand-name` to platform for consistency with React

### Changelog

**Changed**

- property name from `brand-name` to `platform`
- knobs from `brandName` to `platform`


asudoh (commented):
LGTM basically. The change in attribute name is a breaking change and wanted to have @jeffchew to chime in. Thanks @ariellalgilmore!

jeffchew:
> LGTM basically. The change in attribute name is a breaking change and wanted to have @jeffchew to chime in. Thanks @ariellalgilmore!

@ariellalgilmore @asudoh 
Yes, this is a breaking change. It's unknown how many people are using this now in web components. But I'm wondering if we can introduce `brand-name`, and have any deprecation notice that `platform-name` will be removed in a future version?

ariellalgilmore:
> > LGTM basically. The change in attribute name is a breaking change and wanted to have @jeffchew to chime in. Thanks @ariellalgilmore!
> 
> @ariellalgilmore @asudoh
> Yes, this is a breaking change. It's unknown how many people are using this now in web components. But I'm wondering if we can introduce `brand-name`, and have any deprecation notice that `platform` will be removed in a future version?

Okay sounds good. So keep brand-name for now and just change the story knob to platform?

asudoh:
Given @ariellalgilmore is changing the attribute name from `brand-name` to `platform` instead of the way around, I'm wondering if @jeffchew meant to deprecate `brand-name`. @jeffchew Is it the case...? Thanks!

jeffchew:
> Given @ariellalgilmore is changing the attribute name from `brand-name` to `platform` instead of the way around, I'm wondering if @jeffchew meant to deprecate `brand-name`. @jeffchew Is it the case...? Thanks!

Yes, if the purpose of this is to align to the prop name in React. We shouldn't ditch `brand-name` entirely though as it's breaking, I was suggesting keeping it as a deprecated attribute and would throw a warning in console if it's being used.

ariellalgilmore:
> > Given @ariellalgilmore is changing the attribute name from `brand-name` to `platform` instead of the way around, I'm wondering if @jeffchew meant to deprecate `brand-name`. @jeffchew Is it the case...? Thanks!
> 
> Yes, if the purpose of this is to align to the prop name in React. We shouldn't ditch `brand-name` entirely though as it's breaking, I was suggesting keeping it as a deprecated attribute and would throw a warning in console if it's being used.

ah I see! Thanks @jeffchew, just updated!

kennylam (approved):
LGTM thanks @ariellalgilmore!

Issue #5287, opened 2021-02-25
MH Search typeahead is being affected on us-en ibm.com landing page


jmkraus213:
Brief description:  We have been noticing an issue with typeahead in the search box on the us-en ibm.com landing page. It appears queries are getting 'overridden?' somehow in typeahead when the search box is used.  This behavior is not happening all of the time but for me personally it's happening pretty frequently.  We did test other locales and don't see the issue we are seeing on the us-en page.  We fear it may be poorly affecting the overall Click Through Rate for masthead search as this 'bug' is changing the visitors' queries.

Is this a feature request (new component, new icon), a bug, or a general issue? Bug

> Is this issue related to a specific component? Search box 

What did you expect to happen? What happened instead? What would you like to see changed? Queries should not be overridden

What browser are you working in? Chrome

What offering/product do you work on? Any pressing ship or release dates we should be aware of? Search team

## Steps to reproduce the issue

1. Go to https://www.ibm.com
2. Perform a search and press enter

## Additional information

https://user-images.githubusercontent.com/79653961/109180114-8d840900-7758-11eb-8b04-462c8eea5627.mov

jeffchew:
Hi @jmkraus213 ,

As an update, we believe we may have found a fix for this issue, and have kicked off a patch release process for this. The behavior of the search was to highlight the first item in the results list once results came back, which we have removed in this patch. This issue became more noticeable as the time for results to return from the typeahead API has increased.

Here is our pinned release plan which you can track. We will be publishing a patch release and announcing to the support slack channel shortly.

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5288

cc: @RaphaelAmadeu @rsebade @ljcarot @RobertaJHahn

PR #5292, opened 2021-02-25
fix(masthead-search): removed focus on first item prop

RaphaelAmadeu:
### Related Ticket(s)

#5287 

### Description

I've removed the prop that makes Autosuggest focus on the first suggested item, this prop was causing a bug where the typed content would be overwritten by the first suggested item when enter is pressed.


jeffchew (approved):
The fix looks super straight forward, thank you @RaphaelAmadeu !

Issue #5294, opened 2021-02-25
lightbox-media-viewer description displaying HTML tags 


guilhermelMoraes:
## Detailed description

Is this issue related to a specific component?
A. `lightbox-media-viewer` from the web-components package.

What did you expect to happen? What happened instead? What would you like to see changed?
A. The video description in the `lightbox-media-viewer` should not display HTML tags. 

What version of Carbon for IBM.com are you using?
A. Latest.

## Additional information

<img width="368" alt="Screen Shot 2021-02-25 at 17 37 54" src="https://user-images.githubusercontent.com/42848561/109214147-3bf77080-7790-11eb-99dc-62911f02fd9b.png">
<img width="1792" alt="Screen Shot 2021-02-25 at 17 39 02" src="https://user-images.githubusercontent.com/42848561/109214228-5b8e9900-7790-11eb-9d84-de40b384b6da.png">

PR #5295, opened 2021-02-25
fix(lightbox-m-viewer): removing html tags from video description

guilhermelMoraes:
### Related Ticket(s)

lightbox-media-viewer description displaying HTML tags #5294

### Description

Removing html tags from the video description using the `removeHtmlTagEntities` util.


kennylam (approved):
Looks good thanks @guilhermelMoraes!

Issue #4936, opened 2021-01-20
Web component: Masthead - Hamburger menu, user need to scroll down to see the back link.


Praveen-sr:
## Detailed description

Issue:
On Hamburger menu, user need to scroll down to see the 'back' link from sub menus.

## Additional information

![MastheadScr.gif](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/0bcc4aad-645a-4a8c-9c89-520ad2aae74d)

PR #5296, opened 2021-02-25
fix(masthead): keep back button fixed in place upon scroll

IgnacioBecerra:
### Related Ticket(s)
#4936

### Description
In the Masthead hamburger menu, the back button scrolled out of view along with the other elements within the `<ul>`. This change ensures the back button stays in place so there is no need to scroll back up to click the button.

### Changelog

**New**

- added `position: sticky` and few other properties to the back menu link class


jeffchew (approved):
LGTM!

Issue #4890, opened 2021-01-13
Web component: Footer - language selector - Footer region is not present in Web components. Where as footer region is present in React.


Praveen-sr:
## Detailed description

Issue:
Footer region is not present in Web components. Where as footer region is present in React.

Expected:
Just like the react Env, Provide the footer section within the footer region for web component as well.

IgnacioBecerra:
@Praveen-sr Development is finished. There are eight slide effects you can use, select the specific story variation to test out each effect. 

The testing environments can be found here:

Web Components:
https://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-scroll-animations--slide-up

React:
https://ibmdotcom-react-canary.mybluemix.net/?path=/story/components-fade-in-out--slide-up
(In case the previous link doesn't work, try the following one. The previous component title was left in by accident, it might be fixed by the time you get to this)
https://ibmdotcom-react-canary.mybluemix.net/?path=/story/components-scroll-animations--slide-up

The effects can be better appreciated by opening each story in its own tab. Thanks!

cc: @RobertaJHahn

PR #5297, opened 2021-02-25
fix(footer): added aria-label to describe footer

IgnacioBecerra:
### Related Ticket(s)
#4890

### Description
Even though `<dds-footer>` has a `role='footer'`, it appears it's not being detected as such. Added an `aria-label='footer'` to accommodate for this.

### Changelog

**New**

- added `aria-label='footer'` to `<dds-footer>`


ariellalgilmore (approved):
LGTM! Thanks @IgnacioBecerra! Looks like you just need to update snapshots

Issue #4694, opened 2020-12-10
Content block headlines: Same height by CSS


asudoh:
### The problem

Currently content block headlines attempts to ensure all copy contents have the same height, by JavaScript code. It caused FOUC. FOUC causes false positives in Percy depending on when the screenshot is taken.

### The solution

Change it to CSS-based, e.g. `grid-auto-rows: 1fr` in CSS grid.

#### Acceptance Criteria
- [ ] No java script limitation for same height
- [ ] No user observable delay to apply the same height

PR #5319, opened 2021-03-01
fix(content-block-headlines): same height using CSS grid

IgnacioBecerra:
### Related Ticket(s)
#4694

### Description
The Headlines Item container currently uses `display: grid`, but each of the items' CTAs weren't properly aligned to the bottom. This PR addresses that, and also fixes a wrong class name for the heading element so the correct stylings are applied.

Before:
Screen Shot 2021-03-01 at 2 30 19 PM

After:
Screen Shot 2021-03-01 at 2 30 31 PM

### Changelog

**New**

- added `display: flex` and `flex-direction: columns` to the `headlines__item` class
- set custom class name to the previously unnamed CTA div container
- set `margin-top: auto` for the `CTA` div container to ensure proper alignment

**Changed**

- fixed a mislabeled class name for HeadlineItem's `<h4>` to apply proper styles

**Removed**

- removed the dynamic setting of height for the copy elements that used the `setHeight` utility function

### DIFF
--- a/packages/react/src/components/ContentBlockHeadlines/ContentBlockHeadlines.js
+++ b/packages/react/src/components/ContentBlockHeadlines/ContentBlockHeadlines.js
@@ -1,18 +1,16 @@
 /**
- * Copyright IBM Corp. 2016, 2020
+ * Copyright IBM Corp. 2016, 2021
  *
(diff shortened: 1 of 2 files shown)


andysherman2121 (approved):
LGTM 👍 Nice work

PR #5326, opened 2021-03-02
feat(back-to-top): introduces back-to-top web component

kennylam:
### Related Ticket(s)

#4465 

### Description

Adds new `back-to-top` component to both `web components` and `components-react`.

### Changelog

**New**

- `back-to-top` web component and `component-react` versions


packages/web-components/examples/codesandbox/components/back-to-top/index.html
```diff
@@ -0,0 +1,36 @@
...
+    <h1>Hello World! 👋</h1>
+    <div id="app">
+      <dds-btn id="btn">Click me!</dds-btn>
+    </div>
+  </body>
+  <script type="text/javascript">
+    document.getElementById('btn').addEventListener('click', () => {
+      console.log('Button clicked!');
+    });
+  </script>
```
asudoh, line 35: Did we mean to remove this...?

packages/web-components/src/components/back-to-top/back-to-top.ts
```diff
@@ -0,0 +1,98 @@
...
+   * The viewport height
+   */
+  private _windowHeight: any | null = null;
+
+  /**
+   * Button click scrolls to top
+   */
+  // eslint-disable-next-line class-methods-use-this
+  private _handleOnClick() {
+    window.scrollTo({ top: 0, behavior: 'smooth' });
```
asudoh, line 43: Instead of using global `window`, we make it clear the relationship between the component DOM and the window object:

```suggestion
    this.ownerDocument!.defaultView!.scrollTo({ top: 0, behavior: 'smooth' });
```

packages/web-components/src/components/back-to-top/back-to-top.ts
```diff
@@ -0,0 +1,98 @@
...
+   */
+  private _handleOnScroll() {
+    if (this._showBackToTop()) {
+      if (document.body.scrollTop > this._windowHeight || document.documentElement.scrollTop > this._windowHeight) {
+        this.hidden = false;
+      } else {
+        this.hidden = true;
+      }
+    }
+  }
```
asudoh, line 57: Querying for layout for every instance of `onscroll` event is expensive; I suggest throttling it (an easy way to do that is `lodash.throttle`)

packages/web-components/src/components/back-to-top/back-to-top.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+   * Show button only when document height is 3x greater than viewport
+   */
+  private _showBackToTop() {
+    return this._bodyHeight * 3 > this._windowHeight;
+  }
+
+  firstUpdated() {
+    const doc = this.getRootNode() as Document;
```
asudoh, line 67: Starting `this` to get the document object reference is a good approach - We need to pass along `{ composed: true }` to get the actual document object; Otherwise it may return a shadow root. An easier way is using `this.ownerDocument`.

packages/web-components/src/components/back-to-top/back-to-top.ts
```diff
@@ -0,0 +1,98 @@
...
+    this._bodyHeight = doc.documentElement.scrollHeight;
+    this._windowHeight = doc.documentElement.clientHeight;
+    window.onscroll = this._handleOnScroll.bind(this);
```
asudoh, line 70: I suggest adding `@HostListener('window:scroll')` decorator to `._handleOnScroll()` (requires the class inheriting [`HostListenerMixin`](https://github.com/carbon-design-system/carbon-web-components/blob/v1.11.0/src/globals/mixins/host-listener.ts)), instead of setting `onscroll` here. It ensures that the event handler is removed when this component is destroyed.

packages/web-components/src/components/back-to-top/back-to-top.ts
```diff
@@ -0,0 +1,98 @@
...
+   * Show button only when document height is 3x greater than viewport
+   */
+  private _showBackToTop() {
+    return this._bodyHeight * 3 > this._windowHeight;
+  }
+
+  firstUpdated() {
+    const doc = this.getRootNode() as Document;
+    this._bodyHeight = doc.documentElement.scrollHeight;
+    this._windowHeight = doc.documentElement.clientHeight;
```
asudoh, line 69: Looks that we need to take care of window resizing (as well as resizing in scrollable contents). `ResizeObserver` is a recommended way to handle these.

packages/web-components/src/components/back-to-top/back-to-top.ts
```diff
@@ -0,0 +1,98 @@
...
+ * Back to top
+ *
+ * @element dds-back-to-top
+ */
+@customElement(`${ddsPrefix}-back-to-top`)
+class DDSBackToTop extends StableSelectorMixin(LitElement) {
+  /**
+   * The document height
+   */
+  private _bodyHeight: any | null = null;
```
asudoh, line 31:
```suggestion
  private _bodyHeight: number | null = null;
```

packages/web-components/src/components/back-to-top/back-to-top.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+   * The viewport height
+   */
+  private _windowHeight: any | null = null;
```
asudoh, line 36:
```suggestion
  private _windowHeight: number | null = null;
```

packages/web-components/src/components/back-to-top/back-to-top.ts
```diff
@@ -0,0 +1,98 @@
...
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { html, property, customElement, LitElement } from 'lit-element';
+import ddsSettings from '@carbon/ibmdotcom-utilities/es/utilities/settings/settings.js';
+import settings from 'carbon-components/es/globals/js/settings.js';
+import UpToTop20 from 'carbon-web-components/es/icons/up-to-top/20.js';
+import StableSelectorMixin from '../../globals/mixins/stable-selector';
+// import './back-to-top-button';
```
asudoh, line 15: Did you mean to remove this or keep this...?

```suggestion
// import './back-to-top-button';
```

packages/web-components/src/components/back-to-top/__stories__/data/content.js
```diff
@@ -0,0 +1,772 @@
...
+/**
+ * Copyright IBM Corp. 2016, 2021
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+
+// TODO: replace with proper React components once react-wrapper webpack.config.js loaders are correctly configured
```
asudoh, line 11: Do you want to start with renaming this to `.tsx`? In the React wrapper Storybook environment, it's the only extension where Babel sees it as a React file.

asudoh (commented):
Thank you for working on this @kennylam!

kennylam:
Thanks for all the suggestions @asudoh!

packages/web-components/src/components/back-to-top/back-to-top.ts
```diff
@@ -0,0 +1,158 @@
...
+      this._observerResizeRoot = new ResizeObserver(this._observeResizeRoot);
+      this._observerResizeRoot.observe(this.ownerDocument!.documentElement);
+    }
+    // }
+  }
+
+  private _observeResizeRoot = () => {
+    this._bodyHeight = this.ownerDocument!.documentElement.scrollHeight;
+    this._windowHeight = this.ownerDocument!.documentElement.clientHeight;
+  };
```
asudoh, line 96: And we need another `ResizeObserver` that looks at `<body>`, and set `._bodyHeight` in the new observer.

```suggestion
  private _observeResizeRoot = entries => {
    this._windowHeight = entries[entries.length - 1].contentRect.height;
  };
```

asudoh (commented):
Getting close - Thank you for the update @kennylam!

packages/web-components/src/components/back-to-top/back-to-top.ts
```diff
@@ -0,0 +1,165 @@
...
+   * Cleans-up and creats the resize observer for the scrolling container.
+   *
+   * @param [options] The options.
+   * @param [options.create] `true` to create the new resize observer.
+   */
+  private _cleanAndCreateObserverResize({ create }: { create?: boolean } = {}) {
+    if (this._observerResizeRoot) {
+      this._observerResizeRoot.disconnect();
+      this._observerResizeRoot = null;
+    }
```
asudoh, line 84: Let's add clean-up code for `._observerResizeBody`.

asudoh (commented):
Just one thing - Thanks @kennylam!

packages/web-components/src/components/back-to-top/back-to-top.ts
kennylam: Good catch thanks @asudoh 👍

asudoh (approved):
LGTM 👍 - Thanks @kennylam for all the updates!

PR #5330, opened 2021-03-02
chore(toc): fix autoid interpolation

kennylam:
### Description

Removes extra character in `data-autoid` attribute.

### Changelog

**Removed**

- extra `}` in `TOC` mobile options


annawen1 (approved):
good catch

Issue #5058, opened 2021-02-02
React: Leaving IBM - On mobile, landscape mode - continue button is not accessible


Praveen-sr:
## Detailed description

On iPhone iOS 14
 
Issue:
On mobile, landscape mode - continue button is not accessible

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/d8d26756-30ba-4c5c-8589-f9626986a794)

PR #5331, opened 2021-03-02
fix(leavingIbm): max height limit to modal in landscape

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5058

### Changes

Limited the max height of modal to screen height. Affects landscape mode for touch devices only.


IgnacioBecerra (approved):
Just tested the deploy links using an iPhone X, bug has been fixed. Great job Julio!

Issue #5340, opened 2021-03-02
[React] Integration test bundle failures


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

There appears to be failures in the React integration test bundle that does not appear to be related to any component code:

<img width="1602" alt="Screen Shot 2021-03-02 at 4 55 12 PM" src="https://user-images.githubusercontent.com/1641214/109727351-3203b200-7b82-11eb-9c05-5e32a9e77f52.png">

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

React integration bundle test scripts

## Steps to reproduce the issue

1. Go to `/packages/react`
2. run `yarn test:integration:bundle`

## Additional information

- This might be related to an updated sub-dependency, but still needs to be investigated further.
- The test is turned off in PRs as required for now. This should be re-enabled once a fix is found.

PR #5341, opened 2021-03-02
fix(integration-bundle): locking dependency versions for React package

jeffchew:
### Related Ticket(s)

Refs #5340 

### Description

In a recent failed integration bundle test, it was discovered that the
dependency package `@babel/runtime` was updated, which has a
sub-dependency that may have been causing the build failures. The fix
around this is to set the explicit version that is being used by the
React package.

UPDATE: It was discovered that the specific latest version of 
`@babel/preset-env` was causing issues where the version being passed 
was not the correct value, causing a breakage when passed into `semver`.
 This moves the resolution to one patch version back, which appears to 
 fix the issue. This is in the codesandbox examples used in the integration
bundle tests.

### Changelog

**Changed**

- React `package.json`
- csb examples package.json configurations


jeffchew:
This PR is waiting on https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/5350 before this can be merged in. The integration build test picked up an actual bug after the initial bug in the test was fixed, which #5350 is addressing.

asudoh (approved):
LGTM 👍 - Thanks @jeffchew!

kennylam (commented):
@jeffchew Looks like the error is still happening :(

jeffchew:
> @jeffchew Looks like the error is still happening.

Yeah, I noticed, and in a different place. I'm going to take a look if I need to adjust another test too.

jeffchew:
Closed this in favor of https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/5366. Thank you @kennylam !

Issue #5342, opened 2021-03-03
[Web Components] Build compilation error in Web Components test page


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

In the Carbon for IBM.com Web Components HTML test page, a recent change has caused the follow error:

![Screenshot 2021-03-02 at 8 37 17 PM](https://user-images.githubusercontent.com/1641214/109739236-4bfbbf80-7b97-11eb-8dc3-bbf8d22c61a0.png)

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

Unknown

> What did you expect to happen? What happened instead? What would you like to
> see changed?

No more build errors

## Steps to reproduce the issue

1. See: https://github.com/carbon-design-system/carbon-for-ibm-dotcom-web-components-test/runs/2018372267?check_suite_focus=true

## Additional information

- This might be related to a PR that was merged in on 2021-03-02.

PR #5343, opened 2021-03-03
fix(video-player): remove usage of React classnames

asudoh:
### Related Ticket(s)

Refs #5342.

### Description

Replaces usage of React `classnames` library with `lit-html`'s `classMap`, given the latter is the library we use for
`@carbon/ibmdotcom-web-components` and thus our user's environment of `@carbon/ibmdotcom-web-components` won't have React `classnames`.

### Changelog

**Changed**

- Replaces usage of React `classnames` library with `lit-html`'s `classMap`.

### DIFF
--- a/packages/web-components/src/components/video-player/video-player.ts
+++ b/packages/web-components/src/components/video-player/video-player.ts
@@ -8,7 +8,7 @@
  */
 
 import { html, property, customElement, LitElement } from 'lit-element';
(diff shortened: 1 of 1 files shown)


jeffchew (approved):
Thank you for fixing this @asudoh !

jeffchew:
@asudoh can you check on the ci-check failures for web components?

PR #5350, opened 2021-03-03
fix(linkwithicon): remove inverse styles

kennylam:
### Description

Removes unused style that is also causing compile errors for web components.

### Changelog

**Removed**

- `inverse` styling. It is not being used and the scss concatenation is not supported with `:host` pseudo-classes.


jeffchew (approved):
Thanks @kennylam !

Issue #5125, opened 2021-02-09
need a way to set REACT_APP_PROFILE_HOST environment variable


yellowdragonfly:
carbon ibm.com V1.15.0.
We are using ProfileAPI to get the user ibm.com login status. Currently the call points to production environment url. Although the call reads in PROFILE_HOST environment variable to point to different environment, the variable is not a valid REACT environment variable name. So need a way for carbon ibm.com to read in the variable as REACT_APP_PROFILE_HOST in the development environment.

#### Additional Info
- Assess the existing environment variables.  Determine if adding REACT_APP support makes sense.

jeffchew:
This will be in the `v1.17.0` release.

PR #5353, opened 2021-03-04
feat(services): added support for REACT_APP_PROFILE_HOST

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5125

### Description

added support for process.env.REACT_APP_PROFILE_HOST while keeping support for process.env.PROFILE_HOST


annawen1:
@NSDrowned I was thinking we also need to add support for the `KALTURA_PARTNER_ID` and `KALTURA_UICONF_ID` as well possibly so those using create react app can use the video player service

NSDrowned:
@annawen1, sure, let me look into it

packages/services/src/services/VideoPlayer/VideoPlayer.js
```diff
@@ -14,15 +14,21 @@ import root from 'window-or-global';
  * @type {number}
  * @private
  */
-const _partnerId = process.env.KALTURA_PARTNER_ID || 1773841;
+const _partnerId =
+  process.env.REACT_APP_KALTURA_PARTNER_ID ||
```
jeffchew, line 18: @NSDrowned can you add in a check if `process` exists? This would cause errors if not running on Node.

packages/services/src/services/VideoPlayer/VideoPlayer.js
```diff
@@ -14,15 +14,21 @@ import root from 'window-or-global';
...
 
 /**
  * Sets the Kaltura UIConf ID, set by environment variable "KALTURA_UICONF_ID"
  *
  * @type {number}
  * @private
  */
-const _uiConfId = process.env.KALTURA_UICONF_ID || 27941801;
+const _uiConfId =
+  process.env.REACT_APP_KALTURA_UICONF_ID ||
```
jeffchew, line 29: Same for here, check if `process` exists.

packages/services/src/services/VideoPlayer/VideoPlayer.js
NSDrowned: You're right, just added the check.

jeffchew (approved):
LGTM, thank you @NSDrowned !

Issue #5247, opened 2021-02-19
Web Component: [Video Player] Implement the onClick trigger - consulting support


RobertaJHahn:
#### User Story

> As a `[user role below]`:
IBM.com Library developer

> I need to:
change the `Video Player`

> so that I can:
provide the ibm.com adopter developers components they can use to build ibm.com web pages

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- Create with Shadow DOM and Custom Elements standards
- **See the Epic for the Design and Functional specs information**
 - video player should have a knob that triggers Media viewer onClick, instead of playing the video in place.
- [Video Player functional specs](https://ibm.ent.box.com/file/781903492361?s=y8py6laddi9jslvxbt30kvinkvfwfalp)
- Functional specifications issue  (#5246 )
- Prod QA testing issue (#5265 )
- Visual QA issue (#5264)

---- 

- Design is exactly the same as [default Video player](https://ibmdotcom-web-components.mybluemix.net/?path=/story/components-video-player--default), with video title and caption 
- Similar functionality as [CTA - video card](https://ibmdotcom-web-components.mybluemix.net/?path=/story/components-cta--card), trigger lightbox media viewer
- Pulling cover image from Kaltura API 
- Keep original original video's aspect ratio 
- Development by AEM, @RaphaelAmadeu consult.  Engineering 5 story points (Putra assigned) and Raph consulting 3 story points.
- @RaphaelAmadeu @andysherman2121 Link the AEM created PR to this issue and move it though the development pipelines as we normally do.  

#### Acceptance criteria
- [ ] Include README for the web component and corresponding styles
- [ ] Create Web Components styles in styles package
- [ ] Do not create knobs in Storybook that include JSON objects
- [ ] Break out Storybook stories into multiple variation stories, if applicable
- [ ] Create codesandbox example under `/packages/web-components/examples/codesandbox` and include in README
- [ ] Minimum 80% unit test coverage
- [ ] Update the Carbon for ibm.com website component [data file](https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/blob/master/src/data/components.json) to be sure the web site Component Status and Storybook links are correct
- [ ] If a design is provided, the Designer is included as a Reviewer in the Pull Request
- [ ] Provide a direct link to the deploy preview for the designer in the Pull Request description
- [ ] A comment is posted in the Design QA issue, tagging Designer, when development is finished
[...]

Issue #4916, opened 2021-01-19
Web Components:  Document some of the details of the Web Components codebase


asudoh:
### The problem

`@carbon/ibmdotcom-web-components` and `@carbon/ibmdotcom-services-store` have some new technical approaches in their implementation details.

### The solution

Document them so new developers can easily jump into the codebase.

PR #5357, opened 2021-03-04
chore(video-player): misc enhancements/fixes

asudoh:
### Related Ticket(s)

Refs #4916.
Refs #5247.

### Description

* Changes `._activateEmbeddedVideo` from `private` to `protected` so it can be called from the inherited classes
* Supports re-playing the embedded video, for the scenario of lightbox being re-opened
* Changes the approach of "autoplay always" of `<dds-lightbox-video-player-composite>` so it no longer needs `@ts-ignore`
* Defers creating the DOM of embedded video in `<dds-lightbox-video-player-composite>` until the lightbox is first opened
* Supports `dds-video-player-content-state-changed` event in `<dds-lightbox-video-player-composite>` as a way of opening lightbox

Also updates `IMPLEMENTATION_NOTES.md` to explain more behind-the-scene stuffs of video player.

### Changelog

**Changed**

* Changes `._activateEmbeddedVideo` from `private` to `protected` so it can be called from the inherited classes
* Supports re-playing the embedded video, for the scenario of lightbox being re-opened
* Changes the approach of "autoplay always" of `<dds-lightbox-video-player-composite>` so it no longer needs `@ts-ignore`
* Defers creating the DOM of embedded video in `<dds-lightbox-video-player-composite>` until the lightbox is first opened
* Supports `dds-video-player-content-state-changed` event in `<dds-lightbox-video-player-composite>` as a way of opening lightbox


packages/web-components/src/components/video-player/video-player.ts
```diff
@@ -37,7 +37,7 @@ class DDSVideoPlayer extends FocusMixin(LitElement) {
    * Handles `click` event on the video thumbnail.
    */
   private _handleClickOverlay() {
-    this.contentState = VIDEO_PLAYER_CONTENT_STATE.VIDEO;
+    // this.contentState = VIDEO_PLAYER_CONTENT_STATE.VIDEO;
```
kennylam, line 40: Thanks @asudoh! Only thing is it looks like you forgot to remove this?

```suggestion
```

asudoh: It was temporary comment-out and I meant to bring it back. Good catch @kennylam!

kennylam (approved):
LGTM thanks @asudoh!

Issue #4174, opened 2020-10-08
Web Components: Align to non-gutter grid line and use CSS grid


asudoh:
### The problem

Some components, like feature card and video player, does not work unless they are put in a Carbon grid or a similar style. OTOH, some components, like callout, creates grid by itself.

### The solution

Understand the rationals behind it. Some possibilities:

* Remove the requirement of grid if possible
* Clearly document the grid requirement
* Having feature card, video player, etc. render grid by itself

jeffchew:
After investigation, @asudoh notes that this is more like 20+ story points and should be converted into an epic. The team discussed and we are ok to move this out of #4100 and will target a future minor release (tentatively v1.1.0).

@RobertaJHahn let's add this to the next engineering refinement meeting to break this work down, tentatively begin the work in sprint 21-02.

PR #5363, opened 2021-03-04
feat(content-item-horizontal-media): introducing content item horizontal media web component

iwhitcomb:
### Related Ticket(s)

[#5318](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5318)

### Description

This adds a configurable **content-item-horizontal-media** component which is a variant of the **content-item-horizontal** component.

Screen Shot 2021-02-20 at 2 39 36 PM

### Changelog

**New**

- content-item-horizontal-media


jeffchew:
Thank you @iwhitcomb for the contribution! I think @andysherman2121 and @RaphaelAmadeu can provide a thorough review, but just a quick thing if you can check the ci-checks for web components. Also, can you document in the README on how to use the media slot?

packages/styles/scss/components/content-item-horizontal-media/_content-item-horizontal-media.scss
```diff
@@ -0,0 +1,202 @@
...
+          @include carbon--breakpoint(md) {
+            @include carbon--make-col(6, 8);
+          }
+
+          @include carbon--breakpoint(lg) {
+            @include carbon--make-col(7, 10);
+          }
+        }
+      }
+    }
```
asudoh, line 196: In content patterns, we are under [the effort of transitioning to native CSS grid](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/4174). The styling strategy can be found at https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.16.1/packages/web-components/.storybook/_simple-grid.scss#L12-L32 and https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.16.1/packages/web-components/.storybook/container.scss#L77-L141. Would you try...? Thanks!

asudoh (commented):
Thank you for jumping in @iwhitcomb!

packages/styles/scss/components/content-item-horizontal-media/_content-item-horizontal-media.scss
iwhitcomb: @asudoh I went ahead and refactored the styles to use CSS grid. This also allowed me to remove the conditional inside my `content-item-horizontal-media.ts` that was flipping the arrangement of the columns in the markup. The left/right layouts are controlled by the grid now which also makes it easier to place the image on top with the text on the bottom for mobile.

packages/styles/scss/components/content-item-horizontal-media/_content-item-horizontal-media.scss
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  .#{$prefix}--content-item-horizontal__row {
+    display: grid;
+    grid-template-columns: repeat(4, 1fr);
+
+    @include carbon--breakpoint('sm') {
+      grid-template-columns: repeat(4, 1fr);
+      padding-left: 1rem;
+      padding-right: 1rem;
```
asudoh, line 118: Carbon grid doesn't define left/right margin (padding) in `sm` breakpoint, does your design spec different from this...?

packages/styles/scss/components/content-item-horizontal-media/_content-item-horizontal-media.scss
```diff
@@ -0,0 +1,210 @@
...
+          grid-column: 1 / span 4;
+        }
+
+        @include carbon--breakpoint('md') {
+          grid-column: 1 / span 6;
+          padding-right: $carbon--spacing-05;
+        }
+
+        @include carbon--breakpoint('lg') {
+          grid-column: 1 / span 11;
```
asudoh, line 150: Odd number of column span is rare for our Carbon grid usage, wanted to double-check if this was your intent.

packages/styles/scss/components/content-item-horizontal-media/_content-item-horizontal-media.scss
```diff
@@ -0,0 +1,210 @@
...
+
+        @include carbon--breakpoint('md') {
+          grid-row: 1;
+          grid-column: 7 / span 2;
+          padding-left: $carbon--spacing-05;
+        }
+
+        @include carbon--breakpoint('lg') {
+          grid-row: 1;
+          grid-column: 12 / span 5;
```
asudoh, line 168: Odd number of column span is rare for our Carbon grid usage, wanted to double-check if this was your intent.

packages/styles/scss/components/content-item-horizontal-media/_content-item-horizontal-media.scss
```diff
@@ -0,0 +1,210 @@
...
+          @include carbon--breakpoint('sm') {
+            grid-column: 1 / span 4;
+          }
+
+          @include carbon--breakpoint('md') {
+            grid-column: 3 / span 6;
+          }
+
+          @include carbon--breakpoint('lg') {
+            grid-column: 6 / span 11;
```
asudoh, line 186: Odd number of column span is rare for our Carbon grid usage, wanted to double-check if this was your intent.

packages/styles/scss/components/content-item-horizontal-media/_content-item-horizontal-media.scss
```diff
@@ -0,0 +1,210 @@
...
+          @include carbon--breakpoint('sm') {
+            grid-column: 1 / span 4;
+          }
+
+          @include carbon--breakpoint('md') {
+            grid-column: 1 / span 2;
+          }
+
+          @include carbon--breakpoint('lg') {
+            grid-column: 1 / span 5;
```
asudoh, line 200: Odd number of column span is rare for our Carbon grid usage, wanted to double-check if this was your intent.

asudoh (commented):
Thank you for the update @iwhitcomb! At this point I want our designers to chime in.

shixiedesign:
Hi @iwhitcomb thank you! A few issues: 
- [ ] Left alignment toggle is showing text **right** aligned. That's not supposed to happen. Expecting text left align, but positioned on the left. 
image

- [ ] Video caption position was impacted. Should be left aligning to video.
image

- [ ] When media = video, the media position is wrong. Should be top aligning to eyebrow text
image

- [ ] Link list alignment wrong. Should be 32px (layout-03) from the content body copy.
image
image


- [ ] Please name the story with correct capitalization in Storybook left nav: `Content item horizontal` and sub story should be named `with media`. Sorry, this is a total knitpick and yes a lot of errors in the current left nav! I'm trying to get our storybook cleaned up gradually.

iwhitcomb:
- [x] Left alignment toggle is showing text right aligned. That's not supposed to happen. Expecting text left align, but positioned on the left.
- [x] Default should not include Eyebrow text. Please remove from story example. And when Eyebrow text is empty. Top align headline and media
- [x] Missing spacing between headline and body copy
- [x] Video caption position was impacted. Should be left aligning to video.
- [x] When media = video, the media position is wrong. Should be top aligning to eyebrow text
- [x] Link list alignment wrong. Should be 32px (layout-03) from the content body copy.
- [ ] Please name the story with correct capitalization in Storybook left nav: `Content item horizontal` and sub story should be named `with media`. Sorry, this is a total knitpick and yes a lot of errors in the current left nav! I'm trying to get our storybook cleaned up gradually.

@shixiedesign I was able to take care of all of your feedback. I was able to rename the `Content item horizontal` component, but `With Media` appears to be done programmatically across every components so I left it alone.

packages/styles/scss/components/content-item-horizontal-media/_content-item-horizontal-media.scss
```diff
@@ -0,0 +1,162 @@
...
+
+    @include carbon--breakpoint(lg) {
+      min-height: rem(296px);
+      padding-top: $carbon--spacing-07;
+      padding-bottom: $carbon--layout-05;
+    }
+
+    padding-left: $carbon--spacing-05;
+    padding-right: $carbon--spacing-05;
+  }
```
asudoh, line 38: Just to double-check - Did you go back to `carbon--make-row` because of the gutter thing, or something else? If it's the gutter thing, we can use [`gap`](https://developer.mozilla.org/en-US/docs/Web/CSS/gap).

asudoh (commented):
Thank you for the update @iwhitcomb!

jeffchew:
This looks really good, thank you @iwhitcomb !

One minor thing, it doesn't look like CTA 2 is used for the `with Media` variant. Can you suppress those knobs for that story?

+ @RaphaelAmadeu @andysherman2121

packages/web-components/src/components/content-item-horizontal/__stories__/README.stories.mdx
```diff
@@ -18,43 +18,65 @@ Here's a quick example to get you started.
...
+    >Lorem ipsum dolor sit amet, _consectetur_ adipiscing elit. Aenean et ultricies est. Mauris iaculis eget dolor nec hendrerit.
+    Phasellus at elit sollicitudin.</dds-content-item-horizontal-copy
+  >
+  <dds-link-list slot="footer" type="vertical">
+    <dds-link-list-item-cta icon-placement="right" href="https://www.ibm.com" cta-type="local">
+      Learn more
+    </dds-link-list-item-cta>
+    <dds-link-list-item-cta icon-placement="right" href="https://www.ibm.com" cta-type="external">
+      Microservices and containers
+    </dds-link-list-item-cta>
```
jeffchew, line 72: In the example there's only one CTA, perhaps can remove in the Docs as well? I don't recall if it was the intent to have more than one CTA here.

jeffchew: Taking this back, I saw there's more than one CTA in the designs!

shixiedesign:
Thank you Ian! Again, very impressed with the speed and accuracy. A few comment, and some might not apply so let me know. Point 1 & 2 are regarding grid. Is there grid on this component or does it only get the grid when it gets assembled into a section? The other points are just storybook refinement. We are trying to limit what becomes a knob so there appears to be a few extras here.

1. I'm seeing on `sm` breakpoint, the component should not go full width. Expected to leave 16px padding on left and right 
Current:
image
Spec
image

2. On max and above, it's spec'ed to leave a column empty 
Current:
image
Spec
image

3. Storybook knobs refinement
image

iwhitcomb:
@shixiedesign 
- [ ] I'm seeing on sm breakpoint, the component should not go full width. Expected to leave 16px padding on left and right.
  - **I think this is probably an issue with the way the grid is setup. You've got padding setup around the grid at every breakpoint except `sm`. I can certainly go ahead and apply the padding just for this component, but I'm wondering if it would be more appropriate to do it globally instead?** 
- [x]  On max and above, it's spec'ed to leave a column empty
- [x] Storybook knobs refinement

jeffchew (approved):
I think this looks great, thank you for this @iwhitcomb !

photodow:
Quick glance I saw there is a grid alignment issue...

image

iwhitcomb:
@photodow I think what you're seeing is a result of the gutters in the component vs. no gutters in the page container. I had to use a 12-column grid at the max breakpoint to achieve that single empty column after the video/image.

Screen Shot 2021-03-10 at 1 23 56 PM
Screen Shot 2021-03-10 at 1 24 27 PM

If you size down to the `lg` breakpoint the container switches over to a 3-column grid that lines up a bit better with the main grid.

Screen Shot 2021-03-10 at 1 23 35 PM

photodow:
@iwhitcomb it doesn't appear to align on any of the breakpoints. `lg` included. It could be the lack of the gutter from the container. If that's the case, Is there a reason we wouldn't include a container so it falls on the grid? Helps with QA, helps it not look like a mistake.

@jeffchew am I seeing this in a number of our stories? When did we remove the container/grid from our stories?

image

iwhitcomb:
@photodow I think that's exactly what it is. When I apply the same gutter to the container everything lines up.

Screen Shot 2021-03-10 at 3 49 01 PM
Screen Shot 2021-03-10 at 3 49 18 PM

The question then becomes whether or not this is an issue with the container not having gaps, or is the issue that the design spec for this component had gaps?

jeffchew:
> @iwhitcomb it doesn't appear to align on any of the breakpoints. `lg` included. It could be the lack of the gutter from the container. If that's the case, Is there a reason we wouldn't include a container so it falls on the grid? Helps with QA, helps it not look like a mistake.
> 
> @jeffchew am I seeing this in a number of our stories? When did we remove the container/grid from our stories?

This may have been part of the recent grid work we were doing across multiple stories but I may be mistaken.

Either way, I'm wondering if this is good enough at least to get the production level testing started. We can address the grid usage probably in another PR.

asudoh:
Seems that `<dds-content-item-horizontal-media>` itself has the right grid alignment. The component needs to manage the paddings by its own (it comes from the transition from guttered grid in container to non-guttered one as @jeffchew mentioned). As @jeffchew said, addressing it in a separate PR sounds fine to me (probably an issue tracking it would be worth it).

PR #5366, opened 2021-03-04
fix(integration-bundle): use explicit version for parcel-bundler

kennylam:
### Related Ticket(s)

#5340 

### Description

Sets an explicit version for `parcel-bundler` being used Codesandbox examples. The latest version relies on an API from `@babel/present-env`, which is not available and causes this error. Since `integration:bundle` uses our Codesandbox examples in `jest`, the CI workflow would fail.

### Changelog

**Changed**

- all Codesandbox examples explicitly use `parcel-bundle@1.12.3`


jeffchew (approved):
Awesome thank you for getting this in! I can close my PR if you didn't already.

PR #5375, opened 2021-03-04
feat(translation): add env vars to allow diff translation endpoint/sessionkey

annawen1:
### Related Ticket(s)

### Description

In order for the HC team to utilize their own masthead data, we need to allow for different translation service endpoint and session key.

Endpoint:
Use env var `TRANSLATION_ENDPOINT` to specify different path to fetch translation data.

Session Key:
We cache the translation data in session storage with a key in the format of `${key}-${country}-${lang}` (ie. dds-translation-us-en).
Use env var `TRANSLATION_SESSIONKEY` to allow for different key so the data fetched from the HC endpoint can be cached and accessed.

### Changelog

**New**

- `TRANSLATION_SESSIONKEY` env var for translation caching
- `TRANSLATION_ENDPOINT` env var for translation service endpoint

### DIFF
--- a/packages/services/src/services/Translation/Translation.js
+++ b/packages/services/src/services/Translation/Translation.js
@@ -24,7 +24,8 @@ const _host =
  * @type {string}
  * @private
  */
-const _endpoint = `${_host}/common/v18/js/data/jsononly`;
+const _endpoint = `${_host}${(process && process.env.TRANSLATION_ENDPOINT) ||
+  '/common/v18/js/data/jsononly'}`;
 
 /**
  * Session Storage key for translation data
@@ -32,7 +33,8 @@ const _endpoint = `${_host}/common/v18/js/data/jsononly`;
  * @type {string}
  * @private
  */
-const _sessionTranslationKey = 'dds-translation';
+const _sessionTranslationKey =
+  (process && process.env.TRANSLATION_SESSIONKEY) || 'dds-translation';
 
 /**
  * The cache for in-flight or resolved requests for the i18n data, keyed by the initiating locale.


annawen1:
@proeung This is the draft PR for the env var changes to the translation service to allow for HC to pull in masthead data from your own endpoint.

proeung (commented):
@annawen1 Thanks for adding the environment variable that enables us to pass our HC menu endpoint and session key. 

Is there a follow-up PR or commit for exposing an attribute within `<dds-masthead-container>` that fetches the data from env var `TRANSLATION_ENDPOINT`?

cc: @huntermacd @mherchel

annawen1:
> @annawen1 Thanks for adding the environment variable that enables us to pass our HC menu endpoint and session key.
> 
> Is there a follow-up PR or commit for exposing an attribute within `<dds-masthead-container>` that fetches the data from env var `TRANSLATION_ENDPOINT`?
> 
> cc: @huntermacd @mherchel

@proeung yeah good question, I mentioned adding an attribute to the masthead-container yesterday but since we're looking to keep the same data structure just changing the endpoint to the translation service itself should be enough.

proeung:
@annawen1 That's a great point! I think altering the `TRANSLATION_ENDPOINT` variable to point to the new Cloud menu translation data works great in the Storybook preview and etc, however, I think we're looking into exposing an attribute within the `<dds-masthead-container>` that allows any content ecosystems outside of the JS frameworks (Drupal, EI, AEM, etc) to define which data endpoint the masthead container Web Component custom element should be fetched from.

I'm not sure how easy this purposed resolution would be, but I think this gives users/adopters an easy way to use Cloud Masthead/L0 data aside from the dotcom version. Let me know what you think.

```
<dds-masthead-container
            plaform=""
            data-endpoint="cloud" 
            selected-menu-item=""
            user-status=""
          ></dds-masthead-container>
```

proeung:
@annawen1 I added a feature request ticket (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5396) for the work mentioned in this comment (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/5375#issuecomment-791517267). Please feel free to close out this PR or open a new one based on what we've discussed. Thanks! :)

Issue #5396, opened 2021-03-05
feat(masthead-container): Pass in the data endpoint as an attribute within the masthead-container Web Components


proeung:
### The problem
- The Hybrid Cloud team would like to leverage the existing `<dds-masthead-container>` component, however, the existing work does not have an attribute that allows us to pass in the menu data endpoint that's Cloud specific.

### The solution

- If possible, we'd like to expose an attribute within the `<dds-masthead-container>` that allows any content ecosystems (Drupal, EI, AEM, etc) to define which data endpoint the `masthead-container` Web Component custom element should be fetched from.

I'm not sure how easy this purposed resolution would be, but I think this gives users/adopters an easy way to use Cloud Masthead/L0 data aside from the default dotcom version. Let me know what you think.

Example:

`<dds-masthead-container data-endpoint="/common/carbon-for-ibm-dotcom/translations/cloud-masthead"></dds-masthead-container>`

PR #5407, opened 2021-03-08
feat(translate): accept diff endpoint arg for translation service

annawen1:
### Related Ticket(s)

feat(masthead-container): Pass in the data endpoint as an attribute within the masthead-container Web Components #5396

### Description

Allow adopters to set the translation endpoint in the masthead as an attribute (ie. `<dds-masthead-container data-endpoint="/common/carbon-for-ibm-dotcom/translations/cloud-masthead">`)

This requires passing in an `endpoint` arg into the `loadTranslation()` action and then to the translation service methods. The session key required for caching the data will be derived from the endpoint by removing all special characters from the endpoint string.

Also changed default translation endpoint as we have copied over our translation files to a new folder

### Changelog

**New**

- set up new arg `dataEndpoint` for `loadTranslation()` action in the services-store
- set up new arg `endpoint` for `fetchTranslation` and `getTranslation` methods in the translation service

**Changed**

- change default translation endpoint


packages/services/src/services/Translation/Translation.js
```diff
@@ -110,10 +111,11 @@ class TranslationAPI {
    *
    * @param {string} lang Language code
    * @param {string} country Country code
+   * @param {string} endpoint endpoint to fetch data
    * @param {Function} resolve resolves the Promise
    * @param {Function} reject rejects the promise
    */
-  static fetchTranslation(lang, country, resolve, reject) {
+  static fetchTranslation(lang, country, endpoint, resolve, reject) {
```
jeffchew, line 118: @annawen1 should we also offer this up as an environment variable? Perhaps instead of `_dds_endpoint` we open up `DDS_TRANSLATION_ENDPOINT`.

The other thing we might want to track is the sessionStorage key being used, as we don't want a custom endpoint data to conflict with the default. Maybe we can:
1) Set up DDS_TRANSLATION_ENDPOINT
2) If DDS_TRANSLATION_ENDPOINT is not defined, use _dds_endpoint
3) Also have `endpoint` as an optional attribute for the service
4) logic would be `endpoint attribute` used first if defined, if not then use `DDS_TRANSLATION_ENDPOINT`
5) if `DDS_TRANSLATION_ENDPOINT` does not match the default value, set the `_sessionTranslationKey` to something different (perhaps derived somehow from the custom defined `DDS_TRANSLATION_ENDPOINT`

jeffchew:
Screen Shot 2021-03-08 at 1 33 41 PM

Just noticed a couple of private methods are appearing in the JSDoc output, can you set those to `@private`?

jeffchew:
This may or may not be part of the scope of this PR, but noticing that none of the arguments are defined as required in any of the documentation. Have to take a closer look but would any of these methods fail if nothing is passed in? https://ibmdotcom-services.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/5407/TranslationAPI.html

packages/services/src/services/Translation/Translation.js
annawen1: yep I can set the env var up, the session key I'll see what might be the best way to get the key name from the endpoint...

annawen1:
> This may or may not be part of the scope of this PR, but noticing that none of the arguments are defined as required in any of the documentation. Have to take a closer look but would any of these methods fail if nothing is passed in? https://ibmdotcom-services.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/5407/TranslationAPI.html

Don't think they will fail, only thing I can see is the lang codes - but we have them defaulting to us-en if they are not passed in.

annawen1:
@proeung draft PR to set translation endpoint through masthead attribute - still working on how to derive a session key name from endpoint if possible

packages/services/src/services/Translation/Translation.js
```diff
@@ -24,7 +24,11 @@ const _host =
  * @type {string}
  * @private
  */
-const _endpoint = `${_host}/common/v18/js/data/jsononly`;
+const _ddsEndpointDefault =
+  (process &&
+    (process.env.REACT_APP_DDS_TRANSLATION_ENDPOINT ||
+      process.env.DDS_TRANSLATION_ENDPOINT)) ||
+  '/common/v18/js/data/jsononly';
```
jeffchew, line 31: One way to approach the sessionKey logic is to probably have two variables:

```
const _ddsEndpointDefault = '/common/carbon-for-ibm-dotcom/translations/masthead-footer';

const _ddsEndpoint = (process &&
    (process.env.REACT_APP_DDS_TRANSLATION_ENDPOINT ||
      process.env.DDS_TRANSLATION_ENDPOINT)) || _ddsEndpointDefault;
```

Then:

```
if(_ddsEndpointDefault !== _ddsEndpoint) {
   ...do logic to set a different session key...
}
```

annawen1: done!

packages/services/src/services/Translation/__tests__/Translation.test.js
```diff
@@ -98,6 +98,28 @@ describe('TranslationAPI', () => {
     expect(response).toEqual(responseSuccess);
   });
 
+  it('should fetch the i18n data from given endpoint', async () => {
```
jeffchew, line 101: Not sure if this is more ambitious, but might be nice if we could test what is being set in sessionStorage, and if it's either using the default or using a custom key. We'd probably need to mock session storage though in our unit tests: https://www.npmjs.com/package/mock-local-storage

proeung (commented):
@annawen1 The code looks good so far. Thanks for working on this issue https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5396!

We're currently working on getting the `/cloud-masthead` directory added to the  `ibm-dotcom-translations` repo so that we can set a custom endpoint for the translation service to fetch from. I'll keep you posted once that's in place so that we can test out the `data-endpoint` attribute within the masthead-container component.

In the meantime, keep me posted on how you're looking to obtain the session key name from the endpoint.

cc: @huntermacd

jeffchew:
@annawen1 looks like the integration bundle is failing, might be related to the endpoint change perhaps?

packages/services/src/services/Translation/__tests__/Translation.test.js
annawen1: Used the mock session storage that was already set up to test

annawen1:
> @annawen1 looks like the integration bundle is failing, might be related to the endpoint change perhaps?

yep, had to swap the old endpoint with the new one for the integration mocking 👍

jeffchew (approved):
This is great, thank you @annawen1 ! And extra thank you for taking the time to add the unit tests!

proeung (approved):
Looks great! Thank you @annawen1 for taking the time to work on this feature. Appreciate it!

asudoh:
Nice progress @annawen1! One quick question is that whether the endpoint thing is per-component thing or per-system thing. Any thoughts? Thanks!

annawen1:
> Nice progress @annawen1! One quick question is that whether the endpoint thing is per-component thing or per-system thing. Any thoughts? Thanks!

@asudoh Hmm we initially went with the route of only providing the env vars to change the endpoint in the translation service itself, but to accommodate easier instantiation of the masthead in other applications using Drupal, AEM, etc we added the endpoint attribute to the masthead itself. Not sure this will expand further to other components, don't see that happening for now at least.

asudoh:
@annawen1 I see, thank you for your thought; My concern is e.g. masthead container and footer container pointing to different endpoints. Given how our data is cached in memory, it causes an undefined behavior. We can document such caveat, or we can create a global API (that must be called up-front) to defined the endpoint.

Issue #5030, opened 2021-01-28
[card section carousel]: small breakpoint horizontal scrolling


photodow:
## Detailed description

At the `sm` breakpoint in particular there is a horizontal scroll from what appears to be the next hanging card is still around.

## Steps to reproduce the issue

1. navigate to storybook page
2. open up carbon devtools and choose `sm` breakpoint to resize the browser.

## Additional information

<img src="https://user-images.githubusercontent.com/3793636/106178615-9ab0d680-615f-11eb-9cd2-560d9b869252.gif" width="300px" />

RobertaJHahn:
@Praveen-sr Can you retest this bug on the platforms to provide specific detials.

Praveen-sr:
@RobertaJHahn Tested this scenario on Phone, the horizontal scrolling is felt on "Carousel component and Dotcom shell".

![Caro.gif](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/28eae643-f505-444d-9052-57fb024102a1)

Dotcom Shell:
![dotcom.gif](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/d90c5996-7466-4bcd-a7d6-bb1c675d8d99)

PR #5414, opened 2021-03-09
fix(carousel): missing hint size in story

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5030

### Changes

Added missing  `--dds--carousel--overflow-hint-size: 0rem;` to story


packages/web-components/src/components/carousel/__stories__/carousel.stories.scss
```diff
@@ -26,6 +26,10 @@
 
   /* stylelint-disable-next-line selector-type-no-unknown */
   dds-carousel {
+    /* The unit is required to correctly calculate the margin/padding for page overflow UI */
+    /* stylelint-disable-next-line length-zero-no-unit */
+    --dds--carousel--overflow-hint-size: 0rem;
```
asudoh, line 31: Wondering if this should be non-zero for larger breakpoints; This dictates the size of area showing "next card", and such area is required for larger breakpoints AFAIK.

NSDrowned: You are 100% right like always @asudoh :) Just fixed it

asudoh (approved):
LGTM 👍 - Thanks @NSDrowned!

Issue #5410, opened 2021-03-08
Upgrade dependencies to Carbon v10.31.0


jeffchew:
#### User Story

> As a `[user role below]`:
Carbon for IBM.com adopter

> I need to:
load carbon 10.31.0 in my application

> so that I can:
ensure I am using the latest version of Carbon with Carbon for IBM.com

#### Additional information

 - Enable the labels `package: styles`,`RTL`, and `feature flag` to ensure we are capturing testing across
- each package has the npm script `yarn upgrade-carbon` which can be used for this task
- upgrade playlist: https://ibm.ent.box.com/notes/727078207172
- One adopter noted an upstream bug which we may want to take a closer look at once Percy is kicked off, need to evaluate if we want this in `v1.17.0` or hold off until anything gets patched upstream first.
- Separate branch needs to be created to push hotfixes to, before this is merged down to the main branch.

#### Acceptance criteria
 - [ ] All dependencies for react, web-components, and styles are updated to carbon 10.31.0
 - [ ] Activate deploy previews for carbon expressive, RTL, react experimental, and web components experimental

jeffchew:
Just to keep as a record, I have looked into possibly using a Github app in order to automate this work:
https://www.whitesourcesoftware.com/free-developer-tools/renovate
https://docs.renovatebot.com/

However, when testing it out, it ran into disk space issues right away and does not look like this can be used. I have moved on to looking into a home grown automation process for this.

jeffchew:
Putting off the automation work for now, getting this update through should be priority.

PR #5420, opened 2021-03-09
chore(upgrade-carbon): updating scripts for carbon updates

jeffchew:
### Related Ticket(s)

Refs #5410 

### Description

This updates the scripts used to update Carbon, in a way so that the
upgrades can be non-interactive. In addition, there is an introduction
to a dev dependency, syncyarnlock. This tool will align the package.json
 files to match what is currently being used in our lock file. Part of
 the motivation from using this is that when upgrading using `yarn
 upgrade` in a non-interactive way, it upgrades the `yarn.lock` file,
 but does not update our `package.json` files. This effectively doesn't
 show what version we are currently using under the hood. In addition,
 this will help in the long run on what versions are actually being used
  for all other dependencies as well, which may help with the effort to
  remove duplicate dependencies used across the packages.

### Changelog

**New**

- Added `syncyarnlock`

**Changed**

- `upgrade-carbon` npm scripts to be non-interactive
- Updated package.json files to match the versions used in the `yarn.lock` file


jeffchew:
Closing this out for now, will run the usual Carbon dependencies update and will see if I can get automation in place next sprint.

Issue #5430, opened 2021-03-10
[Masthead]: grouped search results title missing background color


kennylam:
`Masthead` grouped search results title is missing a `background-color`, allowing elements/colors behind it to show through.

![Screen Shot 2021-03-10 at 11 39 05 AM](https://user-images.githubusercontent.com/909118/110665117-0a21d900-8196-11eb-9275-f42cb745e407.png)

PR #5431, opened 2021-03-10
fix(masthead): add background color to grouped title

kennylam:
### Related Ticket(s)

#5430 

### Description

Adds background color styles to grouped search result title.

### Changelog

**Changed**

- add background color to grouped result title


annawen1 (approved):
LGTM!

PR #5440, opened 2021-03-10
fix(expressive-modal): toggle aria-hidden upon modal closing

IgnacioBecerra:
### Related Ticket(s)
#4744

### Description
This PR fixes the issue where VoiceOver can still read the Video Player controls even after modal is closed. This fix applies `aria-hidden='true'` to both the Expressive Modal, and an element within the iFrame that isn't affected by the `aria-hidden` from the parent element unless applied directly to it.

### Changelog

**New**

- toggle `aria-hidden='true'` to `<dds-expressive-modal>` and an element within the video iFrame that's unaffected by the first change alone


packages/web-components/src/components/lightbox-media-viewer/lightbox-video-player-composite.ts
```diff
@@ -35,6 +35,20 @@ class DDSLightboxVideoPlayerComposite extends ModalRenderMixin(DDSVideoPlayerCom
...
 
+  /**
+   * Handles aria state depending on the modal's state.
+   */
+  private _handleAriaState = () => {
+    const iFrame = this._videoPlayer?.querySelector('iframe');
+
+    // Handles edge case where screen reader still reads video title within iFrame
+    if (this.open) {
+      window.frames[`${iFrame?.id}`].document.querySelector('.topBarContainer').removeAttribute('aria-hidden');
```
asudoh, line 46: Also I suggest putting this logic in `try` ... `catch` block. Cross-frame DOM access is often restricted:

```suggestion
      iFrame.contentWindow.document.querySelector('.topBarContainer').removeAttribute('aria-hidden');
```

packages/web-components/src/components/lightbox-media-viewer/lightbox-video-player-composite.ts
```diff
@@ -35,6 +35,20 @@ class DDSLightboxVideoPlayerComposite extends ModalRenderMixin(DDSVideoPlayerCom
...
+    } else {
+      window.frames[`${iFrame?.id}`].document.querySelector('.topBarContainer').setAttribute('aria-hidden', 'true');
```
asudoh, line 48:
```suggestion
      iFrame.contentWindow.document.querySelector('.topBarContainer').setAttribute('aria-hidden', 'true');
```

asudoh (commented):

asudoh (approved):
LGTM 👍 with a regular caveat for `aria-hidden` (we may end up with being requested to hide elements that we don't have control over). Thanks @IgnacioBecerra!

Issue #5029, opened 2021-01-28
[carousel]: lacking swiping on mobile devices


photodow:
## Detailed description

I wasn't able to swipe left or right to progress the carousel on mobile devices.

## Steps to reproduce the issue

1. Open storybook component on mobile device
2. Swipe left and right.

## Additional information

<img src="https://user-images.githubusercontent.com/3793636/106179143-3b9f9180-6160-11eb-8da6-5b36855ff0fa.gif" width="400px" />

RobertaJHahn:
@Praveen-sr Could you test this bug to reproduce and provide more details for the platform.

Praveen-sr:
@RobertaJHahn  Tested this on phone and tab on web component & React, swiping cards left or right does not progress the card movement.
Same behavior for card carousel component. 
 Note: This is without using pagination.

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/a195a4de-3c0e-4ba0-bb92-6237061ec3b6)

PR #5444, opened 2021-03-10
fix(carousel): swipe functionality

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5029

### Description

Added swipe functionality, which includes 2 settings:
- `allowedTime`: min time allowed to do a swipe gesture (set to 300ms)
- `threshold`: min distance needed the gesture needs to travel to be considered a swipe (set to 75px)

Tested on Safari, Chrome and Firefox for iPhone.

### Changelog

**New**

- Added `handleTouchStartEvent`  and `handleTouchEndEvent` triggered by touchstart and touchend respectively

**Changed**

- Added same fix as #5030 to story to avoid the horizontal scroll bug that makes the new swipe functionality look like it's malfunctioning.


packages/web-components/src/components/carousel/carousel.ts
```diff
@@ -179,6 +191,35 @@ class DDSCarousel extends HostListenerMixin(LitElement) {
...
+  }
+
+  /**
+   * Handles `touchend` event.
+   */
+  private _handleTouchEndEvent(event: TouchEvent) {
+    const { _startPos, _startTime } = this;
+    const { pageSize, start, _total: total } = this;
+    const allowedTime = 300; // max time allowed to do swipe
+    const threshold = 75; // min distance traveled to be considered swipe
```
asudoh, line 209: Should put these constants in private class fields or out-of-the class variables.

asudoh (commented):
Thank you for working on this @NSDrowned!

packages/web-components/src/components/carousel/carousel.ts
NSDrowned: Done =)

asudoh (commented):
LGTM 👍 code wise, wanted to have designers see what they think. Thanks @NSDrowned!

shixiedesign (approved):
Works on Web components

shixiedesign (commented):
Swiping works on Web components, but not working on React wrapper. If this is expected then we are all good!

jeffchew (approved):
Thank you @NSDrowned !

Issue #5304, opened 2021-03-01
[Card in card] 


shixiedesign:
A stand alone component in Storybook. This is a dependency of Cards group variation requested by AEM (previously Featured card group)

**This component is being developed by HC team. Ref. Putra.**, estimated to be 3 points. 

### Final design spec

Link to design spec [Box folder](https://ibm.ent.box.com/folder/132489332751?s=jad2fd65p56odcme9aoy1lgmuo1xaifu)
Link to Functional specs [Box doc](https://ibm.box.com/s/zd6rq29aolbwr5hf574nxf4jli4z5slj)  

![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/aeef7838-3cd6-4e95-8c92-db8c0e748b71)

### FA notes
- Video should open in Lightbox media viewer 
- Whole card is clickable, aka hover on image/video would trigger `hover` state of "inset" text card
- Card icon should follow current CTA functionality, allowing multiple icon based on link type (email, external link, link, etc)
- Media should be 16x9 on all breakpoints except `sm`, where it is 4x3
- Potentially repurpose CTA - feature https://ibmdotcom-web-components.mybluemix.net/?path=/story/components-cta--feature

oliviaflory:
Updated the spec file on box to include additional information after the review on 3/1/20:

- Interactive states
- with grid overlay to see alignment

proeung:
@shixiedesign @oliviaflory Thanks for updating this ticket to include the FA notes and design spec file.

Here's the JIRA ticket (https://jsw.ibm.com/browse/HC-1647) of this work that the HC team is using to track this contribution work on our end. cc: @RobertaJHahn

Issue #5245, opened 2021-02-19
[Video Player] media viewer onclick enhancement


RobertaJHahn:
#### Overview
AEM team has asked for an enhancement to trigger the video player onClick, instead of playing the video in place.  This component will be used in the Solution template.

This epic will contain all of the linked issues needed to implement this change.

#### Additional Information
- [Video Player functional specs](https://ibm.ent.box.com/file/781903492361?s=y8py6laddi9jslvxbt30kvinkvfwfalp)

RobertaJHahn:
Shixie will reach out to Putra to see whether she can include the Hover state in this existing work.

PR #5447, opened 2021-03-11
feat(video-player): support card as the caption

asudoh:
### Related Ticket(s)

Refs #5304.

### Description

Adds `caption-style="card"` to `<dds-video-player>`, `<dds-video-player-composite>` and `<dds-video-player-container>`. With such new mode, the caption is rendered as a card and is clickable.

### Changelog

**New**

- `caption-style="card"` to `<dds-video-player>`, `<dds-video-player-composite>` and `<dds-video-player-container>`.
- `<dds-video-player-launcher-card-footer>`, used in `<dds-video-player>` internally.


asudoh:
CC @proeung some discussion points:

* Is the size of "caption as card" right?
* Is the size of video player right for this usage? Would separate stories be worthwhile?
* Can somebody in the team or @proeung take further styling work? CC @jeffchew

Thanks!

jeffchew:
Thank you @asudoh ! @RaphaelAmadeu @proeung this looks like this may help for the video variant of the Card in Card? We probably don't need to rush to get this merged in before our code freeze tomorrow though if we want to give ourselves more time to implement this feature.

FYI @IgnacioBecerra @RobertaJHahn

proeung:
@jeffchew Yes, I think this video player fix that @asudoh puts together could help in creating the video variant of the `Card in Card` component. For the `Card in Card` video variant, I think we need to find an elegant way to call for the `<dds-video-player>` that opens up in a lightbox and still adheres to the existing card design. I'll work with @RaphaelAmadeu to see how to best leverage what we have so far and implement this variant in a way that makes sense across the board.

proeung:
@jeffchew @asudoh Looks like there are a good amount of changes that have been applied to the `<dds-video-player>` component within this PR. Feels like this PR should be tested and merged down before this issue/ticket (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5245) can get started so that we can avoid merging conflicts. Let me know what you both think.

asudoh:
@proeung Thank you for asking. You can wait for this PR to be merged if your time allows, or you can directly bring in this code into your Git branch to start your work.

IgnacioBecerra (approved):
LGTM, thank you @asudoh!

shixiedesign (commented):
Sorry it's taking me a moment to review this one - I cannot find a preview link with `Card in card - video` story. If visual review is not needed, happy to rely on our dev's review. The `Card in card` component is added since last sprint. So if I understand correctly, this feature can be demoed in the `Card in card` component.

jeffchew:
> Sorry it's taking me a moment to review this one - I cannot find a preview link with `Card in card - video` story. If visual review is not needed, happy to rely on our dev's review. The `Card in card` component is added since last sprint. So if I understand correctly, this feature can be demoed in the `Card in card` component.

Hi @shixiedesign , the changes were made in Video Player.

oliviaflory:
Thanks for pointing out where to see the updates @jeffchew 

cc @shixiedesign @photodow adding my thoughts below, feel free to counter anything!

1. If this is going to support the Card-in-card video variant, I'd imagine the caption card needs to support the same content from Card-in-card, so when it's paired with Card group there is some consistency. Currently missing the `Label`


 2. The card should move below the video in Medium and Small breakpoints, the overlaid card is not going to work for content on small and medium breakpoints. Reference [Card in card design specs](https://ibm.ent.box.com/folder/132489332751?s=jad2fd65p56odcme9aoy1lgmuo1xaifu)

**Current**
Screen Shot 2021-03-24 at 8 57 19 AM 

**Medium**
Screen Shot 2021-03-24 at 8 58 06 AM 

**Small**
Screen Shot 2021-03-24 at 8 58 14 AM


3.  CTA is pushed into the padding on large breakpoint, needs to maintain the 16px bottom padding
Screen Shot 2021-03-24 at 8 56 35 AM

4. CTA has too much space between the duration text and play icon. It looks like the text has 8px padding on the right, and the icon also has 8px padding on the left = 16px. There should be **8px between text and icon** following our card specs
Screen Shot 2021-03-24 at 8 55 11 AM

5. CTA text should not get underline on hover, and icon is changing color which is not part of our current card specs. 
Screen Shot 2021-03-24 at 8 55 54 AM

6. I'm a little concerned about the size of the video player + card on the large breakpoint in this demo. For Card in card + card group the Card in card spans 12 columns which allows the card to have enough breathing room and content. But if the video player / card in card only spans 8 columns the card consumes so much space and would potentially cover up the play button (see mock up below). 
Not sure what the best way is to handle avoiding this issue, guidelines, restrictions in code 🤷‍♀️  Would love some thoughts from others.

**Card in card – 12 columns**
Screen Shot 2021-03-24 at 10 21 30 AM

**Card in card – 8 column mock up**
Screen Shot 2021-03-24 at 10 21 51 AM


7. Additionally, it looks like the caption card on this demo is only spanning 3 columns, but the intent is to span 4 columns so it aligns with the other cards when paired with Card group

Screen Shot 2021-03-24 at 10 24 30 AM

Screen Shot 2021-03-24 at 10 25 17 AM

Screen Shot 2021-03-24 at 10 28 40 AM

shixiedesign (changes requested):
Hey @asudoh @jeffchew I kinda finally understood what's happening here and I agree with Olivia's concerns. Expected the solve for `Card in card - video` to use [CTA - card video](https://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-cta--card) component. 

Please see [James's fun spec](https://ibm.box.com/s/zd6rq29aolbwr5hf574nxf4jli4z5slj). Using `CTA - card video` would allow Media viewer modal to be triggered onClick without additional work. 

Sorry for the angry red Request changes here. If we allow this change to the Video component, we are gonna negatively affect Content block, Content group and all things media related. 

image

jeffchew:
@asudoh @RaphaelAmadeu @proeung @andysherman2121 based on @shixiedesign and @oliviaflory 's concerns, could the approach be pivoted to be more directly making the variation to the CTA Video Card rather than the Video Player component itself, or are there more fundamental concerns on the engineering side?

asudoh:
@jeffchew `<dds-video-player>` plays key role in showing the thumbnail and thus I went with enhancing it (to stick with the thumbnail). Another route is duplicating `<dds-video-player>` as well as `<dds-video-player-composite>` and `<dds-video-player-container>`, but such duplication would cause maintenance concerns.

asudoh:
Uh... The thing that `<dds-card-cta cta-type="video">` supports video thumbnail totally slipped my mind. Re-layouting that component seems much closer to the feature requested. Withdrawing. Sorry everybody!

jeffchew:
Thanks @asudoh , I was just in the middle of writing a response when you withdrew 👍

Issue #5450, opened 2021-03-11
React/Web component: Back to top - No label for back to top button


Praveen-sr:
## Detailed description

On jaws FF
Issue: There is no label for back to top button

Expected:
Provide label as "back to top"

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/b8a2dc88-65b0-46f7-a59d-80899d9cf494)

PR #5451, opened 2021-03-11
fix(back-to-top): add aria-label to button

kennylam:
### Related Ticket(s)

#5450 

### Description

Add `aria-label` to button.

### Changelog

**New**

- `aria-label` to button


IgnacioBecerra (approved):
LGTM, thanks @kennylam!

Issue #5417, opened 2021-03-09
LeadSpace – Centered g100 not displaying properly in sm breakpoint


kirbyevans:
## Detailed description

> The LeadSpace – centered with no image is displaying a broken image link below the button group when on the sm breakpoint view. For this particular component, there shouldn't be a background image associated with it. Please see attached GIF for a visual of what is happening.
- The breakpoint behavior for small in lead space centered is not the same as react.
- [react link](http://ibmdotcom-react-canary.mybluemix.net/?path=/story/components-leadspace--centered)

## Steps to reproduce the issue

1. Go here: https://ibmdotcom-web-components-experimental.mybluemix.net/?path=/story/components-leadspace--centered
2. Scale browser down to sm breakpoint view

## Additional information
![Mar-08-2021 16-24-59](https://user-images.githubusercontent.com/77290329/110481411-8b556f00-80b5-11eb-93e0-8362021a695b.gif)

PR #5453, opened 2021-03-11
fix(leadspace): remove image from centered leadspace with no image

ariellalgilmore:
### Related Ticket(s)

#5417 

### Description

Image displaying when rendering without image. Added condition to check for image `defaultSrc`

### Changelog

**Changed**

- update condition for `_renderImage()`


IgnacioBecerra:
Screen Shot 2021-03-11 at 8 45 55 AM

Thanks @ariellalgilmore! The broken image doesn't show up anymore, but the background color still defaults to the gradient instead of remaining under the g100 theme.

I think this could be fixed by changing the breakpoint [here](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/packages/web-components/src/components/leadspace/leadspace.scss#L14-L20) to `sm` since the Web Components seems to be handling the gradient differently than in React.

ariellalgilmore:
> Screen Shot 2021-03-11 at 8 45 55 AM
> 
> Thanks @ariellalgilmore! The broken image doesn't show up anymore, but the background color still defaults to the gradient instead of remaining under the g100 theme.
> 
> I think this could be fixed by changing the breakpoint [here](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/packages/web-components/src/components/leadspace/leadspace.scss#L14-L20) to `sm` since the Web Components seems to be handling the gradient differently than in React.

oh forgot to add that! thanks @IgnacioBecerra!

IgnacioBecerra (approved):

jeffchew:
@ariellalgilmore can you check Percy results? Looks like the opacity layer isn't behaving properly in some places.

jeffchew:
Is it me or the opacity layer a bit bolder than it should be?

Screen Shot 2021-03-11 at 6 22 37 PM

ariellalgilmore:
> Is it me or the opacity layer a bit bolder than it should be?
> 
> Screen Shot 2021-03-11 at 6 22 37 PM

haha yeah it's not just you. Working on this. the styles are a bit weird

jeffchew (approved):
Thank you @ariellalgilmore !

Issue #5190, opened 2021-02-17
React: Masthead - Missing Label for search edit combo


Praveen-sr:
## Detailed description

On  JAWS FF/Voiceover iPhone

Issue:
Missing Label for search edit combo

Note: Label is present in Web component staging

PR #5455, opened 2021-03-11
fix(masthead-search): added aria-label to input container

guilhermelMoraes:
### Related Ticket(s)

React: Masthead - Missing Label for search edit combo #5190

### Description

Screen Shot 2021-03-11 at 13 20 42

### Changelog

**Changed**

- `react-autosuggestion` version updated from `9.4.3` to `10.1.0`. For some reason, this was required for the `containerProps` to work. I created a code-sandbox and swapped versions and could reproduce this bug. Looking at the React Autosuggest issues I couldn't find anything related to that though.


packages/react/src/components/Masthead/MastheadSearch.js
```diff
@@ -267,6 +267,15 @@ const MastheadSearch = ({
     className: `${prefix}--header__search--input`,
   };
 
+  /**
+   * Autosuggest will pass through all these props to the container.
+   *
+   * @type {{'aria-label': string}}
+   */
+  const containerProps = {
+    'aria-label': 'IBM search field',
```
kennylam, line 276: Thanks for this @guilhermelMoraes! We can use the default prop `placeHolderText` for the `aria-label` value here. This will allow localized languages to be used when needed.

packages/react/src/components/Masthead/MastheadSearch.js
```diff
@@ -267,6 +267,15 @@ const MastheadSearch = ({
     className: `${prefix}--header__search--input`,
   };
 
+  /**
+   * Autosuggest will pass through all these props to the container.
+   *
+   * @type {{'aria-label': string}}
+   */
+  const containerProps = {
+    'aria-label': placeHolderText,
```
kennylam, line 276: Thanks for this @guilhermelMoraes! Small thing, we can use the `placeHolderText` prop for the `aria-label` value. This will allow localized versions of this text to be passed in for this element.

kennylam (approved):
LGTM thanks @guilhermelMoraes!

PR #5457, opened 2021-03-11
docs(ButtonGroup): add onClick prop and remove href & copy prop requirement

annawen1:
### Related Ticket(s)

### Description

Add `onClick` prop definition so adopters are aware that it can be passed in and remove the requirement from the `href` and `copy` props.


jeffchew (approved):
Thank you @annawen1 !

PR #5462, opened 2021-03-11
chore(typescript): adding typescript v3.8.x as a peer dependency

jeffchew:
### Related Ticket(s)

No related issue

### Description

As more adopters are running into compatibility issues when running applications that uses typescript, it's important to define what version is supported with our library. This adds typescript `v3.8.x` as a peer dependency for web components.

### Changelog

**Changed**

- added `typescript@~3.8.0` peerDependency


yarn.lock
```diff
@@ -19379,9 +19379,9 @@ node-modules-regexp@^1.0.0:
   integrity sha1-jZ2+KJZKSsVxLpExZCEHxx6Q7EA=
 
 node-notifier@>=8.0.1, node-notifier@^6.0.0:
-  version "9.0.0"
-  resolved "https://registry.yarnpkg.com/node-notifier/-/node-notifier-9.0.0.tgz#46c5bbecbb796d4a803f646cea5bc91403f2ff38"
-  integrity sha512-SkwNwGnMMlSPrcoeH4CSo9XyWe72acAHEJGDdPdB+CyBVHsIYaTQ4U/1wk3URsyzC75xZLg2vzU2YaALlqDF1Q==
+  version "9.0.1"
+  resolved "https://registry.yarnpkg.com/node-notifier/-/node-notifier-9.0.1.tgz#cea837f4c5e733936c7b9005e6545cea825d1af4"
+  integrity sha512-fPNFIp2hF/Dq7qLDzSg4vZ0J4e9v60gJR+Qx7RbjbWqzPDdEqeVpEx5CFeDAELIl+A/woaaNn1fQ5nEVerMxJg==
```
asudoh, line 19384: Wondering if this change was actually needed. If so, we need to get in this version of `node-notifier`. Otherwise we can back this change out?

asudoh (approved):
LGTM, just one thing - Thanks @jeffchew!

yarn.lock
jeffchew: I'm pushing up a yarn offline sync change, but this change happened when I did another yarn install. I can double check our settings for this but I don't think we're restricting the version of `node-notifier`

asudoh (approved):
Thanks @jeffchew for the update!

Issue #5293, opened 2021-02-25
Fix ESLint warnings across the Carbon for IBM.com codebase


jeffchew:
#### User Story

> As a `[user role below]`:
Carbon for IBM.com developer

> I need to:
see no ESLint warnings in Carbon for IBM.com

> so that I can:
ensure a clean code base when checking for errors

#### Additional information

 - There are currently 36 warnings appearing in the code base. While this does not cause any major issues, it is good to regularly clean up these warnings. Some warnings are based on incorrect JSDoc documentation, which we would like to maintain consistency across.

#### Acceptance criteria
 - [ ] 0 ESLint warnings

jeffchew:
@RobertaJHahn FYI, we can triage this in engineering refinement as this is a pure engineering issue.

PR #5465, opened 2021-03-12
chore(lint): eslint warning cleanup

kennylam:
### Related Issue

Refs #5293

### Description

Fixes all eslint warnings. This also removes the Github annotations currently polluting the bottom of all our PRs.

### Changelog

**Changed**

- update various `jsDoc` tag types
- move `onKeyDown` from static non-interactive HTML to `useEffect` (lack of a `role` attribute caused a lint warning)

**Removed**

-  remove `packages/react/src/internal` as all components now import from `internal/vendor/carbon-components-react/internal/`


jeffchew (approved):
LGTM!

packages/react/src/components/Masthead/HeaderNavContainer.js
```diff
@@ -134,17 +134,21 @@ const HeaderNavContainer = ({ children }) => {
...
     } else {
       return () => {
         if (io) {
+          navContent.removeEventListener('keydown', handleOnKeyDown);
           io.disconnect();
         }
       };
     }
-  }, [io]);
+  });
```
asudoh, line 151: Seems that the change makes the callback run on every render, was this just for ESLint or of a bigger issue...?

packages/utilities/src/utilities/focuswrap/focuswrap.js
```diff
@@ -15,7 +15,7 @@ import on from 'carbon-components/es/globals/js/misc/on';
  *   The focus sentinel nodes.
  *   If these nodes gets focus, we see it as focus went out of the `element`.
  * @param {string} [eventRequestFocusWrap=dds-request-focus-wrap] The event name.
- * @returns {Handle} The handle to remove the event handler.
+ * @returns {Function} The handle to remove the event handler.
```
asudoh, line 18: Actually this function returns an object instead of a function. What kind of ESLint error did you get?

packages/web-components/tools/babel-plugin-create-react-custom-element-type.js
```diff
@@ -30,7 +30,7 @@ function createMetadataVisitor(api) {
   const { types: t } = api;
 
   /**
-   * @param {Path} path The Babel path what a `@property()` decorator call refers to.
+   * @param {string} path The Babel path what a `@property()` decorator call refers to.
```
asudoh, line 33: `path` in Babel is of AST tree node instead of file path. What kind of ESLint error did you get...?

asudoh (commented):
Thank you for trying to fix this @kennylam!

packages/web-components/tools/babel-plugin-create-react-custom-element-type.js
kennylam: @asudoh This is the lint warning

image

kennylam: Thanks @asudoh, I'll update the `typedef` since I don't think we should suppress valid warnings with the ignore.

asudoh: I see it's good, just in case it helps, `Path` type comes from 3rd party library (for this particular case, Babel). Given we are not on TypeScript for this particular code, we don't have a good way to refer to those 3rd party types (or we define every 3rd party types)...

PR #5480, opened 2021-03-13
chore(content-group): update base class to new layout

asudoh:
### Related Ticket(s)

Refs #4174.

### Description

As we finished migrating all content patterns to the new layout aligning to non-gutter grid, updates the base classes to the new one, that finishes the migration.

### Changelog

**Changed**

- `DDSContentBlock`/`DDSContentGroup` abstract classes to use the new layout of content pattern, aligning to non-gutter grid.


kennylam (approved):
LGTM thanks @asudoh!

Issue #5151, opened 2021-02-11
Masthead:  Persistent search text in Masthead searchbar after search is executed


Daynssjal:
### The problem
Hello, we are using the Masthead component with searchbar. 
When the user types in a search keyword and executes search, the search keyword disappears from the searchbar after the user has been redirected to a new page.

### The solution
We would like the search keyword to be always visible in the searchbar (after the search executes, after a redirect..)

Example can be found on ibm.com/search page,  e.g. https://www.ibm.com/search?lang=en&cc=us&q=api 
The keyword is visible even after search executes and the user can continue writing

### Additional information
Please, be aware that we are using the searchNoRedirect prop, because we do not want our users to be redirected to the default ibm.com/search page. We want them to be redirected to our own search page. 

This feature is requested by the Knowledge Center team.

Expected:

![image](https://user-images.githubusercontent.com/33804471/107654547-08daba80-6c83-11eb-8d4b-0b7a0fd883e6.png)

- Because this feature already exists on ibm.com Masthead search, JC is providing the code to Miroslav Krsjak on the KC team.  As of the March 3 adopter status meeting, the KC team will include this feature in their search to meet their release date.  JC will contribute the code to Carbon for ibm.com.  (According to Miro, JC's code is included in a back level version of the Carbon for ibm.com Masthead.)  DDS will include the feature and will perform the QA required.

- [Design and functional specs Box folder](https://ibm.ent.box.com/folder/59063660753?s=mgb8c6urxlvgpfsbxqvqifb1bj9tm6n1)

RobertaJHahn:
Next Steps:  Reach out to the MH search team to see whether their code can be contributed back to Carbon for ibm.com Library.

Shixie agrees with the design, keeping the search term in the search box when results are presented.

RobertaJHahn:
JC can contribute the code during their next sprint which begins next week.  (Feb 22 - Mar 5)

PR #5501, opened 2021-03-16
feat(masthead-search): add persistent search term

jcpalaci:
### Related Ticket(s)

#5151

### Description

- Added a new prop in `Masthead`, `initialSearchTerm` to pass the initial search value for the search field
- Added a new function in `MastheadSearch`, `getValueFromQueryString` to get (by default) the initial search term from the query string

### Changelog

**New**

- New prop in `Masthead` & `MastheadSearch`: `initialSearchTerm`
- New function in `MastheadSearch`: `getValueFromQueryString`

**Changed**

- `_initialState` in `MastheadSearch` is now calculated as follow:
  - If `initialSearchTerm` is provided, use that value
  - if `initialSearchTerm` is not provided, try to get the search term from the query string (`q`)
  - otherwise, default to an empty string


kennylam (approved):
LGTM thanks @jcpalaci! Just looks like you need to update your snapshots.

Issue #5181, opened 2021-02-16
Wrong type token in Mega menu "the essentials" description text


shixiedesign:
## Detailed description

Currently, the description text in "the essentials" column appears use `$body-short-01`, which has a line height of 22px. Intended design & spec is to use `$body-long-01` which has a line height of 24px. 

![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/f9b3224e-418b-4ce0-8e62-8975874a80d9)

photodow:
@shixiedesign I'm assuming this is a bug in the [component](http://ibmdotcom-web-components-experimental.mybluemix.net/?path=/story/components-masthead--with-custom-data)? I'm tagging it as dev to be triaged for now. Feel free to correct me here if I'm assuming wrong.

PR #5516, opened 2021-03-16
fix(masthead): using correct token for essentials description text

IgnacioBecerra:
### Related Ticket(s)
#5181 

### Description
The previous token was `$body-short-01` while it should've been `$body-long-01`.

### Changelog

**Changed**

- changed the slotted `dds-megamenu-category-group-copy` type from `$body-short-01` to `$body-long-01`


photodow (approved):
:+1:  Appears to be looking good to me. Thanks!

Issue #5400, opened 2021-03-06
[Feature card] Bugs & remove unnecessary knobs


shixiedesign:
## Bug 1. Overflow not showing according to spec

Currently, overflow content is simply hidden, pushing arrow out of card

![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/6302599f-cb84-4f33-be6d-707ec7ada637)

Expecting:

![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/c1717274-44d4-4186-b0c5-84e6d0ee2c9e)

## Bug 2. Hover state not to spec

Currently, card shows no change on hover.
Expecting: 
![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/40672363-42a4-4043-b37c-90cbf757095b)

## Bug 3. Focus state not to spec

![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/641b4ee0-d6c5-4742-9eac-6129c89b57d6)

## Refine storybook
Please remove kobs
- Image src
- Image alt text

![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/d1f464a2-03dc-495e-89d5-794072bc72ee)

Issue #5416, opened 2021-03-09
[Feature card] Deprecate blocks, add large variant to Feature card


shixiedesign:
### Problem

Feature card currently exist in 2 sizes, and as 3 different components: 

- [Feature card block large ](**https://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-feature-card-block-large--default**)
- [Feature card block medium](https://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-feature-card-block-medium--default)
- [Feature card](https://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-feature-card--default)

These are many inconsistencies in design (with / without headline, padding, background color, size of arrow etc). The existence of 3 components also make it hard for adopters to know which one is the correct one to use.

### Proposal

Remove the block level components and treat Feature card as a simple UI component with 2 sizes. If needed, adopters may wrap it in Content group or Content block for additional headlines & body copy.

To remove redundant presentation, and consolidate designs so they are more findable, I propose doing the following: 

1. Feature card block large <--- Deprecate
2. Feature card block medium <-- Deprecate
3. Feature card: rename default to `medium`, fix bugs to spec
4. Feature card: add `large` variant, fix bugs to spec
0. Concurrently, we need FS update FYI @photodow 

### Final **Feature card** component spec

Storybook plam: 
![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/38a6ff23-feb9-4b25-8153-528484a9d207)

https://ibm.box.com/s/292bsxb7tat33ch341u8o4oh6xdx19ct

**Medium**
![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/2c848c49-20f4-4d42-b440-5142d6bd990a)
![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/c04aeaba-c3d1-4bc8-973f-951209f9c7ae)
![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/b4573a62-449c-4b72-8a46-24872607e6b0)

**Large**
![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/4f791099-9ab2-45e8-96f8-84e5af97ba57)
![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/19e6bc78-b978-4a98-92c0-db278b067480)
[...]

PR #5517, opened 2021-03-17
Feature card knobs bugs

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5400

### Description

Multiple bug fixes, including corrected card overflow in smaller screens:
feature-card

### DIFF
--- a/packages/web-components/src/components/feature-card/__stories__/feature-card.stories.ts
+++ b/packages/web-components/src/components/feature-card/__stories__/feature-card.stories.ts
@@ -19,10 +19,10 @@ import readme from './README.stories.mdx';
 import textNullable from '../../../../.storybook/knob-text-nullable';
 
 export const Default = ({ parameters }) => {
-  const { heading, defaultSrc, alt, href } = parameters?.props?.['dds-feature-card'] ?? {};
+  const { heading, href } = parameters?.props?.['dds-feature-card'] ?? {};
   return html`
     <dds-feature-card href=${ifNonNull(href || undefined)}>
-      <dds-image slot="image" alt="${ifNonNull(alt)}" default-src="${ifNonNull(defaultSrc)}"></dds-image>
+      <dds-image slot="image" alt="Image alt text" default-src="${imgLg1x1}"></dds-image>
       <dds-card-heading>${heading}</dds-card-heading>
       <dds-feature-card-footer>
         ${ArrowRight20({ slot: 'icon' })}
--- a/packages/web-components/src/components/feature-card/feature-card.scss
+++ b/packages/web-components/src/components/feature-card/feature-card.scss
@@ -16,6 +16,10 @@
   display: block;
   outline: none;
 
+  .#{$prefix}--card {
+    outline-offset: 1px;
+  }
+
   @include carbon--breakpoint-down('md') {
     .#{$prefix}--card__content {
       padding: 0;
@@ -28,7 +32,9 @@
   ::slotted(#{dds-prefix}-card-heading) {
     background-color: $ui-01;
     margin: 0;
+    width: 100%;
     padding: carbon--mini-units(2);
+    padding-right: 10%;
     padding-bottom: var(--cds-layout-03, 2rem);
   }
 
--- a/packages/web-components/src/components/feature-card/feature-card.scss
+++ b/packages/web-components/src/components/feature-card/feature-card.scss
@@ -1,7 +1,7 @@
 /**
  * @license
  *
- * Copyright IBM Corp. 2020
+ * Copyright IBM Corp. 2020, 2021
  *
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
@@ -14,8 +14,30 @@
 :host(#{$dds-prefix}-feature-card),
 :host(#{$dds-prefix}-feature-cta) {
   display: block;
-
   outline: none;
+
+  @include carbon--breakpoint-down('md') {
+    .#{$prefix}--card__content {
+      padding: 0;
+      justify-content: space-between;
+    }
+  }
+}
+
+@include carbon--breakpoint-down('md') {
+  ::slotted(#{dds-prefix}-card-heading) {
+    background-color: $ui-01;
+    margin: 0;
+    padding: carbon--mini-units(2);
+    padding-bottom: var(--cds-layout-03, 2rem);
+  }
+
+  ::slotted(#{$dds-prefix}-feature-car-footer) {
+    background-color: $ui-01;
+    padding: carbon--mini-units(2);
+    padding-top: 0;
+    margin-top: carbon--mini-units(-4);
+  }
 }
 
 :host(#{$dds-prefix}-feature-card-footer),


photodow (changes requested):
Looking good! It's come a long way. Here are just a couple of additional comments real quick...

1. Image is still missing its hover state.
2. Focus is still aligned to the outside. It should be inset.
3. Can we blow this component in storybook up to 8 columns for demo purposes?

> forgive me, I see the `draft` sign now. I looked for it when I started, but I guess I was looking in the wrong place.
> For the parts you're still working on please ignore. :)

photodow (changes requested):
A couple of comments that would be great to get in here, so let me know if I need to open a new ticket for any of these.

1. This component should be themeable using inverse tokens. Does anyone have any history on why this is locked to the `g90` theme?

```scss
93 114 @include carbon--theme($carbon--theme--g90, true);
```

2. Eyebrow needs to be removed from `FeatureCard` in the `FeatureCardBlockMedium` component.

3. Shouldn't the timing on the [video feature CTA](https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/5517/iframe.html?id=components-cta--feature&knob-CTA%20type%20(cta-type)_FeatureCTA=video&knob-Copy%20text_FeatureCTA=Lorem%20ipsum%20dolor%20sit%20amet&knob-Content%20link%20href%20(href)_FeatureCTA=https://www.example.com&knob-Footer%20copy%20text_FeatureCTAFooter=&knob-Content%20link%20href%20(href)_FeatureCTAFooter=https://www.example.com) be next to the icon like our other cards?

    - CTA feature is also supposed to depend on `FeatureCardBlockMedium`

packages/styles/scss/components/feature-card/_feature-card.scss
```diff
@@ -70,6 +70,27 @@
...
+        background-color: $ui-01;
+        margin: 0;
+        width: 100%;
+        padding: carbon--mini-units(2);
+        padding-right: 10%;
+        padding-bottom: $layout-03;
+      }
+
+      #{$prefix}--card__footer {
+        background-color: $ui-01;
```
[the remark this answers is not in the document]
NSDrowned: Thanks for the comments @photodow. I was testing this because the way the card structure was setup. Not sure if I'm looking at it the correctly but the anchor link (which holds most of the content) has an absolute position. As the text gets longer, this doesn't allow the heading to increase the height of the card in smaller resolutions, causing the overflow problem mentioned in this issue. But I agree, this solution is too messy so I'm working on a new one.

image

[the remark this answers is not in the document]
NSDrowned: Hello @photodow, I was able to solve the overflow bug without any background resets or color rules in general.

NSDrowned:
Hello @photodow, issue is ready for review. I fixed the overflow issue in the react component too and saw what you meant about the multiple background color definitions. To improve this, the styles for this component would need a couple of days of dedications to achieve this.

photodow:
> To improve this, the styles for this component would need a couple of days of dedications to achieve this.

Sorry, to improve what? You said you fixed it in the sentence before, so I'm slightly confused there.

> Hello @photodow, issue is ready for review.

I'm still seeing many of my points above not addressed yet. Let me know if we need to sync up on those.

NSDrowned:
@photodow my apologies if I didn't explain myself well. What I meant was that the overflow problem was fixed, but I would need to dedicate some time to refactor the code to try to avoid the multiple background definitions you mentioned. I think that would need a new issue to be created. 

For the focus, you're right, I thought I had to leave 1px gap because of this screenshot. Gap removed.

image

Fixed the missing image hover too.

shixiedesign (commented):
Hey Julio, thank you. Yeah @photodow I didn't catch the background locked to g90 bug. Will create new issue for it. Scoping to this issue, I caught 2 problems: 

1. hover on `sm` breakpoint seems to break and only lighten half of the card. Expecting to see whole hard hover + image darken. Currently: 
Mar-23-2021 22-12-16

2. focus state on this guy is a bit more similar to the focus state on [Carbon primary button](https://react.carbondesignsystem.com/iframe.html?id=components-button--default). It's expected to have 2 focus outlines with `1px white` then `2px $focus`, but **inset** to the card so focusing doesn't actually make the card look bigger. Please check out the Carbon button and see if it make sense. Here's more documentation on this focus state using button as an example https://carbondesignsystem.com/components/button/style

image

NSDrowned:
@shixiedesign fixed the hover and outline ;)

shixiedesign (approved):
Thank you!

shixiedesign:
I captured [James comment](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/5517#pullrequestreview-614587320) above in a new bug issue: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5601

photodow:
Before I approve did this introduce a grid alignment issue to `FeatureCardBlockMedium`? I'm not seeing it on the [main storybook](http://ibmdotcom-web-components.mybluemix.net/?path=/story/components-feature-card-block-medium--default) site.

image

shixiedesign:
Something changed...yep, but looks like the current (canary) Feature card block medium is not to spec too. The card should be hanging into the gutter. I'm also looking to deprecate the "block" feature cards in general and add the large size to feature card component. See this #5416 @photodow. 

**Current (canary) Feature card block medium** below

image

packages/styles/scss/components/feature-card/_feature-card.scss
```diff
@@ -27,15 +27,33 @@
...
+      right: 0;
+      bottom: 0;
+      border: 2px solid $focus;
+      outline: 1px solid #ffffff;
+      outline-offset: -3px;
+
+      @include carbon--breakpoint-down('md') {
+        min-height: 292px;
+      }
+    }
```
kennylam, line 55: Thanks @NSDrowned! Small thing about `focus` styles, the `before` pseudo-element isn't needed. See buttons in [button group](http://ibmdotcom-react.mybluemix.net/?path=/story/components-buttongroup--default) for an example.

NSDrowned: Thanks @kennylam, the problem I found with box-shadow was that the shadow has the inset option, so it shows behind the content and hidden from the user. 

`box-shadow: inset 0 0 0 1px $focus, inset 0 0 0 2px #ffffff;`

The image and heading child elements show on top of it, like this example (image with 0.5 opacity so we can see the inner shadow): 

image

It works in the button because the content is centered in the element, far away from the shadow. 

I looked at other components but they seem to have the same problem with the image being on top:

image

And the outline is outside in the picture element:

image

I'm open to any suggestions on how to achieve this without using a pseudo element.

kennylam: @NSDrowned Hm ok, I see what you mean. Can approve for now and go back to check an alternative way of doing this without the pseudo elements. 👍

kennylam (approved):

PR #5524, opened 2021-03-17
feat(masthead): add persistent search

kennylam:
### Related Ticket(s)

#5204

### Description

See [`Search Open Onload No Blur`](https://ibmdotcom-react.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/5524/index.html?path=/story/components-masthead--search-open-onload-no-blur) story.

This adds an option for masthead search to be open on load and cannot be closed `onBlur`. When this option is enabled the search can no longer be closed by clicking outside the masthead. Clicking `X` clears the input and closes the search as normal.

### Changelog

**New**

- `searchOpenOnloadNoBlur` option


packages/react/src/components/Masthead/README.stories.mdx
```diff
@@ -242,6 +242,16 @@ const platformData = {
...
 
+## searchAlwaysOpen
+
+Search will always be open, it cannot be closed.
+Clicking the close button will clear and focus the search input only.
+> This should only be used when there are no navigation items in the masthead.
+
+```javascript
+<Masthead navigation="default" searchAlwaysOpen />;
+```
```
photodow, line 253: So I'm wondering if there was a disconnect after our conversation the other day, and maybe it's a misunderstanding on my part of the need. The `searchAlwaysOpen` needs design exploration on how to handle that `clear`/`close` user interaction. Are we ok with the `x` icon to clear when other places of the system it's used to `close`, or should we have differential indicator. The way I wrote the functional specs is if the `openOnLoad` is true then `close` on `blur` shouldn't happen, and if the user closes it at any point it should retain the user's value.

photodow (changes requested):
I see this is still in draft, so ignore any of these if you're already working towards it.

1. The overflow carrot is visible when it opens on load.

image

2. It's supposed to retain the value after closing and opening again.

Mar-17-2021 14-46-12

kennylam:
> 1. It's supposed to retain the value after closing and opening again.

@photodow are you looking at the correct story? A new [`searchAlwaysOpen`](https://ibmdotcom-react.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/5524/index.html?path=/story/components-masthead--search-always-open) story was created.

photodow:
> @photodow are you looking at the correct story? A new searchAlwaysOpen story was created.

It wasn't there when I was looking before for some reason. That behavior still isn't per functional specs, but also see https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/5524#discussion_r596337891

packages/react/src/components/Masthead/README.stories.mdx
kennylam: No disconnect on my part, this is exactly the behavior I proposed. The purpose of this draft is so the design team can review this option. If a different icon should be used to clear the field, we can change that  (or maybe something like a tooltip that says `Clear` when it doesn't close the input).

For `searchOpenOnload`, an accessibility issue was opened a little while back that required it to close on `blur`, I'll have to go digging to find it. But if that's still the intended behavior we should probably track that in a separate issue.

kennylam:
> That behavior still isn't per functional specs

Do we have functional specs for `searchAlwaysOpen`?

packages/react/src/components/Masthead/README.stories.mdx
photodow: Right this is what you proposed, but we haven't made any decisions/need to move forward with that proposal just yet. No designs, or functional specs around it yet.

photodow:
Do we have functional specs for `searchAlwaysOpen`? No, because design hasn't had a chance to look at it yet.

photodow:
@kennylam I was talking to @shixiedesign and @oliviaflory about this earlier. It sounded like for the time being we want to stick with `searchOpenOnload` and allowing the `x` button to continue to close and clear the search. According to specs and requests from KC we still want to update it to prevent close `onBlur` like IBM Search is doing. Let's keep a pulse on this though. If this is something we continue to see a need for we might think about a component separate from Masthead.

@shixiedesign @oliviaflory feel free to add anything.

kennylam:
@photodow Ok sounds good. So the only difference betwwn `searchOpenOnload` and this is that `searchAlwaysOpen` cannot be close on `blur` and only with the close button?

photodow:
Yeah I think so, does this look right?

`searchOpenOnload`
— `onBlur` remains open `UPDATE NEEDED`
— `x` closes search `&&` clears value

~`searchAlwaysOpen`
— `onBlur` remains open 
— `x` remains open `&&` clears value~

shixiedesign:
Just checking in on this one. I notice @kennylam you already have the `Always open` done in React and I think we preferred to update `Open on load` because, a functional X button for close & clear are still desirable for 

1. integrity of the search component, trying to avoid having to take it apart 
2. it doesn't hurt as long as we have the `onBlur` stay open done. 
3. I could be wrong but I am under the impression `Open on load` is specifically for Search and KC these 2 teams needs, we can modify them however we want, so no need to create new prop. 

Then the only other thing I notice is in Web component there's that overflow caret button James pointed out, but that seems like a different bug/issue. Feel free to exclude from this PR if that makes more sense. Lemme know if these help. Thanks!

kennylam:
@photodow @shixiedesign I think I need to understand this better. Currently `searchOpenOnLoad` does nothing but that. It's open when the page loads, then behaves no differently from the default search. The `X` will clear the input and close the search, `blur` will close the search only if there is no value in the input field.

So to borrow James' list:

`searchOpenOnload`
— `blur`: remains open if input value exists, closes otherwise
— `x`: closes search and clears value

KC's feature request (and current behavior of IBM search) would be:

`searchOpenOnloadNoBlur` (proposed new name of this feature)
— `blur`: remains open regardless of input value  (this is the only difference from `searchOpenOnload`)
— `x`: closes search and clears value

This is what I understood from the discussion with James above. I've updated the PR with these new changes, let me know if we're talking about the same thing. Thanks!

photodow:
@kennylam I think you may have misunderstood me. It's not determined by the input value. We don't want `searchOpenOnload` to close on blur at all. We shouldn't need a new property for `searchOpenOnloadNoBlur` since that's how `searchOpenOnload` should behave.

kennylam:
> We don't want `searchOpenOnload` to close on blur at all.

@photodow Got it, this part is what i wasn't getting. I'll make the change.

shixiedesign (approved):
I see that No Blur story is added and works as expected! Thanks for juggling all the issues Kenny! I noticed 1 small bug. The overflow carets flashes for a brief second _when search bar closes_. 
The sprint is closing and I know there's a lot on your plate, if there's no time to fix the caret - since this PR delivers the feature requested - I'm okay with merge now, and open a bug to fix the caret. 
Other refinements include just remove one of the "Open onLoads" coz if I'm not wrong, no one requested the one that close on blur.

Mar-26-2021 09-42-03

kennylam:
@photodow @shixiedesign Thanks guys for all the time you've taken to go through this! Taking out of draft so we can get this in. The overflow issues were addressed in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/5680.

shixiedesign (approved):
Looks to expected! Thank you!

photodow (approved):

packages/react/src/components/Masthead/MastheadSearch.js
```diff
@@ -575,6 +608,12 @@ MastheadSearch.propTypes = {
    */
   searchOpenOnload: PropTypes.bool,
 
+  /**
+   * `true` to make the search field open in the initial state.
+   * Search will not close on blur, but with close button only.
+   */
+  searchOpenOnloadNoBlur: PropTypes.bool,
```
jeffchew, line 615: @kennylam strangely I'm not seeing this in the props table in the documentation...

kennylam: @jeffchew I just noticed i forgot to remove it as well, good catch!

jeffchew (approved):
LGTM!

Issue #5386, opened 2021-03-05
[Call out] should use inverse tokens


oliviaflory:
## Detailed description

> Describe in detail the issue you're having.

Call out quote uses G-90 theme when it should be using inverse tokens for background.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

bug (?) not following specs

> Is this issue related to a specific component?

Callout

cascades to:
- Callout with media
- Callout quote

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Call out is using g-90 theme, should follow specs and use inverse
![Screen Shot 2021-03-05 at 11.26.11 AM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/aeb80db9-08f2-4f4d-85f9-c99ac73baff0)

[call out scss](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/packages/styles/scss/internal/callout/_callout.scss)

I also noticed the g-90 theme doesn't seem to be working on the NextJS templates, assuming the use of inverse fixes this:
[Solutions example](http://ibmdotcom-nextjs-test.mybluemix.net/solutions.html) at the bottom of Content Block Segmented (2)

![Screen Shot 2021-03-05 at 11.23.02 AM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/4ba6650f-9d76-43b1-b594-78828f51ad00)

> What browser are you working in?

Chrome

> What version of Carbon for IBM.com are you using?

Latest Web components

Seen in [Call out with media](https://ibmdotcom-web-components.mybluemix.net/?path=/story/components-callout-with-media--default) and [Call out quote](https://ibmdotcom-web-components.mybluemix.net/?path=/story/components-callout-quote--default)

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

[Visual specs](https://ibm.ent.box.com/folder/110530875791?s=6b6pdgql2y3zxja2soj058t0ev8zwzik)

Updated March 3, 2020, original specs also used inverse tokens.

## Possible questions:
 1. Is call out using gray 90 theme because Call out with media and call out quote use default themes vs inverse?
2. If yes to above, how can we ensure that call out is always 'inverse' when applied in a template to follow design intent. Reference example above in NextJS where call out with media defaults to white theme tokens.

PR #5528, opened 2021-03-17
fix(callout): changing background-color token

guilhermelMoraes:
### Related Ticket(s)

[Call out] should use inverse tokens #5386

### Description

changing background-color token to `$inverse-02` as per the visual specs.


photodow (changes requested):
Thank you! Some much needed TLC.

1. Do you plan to address the downstream effects in a separate issue, or would you like feedback on some of them here? For example, the copy and links and such don't have inverse tokens.

image

2. tangentially related, the knobs don't seem to be working in `callout with media`, and theming is not available.

guilhermelMoraes:
@photodow feel free to provide all the feedback here, please.

oliviaflory:
@guilhermelMoraes thanks, background color looks good! 

As James mentioned the link in Call out quote should be $inverse-link, and the text should use $inverse-01 in Call out quote and callout with media. 
I may have not been explicit enough in the issue #5386 calling out the text and links as well.

oliviaflory:
@guilhermelMoraes hover color looks great! 👍 

I see a double focus state on the Call out with media, and can this be `$inverse-focus-ui` please?

Screen Shot 2021-03-22 at 3 24 15 PM

jeffchew:
Screen Shot 2021-03-22 at 5 28 36 PM

@guilhermelMoraes I think was mentioned above, looks like Percy picked this up in Callout Quote.

I'm hoping to also get more e2e percy tests in place to capture more theme variations.

shixiedesign (approved):
Looks like all the things Olivia and James picked out are fixed! 💪 Thank you Gui!

jeffchew:
Hey @guilhermelMoraes , seeing it darker here too for some reason:

image

guilhermelMoraes:
@jeffchew updated! Now it should be ready to go.

photodow (changes requested):
- [ ] Need to update the active state real quick.

image

- [x] I'm not sure if this is related or not, but for some reason, the link with icon is changing type size on a theme change... Was it doing that before?

Mar-24-2021 15-56-15

- [ ] Itsy bitsy thing, but can we lower case that `Icon` in the text CTA?

photodow (changes requested):
Hey @guilhermelMoraes sorry there is an issue with hover on the dark themes on the link. I don't think that passes accessibility, and appears to be overridden by an RGB value. Do you see this too?

image

- [ ] Just looked in dotcome shell story, and I'm seeing some issues with callout with media in there. Were we able to get the inverse tokens for that one applied?

image

shixiedesign (commented):
Hey James catches the best bugs - yeah I was able to verify his comments above: 
1. Hover of the Link in inverse is unaccessible. Unfortunately there's no token in Carbon for it. In dark themes, Link on hover should get `$blue-40`
2. @guilhermelMoraes is callout media's bug out of scope of this PR? If so that's cool, we should make a new bug ticket for it. Looks like the body copy in Callout media is not using token. I don't know how it's build so lemme know if new issue is needed. 

Thanks!

jeffchew:
Just double-checking, the background color for g100 is supposed to be white now, correct?

image

Also the link size still looks like it's smaller than it should be according to Percy.

shixiedesign:
@jeffchew @guilhermelMoraes 
- the background color change on Dark theme is correct ✅ 
- Link size change is not though. We have a [previous bug](https://app.zenhub.com/workspaces/carbon-for-ibmcom-5d449f3642eb1962336cbe52/issues/carbon-design-system/carbon-for-ibm-dotcom/5647) of g100 theme affecting link type sizes in general, so I'm okay with this bug being fixed in a separate PR.

guilhermelMoraes:
@shixiedesign
1. updated;
2. that's a good question. The theming button is disabled for the [Callout with Media story] (http://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-callout-with-media--default). Is this a design decision?

shixiedesign:
Thanks Gui for adding the theme tab to `Callout with media` so I could QA the color token used there. It did expose a new issue #5676. Would be done as a separate PR. Thank you! Looks good to me 👍

Issue #5470, opened 2021-03-12
[TOC Default and Horizontal prev Anchor link] Web component missing scroll animation


oliviaflory:
## Detailed description

> Describe in detail the issue you're having.

Both TOC Default and Horizontal are missing the scroll animation that is currently present in [React](https://ibmdotcom-react.mybluemix.net/?path=/story/components-table-of-contents--manually-define-menu-items). 

**Web components – TOC Horizontal**
![TOC no scroll animation Mar-12-2021 09-39-13](https://user-images.githubusercontent.com/45692622/110954937-0e293480-8317-11eb-8fa2-4ba6883b0f7a.gif)

**Web components – TOC Default**
![TOC default no scroll Mar-12-2021 09-51-49](https://user-images.githubusercontent.com/45692622/110956491-ad026080-8318-11eb-991a-149c74504468.gif)

**React – TOC Default scroll animation**
![TOC defaul scroll animation Mar-12-2021 09-40-16](https://user-images.githubusercontent.com/45692622/110955014-2305c800-8317-11eb-9beb-2674631869d2.gif)

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug
- included in original specs
- Default TOC should have been in parody with React component
- Horizontal TOC should also have scroll animation

> Is this issue related to a specific component?
[TOC Default}(https://ibmdotcom-web-components.mybluemix.net/?path=/story/components-table-of-contents--default)
[TOC Horizontal](https://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-table-of-contents--horizontal)

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Should have scroll animation

> What browser are you working in?

Chrome

> What version of Carbon for IBM.com are you using?

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

- Screenshots or code
- Notes

PR #5531, opened 2021-03-17
fix(table-of-contents): added smooth scrolling on  web-components version

RaphaelAmadeu:
### Related Ticket(s)

#5470

### Description

I've added the `smoothScroll` utility from our utilities package in table of contents.


photodow (changes requested):
My only comment here is the scroll jacking that appears to be happening. There needs to be some way to escape. So if it's still scrolling down to the bottom of the page, and I'm trying to scroll up it doesn't scroll up until it finishes scrolling all the way down. I'm putting a request for changes here, let me know if we need to open a separate ticket to that. I can approve once we get that new ticket open and linked here for reference.

@oliviaflory any thoughts?

oliviaflory:
@photodow > There needs to be some way to escape.
Right, forgot about your scrolling best practices research, happy to have that included 👍

jeffchew:
@RaphaelAmadeu any status on this?

Issue #5398, opened 2021-03-05
[Same height utility] content not lining up properly


photodow:
This is an important issue to be included in the April release.

## Detailed description

It looks like at some point our same height utility broke, and some of our content is not lining up the way it should. Cards group, and CTA section are ones I've seen issues with here so far. I wonder if there are any other places here, or if these two places will fix it for us.

> calling it out as same height, it could be with cards group, or cta section implentation of the utility as well.

##### This is what I'm seeing

![image](https://user-images.githubusercontent.com/3793636/110175435-5c927c80-7dc7-11eb-8bf6-6f11c1550bba.png)

![image](https://user-images.githubusercontent.com/3793636/110175862-08d46300-7dc8-11eb-9d53-40535ceb22b2.png)

##### This is what I expect to see

Notice how the headlines, copy and CTA all line up with each other.

![image](https://user-images.githubusercontent.com/3793636/110175508-7c29a500-7dc7-11eb-9de8-9b6ffa03c9c0.png)

![image](https://user-images.githubusercontent.com/3793636/110175776-ec382b00-7dc7-11eb-8429-88077c2687ad.png)

PR #5532, opened 2021-03-18
fix(cardGroup): storybook improvements

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5415

### Description

- [x] Apply grid to card group, so the component should be snap to the grid col 5-16
- [x] Apply different amount of content in cards to demo `same height util` in action:
_Added random lorem phrase functionality_
- [x] New substory: with card in card (already existed in web components)
- [x] New substory: with card in card and image card

PR #5530, opened 2021-03-17
feat(cta-block): introduced cta block web component, added it as cta section variant, removed promo group and promo item

rastogir:
### Related Ticket(s)

- [#5216](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5216)
- [#5220](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5220)
- [#5221](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5221)
- [#5218](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5218)

### Description

- Adds `<dds-cta-block>`. This replaces `<dds-promo-group>` and `<dds-promo-item>`. It is based off of `<dds-content-item>`.
- Adds `CtaBlock` variant to `<dds-cta-section>` component. This variant has an additional prop of `Media Type` to switch between image, video, statistic and pictogram. 
- Removes `<dds-promo-group>` and `<dds-promo-item>`.

### Design Specs
 
- https://ibm.ent.box.com/s/shr0pnhof0yzk2b5lh0nokwopgxxb9rr
Screen Shot 2021-03-17 at 3 54 43 PM

- Additional design spec for statisitc from Kirby :
  - Font: IBM Plex Sans, SemiBold
  - Max: 76px
  - xlg: 60px
  - lg: 54px
  - md and sm: 42px

### Functional Spec

- https://user-images.githubusercontent.com/3793636/108720924-067c2a00-74e7-11eb-8441-44bb8b412dcd.png

### Changelog

**New**

- `<dds-cta-block>` component. 

**Changed**

- `<dds-cta-section>` story updated with `<dds-cta-block>` variant.

**Removed**

- `<dds-promo-group>` and `<dds-promo-item>` component.

PR #5567, opened 2021-03-22
fix(same-height): using same height method for card group & cta section

IgnacioBecerra:
### Related Ticket(s)
#5398

### Description
The current CSS grid implementation for both CardGroup and CTA Section work are able to resize properly dynamically depending on the copy content, however it does not work if the heading is longer than a single line. After some discussion, a CSS-only way might actually not be achievable, thus this PR addresses that by bringing back the `sameHeight` utility function for both `CardGroup` and `CTASection`, only targeting the item headings. 

### Changelog

**Changed**

- re-applying the `sameHeight` utility function for `CardGroup` and `CTASection` targeting item headings


shixiedesign (commented):
Ooof I saw this when I tried to add extra content to heading of first card. 
**Current**
image

**Expecting**
image

IgnacioBecerra:
@shixiedesign When using the same height utility it actually takes the height of the longest heading upon rendering, so when using inspect element it wouldn't actually change. I added a card with a longer heading to the Story:

Screen Shot 2021-03-24 at 6 49 00 AM

 I hadn't added the variation in this PR since @NSDrowned was working on doing that in his PR (#5532), but I'll add it here since it fits the scope.

photodow:
- [ ] Does this work with the `eyebrow`?

image

Looks like the answer is no...
image

There may also be an issue here though that they don't have matching content.

- [ ] And CTA?
image

I'm trying to think if `copy` is needed too... My brain isn't workingo n this one, but I think originally they all were supposed to be.

- [ ] Shouldn't these update on screen resize? I'm not seeing that happen.

photodow:
Have you talked to @Praveen-sr about this update to `CTA Section`? I know she's been working that component over at #5530

IgnacioBecerra:
@photodow Updated to include eyebrow and CTA footer + w/resize. I'm not sure if copy is needed since the heading/footer would ensure copy stays the same throughout. 

Examples: 
Screen Shot 2021-03-25 at 2 56 43 PM

Screen Shot 2021-03-25 at 2 57 07 PM

photodow (changes requested):
@IgnacioBecerra looking a lot better.


- [ ] Any idea when I add content to the `heading` in the content items for `CTA section` that the `copy` expands too?

image

- [ ] Also looking at the smallest breakpoint the content is overflowing. At this point I don't think it needs to be the same size so the content should just flow normally.

image

- [ ] I think we should also apply this to react as well? It doesn't look like it's working there right now.

image

IgnacioBecerra:
@photodow I believe the reason for that weird issue with heading increasing height copy might have to do with the current css grid setup. We weren't able to find a workaround for that, so I'm adding copy to the CTA Section's sameHeight utility. Also added functionality to React, though working on fixing the CI Checks.

photodow (changes requested):
It looks like I'm still seeing the second point above on the card group still. I just loaded it, and resized the browser down and now cards has some extra spacing. It should just be the old natural content lead spacing.

image

shixiedesign:
Hey @IgnacioBecerra, awesome to see it working! Thanks for the random content in there to demo the util. I'll leave James to take a final look later next week but it looks good to me! 👍 
I'm seeing the tons of Percy diffs but I know it's probably because we are working on so many things this sprint. They should be all expected...but yeah hard to be sure. Lemme know if you've got a better way to look through them.

jeffchew:
@IgnacioBecerra any update on this?

photodow (approved):
@jeffchew I've been talking to Ignacio offline about this around the issue mentioned above. I think we decided to create a new ticket around that since it appears to be based on some sort of aspect ratio is applied. I'll create that ticket (#5710), and go ahead and approve this for now.

Issue #5486, opened 2021-03-15
[content item horizontal] align media right by default


photodow:
Andy will reach out to Ian to see whether he can make this fix and contribute it.

### As-is
Looks like the code released with `left` as the default.

> If I misread that than we can ignore and close this issue.

![image](https://user-images.githubusercontent.com/3793636/111171219-b8e95f00-8572-11eb-8e98-052167b064ee.png)

### Change expected
I think we just want to change it to `right` align as default for the media. It's a small change, but generally speaking, our design system aligns content left. The media `left` align here is a secondary subtle break in the page.

![image](https://user-images.githubusercontent.com/3793636/111171189-af5ff700-8572-11eb-8604-6e058d3f7f64.png)

### Additional information
 - [functional specs](https://ibm.box.com/s/64h13x3utcps7lva14abxbx5t41xx9w1)

PR #5571, opened 2021-03-23
fix(content-item-horizontal-media): Switched the default alignment fr…

iwhitcomb:
…om left to right.

### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5486

### Description

I changed the alignment knob from left to right to adhere to the design system standard.

### Changelog

**Changed**

- Simple change of the default left variable to right on the alignment knob.


andysherman2121 (approved):
Looks good to me! Thanks  @iwhitcomb

photodow (changes requested):
I think we need to update more than just the story. We need to set the component's default itself.

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/c1b9bdeb1acfa3dada2b694b8fd131bf3a4c1cbf/packages/web-components/src/components/content-item-horizontal/content-item-horizontal-media.ts#L28-L36

shixiedesign:
Low priority fix, but seems like the `Copy` knob isn't working? When I add or remove content in text input of `Copy`, the body copy below headline remains unchanged.

image

andysherman2121 (changes requested):
Agreeing with @shixiedesign The copy knob isn't working

photodow (approved):
Looks like my request for changes from my expectations is good as far as my understanding goes so I'm going to go ahead and approve from my perspective to avoid me being a blocker. I'll lean on Andy to double-check the code and make sure it works as expected, but it sounds like he's been working with you on this so I trust it's good.

iwhitcomb:
@andysherman2121 this appears to be an issue with the default display as well.

andysherman2121:
> @andysherman2121 this appears to be an issue with the default display as well.

@iwhitcomb what appears to be an issue on default display?

iwhitcomb:
> @iwhitcomb what appears to be an issue on default display?

@andysherman2121 the copy knob isn't working. It's definitely an issue somewhere higher up in the hierarchy so possibly more widespread than just these 2 components. I haven't been able to pinpoint the issue yet though.

andysherman2121:
@iwhitcomb You're right, the copy knob isn't working in a few components. (content group simple). @shixiedesign Seems like the copy issue is a more widespread problem not specific to this component. I think we should make a new story for it to be fixed throughout the codebase

andysherman2121 (approved):

shixiedesign (approved):
Thank you! BTW @iwhitcomb I did notice some grid issues with the component but it's not relevant to this PR's issue. It could also have slipped through when we first added it because our design team haven't got the standards figured out. It already looks great but in general we just want to make sure it renders on the grid, even in storybook. Thanks! https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5649

Issue #5144, opened 2021-02-11
Web component: Masthead-With custom data - Search auto suggestions is not working.  


Praveen-sr:
## Detailed description

Issue: 
Masthead-With custom data - Search auto suggestions feature is not working.  

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/f90cf393-1c9a-4114-94a1-27055e01afdd)

PR #5582, opened 2021-03-23
fix(masthead): enabled search suggestions with custom data

IgnacioBecerra:
### Related Ticket(s)
#5144

### Description
The Masthead with Custom Navigation story now uses `masthead-search-container` in order for the Search API redux actions to be loaded in alongside `masthead-composite`. 

### Changelog

**Changed**

- `<dds-masthead-composite>` now uses `dds-masthead-search-container` to load in SearchAPI redux actions


packages/web-components/src/components/masthead/masthead-composite.ts
```diff
@@ -699,15 +699,15 @@ class DDSMastheadComposite extends LitElement {
         ${!hasSearch
           ? undefined
           : html`
-              <dds-masthead-search-composite
+              <dds-masthead-search-container
```
asudoh, line 702: Composite component should never depend on container components. Do you have any background on this change? In general, container component should connect backend API to composite components.

IgnacioBecerra: The reason behind this change is that the Masthead with Custom Data story is currently using the masthead composite directly in order to bypass the retrieved navLinks from the Translations API. However, this has the side effect of not loading the SearchAPI actions that usually come from masthead container. By using search-container over search-composite we are able to have autosuggestions when using masthead-composite. 

I've looked into adding the Search API into masthead-composite itself, though not sure if that's possible -- please let me know otherwise.

asudoh: For that, you can create a new container component that works only with search results. You can start with creating a copy of `masthead-container.ts`. In the new copy (make sure assigning a new class name as well as a new custom element tag name), `mapStateToProps()` should pick only [`currentSearchResults`](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.16.1/packages/web-components/src/components/masthead/masthead-container.ts#L101) and `mapDispatchToProps()` should pick only [`_loadSearchResults`](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.16.1/packages/web-components/src/components/masthead/masthead-container.ts#L118). In this way, the only API the new container component interacts with will be `SearchAPI`.

IgnacioBecerra: I see, that makes sense, it's essentially a "variation" of the container. I was focused on finding a solution between these two components that I hadn't considered creating a third. Thanks for letting me know!

packages/web-components/src/components/masthead/masthead-container-search-only.ts
```diff
@@ -11,19 +11,43 @@ import pickBy from 'lodash-es/pickBy.js';
...
 import ConnectMixin from '../../globals/mixins/connect';
+import { MastheadSearchContainerState } from './masthead-search-container';
 import DDSMastheadComposite from './masthead-composite';
 
 const { stablePrefix: ddsPrefix } = ddsSettings;
 
+/**
+ * The Redux state used for `<dds-masthead-container>`.
+ */
+export interface MastheadContainerState extends MastheadSearchContainerState {
```
asudoh, line 29:
```suggestion
export interface MastheadContainerSearchOnlyState extends MastheadSearchContainerState {
```

packages/web-components/src/components/masthead/masthead-container-search-only.ts
```diff
@@ -11,19 +11,43 @@ import pickBy from 'lodash-es/pickBy.js';
...
+
+  /**
+   * The Redux state for `TranslateAPI`.
+   */
+  translateAPI?: TranslateAPIState;
+
+  /**
+   * The Redux state for `ProfileAPI`.
+   */
+  profileAPI?: ProfileAPIState;
```
asudoh, line 43:
```suggestion
```

asudoh (commented):

asudoh (approved):

packages/web-components/src/components/masthead/masthead-composite.ts
```diff
@@ -53,7 +53,7 @@ import './left-nav-menu';
 import './left-nav-menu-item';
 import './left-nav-menu-category-heading';
 import './left-nav-overlay';
-import './masthead-search-composite';
+import './masthead-search-container';
```
annawen1, line 56: @IgnacioBecerra do we still need this import here?

annawen1 (approved):

PR #5593, opened 2021-03-24
feat(cloud-masthead): add Cloud-specific container and composite components

huntermacd:
### Related Ticket(s)

https://jsw.ibm.com/browse/HC-1681

### Description

This diff adds various Cloud-specific versions of existing components. This should be kept as a draft since there is a lot more work to do, but I need some eyes from the team on the work so far for guidance.

### Changelog

**New**

- cloud-masthead-container component
- cloud-masthead-composite component
- cloud-masthead-profile component

**Changed**

- add Storybook story for Cloud variant masthead
- added selector to masthead.scss to include Cloud-specific masthead


packages/web-components/src/components/masthead/__stories__/masthead.stories.ts
```diff
@@ -286,6 +287,28 @@ withAlternateLogoAndTooltip.story = {
...
+      platform-url="${ifNonNull(platformData.url)}"
+      selected-menu-item="${ifNonNull(selectedMenuItem)}"
+      user-status="${ifNonNull(userStatus)}"
+      searchPlaceholder="${ifNonNull(searchPlaceholder)}"
+      .authenticatedProfileItems="${ifNonNull(authenticatedProfileItems)}"
+      .navLinks="${customLinks}"
+      ?has-profile="${hasProfile}"
+      ?has-search="${hasSearch}"
+      .unauthenticatedProfileItems="${ifNonNull(unauthenticatedProfileItems)}"
+      data-endpoint="/common/carbon-for-ibm-dotcom/translations/cloud-masthead"
```
annawen1, line 307: was double checking why the endpoint was resulting in a 404 - turns out I forgot we need to add /jsononly to the path. We should be able to get the data using /common/carbon-for-ibm-dotcom/translations/cloud-masthead/jsononly

packages/web-components/src/components/masthead/cloud/cloud-masthead-container.ts
```diff
@@ -0,0 +1,24 @@
...
+
+const { stablePrefix: ddsPrefix } = ddsSettings;
+
+/**
+ * Container component for masthead.
+ *
+ * @element dds-cloud-masthead-container
+ */
+@customElement(`${ddsPrefix}-cloud-masthead-container`)
+class DDSCloudMastheadContainer extends DDSMastheadContainer {}
```
annawen1, line 22: I mentioned extending the masthead-container, but it's more like importing the methods from there. Since we are removing the `ProfileAPI`, the `mapDispatchToProps` needs to be defined. We also need to map the props to `DDSCloudMastheadComposite` otherwise you'll be rendering the regular `<dds-masthead-composite>`

like so:
```
/**
 * @license
 *
 * Copyright IBM Corp. 2021
 *
 */

import { customElement } from 'lit-element';
import { ActionCreatorsMapObject, Dispatch, Store, bindActionCreators } from 'redux';
import store from '../../../internal/vendor/@carbon/ibmdotcom-services-store/store';
import { LocaleAPIActions } from '../../../internal/vendor/@carbon/ibmdotcom-services-store/actions/localeAPI.d';
import { TranslateAPIActions } from '../../../internal/vendor/@carbon/ibmdotcom-services-store/actions/translateAPI.d';
import { SearchAPIActions } from '../../../internal/vendor/@carbon/ibmdotcom-services-store/actions/searchAPI.d';
import ConnectMixin from '../../../globals/mixins/connect';
import { loadTranslation } from '../../../internal/vendor/@carbon/ibmdotcom-services-store/actions/translateAPI';
import { loadLanguage, setLanguage } from '../../../internal/vendor/@carbon/ibmdotcom-services-store/actions/localeAPI';
import { loadSearchResults } from '../../../internal/vendor/@carbon/ibmdotcom-services-store/actions/searchAPI';
import {
  MastheadContainerState,
  MastheadContainerStateProps,
  MastheadContainerActions,
  mapStateToProps,
} from '../masthead-container';

import DDSCloudMastheadComposite from './cloud-masthead-composite';

/**
 * @param dispatch The Redux `dispatch()` API.
 * @returns The methods in `<dds-cloud-masthead-container>` to dispatch Redux actions.
 */
export function mapDispatchToProps(dispatch: Dispatch<LocaleAPIActions | TranslateAPIActions>) {
  return bindActionCreators<MastheadContainerActions, ActionCreatorsMapObject<MastheadContainerActions>>(
    {
      _loadLanguage: loadLanguage,
      _setLanguage: setLanguage,
      _loadTranslation: loadTranslation,
      _loadSearchResults: loadSearchResults,
    },
    dispatch as Dispatch // TS definition of `bindActionCreators()` seems to have no templated `Dispatch`
  );
}

/**
 * Container component for masthead.
 *
 * @element dds-cloud-masthead-container
 */
@customElement(`${ddsPrefix}-cloud-masthead-container`)
class DDSCloudMastheadContainer extends ConnectMixin<
  MastheadContainerState,
  LocaleAPIActions | TranslateAPIActions | SearchAPIActions,
  MastheadContainerStateProps,
  ActionCreatorsMapObject<MastheadContainerActions>
>(
  store as Store<MastheadContainerState, LocaleAPIActions | TranslateAPIActions | SearchAPIActions>,
  mapStateToProps,
  mapDispatchToProps
)(DDSCloudMastheadComposite) {}

export default DDSCloudMastheadContainer;
```

packages/web-components/src/components/masthead/__stories__/masthead.stories.ts
```diff
@@ -286,6 +287,28 @@ withAlternateLogoAndTooltip.story = {
   },
 };
 
+export const WithCloudVariant = ({ parameters }) => {
+  const { selectedMenuItem, userStatus, searchPlaceholder, hasProfile, hasSearch } = parameters?.props?.MastheadComposite ?? {};
+  return html`
```
proeung, line 292: @huntermacd Let's use the `useMock` variable so that we're keeping to the same method for Percy's visual testing for cloud-masthead components as well.

```suggestion
  const { useMock } = parameters?.props?.Other ?? {};
  return html`
    <style>
      ${styles}
    </style>
    ${useMock
      ? html`
          <dds-masthead-composite
            platform="${ifNonNull(platformData.name)}"
            platform-url="${ifNonNull(platformData.url)}"
            selected-menu-item="${ifNonNull(selectedMenuItem)}"
            user-status="${ifNonNull(userStatus)}"
            searchPlaceholder="${ifNonNull(searchPlaceholder)}"
            .authenticatedProfileItems="${ifNonNull(authenticatedProfileItems)}"
            .navLinks="${navLinks}"
            ?has-profile="${hasProfile}"
            ?has-search="${hasSearch}"
            .unauthenticatedProfileItems="${ifNonNull(unauthenticatedProfileItems)}"
          ></dds-masthead-composite>
        `
      : html`
          <dds-masthead-container
            platform="Cloud"
            platform-url="${ifNonNull(platformData.url)}"
            selected-menu-item="${ifNonNull(selectedMenuItem)}"
            user-status="${ifNonNull(userStatus)}"
            searchPlaceholder="${ifNonNull(searchPlaceholder)}"
            .navLinks="${navLinks}"
            ?has-profile="${hasProfile}"
            ?has-search="${hasSearch}"
            data-endpoint="/common/carbon-for-ibm-dotcom/translations/cloud-masthead/jsononly"
          ></dds-masthead-container>
        `}
  `;
```

packages/web-components/src/components/masthead/cloud/cloud-masthead-composite.ts
```diff
@@ -0,0 +1,150 @@
...
+class DDSCloudMastheadComposite extends DDSMastheadComposite {
+  /**
+   * Returns user login status (authenticated or anonymous)
+   */
+  private _getUserLoginStatus() {
+    return Cookies.get(this.loginStatusCookieName) === '1' ? { user: 'authenticated' } : { user: 'anonymous' };
+  }
+
+  /**
+   * The cookie name for determining user login status.
```
proeung, line 64: A bit of nitpicking, but let's adjust the comment so that folks know that this cookie status is for `cloud.ibm.com`.

packages/web-components/src/components/masthead/__stories__/masthead.stories.ts
```diff
@@ -286,6 +287,28 @@ withAlternateLogoAndTooltip.story = {
   },
 };
 
+export const WithCloudVariant = ({ parameters }) => {
```
proeung, line 290: Low hanging fruit change - Let's rename this storybook variant from `WithCloudVariant` to `WithCloudNavigation` as I haven't seen the word "Variant" being used anywhere within the different previews. Also, keep in mind that the name might be adjusted in the future.

packages/web-components/src/components/masthead/__stories__/masthead.stories.ts
proeung: @annawen1 This is good to know. Thanks for pointing this out. Also, since we're using the JSON-only directory, I'm guessing we don't need the `.js` within the dotcom translation repo? https://github.ibm.com/webstandards/ibm-dotcom-translations/blob/master/translations/cloud-masthead/usen.js

proeung: I just wanted to bring this up as this will play in our translation content population workflow and if we're only using the JSON files that would help come down in the formatting of the `.js` versions. Let me know what you think.

annawen1: Yeah we don't need the `.js` file for the translations, I think we initially kept it for consistency when copying over the current ibm.com translations folder structure? Not sure if removing it will cause any issues. @jeffchew what do you think?

packages/web-components/src/components/masthead/cloud/cloud-masthead-composite.ts
```diff
@@ -0,0 +1,150 @@
...
+ *
+ * @element dds-cloud-masthead-composite
+ */
+@customElement(`${ddsPrefix}-cloud-masthead-composite`)
+class DDSCloudMastheadComposite extends DDSMastheadComposite {
+  /**
+   * Returns user login status (authenticated or anonymous)
+   */
+  private _getUserLoginStatus() {
+    return Cookies.get(this.loginStatusCookieName) === '1' ? { user: 'authenticated' } : { user: 'anonymous' };
```
annawen1, line 60: Just leaving a note here for other folks looking at this PR, we do intend to move this logic into the services store so there is a separation between the data fetching and ui rendering

packages/web-components/src/components/masthead/cloud/cloud-masthead-composite.ts
```diff
@@ -0,0 +1,150 @@
...
+import '../top-nav-menu';
+import '../top-nav-menu-item';
+import '../left-nav';
+import '../left-nav-name';
+import '../left-nav-item';
+import '../left-nav-menu';
+import '../left-nav-menu-item';
+import '../left-nav-menu-category-heading';
+import '../left-nav-overlay';
+import '../masthead-search-composite';
```
annawen1, line 43: We can remove all the component imports here except for `import './cloud-masthead-profile';` since we're extending from `dds-masthead-composite`

packages/web-components/src/components/masthead/cloud/cloud-masthead-composite.ts
```diff
@@ -0,0 +1,150 @@
...
+          placeholder="${ifNonNull(searchPlaceholder)}"
+          .currentSearchResults="${ifNonNull(currentSearchResults)}"
+          ._loadSearchResults="${ifNonNull(loadSearchResults)}"
+        ></dds-masthead-search-composite>
+        <dds-masthead-global-bar>
+          <dds-cloud-masthead-profile ?authenticated="${loginStatus.user === 'authenticated'}">
+            ${profileItems?.map(
+              ({ title, url }) =>
+                html`
+                  <dds-masthead-profile-item href="${ifNonNull(url)}">${title}</dds-masthead-profile-item>
```
proeung, line 136: As part of the new Cloud masthead work, we need to expose a CTA button component that links out to the console. This is a new feature that we'll need to account for within the authenticated state of the `dds-cloud-masthead-profile` component (see specs - https://ibm.ent.box.com/file/786859527229). 

component--cloud-masthead-l0--design-specs--1_0_0_pdf___Powered_by_Box

As for the data/translation, we'll need to add this to our JSON file as well. @annawen1 @jeffchew Let me know if there are fields within the existing data structure that we can reuse or if this is something new that we need to add. https://1.www.s81c.com/common/carbon-for-ibm-dotcom/translations/cloud-masthead/jsononly/usen.json

proeung (changes requested):
@huntermacd Overall, this is a great first draft of the work, and thanks for putting this PR together. @annawen1 and I left some feedback. Let me if you have any questions/concerns.

packages/web-components/src/components/masthead/cloud/cloud-masthead-profile.ts
```diff
@@ -0,0 +1,140 @@
...
+const { prefix } = settings;
+const { stablePrefix: ddsPrefix } = ddsSettings;
+
+/**
+ * The Cloud-specific profile menu UI in the masthead.
+ *
+ * @element dds-cloud-masthead-profile
+ */
+@customElement(`${ddsPrefix}-cloud-masthead-profile`)
+class DDSCloudMastheadProfile extends HostListenerMixin(FocusMixin(LitElement)) {
```
asudoh, line 30: Wondering if this class will become simpler by extending `DDSMastheadProfile`. Seems that most properties/methods are copies.

asudoh (commented):
Nice contribution - Thanks @huntermacd!

packages/web-components/src/components/masthead/cloud/cloud-masthead-composite.ts
annawen1: hmm don't see any existing fields we can utilize for the `console` button so I think we need to add it. Possibly in 

```
"masthead": {
    "search": {
      "btnSearchClosed": "Open IBM search field",
      "btnSearchOpen": "Search all of IBM",
      "btnClose": "Close IBM search field",
      "placeHolderText": "Search all of IBM"
    },
    --> "not sure key name": "Console"
```

packages/web-components/src/components/masthead/__stories__/masthead.stories.ts
proeung: @annawen1 Thanks! @jeffchew Let me know if we need to account for the `.js` files within the Cloud Masthead translation directory.

packages/web-components/src/components/masthead/cloud/cloud-masthead-composite.ts
proeung: @annawen1 Looking at the existing structure, I think it might be good to add this button within the `[profileMenu][signedin]` array, as this button is intended to show up in the authenticated state (see attached ). This will be managed outside of the `[profileMenu][signedin][links]`, which is being used to display the user drop-down menu. Let me know what you think.

https___1_www_s81c_com_common_carbon-for-ibm-dotcom_translations_cloud-masthead_jsononly_usen_json

@huntermacd let me know what you think as well.

annawen1: yep, that looks good! Just note we need to edit the type files in the services store package! Let me know if you need help with that.

proeung: @annawen1 Great! I'll work on opening a PR to add this new `ctaButton` array. As for accounting for these types in the services store package, perhaps this is something you can help us add once the data is in place?

annawen1: sure!

huntermacd:
cc @annawen1 @asudoh @jeffchew 

I could use some assistance wrapping these components behind a feature flag, being that they're interconnected.

Also, regarding the cookie check, this needs to happen on each load of the component. I'm not certain that there's a need to put this value into the store. The goal is to check the cookie for determining which version of the Cloud Masthead Profile to show, either "authenticated" or "anonymous". Putting this functionality into an importable service, a la the ProfileAPI, also seems like overkill when it boils down to a single, simple function. Please let me know your thoughts on the matter.

annawen1:
> cc @annawen1 @asudoh @jeffchew
> 
> I could use some assistance wrapping these components behind a feature flag, being that they're interconnected. My latest commit [70085a3](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/commit/70085a3faf7a8f3a56e135966816b9ca012078e4) shows my progress on doing that. Currently it's broken. I may have a circular import happening.
> 
> Also, regarding the cookie check, this needs to happen on each load of the component. I'm not certain that there's a need to put this value into the store. The goal is to check the cookie for determining which version of the Cloud Masthead Profile to show, either "authenticated" or "anonymous". Putting this functionality into an importable service, a la the ProfileAPI, also seems like overkill when it boils down to a single, simple function. Please let me know your thoughts on the matter.

@asudoh wonder what your thoughts are on where the cookie fetching logic lives?

packages/web-components/src/components/masthead/cloud/cloud-masthead-composite.ts
proeung: Here's the PR to adjust the data structure based on our conversation here - https://github.ibm.com/webstandards/ibm-dotcom-translations/pull/295.

annawen1:
> @annawen1 Thank you for asking. If the team thinks maintaining the login status in `@carbon/ibmdotcom-services-store`, they can manually change `userStatus` property in `<dds-masthead-composite>` or one in `<dds-masthead-container>`. In other words, cookie handling code should not be baked in to `<dds-masthead-*>`.

Thanks @asudoh! @huntermacd, based on Akira's comment I think it may be good to fetch the cookie data within `dds-cloud-masthead-container` and set cookie value to `userStatus`. This way we won't have to put the logic in the `services-store` package and we won't be baking it into `dds-masthead-*` components

proeung:
@huntermacd Thanks for addressing all of the feedback and changed the status of this PR `draft` to `publish`. One thing that we're still missing within the cope of [HC-1681](https://jsw.ibm.com/browse/HC-1681) is the CTA buttons that are needed for the authenticated and anonymous states of the masthead profile component.

Screen_Shot_2021-04-02_at_10_47_24_AM

Here's the Cloud Masthead endpoint ([see link](https://1.www.s81c.com/common/carbon-for-ibm-dotcom/translations/cloud-masthead/jsononly/usen.json)) that includes the latest data for the `profileMenu` anonymous and authenticated states (see attached).

https___1_www_s81c_com_common_carbon-for-ibm-dotcom_translations_cloud-masthead_jsononly_usen_json-2

For these CTA buttons, we'll be using the CTA component referencing the button type (eg. https://ibmdotcom-web-components-experimental.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/5593/index.html?path=/story/components-cta--button)

Also, here's the link to the design specs. https://ibm.ent.box.com/file/786859527229

packages/web-components/src/components/cta/button-cta.ts
```diff
@@ -29,7 +29,7 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
  * @element dds-button-cta
  */
 @customElement(`${ddsPrefix}-button-cta`)
-class DDSButonCTA extends VideoCTAMixin(CTAMixin(DDSButtonGroupItem)) {
```
annawen1, line 32: Good Catch!

packages/web-components/src/components/masthead/cloud/cloud-masthead-container.ts
```diff
@@ -0,0 +1,98 @@
...
+  for (let { length = 0 } = currentSearchQueryString ?? {}; !currentSearchResults && length > 0; --length) {
+    currentSearchResults = searchResults?.[currentSearchQueryString!.substr(0, length)]?.[language!];
+  }
+  return pickBy(
+    {
+      authenticatedProfileItems: !language ? undefined : translations?.[language]?.masthead?.profileMenu.signedin.links,
+      authenticatedCtaButtons: !language ? undefined : translations?.[language]?.masthead?.profileMenu.signedin.ctaButtons,
+      navLinks: !language ? undefined : translations?.[language]?.mastheadNav?.links,
+      unauthenticatedProfileItems: !language ? undefined : translations?.[language]?.masthead?.profileMenu.signedout.links,
+      unauthenticatedCtaButtons: !language ? undefined : translations?.[language]?.masthead?.profileMenu.signedout.ctaButtons,
```
annawen1, line 58: Since the `ctaButtons` key is new, I think we will have to add this to the `packages/services-store/src/types/translateAPI.ts` file

packages/web-components/src/components/masthead/__stories__/cloud-masthead.stories.ts
```diff
@@ -0,0 +1,110 @@
...
+                .navLinks="${navLinks}"
+                ?has-profile="${hasProfile}"
+                ?has-search="${hasSearch}"
+                data-endpoint="/common/carbon-for-ibm-dotcom/translations/cloud-masthead/jsononly"
+              ></dds-cloud-masthead-container>
+            `}
+      `;
+    };
+
+export default {
```
annawen1, line 73: @huntermacd can you add a flag check here too? I think not having it is bombing all the stories currently

```
export default !DDS_CLOUD_MASTHEAD
  ? undefined
  : {
      title: 'Components/Cloud Masthead',
      .....
      
  }
  ```

proeung (approved):
@huntermacd Looks great! I'm now able to see the CTA button for the anonymous and authenticated states!

@annawen1 @jeffchew I think this PR is good to merge down to the main branch and available for view via the experimental environment. What do you both think?

annawen1:
@huntermacd looks good! 🎉

One thing - I do see it's failing on one of the tests, can you take a look?
Screen Shot 2021-04-08 at 5 34 38 PM

proeung:
Looks like all of the CI checks are cleared! @annawen1 Are we good to merge this PR down?

annawen1:
> Looks like all of the CI checks are cleared! @annawen1 Are we good to merge this PR down?

Everything looks good, one thing though sorry! @jeffchew you mentioned about the `js-cookie` dependency and if it would make more sense to have the cookie fetching logic as part of the utilities package where we already have `js-cookie`?

We have a similar util called `ipcInfoCookie.js` there.

jeffchew:
> > Looks like all of the CI checks are cleared! @annawen1 Are we good to merge this PR down?
> 
> Everything looks good, one thing though sorry! @jeffchew you mentioned about the `js-cookie` dependency and if it would make more sense to have the cookie fetching logic as part of the utilities package where we already have `js-cookie`?
> 
> We have a similar util called `ipcInfoCookie.js` there.

That's right, and sorry that I'm catching up with this PR now. What caught my attention was the `js-cookie` dependency getting added to `web-components`, where we have separated out the cookie setting and fetching logic (to @annawen1 's point) in our `utilities` package. It may be nice to also have it there so that the handling of that is centralized.

@asudoh to your question about PAL specific things, I agree that that would be the ultimate goal. This is a first iteration of the Cloud masthead which is why we have suggested putting behind a feature flag. At least the way I'm thinking of, it would be nice to be able to ultimately abstract some of the things happening in this variation of the Masthead so that other teams/BUs can ultimately take advantage of some of the things here, and Cloud would also ultimately be using this abstracted version for their needs too. But for now it can be safely iterated behind the flag while also getting something that can hit the glass for the Cloud team.

proeung:
@jeffchew Thanks for the response! Sounds to me like we have a good direction on how to manage the cookie settings and fetching logic within the existing `@carbon/ibmdotcom-utilities` package for future implementation. In the meantime, I think we can proceed with the feature flag and account for the abstraction at a later time. Let me know if you'd like me to create a GH issue for this feature.

jeffchew (changes requested):
@proeung yes, would typically agree that if we're behind a feature flag, we can safely continue to iterate behind it. However, I think the added dependency in the package itself, regardless if the intention is to move to another package later on, is something we should try to avoid. 

I think if at minimum we can move the cookie library dependency out of the `web-components` package, I think that can be safe to then merge this in.

proeung:
@jeffchew Sounds good! I'll create a ticket for us to track the work for moving the js-cookie dependency and fetching logic to the existing `@carbon/ibmdotcom-utilities` package. In the meantime, I think this recent commit (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/5593/commits/33cf59064d4e8f92deffb0fd7cd78409699ee73e) that @huntermacd made should allow this PR to move forward with merging.

@shixiedesign I don't believe this PR requires a design QA at the moment. We're working on getting the baseline changes for the new Cloud masthead merged down so that we can add in the other design enchantments. Also, the changes in this PR are behind a feature flag. We'll conduct a proper design QA review once we have all of the elements assembled that you, Lila Title, and others can take a look at.

huntermacd:
With the removal of the js-cookie dependency in the web components package, we're getting an `import/no-extraneous-dependencies` eslint error which is causing CI the fail. I'm not terribly familiar with Lerna or Yarn's workspaces feature, but I'm assuming we're setup to share dependencies in exactly this way. Questions: am I correct in that assumption? And if so, what's the preferred method for getting around this eslint error in cases like this? @jeffchew

annawen1:
Pushed changes to move the cookie fetching logic in the utilities package!

proeung:
@annawen1 Awesome job moving the js-cookie fetching logic in the `@carbon/ibmdotcom-utilitie`s package. You saved us a lot of time with creating the ticket and implementing this work, so thank you so much! 🥇 

@jeffchew With the latest commit (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/5593/commits/864d8930920320ef9cca0aec649d32b026d681ee) that @annawen1 pushed up, I think we're good for another code review and I think we can safely merge this PR down. What do you think?

annawen1 (approved):
LGTM!

jeffchew (approved):

Issue #5473, opened 2021-03-12
[TOC Horizontal prev Anchor link] stacking to two lines, missing overflow feature


oliviaflory:
## Detailed description

> Describe in detail the issue you're having.

Looks like the text is stacking to two lines between 1055–1160px, we had the overflow feature from the masthead in the specs. ~Realizing this might be a bit tight to get into the release and fully test, so LMK if it's too much, maybe we could have the Select kick in early if the content runs out of room?~

![Screen Shot 2021-03-11 at 5 06 34 PM](https://user-images.githubusercontent.com/45692622/110863283-76800380-828e-11eb-9f0f-1b401e3fd3fb.png)

<img width="1254" alt="Screen Shot 2021-03-11 at 5 27 06 PM" src="https://user-images.githubusercontent.com/45692622/110863685-08880c00-828f-11eb-87d8-d85deaf93952.png">

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

I'm not positive if the overflow was put on hold to get into the March release, or missed in the specs so added `bug` and `feature request` label

> Is this issue related to a specific component?

[TOC Horizontal](https://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-table-of-contents--default)

> What did you expect to happen? What happened instead? What would you like to
> see changed?

No stacking to two lines

> What browser are you working in?

Chrome

> What version of Carbon for IBM.com are you using?

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. Open TOC Horizontal storybook
2. Scale browser down
3. Step three
4. etc.

## Additional information

- Screenshots or code
- Notes

PR #5603, opened 2021-03-24
fix(toc): add overflow to horizontal variation 

ariellalgilmore:
### Related Ticket(s)

#5473 

### Description

Added overflow to TOC horizontal, so stacking of words no longer happens. 

Screen Shot 2021-03-24 at 1 31 39 PM

### Changelog

**New**

- added left and right arrow buttons
- intersectionObserver to detect when buttons are visible or hidden
- paginateRight/Left () to calculate how much to scroll

**Changed**

- adjusted multiple style changes to account for overflow changes
- updated storybook content to match design spec


photodow (changes requested):
- [ ] Alignment to grid is off. Not sure if it was introduced in this PR, or if we should tackle this in a sperate issue.

image

- [ ] Tabbing doesn't reveal hidden links, and last tab doesn't remove the last button. Probably related.

image

- [ ] Something weird happening with bottom spacing, selected state, and focus state. In the gif below, see how they kind of push each other out of the way?

Mar-26-2021 16-54-10

shixiedesign (commented):
Thanks @ariellalgilmore! You might be addressing James's comments already and I only a few comments:

The grid issue James already mentioned, but to be more specific, there are a few separate issues:  

1. Beyond `max` breakpoint, the ToC content should be within the max width. 
image

2. Within `max` breakpoint, the **Body** should start at the first column
image

3. `sm` and `md` breakpoints, the ToC and Body both don't align with start of grid. In this case, any text content should be snapping to the left (or beginning) of the column. 
image

None-grid issues are: 

4. Padding under the ToC should be 32px. Currently some of the padding seems to be under the ToC
image

5. Might be related to James's comment too and you might be addressing it already... I'm noticing the ToC content blinking on click
Mar-28-2021 22-23-40

Thank you!

ariellalgilmore:
closing this due to git issues. will create new PR

Issue #5493, opened 2021-03-15
[FeatureCardBlockMedium] icon does not get updated correctly


Claire-ibm-toulouse:
Hello all, here is my code: : 
https://github.com/Claire-ibm-toulouse/ois

and the page in staging: 
https://wwwstage.ibm.com/strategic-partnerships/fr-fr/ois/

On the last component "ContentBlockMedia", I am using the "feature" design, and trying to call the appropriate icon called [Launch20](https://carbon-elements.netlify.app/icons/examples/preview/#Launch20), but it keeps calling the "ArrowRight20" that was initially in the template.

This is a minor bug but it gives me a compliance issue. Please help, thanks.

<img width="637" alt="image" src="https://user-images.githubusercontent.com/74964904/111184624-6d09da80-85b1-11eb-87e3-5ddbed8393c5.png">

Also, do you happen to know why the background color of my "feature" card is dark grey when I deploy it (see above), instead of light grey when I run the code locally (see below)?
<img width="633" alt="image" src="https://user-images.githubusercontent.com/74964904/111185490-4e581380-85b2-11eb-88cc-a05bdb4452d5.png">

RobertaJHahn:
Kenny to respond as part of fireline.  May not be a bug.  Needs investigation.

Claire-ibm-toulouse:
Hello all, unfortunately the bug remains on my side. My code is in GitHub. Please help thanks.


Claire Lebastard

OFF on fridays
Risk & Compliance -  SE&TS (Sales Execution and Transaction Support)
IBM CIO  - Finance and Operations
+33 (0)562181617 – Toulouse - France


From: RobertaJHahn ***@***.***>
Reply to: carbon-design-system/carbon-for-ibm-dotcom ***@***.***>
Date: Tuesday 16 March 2021 at 19:08
To: carbon-design-system/carbon-for-ibm-dotcom ***@***.***>
Cc: Claire Lebastard ***@***.***>, Author ***@***.***>
Subject: [EXTERNAL] Re: [carbon-design-system/carbon-for-ibm-dotcom] [FeatureCardBlockMedium] icon does not get updated correctly (#5493)

[...]

PR #5608, opened 2021-03-24
fix(ContentBlockMedia): feature card cta icon src was being overwritten

annawen1:
### Related Ticket(s)

[FeatureCardBlockMedium] icon does not get updated correctly #5493

### Description

The feature card CTA icon within `ContentBlockMedia` was constantly overwritten in the `_renderFeatureCard()` method. There is no need to pass in the `icon.src` within the props for `FeatureCTA` or `ContentBlockMedia` utilizing `FeatureCTA` as the icon will be determined by the `type` (local, jump, external, etc.) being passed in instead. 

NOTE: There are some components where we define which icon to use and pass the icon as part of the props (FeatureCardBlockMedium, FeatureCardBlockLarge, Card, etc).

We can now set the external, jump, download, etc icon instead of it being overwritten to the default ArrowRight20.
Screen Shot 2021-03-24 at 5 13 38 PM

### Changelog

**Changed**

- make sure the `type` is passed in under the `card` object to ensure the correct icon is rendered in `_renderFeatureCard()`

**Removed**

- documentation in CTA (Feature Card CTA) and ContentBlockMedia where we pass the icon src


jeffchew (approved):
LGTM!

Issue #5150, opened 2021-02-11
Masthead currently allows theming where it should only allow the white theme


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

We have had numerous adopters run in issues where non-white theming is interfering with the Masthead, causing a dark masthead to appear.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

Masthead

> What did you expect to happen? What happened instead? What would you like to
> see changed?

We should lock in the white theme for the masthead and add documentation.

PR #5609, opened 2021-03-24
chore(masthead): lock masthead in white theme

IgnacioBecerra:
### Related Ticket(s)
#5150

### Description
By design, Masthead should only be using the Carbon White theme. However, there is an issue where it is still possible to change its color scheme using the other themes. This PR locks the Masthead into only using the White theme no matter what other themes are being applied to the page. This also applies to Masthead L1.

### Changelog

**New**

- ensured masthead classes use the Carbon White theme
- added note regarding this change to the Masthead documentation


annawen1 (approved):
LGTM!

Issue #5595, opened 2021-03-24
Nav Menu cut off when in Arabic language


yellowdragonfly:
> What version of Carbon for IBM.com are you using?
1.15.0
> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?
It is in production. So it would be nice to get fixed whenever possible.

## Steps to reproduce the issue

1. go to web site: https://proxy1.marketplace-test.ibmcloud.com/profile/dataprivacypreferences/ae-ar
2. Shrink the window
3. Click on "persona" icon
4. you will see the menu is cut off by the screen

## Additional information
![image (2)](https://user-images.githubusercontent.com/49878979/112362065-6adb0680-8caa-11eb-9b2c-2476b77c96df.png)
![image (3)](https://user-images.githubusercontent.com/49878979/112362100-762e3200-8caa-11eb-9202-760830d44ee9.png)

- Screenshots or code
- Notes

annawen1:
installing the `rtlcss` npm package as mentioned in our rtl docs here: https://ibmdotcom-react.mybluemix.net/?path=/docs/overview-enable-right-to-left-rtl--page might solve the issue

RobertaJHahn:
@yellowdragonfly  Did the suggestion by Anna fix this problem?

RobertaJHahn:
@yellowdragonfly we are closing this issue, assuming the that the issue has been resolved as a result of Anna's comments above.

PR #5638, opened 2021-03-25
fix(MastheadProfile): render profile menu flipped when html dir is rtl

annawen1:
### Related Ticket(s)

Nav Menu cut off when in Arabic language #5595

### Description

On rtl languages, the profile menu needs to be flipped so that it is not cut off. Solution is to detect the `dir` attribute of the document and set the `flipped` prop of the masthead profile overflow menu appropriately. 

Screen Shot 2021-03-25 at 5 42 16 PM

### Changelog

**Changed**

- set `flipped` prop to masthead profile overflow menu based on the `dir` attribute of the html


jeffchew:
Wonder if we should start doing some AVT over the RTL environment?

annawen1:
> Wonder if we should start doing some AVT over the RTL environment?

seems like we are getting some RTL bugs coming in now so it might be something I guess we can consider

ariellalgilmore:
looks like this actually isn't a bug and appearing fine on storybook. The person who submitted this did not install the rtlcss npm package

jeffchew:
@annawen1 just checked and doesn't look like overflow is working in RTL: https://ibmdotcom-react-rtl.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/5638/index.html?path=/story/components-masthead--with-custom-navigation, clicking the arrow doesn't appear to do anything.

jeffchew:
@annawen1 should we close this if the adopter needs to install `rtlcss` instead?

annawen1:
> @annawen1 should we close this if the adopter needs to install `rtlcss` instead?

Still waiting on adopter to confirm is installing `rtlcss` works, but we can just close this in the meantime

Issue #2460, opened 2020-05-20
Component: LightboxMediaViewer and CTA - greyed out scrollbar is displayed on ratio 2:1 (1024*512)


Praveen-sr:
## Detailed description

Device : Windows 
Browser : Chrome

Issue on LightboxMediaViewer and CTA sections.

Issue :- Scroll bar is displayed on ratio 2:1 (1024*512),  which is not required.

Expected : When the copy text is large the scroll option must be activated.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/33e66959-8606-4c72-b5af-0698b84f6648)

kennylam:
@Praveen-sr Which storybook environment are you seeing this on? I tried both below and cannot reproduce in Windows Chrome.

http://ibmdotcom-react.mybluemix.net/?path=/story/components-lightboxmediaviewer--default

http://ibmdotcom-react-staging.mybluemix.net/?path=/story/components-lightboxmediaviewer--default

RaphaelAmadeu:
This issue is happening due to `overflow`, as it's set to auto, it's going to have some resolutions that will display a blank scrollbar, I don't know if this is a bug by itself.

cc: @kennylam @jeffchew

NSDrowned:
Hello @Praveen-sr, does this issue still apply? I tested but couldn't reproduce.

NSDrowned:
Hello @Praveen-sr, any update on this one?

Praveen-sr:
@NSDrowned Re tested this on web components staging, still the issue exists. Scroll bar appears for 2:1 ratio

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/a8d92abe-5401-4b12-9080-4c199533cc1c)

On React for 2:1
![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/f301f864-09b8-496f-abe4-ccc85954068d)

CC @RobertaJHahn

RobertaJHahn:
Closing based on the comments in the PR 5640.  Not a suitable fix.

PR #5640, opened 2021-03-26
fix(lightbox-media-viewer): scrollbar disabled fix

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/2460

### Description

This is a very unusual problem that only occurs at a very small resolution range and from my tests, only on chrome for windows since macos hides scrollbars while not in use. Might be a browser bug but I couldn't find anything online.

To replicate you need to resize the window until the content is about to overflow. It will show a disabled scrollbar for a small range until it actually needs to enable it even if `overflow-y` is set to `auto`, which shouldn't happen.

scrollbar-bug

The range is always 16px which is exactly the width of the scrollbar.

For example, I found this 2 ranges where it happens: 
`1002px x 749px` to `1018px x 749px`
`1200px x 687px` to `1216px x 687px`

It's happening on both lightbox media viewer stories:

scrollbar-disabled

scrollbar-disabled-2

The only way I found to fix it was to use the deprecated css rule `overflow-y: overlay` which overlays the scrollbar instead of taking up space. It's still fully supported by webkit/blink-based browsers and from my tests doesn't affect other browsers but content might appear behind the scrollbar. This had to be applied to the expressive modal container.

### DIFF
--- a/packages/web-components/src/components/expressive-modal/expressive-modal.scss
+++ b/packages/web-components/src/components/expressive-modal/expressive-modal.scss
@@ -40,6 +40,8 @@
   .#{$prefix}--modal-content {
     display: grid;
     grid-template-rows: auto 1fr auto;
+    overflow: hidden;
+    overflow-y: overlay;
   }
 
   .#{$dds-prefix}-ce--modal__hedaer--with-body {


asudoh (commented):
Hi @NSDrowned thank you for your attempt to fix this. I don't think we want to change the scrolling behavior of the modal content; applications wants to have control over there instead. Also, I guess `overflow-y: overlay` is Mac-specific thing?

NSDrowned:
@asudoh, from my research only supported in WebKit-based (e.g., Safari) and Blink-based (e.g., Chrome or Opera) browsers, but deprecated. It worked on Chrome on my PC. Couldn't test it Mac OS since hides the scrollbars while not in use.

kennylam:
> Couldn't test it Mac OS since hides the scrollbars while not in use.

@NSDrowned You can change this behavior so you can test it.

System Prefs > General > `Show scroll bars`

image

NSDrowned:
@kennylam, thanks! Just tested it:

image

Seems to happen in chrome for mac os too, but not happening in Safari. I'm thinking it might be a Chrome bug. There's no reason for a disabled scrollbar to show if it's not needed.

NSDrowned:
@asudoh and @kennylam, I tried to find another solution but couldn't find anything that worked. Almost convinced it's bug in Chrome. Not sure what can be done about it or if it's worth waiting for a fix since it's a very small range when it happens. Any thoughts?

jeffchew:
@NSDrowned this can wait. Let's get the Carbon upgrade bugs in first.

jeffchew:
@NSDrowned just checking on this, is this PR still active?

NSDrowned:
Closing... not a suitable fix.

PR #5645, opened 2021-03-26
fix(table-of-contents): added smooth scroll in css

RaphaelAmadeu:
### Related Ticket(s)

#5470

### Description

I've applied `scroll-behavior: smooth` to our TOC stories, and updated the docs, so adopters can use that to have this behavior implemented.


packages/web-components/src/components/table-of-contents/__stories__/README.stories.mdx
```diff
@@ -30,8 +30,15 @@ import '@carbon/ibmdotcom-web-components/es/components/table-of-contents/table-o
 ```scss
 @import '@carbon/ibmdotcom-styles/scss/globals/vars';
 @import '@carbon/ibmdotcom-styles/scss/components/tableofcontents/tableofcontents';
+
+body {
+  scroll-behaviour: smooth;
+  overflow-y: scroll;
```
annawen1, line 36: I'm wondering if this will mess with the masthead megamenu behavior. When the megamenu is open, we disable scroll in the body tag

photodow (changes requested):
Having a couple of problems. Some of these may have already existed.

- [ ] Can we add some extra spacing to the bottom of this story maybe? The last item doesn't work.

Mar-26-2021 17-05-39

- [ ] Maybe a bit of an edge case, but if I click on a lower item, I thne try to scroll up in the middle of it canceling the smooth scroll perfectly, and then click on it again it doesn't work this time.


Mar-26-2021 17-09-26

- [ ] I'm not seeing a smooth scroll when I go over to one of the `scroll animation`, or `ui shell` stories that has table of contents in it as well.

Mar-26-2021 17-11-13

packages/web-components/.storybook/container.scss
```diff
@@ -40,6 +40,8 @@ body,
 [name='main-content'] {
   position: relative;
   height: 100%;
+  overflow-y: scroll;
+  scroll-behavior: smooth;
```
jeffchew, line 44: Adding this to all stories seems to have done some weird things across all stories, can this be set only for the story that needs it?

jeffchew (changes requested):

packages/web-components/src/components/table-of-contents/__stories__/README.stories.mdx
annawen1: I guess it should be fine since we have the class to disable it

packages/web-components/.storybook/container.scss
photodow: @jeffchew what weird things are you seeing out of curiosity? If it adds weird things, would that create weird things happening when someone builds a page with Table of contents?

jeffchew: Hey @photodow it might be more related to how the headless browser sets up the snapshots. The configuration only allows setting width and not height, and chrome driver specifically could be setting the frame with a shorter height as a side effect with the outer body element being set to overflow. So it might be a limitation of the testing suite, or maybe it's surfacing another issue.

RaphaelAmadeu: @jeffchew this problem was fixed, and now the changes Percy's pointing are intended, as per @photodow request.

jeffchew (approved):
Thanks @RaphaelAmadeu !

Issue #5615, opened 2021-03-25
Video Player Web Component not rendering video thumbnail


stephcard:
Hello! I'm super excited to be working with Carbon for IBM Web Components.

I noticed the sample video player component renders the video image by default (which is awesome) but that doesn't seem to be the case when implementing the component - the picture element and image does not render in the shadow dom. Nor does it render the picture element and image when specifying `thumbnail-url`.

Is this a bug? Any advice is much appreciated.

Screen grab attached.
![Screen Shot 2021-03-25 at 10 21 03 AM](https://user-images.githubusercontent.com/28735880/112501475-a8e63200-8d5f-11eb-97fd-316b187d7fe4.png)

Staging here: https://wwwpreview.ibm.com/repeatables/simple-scroll/

stephcard:
Hello. I was able to resolve this by including the image component. Please consider updating the video player component's documentation to mention this dependency. Thank you!

annawen1:
Hello @stephcard we will update the video player component so you will not have to import the image component yourself. Fix will come soon.

PR #5654, opened 2021-03-29
fix(video-player): import image component within video-player component

annawen1:
### Related Ticket(s)

Video Player Web Component not rendering video thumbnail #5615

### Description

Video player needs the image import otherwise the video image overlay does not appear.

### Changelog

**Changed**

- add image component import


jeffchew (approved):
Thanks @annawen1 !

jeffchew:
Just as an fyi @annawen1 @shixiedesign , I'm seeing some false positives around the async data for the video player, I'm going to add some skips for percy for those particular elements so we can avoid the false positives in the future.

image

jeffchew:
PR created: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/5656

PR #5656, opened 2021-03-29
chore(percy): skipping card group with mixed media in AVT

jeffchew:
### Related Ticket(s)

No related issue

### Description

The API calls for video data are causing false positives in the
automated visual regression testing. Because visual testing is already
captured in other stories, this story can probably be skipped in order
to avoid the false positives.

### Changelog

**Changed**

- Skipping Card Group with mixed media story in Percy


ariellalgilmore (approved):
LGTM!

Issue #5505, opened 2021-03-16
[Dropdown/Combobox] issues on screen reader (JAWS) in Firefox


RobertaJHahn:
## Detailed description
This issue has been resolved in Chrome.  This work was managed under issue (#4889 ).  

Copy of the final comment with accessibility tester, Ram:

_Ram has tested the "edit combo" in both chrome and FF.
The behavior you mentioned is found in Chrome.
In FF, the Edit combo role is not identified by jaws. And also the selected value is not announced after making a selection. hence the user would not know the currently selected option.
Regarding the jaws announcement of the input value along with each option in chrome, I think that is tolerable as long as we can achieve the same behavior in FF as in chrome. I would mark that as low sev or even a best practice._

----

Issue:
1. There is no label
2. Screen reader announces this as a list box. Hence the user does not know that this is an Edit combo.
3. When we navigate within the combo box using up/down arrow keys, Screen reader does not announce anything. Hence Unable to select the preferred option 

In React staging, The language selector control is not identified by jaws in arrow key navigation. An issue is already opened on the same and is being tracked.

Expected:
1. Provide appropriate label
2. If this is an editable combo box, Screen reader should convey the "edit combo" role to the user 
3. Screen reader Users should be able to navigate and make the selection within the combo box.

PR #5658, opened 2021-03-30
fix(language-selector): reimplementing components to fix jaws bug

IgnacioBecerra:
### Related Ticket(s)
#4889

### Description
The original Language Selector component extended the `BXComboBox` component from `carbon-web-components`, which has a bug with JAWS where the dropdown options weren't being read when using the up/down arrow keys. 

There was a [PR](https://github.com/carbon-design-system/carbon-web-components/pull/582) in that repo fixing this bug in Chrome + FF (there is another issue (#5505) tracking a fix for Firefox specifically), but upon further discussion with @asudoh we decided it would be better to have the current available fix for Chrome within Carbon for IBM.com until a solution for both browsers can be found, which would then be addressed within `carbon-web-components`.

This current PR extends `BXDropdown` and `BXComboBox` from `carbon-web-components` in order to be used with `language-selector-desktop`, only adding a few lines to the existing code to apply the aforementioned fix.

### Changelog

**New**

- introduced slight variations of `BXDropdown` and `BXComboBox` to fix a JAWS accessibility bug with Chrome for use with Language Selector

**Changed**

- `language-selector-desktop` now extends `DDSComboBox`


packages/web-components/src/components/footer/defs.ts
```diff
@@ -27,6 +29,8 @@ export enum FOOTER_SIZE {
   MICRO = 'micro',
 }
 
+// TODO: Remove when JAWS bug has been fixed
```
ariellalgilmore, line 32: can this TODO be removed now?

IgnacioBecerra: I added that; those new lines come from the `carbon-web-components`'s defs file. Only added as a note so whenever the bug gets fixed in that repo we can remove all the additions from this PR, including the stuff here in `defs`

ariellalgilmore (approved):

Issue #5568, opened 2021-03-23
React: Leadspace block - On hover state, link don't get the grey background, instead links are underlined.


Praveen-sr:
## Detailed description

Issue:
On hover state, link don't get the grey background, instead links are underlined.

## Additional information

Actual:
![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/99f052ca-266b-4e4c-89e1-d9ad0e2736b4)

Expected:

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/0d124d61-b55a-46ca-b534-71d0f8c8a254)

PR #5659, opened 2021-03-30
fix(link-list): added proper background on hover and removed underline

IgnacioBecerra:
### Related Ticket(s)
#5568

### Description
In React, `LinkList` CTA class lacked a background change on hover, underlining the text instead. This PR addresses this issue.

Before:
Screen Shot 2021-03-29 at 5 29 01 PM

After:
Screen Shot 2021-03-29 at 5 29 09 PM

### Changelog

**New**

- ensured proper background color appeared upon hover
- removed underline upon hover with LinkList


shixiedesign:
I wonder if we can fix this in someway so Percy stop seeing the hover text. But anyhow, I approved this diff. 

image

shixiedesign (commented):
I understand this issue is for hover state, so this bug might be out of scope - I noticed the focus state of this component is wrong. Lemme know if you can fix it here or I can make a new issue: 

**Current:**
image

**Expected**
`2px solid inside outline in $focus`

image

IgnacioBecerra:
@shixiedesign Added a fix for #5674 into this PR. The underlining shouldn't happen anymore in Card.

shixiedesign (approved):
Looks good! Thank you so much! 👍

jeffchew (approved):
LGTM!

PR #5570, opened 2021-03-23
fix(table-of-contents): spacing fixes for toc

kennylam:
### Related Ticket(s)

#5500 
#5564 

### Description

Fixes spacing in `TableOfContents` for React and web components.

### Changelog

**Changed**

- mobile select `padding-left` is now `16px`
- fixes header spacing in Storybook stories for React and web components

PR #5662, opened 2021-03-30
fix(toc): add overflow to horizontal

ariellalgilmore:
### Related Ticket(s)

#5473 

### Description

Added overflow to TOC horizontal, so stacking of words no longer happens. 

Screen Shot 2021-03-24 at 1 31 39 PM

### Changelog

**New**

- added left and right arrow buttons
- intersectionObserver to detect when buttons are visible or hidden
- paginateRight/Left () to calculate how much to scroll
- onHandleKeyDown() for tabbing accessibiility

**Changed**

- adjusted multiple style changes to account for overflow changes
- updated storybook content to match design spec


photodow (approved):
Odd spacing at the top for mobile, but I think we have an issue for that (#5570). This has come a long way, and is looking good. Thanks @ariellalgilmore. @shixiedesign any thoughts?

shixiedesign (approved):
Awesome work! Thank you Ari 👍

andysherman2121 (approved):

Issue #5145, opened 2021-02-11
Web component: Masthead-With custom data - We are able to reach to the "Back button" but there is no visual focus at all.


Praveen-sr:
## Detailed description

Back button in 200% zoom text
Issue: We are able to reach to the "Back button" but there is no visual focus at all.

Steps:-
1) Activate hamburger menu 
2) Activate any of the elements for example (Products & Solutions Services & Consulting, Learn & support, Explore more, Covid19).
3) After activating any of the above element respective section will open with the "Back button"

Expected Result:
Provide thick and strong visual focus.

PR #5672, opened 2021-03-30
fix(left-nav-menu): adding focus to back button

guilhermelMoraes:
### Related Ticket(s)

Web component: Masthead-With custom data - We are able to reach the "Back button" but there is no visual focus at all.#5145

### Description

Added `tabindex="0"` to back button element (that is currently an anchor tag without an `href`) and a handler for `keypress`.

### DIFF
--- a/packages/web-components/src/components/masthead/left-nav-menu.ts
+++ b/packages/web-components/src/components/masthead/left-nav-menu.ts
@@ -60,6 +60,15 @@ class DDSLeftNavMenu extends FocusMixin(LitElement) {
   }
 
   /**
+   * Handler for the `keypress` event on the expando button.
+   */
+  private _handleKeydownExpando({ key }: KeyboardEvent) {
+    if (key === 'Enter') {
+      this._handleUserInitiatedToggle();
+    }
+  }
+
+  /**
    * `true` if the menu should be in its active state.
    */
   @property({ type: Boolean, reflect: true })
@@ -94,7 +103,14 @@ class DDSLeftNavMenu extends FocusMixin(LitElement) {
   }
 
   render() {
-    const { active, backButtonText, expanded, title, _handleClickExpando: handleClickExpando } = this;
+    const {
+      active,
+      backButtonText,
+      expanded,
+      title,
+      _handleClickExpando: handleClickExpando,
+      _handleKeydownExpando: handleKeydownExpando,
+    } = this;
     const buttonClasses = classMap({
       [`${prefix}--side-nav__submenu`]: true,
       [`${prefix}--masthead__side-nav--submemu--selected`]: active,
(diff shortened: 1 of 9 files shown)


packages/web-components/src/components/masthead/left-nav-menu.ts
```diff
@@ -115,8 +131,14 @@ class DDSLeftNavMenu extends FocusMixin(LitElement) {
...
       <ul class="${prefix}--side-nav__menu" role="menu">
-        <li class="bx--side-nav__menu-item bx--masthead__side-nav--submemu-back" role="none">
-          <a class="bx--side-nav__link" role="menuitem" @click="${handleClickExpando}">
+        <li class="bx--side-nav__menu-item bx--masthead__side-nav--submemu-back">
+          <a
+            class="bx--side-nav__link"
+            role="menuitem"
+            @keydown="${handleKeydownExpando}"
+            @click="${handleClickExpando}"
+            tabindex="0"
```
asudoh, line 140: Weighing in; `<a>` works as `tabindex="0"` by default.

guilhermelMoraes: It doesn't if you remove the `href`, @asudoh 

https://codesandbox.io/s/wizardly-matsumoto-4r1t5?file=/src/App.js

asudoh: In such case, do you want to add `href` back? The `<a>` features we lose are more than focusing if we remove `href`.

guilhermelMoraes: This element never had an `href`  attribute. How can we use it?

asudoh: Would you want to try something like [this](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.16.1/packages/web-components/src/components/masthead/left-nav.ts#L136)?

guilhermelMoraes: I did it and it makes it stop working. Do you know why we're using an `anchor` tag instead of a `button type="button"`?

asudoh: I don't know. If `<button>` is OK for us I think it'll definitely be a good idea.

guilhermelMoraes: @asudoh coming back on this again before any changes: the React version is currently using an `a` tag. Also, changing the tag would require updating the styles as well. So I'm wondering if we need to change it.

asudoh: In such case, do you want to deep-dive why adding `href` broke things?

annawen1:
@guilhermelMoraes I'm refactoring the MastheadLeftNav a bit and changed the back button to use the `button` element instead in React so far: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/5677

I can do the same for web components as well in another PR I'm working on

packages/web-components/src/components/masthead/left-nav-menu.ts
guilhermelMoraes: @asudoh I've been testing it but couldn't make it work with the `href`.

guilhermelMoraes:
@annawen1 got it. Do you want me to close this PR?

annawen1:
> @annawen1 got it. Do you want me to close this PR?

yeah I think we can close this PR

PR #5677, opened 2021-03-31
feat(MastheadLeftNav): REACT - refactor mobile left nav and add animations

annawen1:
### Related Ticket(s)

React: Change Mega Menu to include animation (back button & forward motion) #4835

Web component: Masthead-With custom data - We are able to reach to the "Back button" but there is no visual focus at all. #5145

### Description

Refactoring work on the mobile menu nav to add animation when traversing through the menu and also to address scroll issues that we have seen/still see.

Previously the menus were nested within each other, this PR pulls all the menu sections out into the same level

For reference: [Masthead Animation BOX folder](https://ibm.ent.box.com/folder/128256664360?s=bqn9xggrys3gcyuy0q0ubmf0w1b3gwvg)

animation

In order to get tabbing to work, I set the focus on the back button when user opens up a submenu. Let me know if this is an issue.

### Changelog

**New**

- restructure the menu to all the menus are separate and on the same level instead of submenu in submenu in menu.
- `SideNavMenuBackButton` component
- `SideNavMenuSection` component

**Changed**

- switch the "Back" button in WC masthead nav to use `button` element instead of `anchor` so that it is focusable


photodow:
Does this work impact what's going on over at #5672?

annawen1:
> Does this work impact what's going on over at #5672?

This is changing just the React side for now, so it won't impact the change for the WC version. I see they are changing the `Back` button to use the `button` element instead of `anchor` tag - I'll switch it over to button on this PR as well

shixiedesign:
Hi Anna! So glad to see this work done! It's a big lift. I found a few issues and would like to keep looking closely. High visibility component. 

1. Most severe bug I noticed is - on mobile, esp. `Learn & Support > What is...` vertical scroll quite work. User cannot access the last two menu options on the list. See video. I'm on iPhone10, Safari 
https://user-images.githubusercontent.com/15144993/113649845-2f065080-965d-11eb-8f72-5b1b6cfc7dbe.MP4

2. Focus on `X` should not be constant. Currently `X` button always get the blue-60 2px outline, but this should only show when the button is focused. Two items being focused should never happen as it's also an accessibility issue. I did notice the motion prototype video had `X` always focused and that's a mistake

image

3. Motion wise, it's always tricky because I'm also not sure what's possible. I'm still seeing a bit of overlap between the panels and I'm wondering if it can be avoided. 

4. We can skip animating the chevron rotating. It's better to stay still while the panel moves it out. Having a moving element on the moving panel creates a lot of noise.

Apr-05-2021 22-43-40

5. Something happened to web components... is this related? Current WC canary doesn't show below errors
image

annawen1:
@shixiedesign yeah the WC version I am still working on and is going to be a different PR so only React changes for now. Plan is to branch the WC PR off this one and merge at the same time so that both versions will be the same.

annawen1:
> Hi Anna! So glad to see this work done! It's a big lift. I found a few issues and would like to keep looking closely. High visibility component.
> 
> 1. Most severe bug I noticed is - on mobile, esp. `Learn & Support > What is...` vertical scroll quite work. User cannot access the last two menu options on the list. See video. I'm on iPhone10, Safari

Ok, looks like I'll need to add some bottom padding to the menu panels

>    https://user-images.githubusercontent.com/15144993/113649845-2f065080-965d-11eb-8f72-5b1b6cfc7dbe.MP4
> 2. Focus on `X` should not be constant. Currently `X` button always get the blue-60 2px outline, but this should only show when the button is focused. Two items being focused should never happen as it's also an accessibility issue. I did notice the motion prototype video had `X` always focused and that's a mistake

This may be tricky as the way I refactored it, each menu panel is separate from the other so in order for tabbing to work I need to have something in each visible menu panel focused. I will see what I can do...

> 
> image
> 
> 1. Motion wise, it's always tricky because I'm also not sure what's possible. I'm still seeing a bit of overlap between the panels and I'm wondering if it can be avoided.

This is happening because each panel only knows two states - unexpanded and expanded. Since there's two different speeds for back and forward motion, it looks like it's overlapping. Can we change to only one speed?

> 2. We can skip animating the chevron rotating. It's better to stay still while the panel moves it out. Having a moving element on the moving panel creates a lot of noise.

Sure, I'll take that out

annawen1:
@shixiedesign  
FIXED:
>  1. Most severe bug I noticed is - on mobile, esp. `Learn & Support > What is...` vertical scroll quite work. User cannot access the last two menu options on the list. See video. I'm on iPhone10, Safari

> 2. Motion wise, it's always tricky because I'm also not sure what's possible. I'm still seeing a bit of overlap between the panels and I'm wondering if it can be avoided.

^ Kinda fixed this, had to set back and forth motion to the same speed for it to not overlap.

> 3. We can skip animating the chevron rotating. It's better to stay still while the panel moves it out. Having a moving element on the moving panel creates a lot of noise.

-------------------------------------------
> 1. Something happened to web components... is this related? Current WC canary doesn't show below errors

This PR is only for React, WC version coming in another PR. So ignore any WC changes here.

> 2. Focus on `X` should not be constant. Currently `X` button always get the blue-60 2px outline, but this should only show when the button is focused. Two items being focused should never happen as it's also an accessibility issue. I did notice the motion prototype video had `X` always focused and that's a mistake

This is something I see specifically in our masthead specs: https://ibm.ent.box.com/file/716827406287?s=00ieobl9wsxjsxb3rhs7y6o7mccurx5r Can you double check and let me know if we keep the focus on the 'X' at all times when menu is open?

jeffchew:
@annawen1 is there supposed to be focus after a user clicks the menu button? 

image

annawen1:
> @annawen1 is there supposed to be focus after a user clicks the menu button?
> 
> image

Since it's on the `Products & Solutions` menu panel, the focus should actually be on the "Back" button. I've set it with js currently so that after the transition event ends, the "Back" button becomes focused.

jeffchew:
When navigating, shift-tab is doing some weird things:

Kapture 2021-04-07 at 15 02 09

annawen1:
> When navigating, shift-tab is doing some weird things:
>

annawen1:
> When navigating, shift-tab is doing some weird things:

Adjusted it so that we user does a Shift + Tab when on the "Back" button, it traps the focus within the menu panel - bringing the focus to the last menu item.

shift and tab

jeffchew:
This looks good to me from an engineering perspective. Since this is a pretty huge change might be good to get a second pair of eyes. @shixiedesign @photodow any additional comments for design?

jeffchew (approved):

shixiedesign (commented):
Hi Anna! I have found only an existing issue (also in react prod) - is it outside the scope of this PR? Because if it's outside scope, then THIS PR IS PERFECT!!! 🎉 🕺 🎉 💃 🎉 🎆 🎉  (tested on Chrome and my iPhone 10s Safari)

Underline under link came back on hover... "View all products" is okay though. Only in "essentials" panel: 
image
Expecting: 
image

shixiedesign (approved):
Aight I'm gonna say that underline in link is out of scope. Thank you Anna for the amazing work! (I spent another 20min testing and couldn't find bugs, easing numbers and durations and everything 👍 )

oliviaflory (approved):
LGTM!

Issue #5657, opened 2021-03-29
Bug when implementing back-to-top component


amanda-mendes-1998:
## Detailed description

> Describe in detail the issue you're having.
The back-to-top component renders on my DOM but it automatically sets a "hidden" value to the tag even if I add the "hidden='false' on it

> Is this a feature request (new component, new icon), a bug, or a general
> issue?
a bug

> Is this issue related to a specific component?
yes, back-to-top component

> What did you expect to happen? What happened instead? What would you like to
> see changed?
I wanted the SVG icon to render and the back-to-top action to happen but none of these are working.

> What browser are you working in?
Using Chrome but also tested on Firefox

> What version of Carbon for IBM.com are you using?
here's a screenshot of my package-json carbon dependencies:
<img width="501" alt="Screen Shot 2021-03-29 at 15 27 55" src="https://user-images.githubusercontent.com/80002122/112882429-57ada980-90a3-11eb-983f-0ab0610b50c4.png">

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?
The request for this page is meant to be published on March 31 so is very urgent. Currently working for ibm.com/partners segment.

## Steps to reproduce the issue

1. Step one: Add the back to top component (mine was added within my main content <div>)
2. Step two: check your localhost to see that the svg doesn't show up

## Additional information

- Screenshots or code
- how it looks when I inspect the page:
![image](https://user-images.githubusercontent.com/80002122/112882703-ace9bb00-90a3-11eb-8b00-e006cc291c3e.png)
- here's a example snippet of my code:
```
<div>
//main content goes here
<dds-back-to-top></dds-back-to-top>
</div>
```

- Notes
I've already sent a message on the support Slack channel and was oriented to open this issue.

PR #5699, opened 2021-04-01
fix(back-to-top): fix visibility toggling

kennylam:
### Related Ticket(s)

#5657 

### Description

Changes the implementation of `back-to-top`, which works in Storybook but not in standalone apps. Perhaps due to Storybook's DOM structure, certain elements targeted by the component did not return the correct values to trigger the `back-to-top` button when used in standalone applications.

### Changelog

**Changed**

- use `scrollingElement` instead of `documentElement` to observe `scrollTop` value
- update Codesandbox example

**Removed**

- `ResizeObserver` for root element (`this.ownerDocument.documentElement`). This element would return inconsistent values when not used in Storybook.


jeffchew (approved):
LGTM!

PR #5700, opened 2021-04-02
chore(dependencies): upgrade `sass-loader`

asudoh:
### Description

Upgrades `sass-loader` to the latest version that support WebPack4, so that we can use `webpackImporter: false` option  to disable `sass-loader`'s custom Sass importer that was one of the bottlenecks of Storybook build.

`sass-loader`'s custom Sass importer is meant for non-standard `@import` with `~`, what resolves dependencies in `node_modules`. However, we are guiding our users to use `SASS_PATH` environment variable or `includePaths` option in `node-sass`.

Also:

* Introduces `cache-loader` in there to further boost the performance.
* Upgrades `css-loader`, `node-sass` and `style-loader`.

### Changelog

**Changed**

- Uses `cache-loader` for Sass build.
- Starts using `sass-loader` instead of `fast-sass-loader`, for the sake of `webpackImporter: false`.
- Upgrades `css-loader`, `node-sass` and `style-loader`.


asudoh:
Note: http://ibmdotcom-react-canary.mybluemix.net/ does not seem to reflect carbon-design-system/carbon#7093 and carbon-design-system/carbon#7518 somehow which seems to have caused Percy errors.

jeffchew:
> Note: http://ibmdotcom-react-canary.mybluemix.net/ does not seem to reflect [carbon-design-system/carbon#7093](https://github.com/carbon-design-system/carbon/pull/7093) and [carbon-design-system/carbon#7518](https://github.com/carbon-design-system/carbon/pull/7518) somehow which seems to have caused Percy errors.

All of the icons are bumped down a pixel or two, which is why we're seeing all the diffs. Any thoughts why this is happening?

asudoh:
> All of the icons are bumped down a pixel or two, which is why we're seeing all the diffs. Any thoughts why this is happening?

@jeffchew The humber icon diff is caused by https://github.com/carbon-design-system/carbon/pull/7093/files#diff-89fd4df83e945a84c57fde50daa364196c3ee93c104fc29e883f1f511d251ac6R95-R97. The lang selector X icon diff is caused by https://github.com/carbon-design-system/carbon/pull/7518/files#diff-2d8f4b8c5cc37d2869d1f6ac7a1a1ceb847d8634c26c7bf48fe69f0305dcd084R447.

shixiedesign:
It appears everything with an `x` icon is affected? This is seen in Web components. 
image

shixiedesign:
Thank you! Please give me a min to go through the diffs. Maybe they are intentional since the fixes seem to be made for a reason. just wanna double check.

jeffchew:
> Thank you! Please give me a min to go through the diffs. Maybe they are intentional since the fixes seem to be made for a reason. just wanna double check.

Thank you @shixiedesign, I'm holding back on approving the diffs but otherwise the webpack changes look fine to me @asudoh 👍

jeffchew (approved):
Approving from an engineering perspective. Having design look at the visual diffs.

asudoh:
Thanks @shixiedesign and @jeffchew for reviewing - And expressive button is great to bring up, seems that http://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-expressive-modal--default has wrong paddings (and thus wrong height). The deploy preview one has correct height (`48px` IIUC).

shixiedesign:
Yes there are some good diffs! But these two are not right. Should we create issues? 

- Hamburger icon is not centered any more. Original is correct 
image

- `X` icon in dropdown got brighter. Original is correct 
image

asudoh:
Thank you for the details @shixiedesign!

> - Hamburger icon is not centered any more. Original is correct 

Found that our code has [an adjustment](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1.17.0/packages/styles/scss/components/masthead/_masthead.scss#L361) that doesn't work with [upstream Carbon core change](https://github.com/carbon-design-system/carbon/pull/7093/files#diff-89fd4df83e945a84c57fde50daa364196c3ee93c104fc29e883f1f511d251ac6R95-R97) (change in approach wrt how to center the icon). Fixed. 

- `X` icon in dropdown got brighter. This affects Micro footer and Dotcom shell. Original is correct

[Carbon core made it from `$icon-02` to `$icon-01`](https://github.com/carbon-design-system/carbon/pull/7518/files#diff-2d8f4b8c5cc37d2869d1f6ac7a1a1ceb847d8634c26c7bf48fe69f0305dcd084R447), do you mean we should make a customization to change it back...?

shixiedesign:
Thank you for catching me up. By the look of it, the icon token change should be intentional. Lemme trace back to Carbon designers just to confirm.

shixiedesign:
Hey @asudoh looks like hamburger is fixed but now Search icon is bumped 1px too high 😅 Can it be looked at again? 
image

And I confirmed `icon-01` update is intentional and approved the diff. Thanks!!

shixiedesign (approved):
This is amazing. Thank you so much!

Issue #5415, opened 2021-03-09
[Card group] Several storybook improvements


shixiedesign:
## Tasks
Card group storybook example should be improved in the following way to better demo the design and function of this component:
- [ ] Apply grid to card group, so the component should be snap to the grid col 5-16
- [ ] Apply different amount of content in cards to demo `same height util` in action

Also just noticed the lovely thin 1px lines between cards! 👍 Not sure when this was done but great job! 

### New sub-stories:
**Card group - with card in card**
![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/80ca21ae-d51a-4328-a637-c086ba42385b)

**Card group - with card in card and image card**
![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/2384dc76-a754-4c1c-be45-8f3d1aeb9f1e)

PR #5703, opened 2021-04-02
fix(card-group): random title percy check

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5415

### Description

Added `inPercy` check for random title lengths.

### DIFF
--- a/packages/web-components/src/components/card-group/__stories__/card-group.stories.ts
+++ b/packages/web-components/src/components/card-group/__stories__/card-group.stories.ts
@@ -19,6 +19,7 @@ import ArrowRight20 from 'carbon-web-components/es/icons/arrow--right/20';
 import { html } from 'lit-element';
 import { select, number } from '@storybook/addon-knobs';
 import ifNonNull from 'carbon-web-components/es/globals/directives/if-non-null.js';
+import inPercy from '@percy-io/in-percy';
 // eslint-disable-next-line sort-imports
 import imgXlg4x3 from '../../../../../storybook-images/assets/1312/fpo--4x3--1312x984--003.jpg';
 import imgXlg16x9 from '../../../../../storybook-images/assets/1312/fpo--16x9--1312x738--005.jpg';
@@ -29,7 +30,7 @@ import styles from './card-group.stories.scss';
 
 import readme from './README.stories.mdx';
 
-const cardRandomPhrase = () => {
+const cardRandomPhrase = randomTitles => {
   const phraseArray = [
     'Lorem ipsum dolor sit amet',
     'Nunc convallis lobortis',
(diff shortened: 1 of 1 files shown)


packages/web-components/src/components/card-group/__stories__/card-group.stories.ts
```diff
@@ -243,7 +247,7 @@ export default {
       CardGroup: ({ groupId }) => ({
         cards: Array.from({
           length: number('Number of cards', 5, {}, groupId),
-        }).map(() => cardRandomPhrase()),
+        }).map(() => cardRandomPhrase(inPercy())),
```
jeffchew, line 250: Is it necessary to pass this as an argument? Can't you just use it in the method itself?

NSDrowned: Yeah, I can do that. I'll change it right away.

NSDrowned: Done 👍

packages/web-components/src/components/card-group/__stories__/card-group.stories.ts
```diff
@@ -37,7 +38,10 @@ const cardRandomPhrase = () => {
     'Te sint disputando pri, at his aliquip corrumpit',
   ];
 
-  const randomSampleText = phraseArray[Math.floor(Math.random() * phraseArray.length)];
+  const randomSampleText = inPercy()
+    ? 'Lorem ipsum dolor sit amet, consectetur'
+    : phraseArray[Math.floor(Math.random() * phraseArray.length)];
+
```
jeffchew, line 44: Thanks @NSDrowned !

I'm wondering if we should still pass in varying text sizes, but keep them static, just so we can also capture varying heights in Percy too.

NSDrowned: Working on that 👍

packages/web-components/src/components/card-group/__stories__/card-group.stories.ts
```diff
@@ -29,18 +29,19 @@ import styles from './card-group.stories.scss';
...
+const phraseArray = [
+  'Lorem ipsum dolor sit amet',
+  'Nunc convallis lobortis',
+  'Lorem ipsum dolor sit amet, consectetur.',
+  'Te sint disputando pri, at his aliquip corrumpit',
+  'Disputando lorem covallis',
+];
 
-  const randomSampleText = phraseArray[Math.floor(Math.random() * phraseArray.length)];
+const cardsDiffLenghtPhrase = () => {
```
jeffchew, line 41: minor thing, `length` is misspelled...

NSDrowned: Whoops, fixed 😳

packages/web-components/src/components/card-group/__stories__/card-group.stories.ts
NSDrowned: I removed the random phrase functionality and instead I just loop through the different phrases in order so it always shows the same cards for users and for percy. When ready, we'll still be able to see the `same height` function in action and increasing the card count still works normally.

jeffchew (approved):
Looks good! One thing we might want to look at later on (not really imperative for this PR) is to maybe cycle through the phrase Array in case a developer goes in and adds more cards for any reason, but I guess if that happens they can add the functionality when needed.

NSDrowned:
@jeffchew it does cycle through the array, they can add as many cards as they want 👍

Issue #5397, opened 2021-03-05
[CTA] Card link heading style error


photodow:
## Detailed description
Our CTA cards under content blocks, and such should be using Card link, and card link is visually incorrect. We probably should make some adjustments to the visuals in the library, and add a style type specifically for card link within CTA so we can continue using Card a little more generically.

Affected components: [Card link](https://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-card-link--default), [CTA card](https://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-cta--card), 

**Currently**
Heading style is `$body-long-02`, which is 16px regular.
![image](https://user-images.githubusercontent.com/3793636/110172145-5352e100-7dc2-11eb-8934-0c2d00a614cd.png)

**Expected**
Heading style should be `$expressive-heading-01`, which is 16px semibold.
![image](https://user-images.githubusercontent.com/3793636/110172336-9a40d680-7dc2-11eb-8522-9fbdcd5f4199.png)

## Things to do

- [ ] visually correct card link [development]
- [ ] update CTA component with card link style [functional specs] [development]
- [ ] update components like content group and content block to use card link with card link style. [development]

## Additional information

- [CTA](https://ibm.box.com/s/unvd38oewwguycqoqg0et75eluiiqhg6)
- [Card link specs](https://ibm.box.com/s/1u2m0fqdtrumog0ie2c4tfyyyn1clyj7)
- Card link storybook [[web components](http://ibmdotcom-web-components.mybluemix.net/?path=/story/components-card-link--default)] [[react](http://ibmdotcom-react.mybluemix.net/?path=/story/components-card-link--default)]
- CTA storybook [[web components](http://ibmdotcom-web-components.mybluemix.net/?path=/story/components-cta--text)] [[react](http://ibmdotcom-react.mybluemix.net/?path=/story/components-cta--text)]

PR #5711, opened 2021-04-05
fix(card-link): changed style to expressive-heading-02

IgnacioBecerra:
### Related Ticket(s)
#5397

### Description
The current font style for `CardLink` and `CTACard` was using `$body-long-02`, while it should be using `$expressive-heading-02`. This PR changes the styles for both React and Web Components.

### Changelog

**Changed**

- changed `CardLink` and `CTACard` styles to using `$expressive-heading-02`
- updated `CardLink` Web Component README to remove the use of `<p>` for proper stylings to be applied


oliviaflory:
1. Text alignment is off in the heading of Content block simple with aside elements in React (actually all layout components "with aside elements". Not sure if this is caused in this PR, so lmk if I should open a separate issue 
Screen Shot 2021-04-07 at 12 59 46 PM

2. The CTA card or Card link is not spanning full width on mobile as I expected in most layout component examples
Preview
Screen Shot 2021-04-07 at 1 00 48 PM
Expectation
Screen Shot 2021-04-07 at 1 01 35 PM

3. Commented in Percy, the Card group body text changed to $expressive-heading-02, but should remain as $body-long-02
Screen Shot 2021-04-07 at 1 05 49 PM

IgnacioBecerra:
Hi @oliviaflory! For the heading with aside element issue, it's not caused by this PR, but I think I may know what's actually causing it. If you open a separate issue for this, you can assign me to it. 

And the other two issues have been fixed in the latest commit 👍

photodow:
Are we just changing the font size? Should we be using the card link component here instead? Otherwise why do we have the component? Does that make sense?

jeffchew (approved):
LGTM!

IgnacioBecerra:
With the latest commit, I made it so only the Web Components Card Link is updated. I also left over the fix for CTA not spanning full width.

oliviaflory (approved):
Approving with understanding that other issues will be fixed in #5740 and logo grid PR will fix those Percy differences

Issue #5060, opened 2021-02-02
Web component: Leaving IBM - There is no label for the modal dialog


Praveen-sr:
## Detailed description

On JAWS FF/Voiceover iPhone
Issue:
There is no label for the modal dialog

Expected:
Provide the heading present in the dialog as the label by associating the same using aria-labelledby

andysherman2121:
@Praveen-sr Can you retest this now that we've applied the solution?

Praveen-sr:
@andysherman2121 Re tested this Web component/React -
Jaws FF
In react wrapper, Screen reader does not identify the dialog label when we open the modal for the first time. But when we refresh the page, the label is identified. 
In web component Canary the label is identified hence working fine.
iPhone VO
Dialog label not identified in both the environments.

RobertaJHahn:
Fixed in Web Components v1.18.0. Needs to be fixed in React.

PR #5712, opened 2021-04-06
chore(dependencies): upgrade `carbon-web-components`

asudoh:
Refs #5060 

### Description

To `v1.13.0`.

### Changelog

**Changed**

- Upgrades `carbon-web-components` to `v1.13.0`.


andysherman2121 (changes requested):
@oliviaflory Left some comments on the Percy build. Looks like the button on Modal isn't right

asudoh:
Good catch @oliviaflory - Fixed.

asudoh:
Seems that Percy flakiness happening for cards with `sameHeight()`. CC @jeffchew

jeffchew (approved):
LGTM!

@asudoh yes, we never had any luck with same height in Percy, I'm thinking of skipping those in Percy tests.

andysherman2121 (approved):

PR #5716, opened 2021-04-06
fix(masthead): global styles updates

mherchel:
### Related Ticket(s)
- No public GitHub ticket.
- JIRA ticket from IBM - [[FE] Update the existing Carbon Masthead/L0 component to reflect the new design](https://jsw.ibm.com/browse/HC-1679)

### Description
- [x] All L0 menu item text reduced to 14px (from 16px) now using $body-short-01 
- [x] All drop down menu item text sizes reduced to 14px (from 16px) now using $body-short-01 
- [x] All L0 menu icons reduced to 16px (from 20px)
- [x] All L0 menu items using $text-02 for initial state (changes to $text-01 on hover and selected states)
- [x] All drop down menu items using $text-02 for initial state (changes to $text-01 on hover and selected states)
- [ ] ~"Log in" text for logged out state of the menu (no drop down in logged out state)~
- [ ] ~User icon for account drop down once logged in (menu items will vary between IBM.com and Cloud)~

### Design Specs

- Invision Prototype - https://ibm.invisionapp.com/share/G6O0VIFJPNB#/screens/319853575
- Please see this Project View doc (https://ibm.ent.box.com/notes/775640792107) for the links to the Masthead and Mega Menu components directories that are housing the design specs and sketch files.

image

### Changelog

**New**

- {{new thing}}

**Changed**

- {{changed thing}}

**Removed**

- {{removed thing}}


mherchel:
Note that I didn't make any changes to the chevron down icon within the desktop navigation. This icon is imported through a different package (see https://github.com/carbon-design-system/carbon-web-components/blob/master/src/components/ui-shell/header-menu.ts#L13). 

However, the icons are 10x6, so they shouldn't require any changes.

proeung:
@mherchel Thanks for fixing up the CI issues. I think this PR is ready for code and design QA, so please change the status of this PR from a draft to publish.

> "Log in" text for logged out state of the menu (no drop down in logged out state)
>  User icon for account drop down once logged in (menu items will vary between IBM.com and Cloud)

Also, the two items that you've listed above have already been accounted for in @huntermacd's [PR ](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/5593)that got merged as an experimental feature flag.

shixiedesign (changes requested):
1. I'm seeing all the links in the library are impacted by this PR and that should not happen. 

image

2. Corporate masthead should not be affected. 

image


image

packages/styles/scss/themes/expressive/_tokens.scss
```diff
@@ -116,13 +116,13 @@ $carbon-productive-type-tokens: (
 /// @group carbon-expressive
 $carbon-expressive-type-tokens: (
   body-short-01: (
-    font-size: carbon--type-scale(3),
+    font-size: carbon--type-scale(2),
```
annawen1, line 119: Don't think we should be changing the font size here as it will affect every component using these tokens.

annawen1: If we have to use `body-short-01` but we want 14px instead of 16px - there is a `@include use-carbon-productive-tokens();` mixin that we use.

shixiedesign:
Had a quick chat with @proeung and I understood how this happened. So FYI @mherchel the current `body-short` maps to 16px is correct because in dotcom we utilize an expressive theme that bumps up original Carbon token values ([see explanation here](https://www.ibm.com/standards/web/carbon-for-ibm-dotcom/guidelines/expressive-theme)) . We need to keep that unchanged for now so the rest of Dotcom is unaffected. 

Please ping whenever the changes are in place. Thank you!

proeung:
I think we can close out this PR. I'm going to create a new one that we're using for the overall design QA of the Cloud masthead work.

annawen1:
Closing this PR as @proeung will be opening another one

Issue #5382, opened 2021-03-05
[Tabs extended]


RobertaJHahn:
### Overview

**Tabs extended** `ui`
![image](https://user-images.githubusercontent.com/3793636/110023769-28e62280-7cf3-11eb-9d50-af5e4fe9c6c0.png)

### Dev
- Assigned to Ian, estimated to be 5 
----
### BA
- [Tabs extended Functional specs](https://ibm.ent.box.com/folder/133802998765?s=26zrugvym00408zhpp23gqwnudzmgdd1)

PR #5720, opened 2021-04-07
feat(tabs-extended): Tabs extended component

iwhitcomb:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5382
https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5411

### Description

We're creating a new component that mirrors the BXTabs behavior but collapse down into a BXAccordion on mobile screens.

The challenge here is that I can't simply extend BXTabs/BXAccordion because we need it to render the markup for BOTH tabs and accordions. The BXTabs markup also isn't structured in a way that supports this type of implementation. In order to get this working, I created two new LitElement components called `<dds-tabs-extended>` and `<dds-tab>`. The dds-tab is just the pane content and contains parameters for the tabs/accordion. The dds-tabs-extended has a little bit of logic to parse through the dds-tab components, and create tabs and the accordion using Carbon classes. It also contains a little bit of logic for switching the active tab on click.

### Changelog

**New**

- Tabs extended
- Tab


packages/web-components/src/components/tabs-extended/tab.ts
```diff
@@ -0,0 +1,95 @@
...
+           aria-hidden="${!this.selected}">
+        <slot></slot>
+      </div>
+    ` : html`
+      <div id="tab-panel-${this.index}-default"
+           class="tab-${this.index}-container"
+           role="tabpanel"
+           aria-labelledby="tab-link-${this.index}-default"
+           aria-hidden="${!this.selected}"
+           hidden>
```
asudoh, line 82: In this way I guess conditional rendering is not needed?

```suggestion
           ?hidden="${!this.selected}">
```

packages/web-components/src/components/tabs-extended/tab.ts
```diff
@@ -0,0 +1,95 @@
...
+   * Defines the selected state of the tab.
+   */
+  @property({ reflect: true })
+  selected: Boolean = false;
+
+  /**
+   * Defines the index of the tab relative to other tabs.
+   */
+  @internalProperty()
+  private index: Number;
```
asudoh, line 58: Prefixing private properties with `_` is our naming convention:

```suggestion
  private _index: Number;
```

packages/web-components/src/components/tabs-extended/tabs-extended.ts
```diff
@@ -0,0 +1,134 @@
...
+                  <div id="pane-${index}" class="bx--accordion__content">
+                    ${tab.innerHTML}
+                  </div>
+                </li>
+              `;
+            })}
+          </ul>
+        </div>
+
+        <div data-tabs class="${prefix}--tabs">
```
asudoh, line 97: `data-tabs` for Carbon vanilla API, so I think it's not needed?

packages/web-components/src/components/tabs-extended/tabs-extended.ts
```diff
@@ -0,0 +1,134 @@
...
+        <div class="${prefix}--accordion">
+          <ul data-accordion class="bx--accordion">
+            ${tabs.map((tab:DDSTab, index) => {
+              const classes = [
+                index === this.active && `${prefix}--accordion__item--active` || null,
+              ];
+              return html`
+                <li data-accordion-item class="bx--accordion__item ${classes.join(' ')}">
+                  <button class="bx--accordion__heading" aria-expanded="${index === this.active}" aria-controls="pane-${index}" @click="${(e) => this._setActiveTab(index)}">
+                    <svg focusable="false" preserveAspectRatio="xMidYMid meet" style="will-change: transform;" xmlns="http://www.w3.org/2000/svg" class="bx--accordion__arrow" width="16" height="16" viewBox="0 0 16 16" aria-hidden="true"><path d="M11 8L6 13 5.3 12.3 9.6 8 5.3 3.7 6 3z"></path></svg>
```
asudoh, line 85: Does this help? https://github.com/carbon-design-system/carbon-web-components/blob/v1.13.0/src/components/accordion/accordion-item.ts#L175-L178

packages/web-components/src/components/tabs-extended/tabs-extended.ts
```diff
@@ -0,0 +1,134 @@
...
+    } = this;
+    return html`
+      <div class="${prefix}--tabs-extended">
+
+        <div class="${prefix}--accordion">
+          <ul data-accordion class="bx--accordion">
+            ${tabs.map((tab:DDSTab, index) => {
+              const classes = [
+                index === this.active && `${prefix}--accordion__item--active` || null,
+              ];
```
asudoh, line 81: Should use: https://lit-html.polymer-project.org/guide/template-reference#classmap

packages/web-components/src/components/tabs-extended/tabs-extended.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  render() {
+    const {
+      _tabItems: tabs,
+    } = this;
+    return html`
+      <div class="${prefix}--tabs-extended">
+
+        <div class="${prefix}--accordion">
+          <ul data-accordion class="bx--accordion">
```
asudoh, line 77: `data-accordion` is for Carbon vanilla API , so it's not needed?

packages/web-components/src/components/tabs-extended/__stories__/README.stories.mdx
```diff
@@ -0,0 +1,43 @@
+import { Props, Description } from '@storybook/addon-docs/blocks';
+import contributing from '../../../../../../docs/contributing-license.md';
+import markdownContents from '../../../../docs/markdown-contents.md';
+
+# `<dds-content-item-horizontal>`
```
asudoh, line 5:
```suggestion
# `<dds-tabs-extended>`
```

asudoh (commented):
Just weigh in with some comments - Thanks @iwhitcomb!

packages/web-components/src/components/tabs-extended/tab.ts
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * @element dds-tab
+ */
+@customElement(`${ddsPrefix}-tab`)
+class DDSTab extends StableSelectorMixin(LitElement) {
+  /**
+   * Defines label of the tab.
+   */
+  @property({ reflect: true })
+  label = null;
```
asudoh, line 28: This line may yield to type inference to `null`, though it looks like a string. `label = ''` may be more desirable?

packages/web-components/src/components/tabs-extended/tab.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  /**
+   * Defines title of the tab.
+   */
+  @property({ reflect: true })
+  title = null;
```
asudoh, line 34: This line may yield to type inference to `null`, though it looks like a string. `title = ''` may be more desirable?

packages/web-components/src/components/tabs-extended/tabs-extended.ts
```diff
@@ -0,0 +1,151 @@
...
+                @click="${e => this._setActiveItem(index)}"
+              >
+                ${ChevronRight16({
+                  part: 'expando-icon',
+                  class: `${prefix}--accordion__arrow`,
+                })}
+                <div class="${prefix}--accordion__title">${tab.label}</div>
+              </button>
+              <div id="pane-${index}" class="${prefix}--accordion__content">
+                ${tab.innerHTML}
```
asudoh, line 90: I guess `.innerHTML` here is rendered as a string instead of a HTML? https://github.com/carbon-design-system/carbon-web-components/blob/v1.13.0/src/components/select/select.ts#L78-L111 may help.

asudoh (commented):
Thanks @iwhitcomb for the update! Wanted to have other devs chime in from now on.

shixiedesign:
Hi Ian! A few things I noticed: 

1. An important keyboard nav bug: `tab` and `shift+tab` is not letting me get to the first tab 🤔 

Apr-08-2021 14-19-19

2. HTML didn't render in accordion: 

Apr-08-2021 15-19-33

3. Accordion's chevron is too small. Should be 20px: 

image

4. The other points below are storybook refinements so less critical. The general rule is we should render components left aligned. See more [storybook standards](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/wiki/Storybook-format-standard#3-rendering-in-viewport). We are still in the process of getting better at these ourselves. 

image

3.

packages/web-components/src/components/tabs-extended/tabs-extended.ts
iwhitcomb: Hey @asudoh I like what you're doing here, although this appears to strip the markup as well. Prior to this, `.innerHTML` was encoding all of the HTML so we we're seeing the markup in the rendered content. We got around this using `${unsafeHTML((tab as DDSTab).innerHTML)}`

jeffchew:
@iwhitcomb it's looking good!

The docs is showing up weird, mainly the proptable isn't rendering as intended:

image

There's also some lint errors showing up if you can take a look...

shixiedesign:
Thank you for the quick fixes @iwhitcomb ! Here are a few more things: 

1. Container height currently is 40px. Expected 48px. Are you not using the [expressive theme tabs](https://carbon-expressive.mybluemix.net/?path=/story/tabs--default)? Our expressive theme tab is showing a bug too and rendering at 44px, but this tab component should just be pulling that one in. Just confused how it got to 40px 😅 If possible, please adjust the spacing below text so we get 48px total (including the 3px border bottom). 

image

According to the design it should be 48px 
image

shixiedesign:
sorry accidental close!

photodow (commented):
Might be able to get by merging, and addressing this later, but the Carbon accordion has a transition when it opens and closes that I'm not seeing here.

###### Ours

Apr-09-2021 09-15-06


###### Carbon's

Apr-09-2021 09-32-33

jeffchew:
> Might be able to get by merging, and addressing this later, but the Carbon accordion has a transition when it opens and closes that I'm not seeing here.

Maybe it's an upstream thing, are you looking at `carbon-web-components`? 
https://web-components.carbondesignsystem.com/?path=/story/components-accordion--default

iwhitcomb:
@jeffchew no, it's definitely an issue with tabs-extended because I'm not able to directly extend the tabs or the accordion objects. Instead, I had to basically reproduce the functionality from scratch and I haven't gotten to the accordion transitions yet.

jeffchew:
> @jeffchew no, it's definitely an issue with tabs-extended because I'm not able to directly extend the tabs or the accordion objects. Instead, I had to basically reproduce the functionality from scratch and I haven't gotten to the accordion transitions yet.

As discussed offline with @iwhitcomb , let's round back and see if this will require functional spec updates, @photodow . As Ian also mentioned, Carbon accordion allows for multiple panes to be open where this might be an awkward experience if we want the accordion to behave more like the Tabs if we only want one pane open at a time. I think we should discuss further on this one but doesn't seem to be launch gating to me and at least get this through for the release.

photodow:
As mentioned above I don't think my comments above should block this. The accordion item open one at a time is per functional specs, so that's good. It's just missing the animation which I wouldn't call a blocker. My concern which sounds like Ian has at least commented on is that because it's not there something is broken somewhere. It sounds like that broken bit is he had to recreate it from scratch which causes hesitation on my part. Again I don't think I would block this PR from going in over that at the moment.

photodow:
I don't think functional specs need to be updated on any of this, but do let me know if y'all think otherwise.

shixiedesign:
Hi @iwhitcomb 
1. another small bug: after the accordion is clicked on, focus state there is correct, but the accordion title shouldn't stay gray, but should return to default background color. 
Apr-09-2021 11-32-59

2. so sorry to ask but could you give it 16px padding from top in the view port? It helps to let us see the top rule of accordion especially. 
image

iwhitcomb:
@shixiedesign I fixed the viewport margin. As for the gray hover/active state... what you're seeing is the difference between mobile and desktop. You're testing the tabs-extended on a mobile display(which won't have a hover state) and the carbon accordion on desktop(which does have hover state).

If you test both on mobile or both on desktop you'll see that they match. :)

shixiedesign (approved):
Thanks Ian!!

annawen1 (approved):
LGTM!

Issue #5599, opened 2021-03-24
[Carbon v10.31.0 Upgrade | Expressive] UI Header icons look smaller


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

![image](https://user-images.githubusercontent.com/1641214/112371249-c5796000-8cb4-11eb-863c-c314f67f5c5d.png)

> Is this issue related to a specific component?

UI Header (and possibly Masthead too)

> What did you expect to happen? What happened instead? What would you like to
> see changed?

It seems like icons are a bit smaller, should take a look and adjust this.

> What version of Carbon for IBM.com are you using?

Current branch `chore/carbon-10.31.0-upgrade`. **Fixes should be targeted to this branch.**

## Additional information

1. Original PR: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/5566
2. Deploy Preview: https://carbon-expressive.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/5566/index.html
3. Percy Build: https://percy.io/538fc19a/Carbon-Expressive-Theme/builds/9518540?utm_campaign=538fc19a&utm_content=Carbon-Expressive-Theme&utm_source=github_status_public

PR #5722, opened 2021-04-07
fix(ui-header): removed unnecessary padding around icons

IgnacioBecerra:
### Related Ticket(s)
#5599

### Description
The new carbon upgrade adds some additional classes to the `button` elements, one of which adds some padding calculation around the icons, making them smaller. This PR ensures that the UI Header Icons remain the same size as they were before the upgrade by setting the padding to 0.

### Changelog

**New**

- set `padding: 0` under `bx--header__action` class to ensure icons stay the same size


shixiedesign:
Almost! When the UI shell icons are selected, the icon still moved a bit: 
Apr-07-2021 15-51-34

IgnacioBecerra:
@shixiedesign I'm comparing between the current version and the new, and the behavior is similar for both. There is `.bx--header__action-active` that adds a 1px border everywhere except top. There is currently a 2px border surrounding icons if they're not active, wondering if the correct behavior should be always having a 2px border, even when active?

packages/styles/scss/themes/expressive/components/_ui-shell.scss
```diff
@@ -34,4 +34,19 @@
...
+
+    &--active {
+      border: rem(2px) solid transparent;
+      outline: none;
+    }
+  }
+
+  .#{$prefix}--btn.#{$prefix}--btn--icon-only.#{$prefix}--tooltip__trigger:focus {
+    outline: none;
+    border-color: #f4f4f4;
```
jeffchew, line 49: We don't have a token for this?

IgnacioBecerra: Yep, I just found it, `$ui-01`. Just pushed a commit using it.

IgnacioBecerra: Yeah, I removed the blue focus since that didn't seem to be present in the [current Expressive example](http://carbon-expressive.mybluemix.net/?path=/story/ui-shell--header-base-w-actions-and-right-panel). 

Let me know what's the proper way to handle the new focus

oliviaflory: Previously we've tried to maintain as much parody with the changes that come from Carbon directly so that we are not maintaining unnecessary overrides on our end. _Unless_ something is very broken or not accessible for whatever reason, and if/when that happens we can open bugs in their repo to fix something too. 

I don't have insight into exact reasons why they altered the focus state to include the blue+white, but we've previously trusted the updates to be correct.

I hope that makes sense!

oliviaflory: Alright, @shixiedesign just checked with them and apparently it's a 🐛 

So removing the blue focus was the correct move, ignore above comments.

jeffchew:
From an engineering perspective, things look ok, but would love to see a percy diff before we move this forward. Hopefully that will be back online soon.

shixiedesign (approved):
I see all the fixes and thank you Ignacio! I also misses percy very much so I leave the merge tag for Jeff to decide. Thanks!

jeffchew (approved):
LGTM!

Issue #5598, opened 2021-03-24
[Carbon v10.31.0 Upgrade | Expressive] Style updates needed for new NumberInput/FormItem


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

![image](https://user-images.githubusercontent.com/1641214/112369680-d6c16d00-8cb2-11eb-9229-4eaff6da59d4.png)

![image](https://user-images.githubusercontent.com/1641214/112370076-57806900-8cb3-11eb-88d8-acb1b64bfa3b.png)

- [ ] icons are super tiny, should be 20px per expressive theme
- [ ] extra bottom border appearing within the container
- [ ] bottom border has spaces in it, should be continuous line
- [ ] hover state does not extend to whole container height

> Is this issue related to a specific component?

NumberInput

> What version of Carbon for IBM.com are you using?

Current branch `chore/carbon-10.31.0-upgrade`. **Fixes should be targeted to this branch.**

## Additional information

1. Original PR: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/5566
2. Deploy Preview: https://carbon-expressive.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/5566/index.html
3. Percy Build: https://percy.io/538fc19a/Carbon-Expressive-Theme/builds/9518540/changed/537658459?browser=chrome&browser_ids=16&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1280&widths=320%2C1280
4. Carbon update for reference: https://react.carbondesignsystem.com/?path=/story/components-numberinput--default

![Screen Shot 2021-03-24 at 4 43 24 PM](https://user-images.githubusercontent.com/45692622/112381973-a3d2a580-8cc1-11eb-8946-547b99ef0b00.png)
![Screen Shot 2021-03-24 at 4 47 35 PM](https://user-images.githubusercontent.com/45692622/112381991-a8975980-8cc1-11eb-9e2e-b1216f5740f9.png)

PR #5729, opened 2021-04-07
fix(number-input): ensured icon and container fit within parent properly

IgnacioBecerra:
### Related Ticket(s)
#5598

### Description
With the newest Carbon upgrades, the FormItem/NumberInput components were slightly reworked, which clashed with our current dimension styles for both the button and svg. This update aligns our Expressive theme and the new component updates.

### Changelog

**Changed**

- made the number control buttons fit their container
- increased the svg's dimensions to 20px


oliviaflory:
Hey @IgnacioBecerra I was actually expecting that in the upgrade we would be adopting the new version of the Number input from Carbon to align with the `-` and `+` symbols. It looks like we've reverted to the previous design with the carets?

Screen Shot 2021-04-08 at 10 53 17 AM

IgnacioBecerra:
@oliviaflory Hm, I see the `-` and `+` symbols when working on this locally, I'll take a look as to why this might be happening on the preview.

IgnacioBecerra:
@oliviaflory There was a dependency bumped down in an upgrade (that contained the new icons) which made the preview use the previous icons. After talking to Jeff, we pushed a change for that and the preview now contains the correct, updated icons. 👍

oliviaflory (approved):
Looks good to me, thanks @IgnacioBecerra 🎉 

Guessing I'm having trouble logging into Percy to approve changes because of the migration Jeff mentioned 😞

shixiedesign (approved):
👍 Percy diffs look good too

jeffchew (approved):
LGTM!

PR #5742, opened 2021-04-09
feat(announcement-banner): Think 2021 announcement banner code

rsebade:
### Related Ticket(s)

#5512 

### Description

Adds an AnnouncementBanner component that pulls it's data from the v18 masthead banner JSON (https://www.ibm.com/common/v18/mastheadbanner/masthead-banner.json) so that the DDS team can enable announcements that are synced between v18 and carbon. The PR includes both a component itself that can be added to pages, and a masthead prop, hasAnnouncementBanner, that renders the announcement banner above the masthead, in the same place the v18 one currently exists. 

### Changelog

**New**

- AnnouncementBanner component, service, and styles. 

**Changed**

- Masthead files, added prop hasAnnouncementBanner and code to shift the masthead below the banner when enabled. Edited the onScroll handler in the masthead.js file to fix the masthead once it reaches the top of the viewport.


jeffchew:
@rsebade as discussed, we will be creating a standalone package architecture for this, so will close this PR. Thank you @rsebade !

Issue #5596, opened 2021-03-24
[Carbon v10.31.0 Upgrade | React] Masthead icons are white


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

![image](https://user-images.githubusercontent.com/1641214/112366909-b17f2f80-8caf-11eb-80e8-33e0c275890b.png)

![image](https://user-images.githubusercontent.com/1641214/112367848-d1632300-8cb0-11eb-8f2e-630ab2717e17.png)

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

Masthead

> What did you expect to happen? What happened instead? What would you like to
> see changed?

The icons to not be white.

> What version of Carbon for IBM.com are you using?

Current branch `chore/carbon-10.31.0-upgrade`. **Fixes should be targeted to this branch.**

## Additional information

1. Deploy Preview: https://ibmdotcom-react.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/5566/index.html?path=/story/components-masthead--default
2. Percy Build: https://percy.io/538fc19a/Carbon-for-IBM.com-React/builds/9518436/changed/537649382?browser=chrome&browser_ids=16&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1280&widths=320%2C1280

PR #5756, opened 2021-04-12
fix(masthead-search): increasing css selectors specificity

guilhermelMoraes:
### Related Ticket(s)

[Carbon v10.31.0 Upgrade | React] Masthead icons are white #5596

### Description

The Carbon upgrade added some new selectors to the `HeaderGlobalAction` component that were overriding the styles set at the `MastheadSearch` buttons. This was fixed by increasing the specificity of the previous selectors.

### DIFF
--- a/packages/styles/scss/components/masthead/_masthead-search.scss
+++ b/packages/styles/scss/components/masthead/_masthead-search.scss
@@ -125,6 +125,11 @@
     position: absolute;
     top: 0;
     right: 0;
+
+    .#{$prefix}--header__search--search,
+    .#{$prefix}--header__search--close {
+      color: #000000;
+    }
   }
 
   // search container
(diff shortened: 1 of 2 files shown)


jeffchew:
Looks like the initial build bundle test passed, but another one needs to be addressed. I'll take a look at this after this PR is merged in:

image

jeffchew:
@guilhermelMoraes cannot approve Percy since we're locked out of it but the diff looks good!

jeffchew (approved):
Thank you @guilhermelMoraes for the fix! 

As stated above, the Percy diff cannot be approved since we're locked out of the projects.

packages/styles/scss/components/masthead/_masthead-search.scss
```diff
@@ -125,6 +125,11 @@
     position: absolute;
     top: 0;
     right: 0;
+
+    .#{$prefix}--header__search--search,
+    .#{$prefix}--header__search--close {
+      color: #000000;
```
annawen1, line 131: think we can use the `$ui-05` token instead?

jeffchew:
```suggestion
      color: $icon-01;
```

Issue #5305, opened 2021-03-01
[Tabs extended - with media] a.k.a content item horizontal with media - in tabs


shixiedesign:
### Overview

Final design spec: https://ibm.box.com/s/ios0b4wzrkxou1aquxitacngx4nyz345

![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/b02b56f1-349e-41d9-a460-4612bdf9c29b)
![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/5977c976-bee3-4f78-9724-3f2772a306b9)

### Dev
Assigned to Ian, estimated to be 5.

----
#### BA additional info

**Tabs extended** `ui`
![image](https://user-images.githubusercontent.com/3793636/110023769-28e62280-7cf3-11eb-9d50-af5e4fe9c6c0.png)

**Tabs extended — with media** `block` `layout`
![image](https://user-images.githubusercontent.com/3793636/110023782-2f749a00-7cf3-11eb-9dba-08e80cb7ba5e.png)

> I'm going to suggest we don't build the `section` level equivalent of these components since it's simply putting `content section` together with these. Definitely open to hearing others' opinions on this though.
- [Tabs extended with media functional specs](https://ibm.ent.box.com/folder/133803243554?s=3sun8lo3gu2b4uaa4oxgpxrqit1m7t6i)

RobertaJHahn:
Issues needed:  component naming for the tabs, Func specs, wc dev issue, Visual QA, Prod QA

PR #5759, opened 2021-04-12
feat(tabs-extended-media): introduce tabs extended with media component

iwhitcomb:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5305
https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5422

### Changelog

**New**

- {{new thing}}

* Tabs extended w/ Media

- {{changed thing}}

**Removed**

- {{removed thing}}


packages/web-components/src/components/tabs-extended/__stories__/README.stories.mdx
```diff
@@ -20,6 +20,7 @@ Here's a quick example to get you started.
 
 ```javascript
 import '@carbon/ibmdotcom-web-components/es/components/tabs-extended/tabs-extended.js';
+import '@carbon/ibmdotcom-web-components/es/components/tabs-extended/tabs-extended-media.js';
```
jeffchew, line 23: I'm wondering if there should be two separate sections in the documentation for just `dds-tabs-extended` and for `dds-tabs-extended-media`. If an adopter only wants `dds-tabs-extended`, they won't have any need for the extended with media version so wouldn't be something they need to import.

packages/web-components/src/components/tabs-extended/__stories__/README.stories.mdx
```diff
@@ -44,6 +45,48 @@ import '@carbon/ibmdotcom-web-components/es/components/tabs-extended/tabs-extend
...
+  </dds-tab>
+  <dds-tab label="Fourth tab">
+    <p>Content for fourth tab goes here.</p>
+  </dds-tab>
+  <dds-tab label="Fifth tab" disabled="true">
+    <p>Content for fifth tab goes here.</p>
+  </dds-tab>
+</dds-tabs-extended-media>
+```
+
```
jeffchew, line 89: Can't target below this, but a couple of notes:

1. Need to add `dds-tabs-extended-media` to stable selectors
2. Just doublechecking if the `markdownContents` is needed here?

packages/styles/scss/components/tabs-extended-media/_tabs-extended-media.scss
```diff
@@ -0,0 +1,73 @@
...
+  :host(#{$dds-prefix}-tabs-extended-media) {
+    margin: 0;
+    grid-column: 1 / span 16;
+    @include carbon--breakpoint(sm) {
+      padding-top: $carbon--spacing-07;
+      padding-bottom: $carbon--spacing-12;
+    }
+    @include carbon--breakpoint(lg) {
+      padding-top: $carbon--spacing-10;
+      padding-bottom: 10rem;
```
kennylam, line 28: Can use `$spacing-13` here.

kennylam (commented):
Thanks for this @iwhitcomb! Just a small comment.

shixiedesign:
Thank you Ian! Here are a few visual issues: 

1. Grid not aligned. The Content section title and tabs should all snap to grid: 
Current: 
image
Expected:
image


2. content section heading should top align with the Tabs labels - I've heard from Putra our content section might not be set up in the best way in terms of spacing and I'm interested in improving it. For now, if it's relatively easy, please make sure the headline and tab labels are top aligned.  
image

3. Please default `item horizontal with media` to media on the left

4. Please fill the other tabs with the `item horizontal with media` as well. Same placeholder content is fine. We wanna avoid the impression that mixing content type in tabs is allowed where in fact it's not. 

5. on mobile, content item horiz should have bottom spacing 48px. Currently too small. 
image

kirbyevans:
1. Medium breakpoint: The media should have a max-width of 640px.
Current:
Pasted_Image_4_13_21__9_39_AM

Expected:
image

2. Small breakpoint: Media should span all four columns.
Current:
image

Expected:
image

packages/styles/scss/components/tabs-extended-media/_tabs-extended-media.scss
iwhitcomb: It doesn't look like `$carbon--spacing-13` exists.

iwhitcomb:
Per my convo with @kirbyevans it sounds like we're going to leave the image max-width alone since it's primarily a result of the accordion styles coming from Carbon. Rather than trying to override this, we're going to stick with what Carbon provides.

> Ian Whitcomb:peanutbutterjelly:  9:47 AM
> I'm awaiting some feedback on how to answer your Q's about the max-width 640 on the image there.... that's an interesting problem.
> 
> Kirby Evans  10:07 AM
> ok cool cool. Thanks for the update.
> 
> Ian Whitcomb:peanutbutterjelly:  10:07 AM
> Basically, there are a couple things going on. I remember correctly, you can either enforce a 640px max-width OR have it lined up to the grid. There was that extra grid column with the content-item-horizontal-media that pretty enforced this on it's own so it was an easy answer...
> 
> Kirby Evans  10:08 AM
> Ah ok. hmm.
> 10:08
> so in other places it's used as an either or?
> 
> Ian Whitcomb:peanutbutterjelly:  10:09 AM
> In this particular case I'm nesting a content-item-horizontal-media which has it's own styling and rules, inside the tabs-extended-media which also has it's own set of styling and the question is whether or not the higher level component can/should dictate styles to the lower level item, first of all.
> 10:09
> But also the accordion inside tabs-extended-media has padding that's coming from Carbon. So should we attempt to change that. :man-shrugging::skin-tone-2:
> 10:09
> Override*
> 
> Kirby Evans  10:10 AM
> Ah. got it. Yeah I think the max-width rules would apply at the content-item-horizontal-media level.
> 
> Ian Whitcomb:peanutbutterjelly:  10:10 AM
> At that particular breakpoint, the accordion adds 25% padding on the right... it's weird.
> 
> 1 reply
> Today at 10:11 AMView thread
> 
> Kirby Evans  10:11 AM
> In terms of the carbon padding issue, not sure what to do there. Maybe we keep consistent with Carbon and disregard my feedback on that?
> 
> Ian Whitcomb:peanutbutterjelly:  10:11 AM
> That's exactly my question too.
> 10:11
> I'm just not sure how they typically address things like that.
> 
> Kirby Evans  10:12 AM
> Yeah, I lean staying consistent with Carbon for now. It's just a minor visual difference so I'd say let's stay consistent.
> 
> Ian Whitcomb:peanutbutterjelly:  10:13 AM
> ok, yeah.. I'll leave it alone. It seems like we're seeing a number of other issues that need to be addressed as well(i.e. the grid problem)
> 
> Kirby Evans  10:13 AM
> yeah, I think those are more important.

shixiedesign:
Gotcha... thanks for that detailed explanation and love the diagram! Glad you are in convo with Kirby on this design. Lemme know whenever I should take another peak. It's really not far and looking great :)

andysherman2121:
@shixiedesign Look's like Ian fixed a few of your comments. Could you provide a re-review? 
@kennylam is `$carbon--spacing-13` a usable token?

shixiedesign (approved):
Wooohooo! Thank you!

packages/styles/scss/components/tabs-extended-media/_tabs-extended-media.scss
jeffchew: strange, there should be here: https://carbondesignsystem.com/guidelines/spacing/overview/

andysherman2121: @jeffchew it looks like it only goes up to `$carbon--spacing-13`. I just checked locally

jeffchew (approved):
Thank you so much @iwhitcomb !

PR #5761, opened 2021-04-12
fix(MastheadSearch): carbon upgrade - masthead search button tooltip

annawen1:
### Related Ticket(s)

### Description

Carbon upgrade made the tooltip appear on hover for icon only buttons which is the masthead search buttons are using. Set the tooltip to be right aligned so it doesn't get cut off in smaller breakpoints

BEFORE:
Screen Shot 2021-04-12 at 12 51 29 PM

AFTER:
Screen Shot 2021-04-12 at 3 35 26 PM

### Changelog

**Changed**

- set tooltip alignment prop to 'end'


jeffchew (approved):
Thanks @annawen1 !

annawen1:
discussing with design at the moment to see if this is actually something we want to keep

annawen1:
spoke with designers, we are keeping the tooltip - just changing the alignment so it doesn't get cut off in smaller breakpoints

shixiedesign (approved):
Looking good!

ariellalgilmore (commented):
looks like the blue focus is missing now? 
Screen Shot 2021-04-12 at 3 00 31 PM

PR #5765, opened 2021-04-12
fix(footer): language selector icon alignment fix

NSDrowned:
### Description

Icon alignment was off in language selectors:

https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/5760/index.html?path=/story/components-footer--default-language-only

Fixed in micro footer too.


jeffchew (approved):
Thanks @NSDrowned !

PR #5770, opened 2021-04-13
fix(MastheadLeftNav): set aria-hidden to unseen menu panels

annawen1:
### Related Ticket(s)

[Mobile Menu]: VO reads all menu items including ones not visually seen when menu is open #5771

### Description

Set `aria-hidden` when the menu panel is not visible so that VO does not read those panels.

### Changelog

**Changed**

- set `aria-hidden` when menu panel is not showing


jeffchew (approved):
Looks good to me!

IgnacioBecerra (approved):
Tested it out, works as expected. Thanks @annawen1!

Issue #4614, opened 2020-12-03
Web component: Card section - Images - beginning and ending boundary for the card is not conveyed for the screen reader user.


Praveen-sr:
## Detailed description

Issue:
1. The beginning and ending boundary for the card is not conveyed for the screen reader user.
2. In tab / shift tab navigation, when focus reaches the link in each card, Screen reader announces a lengthy text which makes it oververbose.

Expected:
1. Provide each card within the landmark region and give appropriate label.
2. It may be possible if the entire content in the card is given in anchor tag. In this scenario, only the heading can be given in anchor tag. Note: the link text/label should be concise and convey only the relevant information.

Note: working fine on react version.

RobertaJHahn:
@ariellalgilmore Added Ari to issue (#4613 ) as the same solution should be implemented for all Card variants.  Once the fix is done, Ari, can you check to be sure that the fix addresses this bug and move to Done.  Thanks.

Praveen-sr:
@ariellalgilmore Ram re tested this on both JAWS and Voiceover on the [ link ]( https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/6642/index.html?path=/story/overview-getting-started--page)  - Working as expected.

Issue #4613, opened 2020-12-03
React: Card section - simple - beginning and ending boundary for the card is not conveyed for the screen reader user. 


Praveen-sr:
## Detailed description

Issue: 
1. The beginning and ending boundary for the card is not conveyed for the screen reader user. 
2. In tab / shift tab navigation, when focus reaches the link in each card, Screen reader announces a lengthy text which makes it oververbose.

Expected:
1. Provide each card within the landmark region and give appropriate label.
2. It may be possible if the entire content in the card is given in anchor tag. In this scenario, only the heading can be given in anchor tag. Note: the link text/label should be concise and convey only the relevant information. 

Note: working fine on react version.

## Additional information
- this fix should address the bug included in issue (#5504 ). Engineer should check to be sure this fix addresses the bug included in (#5504 )
 - Reach out to kenny for additional details.

kennylam:
@Praveen-sr I'm not sure this is the correct approach for card links. As I understand it, `landmark` roles are meant to identify larger regions of a page, not individual elements within it. I also don't understand why we'd want to reduce the amount of copy read by screen readers, since it is meant to inform users of the content they'd expect to see when clicking on it.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Region_role#Accessibility_concerns

cc: @jeffchew @RobertaJHahn

Praveen-sr:
@ariellalgilmore Ram re tested the PR provided, here is the report 

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/51fb7e45-af3a-4b94-bf7b-305df78121dd)

RobertaJHahn:
@ariellalgilmore Can this issue and (#5504 ) be closed?

ariellalgilmore:
@RobertaJHahn We just confirmed with the team the changes that will be made for this issue to be closed, so next sprint I will finish this issue. 5504 and mostly all the other card accessibility bugs require this fix being made.

ariellalgilmore:
@RobertaJHahn only the React version has been completed

RobertaJHahn:
@ariellalgilmore Thanks for reminding me.  I have created issue (#6005 ) and assigned it to you in the next sprint.  Closing this issue.

PR #5777, opened 2021-04-14
fix(card): refactor for accessibility updates

ariellalgilmore:
### Related Ticket(s)

#4613 #4614 
### Description

REACT - Due to Card previously using Carbon's Tile Clickable  component,  an `<a>` is wrapped around the entire card causing over verbose issues while tabbing. This refactor uses Carbon `Tile: Default` and Carbon `Link` in the footer. This allows JAWS/VO users to tab which reads just the link information and also have the capability to use arrow keys to reach eyebrow, heading, copy, etc. 

### Changelog

**New**

- Card using Default Tile and Carbon Link for footer
- added logic if there is no text in footer for aria label to use heading instead and if no heading then copy.
- Temporarily added web component styles for Card focus

**Changed**

- updated styles, so Cards and Feature Card behavior perform as before

**Removed**

- ariaProps in CTA


IgnacioBecerra (approved):
@ariellalgilmore LGTM, nice work! Tested in JAWS and both MacBook/iPhone VO, works as expected. Thank you!

Issue #5778, opened 2021-04-14
Custom Masthead searchbar tooltip text


Daynssjal:
### The problem
While testing the latest release candidate **v1.18.0-rc.0**, I noted there is a new **tooltip in the Masthead searchbar**, which is displayed when the user hovers over one of the searchbar icons, e.g the magnifier. The tooltip for the searchbar magnifier icon says: **"Search in all of IBM"**. This text is not applicable for us, since **we do not perform search in all of IBM.com,** but in our own application - IBM Documentation.

### The solution
Have a possibility to pass custom text into the tooltip.

### Additional information
https://ibmdotcom-react-staging.mybluemix.net/?path=/story/components-masthead--search-open-onload

![searchbar tooltip](https://user-images.githubusercontent.com/33804471/114721522-48805a00-9d39-11eb-91eb-70745c24bed4.png)

This request is submitted by the Knowledge Center team.

PR #5780, opened 2021-04-14
fix(MastheadSearch): remove tooltip from search button

annawen1:
### Related Ticket(s)

Custom Masthead searchbar tooltip text #5778

### Description

Hide the tooltip from the masthead search buttons

### Changelog

**Changed**

- set tooltip to `display: none`


jeffchew (approved):
Thank you @annawen1 !

Issue #5433, opened 2021-03-10
[Quote] Hanging quote and general alignment


oliviaflory:
#### User Story

The specs state that the starting quote should hang off the left of the column, and all other text should be aligned to the column. 
![Screen Shot 2021-03-10 at 1.17.47 PM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/8ed563a8-62dc-4aba-84c6-466b2566e1cb)

Currently, the Quote component does not hang the starting quotation mark on the left side of the column. The starting quote is aligned to the column, and all other text is inset within the column structure.
![Screen Shot 2021-03-10 at 1.20.03 PM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/fa8ba9e9-ef07-4203-b93f-be8fa878e0ed)

### Question:
Is it possible for the starting quote to hang to the left of the column? 

1. If yes, this can be a development story
2. If not, we may need to update the visual specs, or a combination of spec + dev updates to achieve the desired outcome.

#### Additional information

 - Discovered this during our clean up
 - Potentially affects Call out quote specs

#### Acceptance criteria
 - [ ] Dev & design collaboration to understand how to achieve desired design intent
 - [ ] Specs and/or story updated
 - [ ] Dev and design review

RaphaelAmadeu:
@oliviaflory 
I've opened a PR to address this issue.

shixiedesign:
Also noticed `Callout` component could need some update to maintain the changing quote.

PR #5782, opened 2021-04-14
fix(quote): added hang to component

RaphaelAmadeu:
### Related Ticket(s)

#5433 

### Description

I've updated the styles for `DDSQuote` so the component matches the design specs with the hang in the quote mark and line.

Before:
image

After:
image


shixiedesign:
Almost! Need to move to the right 1 more column: 

**Current quote start on column 5:** 
image

**Expecting quote to start on column 6:**
image

RaphaelAmadeu:
@shixiedesign I've updated the story, so the Quote component starts on column 6 now.

shixiedesign:
Also noticed `Callout` component could need some update to maintain the changing quote.

kennylam:
@RaphaelAmadeu Did you maybe create this PR off the wrong branch? There are a lot of offline tarballs included.

Issue #5510, opened 2021-03-16
Web Components:  Masthead search open on default and entering search keyword - overflow 


ariellalgilmore:
## Detailed description

The Masthead search open on default storybook and also when on default - entering a search term (ie. "red hat") triggers overflow.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

bug

> Is this issue related to a specific component?

masthead

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Should not see overflow arrows. Works correctly once you close the search then open again. 

## Additional information

- This issue should also be fixed with (#5464 )

<img width="1213" alt="Screen Shot 2021-03-16 at 11 33 36 AM" src="https://user-images.githubusercontent.com/20210594/111361917-75aeef00-864b-11eb-9862-aff27a9c535d.png">

PR #5790, opened 2021-04-15
fix(masthead): search triggering overflow

ariellalgilmore:
### Related Ticket(s)

#5510 
 
### Description

web component masthead was triggering overflow when search is open on default and when user typed phrases with a spacebar. 

### Changelog

**Changed**

- added conditional in top-nav checking if `event` is defined
- changed `hideNav` from private to public so `masthead-composite` can pass in rendered value of `activateSearch`


andysherman2121 (approved):
Nice work Ari!

shixiedesign (approved):
LGTM!

Issue #5423, opened 2021-03-09
Video component within Content block simple with video should be using a max width of 640px


kirbyevans:
## Detailed description

> According to the design specs the video component within the Content block simple with video should have a max-width of 640px. Right now on Storybook, it doesn't appear to have a max-width applied. 

## Additional information

- Content block simple with video not using a max-width: https://ibm.box.com/s/fjt3gl9b3qkyhhshin2z74az62bz795x
- Content block simple with image appears to be using a max-width of 640px: https://ibmdotcom-react.mybluemix.net/?path=/story/components-contentblocksimple--with-image
- Design specs state that video should have a max-width of 640px: https://ibm.box.com/s/fjt3gl9b3qkyhhshin2z74az62bz795x

RobertaJHahn:
dev - test in both react and WC to see where the problem is.

shixiedesign:
hey @RaphaelAmadeu Web component fixed, but problem still in React https://ibmdotcom-web-components-canary.mybluemix.net/iframe.html?id=components-content-block-simple--with-video

PR #5792, opened 2021-04-15
fix(content-block-simple): added max-width to media components

RaphaelAmadeu:
### Related Ticket(s)

#5423 

### Description

I've added max-width to the slotted media components on `ContentBlockSimple`.
This change affects React and Web Components.


shixiedesign (approved):
Looks good in the preview of component builds, but not seeing any Percy diffs. I'm guessing is Julio's PR for name change happened which made Percy think these are different components. Make sure to get a dev verification before merge.

jeffchew (approved):
LGTM!

Issue #5498, opened 2021-03-15
Web components [TOC] active section


oliviaflory:
## Detailed description

> Describe in detail the issue you're having.

### Web components 'active' section within the TOC should follow React behavior.

**React (desired)** 
- 2nd section does not become 'active' until title Praesent fermentum is near the top of the browser
- As user scrolls back up, when above section Cras molestie comes into view it activates 1st option

**Web components**
- 2nd section becomes 'active' as soon as title Praesent fermentum is in view, but 1st section body text is still taking up majority of screen
- As user srcolls back up, when 1st section is in view, 2nd title is still active until it scrolls out of view.

| React | Web components |
| --- | --- |
| ![react-scroll-selected-section-Mar-15-2021 15-04-29](https://user-images.githubusercontent.com/45692622/111207681-4db68180-85a0-11eb-86b9-6a7edb1ff6cf.gif)  |  ![webcomponent-scroll-selected-section-Mar-15-2021 15-05-51](https://user-images.githubusercontent.com/45692622/111207628-3f686580-85a0-11eb-8875-cf713d42638d.gif) |

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

TOC default and horizontal

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Should follow React behavior

> What browser are you working in?

Chrome

> What version of Carbon for IBM.com are you using?

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. Scroll through React TOC
2. Scroll through Web components TOC
3. Compare 'active' sections and when they change
4. etc.

## Additional information

- Screenshots or code
- Notes

PR #5794, opened 2021-04-15
fix(table-of-contents): fixed active section

IgnacioBecerra:
### Related Ticket(s)
#5498

### Description
The current Table of Contents changes the active section on the left the moment a new anchor tag appears in the viewport, even if it hasn't reached the top yet. 

| React | Web components |
| --- | --- |
| react-scroll-selected-section-Mar-15-2021 15-04-29  |  webcomponent-scroll-selected-section-Mar-15-2021 15-05-51 |

This PR changes this behavior so the current section is only set to "active" upon it reaches the top, and if a user scrolls down a bit further, the previous section will be the one becoming active, as shown in React.

The Web Component version of ToC uses Intersection Observer API for this functionality. However, after various attempts to modify the existing observer to apply this effect properly, it seems that there are times when the Intersection Observer API fails to detect the element coming into the viewport.

For example, one can be scrolling through Section 1, pass through Section 2 without 2 becoming active, and upon reaching Section 3 this is the one that becomes active, i.e. it ignores Section 2 completely. This happens randomly, and due to the spontaneous nature of this bug, it might be better to stick with using a Scroll Listener, like React is already doing.

### Changelog

**Changed**

- replaced Intersection Observer with Scroll Listener to align with React behavior


packages/web-components/src/components/table-of-contents/table-of-contents.ts
```diff
@@ -180,23 +174,18 @@ class DDSTableOfContents extends StableSelectorMixin(LitElement) {
   }
 
   /**
-   * Cleans-up and creats the inersection observer for the intersection of target `<a>`s with the viewport.
+   * Cleans-up and creates the Scroll Listener for the intersection of target <a>'s with the viewport.
```
kennylam, line 177: Since `IntersectionObserver` is no longer being used, suggest using [HostListenerMixin](https://github.com/carbon-design-system/carbon-web-components/blob/v1.11.0/src/globals/mixins/host-listener.ts)'s `@HostListener` decorator. This automates adding/removing event handlers.

Additionally, lodash's [`throttle()`](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/packages/web-components/src/components/back-to-top/back-to-top.ts#L113-L117) can be used to reduce polling time when scrolling occurs.

packages/web-components/src/components/table-of-contents/table-of-contents.ts
```diff
@@ -251,19 +240,34 @@ class DDSTableOfContents extends StableSelectorMixin(LitElement) {
     }
   }
 
+  /**
+   * Handles intersection of target `<a>`s with the viewport by checking which target's
+   * immediate siblings are close to the viewport, and set the active target depending
+   * on their positions.
+   */
+  private _handleScroll = () => {
+    window.requestAnimationFrame(() => {
```
kennylam, line 249:
```suggestion
    this.ownerDocument!.defaultView!.requestAnimationFrame(() => {
```

Suggest component DOM instead of global `window` whenever possible.

packages/web-components/src/components/table-of-contents/table-of-contents.ts
```diff
@@ -445,13 +427,13 @@ class DDSTableOfContents extends StableSelectorMixin(LitElement) {
   connectedCallback() {
     super.connectedCallback();
     this._cleanAndCreateObserverResizeMobileContainer({ create: true });
-    this._cleanAndCreateObserverIntersectionTarget({ create: true });
+    this._cleanAndCreateScrollListener({ create: true });
```
kennylam, line 430: Same as previous `lodash` comment, can use [`throttle`](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/packages/web-components/src/components/back-to-top/back-to-top.ts#L119-L125) here.

kennylam (commented):
Looking good @IgnacioBecerra! Just a few comments.

kennylam:
@IgnacioBecerra @oliviaflory In both versions, depending on the length of the final section there is a chance that the last item of the left nav will not be highlighted since the scrolling conditions don't occur. Should the last item be highlighted if the end of the page is reached?

image

oliviaflory:
> In both versions, depending on the length of the final section there is a chance that the last item of the left nav will not be highlighted since the scrolling conditions don't occur. Should the last item be highlighted if the end of the page is reached?

@kennylam good question... 

I would suggest yes, so far we haven't seen too many very short sections, but I'd expect to still show the last item highlighted/active at the end of the page. 

There is also no guarantee that a page will always use the default footer which would add more height to the page.

kennylam (approved):
LGTM thanks @IgnacioBecerra!

oliviaflory:
Looks great! 

@IgnacioBecerra I just want to double check, it looks like the colors are different in Locale modal story are different, is this a fluke in the snapshot?
Screen Shot 2021-04-20 at 4 12 15 PM

IgnacioBecerra:
@oliviaflory Yeah I believe so, this PR doesn't modify any styles

oliviaflory (approved):
active sections look good, thanks Ignacio!

Issue #5795, opened 2021-04-16
[Card] Empty eyebrow text should not leave space above card headline


shixiedesign:
Web components canary, not in React:
https://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-card--default

![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/1df5a97a-dcec-4764-ab6f-a0abf7399ad1)

PR #5798, opened 2021-04-16
fix(card): add eyebrow to card pictogram story

kennylam:
### Related Ticket(s)

#5795

### Description

Adds missing `eyebrow` copy to default web component `Card` story.

### Changelog

**New**

- `eyebrow` copy added to default story.


shixiedesign:
Hey Kenny! Thanks for the fix. I'm afraid this won't fix the situ when eyebrow text is absent though. Can we restore the knobs so I can test the cases where some optional texts are absent? Left side is Web components, right side React. React card's knobs are expected. 

We recently cleaned up the pictogram cards in Web component (`Card : with pictogram`) and I asked for some knobs to be removed. Maybe the changes were accidentally applied here in the default card as well. 

image

kennylam:
@shixiedesign There is no knob for eyebrow so if you remove it via the inspector, or even a Storybook knob, the `eyebrow` custom element will remain. This will create the extra space we don't want. One thing we could do is have a checkbox which shows/hides the eyebrow – that would remove the element entirely if it is unchecked.

shixiedesign (approved):
Kewl. For now let's do this then. Thank you! When we get to Card's storybook clean up we can potentially add a knob.

Issue #5508, opened 2021-03-16
Text line length change to `# column - 32px`


shixiedesign:
### Text width changing:
We'd like to change how line lengths are calculated in containers
**Current**: 90% of container
**Change to**: 100% of container - 32px ($layout-03)

Note: 
1. **Mobile text width is still 100%**
2. Max width 640px is also unchanged

### This change needs to happen here: 

- [ ] content item
- [ ] content group
- [ ] content block
- [ ] content section 
- [ ] card
- [ ] quote

#### Example, see green/teal colored padding: 

![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/1528a290-1066-4ab3-9c42-ff49c203f65a)

PR #5799, opened 2021-04-16
fix: text line length change

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5508

### Description

Changed `content-width` mixin to 100% of container - 32px ($layout-03). Added same width calculation to quote (which doesn't use the same mixin because of the width limitation). 

- [x] content item
- [x] content group
- [x] content block
- [x] content section 
- [x] card
- [x] quote


NSDrowned:
@shixiedesign let me know if this is what you asked for since modifying this mixin changes a lot of components. If not, I'll add overrides just for the components mentioned in the original issue and leave the mixin intact.

shixiedesign:
The text line length change should not affect `sm`. On `sm` breakpoint, we are still expecting 100% container width for text width: 
image

NSDrowned:
@shixiedesign, you're right, let me fix that.

shixiedesign:
Noticing a small thing - the rule inside Quote seems to have gotten shorter. Don't think this is expected. 

image

kennylam (approved):
Looks good thanks @NSDrowned!

shixiedesign (approved):
Thank you! Looking good 🌮

Issue #5819, opened 2021-04-19
[Masthead - React]: search icon closes search field when input is empty


kennylam:
In the React version of the `Masthead`, clicking the search icon closes the entire search component when the input field is empty. The expected behavior is for the search component to remain open (**not** perform a search) and basically do nothing.

![Apr-19-2021 10-19-51](https://user-images.githubusercontent.com/909118/115251873-f48ac200-a0f8-11eb-93c5-8ac38cda1c45.gif)

cc: @Praveen-sr @RobertaJHahn

PR #5820, opened 2021-04-19
fix(masthead): search icon does nothing when input empty

kennylam:
### Related Ticket(s)

#5819 

### Description

Fixes an issue in the `default` React masthead where clicking the search icon while the search component is open with no search value would close the entire component. The expected behavior is for the search component to remain open and do nothing.

### Changelog

**Changed**

- search visibility behavior in `MastheadSearch`

### DIFF
--- a/packages/react/src/components/Masthead/MastheadSearch.js
+++ b/packages/react/src/components/Masthead/MastheadSearch.js
@@ -312,6 +312,9 @@ const MastheadSearch = ({
       });
 
       event.currentTarget.dispatchEvent(onOpenSearch);
+
+      dispatch({ type: 'setSearchOpen' });
+      handleChangeSearchActive();
     }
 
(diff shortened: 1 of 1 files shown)


shixiedesign (approved):
Looks fixed in my testing in Chrome! Thanks Kenny ⚡

shixiedesign:
Percy found this and seems like a it's probably not a real diff. React doesn't have this bug so shouldn't be affected. 

image

jeffchew:
> Percy found this and seems like a it's probably not a real diff. Prod React is behaving as expected. React doesn't have this bug so shouldn't be affected.
> 
> image

Yes, it's a false positive that's sporadic, haven't found a fix for this yet.

jeffchew (approved):
LGTM!

Issue #5789, opened 2021-04-15
NextJS page: Masthead Search bar - On hover search icon and close icon are hidden


Praveen-sr:
## Detailed description

Issue: 
On Search bar - On hover search icon and close icon are hidden.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/8d451293-1931-485a-8797-1ce6e24e9429)

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/223d3ebb-c2f0-4013-9679-86866383ff0b)

RobertaJHahn:
Slack post from Ignacio:  I took a quick look and it seems the carbon-component styles are being loaded later, so they override our previous fixes that were done for the upgrade. It may have to do with chunking, or the styles changes may possibly need to have a more specific selector overall. Not sure if it’s NextJS specific or it may affect other sites.

RobertaJHahn:
reopened because tooltip fix is not included in the merged PR.

PR #5821, opened 2021-04-19
fix(Masthead): use specific selectors to remove tooltip

annawen1:
### Related Ticket(s)

NextJS page: Masthead Search bar - On hover search icon and close icon are hidden #5789

### Description

Make sure that the tooltip for the search icons are not displaying on hover and click of the icons. Need to use more specific selectors for the NextJS test pages.

### Changelog

**Changed**

- use more specific selectors


jeffchew (approved):
LGTM!

PR #5830, opened 2021-04-20
fix(tabs-extended-with-media): videos are now working

RaphaelAmadeu:
### Description

I've added a quick fix to our video player service so videos work properly on Tabs extended with media.

### DIFF
--- a/packages/services/src/services/VideoPlayer/VideoPlayer.js
+++ b/packages/services/src/services/VideoPlayer/VideoPlayer.js
@@ -193,6 +193,17 @@ class VideoPlayerAPI {
     const fireEvent = this.fireEvent;
     return await this.checkScript().then(() => {
       const promiseKWidget = new Promise(resolve => {
+        let isCustomCreated;
+
+        if (!document.getElementById(targetId)) {
+          const newVideoDiv = document.createElement('div');
+          newVideoDiv.classList.add(`bx--video-player__video`);
+          newVideoDiv.setAttribute('id', targetId);
+          document.body.append(newVideoDiv);
+          console.log(document.getElementById(targetId));
+          isCustomCreated = true;
+        }
+
         root.kWidget.embed({
           targetId: targetId,
           wid: '_' + _partnerId,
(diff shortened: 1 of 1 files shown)


packages/services/src/services/VideoPlayer/VideoPlayer.js
```diff
@@ -193,6 +193,17 @@ class VideoPlayerAPI {
...
     return await this.checkScript().then(() => {
       const promiseKWidget = new Promise(resolve => {
+        let isCustomCreated;
+
+        if (!document.getElementById(targetId)) {
+          const newVideoDiv = document.createElement('div');
+          newVideoDiv.classList.add(`bx--video-player__video`);
+          newVideoDiv.setAttribute('id', targetId);
+          document.body.append(newVideoDiv);
+          console.log(document.getElementById(targetId));
```
jeffchew, line 203: stray console.log here @RaphaelAmadeu

packages/services/src/services/VideoPlayer/VideoPlayer.js
```diff
@@ -193,6 +193,17 @@ class VideoPlayerAPI {
     const fireEvent = this.fireEvent;
     return await this.checkScript().then(() => {
       const promiseKWidget = new Promise(resolve => {
+        let isCustomCreated;
+
+        if (!document.getElementById(targetId)) {
```
jeffchew, line 198: @RaphaelAmadeu I'm wondering if we should check if `!document.getElementById(targetId)` as well as if `document.querySelector('dds-tabs-extended-media')` exists? That way in any weird cases where the target ID doesn't exist, it won't just append a video to the end of the page and not do anything else after that (hopefully that's an edge case but still...)

RaphaelAmadeu: yeah, I think it would be good! I'm adding this in my new commit

packages/services/src/services/VideoPlayer/VideoPlayer.js
```diff
@@ -226,6 +237,17 @@ class VideoPlayerAPI {
             resolve(kdp);
           },
         });
+
+        if (isCustomCreated) {
+          const previousVideoDiv = document
```
jeffchew, line 244: At this point, does `document.querySelector('dds-tabs-extended-media').shadowRoot.querySelector(targetId)` exist yet? Otherwise if this works we can stick with this for now since this is only a temporary measure.

packages/services/src/services/VideoPlayer/VideoPlayer.js
jeffchew: Thanks @RaphaelAmadeu , looks great!

jeffchew (approved):
Slick fix, thank you @RaphaelAmadeu !

Issue #5541, opened 2021-03-18
[CTA] Card - should show card link


RobertaJHahn:
#### Overview
The CTA components inside the [React](https://ibmdotcom-react.mybluemix.net/?path=/story/components-cta--button) and [Web Components](http://ibmdotcom-web-components.mybluemix.net/?path=/story/components-cta--text) Storybooks have discrepancies around the naming and the examples between the two Storybooks (see screenshots below). 

After a [discussion](https://cognitive-app.slack.com/archives/GBG66T9NK/p1614887714098800) on the team Slack channel, the React version is wrong based on James's functional spec.

#### Additional Information
- [ ] Card - should show card link. (React: should be left aligned in viewport. WC: missing corner icon).

`For the examples:`
![Storybook samples](https://user-images.githubusercontent.com/45213839/110166697-56e26a00-7dba-11eb-825e-5da90856b283.jpg)

#### Acceptance Criteria
 - [ ] Both Storybooks show the same labels and examples for `CTA - card` component

PR #5831, opened 2021-04-20
fix(cta): card link and other story fixes

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5541

### Description

- React: Aligned to the left and added same padding as web components
- Web Component: Added missing corner icon


packages/react/src/components/CTA/__stories__/cta.stories.scss
```diff
@@ -0,0 +1,11 @@
...
+ * Copyright IBM Corp. 2016, 2021
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+.bx--card {
+  margin-top: 1rem;
+  margin-left: -1rem;
+}
```
kennylam, line 11: I don't think these are needed? Removing them makes the story layout identical to the web components version.

kennylam (commented):
Thanks for this @NSDrowned! Just a small observation

packages/react/src/components/CTA/__stories__/cta.stories.scss
NSDrowned: You're right 🤔 Weird... my fork was bit behind when I worked on it. Perhaps that's why I added them cause I see no reason now lol. Thanks!

shixiedesign (approved):
Visually unchanged. Goal is to make sure`CTA - card` uses `Card link`. Please get additional dev approval!

PR #5832, opened 2021-04-20
fix(lerna): increasing max buffer for lerna ci-check

jeffchew:
### Related Ticket(s)

No related issue

### Description

Github actions is running into memory buffer issues, increasing the max
buffer in hopes that this will fix things.

### Changelog

**Changed**

- ci-check.yml


jeffchew:
This is possibly being caused by something else, closing this PR.

Issue #5484, opened 2021-03-15
Masthead with Platform - Divider activates hover state on platform


matoushavlena:
## Detailed description

In Masthead with Platform, when you mouse hover the divider between the IBM logo and platform, the hover state gets activated on the platform.

## Steps to reproduce the issue

1. Open Masthead with Platform
2. Move your mouse a few pixels right next to the divider between the logo and the platform.

## Additional information

https://user-images.githubusercontent.com/9358330/111163258-42158b80-859d-11eb-9176-c306f34ab54e.mov

PR #5835, opened 2021-04-20
fix(Masthead): remove pseudo element from platform name

annawen1:
### Related Ticket(s)

Masthead with Platform - Divider activates hover state on platform #5484

### Description

Space between the platform name and divider is clickable - removing the anchor link pseudo element resolves this.

BEFORE:
BEFORE

AFTER:
AFTER

### Changelog

**Removed**

- remove anchor link `after` pseudo element

### DIFF
--- a/packages/styles/scss/components/masthead/_masthead-search.scss
+++ b/packages/styles/scss/components/masthead/_masthead-search.scss
@@ -45,17 +45,6 @@
     background-color: #dcdcdc;
   }
 
-  &::after {
-    content: '';
-    display: block;
-    position: absolute;
-    width: 15px;
-    height: $spacing-09;
-    left: calc(-1rem - 1px);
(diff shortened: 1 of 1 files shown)


jeffchew (approved):
LGTM!

ariellalgilmore:
Screen Shot 2021-04-22 at 10 34 16 AM

removing the psuedo element brings back the overflow :'(  visibility. Wondering if we can keep the psuedo element and add `pointer-events: none;` ?

annawen1:
> Screen Shot 2021-04-22 at 10 34 16 AM
> 
> removing the psuedo element brings back the overflow :'( visibility. Wondering if we can keep the psuedo element and add `pointer-events: none;` ?

oh damn! good catch @ariellalgilmore I'll push a fix for that

Issue #5732, opened 2021-04-08
[Content group cards] Storybook - Web Comp - card focus outline incorrect


DragosRistici:
The Web Comp Storybook card focus outline is not correct:
![image](https://user-images.githubusercontent.com/48327647/114030655-49ba0e80-9883-11eb-93df-f17b4283fad3.png)

IgnacioBecerra:
@DragosRistici Seems like this has been fixed already, would you be able to confirm?

DragosRistici:
@IgnacioBecerra There's still a double outline for the cards in the content group cards. Check the others, like Card section simple:


This one is Card section simple:
![Screenshot 2021-04-20 at 19.30.59.png](https://images.zenhubusercontent.com/6024dfe2f687e36aef6881c5/326fa60c-fc97-40fd-83c8-3be442ff91bf)

This one is Content group cards:
![Screenshot 2021-04-20 at 19.30.45.png](https://images.zenhubusercontent.com/6024dfe2f687e36aef6881c5/02bdee80-186f-4df5-b703-d8008b8ed8b9)

I'm checking here: http://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-content-group-cards--default

Thanks, 
Dragos

PR #5840, opened 2021-04-20
fix(content-group-cards): fixed bold outline

IgnacioBecerra:
### Related Ticket(s)
#5732

### Description
Removed additional outline that was present when focusing on the `ContentGroupCardItem`

### Changelog

**Changed**

- `outline: none` when focused on `<dds-content-group-card-item>`


ariellalgilmore (approved):
looks good!

Issue #5691, opened 2021-03-31
[Performance] Investigate long build times in web components package


jeffchew:
#### User Story

> As a `[user role below]`:
Carbon for IBM.com developer

> I need to:
build the web components storybook locally and quickly

> so that I can:
be efficient with my development work on the web components package

#### Additional information

 - Current builds are a bit long, this is a story to investigate what is happening and if there are ways to improve the performance of the build times

#### Acceptance criteria
 - [ ] Find the reason to why build times are long for the web components package
 - [ ] Either address the issue (if it is a quick fix) or create followup issues on how to address

jeffchew:
@RobertaJHahn we can roll this in for the sprint where we are looking at backlog items. We still need to size this.

jeffchew:
Initial look at doing a `yarn build` for `web-components`, and looks like building css is the heaviest in the build:

![image](https://user-images.githubusercontent.com/1641214/115464219-be356b80-a1fa-11eb-9771-a473ef817f11.png)

PR #5845, opened 2021-04-20
chore(github-actions): adjusting scope of yarn builds

jeffchew:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5691

### Description

This is an attempt to reduce the build times around various jobs in
Github Actions, particularly reducing scope of the `yarn build` and
identifying specific packages that do not need to be built, rather than
building the entire project.

### Changelog

**Changed**

- github actions updates


IgnacioBecerra (approved):
LGTM, Thanks @jeffchew!

Issue #5850, opened 2021-04-21
Web component: Masthead - Search bar - On click,  Keywords does not get activated 


Praveen-sr:
## Detailed description

Issue:
Masthead - Search bar - On click,  Keywords does not get activated. Search bar closes, user has to re open the search field.

Note: 
1. When user uses enter key, it works
2. Working fine on react version.

## Additional information

![Search_key.gif](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/1d2d9be1-6c33-414d-bedc-24be303e56b7)

PR #5851, opened 2021-04-21
fix(masthead-search): ensure clicking on suggestion works

IgnacioBecerra:
### Related Ticket(s)
#5850

### Description
This PR ensures that clicking on any of the search suggestion items doesn't close the search bar and goes directly to the search results page as intended.

### Changelog

**Changed**

- quick check to ensure the blur event doesn't occur if clicking a masthead search result


ariellalgilmore (approved):
LGTM! thanks for the quick fix!

PR #5861, opened 2021-04-21
fix(masthead): ensure only search bar displays on load upon search open

IgnacioBecerra:
### Related Ticket(s)
Mentioned in the QA meeting this morning, couldn't find an open issue for this

### Description
The Masthead Menu button and IBM logo still showed up on load on the Masthead - Search Open by Default story. The previous fix which ensures toggling the search bar still works only, though only after closing it once.This fix ensures these two components hide upon load. 

### Changelog

**New**

- hide booleans for `<masthead-logo>` and `<masthead-menu-button>` to ensure they are hidden upon load the same way `<masthead-top-nav>` does it


jeffchew (approved):
LGTM!

Issue #5867, opened 2021-04-22
Issues upgrading to Carbon for IBM.com v1.18.0


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

From a couple of adopters:

```We are getting build issue w.r.t ibmdotcom-styles package.```

![image](https://user-images.githubusercontent.com/1641214/115738751-55193980-a35b-11eb-9984-79262a98bbb0.png)

```anyone else getting this issue when trying to upgrade to v1.18.0?```

![image](https://user-images.githubusercontent.com/1641214/115738839-6d895400-a35b-11eb-9417-7d0c82175073.png)

```
warning "@carbon/ibmdotcom-react@1.18.0" has incorrect peer dependency "carbon-components@10.32.1".
warning "@carbon/ibmdotcom-web-components > carbon-web-components@1.12.0" has incorrect peer dependency "carbon-components@~10.30.0".
```

```
➜  think-2021-nextjs git:(master) ✗ yarn install
Couldn't find any versions for "carbon-components-react" that matches "^7.33.1"
? Please choose a version of "carbon-components-react" from this list: (Use arrow keys)
❯ 7.33.0
  7.33.0-rc.0
```

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

Application install level

> What did you expect to happen? What happened instead? What would you like to
> see changed?

No issues upgrading to latest `Carbon for IBM.com`

> What version of Carbon for IBM.com are you using?

v1.18.0

RobertaJHahn:
@jeffchew Can this issue be closed?  There is a merged PR.

jeffchew:
@RobertaJHahn yes. And why are you working on a Sunday?

PR #5871, opened 2021-04-22
fix(dependencies): changed @carbon/telemetry to be a dependency

jeffchew:
### Related Ticket(s)

#5867 

### Description

Adopters were providing feedback that the `carbon-telemetry` command was
 not found, which is due to the case that @carbon/telemetry was only a
 dev dependency. This change moves it to the dependencies list.

 In addition, the upstream Carbon dependencies that were originally
 identified as peer dependencies are now optional dependencies.

### Changelog

**Changed**

- `@carbon/telemetry` now a dependency for public packages
- Moved upstream Carbon peer dependencies to optional dependencies


andysherman2121 (approved):
Nice work Jeff

PR #5886, opened 2021-04-23
fix(callout-quote): added hang

RaphaelAmadeu:
### Related Ticket(s)

#5433 

### Description

Hang was added to `Quote` and `CalloutQuote`, so it matches the designs specs

### DIFF
--- a/packages/styles/scss/components/callout-quote/index.scss
+++ b/packages/styles/scss/components/callout-quote/index.scss
@@ -39,6 +39,11 @@
       padding-left: $spacing-07;
       outline: transparent;
     }
+
+    .bx--quote__container,
+    .bx--quote__wrapper {
+      @include hang;
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


shixiedesign:
Behavior between `max` and `sm` breakpoint looks good! But once above `1584px`, component stops following grid: 

Apr-25-2021 15-38-14

oliviaflory:
@RaphaelAmadeu This is looking really good, just have a few questions / comments

1. the rule is hanging more than it should from the design specs, should be just 16px hang on either side, but looks like it might be 32px. Seeing in Quote and in Call out quote.



2. Small breakpoint has some differences between react and web components, I attached a screen shot for reference
   a. the container for the text is different Web components 256px vs React 272px, I noticed because the left padding looks different between the two stories
   b. the grid seems different (blue dotted line), React doesn't have as much margin on the right side 

Screen Shot 2021-05-06 at 1 20 32 PM 


Let us know if this cannot be fixed by code freeze tomorrow, I think we could merge this and open new issues to fix the things above.

oliviaflory (approved):
Looks good, thanks @RaphaelAmadeu for all the adjustments! 🎉

guilhermelMoraes (approved):
LGTM! 👍🏻

Issue #5890, opened 2021-04-23
Masthead and Footer components not supporting locales like en-cn or en-gb


mprab80:
Im seeing an issue with Masthead component (https://ibmdotcom-web-components.mybluemix.net/?path=/docs/components-masthead--with-l-1) - It renders the header/navigation on en-us and fr-ca page, but I see that its not happening on en-gb page. 

## Detailed description

> Describe in detail the issue you're having.
Anna says "we have a service to grab the locale from the html lang attribute and we verify the locale by making a call to get the countrylist. If a countrylist does not exist for the locale we use our default (us-en) to get the countrylist, but we never pass the default locale to the translation api call which is why it still calls gb-en, which does not exist, and returns empty. "
> Is this a feature request (new component, new icon), a bug, or a general issue?
Its a bug
> Is this issue related to a specific component?
Its for translations. 
> What did you expect to happen? What happened instead? What would you like to
> see changed?
I expect header and footer to be populated with default locale i.e., en-us but that is not happening. 
> What browser are you working in?
Chrome
> What version of Carbon for IBM.com are you using?
latest
> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?
No
## Steps to reproduce the issue

1. Compare these 2 pages: 
2.https://dev-publish.np-roks.cms.cis.ibm.net/gb-en/test-learn-page
https://dev-publish.np-roks.cms.cis.ibm.net/uk-en/test-learn-page 

## Additional information

- Screenshots or code
![image](https://user-images.githubusercontent.com/55397410/115927198-0e5a3b00-a452-11eb-8f22-0d5f1686a1ba.png)
![image](https://user-images.githubusercontent.com/55397410/115927220-15814900-a452-11eb-98f2-e0802224844a.png)

- Notes

PR #5907, opened 2021-04-27
fix(locale): use getLocale method to verify locale before passing into translation call

annawen1:
### Related Ticket(s)

Masthead and Footer components not supporting locales like en-cn or en-gb #5890

### Description

In web-components masthead, the masthead returns blank when using a non-existent locale (ie. cn-en).

For web-components, we are fetching the locale using `LocaleAPI.getLang()` which doesn't verify the locale and passing the unverified locale directly to the translation call as a parameter. When passing the locale as a parameter, there are no checks are the translation call is made with the unverified locale. 

Solution is to use `LocaleAPI.getLocale()` instead which verifies the locale before returning it.

### Changelog

**Changed**

- For `loadLanguage()` redux action in services-store, use `getLocale()` instead to make sure the locale retrieved from DDO is valid.

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -41,6 +41,16 @@ jobs:
           cf-app: ibmdotcom-react-staging
           cf-manifest: manifest-staging.yml
           deploy-dir: packages/react
+      - name: Set env vars for e2e percy snapshots
+        uses: ./.github/actions/set-dotenv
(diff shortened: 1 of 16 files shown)


IgnacioBecerra (approved):
Nice work, @annawen1!

ariellalgilmore (approved):
looks good!! thanks @annawen1!

Issue #5815, opened 2021-04-19
Web component: Tabs extended - The Fifth tab is disabled. But navigable through keyboard and operable as well. 


Praveen-sr:
## Detailed description

Issue:
1. The Fifth tab is disabled. But navigable through keyboard and operable as well. 
2. This tab looks disabled visually. But for screen reader users the disabled status is not conveyed. 

Expected:
1. The disabled tab should not be Keyboard operable and should not receive tab stop
2. Convey the disabled status for the screen reader using HTML disabled attribute

Issue #5814, opened 2021-04-19
Web component: Tabs extended - with media - selected attribute is announced for all the tabs


Praveen-sr:
## Detailed description

On JAWS FF
Issue:
Selected attribute is announced for all the tabs. 
Note: Same issue is present in "tabs extended as well" 

Expected:
Screen reader should announce the selected attribute only for the tab which is currently selected. Use aria-selected ="true" only for the currently selected tab.

PR #5917, opened 2021-04-27
fix(tabs-extended): update aria-selected label and disabled tab

ariellalgilmore:
### Related Ticket(s)

#5814 #5815

### Description

Use `active` state to set `aria-selected` label. Adjust `tabindex` for `disabled` state


IgnacioBecerra (approved):

Issue #5863, opened 2021-04-22
[Content blocks and groups] - React Storybook - 


DragosRistici:
The ordered list in all the examples in React Storybook (every block and group page that has that piece of content) has an overhang in the content from this css:
<img width="578" alt="Screenshot 2021-04-22 at 14 18 30" src="https://user-images.githubusercontent.com/48327647/115706528-6de31880-a376-11eb-9b82-c0b669a140cc.png">

What it looks like 
<img width="468" alt="Screenshot 2021-04-22 at 14 13 45" src="https://user-images.githubusercontent.com/48327647/115706708-9e2ab700-a376-11eb-8bbc-cfe4131c6f67.png">

The issue is minor, but it is present on multiple pages of the React Storybook. 

The issue was opened separately as per @IgnacioBecerra  request.

PR #5919, opened 2021-04-27
fix: list margin fix

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5863

### Description

Carbon list had a bug in which ordered list items required an extra left margin to align correctly. This bug has fixed now and the extra margin is no longer needed.

**Removed**

- Removed extra margin from carbon expressive theme
- Removed `_list.scss` file from theme since it's no longer needed


DragosRistici:
Thank you, Julio!! This looks awesome! One little thing, if you could check why the React Wrapper has that grid misalignment for the lead space block on the page:
Screenshot 2021-04-28 at 11 53 29
Screenshot 2021-04-28 at 11 53 40

Thank you so much. If you think we should open a new issue for this, let me know. Anyway, the lists are looking G R E A T !

DragosRistici (approved):

kennylam (approved):
LGTM thanks @NSDrowned!

@DragosRistici the `LeadspaceBlock` grid alignment issue is addressed in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/5921

Issue #5560, opened 2021-03-21
Storybook refinement: use proper Carbon theme names 


shixiedesign:
Theme names should not be abbreviated. These abbreviations are not standard and could cause confusion. Reference[ Carbon website](https://carbondesignsystem.com/components/accordion/usage)

**Current**
![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/c002798e-d26b-475f-96c7-56f67eb42326)

**Expecting** in exact order, spelling, and capitalization
```
Carbon theme
|--- White 
|--- Gray 10
|--- Gray 90
|--- Gray 100

PR #5923, opened 2021-04-28
fix(storybook): update theme names

ariellalgilmore:
### Related Ticket(s)

#5560 

### Description

Use proper theme names. For example `g10` -> `Gray 10` in storybook carbon theme knobs


annawen1 (approved):
Nice! LGTM!

Issue #5623, opened 2021-03-25
[Card group with CTA] inverse hover state incorrect 


oliviaflory:
## Detailed description

> Describe in detail the issue you're having.

Hover state is wrong, and likely not accessible (white text on light gray)
- should be `$inverse-hover-ui`

![card-group-ctaMar-25-2021 14-54-12](https://user-images.githubusercontent.com/45692622/112528406-7d247580-8d7a-11eb-8721-ba7a396ec5ca.gif)

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

> Is this issue related to a specific component?

> What did you expect to happen? What happened instead? What would you like to
> see changed?

> What browser are you working in?

> What version of Carbon for IBM.com are you using?

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information
https://ibmdotcom-web-components.mybluemix.net/?path=/story/components-card-group--with-cta
https://ibmdotcom-web-components.mybluemix.net/?path=/story/components-card-group--with-images-and-cta

PR #5935, opened 2021-04-29
fix (card): updating hover token for Card with the inverse theme

guilhermelMoraes:
### Related Ticket(s)

CTA inverse card in CardSectionSimple component: heading color does not change on hover#5663
[Card group with CTA] inverse hover state incorrect #5623


IgnacioBecerra (approved):
LGTM, thanks @guilhermelMoraes!

Issue #5913, opened 2021-04-27
TOC Content Div is set to 75% and doesn't allow content to span appropriate columns


kirbyevans:
## Detailed description

> Describe in detail the issue you're having.
**There is extra space (4 columns) on the right side of our page template. The TOC content div has max-width and flex-basis set to 75%. Should this be fixed since shadow DOM styling can't be overridden?**

> Is this a feature request (new component, new icon), a bug, or a general
> issue?
**Bug**

> Is this issue related to a specific component?
**TOC Content DIV**

> What did you expect to happen? What happened instead? What would you like to
> see changed?
**I expected our page content to not have an extra 4 columns of space on the right side.**

> What browser are you working in?
**Chrome**

> What version of Carbon for IBM.com are you using?
**On AEM v.17.0 and on storybook v1.18.0**

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?
**Afterburner / AEM**

## Additional information
– Jira ticket on Afterburner CMS: https://jsw.ibm.com/browse/ADCMS-1068
– Please reach out to Raveena Rastogi for further details.

![SolutionPage](https://user-images.githubusercontent.com/77290329/116268144-08fb3a00-a74b-11eb-87b6-2c180895d29a.png)

![Screen Shot 2021-04-27 at 7 31 21 AM (1)](https://user-images.githubusercontent.com/77290329/116268240-23351800-a74b-11eb-9839-3a5ba1eb8378.png)

RobertaJHahn:
To be investigated as part of Fireline.

PR #5951, opened 2021-05-04
fix(toc-horizontal): removed max-width and flex styles for content

RaphaelAmadeu:
### Related Ticket(s)

#5913

### Description

Removed `max-width` and added `flex: 1` to the Horizontal Top content, so it fills the available width.

image


kennylam (approved):
LGTM thanks @RaphaelAmadeu! Looks like you just need to update the snapshots.

jeffchew:
@RaphaelAmadeu just checking percy, this is supposed to be this large?

image

And this looks like pre-existing, we may need to look at what's going on here with the theming here:

image

jeffchew:
Also @RaphaelAmadeu checking the web components percy snapshots, both the regular and horizontal are bumped down, was this intended:

image

image

Issue #5942, opened 2021-05-03
Masthead search typeahead - multisection - section label not showing when there is just 1 result


Daynssjal:
Multiple search result sections issue:

**Detailed description**

 We are using customTypeAheadApi and multiSection property in the Masthead search. 
The issue occurs in the search results dropdown:
If there is just 1 result in the second section, the section label does not show in the result dropdown. It then looks as if that 1 result was a part of the previous section. Please see attached images (incorrect, correct)

> **Is this issue related to a specific component?**
Masthead

**What did you expect to happen? What happened instead? What would you like to
see changed?**
**We would like the section label to be visible even when there is only 1 result.**

> What browser are you working in?
FF, Chrome, Safari, Edge

> What version of Carbon for IBM.com are you using?
1.17.0, 1.18.1

> **What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?**
**Knowledge Center team, IBM Docs project**

## Steps to reproduce the issue

1. Go to https://www.ibm.com/docs/en
2. Open Masthead search
3. Type in "spss statistics"
4. See the results. The last (7th) result - "SPSS Statistics" is actually the 1st result of a new section. The label for that section is missing

## Additional information

![search-results-INCORRECT](https://user-images.githubusercontent.com/33804471/116847878-c2df2400-abeb-11eb-83f4-8e1752534ef5.png)

<img width="734" alt="search-results-CORRECT" src="https://user-images.githubusercontent.com/33804471/116847351-a8587b00-abea-11eb-8868-6ef155684daa.PNG">

PR #5956, opened 2021-05-04
fix(masthead): multi-section search headline

kennylam:
### Related Ticket(s)

#5942 

### Description

Fixes an issue where search result section title would not show unless more than one item is in the section. The title should show when there is one or more items in the section.

### Changelog

**Changed**

- updated section title to show when one or more results are present


ariellalgilmore (approved):
LGTM!

Issue #5847, opened 2021-04-21
Web component: Dotcom shell Footer - From the second modal, when we activate "Select geographic area" link, the First modal is not displayed. 


Praveen-sr:
## Detailed description

On JAWS FF
Issue:
From the second modal, when we activate "Select geographic area" link, the First modal is not displayed. 
Instead the screen reader focus goes to the parent page.

Expected:
Screen reader should identify the first modal.

kennylam:
@Praveen-sr @RobertaJHahn Issue is fixed in #5964.

Praveen-sr:
@kennylam Re tested this, The original issue is fixed. 
But, now jaws does not identify the dialog role for both the dialogs hence the user would not know that this is a modal dialog.

PR #5964, opened 2021-05-05
fix(locale-modal): re-focus on region item when navigating within modal

kennylam:
### Related Ticket(s)

#5847

### Description

This fixes a bug where the current `locale-modal` does not re-focus on the first `region-item` button when a user navigates back to the first pane from the country filter.

https://user-images.githubusercontent.com/909118/117159277-23a66200-ad8e-11eb-82e5-acb18af403d5.mp4

### Changelog

**New**

- add focus state to first geographic `region-item` when navigating back to the first modal pane from search pane


ariellalgilmore (approved):
Looks good! Thanks @kennylam!

PR #5966, opened 2021-05-05
fix(snapshot): fixing React snapshot file

jeffchew:
### Related Ticket(s)

No related issue

### Description

This fixes the out-of-sync React snapshot file which is causing preflight checks to fail. Not sure how this slipped previously.

### Changelog

**Changed**

- React snapshot file


kennylam (approved):
LGTM thanks @jeffchew!

PR #5970, opened 2021-05-05
feat(masthead): search-open-onload-wc

kennylam:
### Related Ticket(s)

#5232 
#5233 

### Description

This adds two features to web components `masthead`:

1. `search-open-on-load` opens the masthead search on page load. Just like the React version, the only way to close the search is by clicking the close button – `blur` does nothing.
2. If `search-open-on-load` is true and the query parameter `q` is present, the search input will take that value, i.e. if a user searches for `cloud` and the search results page is using `search-open-on-load`, the search input will have the value `cloud` already populated.

[Here](https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/5970/iframe.html?id=components-masthead--search-open-onload&q=cloud) is an example of both features in action.

### Changelog

**New**

- `search-open-on-load` attribute opens search on page load. Can only be closed via close button.
- search input value will auto-populate if query param `q` is detected and `search-open-on-load` is true

**Changed**

- update docs


IgnacioBecerra:
Changes look great! One question though: upon closing and reopening the search bar, should the search input be repopulated with the query parameter? The search input currently gets cleared upon reopen as usual, but I'm not sure if this should be different when using `search-open-on-load`.

kennylam:
@IgnacioBecerra No, once the search is manually closed the pre-populated input value gets reset. It is not meant to be persistent to that degree.

IgnacioBecerra (approved):
@kennylam Thanks for confirming functionality, LGTM!

Issue #5924, opened 2021-04-28
5152 - Focus state is problematic on mobile


aiproulx:
As per discussion, on page load the focus is on the search field and this loads the keyboard automatically. it is not a user prompt to load the keyboard.

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

ACTION:  KC will submit a bug to fix the fact that the keyboard is open when the user enters the page and the search bar is in focus in mobile.

PR #5971, opened 2021-05-05
fix(masthead-search): disable focus for search open on load

IgnacioBecerra:
### Related Ticket(s)
#5924

### Description
Due to the way the `MastheadSearch` and `MastheadSearchInput` components are built in React, whenever the Search bar is open it's always assumed it's open state came from clicking the Search button to bring up the searchbar. This causes for the focus state go into the bar the moment the "Search Open on Load" story loads. 

This PR adds an additional check before the focus function is called in case `disableFocus` is set to `true`. After `MastheadSearchInput` closes, this boolean is changed to `false` in order for the focus feature to activate as normal upon reopen.

Note that using boolean React states for this would need a bigger refactoring, since whenever `MastheadSearchInput` is closed, its boolean values coming from `MastheadSearch` are reset to what they originally were, meaning that if `disableFocus` was set to `false` upon closing, it would reset back to `true` the next time it opens. With the current fix this boolean is only set to `true` in the "Search Open on Load" story.

### Changelog

**New**

- `disableFocus` prop in `MastheadSearchInput`


kennylam (approved):
LGTM thanks @IgnacioBecerra! Looks like you need to update the snapshots though.

kennylam (commented):
LGTM thanks @IgnacioBecerra!

Issue #5849, opened 2021-04-21
Web component: Masthead - On iPhone, with theme g90 , g100, menu option are not visible for user.


Praveen-sr:
## Detailed description

Issue;
Masthead - On iPhone, with theme g90 , g100, menu option are not visible for user.
The text is same as the background color.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/f177e694-23b7-4a43-871b-a985a07e27d5)

PR #5973, opened 2021-05-05
fix(masthead): lock down menu to white theme on mobile

IgnacioBecerra:
### Related Ticket(s)
#5849

### Description
Locked down the `dds-left-nav` menu and its options to using the white theme in order for the text to be readable using the proper colors.

### Changelog

**New**

- locked down `.bx--side-nav__submenu` to using the white theme


kennylam (approved):
LGTM thanks @IgnacioBecerra!

Issue #5730, opened 2021-04-08
[Card group] Grid mode knob, narrow grid style error


shixiedesign:
This is a recent component and might got affected when we updated Card group. 

### The bug is happening when `Grid mode = Narrow` 🐛 

![Screen Shot 2021-04-07 at 4 37 17 PM](https://user-images.githubusercontent.com/15144993/113956059-e54b7080-97ea-11eb-9618-7c2d562e68be.png)

### `Grid mode = Condensed` is all good 👍 
![Screen Shot 2021-04-07 at 4 37 55 PM](https://user-images.githubusercontent.com/15144993/113956374-5be86e00-97eb-11eb-8213-bd7accde9659.png)

PR #5974, opened 2021-05-06
fix(card-group): card grid alignment

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5730

### Description

@shixiedesign I can fix the alignment perfectly when the card only includes text. 

The problem happens when the card includes an image. 

From my tests, it's mathematically impossible to make it work unless we add a javascript fix that makes all images inside the card same height. To make the last card's content align with the grid, it needs to be the same width as the top card, which takes 33.333...% of the width. If you substract that width plus the two 16px gutters from 100%, we don't have enough space left to make the other two cards in the same row the same width as the last card, which causes their images to be shorter than the last card. 

I can fix that problem by making them all the same width, but the last card wouldn't hang like the top card and image.

### DIFF
--- a/packages/styles/scss/components/card-group/_card-group.scss
+++ b/packages/styles/scss/components/card-group/_card-group.scss
@@ -120,6 +120,13 @@
   // Add Grid Mode Narrow (16px gutter) style.
   :host(#{$dds-prefix}-card-group)[grid-mode='narrow'],
   .#{$prefix}--card-group--narrow {
+    @include carbon--breakpoint('lg') {
+      grid-template-columns: repeat(
+        3,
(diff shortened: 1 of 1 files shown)


shixiedesign:
Talked to Julio on Slack, the Card in card (image included) and all right-most card should not go beyond the pink line. So `narrow grid` mean 16px gutter and nothing hang-right

image

NSDrowned:
@shixiedesign fixed it! Now everything is aligned to grid, with no hang-right 👍

shixiedesign (approved):
Thank you! Looking good 👍

kennylam (approved):
LGTM thanks @NSDrowned!

Issue #5981, opened 2021-05-06
[Modal | Web Components] Background color missing after Carbon v10.34.0 upgrade


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

![image](https://user-images.githubusercontent.com/1641214/117305989-48154380-ae4d-11eb-8ab0-3820f455b7d3.png)

This happens after upgrading to Carbon v10.34.0. Creating this issue so we can approve the Percy diff for now and do a fast follow fix before we go into code freeze.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

Modal

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Restore the background color.

jeffchew:
Should have this PR merged in first before addressing this:
https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/5972

PR #5982, opened 2021-05-06
fix(leaving-ibm): add modal bg color

kennylam:
### Related Ticket(s)

#5981 

### Description

The latest Carbon upgrade (v10.34.0) includes a [change](https://github.com/carbon-design-system/carbon/pull/8486/files#diff-41b303d40b12eaf5e8b1c30c4131aa40d0205342c85fe50c9b75c9079a0510aeL43) which removes the background color from the `modal` component. Still confirming with the Carbon team if this is intentional, but in the meantime this adds it back.

 Can be removed later if it is added back to `carbon-components`. If this change is intentional, the correct place for this change is [carbon-web-components](https://github.com/carbon-design-system/carbon-web-components)

### Changelog

**New**

- add `background-color` to `leaving-ibm-modal` and `expressive-modal`


jeffchew (approved):
LGTM!

jeffchew:
Realized this only fixes Leaving IBM, need to address the expressive modal (and local modale) too.

kennylam:
> Realized this only fixes Leaving IBM, need to address the expressive modal (and local modale) too.

@jeffchew Good catch, fixed.

Issue #5648, opened 2021-03-29
Storybook's "copy" knob isn't working in several stories


shixiedesign:
@iwhitcomb You're right, the copy knob isn't working in a few components. (content group simple). @shixiedesign Seems like the copy issue is a more widespread problem not specific to this component. I think we should make a new story for it to be fixed throughout the codebase

_Originally posted by @andysherman2121 in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/5571#issuecomment-808268788_

ariellalgilmore:
Just a storybook error. After doing some digging, I realized the components effected are any component extending content-block, content-group, and content-item. The copy text for those components are being rendered automatically within the component (similar to how videoPlayer title is rendered in the card heading): https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/packages/web-components/src/components/content-block/content-block-copy.ts#L35-L41. 

![Screen Shot 2021-05-06 at 9.30.08 AM.png](https://images.zenhubusercontent.com/5ee7ae1cc40147132b286905/ac47c0fe-1669-4e19-ab30-2872edafbde4)

Moving this issue to icebox and creating a PR removing all copy knobs effected in web components.

RobertaJHahn:
Closing

PR #5992, opened 2021-05-06
chore(storybook): remove copy knobs not re-rendering in storybook

ariellalgilmore:
### Related Ticket(s)

#5648

### Description

Removing copy knobs for components using `content-block-copy`, `content-group-copy`, and `content-item-copy`. This is a storybook issue where the components don't re-render if the copy input changes; the copy components work properly otherwise. 
### Changelog

**Removed**

- copy knobs


IgnacioBecerra (approved):
LGTM, thanks @ariellalgilmore !

PR #6001, opened 2021-05-07
fix(simple-benefits): removes javascript sameheight

andysherman2121:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/4697

### Description

All of the `simplebenefits-item.scss` styles weren't being pulled in due to lack of import within `ibm-dotcom-styles.scss`. 
This was causing other incorrect changes to be incorporated on-top of this. Also deleted the Sameheight js functionality and added a temporary height to keep everything in place. This component is also still behind a feature flag.


shixiedesign:
First time seeing this component! Based on spec I found, a few spacing issues: 
1. Space under individual headline is doubled. Should be 32px only, followed by the short horizontal rule
image


### Here's the spec:

andysherman2121:
@shixiedesign thanks for the feedback! working  on the fixes now

packages/styles/scss/components/simplebenefits/_simplebenefits-item.scss
```diff
@@ -25,6 +28,8 @@
     }
 
     &__title {
+      overflow: hidden;
+      height: $spacing-11;
```
kennylam, line 32: @andysherman2121 We shouldn't be hard-coding this. Suggest switching to CSS `grid` and having it set the heights equally or we continue to use `sameHeight`.

shixiedesign:
Hey there, given this #6183, should we best merge, or close and deprecate later?

kennylam:
> Hey there, given this #6183, should we best merge, or close and deprecate later?

@shixiedesign Close and deprecate – good idea.

Issue #5954, opened 2021-05-04
[Masthead L1]: L1 Megamenu position is off when user first scrolls page and then opens L1 megamenu


annawen1:
## Detailed description

> Describe in detail the issue you're having.

When scrolling the page and then opening the L1 megamenu, the megamenu top position is off, creating a gap between the menu item and megamenu

## Steps to reproduce the issue

1. go react storybook, DotcomShell component with L1
2. scroll page
3. open menu item with megamenu on the L1

## Additional information

<img width="1461" alt="Screen Shot 2021-05-04 at 12 27 00 PM" src="https://user-images.githubusercontent.com/54281166/117039517-dca96580-acd6-11eb-8889-0ba90866b68a.png">

RobertaJHahn:
@annawen1 We shouldn't be using the mega menu data for the L1.  This isn't a bug, engineering should stop pointing to the L0 data file.  Reach out to @oliviaflory for more details, if needed.

PR #6019, opened 2021-05-10
fix(storybook): remove megamenu from L1 custom data storybook

annawen1:
### Related Ticket(s)

[Masthead L1]: L1 Megamenu position is off when user first scrolls page and then opens L1 megamenu #5954

### Description

As explained in the comments of the bug ticket, we shouldn’t apply mega menus or the L0 data to L1 because the IA guild hasn’t determined if L1s should get mega menus yet. So removing the megamenu data from our custom data file for storybook.

### Changelog

**Removed**

- removed `Products & Solutions` megamenu content data and switching the `hasMegapanel` field to `false`


IgnacioBecerra (approved):
LGTM, thanks @annawen1!

PR #6037, opened 2021-05-11
fix(horizontal-rule): update grid layout

kennylam:
### Description

Currently `horizontal rule` does not align to the grid in some components. This fixes that.

**Before**
image

**After**
image

### Changelog

**Changed**

- update `horizontal rule` grid styles in `leadspace block`, `content block`, and `table of contents` components.


shixiedesign (commented):
Hmm I'm seeing the rule is longer than expected? Screenshot in your description is correct. This is seen in Lead space block **Web components**. React looked good. 

image

kennylam:
@shixiedesign Good catch! That's actually due to web components `leadspace-block` having incorrect grid styles – fixed.

ariellalgilmore (commented):
looks like the column spacing in link list section is being effected? 
Screen Shot 2021-05-17 at 8 42 10 AM

Screen Shot 2021-05-17 at 8 43 28 AM

kennylam:
> looks like the column spacing in link list section is being effected?

Looks like that comes from CSS grid applying same height to all those items. Otherwise the row in the middle would be shorter than the top/bottom.

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/c5b02494f7ba54d2dc82b7ff8275145464ebd078/packages/web-components/src/components/link-list/link-list.scss#L69

shixiedesign:
I was noticing Link list section too - not height of the links, but between the columns of lists, there should be 32px gap. But I checked Canary, looks like Link list section itself is already not to spec, so this PR probably fixed part of it, but we still need a bug fix on Link list outside this PR. Lemme know if sounds right. 

**Design spec**
image

**This PR**
image

**Web comp canary**
image

shixiedesign (approved):
I created a bug issue for link list to fix outside this PR. Lots of clean up here! 🎉

ariellalgilmore (approved):
Thanks @kennylam!!

Issue #6011, opened 2021-05-10
Web component & React: Table of contents - On g90 & g100 theme, menu option are not visible for the user.


Praveen-sr:
## Detailed description

Issue: On tab view, 
On g90 & g100 theme, menu option are not visible for the user. As the text color is same as background color.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/7043f83d-41cd-4377-9065-f1c18e72f7a5)

PR #6041, opened 2021-05-11
fix(table-of-contents): ensure option text color is readable

IgnacioBecerra:
### Related Ticket(s)
#6011

### Description
When using any of the dark themes, the options' color is set to white, which blends with the background. This ensures that the options use the White theme, which uses darker text color tokens and thus will be readable on Windows systems.

### Changelog

**New**

- locked down the ToC option class to using the white theme


packages/utilities/src/utilities/settings/settings.js
```diff
@@ -11,11 +11,11 @@
  * @exports ibmdotcom.settings
  * @type {object} Settings object
  * @property {string} [prefix=dds]
- * Carbon for IBM.com v1.19.0',
+ * Carbon for IBM.com v1.19.1',
```
jeffchew, line 14:
```suggestion
 * Carbon for IBM.com v1.19.0',
```

Looks like this was based on the `v1.19.1` patch release branch.

packages/utilities/src/utilities/settings/settings.js
```diff
@@ -11,11 +11,11 @@
...
  *
  */
 const settings = {
-  version: 'Carbon for IBM.com v1.19.0',
+  version: 'Carbon for IBM.com v1.19.1',
```
jeffchew, line 18:
```suggestion
  version: 'Carbon for IBM.com v1.19.0',
```

Looks like this was based on the `v1.19.1` patch release branch.

jeffchew (approved):
LGTM! We might want to capture this as an e2e test for percy later on.

Issue #5536, opened 2021-03-18
[Content block segmented] Storybooks QA


DragosRistici:
## General issues

Please change this: 
<img width="209" alt="Screenshot 2021-03-18 at 16 07 36" src="https://user-images.githubusercontent.com/48327647/111640569-f9600b80-8804-11eb-9bd5-7a8a5a3511bb.png">

to this:
```
Content block – segmented
    Default
    With link list

```

Rename the themes by their full name, from this: 
<img width="359" alt="Screenshot 2021-03-18 at 14 43 31" src="https://user-images.githubusercontent.com/48327647/111627967-5608f980-87f8-11eb-9bf4-c0424146bd65.png">

to this:
```
White
Gray 10
Gray 90
Gray 100
```

Add the same knobs from the Default story to the third story also, besides what it already has. Make sure you add the specific knobs at the end (link list heading, and maybe add the number of links, 2 or 3).

And also add this, but change the first one to `Container bottom border` and make the options be: `With border` and `Without border`. Make the first one, `With border`, selected by default, so the border is present from the start: 
<img width="635" alt="Screenshot 2021-03-17 at 15 30 43" src="https://user-images.githubusercontent.com/48327647/111521957-6bcbdf80-8762-11eb-8946-0c1384c4c4a1.png">

In React Storybook, for the `With link list` story, the content text CTAs are missing.
<img width="576" alt="Screenshot 2021-03-18 at 16 24 38" src="https://user-images.githubusercontent.com/48327647/111642114-79d33c00-8806-11eb-9bf7-546ec0f19048.png">
<img width="578" alt="Screenshot 2021-03-18 at 16 24 32" src="https://user-images.githubusercontent.com/48327647/111642124-7c359600-8806-11eb-8991-a995b8c5d0b3.png">

1. Again, on React Storybook, the heading is off the grid
<img width="365" alt="Screenshot 2021-03-18 at 16 25 50" src="https://user-images.githubusercontent.com/48327647/111642324-a8511700-8806-11eb-913f-2758125c29d8.png">

2. in Web Components Storybook, make sure the spacing below the heading is 32px and the H3 intro text is present, just like on the other stories. 
<img width="1047" alt="Screenshot 2021-03-18 at 16 26 50" src="https://user-images.githubusercontent.com/48327647/111642798-1269bc00-8807-11eb-97c3-6a58a3e7c927.png">

3. Please apply card max-width of 320px throughout. 
[...]

PR #6072, opened 2021-05-12
fix(content-block-segmented): align react and wc versions

IgnacioBecerra:
### Related Ticket(s)
#5536

### Description
Alignment between the two Content Block Segmented stories in React and WC.

### Changelog

**New**

- added knobs to change number of ListList items and add border knob to all stories

**Removed**

- the video story variation


kennylam:
@IgnacioBecerra The web components version has a video in the default story but the React version doesn't.

ariellalgilmore:
looks great! just a few small things: 
1. The spacing between the border and the cta looks different in react and web components
2. the copy in web components is not in the story `with link list`
3. should the additional story be called `with link list` (web components) or `with aside elements`

IgnacioBecerra:
@ariellalgilmore fixed those issues!

packages/react/src/internal/components/ContentBlock/ContentBlock.js
```diff
@@ -65,7 +65,14 @@ const ContentBlock = ({
...
     <div data-autoid={`${stablePrefix}--content-block`} className={classnames}>
-      {title}
+      {aside && aside.items
+        ? _layoutWrap(
+            <>
+              {title}
+              <div></div>
+            </>
+          )
+        : title}
```
ariellalgilmore, line 75: hmm wondering why this is being added back? It causes the title to go off grid with aside elements. Otherwise looks good!!!
Screen Shot 2021-06-02 at 7 20 45 AM

IgnacioBecerra: I removed what I added back since it's more noticeable with that code. I've looked a bit more into this, and I think this is affecting all Content Blocks whenever there's the LinkList open. I'll be opening a new issue capturing this layout problem if there's not one already so it can be addressed properly.

IgnacioBecerra: update: after merging latest changes, I was able to remove the left padding from the content and everything renders as expected. No need to create an issue as it's been already fixed.

oliviaflory:
Hey @IgnacioBecerra, looking good!

A few comments about the knobs:

1. Knobs are not the same between Web components and React, I assume from the issue they should be the same.
   - In Web components Default and With link list are missing `Copy` knob
 
2. Can we control the order of the knobs? Some of the knobs don't seem sequential and are confusing. Suggested order:
**Default**
   - Heading
   - Copy
   - CTA style
   - CTA type
   - Container bottom border

**Link list**
   - Heading
   - Copy
   - Link list heading
   - Number of links
   - CTA style
   - CTA type
   - Container bottom border


3. React small breakpoint: content is shifting off of the columns and is pressed to the edge of the browser, mostly in the React With link list variant. But also seeing it happen to the CTAs in the Default variant

| small | medium|
| --- | --- |
| Screen Shot 2021-06-03 at 5 00 41 PM | Screen Shot 2021-06-03 at 5 00 27 PM |

4. Link list alignment medium breakpoint
In Web components link list hangs on the left side in medium breakpoints (left)
In React link list is inset in the medium breakpoints (right)

From the spec it looks like link list should be inset on the medium breakpoints. I know there is another PR adjusting something in Link list as well...lmk if there should be a separate issue.

Screen Shot 2021-06-03 at 3 04 54 PM

5.React  Link list alignment lg, xl, max
List list isn't aligned to the columns in react, it's hanging on the left side. idk if this was originally captured so LMK if that should be a separate issue.
I believe there is _another_ issue to fix the number of columns link list spans...It should be 3 not 4 :/ 

Screen Shot 2021-06-03 at 5 24 20 PM

IgnacioBecerra:
Thanks for the feedback @oliviaflory!

About the copy knob comment, there's an issue in Web Components in which the copy knob for Content Block stories (and some others as well) aren't able to be updated due to the way they are being rendered. We made a decision to icebox this for now, ([here's the Slack thread for more info](https://cognitive-app.slack.com/archives/GBG66T9NK/p1619795769013200?thread_ts=1619723263.090700&cid=GBG66T9NK))

About the Link List components, I think they could be fixed in another issue since they'll be affecting other components, and I think those other issues may be addressing this as well. 

Otherwise, the other comments have been addressed!

oliviaflory:
Notes from Slack convo with Ignacio:
- Fixing missing bottom border in react
- Fixing padding

IgnacioBecerra:
@oliviaflory These have been fixed in the latest commit.

oliviaflory (approved):
Looks good @IgnacioBecerra thanks for the updates!

Approved the Percy build even though it's not picking up the video in some story, as we discussed over slack we aren't sure why but the storybook looks good 👍

jeffchew:
@IgnacioBecerra Can you check prettier formatting, looks like that's failing CI checks.

@oliviaflory @IgnacioBecerra I believe we have the video story suppressed in percy since it is loaded asynchronously, it sometimes would cause false positive reports.

jeffchew (approved):
LGTM!

@IgnacioBecerra @oliviaflory not sure if design needs to be re-approved, there weren't any changes that happened that would require another review since it was just fixing lint issues.

PR #6074, opened 2021-05-13
chore(dependabot): adding workflow to auto update offline mirror

jeffchew:
### Related Ticket(s)

No related issue

### Description

One of the challenges of dependabot updates is that it does not update
the yarn offline mirror when adjusting the yarn.lock file. This will
hopefully detect any dependabot PRs, then will update the offline mirror.

### Changelog

**New**

- dependabot-yarn-mirror.yml


ariellalgilmore (approved):
thanks @jeffchew !!

Issue #6060, opened 2021-05-12
React: Masthead - Search - On empty search bar, clicking enter key, takes the user to results page.


RobertaJHahn:
## Detailed description

Issue:
Masthead: Search - On empty search bar, clicking enter key, takes the user to results page.

## Additional information
- This fix will be contributed by lathaMaramganti from Career Quest.

![Search_enter.gif](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/3ad3de6e-8840-46a7-a6ef-94844fec3fb7)

PR #6082, opened 2021-05-13
fix(masthead): prevent enter key search when search field is empty

lathaMaramganti:
### Related Ticket(s)

#6060 

### Description

Disabled the search on enter key when search field in Master head is empty

### Changelog

**New**

-  Added condition in onKeyDown method  (Specific to Enter key) to check if the search field is empty


IgnacioBecerra (approved):
Looks good to me, nice work Latha!

Issue #6040, opened 2021-05-11
[Logo grid] Storybook cleanup 


shixiedesign:
## Logo grid storybook example can be improved. 
- Should be on grid! 
- Remove CTA from example. It is an optional CTA but should not be in default
- Ideally, instead of images we use logos. [Here's a folder of logos ](https://ibm.box.com/s/c51wajohvdl92q0ywhhy2t5yuglj18ne)

### Current screenshot with expected position of logo images in pink:
![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/efe6b570-af11-4585-b190-3efebb9bb2d3)

DragosRistici:
@shixiedesign what do you think, should we update the spacing below the heading here, too, or not? From 64 to 32?

shixiedesign:
Good point @DragosRistici ... spec is not up to date, lemme update to 32px for now.

PR #6097, opened 2021-05-14
fix(logo-grid): styles and functionality fixes

kennylam:
### Related Ticket(s)

#6040 

### Description

This aligns both React and web components `logo-grid` to the grid, as well as various other styling fixes. Also includes fixes to the web components version behavior where a space would be rendered for the CTA even when it did not exist.

### Changelog

**New**

- company logos

**Changed**

- align both React and web component versions to the grid
- spacing fixes
- fixes web components CTA footer rendering behavior
- use `dds-hr` instead of CSS border

**Removed**

- CTA from `logo grid` stories


IgnacioBecerra:
@kennylam Looking great with the new logo images! One quick thing, since we're removing the CTA from the stories, maybe we should also remove the CTA knobs in the Web Components story to match React's.

kennylam:
> maybe we should also remove the CTA knobs in the Web Components story to match React's.
@IgnacioBecerra Good catch, thanks!

shixiedesign:
Grid seems a bit off - rule is 16px too long on each side, and content should start on Column, currently hanging: 

image

kennylam:
> Grid seems a bit off

@shixiedesign Good catch, this was being caused by extra spacing in the Storybook story, which I've now removed.

shixiedesign (approved):
Looking cleeeaan! Thank you!

Issue #4083, opened 2020-09-30
Web Component:  Audio Player


tweenn:
This component will be developed in both Web Components and React as a wrapper.  

-  Corresponding React epic (#5106)

### The problem

Marketing need a way to deliver podcasts and other audio files to their clients.
As Mediacenter/Kaltura is an approved source for hosting audio, we need a way to deliver the kaltura player - within an audio experience - to our clients.

### The solution

1 - We need an IDL compliant layout to build the audio player within Kaltura.
2 - This player, that is already leveraged as the video player, will automagically know based on the media-type that it is rendering an audio (or a video) - in order to render the right visualization / layout
3 - Will be consumed and made available as a component in Carbon For IBM.com

### Additional information

It would be great to have this specs take it account:
- SRT (closed caption) display capable and translatable - optional 
- Thumbnail inclusion - mandatory
- Large scrub bar - larger than a video scrub bar
- Large play/pause interface
- 1.5x, 2x playback rate options
- Current time point displayed on scrub bar
- Markered scrub timeline (1/10+) 
- This audio file will be hosted in the Kaltura application and will be implemented using the Kaltura API.
- [Northstar audio player](https://www.ibm.com/standards/web/v18/design/audio-player/)

This is how Mediacenter's player act when playing an audio-file:
https://pages.github.ibm.com/mediacenter/docs/examples/audio/

- The Kaltura API includes an event within the player that includes the media type.  Media type codes:  1 = video, 2 = image, 5 = audio, 201 = Live stream flash, 202 = Live stream windows media, 203 = Livestream real media and 204 = Livestream quicktime
- The Carbon for ibm.com calls the API to get the description etc...
- The Carbon for ibm.com has to be able to handle both a video and audio source file.  This component may need to be renamed. (possibly supporting video, audio and images)
- [Box folder with sample audio files](https://ibm.ent.box.com/s/paut8w74l9sjqzkkzbqch8vbn3byoba9)
-[Audio Player Box folder](https://ibm.box.com/s/05ndo5nsg3sonc3wyskcbkoi93lhdoor)

Test Audio files included under Partner ID:  1773841
     - [Test Audio - IBM Elevator Pitch Series EP1 - Supply Chain](https://mediacenter.ibm.com/media/Test+Audio+-+IBM+Elevator+Pitch+Series+EP1+Supply+Chain/1_gp572bda)
[...]

PR #6102, opened 2021-05-14
feat: audio-player and kaltura player service

tweenn:
### Related Ticket(s)

I believe the one I've opened (#4083) was closed due to #5105

### Description

This PR implements 3 things:
1. A Kaltura Player Based Audio Player Component.
2. A Kaltura Player Service / API
3. A tweak to the Video Player Service embed method

* Kaltura Player Service / API
As an Audio Player Service had to be created in order for the component to work properly and due to the similarities to the already existing Video Player Service, we agreed in May 5 office hours that instead of duplicating it, a Kaltura Player Service would be created and in a future effort, either the Video Player component or the Video Player Service would be migrated to this new service.

* Video Player Service Tweak
I've added a parameter into the flashvars object of the embed method that works specifically with the mediacenter-plugin and is quite harmless for the overall Kaltura Experience as it is an object, whose key is ibm - in order to tell that the player should render the IBM Design Language rulesets (colors, icons and other things)

### Changelog

**New**

- Kaltura Player Based Audio Player Component.
- A Kaltura Player Service / API

**Changed**

- Video Player Service


jeffchew:
@shixiedesign fyi, audio player is behind a feature flag: https://ibmdotcom-react-experimental.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/6102/index.html?path=/story/components-audioplayer--default

shixiedesign:
Hi @tweenn, this is looking great! I noticed some visual adjustments you've made and agree with them, so I can update the design spec to align. But here are a few changes I think we should make: 

### Thumbnail related

1. Thumbnail seems to be getting an extra border? Should be inset and flush to player background.
May-19-2021 09-29-28

2. Thumbnail is 16px from left of the player. Expecting no padding left of thumbnail: 
**Current:**
image

**Expecting:**
image

### Play/pause button related

3. Play/pause button should be the 32px icon. This requires a bit of customization over Carbon Icon buttons. But visually it's important to make it prominent. 

**Current:** 
image
**Expecting:**
image

4. Play/pause should not become `Replay`. After audio ends, the Play icon should be visible again. 
image

### Functionality wise

5. Onload, current behavior shows auto play, but on mute. Mute is not visible until user click on volumn slider. This is a buggy experience. Do not default to play on mute with no visual indicator. 
 
**Expected behavior is, onload, the player is at default state, not playing, and play icon should be visible. Then, when Play is clicked, audio should play at set volumn.** 
**Design spec default state**
image

tweenn:
> 0. Onload, current behavior is auto play, but on mute, while volume slider shows full volume. This is a buggy experience. Do not default to play on mute with no visual indicator. Expected behavior is, onload, the player is at default state, not playing, and play icon should be visible. Then, when Play is clicked, audio should play at set volume.

Currently this is a problem related to the kaltura player, not the carbon audio-or-video-player.
If you go into the [PR preview video-player page](https://ibmdotcom-react-experimental.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/6102/index.html?path=/story/components-video-player--default) and try to click on the video, it will play muted - with the un-mute button on top left as in the screenshot below.

audio-player-unmute-issue-on-video-player min

Which is not the behavior that happens on any carbon-based page on IBM.com - [like this one](https://www.ibm.com/sports/overwatchleague/) as you can see in the leadspace CTA.

We "fix" that behaviour in mediacenter through a player-plugin (so it work on northstar, carbon, etc), but as we speak, there is a domain restriction for such behaviour / fix in our end.
We've discussed this with our security overseer and I'm dropping this restriction in order to make the player act correctly everywhere where the player can play - buuuut that's still in the deployment pipeline to go live (which I'm not sure will be done by the time you've check this)

-----

> Itens 1 ~ 7

Should be resolved

-----

> Closed caption
> Any reason this is wrapped to 2 lines? Should be enough space without having to wrap text.

In this video, they're not being wrapped, it's just that sometimes the caption text is broken into two or more lines, as in the screenshot below

audio-player-caption-breakpoint min

shixiedesign:
Hi @tweenn sorry just 1 more observation. Noticing the "Disallow" cursor. Is this a change from before? The buttons should still be clickable before play, but the effect won't be visible until play is happening. 

May-27-2021 16-20-31

jeffchew:
Hi @tweenn ! Just checking if you saw @shixiedesign 's comment above? We'll see if we can get this merged in before end of week, which is our code freeze.

tweenn:
@jeffchew & @shixiedesign 

> We'll see if we can get this merged in before end of week, which is our code freeze.

Much appreciated for the help and attention!
A huge thank you!

-------

> The buttons should still be clickable before play, but the effect won't be apparent until play starts.

I'll deal with the idea of being able to change volume and maybe speed-rates and closed captions.

For the first two itens, I've blocked interaction before the real player gets instantiated behind the scenes because when the player loads, it has some dire rules on deciding which configuration it will follow.
To put it simple, the player will check if there is a configuration-cookie, then if it was instantiated with a parameter, check if there is a player configuration default and finally assume the kaltura default for that.

Which mean that if the user interacted with said configuration before, let's use volume for the sake of example, the player will try to load the cookie set from last session as default.
If the user manually changed the player volume before hitting play, there is some extra steps needed to ensure that such volume will be enforced when the player is ready and before playing.

Now, for closed captions, there are asynchronous rules for loading then, that will trigger way after the player is ready to play - and probably playing - and in order to be able to interact before the real player load, may be a spike that I have to investigate if the api for listing CC's is public.

All of it can be done, but may need some harsh updates to the player load behind the scenes.
Again, not saying no, just explaining that it may need some time to tweak the behaviour.

-------
Now for the color change over the disabled iconography:

I had to `yarn reset` my project in order to reproduce the behaviour and it seems to be a side-effect of a button with the combination of being ghost and hasIconOnly.
```
<Button
  kind="ghost"
  hasIconOnly
>
```

To be sure this wasn't something I did in the audio-player component, I went into the ButtonGroup component and changed the props until I got this reproduction:

| *Disabled Prop Active:* | *Not disabled:* |
| ----------------------- | ---------------- 
| image | image |

-------

From what I was able to debug, the `<path>` elements inside the `<svg>` of the icon prop will have a css rule that force then to be:
```
.bx--btn--ghost.bx--btn--icon-only .bx--btn__icon path:not([data-icon-path]):not([fill='none']),
.bx--btn--ghost.bx--btn--icon-only .bx--btn__icon {
	fill: var(--cds-icon-01, #161616);
}
```

Aside from that finding, I couldn't figure out where was the problem as I'm not well versed in react and storybook to figure that out, but I'm pretty sure it was introduced on the master-branch merge #597f755 
Oh, and I've tried searching for key workds like `btn--ghost` or `btn__icon` in the scss folder to see if there was something, but wasn't able to figure it out, so in my limited knowledge, I'm inclined to say it's something on @carbon, right?

-------

tweenn:
Quick Update on the color change,

I've went into [carbon's react storybook](https://react.carbondesignsystem.com/?path=/story/components-button--playground) and on the button-playground, the behaviour was exact the same as related on [my comment above](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/6102#issuecomment-852720916) - ghost button with icon-only get `<path>` fill color to `#161616`

image


-----

With that said, @shixiedesign should I enforce manually the icons to be `$ui-icons-03` when disabled?

packages/react/src/components/AudioPlayer/AudioPlayerVolumeControl.js
```diff
@@ -0,0 +1,123 @@
...
+    }
+    return VolumeUp32;
+  };
+
+  const isTooltipVisible = () => {
+    if (displayVolumeControl) {
+      return `${prefix}--force-tooltip-hidden ${prefix}--menu--open`;
+    }
+    return '';
+  };
```
kennylam, line 55: Whenever possible we prefer to use `cx` to set classes, e.g.
```suggestion
  const buttonClasses = cx({
    [`${prefix}--force-tooltip-hidden ${prefix}--menu--open`]: displayVolumeControl,
  });
```

packages/react/src/components/AudioPlayer/README.stories.mdx
```diff
@@ -0,0 +1,124 @@
...
+  return (
+    <AudioPlayer
+      videoId="your-video-id"
+    />
+  );
+}
+ReactDOM.render(<App />, document.querySelector('#app'));
+```
+
+Add the following line in your `.env` file at the root of your project.
```
kennylam, line 52: A section can be added here regarding the environment variable needed since it's behind a feature flag. See [this](http://ibmdotcom-react-experimental.mybluemix.net/?path=/docs/components-callout-data--default#feature-flag) example.

packages/react/src/components/AudioPlayer/README.stories.mdx
```diff
@@ -0,0 +1,124 @@
...
+```
+KALTURA_PARTNER_ID=243342
+KALTURA_UICONF_ID=12905712
+```
+
+## Manual usage of Kaltura API
+
+With `@carbon/ibmdotcom-services`, you can keep track of the readiness of
+[Kaltura API](http://player.kaltura.com/docs/api) and use it when it is ready.
+Here is an example of stopping the video player when it is hidden in DOM:
```
kennylam, line 78:
```suggestion
Here is an example of stopping the audio player when it is hidden in DOM:
```

packages/react/src/components/AudioPlayer/README.stories.mdx
```diff
@@ -0,0 +1,124 @@
...
+  entries.forEach(entry => {
+    const { width, height } = entry.contentRect;
+    if (width === 0 && height === 0) {
+      // Uses Kaltura API to stop the video
+      entry.target.sendNotification('doPause');
+    }
+  });
+});
+
+VideoPlayerAPI.checkScript().then(() => {
```
kennylam, line 94:
```suggestion
KalturaPlayerAPI.checkScript().then(() => {
```

kennylam (changes requested):
@tweenn Thanks for this, looks amazing! Just a few comments below. Also as a follow-up we probably want to open up certain labels, such as `iconDescription`, as props to allow localization. But that can be addressed in a future PR.

packages/styles/scss/components/audio-player/_audio-player.scss
```diff
@@ -0,0 +1,152 @@
...
+@mixin audio-player {
+  .#{$prefix}--audio-player {
+    position: relative;
+    display: flex;
+    align-items: center;
+    height: 48px;
+    padding: 0 8px 0 0;
+    background-color: $ui-01;
+
+    &__captions {
```
ariellalgilmore, line 21: looks great!! Just to add one small thing to @kennylam's comments. Wondering if we can avoid concatenation in the styles. For example, `.#{$prefix}--audio-player__captions` instead. This will help us in the future when building the web component version

shixiedesign:
@tweenn That's weird. I will look into it. Yes I think `enforce manually the icons to be $ui-icons-03 when disabled` is a good solution for now! Thank you! This looks really good!

packages/react/src/components/AudioPlayer/README.stories.mdx
tweenn: Oh, I totally forgot to come back into that file and reword `video` to either `media` or `audio`.

packages/react/src/components/AudioPlayer/README.stories.mdx
tweenn: There is another code error in the examples on line 44~46 that I'm fixing
```
<AudioPlayer
  videoId="your-video-id"
/>
```

packages/styles/scss/components/audio-player/_audio-player.scss
```diff
@@ -0,0 +1,152 @@
...
+@import '../../globals/imports';
+@import '../../globals/utils/ratio-base';
+@import 'carbon-components/scss/components/slider/slider';
+
+@mixin audio-player {
+  .#{$prefix}--audio-player {
+    position: relative;
+    display: flex;
+    align-items: center;
+    height: 48px;
```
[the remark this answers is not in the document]
tweenn: The only value that I couldn't convert directly or that I thought that using a calc() wasn't that nice was the width of the scrubbers (audio time and volume), so I've left then in `rem`.

And the borders of 1px and a box shadow of 6px from the overflowmenu that I've copy-pasted from carbon directly.

tweenn:
@ariellalgilmore @kennylam & @shixiedesign 

> Thanks for this, looks amazing!

Thank you!
Much appreciated!

> Just a few comments below.

I believe that my last commits have addressed all the conversations above - and a little more things that I've found while refining the code.

Like a flickering in the width that happened in the transition of disabled buttons for the overFlowMenu component after audio started.

> Also as a follow-up we probably want to open up certain labels, such as iconDescription, as props to allow localization.

Sure this would be very handy!
Also, we could use the same idea/behaviour for the `<video-player>` component as the kaltura player has a feature for localization that can be set either through `flashvars` over the api.embed (which may not be the best approach) or manually set through readyCallback.

shixiedesign (approved):
Awesome!

kennylam (approved):
LGTM thanks for this @tweenn!

Issue #6104, opened 2021-05-17
[Footer] combo-box import breaking builds


kennylam:
There is a [`combo-box.scss`](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/packages/web-components/src/components/footer/combo-box.ts#L14) import breaking builds when a bundler such as Webpack is not used. This import is unnecessary since the styles for `combo-box` are already import [here](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/packages/web-components/src/components/footer/footer.scss#L12), and bundlers are not a requirement when using web components.

This breaks the build when using either `footer` or `dotcom-shell` components.

![image](https://user-images.githubusercontent.com/909118/118487810-158b0680-b6e9-11eb-95d1-fd2207252009.png)

PR #6105, opened 2021-05-17
fix(footer): remove combo-box.scss import

kennylam:
### Related Ticket(s)

#6104 

### Description

Removes an unneeded `scss` import which is breaking builds when using `footer` or `dotcom-shell` components.

### Changelog

**Removed**

- extra `combo-box.scss` import


jeffchew (approved):
Thanks @kennylam !

PR #6133, opened 2021-05-19
feat(card): introducing static variant with tag-group

IgnacioBecerra:
### Related Ticket(s)
#5962

### Description
`DDSCard` will becoming "static" if no link is passed to the `href` attribute. This new variant includes the light theme, with optional tags.

### Changelog

**New**

- Card static variant when no `href` is passed in

**Changed**

- set `delegateFocus` in `this.attachShadow` to false due to focus bug only happening in Chrome


kennylam (approved):
LGTM thanks @IgnacioBecerra!

shixiedesign:
- Shouldn't the Static card get the outline from Card group? 
- Hover on text link should get the underline (pls double check to make sure this doesn't affect the other components! I think we've seen the underline in link too often in the wrong places) 
May-24-2021 11-56-25

- Knobs should have: these options: 
image

photodow:
In addition to some of Shixie's comments, can we get a knob to toggle the light variant theme on and off, and maybe able to see it with an image and eyebrow. I'm really curious about how it's impacted with an image in there.

shixiedesign (approved):
Looking good!

PR #6152, opened 2021-05-21
fix(card-group): optional border for card group default

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5993

### Description

Include an optional border for `Cards group - Default`


packages/styles/scss/components/card-group/_card-group.scss
```diff
@@ -162,6 +161,63 @@
     }
   }
 
+  :host(#{$dds-prefix}-card-group)[border='visible'] {
```
IgnacioBecerra, line 164: This would be changed to
```suggestion
  :host(#{$dds-prefix}-card-group)[border] {
```

packages/web-components/src/components/card-group/card-group.ts
```diff
@@ -166,6 +166,14 @@ class DDSCardGroup extends LitElement {
   @property({ attribute: 'grid-mode', reflect: true })
   gridMode = GRID_MODE.COLLAPSED;
 
+  /**
+   * An optional 1px border surrounding the component.
+   * Hidden/0px (default) | Visible/1px).
+   */
+
+  @property({ attribute: 'border', reflect: true })
+  optionalBorder = OPTIONAL_BORDER.HIDDEN;
```
IgnacioBecerra, line 175: Just a little nitpick, I think it would be easier for adopters to only have to pass in a boolean for enabling the border. So instead of having to use border='visible' they can simply pass in border to enable, or omit to have the default borderless group.

```suggestion
  @property({ type: Boolean, reflect: true })
  border = false;
```

packages/web-components/src/components/card-group/card-group.ts
```diff
@@ -185,6 +193,7 @@ class DDSCardGroup extends LitElement {
     const slotClasses = classMap({
       [`${prefix}--card-group--narrow`]: this.gridMode === GRID_MODE.NARROW,
       [`${prefix}--card-group--collapsed`]: this.gridMode === GRID_MODE.COLLAPSED,
+      [`${prefix}--card-group--border`]: this.optionalBorder === OPTIONAL_BORDER.VISIBLE,
```
IgnacioBecerra, line 196:
```suggestion
      [`${prefix}--card-group--border`]: this.border,
```

packages/web-components/src/components/card-group/defs.ts
```diff
@@ -21,3 +21,18 @@ export enum GRID_MODE {
...
+  /**
+   * Show 1px border
+   */
+  VISIBLE = 'visible',
+
+  /**
+   * Hide 1px border
+   */
+  HIDDEN = 'hidden',
+}
```
IgnacioBecerra, line 38: I think we can get rid of these since we're just dealing with a boolean
```suggestion
```

packages/web-components/src/components/card-group/card-group.ts
```diff
@@ -13,10 +13,10 @@ import settings from 'carbon-components/es/globals/js/settings';
 import ddsSettings from '@carbon/ibmdotcom-utilities/es/utilities/settings/settings.js';
 import { sameHeight } from '@carbon/ibmdotcom-utilities';
 import { baseFontSize, breakpoints } from '@carbon/layout';
-import { GRID_MODE } from './defs';
+import { GRID_MODE, OPTIONAL_BORDER } from './defs';
```
IgnacioBecerra, line 16:
```suggestion
import { GRID_MODE } from './defs';
```

packages/web-components/src/components/card-group/__stories__/card-group.stories.ts
```diff
@@ -24,11 +24,16 @@ import imgXlg4x3 from '../../../../../storybook-images/assets/1312/fpo--4x3--131
...
 import imgSm4x3 from '../../../../../storybook-images/assets/480/fpo--4x3--480x360--005.jpg';
-import { GRID_MODE } from '../defs';
+import { GRID_MODE, OPTIONAL_BORDER } from '../defs';
 import styles from './card-group.stories.scss';
 
 import readme from './README.stories.mdx';
 
+const optionalBorders = {
+  [`Visible`]: OPTIONAL_BORDER.VISIBLE,
+  [`Hidden`]: OPTIONAL_BORDER.HIDDEN,
```
IgnacioBecerra, line 34:
```suggestion
  [`Visible`]: true,
  [`Hidden`]: false,
```

IgnacioBecerra (commented):
It's looking great, @NSDrowned!

packages/web-components/src/components/card-group/card-group.ts
NSDrowned: Great suggestion, just committed the changes 👍

packages/web-components/src/components/card-group/card-group.ts
```diff
@@ -185,6 +193,7 @@ class DDSCardGroup extends LitElement {
     const slotClasses = classMap({
       [`${prefix}--card-group--narrow`]: this.gridMode === GRID_MODE.NARROW,
       [`${prefix}--card-group--collapsed`]: this.gridMode === GRID_MODE.COLLAPSED,
+      [`${prefix}--card-group--border`]: this.border,
```
photodow, line 196: Is this separate from the grid mode?

image

packages/styles/scss/components/card-group/_card-group.scss
```diff
@@ -162,6 +161,68 @@
...
   }
 
+  :host(#{$dds-prefix}-card-group)[border] {
+    &::after {
+      content: none;
+      box-shadow: none;
+    }
+
+    ::slotted(#{$dds-prefix}-card-group-item) {
+      border-right: 1px solid $ui-03;
```
photodow, line 171: Is this work overwriting what Ignacio, is doing over in #6133? Or is it going to leverage that same work?

packages/web-components/src/components/card-group/card-group.ts
NSDrowned: Yes, completely independent 👍

packages/styles/scss/components/card-group/_card-group.scss
NSDrowned: Everything is contained inside the border attribute, it shouldn't affect his work.

photodow: But, why not leverage his work. Why maintain the same border in two places?

packages/web-components/src/components/card-group/card-group.ts
photodow: It's specced out as a grid mode. Not separate or independent of them.

packages/styles/scss/components/card-group/_card-group.scss
NSDrowned: From what I discussed with @IgnacioBecerra, his cards only get a border if they don't have a href. So, for example, if you fill a card group with cards with links in them, the border would't be visible without this new attribute.

photodow: > his cards only get a border if they don't have a href

I don't think that's supposed to be the case. The border is supposed to be optional for the card as a whole.

packages/web-components/src/components/card-group/card-group.ts
NSDrowned: Ok, let me fix that 👍

NSDrowned: Fixed this 👍

packages/styles/scss/components/card-group/_card-group.scss
kennylam: > The border is supposed to be optional for the card as a whole.

If this is the case then shouldn't @IgnacioBecerra be leveraging this work, since his PR is specifically targeting card without an `href`, whereas this is for `card-group-item` in general.

IgnacioBecerra: We're working together on making sure things don't overlap, just waiting for some design confirmation.

shixiedesign (commented):
So close! 
1. Border is not showing up on the left of Card group: 

image

2. Ideally, the knob would be set as: 
```
Outlined card [ Yes / No ]
```
where outlined cards are using Light variant (background $field-02). Here's a reference from the Card static PR. Just wanna align the appearance of the card & knobs: 

image

shixiedesign:
Hmm just checked again, the border left is showing when I view the component outside storybook. FYI @NSDrowned.

DragosRistici:
A few things from me here @NSDrowned : 

#### 1. On Web Components `Card group - default` showing 6 cards but the knob says 5. This difference maintains if I add or subtract cards.

#### 2. The border knob triggers the cards' content change like it loads another group of different cards in its place, instead of just adding a border.

https://user-images.githubusercontent.com/48327647/119662595-3b778000-be3a-11eb-995f-a101e8230899.mov


#### 3. On React, there is no `border` knob.

NSDrowned:
@shixiedesign just pushed a fix for the left margin. To further explain, the original card group is using padding to create the inner borders and the calculation creates a lot of decimals. When I use the same method in the outer borders, chrome rounded those decimals down in some resolutions, so the padding was hidden. That's why it worked outside of storybook. 

I changed everything to use css borders instead of padding 👍

NSDrowned:
@DragosRistici I fixed the knob count and the made the random titles static so the cards don't get reloaded (well, they do get reloaded but it's unnoticeable). This issue only applies for Web Components. The React version is coming in #6049 👍

photodow:
Is this a separate issue? I'm seeing alignment issues on the body copy.

image

photodow:
Last item with images appear to have a px gap, and a px shorter.

image
image

photodow:
Separate issue, but our card in card example is outside of card group.

image

It might also be causing some of the problems here...
- no border on card in card
- gap on right side of card in card
- gap below card in card

image
image
image

photodow:
@shixiedesign correct me if I'm wrong, but I think she wanted the light variant in addition to the border being chosen when the outlined cards knob is turned on.

So outline cards knob true sets the grid mode to border, and sets the cards to the light color variant.

image

image

NSDrowned:
Found a bug in `card in card` and `card in card with images`

photodow:
1. Clickable light cards are missing the interactive states. Maybe you haven't pulled in #6213 yet?
2. [There are still some alignment issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/6152#issuecomment-849021870) that haven't been addressed when Card in card is present. Card in card doesn't get the border treatment, and it has a separate grid mode because it's separate from the card group for some odd reason.

> I'm not seeing the alignment issue anymore either. I must have hit a really big edge case with that on accident or did something myself to make it misalign.

NSDrowned:
@photodow I pulled the latest code and got interactions. As for the border issue, Dragos opened a separate issue about it #6234 since it's related to another variant (and happening in canary)

photodow (approved):
I added some additional items over in #6234 that need to be fixed. Focus state is still missing but I think there was a plan to fix that already in motion. Other than those things this works.

IgnacioBecerra (approved):
LGTM, thanks @NSDrowned!!

Issue #5852, opened 2021-04-21
TypeScript 3.8 breaking change


whoisbob:
Similar to this issue:
 https://github.com/carbon-design-system/carbon-web-components/issues/328

And I quote (from above):
> With the launch of the TypeScript 3.8 version Microsoft introduced a breaking that affects to the carbon-custom-elements. As we can see in the next image:
<img width="1305" alt="Screen Shot 2021-04-21 at 9 32 02 AM" src="https://user-images.githubusercontent.com/71522536/115562398-70fddc00-a284-11eb-8103-3d49d5a7c46a.png"> (taken from my machine)

> It seems that after 3.8 version the index must contain a strict type and we are using in the addEventListeners and removeEventListeners (from what I could investigate) a K or K_1 type to index.

RobertaJHahn:
Teh scope of this issue is to perform the investigation of the impact and future work.

whoisbob:
We have a client-facing project that needs to align with carbon standards, is there any chance this could be prioritized @RobertaJHahn

whoisbob:
F.Y.I. current version effectively breaks support for angular 10+

RobertaJHahn:
Do you have a time frame...Date needed by?

whoisbob:
We have a MTP next Friday, May 7th. Would it be possible to include this then?  @RobertaJHahn

RobertaJHahn:
Assigned to Raph as part of Fireline support.  More answers needed.

jeffchew:
Just to capture some of the findings, this appears to be the usage of `new CustomEvent` in various components, which has been brought up as an issue here:

https://github.com/Microsoft/TypeScript/issues/28357#issuecomment-818069744

The fix may be going through and creating an interface for these CustomEvents so that Typescript won't be upset about these.

@RaphaelAmadeu @annawen1

omkardixit:
Is there any target date for fixing this issue?

PR #6158, opened 2021-05-21
fix(custom-events): onloadend type defined

RaphaelAmadeu:
Related Ticket(s)
#5852

Description
I've added a type definition for loadend on the global ElementEventMap, so we fix the error that was blocking Angular 10 development.


jeffchew (approved):
thank you for finding the fix!!

Issue #5678, opened 2021-03-31
[Performance] Investigate performance issues with web components pages


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

Performance runs on AEM test pages as well as the Partners page are showing poor performance scores in Lighthouse.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

General

> What did you expect to happen? What happened instead? What would you like to
> see changed?

The script evaluation time appears to be high. There needs to be an investigation on what is causing this, if this is application level improvements that need to be done or if there are any identified issues in the components themselves.

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

Brought up by AEM afterburner team

## Steps to reproduce the issue

1. Go to https://www.ibm.com/aem-test/edi-electronic-data-interchange or https://www.ibm.com/partners/
2. Run a Lighthouse performance test (mobile setting)
3. See outcome

## Additional information

- One thing that stands out is a lack of code splitting strategy on the application side. Suggested reading: https://medium.com/front-end-weekly/webpack-and-dynamic-imports-doing-it-right-72549ff49234
- Should also look to see if imagery can be improved on the Partners page, as the assets are quite large.
- Ensure that `ibm-common.js` is not blocking

jeffchew:
Closing this, all issues attached have been addressed.

PR #6164, opened 2021-05-24
feat(image): adding `loading="lazy"` to images in web-components

jeffchew:
### Related Ticket(s)

Refs https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5678

### Description

This is a feature that would include native lazy loading of imagery.
While it isn't supported across all browsers, it will work in prevelent
browsers like Chrome.

### Changelog

**Changed**

- added `loading="lazy"` to ``


kennylam (approved):
LGTM thanks @jeffchew!

jeffchew:
> LGTM thanks @jeffchew!

Thanks @kennylam !

One of the things I noticed is that images aren't fully loaded in the long pages (like the dotcom stories), wondering if we want to skip some of the dotcom stories, or at least suppress the middle section?

Issue #6086, opened 2021-05-13
[Link list] CTA type - Video is broken


shixiedesign:
When Link list's CTA type is set to `Video`, the styling is wrong. Currently showing heading-03, or card title style, but it should be still the link style: 

### Current 
Web components: https://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-link-list--default
![May-13-2021 14-29-16.gif](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/5c17fb09-dfc4-4cf9-99b6-b9ad620602a4)

React: https://ibmdotcom-react-canary.mybluemix.net/?path=/story/components-link-list--default
![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/1c4ccdf4-03c1-43a6-85ed-8b507cef6d40)

### Expecting

![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/1ffac59b-a98a-4052-b4fb-81dff2246ae4)

RobertaJHahn:
The linked PR is for React only.  Another PR is needed for the Web Components fix.

PR #6184, opened 2021-05-25
fix(linklist): video heading link style is broken

lathaMaramganti:
### Related Ticket(s)

#6086 

### Description

add the proper link style to the heading of CTA type video

Before fix screenshot : 

image

After the fix screenshot : 
image

### Changelog

**New**

- {{new thing}}
   Added new class in scss to apply link style for the heading when CTA type is video


shixiedesign:
Thank you! React is fixed 🎉 . Would Web components be split out as a new issue?

shixiedesign (approved):
👍 please just make sure not to close the issue yet. Web components to be fixed using the same issue later.

ariellalgilmore:
@jeffchew we were just focusing on the react issue

> @ariellalgilmore are you working with @lathaMaramganti to fix in web components in a separate PR?

we have just been focusing on react only

Issue #6173, opened 2021-05-25
Web component: Masthead - focus on the search bar should be removed when open by default


Praveen-sr:
## Detailed description

Issue:
Masthead - On opening search edit bar, Default Focus on the search edit bar should be removed.

## Additional information

https://images.zenhubusercontent.com/193129061/ad3285be-99cc-44e4-85d4-c83f4a32561d/rpreplay_final1621852603.mp4

ariellalgilmore:
This has been fixed and should be working in canary: http://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-masthead--default

RobertaJHahn:
Closing based on Ari's comments above.

Praveen-sr:
@ariellalgilmore Another issues related to focus position on search edit bar :- 
On Mobile, when bar is opened initially there is a delay in focus position set to left of search edit field.

PR #6208, opened 2021-05-26
fix(masthead-search): delay position for cursor focus

ariellalgilmore:
### Related Ticket(s)

#6173 

### Description

Cursor was rendering in center of the search for a second before jumping to the left align position. 
https://user-images.githubusercontent.com/20210594/119724167-378c3380-be23-11eb-9ced-0879aca47544.MP4

### Changelog

**Changed**

- move focus call to end of function


kennylam:
@ariellalgilmore This was only occurring with the `search open onLoad` story and it would autofocus on the search bar, correct? Since the autofocus has been removed, is this change still needed?

ariellalgilmore:
> @ariellalgilmore This was only occurring with the `search open onLoad` story and it would autofocus on the search bar, correct? Since the autofocus has been removed, is this change still needed?

This happens on all the masthead stories on mobile when you open the search the cursor flashes for a second in the middle then realigns to the left.

IgnacioBecerra (approved):
Tested out on my phone, works as expected. Thanks @ariellalgilmore!

kennylam:
> This happens on all the masthead stories on mobile when you open the search the cursor flashes for a second in the middle then realigns to the left.

Got it, I see.

PR #6213, opened 2021-05-26
fix(card-static): change background color on hover

IgnacioBecerra:
### Related Ticket(s)
None.

### Description
Allow `background-color` to change upon hover if light mode is active and Card is clickable.

### Changelog

**Changed**

- changed hover styles


photodow:
Looks like it's missing focus and transitions. Not only the light variation, but the default version as well.

May-26-2021 17-05-16

IgnacioBecerra:
Reverted outline changes as discussed offline, will have them ready for follow-up PR after confirmation on the focus behavior on affected components.

packages/styles/scss/components/card/index.scss
```diff
@@ -133,14 +134,27 @@
...
-        margin-bottom: 1rem;
+        margin-bottom: $spacing-05;
       }
     }
     &[color-scheme='light'] {
       &,
       &:hover {
         .#{$prefix}--card__wrapper {
           background-color: $ui-02;
+          transition: $duration--moderate-01 motion(standard, productive);
```
photodow, line 145: Am I seeing this right? Looks like at least on the default color scheme the transition only applies to on mouse out and not mouse in.

May-27-2021 10-49-11

IgnacioBecerra: Needed to add the transition property to the wrapper class as well, pushed the changes 👍

photodow (commented):
That looks better. :+1:

photodow (approved):

Issue #6218, opened 2021-05-27
[Video player] Web Components: Video player configured for inline mode will still open a lightbox


ccwilson-us-ibm-com:
## Detailed description

> Describe in detail the issue you're having.

When a `<dds-lightbox-video-player-container></dds-lightbox-video-player-container>` is included on a page, clicking any `dds-video-player-container` will open the lightbox modal even if mode is configured as `inline`.  This means that any page that includes both `inline` and `lightbox` mode video players with open the lightbox when any of those videos are clicked.

Either the lightbox component isn't ignoring a video player event when the mode is inline.  ...Or the video player is dispatching an event it shouldn't when the mode is inline.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

> Is this issue related to a specific component?

The video player and lightbox video player components.

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Expected: Inline configured video players should play inline (not in the lightbox modal)

Actual: The lightbox modal opens for inline videos when `<dds-lightbox-video-player-container></dds-lightbox-video-player-container>` (Which might be necessary for other lightbox videos on a page.)

> What browser are you working in?

Chrome; Firefox

> What version of Carbon for IBM.com are you using?

Laster master branch.

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

AEM

## Steps to reproduce the issue

The following markup in storybook should replicate the issue:
```
<dds-video-player-container playing-mode="inline" video-id="1_9h94wo6b"></dds-video-player-container>
<dds-lightbox-video-player-container></dds-lightbox-video-player-container>
```

Issue #6217, opened 2021-05-27
[Video player] Web Components: Lightbox mode for video player only supports a single video on a page


ccwilson-us-ibm-com:
## Detailed description

> Describe in detail the issue you're having.

Bug occurs when rendering two instances of the video player container web component in lightbox playing mode with different video ids.  Clicking on one video will open the lightbox modal and play the correct video as expected.  Clicking on the second video will open the lightbox model, but only the first video clicked will play in the modal.  The second video id (or any video id beyond the first one clicked) will not play.

The issue revolves around this line: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/packages/web-components/src/components/lightbox-media-viewer/lightbox-video-player-composite.ts#L77

This if conditional will prevent a different video id from being loaded in the lightbox video player.  And I haven't found another line of code that clears the original video id.  So only the first video loaded will play in the lightbox and no other after that.

Should there be more than one `dds-lightbox-video-player-container` to handle more than one `dds-video-player-container` in lightbox mode?

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

A bug. (maybe)

> Is this issue related to a specific component?

Video player and lightbox video player components.

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Expected: Multiple video players in lightbox mode can be used on a page.  Each video player will load the video associated with the video id attribute.

Currently: Only the first video clicked will play the correct video.  Clicking any other video player will only play that same first video regardless of the video id attribute.

> What browser are you working in?

Chrome, Firefox

> What version of Carbon for IBM.com are you using?

Latest master branch.

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

AEM

## Steps to reproduce the issue

The following markup in storybook should replicate the behavior:
```
<dds-video-player-container video-id="1_9h94wo6b" aspect-ratio=${aspectRatio} playing-mode="lightbox">
</dds-video-player-container>
<dds-video-player-container video-id=1_w5drm2rz aspect-ratio=${aspectRatio} playing-mode="lightbox">
</dds-video-player-container>
<dds-lightbox-video-player-container></dds-lightbox-video-player-container>
```

RobertaJHahn:
@ccwilson-us-ibm-com  Can you provide a link to the page where you have implemented so that we can reproduce the issue.  Thanks.

virtuakazib:
[...]

PR #6226, opened 2021-06-01
fix(lightbox-video-player): update videoID and only open if playing mode is lightbox

ariellalgilmore:
### Related Ticket(s)

#6217 and #6218

### Description

fix lightbox video player so that it updates the videoID if a new video is played and only opens the lightbox video player if `playing-mode = lightbox`
### Changelog

**Changed**

- added additional check to see if `playing-mode` is `lightbox`
- always update videoID


andysherman2121 (approved):
LGTM!

PR #6235, opened 2021-06-02
fix(utilities): use util package for deprecate instead of relative path

annawen1:
### Related Ticket(s)

Refs https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6236

### Description

use the utilities package instead of relative path

### Changelog

**New**

- {{new thing}}

**Changed**

- {{changed thing}}

**Removed**

- {{removed thing}}


jeffchew (approved):
thank you @annawen1 !

kennylam (approved):

Issue #6254, opened 2021-06-03
Multiple lightboxes appear when using both the VideoPlayerContainer and VideoCTAContainer


ccwilson-us-ibm-com:
## Detailed description

> Describe in detail the issue you're having.

When using `<dds-video-player-container mode="lightbox">` is the same page with `<dds-video-cta-container`> there are two issues depending on the proper usage of these two components.

Scenario 1: The page has a high level `<dds-video-cta-container>` to catch and handle all video cta/links in the page.  And also contains multiple `<dds-video-player-container mode="lightbox">`.

Example:
```
<dds-video-cta-container>
    <dds-text-cta cta-type="video" href="1_9h94wo6b">
        test 1
    </dds-text-cta>
    <dds-video-player-container video-id="1_9h94wo6b" aspect-ratio="4x3" playing-mode="lightbox"> </dds-video-player-container> 
    <dds-video-player-container video-id="1_w5drm2rz" aspect-ratio="4x3" playing-mode="lightbox"> </dds-video-player-container> 
  </dds-video-cta-container>
```

Issue: The first VideoPlayerContainer element clicked will properly play the video in a lightbox, but clicking any other VideoPlayerContainer on the page won't play.

Scenario 2: Each video cta/link in a page has its own `<dds-video-cta-container>` wrapper.  And each video player container has a corresponding `<dds-lightbox-video-player-container>` with the matching `video-id` attributes:

Example:
```
<dds-video-cta-container>
    <dds-text-cta cta-type="video" href="1_w5drm2rz">
         test 2
   </dds-text-cta>
</dds-video-cta-container>
<dds-video-player-container video-id="1_9h94wo6b" aspect-ratio="4x3" playing-mode="lightbox"> </dds-video-player-container> 
<dds-video-player-container video-id="1_w5drm2rz" aspect-ratio="4x3" playing-mode="lightbox"> </dds-video-player-container> 
<dds-lightbox-video-player-container video-id="1_9h94wo6b"></dds-lightbox-video-player-container> 
<dds-lightbox-video-player-container video-id="1_w5drm2rz"></dds-lightbox-video-player-container>
```

Issue: All the videos will open a lightbox and play.  But some video clicks cause multiple lightboxes to open at the same time.

I'm including both of these scenarios because I'm not sure which is the intended usage when both video cta/links and a lightbox video player are on the same page.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

Bug

[...]

PR #6260, opened 2021-06-03
fix(lightbox-video-player): add check for video cta specific modal to open and prevent multiple modals from opening

annawen1:
### Related Ticket(s)

Multiple lightboxes appear when using both the VideoPlayerContainer and VideoCTAContainer #6254

### Description

Issue 1:
`video-cta-container` renders the `lightbox-video-player-composite` component and triggers the opening of the modal itself. When placed with `video-player-container` that requires `lightbox-video-player-container` to be applied as well, the custom event triggers both lightbox modals (one for the `video-player-container` and one for `video-cta-container`) to open.

Solution is to pass a `video-cta-lightbox` attribute to `lightbox-video-player-composite` within `video-cta-composite` and check whether to open the lightbox modal based off the attribute. 

Issue 2:
When having multiple video players that open to the lightbox, after opening each one and closing if user opens the first video again,  the video is that of the previously closed.

Solution is to set the hidden state to previously closed videos. Looks like there was code to do this in the `video-player-composite` that wasn't working properly. That logic was tweaked and moved to the `lightbox-video-player-composite` instead.

```
<dds-video-cta-container>
   <dds-text-cta cta-type="video" href="1_8ij4i1qy">
        test 2
    </dds-text-cta>
</dds-video-cta-container>
<dds-video-player-container video-id="1_9h94wo6b" aspect-ratio="4x3" playing-mode="lightbox"> </dds-video-player-container>
<dds-video-player-container video-id="1_w5drm2rz" aspect-ratio="4x3" playing-mode="lightbox"> </dds-video-player-container>
<dds-lightbox-video-player-container></dds-lightbox-video-player-container>
```

### Changelog

**New**

- pass `video-cta-lightbox` attribute to `lightbox-video-player-composite` to identify video-cta specific lightbox modal

**Changed**

- place hiding video logic into `lightbox-video-box-composite` instead

**Removed**

- hiding video logic from `video-player-composite`


jeffchew (approved):
LGTM!

PR #6285, opened 2021-06-07
chore(cloud-masthead): update styles flex value

kennylam:
### Description

Removes shorthand flex value causing build warnings.

image

### Changelog

**Changed**

- `start` to `flex-start`


jeffchew (approved):
Thanks @kennylam

PR #6310, opened 2021-06-08
fix(storybook): adjust dotcomshell and masthead knobs for platform

annawen1:
### Related Ticket(s)

### Description

Add `platform` knob to DotcomShell story - this will prevent us from seeing the `[Object Object]` bug observed when going from L1 story to default. 

Removed the `L1PlatfromData` from the Masthead/DotcomShell L1 story as it is already pulling that data from the data file. There is no need to pass in `platform` and `platform-url` attributes to that particular story as those attributes will not be used. 

### Changelog

**New**

- `platform` knob in DotcomShell WC story, to toggle showing platform name

**Removed**

- unnecessarily passing in the `platform` attributes from Masthead/DotcomShell L1 story as it is already rendering the L1 platform from data file.


kennylam (approved):
LGTM thanks @annawen1!

PR #6331, opened 2021-06-09
fix(masthead-l1): match data structure to react and reuse general `renderNavItems` function

annawen1:
### Related Ticket(s)

Web Component: Change Mega Menu to include animation (back button & forward motion) #4831

### Description

React version data structure for L1 menu items is the same as L0 menu items from translation API. This should match expected structure of L1 in web-components.

Use the same structure and remove the L1 specific left nav item function, instead use the same function used to render the L0 top nav and left nav items.

Use same custom data as React storybook example

### Changelog

**Changed**

- use custom data from React example.
- edit `MastheadL1` type in the services-store translateAPI to use `MastheadLink[]` as the data structure shouldn't differ

**Removed**

- `renderL1Items()` and ulitize `renderNavItems()` instead as the structure of the L1 data and L0 data is the same

### DIFF
--- a/packages/services-store/src/types/translateAPI.ts
+++ b/packages/services-store/src/types/translateAPI.ts
@@ -87,17 +87,7 @@ export interface MastheadLink {
 export interface MastheadL1 {
   title: string;
   url?: string;
-  menuItems?: MastheadL1Item[];
-}
-/**
- * An item for MastheadL1
- */
-
-export interface MastheadL1Item {
-  title: string;
-  titleEnglish?: string;
-  url?: string;
-  menuItems?: MastheadL1Item[];
+  menuItems?: MastheadLink[];
 }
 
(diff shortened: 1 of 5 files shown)


ariellalgilmore:
looks good! just small thing looks like the color went back to the darker color just for the L1 in web components

annawen1:
> looks good! just small thing looks like the color went back to the darker color just for the L1 in web components

weird I don't see the difference when comparing the wc canary and the deploy preview...

WEB COMPONENTS CANARY:
Screen Shot 2021-06-09 at 6 14 25 PM

DEPLOY PREVIEW:
Screen Shot 2021-06-09 at 6 15 03 PM

ariellalgilmore (approved):
LGTM! not sure why percy looks different

Issue #5425, opened 2021-03-09
Styles Package: Update spacing above and below the [content group] 


RobertaJHahn:
#### User Story

> As a `[user role below]`:
IBM.com Library developer

> I need to:
create/change the `Content group`

> so that I can:
provide the ibm.com adopter developers components they can use to build ibm.com web pages

#### Additional information

- Update the **Content group** to fix it globally and it will cascade to the components that use content group
- change the spacing above and below from 96 to 64 [layout 05] for breakpoint large and above
- [layout spacing design background](https://app.zenhub.com/workspaces/carbon-for-ibmcom-5d449f3642eb1962336cbe52/issues/carbon-design-system/carbon-for-ibm-dotcom/5376)
- **Design analysis details are also included in issue (#5376 )**
- **Storybook bugs are also included in issue (#5667 )**
- Visual QA testing issue (#5727)
- Prod QA testing issue (#5226)

#### Summary of changes:  
- Update spacing below section headers from 48px to 32px.
- Update 2nd-tier spacing from 96px to 64px.

### B. **Content Groups** between heading-04 and body copy
See green arrow for emphasis
![Screen Shot 2021-03-05 at 5 02 45 PM](https://user-images.githubusercontent.com/45692622/110179145-8eaadb00-7dd5-11eb-8209-1fb022d0695d.png)

List of components:
*I believe all content groups are already 32px?* 

### B. Spacing between two Content Groups
See green arrow for emphasis
<img width="834" alt="Screen Shot 2021-03-04 at 5 22 18 PM" src="https://user-images.githubusercontent.com/45692622/110039281-fcd89a80-7d0e-11eb-9f58-ba1f966c4592.png">

List of components:
- [ ] Content group cards (top spacing)
- [ ] Content group simple (default, with image, with video) (top spacing)
- [ ] Content block media (with aside elements)
- [ ] Content block mixed (with aside elements)
- [ ] Content block segmented (default, with video, aside elements)

#### Acceptance criteria
- [ ] Include README for the web component and corresponding styles
- [ ] Create Web Components styles in styles package
- [ ] Do not create knobs in Storybook that include JSON objects
- [ ] Break out Storybook stories into multiple variation stories, if applicable
- [ ] Create codesandbox example under `/packages/web-components/examples/codesandbox` and include in README
- [ ] Minimum 80% unit test coverage
- [ ] **engineer:  visually inspect all other components using content group**
[...]

PR #6337, opened 2021-06-10
fix(content-group): spacing above and below

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5425

### Description

Update 2nd-tier spacing from 96px to 64px.


shixiedesign:
Hey Julio, this issue is a bit confusing. Lemme try to clarify: 

#### The idea is to change this current spacing (64 top, 64 bottom): 

image

#### To this [spacing specified here](https://www.ibm.com/standards/web/carbon-for-ibm-dotcom/guidelines/type-pairing-and-spacing#content-group) **(32 top, 48 bottom)**: 

image

## So this PR haven't addressed the problem: 

Example, `Content group pictogram` margin top didn't change but is expected to reduce to 32px. Margin bottom disappeared completely but is expected to 48px.

image
image

NSDrowned:
@shixiedesign ohhh, ok, I understand. I thought it was 64px all around. Let me fix that.

NSDrowned:
Hello @shixiedesign, I have updated all margins for all components related to `content-block` and `content-group`. I removed any paddings the components had so they will all be ruled by the same set of margins (hope that's ok). I also used `spacing-07` for both top and bottom margins in `sm` like the table pointed out.

The only story that doesn't appear to have the same spacing is `content-group-pictograms` because it comes with extra padding caused by this class (not related to the component):

image

Let me know if you want it gone ⚰️

shixiedesign:
Discussing on Slack with Julio... more soon

NSDrowned:
@shixiedesign updates are up 👍

shixiedesign:
All looks good except the `content-group-pictograms` in Web components. It's probably what you described in the previous comment? I think we just need to get rid of the extra spacing from the demo class and we should be good 👍 

image

NSDrowned:
@shixiedesign removed the padding! 👍

shixiedesign (approved):
Awesome! Thank you!

jeffchew (approved):
Thanks @NSDrowned !

Issue #6016, opened 2021-05-10
[card, cardgroup] enable border to allow gradient and offset items


photodow:
Follow onto #5962 and #5993 

### The problem
Need the ability to add border gradient applied to card group cards in between cards, beginning in one corner to the other corner without breaking or starting over. Also needs to skip over blank/empty/offset columns.

- border images don't work great because we need them in the grid gutter between cards.
- things like box shadows don't allow you to add gradients properly.
- needs to be applied to the card group as a whole so it has a consistent beginning to end flow.
- needs to also be applied to a single card for certain situations.

### Possible solution
- See the prototype for how I approached it. There maybe a better way, but at least this will kind of demonstrate what we're looking for.
- Empty cards for offsetting
- Border using padding on card
- linear background applied to the card, and card group depending on the effect we're going for.
- Gutters in-between two offsets shouldn't exist both horizontally, and vertically
- Empty items don't have borders

![image](https://user-images.githubusercontent.com/3793636/117724295-2ac3da80-b1a9-11eb-8239-6098b20c5b4b.png)

### Additional information
 - See a prototype exploring some of the different approaches. https://codepen.io/photodow/full/MWpYKmL

RobertaJHahn:
@ariellalgilmore Can you take a look at this work and decide what the sizing would be.  Would like to know if this work can be done by the code freeze June 25.

Issue #5868, opened 2021-04-22
[Card group] Change to same height util to apply to whole group


shixiedesign:
After review, design decided that same height util should not be applied per row. It creates the impression that some content are associated somehow because they in bigger cards. This is a feature request. Sorry I know Ignacio recently fixed it. 

### Details of the change

- Change `same height util to apply` to whole group, so cards in the group always the same size.
- Same height util should turn OFF on `sm` breakpoint where cards are 1 column stacked

PR #6367, opened 2021-06-14
feat(card-group): add option for 2-4 cards per row

kennylam:
### Related Ticket(s)

#5286 #5520 

### Description

This adds the option to set the number of cards per row (2-4 cards) in `CardGroup`/`card-group`. The default number is 3.

### Changelog

**New**

- users can now set the number of cards per row in `CardGroup`/`card-group`


IgnacioBecerra:
@kennylam So far looking good! I found a couple issues in both versions:

In Web Components, when choosing 4 cards per row and enabling the outline cards knob, it appears that there's a card that's missing the bottom border in the first row:
Screen Shot 2021-06-15 at 10 14 17 AM


The same border bug can also be seen In React with the 4 cards per row, but the footer seems to be clipping into the copy as well:
Screen Shot 2021-06-15 at 10 09 31 AM


Also in the first picture the weird copy height might be caused from the sameHeight utility being applied to each row individually, and the logic for that takes 2/3 cards per row into account. I know #5868 will be fixing this, so not sure if we sure be updating the individual sameHeight logic to consider 4 or let it go for now since it'll get removed anyway.

kennylam:
>I found a couple issues in both versions

@IgnacioBecerra Thanks for looking, yeah I'm aware of these issues as well. The React clipping is caused by the `sameHeight` utility not firing a second time when the card number is changed. This is a common problem in Storybook and one of the reasons we sometimes need to break out variations like this into their own story. I'll take a look at the border issue since this was addressed in a separate PR. If it is too far out of the scope of this I'm going to create an issue and track separately.

kennylam:
@IgnacioBecerra looks like there's already an issue (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6016) open for the borders.

shixiedesign:
Hey @kennylam this is working and good, one challenge is that because the “number of cards per row” is unaware of the grid, changing the values causes cards in the grid to squeeze and stretch. How would adopter achieve the 2x2 layout without having the cards stretch to 5 columns each? 
image

kennylam:
@shixiedesign Grid layout has been adjusted for when there are two cards per column.

shixiedesign:
Thank you!! 

1. Border on 2nd card became a bit broken. Issue persists if I open the canvas in new tab. 
image

2. If we display Card group starting first column of the grid, would we avoid seeing the squished cards issue? If so, It's totally fine to change the display to default starting in column 1
3. Seeing the gray empty card slot sometimes
4. Same height util might need an update? 
image

kennylam:
@shixiedesign Good catch! Had to update the component with the outline work from #6435.

shixiedesign:
1. Outline issue is resolved! 🎉 
2. How about this one? For with it's 4 cards per row, re-align storybook story to start on column 1 instead of column 5 could help?  

> If we display Card group starting first column of the grid, would we avoid seeing the squished cards issue? If so, It's totally fine to change the display to default starting in column 1

3. Try 5 cards, 3 per row, 0 offset. I see the card's body copy are not starting at the same height: 
image

kennylam:
@shixiedesign #2 and 3 done!

shixiedesign:
💣 👍

shixiedesign (approved):

annawen1 (approved):
LGTM!

PR #6385, opened 2021-06-16
docs(cdn): adding reusable CDN doc partials for web components

jeffchew:
### Related Ticket(s)

No related issue

### Description

The motivation for creating these as partials is that there was a lot of
 repeated content that made more sense to have it centralized. In
 addition, the specific version is now set based on the current
 `package.json` version. While this would work well for the production
 output of Storybook, this would generate broken links for the `next`
 versions, but I think is probably acceptable since our adopters largely
  are only looking at production.

### Changelog

**Changed**

- Updated CDN docs to use a centralized set of partials

**Removed**

- `plex.md`


ariellalgilmore (approved):
looks great!!! thanks @jeffchew

PR #6408, opened 2021-06-17
docs(cdn): updated to point to the new cdn urls

jeffchew:
### Related Ticket(s)

No related issue

### Description

After consideration, have decided to add more specificity to the CDN urls for tag releases. This is in anticipation of any future major releases of `Carbon for IBM.com`. URLs are now:

tag version:
```https://1.www.s81c.com/common/carbon-for-ibm-dotcom/tag/v1/[tag name]/[component name].min.js```

specific version:
```https://1.www.s81c.com/common/carbon-for-ibm-dotcom/version/[version number]/[component name].min.js```

Note, I already kicked off devops jobs to fill in the tag releases with at least the latest version on master, but should be populated with the proper tag releases once those are scheduled to auto-run.

### Changelog

**Changed**

- Multiple documentation for CDN urls

### DIFF
--- a/packages/web-components/README.md
+++ b/packages/web-components/README.md
@@ -171,7 +171,7 @@ Here is an example of implementing the `dotcom-shell`:
 <html>
   <head>
     <script type="module">
-      import 'https://1.www.s81c.com/common/carbon-for-ibm-dotcom/latest/dotcom-shell.min.js';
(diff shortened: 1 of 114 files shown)


IgnacioBecerra (approved):
Thanks @jeffchew!!

PR #6416, opened 2021-06-18
feat(cloud-masthead): HC-1993 Expose a new authentication checking method for the docs/console team

huntermacd:
### Related Ticket(s)

https://jsw.ibm.com/browse/HC-1993

### Description

The goal of this PR is to expand the utility for checking the Cloud login cookie to also provide an option for checking against the console API. Additionally, a prop should be added to the cloud-masthead-container component that allows for selecting between the two auth methods.

### Changelog

**New**

- added cloudAccountAuth service, which encompasses both cookie and api auth checking methods
- added various files for actions, types, reducers, and tests for new service

**Changed**

- adds prop for selecting auth method, updates cloud-masthead components to hook into new service

**Removed**

- removed cloudLoginStatusCookie files since that functionality has been moved into new service


huntermacd:
@annawen1 @jeffchew @proeung 

Posting this work in progress in order to get some guidance. I've expanded the cloud login utility to include an option for checking against the console API instead of using the cookie method. The goal is to provide a prop to cloud-masthead-container that allows for selecting between the two options. I'm not certain how to provide this prop and could use some help. Please look over what I have so far and advise.

huntermacd:
@annawen1 @jeffchew @proeung @asudoh

Ok, I've added the prop we need. Curious how complex this needs to get with regard to the new auth API method, being an async operation. Should this "utility" instead be wrapped into a larger "service" a la the ProfileAPI and tied into the store as that is?

annawen1:
> @annawen1 @jeffchew @proeung
> 
> Ok, I've added the prop we need. Curious how complex this needs to get with regard to the new auth API method, being an async operation. Should this "utility" instead be wrapped into a larger "service" a la the ProfileAPI and tied into the store as that is?

Yeah, I think we should follow with what we've done for the global masthead. Maybe set up a service specifically for cloud console for now, and then tie that into the store.

packages/web-components/src/components/masthead/__stories__/cloud-masthead.stories.ts
```diff
@@ -62,6 +62,7 @@ export const Default = !DDS_CLOUD_MASTHEAD
                 platform-url="${ifNonNull(platformData.url)}"
                 selected-menu-item="${ifNonNull(selectedMenuItem)}"
                 ?has-contact="${hasContact}"
+                auth-method="cookie"
```
proeung, line 65: Thanks for adding this new attribute/prop `auth-method` and made it available in the `cloud-masthead-container`. This is an enhancement, but I think it'll be nice to expose these Authentication method options as a knob select within the storybook preview.

Also, we should note this attribute/prop in the `README.stories.mdx` file as well.

proeung (commented):
Looks good overall! I just have a minor comment about documentation and exposing the `auth-method` in the storybook knobs.

@gptt916 Can you take a look at this PR and let us know if you have any questions.

packages/utilities/src/utilities/cloudAccountAuthentication/cloudAccountAuthentication.js
```diff
@@ -0,0 +1,70 @@
...
+   * @example
+   * import { cloudAccountAuthentication } from '@carbon/ibmdotcom-utilities';
+   *
+   * const status = cloudAccountAuthentication.checkAPI();
+   *
+   * @returns {string} string determining login status
+   */
+  static async checkAPI() {
+    const cloudLogin = await fetch(
+      'https://cloud.ibm.com/api/v6/selected-account?profile=true',
```
gptt916, line 50: Seems like this is hard coded to call production, we have other environments like test.cloud.ibm.com and dev.test.console.cloud.ibm.com. This should be a relative url to make sure those envs work as well.

packages/web-components/src/components/masthead/__stories__/cloud-masthead.stories.ts
huntermacd: Updated the README. I'll need to think more on adding a knob to control auth method, since it currently won't work since the method needs to fire on load.

huntermacd:
Ok. Ready for proper review.

packages/web-components/src/components/masthead/__stories__/cloud-masthead.stories.ts
proeung: @huntermacd No worries! The auth method knob is just a fancy addition, I don't think it's a blocker for this PR. As long as we have this attribute accounted for in the dev README, I'm fine with it. 👍

proeung:
@huntermacd Thanks for the update!

@annawen1 and @gptt916, can you both help review this PR? Thanks!

gptt916 (approved):

annawen1 (approved):
looks good!

proeung (approved):
LGTM as well!

Issue #6110, opened 2021-05-17
[Content item horizontal - with thumbnail] New variation for the the Content item horizontal


pavolprok:
### The problem
We are currently missing a Content item horizontal component variation with small thumbnail images. We only have the variation with media where the image is too dominant. 

### The solution
I have prepared (together with Lila Title and also with guidance from Shixie) design specs for this new variation with a 4:3 thumbnail image that aligns to the right and takes up 100% of the last 3 columns with max. width of 256px.
https://ibm.box.com/s/bvpjhnwwpsi8uo51s1993v5ywbpakxcf

This variation is currently being used on the Learn ecosystem prototype. 

![image](https://user-images.githubusercontent.com/70329414/118513102-a8509300-b733-11eb-97cc-4ccf8b7b5880.png)

### Additional information

Design specs:

https://ibm.ent.box.com/s/bvpjhnwwpsi8uo51s1993v5ywbpakxcf

Functional specs:

https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/wiki/Content-item-horizontal

![content-item-horizontal--thumbnail](https://user-images.githubusercontent.com/3793636/121090652-85535500-c7ae-11eb-8bb8-4b329b9897cb.gif)

----
This component will be developed by a Hybrid Cloud engineer.  

- Corresponding Jira issue:  https://jsw.ibm.com/browse/HC-1969

RobertaJHahn:
AEM will develop this component behind a feature flag and contribute it.  Design specs are finished.  

The topic of maintaining the DDS experimental components to the May 18 Team Huddle.

RobertaJHahn:
Met with Lily, Lila and Sandeep.  They asked whether James can create the functional specs for this component.  AEM will still develop it.

RobertaJHahn:
Will be developed as a component in the Carbon for ibm.com Library:  Needs epic, func specs, DDS dev consulting, (AEM developer), Visual QA (AEM designer), Prod QA

proeung:
@RobertaJHahn Here's the JIRA ticket for this component that we're (HC) tracking on our end - https://jsw.ibm.com/browse/HC-1969.

RobertaJHahn:
Epics do not belong in the Review pipeline.  Moved back to the Prioritized epics pipeline.

PR #6425, opened 2021-06-20
feat(content-item-horizontal-thumbnail): add thumbnail variation

mwanberg:
### Related Ticket(s)

DDS ticket: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6244
HC Jira ticket: https://jsw.ibm.com/browse/HC-1969

### Description

This PR adds a new variation to the "Content item horizontal" web component. It is very similar to the default pattern except: 

1) The eyebrow is removed
2) A smaller image has been added
3) On lg+ screens the layout goes to three columns, with the thumbnail on the right side.

Screen Shot 2021-06-19 at 8 46 10 PM

### Changelog

**New**

- N/A

**Changed**

- Added a "With Thumbnail" variation to `<dds-content-item-horizontal>`
- Small width style fix that affects the React `<dds-content-item-horizontal>` component

**Removed**

- N/A

### React component style fix

The current default variation for both the Web Component and React  `<dds-content-item-horizontal>` component has a small mis-alignment on md+ breakpoints. This PR introduces a fix for that, so we will see a small horizontal visual diff for the both the Web Component and React default variation.

**Current:**

Screen Shot 2021-06-24 at 4 45 06 PM

**Fixed:**

Screen Shot 2021-06-24 at 4 46 34 PM


packages/styles/scss/components/content-item-horizontal-thumbnail/_content-item-horizontal-thumbnail.scss
```diff
@@ -0,0 +1,159 @@
+/**
```
annawen1, line 1: don't think we need to have a separate `scss` file in the styles package, can you move this to the `scss` file you created within the `content-item-horizontal` web-components folder?

packages/styles/scss/components/content-item-horizontal-thumbnail/_content-item-horizontal-thumbnail.scss
```diff
@@ -0,0 +1,159 @@
...
+ * Copyright IBM Corp. 2016, 2021
```
annawen1, line 2: don't think we need to have a separate scss file in the styles package, can you move this to the scss file you created within the content-item-horizontal web-components folder? It's a little confusing to have 2 different files.

packages/web-components/src/components/content-item-horizontal/__stories__/README.stories.mdx
```diff
@@ -118,6 +118,34 @@ import '@carbon/ibmdotcom-web-components/es/components/content-item-horizontal/c
 </dds-content-item-horizontal-media>
 ```
 
+## With Thumbnail
+
+### JS
+
+```javascript
+import '@carbon/ibmdotcom-web-components/es/components/content-item-horizontal/index.js';
+import '@carbon/ibmdotcom-web-components/es/components/content-item-horizontal/content-item-horizontal-thumbnail';
```
annawen1, line 127: since you've already import `content-item-horizontal-thumbnail` in `content-item-horizontal/index.js`, I don't think we need to add this import statement again in the readme.

packages/styles/scss/components/content-item-horizontal-thumbnail/_content-item-horizontal-thumbnail.scss
```diff
@@ -0,0 +1,159 @@
...
+      padding-top: $carbon--spacing-07;
+      padding-bottom: $carbon--layout-05;
+    }
+
+    padding-left: $carbon--spacing-05;
+    padding-right: $carbon--spacing-05;
+  }
+
+  :host(#{$dds-prefix}-content-item-horizontal) ::slotted([slot='heading']),
+  :host(#{$dds-prefix}-content-item-horizontal-copy) ::slotted(:not([slot])),
```
annawen1, line 40: Haven't seen this `::slotted(:not([slot]))`, can you tell me what this selector is targeting?

packages/web-components/src/components/content-item-horizontal/content-item-horizontal-thumbnail.scss
```diff
@@ -0,0 +1,148 @@
...
+
+  :host(#{$dds-prefix}-content-item-horizontal-thumbnail-copy) ::slotted(:not([slot])) {
+    margin-bottom: $carbon--spacing-07;
+
+    p {
+      color: $text-01;
+    }
+  }
+
+  :host(#{$dds-prefix}-link-list) ::slotted([slot='footer']) {
```
annawen1, line 133: I think this may be changing the layout of components using this element, can you contain it to just the `content-item-horizontal-thumbnail`?

packages/web-components/src/components/content-item-horizontal/content-item-horizontal-thumbnail.scss
```diff
@@ -0,0 +1,148 @@
...
+  .#{$prefix}--content-item-horizontal-thumbnail__content-wrapper {
+    @include carbon--breakpoint(md) {
+      padding-right: $carbon--spacing-07;
+    }
+    @include carbon--breakpoint(lg) {
+      @include carbon--make-col(4, 8);
+    }
+  }
+
+  .#{$prefix}--content-item-horizontal-thumbnail__heading-wrapper {
```
annawen1, line 98: instead of using a wrapper class, could you target the heading with the `:host(#{$dds-prefix}-content-item-horizontal-thumbnail) ::slotted([slot='heading'])` selector that you have below?

annawen1 (commented):
Thanks @mwanberg for getting this set up! Have a few comments, but otherwise it's looking good!

packages/web-components/src/components/content-item-horizontal/content-item-horizontal-thumbnail.ts
```diff
@@ -0,0 +1,53 @@
...
+const { prefix } = settings;
+const { stablePrefix: ddsPrefix } = ddsSettings;
+
+/**
+ * A component to present content in a horizontal orientation with thumbnail.
+ *
+ * @element dds-content-item-horizontal-thumbnail
+ */
+@customElement(`${ddsPrefix}-content-item-horizontal-thumbnail`)
+class DDSContentItemHorizontalThumbnail extends DDSContentItem {
```
proeung, line 25: I don't think this new component `<dds-content-item-horizontal-thumbnail>` is needed for the work outlined in this issue (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6110). It seems to me like we just need to include the [dds-mage](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/packages/web-components/src/components/image/image.ts#L26) component within the [dds-content-item-horizontal](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/packages/web-components/src/components/content-item-horizontal/content-item-horizontal.ts#L25) component and defining the `image` slot within the HTML return of the shadow dom, unless there's something else that I'm missing here.

@annawen1 @photodow Can you let me know if what I'm suggesting is in alignment with what you guys are thinking as well?

https://ibm.ent.box.com/file/812734380200

packages/web-components/src/components/content-item-horizontal/content-item-horizontal-thumbnail-copy.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const { stablePrefix: ddsPrefix } = ddsSettings;
+
+/**
+ * The copy content of content item horizontal with thumbnail.
+ *
+ * @element dds-content-item-horizontal-thumbnail-copy
+ */
+@customElement(`${ddsPrefix}-content-item-horizontal-thumbnail-copy`)
+class DDSContentItemHorizontalThumbnailCopy extends StableSelectorMixin(DDSContentItemCopy) {
```
proeung, line 24: I don't think this component is needed as the `dds-content-item-horizontal-copy` has already existed and can be used within the variant of `content-item-horizontal` layout component with thumbnail. https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/packages/web-components/src/components/content-item-horizontal/content-item-horizontal-copy.ts

Similar comment as noted here - https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/6425/files#diff-962566a3c5ebb5bad8f6364e5cdadebedeb244446091770dc9b678a89b3f2743R25

proeung (changes requested):
@mwanberg This is a great first draft! You've definitely have put a lot of work and effort into this enhancement ticket, however, there are some implementation questions that I have surrounding extending some components to showcase a variant with a thumbnail and video player. I'll defer this to @annawen1 as perhaps there might be a good reason why extending these components ( eg. `dds-content-item-horizontal-thumbnail`, `dds-content-item-horizontal-thumbnail-copy`) are warranted.

packages/web-components/src/components/content-item-horizontal/content-item-horizontal-thumbnail.ts
annawen1: @proeung that's a good point, from the functional specs I think we could work off of `content-item-horizontal` itself and just add an attribute to render the `thumbnail` styles

packages/styles/scss/components/content-item-horizontal-thumbnail/_content-item-horizontal-thumbnail.scss
mwanberg: @annawen1 I was using the `<dds-content-item-horizontal-media>` component/variation as an example, which had a setup similar to what I created. But I can move these styles over as you suggest!

mwanberg: Also, I totally did make them duplicate. 🙈

packages/styles/scss/components/content-item-horizontal-thumbnail/_content-item-horizontal-thumbnail.scss
mwanberg: @annawen1 TBH that's a new one to me, too. I was once again copying what I was seeing already existing, e.g. https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/0d0571e43b1732086681ac59f1f7193765b9d83b/packages/styles/scss/components/content-item-horizontal-media/_content-item-horizontal-media.scss#L30

mwanberg:
@annawen1 @proeung Based on your suggestions, I refactored, editing the default component instead of creating a new one. Please let me know if I'm on the right track.

I made changes directly to the original content item horizontal component, so I'm guessing we should check for any regressions. There will likely be one visible change, which is a slight adjustment in how wide the whole component is. There was a bug in the original code that made the width slightly less than what it should be. There should be no other visual changes for the default variation.

Because I needed to modify the default markup to accommodate the new thumbnail image and slightly different layout and margins of the new variation, there is some moderately significant changes in both the markup structure and the styles. I tried to keep things simple while changing as little of the original as possible, so it's possible the whole thing could use some more streamlining. I'm open to any observations or suggestions.

One problem I ran into is that I found it difficult to target the copy and the CTAs in the styling. I am still working to figure that out.

A few questions:
- Do I still need to do the new feature flag?
- Should I be testing this in RTL?
- Should I be writing any new tests?

pavolprok:
Hi there. Thanks for putting this component together. I had a look at it and found some issues:
1. On max, xlg, lg breakpoints the component starts on the 4th column from the left but it is suppose to start on the 5th column from the left. The 4:3 thumbnail mage is suppose to be placed on the last 3 right columns. 
Screen Shot 2021-06-21 at 17 39 07

2. On the small breakpoint the 4:3 image is suppose to be left aligned and span max 3 columns.
Screen Shot 2021-06-21 at 17 57 40


3. I do not see the possibility to change the copy text in the knobs. There is only the possibility to change headline and CTAs.
Screen Shot 2021-06-21 at 17 54 58

annawen1:
@mwanberg 

> There will likely be one visible change, which is a slight adjustment in how wide the whole component is. There was a bug in the original code that made the width slightly less than what it should be. There should be no other visual changes for the default variation.

That's a good catch!

> One problem I ran into is that I found it difficult to target the copy and the CTAs in the styling. I am still working to figure that out.

Depending on what you're trying to style, you may need to extend a component if necessary. Let me know if you need help with that.

>  A few questions:
> 
> - Do I still need to do the new feature flag?
> - Should I be testing this in RTL?
> - Should I be writing any new tests?

- I don't remember seeing that this needed to be behind a feature flag @shixiedesign can you confirm?
- We haven't actively been testing the components in RTL, but that's definitely something to consider. If you add the `RTL` label to this PR, it should build out a new deploy preview.
- Yes, could you take a look at other component tests we've written and follow those as a guide?

packages/web-components/src/components/content-item-horizontal/__stories__/content-item-horizontal.stories.ts
```diff
@@ -63,6 +64,34 @@ export const Default = ({ parameters }) => {
   `;
 };
 
+export const WithThumbnail = ({ parameters }) => {
+  const { alt, heading, copy, ctaType1, ctaCopy1, href1, ctaType2, ctaCopy2, href2 } =
+    parameters?.props?.ContentItemHorizontal ?? {};
+  return html`
+    <dds-content-item-horizontal thumbnail="true">
```
proeung, line 71: Where is this `thumbnail` attribute being defined? I'm not seeing this prop defined in the `dds-content-item-horizontal` component.

packages/web-components/src/components/content-item-horizontal/content-item-horizontal.ts
```diff
@@ -26,12 +26,17 @@ class DDSContentItemHorizontal extends DDSContentItem {
...
+            <slot name="eyebrow" @slotchange="${this._handleSlotChange}"></slot>
+            <slot name="heading"></slot>
+          </div>
+          <div class="${prefix}--content-item-horizontal__content-wrapper">
+            ${this._renderBody()}${this._renderFooter()}
+          </div>
         </div>
-        <div class="${prefix}--content-item-horizontal__col">
-          ${this._renderBody()}${this._renderFooter()}
+        <div class="${prefix}--content-item-horizontal__col ${prefix}--content-item-horizontal__col--2">
```
proeung, line 38: Thanks for defining a slot for the thumbnail/image in the HTML return. I think we'll need to add a condition check (eg. https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/packages/web-components/src/components/card/card.ts#L130) within the result that adds in the `div` wrapper when the `thumbnail` or `image` slot is defined within the markup of the `<dds-content-item-horizontal>`. @annawen1 Let me know if you agree.

Also, @shixiedesign correct me if I'm wrong, but we'll need to account for the `<dds-video-player>` component that opens up the lightbox media viewer (https://ibmdotcom-web-components.mybluemix.net/?path=/story/components-video-player--with-lightbox-media-viewer)? if not, feel free to disregard my comment.

packages/web-components/src/components/content-item-horizontal/__stories__/README.stories.mdx
```diff
@@ -131,6 +150,7 @@ See [our README](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/t
 | ------------------------------------- | -------------------------------------------------- | ----------- |
 | `<dds-content-item-horizontal>`       | `data-autoid="dds--content-item-horizontal"`       | Component   |
 | `<dds-content-item-horizontal-media>` | `data-autoid="dds--content-item-horizontal-media"` | Component   |
+| `<dds-content-item-horizontal-thumbnail>` | `data-autoid="dds--content-item-horizontal-thumbnail"` | Component   |
```
proeung, line 153: This line can be removed as we're no longer defining a component for `<dds-content-item-horizontal-thumbnail>`.

packages/web-components/src/components/content-item-horizontal/__stories__/README.stories.mdx
```diff
@@ -60,6 +60,25 @@ import '@carbon/ibmdotcom-web-components/es/components/content-item-horizontal/i
...
+    Phasellus at elit sollicitudin.</dds-content-item-horizontal-copy>
+  <dds-link-list slot="footer" type="vertical">
+    <dds-link-list-item-cta icon-placement="right" href="https://www.ibm.com" cta-type="local">
+      Learn more
+    </dds-link-list-item-cta>
+    <dds-link-list-item-cta icon-placement="right" href="https://www.ibm.com" cta-type="external">
+      Microservices and containers
+    </dds-link-list-item-cta>
+  </dds-link-list>
+  <dds-image slot="thumbnail" alt="Alt text" default-src="image.jpg"></dds-image>
```
proeung, line 78:
```suggestion
  <dds-image slot="thumbnail" alt="Alt text" default-src="https://fpoimg.com/960x720?text=4:3&bg_color=ee5396&text_color=161616"></dds-image>
```

Let's use FPO image for easy copy and paste of this WC markup so that adopters can see what this variant will look like instead of defending the URL of the image.

packages/web-components/src/components/content-item-horizontal/__stories__/content-item-horizontal.stories.ts
```diff
@@ -63,6 +64,34 @@ export const Default = ({ parameters }) => {
...
+        </dds-link-list-item-cta>
+        <dds-link-list-item-cta
+          icon-placement="${ICON_PLACEMENT.RIGHT}"
+          href="${ifNonNull(href2)}"
+          cta-type="${ifNonNull(ctaType2)}"
+        >
+          ${ctaCopy2}
+        </dds-link-list-item-cta>
+      </dds-link-list>
+      <dds-image slot="thumbnail" alt="${ifNonNull(alt)}" default-src="${imgMd4x3}"></dds-image>
```
proeung, line 90: @shixiedesign or @pavolprok Do we need to account for media queries for this thumbnail image? I'm not sure if there are different ratios or dimensions that we need to serve on different breakpoints.

proeung (changes requested):
@mwanberg Great revision! Thanks for applying the refactoring work that I've suggested. I just have a few comments, but overall looks good from a code perspective. @annawen1 or @IgnacioBecerra let me know if you guys have other feedback on this PR.

packages/web-components/src/components/content-item-horizontal/content-item-horizontal.ts
annawen1: > Thanks for defining a slot for the thumbnail/image in the HTML return. I think we'll need to add a condition check (eg. https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/packages/web-components/src/components/card/card.ts#L130) within the result that adds in the div wrapper when the thumbnail or image slot is defined within the markup of the <dds-content-item-horizontal>. @annawen1 Let me know if you agree.

yep, same for the `eyebrow` slot. We can check if the `thumbnail` attribute exists/is true to render the `thumbnail` and `eyebrow` slots.

```
{this.thumbnail ? <slot name="eyebrow" @slotchange="${this._handleSlotChange}"></slot> : undefined}
```

annawen1: @pavolprok @shixiedesign want to clarify again if we need to account for `video` as well or will it only be `image` for thumbnail?

pavolprok: @annawen1 the thumbnails will only be images. No videos. ;-)

proeung: @pavolprok Thanks for the confirmation!

mwanberg:
@pavolprok Thank you for reviewing this PR! The branch has been updated with fixes from your feedback.

1. I have adjusted the component so it starts on the 5th column on lg+ and the thumbnail now is aligned right with the last column.
2. The thumbnail now fits inside the first three columns on mobile/sm.
3. After attempting to add a "copy" knob and then looking into it with @annawen1, it seems that there is currently an [open issue](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5648) where some "copy" knobs aren't updating properly. So unfortunately I think we will need to leave it out of this version.

shixiedesign (approved):
This looks good! Thank you for the contribution!

proeung (approved):
@mwanberg Code changes look good overall! Thanks for addressing my feedback!

@annawen1 @shixiedesign I'm approving this PR from a code review perspective, however, let us know if you see anything else that needs adjusting.

shixiedesign:
Hi! Almost approved - noticed a few things got affected: 

Looks like the changes are done to style package, and `content item horizontal -- default` in React is affected: 
**Current:** 
image

**Expecting**
image

mwanberg:
@shixiedesign Good catch! I've modified the styling to fix that regression. One thing that will still be different for both the Web Component and React default variation is a slight adjustment in the width of the component on md+ breakpoints. This fixes an existing bug. Please let me know if you'd like me to undo that fix.

**Current:**

Screen Shot 2021-06-24 at 4 45 06 PM

**Fixed:**

Screen Shot 2021-06-24 at 4 46 34 PM


@proeung and @annawen1 With Anna's direction I updated a little of how the `thumbnail` property is working so now it's a little cleaner AND the eyebrow shows up now on the default variation.

What remains now is writing up tests.

packages/web-components/src/components/content-item-horizontal/__stories__/content-item-horizontal.stories.ts
pavolprok: Hi Putra. There will be only 4:3 ratio for the thumbnail images. The thumbnail images should span across 3 last columns  on max, xlg, lg, md. The only chage it rems of placement of the images is on the small breakpoint where the images spans across 3 columns but is aligned to the left and is placed below the CTA.

pavolprok:
> Link to preview: https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/6425/index.html?path=/story/components-content-item-horizontal--with-thumbnail
> 
> @pavolprok Thank you for reviewing this PR! The branch has been updated with fixes from your feedback.
> 
> 1. I have adjusted the component so it starts on the 5th column on lg+ and the thumbnail now is aligned right with the last column.
> 
> Screen Shot 2021-06-24 at 11 06 38 AM
> 
> 1. The thumbnail now fits inside the first three columns on mobile/sm.
> 
> Screen Shot 2021-06-24 at 11 05 04 AM
> 
> 1. After attempting to add a "copy" knob and then looking into it with @annawen1, it seems that there is currently an [open issue](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5648) where some "copy" knobs aren't updating properly. So unfortunately I think we will need to leave it out of this version.

Thank you for the updates. The component now looks great.

pavolprok (approved):
Thanks for addressing the alignment issues. The component now looks correct.

packages/web-components/src/components/content-item-horizontal/content-item-horizontal.ts
```diff
@@ -23,16 +23,35 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
...
       <div class="${prefix}--content-item-horizontal__row">
-        <div class="${prefix}--content-item-horizontal__col">
-          <slot name="eyebrow" @slotchange="${this._handleSlotChange}"></slot>
-          <slot name="heading"></slot>
-        </div>
-        <div class="${prefix}--content-item-horizontal__col">
-          ${this._renderBody()}${this._renderFooter()}
+        <div class="${prefix}--content-item-horizontal__col--1">
+          <div class="${prefix}--content-item-horizontal__heading-wrapper">
+            ${this.thumbnai
```
annawen1, line 37: missing the `l` in `thumbnail` here

annawen1 (commented):
close! If we could also get the unit tests you've been working on, I think we should be good to go

shixiedesign:
React is fixed! Looks like 

1. web components's version's `content-item-horizontal: default` is still not showing eyebrow: 

**Current:** 
image
**Expecting**
image

2. this is totally not important but if it's easy, can we **nudge the `default` variant in storybook right by one column**? I think this is an existing error so out of scope of this PR. It's just so much more obvious now when the other variants are nice and correct.

shixiedesign (approved):
Woohoo!! And extra thanks for fixing the existing grid issue in the default component !! 🙏 🏅

annawen1 (approved):
Thanks @mwanberg! Looks good!

PR #6428, opened 2021-06-21
chore(leadspace): clean up args and attr

kennylam:
### Description

This cleans up the `Leadspace` component in React which has compile errors and console warnings.

### Changelog

**Changed**

- change `class` to `className`

Screen Shot 2021-06-21 at 10 13 15 AM

**Removed**

- `image` argument from section class name

Screen Shot 2021-06-21 at 10 13 07 AM


jeffchew (approved):
Thanks @kennylam !

PR #6433, opened 2021-06-21
fix(cdn): enable CSS custom properties in CDN artifacts

jeffchew:
### Related Ticket(s)

No related issue

### Description

A bug that came up is that the CDN artifacts did not have Carbon's CSS custom properties flag enabled. This PR enables so that CSS custom properties can be taken advantage of. In addition, there are new theme classes being introduced
as a CDN style helper. This enables the ability of creating theming zones using the theme classes. Documentation has also been added for this.

### Changelog

**New**

- `themes.css` CDN artifact with theme zone classes
- Documentation on `Carbon CDN style helpers`

**Changed**

- Updated rollup config to include css custom properties flag


IgnacioBecerra (approved):
This and the landing page you're working on looks promising, thanks @jeffchew!!

PR #6434, opened 2021-06-21
chore(stable-selectors): add missing selectors

IgnacioBecerra:
### Related Ticket(s)
None.

### Description
This PR adds the stable selector to any of our components that were missing it.

### Changelog

**New**

- added missing stable selector to many components


jeffchew:
Thank you @IgnacioBecerra ! 

It looks like the snapshots need to be updated since they are adding new autoid's to a bunch of the markup now.

Also, I added 2 points to this PR to capture the work, let me know if that sounds accurate to you.

kennylam:
@IgnacioBecerra Thanks for this! One big reason why `auto-id` wasn't added to every element is that they should be, for the most part (and as much as possible), unique. Adding it to `button` or `card-group-item`, for instance, is not especially helpful since there could be very many of them on the page.

Most analytics suites, including the ones we use, automatically listen to an interactive element whether it has an `auto-id` or not. And at that point if the parent component can be identified via `auto-id`, I think that is enough. If components are missing them at the parent level, i.e., component-level, then we should definitely add them. Otherwise I'm not sure repeated elements should have them?

jeffchew:
@kennylam @IgnacioBecerra I think parent level for the larger components for sure. Not sure about the repeatable ones, but at least for targeting these inside larger components could at least making traversing a bit easier? e.g.: `$('[data-autoid="dds--leadspace"] [data-autoid="dds--button-group"]')`

IgnacioBecerra:
@kennylam @jeffchew Yeah while I was working on the Beacon stuff I noticed that having stable selectors would make it easier for checking things (e.g. Card heading), since I had trouble getting each of the inner components' values. But let me know what would be the better option here so I can update if needed!

jeffchew (approved):
Thank you @IgnacioBecerra , I think I'm good with adding the stable selectors, makes it consistent with what we're offering across the board as part of our APIs.

PR #6435, opened 2021-06-22
fix(card-group): fix offset and optional gradient

ariellalgilmore:
### Related Ticket(s)

#6016 

### Description

added offset option by adding empty to property to card-group-item. CSS custom property for gradient value in outline. 

Talked with Shixie and gradient will not be demo'd in storybook yet. 

### Changelog

**New**

- borderAdjustments, and  _fillLastRowWithEmptyCards 
- added update to call resizeHandler when properties change in gridMode

**Changed**

- resizeHandler now calls _setSameHeight, _borderAdjustments, and  _fillLastRowWithEmptyCards

**Removed**

- {{removed thing}}


packages/styles/scss/components/card-group/_card-group.scss
```diff
@@ -116,7 +113,21 @@
...
+  //    width: 100%;
+  //    height: 100%;
+  //    position: absolute;
+  //    left: 0;
+  //    top: 0;
+  //    pointer-events: none;
+  //    box-shadow: inset 0 0 0 1px $ui-background;
+  //  }
+  //}
+
```
kennylam, line 130: @ariellalgilmore Looks like you forgot to remove these.

shixiedesign:
Looks great! only noticed `same height util` is having a bit of trouble. I was playing with the knobs. Could this be a non-issue if the cards are loaded with preset numbers, then it's okay I think. 
image

ariellalgilmore:
@shixiedesign seems like a storybook thing. When you open up the knob changes in a new tab you can see it being displayed correctly.

shixiedesign (approved):
Looks great!!! 🔥

kennylam (approved):
LGTM thanks @ariellalgilmore!

PR #6448, opened 2021-06-23
fix(leadspace): width and theme fix

NSDrowned:
### Changelog

- Added 100% width to `<picture>` for larger screens
- Moved the gradient stop color rule to `themed-items` mixin (how did this ever work before?)
- Added `g10` and `g90` to theme prop types to avoid console log error


jeffchew (approved):
Thank you @NSDrowned !

Issue #6340, opened 2021-06-10
[Cloud Megamenu]: Non-alphabetical titles being selected in the cloud megamenu when using special character languages


annawen1:
## Detailed description

> Describe in detail the issue you're having.

** This fix has to be in the July 7 release.  Code freeze for this release is June 25.**

When using special character  language translations (ie. Korean, chinese, japanese, etc.) the megamenu has selected states for multiple menu items.

This is due to the regex used in [https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/packages[…]nents/src/components/masthead/cloud/cloud-masthead-composite.ts](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/packages[…]nents/src/components/masthead/cloud/cloud-masthead-composite.ts) which removes all special characters. This messes with the tabbing functionality

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

bug

> Is this issue related to a specific component?

This issue is specific to the Cloud Megamenu component.

Additional Notes:
<img width="1494" alt="Screen Shot 2021-06-10 at 3 11 41 PM" src="https://user-images.githubusercontent.com/54281166/121591087-14f34080-ca07-11eb-9334-d786a2a74143.png">

RobertaJHahn:
Will Slack in the Squad channel to get more info.  Is this needed in the July 7 release.  Can HC dev's work on this?

PR #6451, opened 2021-06-23
fix(cloud-megamenu): modify regex to only remove punctuation special characters

annawen1:
### Related Ticket(s)

[Cloud Megamenu]: Non-alphabetical titles being selected in the cloud megamenu when using special character languages #6340

### Description

Regex was removing characters from non latin languages (ie. chinese) which caused issues with the tab component of the cloud megamenu (not knowing which tab was selected, setting all special character items as "selected" state, etc). Solution is to only remove punctuation characters.

BEFORE:
Screen Shot 2021-06-23 at 8 53 34 AM

AFTER:
Screen Shot 2021-06-23 at 9 29 42 AM

### Changelog

**Changed**

- edit regex to only target punctuation characters


proeung (approved):
LGTM as well!

Issue #6280, opened 2021-06-07
Web component: Card static - CTA is not identified by the screen reader in both arrow key navigation and also in tab/shift tab navigation.


Praveen-sr:
## Detailed description

Issue:
The CTA is not identified by the screen reader in both Arrow key navigation and also in tab/shift tab navigation. Hence the user would not know its presence.

Expected:
Screen reader is expected to identify the CTA element in both Tab navigation and Arrow key navigation

PR #6467, opened 2021-06-24
feat(card): ensure card footer is accessible with card static

IgnacioBecerra:
### Related Ticket(s)
#6280

### Description
Ensures that `aria-hidden` is only active when the whole Card is clickable; Card footer is now screen reader accessible when using Card static.

### Changelog

**Changed**

- `aria-hidden` not applied to `dds-card-footer` when using `Card static` variation


kennylam (approved):
LGTM thanks @IgnacioBecerra! Looks like you just need to update the snapshots.

Issue #6465, opened 2021-06-24
[Masthead] search input should use `aria-label` instead of `aria-labelledby`


kennylam:
The `Masthead` search input currently uses `aria-labelledby`, which does not allow screen readers to announce the correct label ("Search all of IBM"). Instead, `aria-label` should be used so this announcement is made every time focus is passed to the input, whether there is a search value or not. When a search value is present, the label will still be announced first.

PR #6468, opened 2021-06-24
fix(masthead): use aria-label instead of labelledby

kennylam:
### Related Ticket(s)

#6465 

### Description

Use `aria-label` for Masthead search input instead of `aria-labelledby`. This allows screen readers to correctly announce an input label regardless of the search input value.

### Changelog

**Changed**

- use `aria-label` instead of `aria-labelledby` for masthead search input


jeffchew (approved):
LGTM!

kennylam:
@RobertaJHahn @Praveen-sr Confirmed with Ram this is working as expected.

PR #6484, opened 2021-06-25
fix(stripHTML): add check if there are no surrounding elements

jeffchew:
### Related Ticket(s)

No related issue

### Description

The stripHTML utility always assumes that there is a wrapping tag. However, in the cases when only a string is passed, it should just pass through without doing anything, rather than throwing an error.

### Changelog

**Changed**

- Add additional string check in `stripHTML`


NSDrowned (approved):
LGTM!

Issue #5405, opened 2021-03-08
[Storybook] Investigate adding a11y tab to Storybook stories


jeffchew:
#### User Story

> As a `[user role below]`:
Carbon for IBM.com user

> I need to:
see automated a11y reports for each storybook story

> so that I can:
ensure that the components available in Carbon for IBM.com passes a11y tests

#### Additional information

 - The repo originally had the storybook plugin https://www.npmjs.com/package/@storybook/addon-a11y, but turned off as the plugin itself was not working at the time
 - We should probably wait on the storybook 6 upgrade first before tackling this
- Example:  https://next--storybookjs.netlify.app/official-storybook/?path=/story/addons-a11y-basebutton--default&globals=measureEnabled:false

#### Acceptance criteria
- [ ] Investigate whether there are any issues to include the addon
- [ ] Determine whether there are any story level adjustments for the plugin
- [ ] Can the plugin support Web Components

PR #6500, opened 2021-06-29
chore(storybook): adds a11y tab to stories

kennylam:
### Related Ticket(s)

#5405 

### Description

This adds the `Accessibility` tab to all Storybook environments.

image

### Changelog

**New**

- `Accessibility` tab added to `react`, `web components` (and `react wrapper`), and `styles` package Storybooks


kennylam:
> It looks like we have to go through and fix some of these violations that the tabs are picking up, a lot seem to be the way we created the stories and not in the components themselves.

@jeffchew For sure, a few things we need to fix.

Issue #6490, opened 2021-06-28
Web component: Card Static - CTA link is not operable on mobile.


Praveen-sr:
## Detailed description

On Mobile
Issue:
Card Static - CTA link is not operable on mobile.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/b9c227ee-5890-4e8d-b819-c50b979c3974)

PR #6504, opened 2021-06-29
fix(card-static): cta link not working on mobile

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6490

### Description

- Footer's `::after` element (not used in this variation) blocking anchor link. Removed pointer events.
- Added `position: relative` to parent so `::after` only occupies the CTA's area, not the whole screen.

https://user-images.githubusercontent.com/4054756/123836782-4ef98900-d8c7-11eb-889f-20e961ad54cf.MP4


packages/web-components/src/components/card/card.scss
```diff
@@ -18,6 +18,14 @@
     color: $inverse-link;
   }
 
+  &[color-scheme='light'] {
```
annawen1, line 21: does the static card always have the `light` color-scheme? what happens if we attach the styles to 

```
:host(#{$dds-prefix}-card-footer),
:host(#{$dds-prefix}-card-cta-footer)
```

jeffchew:
@NSDrowned is this off of the release branch? it looks like you are targeting `master`.

jeffchew (changes requested):
@NSDrowned can you reopen the PR based off the release branch? It looks like this was pulled from master.

packages/web-components/src/components/card/card.scss
NSDrowned: @annawen1, yes it's always on in static. Can't find another way to target static card. Tested it on other variations and they all work fine except for the light schem.

NSDrowned:
@jeffchew whoops, you got it!

kennylam:
@NSDrowned Link functionality working fine for me now. But for some reason Safari adds a `margin-left` to the icon.

Chrome on left, Safari on right.
image

Issue #6513, opened 2021-06-30
[CardLink] disabled cards still behave as links


kennylam:
Currently the React `CardLink` component does not properly disable when `disabled` is used – it still redirects to a URL.

![Jun-30-2021 09-21-32](https://user-images.githubusercontent.com/909118/123967777-a1e54600-d984-11eb-87f1-3b1656295b6c.gif)

PR #6514, opened 2021-06-30
fix(CardLink): fix disabled behavior

kennylam:
### Related Ticket(s)

#6513 

### Description

Adds conditional rendering of `CardLink` footer based on `disabled` state.

### Changelog

**Changed**

- conditionally render footer based on `disabled`


jeffchew (approved):
LGTM!

Issue #6394, opened 2021-06-16
[Background media] New component - DDS consulting


shixiedesign:
## Overview
The `Background media` is used behind content to display image or video, standardizing the **gradient overlay** and **mobile stacking behavior**. MVP is image for now, and later should be enhanced to support video.

This component is used in **Card section offset**.

## Functional spec: 
https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/wiki/Background-media

![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/1d83c702-85e4-4ae5-8664-43d2c02f3e75)

----
- Dev by DDS
- HC Engineer assigned:  Putra Bonaccorsi

RobertaJHahn:
@ariellalgilmore I have not set this issue up for poker planning since you have been assigned and should probably be in the best position to provide the estimate.  This component has been assigned to the July 7 sprint because the content block with L-shaped card group (#6241 ) needs it.

cc:  @jeffchew @shixiedesign

proeung:
@RobertaJHahn Here's the JIRA ticket (https://jsw.ibm.com/browse/HC-2014) that we're using to track the contribution work on our end.

proeung:
@RobertaJHahn @ariellalgilmore We have new work coming in from the HC business that requires @ddollive's time. I'm going to take on the work in this ticket so that we can still try to get this component in for the late July release.

RobertaJHahn:
@proeung Thanks for letting us know.  

cc:  @ariellalgilmore

Issue #6129, opened 2021-05-19
[card section offset] used on the Learn ecosystem a.k.a. Content block with L-shaped card group


pavolprok:
### The problem
We are currently missing a component where the card group in L-shape can be used on top of a background image as part of the main topic that is represented by the Content Block component on the left side.

### The solution

I have prepared design specs for this new component that we are using on the Learn ecosystem prototype.
There are 4 dependencies:
1. Card
2. Card group
3. Content block simple
4. Image component

![image](https://user-images.githubusercontent.com/70329414/118792490-fda5b500-b897-11eb-9ccf-2fefe420c617.png)

### Additional information

Learn ecosystem:

https://xd.adobe.com/view/f260b7b3-e460-4e78-ab17-75237a7968c4-a733/screen/9aad39f9-bf0d-4752-afa4-8bd72ae5ae38?fullscreen&hints=off

Design specs:
To find under: Content block cards
https://ibm.box.com/s/9pweudf9ptucx8jxzgo3e2m5jpyke5qi

Functional specs:
`enhancement` Offset cards in card group:
https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/wiki/Card-group

![card-section-offset--card-group](https://user-images.githubusercontent.com/3793636/121088163-f133be80-c7aa-11eb-8d42-d16dde542bf7.gif)

`new` Background media component:  
https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/wiki/Background-media

![card-section-offset--background-media](https://user-images.githubusercontent.com/3793636/121088143-e8db8380-c7aa-11eb-8779-b14e69866af8.gif)

`new` Card section offset:  
https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/wiki/Card-section-offset

![card-section-offset--card-section-offset](https://user-images.githubusercontent.com/3793636/121088198-fc86ea00-c7aa-11eb-9916-3ef2daad7ef9.gif)

----
This component will be developed by a Hybrid Cloud engineer.

- Corresponding Jira issue:  https://jsw.ibm.com/browse/HC-1970

RobertaJHahn:
AEM will develop this component behind feature flag and contribute it.  

Met with Lily, Lila and Sandeep today.  They asked whether James can create the functional specs.

RobertaJHahn:
[...]

PR #6548, opened 2021-06-30
feat(background-media): add "Background media" UI component

proeung:
### Related Ticket(s)
https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6129

DDS Consulting Issue: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6394
HC JIRA Ticket: https://jsw.ibm.com/browse/HC-2014

### Description
- This PR adds the "Background media" UI component/pattern that designed to display an `image` or `video` with gradient options. For MVP we're looking to expose `image` for now, and later should be enhanced to support video background.
- This component is dependent of the L-shaped card group - https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6129.

### Design & Functional Specs
- https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/wiki/Background-media

68747470733a2f2f696d616765732e7a656e68756275736572636f6e74656e742e636f6d2f3562653039633361363130323930306336643136643939312f31643833633730322d383565342d346165352d383636342d343364326330326633653735

### Changelog

**New Components/Props**

- `<dds-background-media>` – this new layout component is a variant/extension of the `<dds-image>`.
- `gradient-direction` - a prop/attribute (left-to-right (default) | top-to-bottom)
- `mobile-position` - a prop/attribute (top (default) | bottom)

**Removed**

- N/A

**Note**

- Need to figure out how to treat mobile image position where typically this should be defined within the layout component this `<dds-background-media>` is being used.

**Testing Instructions**
- Visit the `Background media` - `/?path=/story/components-background-media--default`
- Check the "Gradient direction" knobs
- Ensure that the component matches with the design specs for all breakpoints.


shixiedesign:
Hey @proeung is this ready for design review?

proeung:
@shixiedesign Yep! I just changed the PR status from `draft` to `ready for review`.

packages/styles/scss/components/background-media/_background-media.scss
```diff
@@ -0,0 +1,51 @@
...
+@import 'carbon-components/scss/globals/scss/vendor/@carbon/elements/scss/themes/mixins';
+@import '../../globals/vars';
+@import '../../globals/imports';
+@import '../image/image';
+
+@mixin background-media {
+  :host(#{$dds-prefix}-background-media),
+  .#{$prefix}--background-media {
+    position: relative;
+  }
```
ariellalgilmore, line 17:
```suggestion
  :host(#{$dds-prefix}-background-media) {
    position: relative;
  }
```

ariellalgilmore (approved):
LGTM!!!! Thank you @proeung

shixiedesign:
Hi @proeung sorry for the wait! The only issue I noticed is **gradient top-to-bottom doesn't have the correct values.** The overlay should fade like this 

image

packages/styles/scss/components/background-media/_background-media.scss
```diff
@@ -0,0 +1,50 @@
...
+        to right,
+        $ui-background 25%,
+        rgba(255, 255, 255, 0) 75%
+      );
+    }
+  }
+
+  .#{$prefix}--background-media--gradient--top-to-bottom {
+    @include carbon--breakpoint('lg') {
+      background: linear-gradient(
```
kennylam, line 39:
```suggestion
      background-image: linear-gradient(
```

packages/web-components/src/components/background-media/background-media.ts
```diff
@@ -0,0 +1,90 @@
...
+@customElement(`${ddsPrefix}-background-media`)
+class DDSBackgroundMedia extends DDSImage {
+  /**
+   * Returns a class-name based on the Gradient Direction type
+   */
+  protected _getGradientClass() {
+    return classMap({
+      [`${prefix}--background-media--gradient`]: true,
+      [`${prefix}--background-media--gradient--left-to-right`]: this.gradientDirection === GRADIENT_DIRECTION.LEFT_TO_RIGHT,
+      [`${prefix}--background-media--gradient--top-to-bottom`]: this.gradientDirection === GRADIENT_DIRECTION.TOP_TO_BOTTOM,
```
kennylam, line 36:
```suggestion
     [`${prefix}--background-media--gradient--${this.gradientDirection}`]: this.gradientDirection,
```

packages/web-components/src/components/background-media/background-media.ts
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  /**
+   * Returns a class-name based on the Mobile Position type
+   */
+  protected _getMobilePositionClass() {
+    return classMap({
+      [`${prefix}--background-media--mobile-position`]: true,
+      [`${prefix}--background-media--mobile-position--top`]: this.mobilePosition === MOBILE_POSITION.TOP,
+      [`${prefix}--background-media--mobile-position--bottom`]: this.mobilePosition === MOBILE_POSITION.BOTTOM,
```
kennylam, line 47:
```suggestion
     [`${prefix}--background-media--mobile-position--${this.mobilePosition}`]: this.mobilePosition,
```

kennylam (commented):
Thanks for this @proeung! Just a few small suggestions.

packages/styles/scss/components/background-media/_background-media.scss
```diff
@@ -0,0 +1,50 @@
...
+    }
+  }
+
+  .#{$prefix}--background-media--gradient--top-to-bottom {
+    @include carbon--breakpoint('lg') {
+      background-image: linear-gradient(
+        to bottom,
+        $ui-background 0%,
+        rgba(255, 255, 255, 0) 86%
+      );
```
ariellalgilmore, line 43: looking at shixie's feedback. I think these are the updates needed for the gradient. Let me know if you agree @proeung 

```suggestion
      background-image: linear-gradient(
        to bottom,
        $ui-background 25%,
        rgba(255, 255, 255, 0) 75%
      );
```

proeung:
@shixiedesign I just pushed up a commit that adjusted the gradient value of the `top-to-bottom` direction to include a third opacity value as defined in the Sketch library. However, in doing so I have to expose a new prop as the current Carbon theme doesn't support light or dark `RGB` tokens.

Let me know if my approach makes sense. @ariellalgilmore and @annawen1 I'd like to get your input on this as well - https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/6548/commits/21df9493384b92b903d32a2f3bd89e838ae1e7eb.

ariellalgilmore:
> @shixiedesign I just pushed up a commit that adjusted the gradient value of the `top-to-bottom` direction to include a third opacity value as defined in the Sketch library. However, in doing so I have to expose a new prop as the current Carbon theme doesn't support light or dark `RGB` values unless there's a way to add to the theme tokens.
> 
> Let me know if my approach makes sense. @ariellalgilmore and @annawen1 I'd like to get your input on this as well - [21df949](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/commit/21df9493384b92b903d32a2f3bd89e838ae1e7eb).

hi @proeung! this workaround definitely works, but ideally we would want the component to change according to the theme and not need the adopter to set the attribute

proeung:
@ariellalgilmore I agree. The workarounds implemented in this PR and in the [LS](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/packages/styles/scss/components/leadspace/_leadspace.scss) are definitely not ideal, but it gets us to the `top-to-bottom` gradient value as defined in the design.

My ideal solution is to add `rgba()` tokens to the Carbon theme variables so that we can apply the theme based on what the adopter defines so that we can eliminate future tech debt and confusion. Also, another solution is to limit the `top-to-bottom` direction to two points as I had it previously and adjust the start and ending percentage.

shixiedesign (approved):
Visually this is good! I understand the current limitation with top-to-bottom gradient and will keep an eye on the Carbon issue. Thanks again Putra!

shixiedesign:
Slightly concerned about Percy result here for Web components e2e test. Any idea why the masthead could be affected? These are seen in `dds-dotcom-shell | cdn | default` which is also not a story in web components storybook. If this is the CDN artifact adopters are getting, it's worth getting it right. But I could be missing something or just a percy error too. Lemme know. 

image
image

jeffchew:
> Slightly concerned about Percy result here for Web components e2e test. Any idea why the masthead could be affected? These are seen in `dds-dotcom-shell | cdn | default` which is also not a story in web components storybook. If this is the CDN artifact adopters are getting, it's worth getting it right. But I could be missing something or just a percy error too. Lemme know.
> 

This is the link that this percy snapshot is built from:

https://webcomponents-codesandbox.s3-web.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/6548/dotcom-shell/cdn.html

This was just spun up pretty recently, but want to keep a close eye on this as the snapshot baseline stabilizes. 

cc: @ariellalgilmore

Issue #6505, opened 2021-06-29
[Masthead] React component: Autosuggest in the searchbox causes runtime error


PatrikHysky:
## Detailed description

> Describe in detail the issue you're having.
Having problems with the Autosuggest function in the searchbox of [Masthead]. A list of suggestions displays as I type in the searchbox but upon clicking a suggestion I get a runtime error. Everything else in the masthead seems to work OK including the search itself (even with custom EventListeners).

> Is this a feature request (new component, new icon), a bug, or a general
> issue?
BUG

> Is this issue related to a specific component?
Masthead

> What did you expect to happen? What happened instead? What would you like to
> see changed?
`Uncaught TypeError: Cannot read property 'focus' of undefined
    at Object.eval [as onClick] (Autosuggest.js?0a0f:208)
    at eval (Item.js?4fd4:114)
    at HTMLUnknownElement.callCallback (react-dom.development.js?61bb:3945)
    at Object.invokeGuardedCallbackDev (react-dom.development.js?61bb:3994)
    at invokeGuardedCallback (react-dom.development.js?61bb:4056)
    at invokeGuardedCallbackAndCatchFirstError (react-dom.development.js?61bb:4070)
    at executeDispatch (react-dom.development.js?61bb:8243)
    at processDispatchQueueItemsInOrder (react-dom.development.js?61bb:8275)
    at processDispatchQueue (react-dom.development.js?61bb:8288)
    at dispatchEventsForPlugins (react-dom.development.js?61bb:8299)
    at eval (react-dom.development.js?61bb:8508)
    at batchedEventUpdates$1 (react-dom.development.js?61bb:22396)
    at batchedEventUpdates (react-dom.development.js?61bb:3745)
    at dispatchEventForPluginEventSystem (react-dom.development.js?61bb:8507)
    at attemptToDispatchEvent (react-dom.development.js?61bb:6005)
    at dispatchEvent (react-dom.development.js?61bb:5924)
    at unstable_runWithPriority (scheduler.development.js?3069:468)
    at runWithPriority$1 (react-dom.development.js?61bb:11276)
    at discreteUpdates$1 (react-dom.development.js?61bb:22413)
    at discreteUpdates (react-dom.development.js?61bb:3756)
    at dispatchDiscreteEvent (react-dom.development.js?61bb:5889)`

> What browser are you working in?
Chrome, Edge

> What version of Carbon for IBM.com are you using?
1.20.0

[...]

PR #6553, opened 2021-07-01
fix(MastheadSearch): set focus on search suggestion to false

annawen1:
### Related Ticket(s)

[Masthead] React component: Autosuggest in the searchbox causes runtime error #6505

### Description

sets the `focusInputOnSuggestionClick` prop for the `Autosuggest` component to false so error doesn't appear when user clicks on search suggestion.

### Changelog

**Changed**

- set `focusInputOnSuggestionClick={false}` for `Autosuggest` component


jeffchew (approved):
Thank you for the quick fix @annawen1 !

We should probably tell Praveen about this fix so he can focus on regression around autocomplete in React.

Issue #6554, opened 2021-07-02
ibmdoctcom-styles - Error in minified css version 1.19.0 and above - we are unable to upgrade dotcom


Daynssjal:
React:

As per the documentation https://www.npmjs.com/package/@carbon/ibmdotcom-styles -> Usage, we are using this import to apply styles from ibm dotcom in our project:

`import '@carbon/ibmdotcom-styles';`

After upgrading to version 1.19.0 and above (we tried 1.20.0 as well as the current rc - 1.21.0-rc.0), we receive an error in the console, due to which our whole build fails:

![image (35)](https://user-images.githubusercontent.com/33804471/124249062-cc213a00-db23-11eb-933c-e61afc9b57f6.png)

We checked the ibmdotcom-styles minified css file, where the error occurs and localized the error:

<img width="704" alt="dotcom-styles-bug" src="https://user-images.githubusercontent.com/33804471/124249331-10143f00-db24-11eb-963b-9b41f9fd4b0c.PNG">

For easier localiziation, we unminified css and found out there are 2 errors like that:

![image (33) (1)](https://user-images.githubusercontent.com/33804471/124249588-5073bd00-db24-11eb-964b-9cebb60c507b.png)
![image (34) (1)](https://user-images.githubusercontent.com/33804471/124249593-51a4ea00-db24-11eb-99b8-8b7cdaf3802d.png)

I had a chat with Anna Wen from your team, who checked this issue and said, that the line which produces this code in the minified css is this one: ` @include carbon--breakpoint-between('md', $TEMP--breakpoint-down--lg)` and that it hasn't been changed since last  year. 

So there might be a deeper underlying issue.

What we tried:
- upgraded not only dotcom react and dotcom styles but also carbon-components and carbon-components-react to the latest version at that time (carbon-components: "10.37.0", carbon-components-react: "7.37.0")
- later we downgraded carbon-components to "10.36.0" and carbon-components-react to "7.36.0" as suggested by Ari in this slack post  https://ibm-content-coe.slack.com/archives/C2PLX8GQ6/p1623682361279700
- with each new release or release candidate of ibmdotcom-react and ibmdotcom-styles, we upgraded to that version just to check if the error is still there - it was

We tried a different strategy too:  **(plan B)**
- we removed the `import '@carbon/ibmdotcom-styles';` - the page loaded okay, as the error was not present anymore, however, of course, there were no dotcom styles in our project anymore
[...]

PR #6557, opened 2021-07-02
fix(styles): temporarily hard-code breakpoint values

kennylam:
### Related Ticket(s)

#6554 

### Description

Temporarily use numerical breakpoint values for the `carbon--breakpoint-between` mixin. Currently there is an error in this mixin that prevents the usage of breakpoint keywords, i.e. `md`, `lg`, etc.

A [PR](https://github.com/carbon-design-system/carbon/pull/9102) for this fix is currently open. Once merged these changes can be reverted to use keywords.

### To-do
Revert these changes when the changes are available in the next `carbon-components` release. [#6558]

### Changelog

**Changed**

- use numerical values instead of keywords for `carbon--breakpoint-between` mixin

### DIFF
--- a/packages/styles/scss/components/footer/_footer-nav.scss
+++ b/packages/styles/scss/components/footer/_footer-nav.scss
@@ -30,7 +30,10 @@
         padding: 0 $carbon--grid-gutter / 2;
       }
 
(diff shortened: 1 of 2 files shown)


jeffchew (approved):
Thank you @kennylam !

annawen1 (approved):
Looks good!

Issue #6251, opened 2021-06-03
[Web Components | e2e testing] Create Cypress e2e tests for isolated component deployments


jeffchew:
#### User Story

> As a `[user role below]`:
Carbon for IBM.com tester

> I need to:
run e2e tests on isolated component builds

> so that I can:
ensure that the components run properly as isolated imports in an application

#### Additional information

 - Will rely on the issue https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6250 before Cypress can be set up
 - Should see if setting up Percy is a possibility, especially if these will not be run on a pull request basis
 - Should run as part of a CI flow (either via pull request or on a daily cron

#### Acceptance criteria
 - [x] Cypress setup for the isolated e2e tests for web components
 - [x] CI setup to run e2e tests

PR #6559, opened 2021-07-02
fix(e2e): added timeout for render time on CI

jeffchew:
### Related Ticket(s)

#6251 

### Description

These are some tweaks to the build so that it runs correctly in the CI. It needed more time to complete rendering.

Also, turning off percy snapshots for video player, as it was causing false positives.

### Changelog

**Changed**

- `cypress.json` config tweaks
- suppress `percySnapshot` for `video-player`


annawen1 (approved):

kennylam (approved):
Thanks @jeffchew!

PR #6581, opened 2021-07-07
chore(storybook): restore react storybook webpack fast-sass-loader

kennylam:
### Description

This restores `fast-sass-loader` in React Storybook for non-production environments. The recent change removing `fast-sass-loader` has led to considerably increased load/reload times in Storybook. Using `fast-sass-loader` decreases load times for both initial/hot -loading by ~30s.

**Before**

orig

**After**

fast-sass

### Changelog

**Changed**

- restore `fast-sass-loader` for non-production environments
- fix concatenated class name which is not supported in web components


jeffchew (approved):
LGTM!

Issue #6178, opened 2021-05-25
[Masthead]: Provide local/globalization support to Platform Name and URL


proeung:
### The problem
- The current attributes for passing a platform name and URL to the `dds-masthead-container` and `dds-cloud-masthead-container` works great, however, it doesn't scale well on localized pages.
- Here's an example of the current Drupal Cloud masthead that has the string "cloud" translated into `arabic` (https://cms.ibm.com/ar/cloud/paks) as well as its homepage URL (https://cms.ibm.com/ar/cloud).
- This feature is needed within the production use of the Unified Cloud Masthead + Mega Menu as well as in AEM.

![Screen_Shot_2021-05-25_at_11_12_26_AM](https://user-images.githubusercontent.com/1815714/119522753-3a672580-bd4a-11eb-8ebc-724018b88d96.jpg)

```
<dds-cloud-masthead-container
  platform="Cloud"
  data-endpoint="/common/carbon-for-ibm-dotcom/translations/cloud-masthead/jsononly"
  platform-url="https://www.ibm.com/cloud"
></dds-cloud-masthead-container>
```

### The solution
- Alter the data fetching method to pull from the endpoint if the data exists.
- We might need to add a new field within the `mastheadNav` array for `mastheadNav[platform]` that manages its translated name and URL. https://github.ibm.com/webstandards/ibm-dotcom-translations/tree/master/translations

jeffchew:
Does the platform name need to be translated if it's an IBM product name?

proeung:
@jeffchew From what I've seen so far, the platform name looks to be translated for non-English speaking countries, please see below for the list of pages that are translating the word "Cloud". 

- https://cms.ibm.com/ar/cloud
- https://cms.ibm.com/jp-ja/cloud
- https://cms.ibm.com/cn-zh/cloud
- https://cms.ibm.com/ar-es/cloud

I'm not sure if this is the standard, however, I think it's still important to expose the `platform-url` string in the API endpoint so that each local can put in their proper homepage link.

![Screen_Shot_2021-05-27_at_10_54_47_AM](https://user-images.githubusercontent.com/1815714/119848696-07539c00-beda-11eb-8efd-f282dfac1d44.jpg)

RobertaJHahn:
Will be added to the 6/7 engineering refinement meeting agenda.

RobertaJHahn:
HC has a work around until this is ready.  Moving to the July 27 release.

[...]

PR #6586, opened 2021-07-07
feat(masthead): globalization support

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6178

### Description

Provide local/globalization support to platform name and URL with `platformObj`. If `language` matches one of the locales in the object, name and url will be used.

`platformObj` example:

```
const platformObject = {
  'en-US': {
    url: 'https://www.cloud.com/us-en',
    name: 'Cloud',
  },
  'fr-FR': {
    url: 'https://www.cloud.com/fr-fr/sample',
    name: 'Nuage',
  },
  'es-MX': {
    url: 'https://www.cloud.com/ibm/sample/es-mx',
    name: 'Nube',
  },
};
```

### Changelog

**New**

- Optional `platformObj` property can be passed to `dds-mastehead-container` and `dds-cloud-masthead-container`


annawen1 (approved):
LGTM! @proeung could you take a look at this as well?

packages/web-components/src/components/masthead/__stories__/cloud-masthead.stories.ts
```diff
@@ -30,6 +30,21 @@ const platformData = {
   url: 'https://www.ibm.com/cloud',
 };
 
+const platformObject = {
+  'en-US': {
+    url: 'https://www.cloud.com/us-en',
+    name: 'Cloud',
```
proeung, line 36: I don't think we need to account for `platformObject.local.name` in the object as from what I understand Product Name doesn't go through the globalization process.

proeung (commented):
Looks good overall! I just have a couple of comments.

packages/web-components/src/components/masthead/__stories__/masthead.stories.ts
```diff
@@ -35,6 +35,21 @@ const platformData = {
...
+    url: 'https://www.cloud.com/us-en',
+    name: 'Cloud',
+  },
+  'fr-FR': {
+    url: 'https://www.cloud.com/fr-fr/sample',
+    name: 'Nuage',
+  },
+  'es-MX': {
+    url: 'https://www.cloud.com/ibm/sample/es-mx',
+    name: 'Nube',
```
annawen1, line 49: @jeffchew do you know if other platform names should be translated? I know for `Cloud` we don't want it translated.

packages/web-components/src/components/masthead/__stories__/cloud-masthead.stories.ts
```diff
@@ -60,6 +75,7 @@ export const Default = !DDS_CLOUD_MASTHEAD
               <dds-cloud-masthead-container
                 platform="Cloud"
                 platform-url="${ifNonNull(platformData.url)}"
```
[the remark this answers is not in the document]
annawen1: ^ @NSDrowned Can you add documentation in the README around this? I can see this being confusing to adopters^

packages/web-components/src/components/masthead/__stories__/cloud-masthead.stories.ts
NSDrowned: @proeung, `name` is optional in `platformObj`. If you want to show the same platform name for all locales with different URLs, you can combine `platformObj` with the `platform` attribute

So the following object would would work as long as `platform` exists in `dds-cloud-masthead-container`:

```
const platformObject = {
  'en-US': {
    url: 'https://www.cloud.com/us-en',
  },
  'fr-FR': {
    url: 'https://www.cloud.com/fr-fr/sample',
  },
  'es-MX': {
    url: 'https://www.cloud.com/ibm/sample/es-mx',
  },
};
```

If this is too confusing, I can remove name completely.

packages/web-components/src/components/masthead/__stories__/cloud-masthead.stories.ts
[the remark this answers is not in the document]
NSDrowned: @annawen1 I updated the readme. I added both examples for `platform-url` and `platformObj`. 

I've was thinking that perhaps we should only keep one property: platform-url or platformObj (perhaps rename this one) and detect if the value is an object or a string and act accordingly. If you need a single URL, send a string. If you want different URLs by locale, send and object. 

Would that be a bit less confusing?

packages/web-components/src/components/masthead/__stories__/README.stories.mdx
```diff
@@ -41,6 +41,32 @@ import '@carbon/ibmdotcom-web-components/es/components/masthead/masthead-contain
...
+    url: 'https://www.example.com/fr-fr/sample',
+  },
+  'es-MX': {
+    url: 'https://www.example.com/sample/es-mx',
+  },
+};
+```
+
+```html
+<dds-masthead-container platform="Lorem Ipsum" platformObj={urlObject}></dds-masthead-container>
```
annawen1, line 67: if we're passing in as a property, I think it needs to have the `.` in front, so `.platformObj={urlObject}`

packages/web-components/src/components/masthead/__stories__/README.stories.mdx
```diff
@@ -41,6 +41,32 @@ import '@carbon/ibmdotcom-web-components/es/components/masthead/masthead-contain
...
 ```
 
+#### Setting Platform and URL
+
+```html
+<dds-masthead-container platform="Lorem Ipsum" platform-url="https://www.example.com"></dds-masthead-container>
+```
+
+#### Setting Platform URL Object (Optional)
+
```
annawen1, line 51: We should probably add some wording here to explain what this is for.

packages/web-components/src/components/masthead/__stories__/cloud-masthead.stories.ts
NSDrowned: Name has been removed 👍

packages/web-components/src/components/masthead/__stories__/cloud-masthead.stories.ts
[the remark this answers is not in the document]
NSDrowned: @annawen1, I updated the code so that the `.platformUrl` property now handles both a string or an object 👍  Docs have also been updated.

packages/web-components/src/components/masthead/__stories__/masthead.stories.ts
NSDrowned: Name doesn't apply anymore 👍

packages/web-components/src/components/masthead/masthead-composite.ts
```diff
@@ -693,13 +693,21 @@ class DDSMastheadComposite extends LitElement {
     } = this;
     const authenticated = userStatus !== UNAUTHENTICATED_STATUS;
     const profileItems = authenticated ? authenticatedProfileItems : unauthenticatedProfileItems;
+    const formattedLang = language?.toLowerCase().replace(/-(.*)/, m => m.toUpperCase());
+    let platformAltUrl = platformUrl;
+    if (platformUrl && formattedLang) {
+      if (typeof platformUrl === 'object' && Object.prototype.hasOwnProperty.call(platformUrl, formattedLang)) {
+        platformAltUrl = platformUrl[formattedLang].url || platformUrl;
```
jeffchew, line 700: @NSDrowned just checking, this seems like an API change? We should still support the old platform url as a string so we don't break it for adopters that are currently using it with the current API. If we detect a string, we should still support but perhaps throw a deprecation notice.

jeffchew: Nevermind, took another look and realized that we are allowing both, thanks @NSDrowned !

Issue #6550, opened 2021-07-01
React - Content block media - Feature card external link opens in same tab


Praveen-sr:
## Detailed description

Issue:
Content block media - Feature card external link opens in same tab

Expected:
Feature card external link should open in new tab.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/7797c8e6-ba86-4a4e-a060-157fa2a3f673)

PR #6590, opened 2021-07-07
fix(FeatureCTA): adjust prop cta structure to set target correctly

annawen1:
### Related Ticket(s)

React - Content block media - Feature card external link opens in same tab #6550

### Description

Change `cta` prop structure for the `FeatureCard` so that it can set the `target` attribute correctly.

### Changelog

**Changed**
- move `type` within `cta` object so the `target` can be set in the child `Card` component
- structure in README


kennylam (commented):
@annawen1 `ContentBlockMedia` looks good, but should we update `CTA - FeatureCard` as well? Right now it opens an external link correctly, but the icon doesn't update based on type.

image

kennylam (approved):
LGTM thanks @annawen1!

PR #6609, opened 2021-07-12
chore(styles): update clean-css to remove duplicates and includes

kennylam:
### Description

This fixes an issue where an unneeded style reset was included in the global imports file. This resulted in hundreds of duplicated blocks in the outputted files found in `/dist`, as well as styles compiled by bundlers such as Webpack. As this reset is included in `carbon-components/scss/globals/scss/css--reset` (which is already imported globally), there is no need to call it again.

Lots and lots (and lots!) of these – among others, but mostly these.

image

Removing this, along with modifying the optimizer used in Gulp to remove duplicates and comments, has reduced the file size of unminified css files by ~3mb and removed ~80k lines of redundant/unneeded code.

**Before**
Screen-Shot-2021-07-12-at-10 59 28-AM
Screen Shot 2021-07-12 at 11 07 00 AM

**After**
Screen-Shot-2021-07-12-at-11 00 42-AM
Screen Shot 2021-07-12 at 11 02 26 AM

### Changelog

**New**

- add optimization levels 1 (remove comments) and 2 (remove duplicate rules) to `gulp-clean-css`

**Removed**

- `@include carbon--type-reset()` removed from global imports


jeffchew (approved):
Great catch!

PR #6630, opened 2021-07-13
chore(content-block-mixed): update import paths in docs

kennylam:
### Description

Update import paths in README.

### Changelog

**Changed**

- update import paths to use `es` modules


annawen1 (approved):
Thanks Kenny!

Issue #6264, opened 2021-06-04
[Services | VideoPlayerAPI] Deprecate the VideoPlayerAPi in favor of KalturaPlayerAPI


jeffchew:
#### User Story

> As a `[user role below]`:
Carbon for IBM.com user

> I need to:
use the KalturaPlayerAPI

> so that I can:
embed a video or audio player using the service

#### Additional information

 - This PR introduced the audio player, which essentially brought in a duplicate of the `VideoPlayerAPI` and renamed it to `KalturaPlayerAPI`: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/6102
 - We should deprecate the VideoPlayerAPI and add messaging to it
 - Any instances of the VideoPlayerAPI should be swapped out with KalturaPlayerAPI
 - The updated VideoPlayerAPI should be wiped out and only extend KalturaPlayerAPI, with the included deprecation notice

#### Acceptance criteria
 - [ ] All player components point to KalturaPlayerAPI
 - [ ] Deprecation notice added to VideoPlayerAPI
 - [ ] VideoPlayerAPI extends KalturaPlayerAPI

PR #6635, opened 2021-07-14
chore(videoPlayerAPI): deprecation in favor of KalturaPlayerAPI

guilhermelMoraes:
### Related Ticket(s)

[Services | VideoPlayerAPI] Deprecate the VideoPlayerAPi in favor of KalturaPlayerAPI #6264

### Description

### Changelog

**New**

- {{new thing}}

**Changed**

- {{changed thing}}

**Removed**

- {{removed thing}}


packages/services/src/services/VideoPlayer/VideoPlayer.js
```diff
@@ -130,219 +17,6 @@ let videoData = {};
...
-    let seconds = Math.floor((duration / 1000) % 60);
-    const minutes = Math.floor((duration / (1000 * 60)) % 60);
-    let hours = Math.floor((duration / (1000 * 60 * 60)) % 24);
-    hours = hours > 0 ? hours + ':' : '';
-    seconds = seconds < 10 ? '0' + seconds : seconds;
-
-    return duration && '(' + hours + minutes + ':' + seconds + ')';
-  }
-}
+class VideoPlayerAPI extends KalturaPlayerAPI {}
```
kennylam, line 20: We should probably add a deprecation warning here.

kennylam (commented):
Looking good @guilhermelMoraes! Looks like there might be a few other places where `VideoPlayerAPI` needs to be swapped out. That's one of  the reasons the `ci-check` is failing. Just one example:

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/806a94d6e96f6b457665d8f331c59d5fd6cede80/packages/web-components/src/components/video-player/video-player-composite.ts#L13

jeffchew (approved):
LGTM!

Issue #6414, opened 2021-06-18
CTA Block wrong alignment when using different media items


dratzenboeck-ibm:
## Detailed description
When mixing image and video content items in the cta-block, they are not aligned. 
![image](https://user-images.githubusercontent.com/85555103/122546700-834f8880-d02f-11eb-8e2b-ad797b785bb5.png)

Most likely, because there a margin top on `.bx--video-player__video-container` which is not there on the image.
![image](https://user-images.githubusercontent.com/85555103/122546962-cf9ac880-d02f-11eb-9cd8-66491e110b61.png)

> Describe in detail the issue you're having.
In AEM we offer the authors to mix image and video content items within one cta block. They look bad when they are not aligned.
> Is this a feature request (new component, new icon), a bug, or a general
> issue?
bug

> Is this issue related to a specific component?
CTA-Block / CTA-Section with content items https://ibmdotcom-web-components.mybluemix.net/?path=/story/components-cta-section--with-content-items 

> What did you expect to happen? What happened instead? What would you like to
> see changed?
The media items should be in one line. The margin top should be removed from the video within the cta-block-item

> What browser are you working in?
Version 91.0.4472.106 (Official Build) (64-bit)

> What version of Carbon for IBM.com are you using?
    "@carbon/ibmdotcom-styles": "1.20.0-rc.1",
    "@carbon/ibmdotcom-web-components": "1.5.0-rc.1",

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?
IBM US Adobe Experience Manager CMS

PR #6641, opened 2021-07-14
fix(cta-block): remove margin top from video container

annawen1:
### Related Ticket(s)

CTA Block wrong alignment when using different media items #6414

### Description

`video-player-container` has a top margin applied to itself, which causes spacing issues when placed within another component. The issue becomes visible when user tries to align both video and images in the `cta-block` or `cta-section` components.

This fix resolves some spacing mismatching in `callout-with-media` as well.

**CURRENT:**
Screen Shot 2021-07-14 at 5 20 38 PM

**AFTER:**
Screen Shot 2021-07-14 at 5 17 46 PM

### Changelog

**Changed**

- set the margin-top styling for `cta-block-item__media` to maintain original spacing
- - set margin top in video-player story

**Removed**

- margin-top styling on the `video-player-container`


DragosRistici:
i'll take this one @annawen1

kennylam (approved):
Thanks @annawen1!

DragosRistici:
Looking good, Anna. This little thing that we've talked about (extra spacing on all items, not only with media) on Slack and we should be good to go. Will check it again in the morning. 
Screenshot 2021-07-15 at 23 09 16

annawen1:
@DragosRistici after looking more into this and looking at the design specs, I believe there was extra spacing added to the `text` and `button` content items that isn't supposed to be there. I removed the spacing so that there is a total of 64px between the CTA section copy and the content items.

Screen Shot 2021-07-15 at 5 13 04 PM

I set the class to be `hidden` when there are no media items passed, so the spacing is now consistent to 64px. (48px bottom padding from the "Browse tutorial" link + 16px from top margin of the content item)
Screen Shot 2021-07-15 at 5 45 26 PM
Screen Shot 2021-07-15 at 5 45 10 PM


Let me know if this is wrong and I'll remove those changes!

DragosRistici:
A little thing, @annawen1 , on mobile, the section heading lost the grid. 

Screenshot 2021-07-16 at 18 16 54

DragosRistici (approved):
LGTM, @annawen1! Thank you!

Issue #6596, opened 2021-07-08
Web component: Feature Section - The Eye brow text is not identified by the screen reader


Praveen-sr:
## Detailed description

Issue:
The Eye brow text is not identified by the screen reader

Expected:
Screen reader is expected to identify the Eye brow text

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/5ee73bfb-923b-4480-ab3f-89acb07dc8fc)

Issue #6512, opened 2021-06-30
Web component: Card group - when focus reaches the link in each card, Screen reader announces a lengthy text which makes it over verbose.


Praveen-sr:
## Detailed description

Issue:
In tab / shift tab navigation, when focus reaches the link in each card, Screen reader announces a lengthy text which makes it over verbose.

Expected:
It may be possible if the entire content in the card is given in anchor tag. In this scenario, only the heading can be given in anchor tag. Note: the link text/label should be concise and convey only the relevant information.

Praveen-sr:
Ram re tested this on both JAWS and Voiceover on the [ link ]( https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/6642/index.html?path=/story/overview-getting-started--page)  - Working as expected.

Issue #6281, opened 2021-06-07
Web component: Card static - Eyebrow - The Eye brow text is not identified by the screen reader


Praveen-sr:
## Detailed description

Issue:
The Eye brow text is not identified by the screen reader

Expected:
Screen reader is expected to identify the Eye brow text

Praveen-sr:
@ariellalgilmore Ram re tested this on both JAWS and Voiceover on the [ link ]( https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/6642/index.html?path=/story/overview-getting-started--page)  - Working as expected.

Issue #6005, opened 2021-05-07
Web Components: Card section - simple - beginning and ending boundary for the card is not conveyed for the screen reader user. 


RobertaJHahn:
## Detailed description

Issue: 
1. The beginning and ending boundary for the card is not conveyed for the screen reader user. 
2. In tab / shift tab navigation, when focus reaches the link in each card, Screen reader announces a lengthy text which makes it oververbose.

Expected:
1. Provide each card within the landmark region and give appropriate label.
2. It may be possible if the entire content in the card is given in anchor tag. In this scenario, only the heading can be given in anchor tag. Note: the link text/label should be concise and convey only the relevant information. 

Note: working fine on react version.

## Additional information
- this fix should address the bug included in issue (#5504 ). Engineer should check to be sure this fix addresses the bug included in (#5504 )
 - Reach out to kenny for additional details.
- The React bug was closed in sprint 21-09 and included in the May v1.19.0 release under issue (#4613 ).  Assigning to Ari as she worked on fixing the React bug.

Praveen-sr:
@ariellalgilmore Ram re tested this on both JAWS and Voiceover on the [ link ]( https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/6642/index.html?path=/story/overview-getting-started--page)  - Working as expected.

PR #6642, opened 2021-07-15
fix(card): update accessibility

ariellalgilmore:
### Related Ticket(s)

#6693 - card in card focus visibility
#6596 - feature section eyebrow text screen reader
#6281 - card static eyebrow text screen reader
#6512 - card group over verbose
#5504 - card in card eyebrow text screen reader
#6005 - card section simple boundaries
#4614 - card section images boundaries
#4714 - content block cards boundaries

### Description

Updated Card component to no longer wrap entire card in `<a>` the `<a>` only wraps around the footer, so that accessibility users can tab through cards and optionally use arrow keys to read the copy, header, or text.

### Changelog

**Changed**

- updated card and card footer components for accessibility use
- updated styles for focus states


packages/web-components/examples/codesandbox/components/card-group/cdn.html
```diff
@@ -52,7 +52,7 @@
               <path
                 d="M11.8 2.8L10.8 3.8 16.2 9.3 1 9.3 1 10.7 16.2 10.7 10.8 16.2 11.8 17.2 19 10z"></path>
             </svg>
-          </dds-card-footer>
+          </dds-card-cta-footer>
```
jeffchew, line 55: @ariellalgilmore is this an API change or a correction in the example?

ariellalgilmore: correction in the example. card-group should always use `<dds-card-cta-footer>`. `<dds-card-group-item>` extends `DDSCardCTA` so the footer should match.

jeffchew:
The alignment of the arrow is now the bottom right in link list, this is intentional? 
image

packages/web-components/examples/codesandbox/components/card-group/cdn.html
jeffchew: Ok, we might want to note it in the next release to change to `dds-card-cta-footer` if this change breaks for anyone using `dds-card-footer` instead.

ariellalgilmore: yeah good idea!

jeffchew:
@ariellalgilmore it looks like cards are unclickable unless it's changed to `dds-card-cta-footer`, wonder if we should add some messaging for this? 

https://ibmdotcom-web-components-html-test-upstream.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/6642/index.html

jeffchew (approved):
LGTM!

As discussed, we should include a PSA on the fixed documentation around card-group in our next release. Might even want adopters to know about it when the first rc comes out.

Issue #6672, opened 2021-07-16
[Masthead (dotcom & Cloud)]: Overflow logic is behaving weirdly on RTL


proeung:
## Detailed description
- While testing the new `<dds-cloud-masthead-container>` within the Drupal pages in RTL, I noticed that the overflow container is constantly adding and removing the `.dds--ce--header__nav-caret-container--hidden` class resulting in flickering behavior where the masthead is unusable on RTL direction.
- This behavior looks to be the same on the dotcom version as well.

## Steps to reproduce the issue

1. Visit this page - _VPN requires_ (https://feat-hc-2021-gwktmubqkviyou8ytu4emiozocilov2b.w3-origin-007522.intranet.ibm.com/ar/cloud/watson-assistant)
2. Or test the bug locally.
3. set `STORYBOOK_IBMDOTCOM_WEB_COMPONENTS_USE_RTL=true` in your `packages/web-components/.env` file

## Additional information

![Screen Recording 2021-07-16 at 3 41 33 PM](https://user-images.githubusercontent.com/1815714/126000705-25842f7e-fba9-4004-8850-70dcc4a7ea4f.gif)

![Jul-16-2021 15-49-28](https://user-images.githubusercontent.com/54281166/126001374-40f65a8d-2f6f-48e1-8a9a-cbeee9168cec.gif)

jeffchew:
This will need to be included in the next release candidate for `v1.22.0`.

PR #6681, opened 2021-07-20
fix(overflow): ensure overflow looks and works properly in RTL

annawen1:
### Related Ticket(s)

[Masthead (dotcom & Cloud)]: Overflow logic is behaving weirdly on RTL #6672

### Description

The overflow and structure is specific to LTR; solution is to detect when the page is RTL from the html `dir` attribute and swap the structure and logic.

**FIX:**
cloud-masthead

### Changelog

**New**

- `_pageIsRTL` internal property to `top-nav` and `table-of-contents` to detect when page is in RTL

**Changed**

- swap `paginateLeft` and `paginateRight` logic when in RTL for overflow
- swap structure of caret containers and the sub-content divs when RTL is detected


jeffchew (approved):
Thank you @annawen1 and @ariellalgilmore !

Issue #6685, opened 2021-07-20
Bug on importing the CTA section sass


leticiafontoura:
## Detailed description

Importing the `@import '@carbon/ibmdotcom-styles/scss/components/ctasection/ctasection'` is throwing back an error 

![image (14)](https://user-images.githubusercontent.com/47862337/126356864-45d334ac-b97f-4de9-b62f-139b599321c0.png)

> Describe in detail the issue you're having.

> Is this a feature request (new component, new icon), a bug, or a general
> issue? Bug

> Is this issue related to a specific component? CTA Section
 
> What did you expect to happen? What happened instead? What would you like to
> see changed?

> What browser are you working in? Chrome

PR #6686, opened 2021-07-20
fix(cta-section): remove style concatenation

kennylam:
### Related Ticket(s)

#6685 

### Description

Remove Sass concatenation for web components styles that is causing build errors.

### Changelog

**Changed**

- change `__cta` to `.#{$prefix}--cta-section__cta`


jeffchew (approved):
Thanks @kennylam !

PR #6688, opened 2021-07-20
chore(feature-card): update sandbox to include css custom props flag

IgnacioBecerra:
### Related Ticket(s)
None

### Description
Codebox Example wasn't styling properly due to not having the CSS custom properties flag set, this PR ensures its on.

### Changelog

**New**

- added CSS Custom property flag within FeatureCard Sandbox Example


ariellalgilmore (approved):
thank you!!

jeffchew (approved):
LGTM!

We should probably add to all of the codesandbox examples eventually.

Issue #6687, opened 2021-07-20
replace all the VideoPlayerAPI instances with the KalturaPlayerAPI


guilhermelMoraes:
## Detailed description

Recently, we've deprecated the `VideoPlayerAPI` in favor of the `KalturaPlayerAPI`, but components are still referencing the deprecated API. 

> Is this issue related to a specific component?

All the components using the `VideoPlayerAPI` (a lot considering react and web components).

> What did you expect to happen? What happened instead? What would you like to
> see changed?

The components' behavior should stay the same. This update is code-related only.

PR #6705, opened 2021-07-21
chore(VideoPlayerAPI): replacing references with KalturaPlayerAPI

guilhermelMoraes:
### Related Ticket(s)

replace all the VideoPlayerAPI instances with the KalturaPlayerAPI #6687

### Description

Not all instances of the `VideoPlayerAPi` could be replaced. Mostly because the `KalturaPlayerAPI` has a different implementation of the `getMediaDuration` method. For the deprecated API, it's converting time from milliseconds to hours but for the Kaltura API, it's converting from seconds to hours.

### Changelog

**New**

- {{new thing}}

**Changed**

- {{changed thing}}

**Removed**

- {{removed thing}}


jeffchew (approved):
LGTM!

annawen1:
Hey @guilhermelMoraes looks good so far. 

> Not all instances of the VideoPlayerAPi could be replaced. Mostly because the KalturaPlayerAPI has a different implementation of the getMediaDuration method. For the deprecated API, it's converting time from milliseconds to hours but for the Kaltura API, it's converting from seconds to hours.

Since the `VideoPlayerAPI` will eventually be removed, we should make sure that things won't break/are properly documented. I'm wondering if we should create another function within `KalturaPlayerAPI` to convert time from milliseconds to hours? Then that function can be used in the instances you mentioned couldn't be replace due to that.

There's other places besides the components where we will need to replace the `VideoPlayerAPI` references as well - theres unit tests and READMEs, also in the `services-store` package where `VideoPlayerAPI` still being used for web-components

jeffchew:
> Hey @guilhermelMoraes looks good so far.
> 
> > Not all instances of the VideoPlayerAPi could be replaced. Mostly because the KalturaPlayerAPI has a different implementation of the getMediaDuration method. For the deprecated API, it's converting time from milliseconds to hours but for the Kaltura API, it's converting from seconds to hours.
> 
> Since the `VideoPlayerAPI` will eventually be removed, we should make sure that things won't break/are properly documented. I'm wondering if we should create another function within `KalturaPlayerAPI` to convert time from milliseconds to hours? Then that function can be used in the instances you mentioned couldn't be replace due to that.
> 
> There's other places besides the components where we will need to replace the `VideoPlayerAPI` references as well - theres unit tests and READMEs, also in the `services-store` package where `VideoPlayerAPI` still being used for web-components

@annawen1 good catches! I'll retract the approval on my end til those are all addressed.

annawen1 (approved):
LGTM! Thanks @guilhermelMoraes for making all those changes!

Issue #6276, opened 2021-06-04
[tabs extended] without expressive theme


oliviaflory:
## Detailed description

Removing expressive caused some changes in tabs extended, this issues fixes that and aligns tabs extended with the approach to the future of expressive theme, utilizing the 14px type size.

**Desktop**
- [ ] update selected state tab label to $productive-heading-01
- [ ] update enabled state tab label to $body-long-01

<img width="1101" alt="Screen Shot 2021-06-04 at 4 28 06 PM" src="https://user-images.githubusercontent.com/45692622/120859249-f21de380-c551-11eb-82d1-7e24f0dd23ca.png">

**Mobile**
No changes

## Additional information

[PR removing expressive](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/6127)
[Web components](https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/6127/index.html?path=/story/components-tabs-extended--default)
React N/A

PR #6716, opened 2021-07-23
fix(tabs-extended): item token update

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6276

### Description

Added `body-long-01` to tab items. Selected items already using `productive-heading-01` and at 14px.


NSDrowned:
@oliviaflory I fixed the text alignment I found 👍

packages/styles/scss/components/tabs-extended/_tabs-extended.scss
```diff
@@ -48,6 +52,7 @@
             @include type-style('productive-heading-01');
 
             color: $text-primary;
+            line-height: 1.375rem;
```
annawen1, line 55: The font tokens should already set the `line-height`, this is something we want to overwrite?

annawen1 (commented):
code-wise looks good, just one question about the `line-height` property being set

packages/styles/scss/components/tabs-extended/_tabs-extended.scss
NSDrowned: @annawen1, the vertical text alignment was a bit off in the selected tab. It seems the line-height set by `productive-heading-01` is different than the other tabs. This is how it looks like without the overwrite:

image

It's 1px highter.

oliviaflory: @NSDrowned thanks for fixing the text alignment! 

I think we could remove the line height override by switching the tab items to use $body-short-01. Carbon has some funky pairings with the heading and body styles that I didn't think about when I wrote the issue initially. 

Apologies for the back and forth, would we be able to remove the line height override in the heading style and switch from $body-long-01 to $body-short-01 ?

NSDrowned: @oliviaflory I switched to `body-short-01` but still have the same problem. The line-height problem seems to lie in `productive-heading-01`.

NSDrowned: Talked with Olivia over slack and we're going with the override for now 👍

oliviaflory:
@NSDrowned overall looks really good!

I wasn't expecting the tabs Content type size to change between desktop and mobile (small & medium breakpoints)
Can we keep the small and medium at 16px (likely $body-long-02)

Screen Shot 2021-08-04 at 9 32 26 AM

oliviaflory (approved):
LGTM, thanks @NSDrowned ! 

Let's work together to open an issue for the Carbon type token line height error and make a note to come back and remove the overrides in Tabs extended later 👍

Issue #6704, opened 2021-07-21
[Cloud Masthead]: Megamenu height is not displaying correctly within the Drupal integration


proeung:
## Detailed description

- While testing the new `<dds-cloud-masthead-container>` Web Components in Drupal, I noticed that the megamenu height is inheriting `min-height: 20rem;` instead of the height shown within the Storybook preview.

## Steps to reproduce the issue

1. Visit this Drupal preview page (VPN requires) - https://wc-masthead-jfbm8ry4yxvz0jplleccg7urlffpem5u.w3-origin-007522.intranet.ibm.com/cloud/watson-assistant
2. Click on the "Products" mega menu item
3. The height is not the same as what is being shown here - (see - http://ibmdotcom-web-components-experimental.mybluemix.net/?path=/story/components-cloud-masthead--default)

![Screen Recording 2021-07-21 at 2 07 40 PM](https://user-images.githubusercontent.com/1815714/126538969-76e16c4d-ba42-40dd-b66c-3e705c9d66a7.gif)
![Screen Recording 2021-07-21 at 2 09 54 PM](https://user-images.githubusercontent.com/1815714/126538989-ef0d7d37-3a81-46e8-853f-ca331ca7b516.gif)

## Additional information

- HC Engineer @proeung will be working on this ticket.
- JIRA ticket: https://jsw.ibm.com/browse/HC-2087

proeung:
@jeffchew I'm currently working on this ticket and will have a PR for review soon.

PR #6718, opened 2021-07-23
fix(megamenu): adjust height for megamenu to work for cloud's drupal integration

annawen1:
### Related Ticket(s)

[Cloud Masthead]: Megamenu height is not displaying correctly within the Drupal integration #6704

### Description

The cloud megamenu is not displaying properly height when expanded in their Drupal integration. The solution is the set the height so it displays properly.

### Changelog

**New**

- set the height to take 100vh - the margin bottom and menu height


ariellalgilmore (approved):
thank you!!!

Issue #6697, opened 2021-07-21
[RTL]: web component package and CDN not deploying rtl styles


ariellalgilmore:
## Detailed description

> Describe in detail the issue you're having.
related to issue #6672. The RTL styles are somehow being stripped from the static builds. For example the deploy previews are rendering the overflow carets in the wrong place without the RTL styles in the scss. We believe this is most likely a webpack issue. This issue is also visible when using the CDN.

<img width="1230" alt="Screen Shot 2021-07-21 at 9 43 29 AM" src="https://user-images.githubusercontent.com/20210594/126526907-5ef4392a-b691-4dee-9310-61e995a7f381.png">

 

> Is this a feature request (new component, new icon), a bug, or a general
> issue?
bug

> Is this issue related to a specific component?
no web component package in general

> What did you expect to happen? What happened instead? What would you like to
> see changed?

see the rtl styles applied in deploy previews and not just on local machine

> What browser are you working in?

> What version of Carbon for IBM.com are you using?

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

- Screenshots or code
- Notes

jeffchew:
@kennylam @annawen1 @ariellalgilmore would this be related? https://github.com/MohammadYounes/rtlcss/issues/221

PR #6729, opened 2021-07-26
fix(rtl): fix to not have rtlcss styles get parsed out

jeffchew:
### Related Ticket(s)

Refs #6697 

### Description

This adds some updates to change the declaration for rtl in scss files so they don't get parsed out.

### Changelog

**Changed**

- Various RTL style declaration changes


ariellalgilmore (approved):
THANK YOU!!!

Issue #6733, opened 2021-07-27
Web component: Card section offset - Cards are not operable with keyboard Tab.


Praveen-sr:
## Detailed description

On FF
Issue:
1. Keyboard Tab focus is not going to the cards so the user is not able to click them.
2. When we navigate with up/down arrow keys, Screen reader announcing links without any label just before the (Eyebrow) text i.e. "label" and at the end of the card. Note visually these links are not operable with the screen reader.

Expected:
1. Tab focus should go to each and every interactive element and it should be keyboard operable.
2. Screen reader is expected to identify the links with appropriate label and the user should be able to activate them.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/293cdff5-b6ba-48c5-bbad-bf32b3b2c26e)

Praveen-sr:
@ariellalgilmore Re tested above issue on both iPhone VO, JAWS and Keyboard focus, Working as expected.
Thanks!

PR #6734, opened 2021-07-27
fix(card-section-offset): update to use card-cta-footer

ariellalgilmore:
### Related Ticket(s)

#6733 
### Description

update components to use `card-cta-footer` instead of just `card-footer` when applicable.


jeffchew (approved):
thank you @ariellalgilmore !

Again, I'm anticipating some confused adopters when the rc and full release goes out for folks that are upgrading, so we'll need to put in some heavy messaging around this change for `v1.23.0`.

Issue #6077, opened 2021-05-13
Web component: Dotcom shell - Locale modal- when we activate "Select geographic area" link, the First modal is not displayed


Praveen-sr:
## Detailed description

On iPhone Voice over
Issue:
From the second modal, when we activate "Select geographic area" link, the First modal is not displayed. Instead the screen reader focus goes to the parent page.

Note: we understood from visual users that first modal appears momentarily and then disappears.
Note: this issue is not found in react staging.

Expected:
Screen reader should identify the first modal

PR #6735, opened 2021-07-27
fix(locale-modal): set region item to be tabbable for voiceover

annawen1:
### Related Ticket(s)

Web component: Dotcom shell - Locale modal- when we activate "Select geographic area" link, the First modal is not displayed #6077

### Description

When using voiceover on iphone, the user isn't taken back to the first modal with all the geographic regions (Americas, Asia Pacific, etc) after selecting the "Select geographic area" link to go back. The "Americas" region card is briefly focused before voiceover announces the `select` component. 

Solution is the set the `tabIndex` of the first `<dds-region-item>` component in the first modal. This way voiceover recognizes the component as tabbable and will then focus properly on it.

### Changelog

**Changed**

- set `tabIndex` of the first region item to 0 to ensure voiceover focuses on it


ariellalgilmore (approved):
LGTM! Thank you @annawen1! Tested on my iphone with voiceover and focus state went back to region item

PR #6751, opened 2021-07-29
fix(megamenu): remove overflow auto

ariellalgilmore:
### Related Ticket(s)

#6752
### Description

`overflow: auto` causing error on application side. Changed to just remove the `overflow` if megamenu is not open


jeffchew (approved):
LGTM!

Issue #6709, opened 2021-07-21
[Footer] Locale Button always defaults to United States with any English locale


IgnacioBecerra:
## Detailed description

> Describe in detail the issue you're having.

When using any other English locale, the Web Components Footer Locale Button always defaults to United States instead of using the locale's country. For example, using `en-uk` should display "United Kingdom - English", or using `en-au` would display "Australia - English", but it always displays "United States - English"

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

A bug

> Is this issue related to a specific component?

Web Components Footer

> What did you expect to happen? What happened instead? What would you like to
> see changed?

The button should display the proper locale's country.

> What version of Carbon for IBM.com are you using?

Happening in Canary, Staging, Prod

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

Issue was brought up by the Adobe CMS team
https://cognitive-app.slack.com/archives/G01JR9YHU3W/p1626881735110900

## Steps to reproduce the issue

1. Ensure the html lang attribute is set to `en-au`
2. Updating the DDO to this locale also works
3. Check the button, should display as "United States - English" no matter the English locale

## Additional information

This may be happening somewhere in `<dds-footer-container>` or `<dds-footer-composite>` with the Redux LocaleAPI calls. Querying the regular LocaleAPI within `<dds-footer-composite>` and setting the results to `this.langDisplay` makes it display properly, this requires further investigation to see if it's a good solution in general.

PR #6770, opened 2021-07-30
fix(footer): remove redux langDisplay

ariellalgilmore:
### Related Ticket(s)

#6709 
### Description

Locale Button always defaults to United States with any English locale. Removed redux and directly use the localeAPI service to load the langDisplay for the locale button. 

### Changelog

**Changed**

- removed redux 
- updated `firstUpdated()` and added `async getLangDisplay`


ariellalgilmore:
Not sure why percy thinks there is a change in the locale display. The current [canary version ](https://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-footer--default)matches with the deploy previews. 
Screen Shot 2021-07-30 at 11 09 14 AM

jeffchew (approved):
Nice one, definitely seems much simpler!

IgnacioBecerra (approved):
Thanks for making this change @ariellalgilmore!

Issue #6022, opened 2021-05-11
Accessibility - label missing for language selector


kyjak:
this element in the footer has no associated label.
please note when adding labels, these need to be translated to all DOCS languages, we can help with that, 
and include the label in our translation request.

`<select class="bx--select-input" id="bx--language-selector" text="English" data-autoid="dds--language-selector__select">`

RobertaJHahn:
@kyjak  Could you provide a link to where you are seeing this issue?  Also provide the names of the tool(s) you are using to identify this issue?  We are forwarding this bug to our accessibility team in India for review.

kyjak:
you can use our dev https://ibmdocs-dev.mybluemix.net/docs/en
and the tool to check errors is "IBM Equal Access Accessibility Checker", plugin for chrome

jeffchew:
@Praveen-sr can you review?

Praveen-sr:
@jeffchew @RobertaJHahn 
Footer - Language selector dropdown ( On JAWS FF & iPhone Voiceover)
Issue: There is no label for this dropdown.
Expected: Provide label for the dropdown using html label tag or aria label.

RobertaJHahn:
@Praveen-sr You mentioned in our meeting yesterday that this bug is a dup of another created by you.  Could you find this issue and make a note of the issue number here?  Thanks.

Praveen-sr:
@RobertaJHahn I searched our board for the issue opened during Oct 2020, but wasn't successful. Once the issue is closed on github, we are unable to see them.

Gopi916:
Issue is repro'ing in the below environment. Hence reopening the defect.

Carbon for IBM.com version
v1.34.0-rc.0

Application/website
Carbon for IBM.com

Package
@carbon/ibmdotcom-web-components

CodeSandbox example
https://www.ibm.com/standards/carbon/web-components/?path=/story/

IgnacioBecerra:
Might be fixed with #9086

PR #6782, opened 2021-08-02
fix(language-selector): added missing label to select

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6022

### Description

Added missing `aria-label` to `select` in React.


jeffchew (approved):
Thanks @NSDrowned !

PR #6809, opened 2021-08-04
fix(card-section-offset): background with 100%

ariellalgilmore:
### Related Ticket(s)

none 
### Description

just noticed the card section offset component background image was not going full width. Not sure what changed? but adding width 100% fixed it. 

### Changelog

**New**

- width: 100%


IgnacioBecerra (approved):
Weird that it was already going full width before adding this, but thanks for the fix @ariellalgilmore!

Issue #5705, opened 2021-04-05
[Link with icon] Web components: icon position when text wraps to 2 lines 


shixiedesign:
Currently link with icon is not showing the correct behavior when link text wraps to 2 lines. Icon snaps to right-aligned to the container, and creates a weird gap between text and icon. This could happen when text is too long, or when screen size is narrow, and is influencing several dependent components such as `Link list`, `Link list section`, and `Content group banner`

This is similar problem to (#5740 )

#### Current

![Apr-02-2021 16-52-42.gif](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/daeba0c3-fb50-4005-8468-6913d6923c9d)

![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/7d7a419e-8c4d-42e1-9050-d6f371b4cf06)

## Proposed design & spec

![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/28c55ded-f65e-4a13-abf5-92216e57dcbd)

jeffchew:
@kennylam going to unassign this from you since you're on vacation for the rest of the sprint, and I want you to focus on Beacon next sprint.

emyarod:
@shixiedesign just want to clarify some of the text wrap behavior I'm currently testing, does this recording match the expected behavior outlined in the spec? depending on the viewport size the icon can wrap to the next line before the last word does

https://user-images.githubusercontent.com/8265238/127699900-c8084648-3f9c-40db-b1bd-6ee8aeaf5fc8.mov

shixiedesign:
Hey @emyarod yep that is the expected behavior. Unfortunately there are awkward moments (when 1 arrow breaks to line 2 before the other one). 

I think if there's possibility to have an "orphan" logic so the arrows is always gonna be attached to 1 word with it, it could be nice. Lemme know if that's even possible. If so, I can add a feature request for the team to look at.

emyarod:
that behavior is tricky but should be possible (although it may be a breaking change). I will need to test more and see if it can be implemented with minimal changes to the existing component

RobertaJHahn:
Bug must also be fixed in React.  Issue (#6837 ) has been created to manage that work.

PR #6812, opened 2021-08-04
fix(link-with-icon): add icon line wrap behavior by default

emyarod:
### Related Ticket(s)

#5705

### Description

This PR changes the default Link with icon component behavior so that the icon will wrap to the next line the component length gets too long

Verify that icons in the following components/variants will wrap with longer link text

* Callout link with icon
* Card footer
* Content block horizontal
* Content block segmented
* Content group banner
* CTA - Text
* Link list (horizontal)
* Link list (vertical)
* Link with icon

image

and Link list section, Link list (default), Link list (End Of Section)  icons do not wrap with longer text (behavior remains unchanged)

image

Since the Link with icon component is a dependency for several other components, also verify that components such as callout link with icon and megamenu link with icon appear correct

### Changelog

**New**

- nonreflected `iconInline` property on Link with icon component (extended to Link list and Card link footer)

**Changed**

- icon line wrap behavior on link with icon and several dependent components


jeffchew:
This diff doesn't seem to look right @emyarod:

image

emyarod:
@jeffchew this is the behavior I mentioned where the icon can wrap before the final word. not sure if we can achieve the original desired behavior without modifying the markup or resorting to some css hacks but I checked with Shixie on if we still want to proceed with this cc @shixiedesign

shixiedesign:
Hey @jeffchew yes the wrapping behavior caught by Percy is expected. 

This bugfix should fix many stances of the wrong wrapping. Currently I'm still seeing the incorrect icon wrap in `CTA`, `Content block segmented`, `Content block horizontal`. I cannot easily test all the content blocks for CTA. Does this mean `Content blocks` are not dependent on Link with icon? Lemme know if these are out side the scope of this PR. 

### Still seeing the wrong icon wrapping behavior in the following components: 

**CTA - Text** 
image

**Content block horizontal**
image

**Content block segmented**
image

emyarod:
> Currently I'm still seeing the incorrect icon wrap in `CTA`, `Content block segmented`, `Content block horizontal`. I cannot easily test all the content blocks for CTA. Does this mean `Content blocks` are not dependent on Link with icon?

from what I can tell they do not depend on Link with icon, and I can look into adding the same behavior to those components. I'm not sure on how we want to scope these changes but for easier testing it may be good to split it up since I'm still unfamiliar with which components depend on which. Interested in hearing more thoughts on if/how we should split up the PR!

emyarod:
CTA - Text, Content block horizontal, and Content block segmented should also see the same text wrap behavior now!

shixiedesign:
Noticed React needs these update as well. Already discussed on Slack.

emyarod:
the React versions should now receive the same changes!

shixiedesign (approved):
Awesome! Thank you for the fix!

jeffchew (approved):
Thank you @emyarod !

RobertaJHahn:
Issue (#6837 ) has been created to manage the React work mentioned in Andrew's comment above.

Issue #6811, opened 2021-08-04
[Services - Masthead/Footer Translations] Translations defaulting to en-US if there is no corresponding country list


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

If there isn't a corresponding country list for a lc/cc combination, the translations will default to us-EN. 

> Is this issue related to a specific component?

Masthead/Footer

> What did you expect to happen? What happened instead? What would you like to
> see changed?

The translation data should be separate from the countrylist. It shoudl be ok to use the en-US country list if one doesn't exist for the current lc/cc combination, but it should definitely still load in the corresponding lc/cc localized navigation content.

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

This issue was also brought up by Hybrid Cloud team.

## Steps to reproduce the issue

1. Can see it reproducable here: https://ibmdotcom-react.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/6802/iframe.html?id=components-dotcom-shell--default&cc=sa&lc=ar
2. Notice the Arabic Masthead/Footer is not getting loaded.

PR #6819, opened 2021-08-05
fix(locale): remove countrylist check for translation

annawen1:
### Related Ticket(s)

[Services - Masthead/Footer Translations] Translations defaulting to en-US if there is no corresponding country list #6811

### Description

Problem: 
Translations for locales that do not have a corresponding countrylist file default to `us-en`.

Solution:
Remove the countrylist check in the `LocaleAPI` and allow the locale to be passed to the `TranslationAPI`. If a countrylist file for the given locale doesn't exist, make another call to retrieve the default `us-en` countrylist. If in the case there is no translation file for the locale, it will also make another call to the `us-en` translation file.

**Expected Result:**
If adopter tries to get translation for non-existent locale (ie. `eh-ar` as show in the screenshot) it will make a call to both the countrylist and translation file with the locale. Once those fail, it will return the default `us-en` translation and countrylist instead.
Screen Shot 2021-08-05 at 10 38 03 AM

### Changelog

**Changed**

- pass locale from the `LocaleAPI.getLang()` without checking against countrylist
- ensure the `fetchTranslation()` call is made again if the translation call fails


jeffchew (approved):
Nice fix, thank you!

kennylam (approved):
LGTM thanks @annawen1!

emyarod (approved):
looks good to me!

Issue #6501, opened 2021-06-29
[Card link] and [Tabs extended] changes to support AEM Geo home pages 


skuruvil:
### The problem
One of the templates that have been prioritized is the IBM geo home page. Using https://www.ibm.com/us-en?ar=1  as a reference, one of the components on the page is essentially a list of links that allow users to access different portions of the site. What we want to use going forward is basically a full-width link list with tabs

### The solution
Created this new layout component design using variants of existing components Tabs extended vertical, Card link, and Content-group (for heading) as a baseline. This component is going to be used in IBM geo home page design.

https://ibm.ent.box.com/folder/140172923780

### Additional information

Content model: https://ibm.box.com/s/2d9wyvho8ue2kgx3sszvjq7103jcf7ln

 - Adding vertical tabs and a modified card link
 - Hybric Cloud Jira ticket link:  https://jsw.ibm.com/browse/HC-2066
 - [Card link functional specs](https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/wiki/Card-link)
 - [Tabs extended functional specs](https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/wiki/Tabs-extended)

**Examples from design spec:**

<img width="1193" alt="Screenshot 2021-06-29 at 7 17 46 PM" src="https://user-images.githubusercontent.com/71765369/123809037-e33b0e80-d90e-11eb-8ea6-bf3be765b14d.png">
<img width="702" alt="Screenshot 2021-06-29 at 7 18 12 PM" src="https://user-images.githubusercontent.com/71765369/123809364-2bf2c780-d90f-11eb-8cfd-59cf6fde59f5.png">
<img width="200" alt="Screenshot 2021-06-29 at 7 18 30 PM" src="https://user-images.githubusercontent.com/71765369/123809424-38772000-d90f-11eb-9d35-809ddfd4190e.png">

RobertaJHahn:
Convert to epic, func specs, dev issues (WC & React Wrapper), all QA & website

RobertaJHahn:
Per 7/12/2021 Slack with Sandeep.  This request is a high priority.  They would like the feature to be included in the Aug 17 release.

RobertaJHahn:
Card link and Tabs extended dev will be  done by HC.

Card LInk WC sizing:  3
Card link React sizing:  set up planning poker

proeung:
[...]

PR #6826, opened 2021-08-05
feat(tab-extended-vertical): add "Vertical" Orientation

proeung:
### Related Ticket(s)

DDS Consulting Issue: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6614
HC JIRA Ticket: https://jsw.ibm.com/browse/HC-2066

### Description
- This PR adds in the "Orientation" prop that allows adopters to set the orientation of the Extended Tabs layout component to display as `horizontal (default)` or `vertical`.

### Design & Functional Specs
- https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/wiki/Tabs-extended#22-orientation 

Screen Shot 2021-08-05 at 3 45 33 PM

### Changelog

**New Components/Props**

- `Orientation` - this prop/attribute allows adopters to set the orientation of the Extended Tabs component to horizontal (default) or vertical.

**Removed**

- N/A

### Note

- The layout component/band that's designed for the GEO page in AEM requires `card-link` component, however, this particular card component doesn't have the same functional specs as the `<dds-card-group-item>` component, which can make it difficult to assemble within AEM.
- We might want to possible extend the `<dds-card-link>` component to the following:
```
<dds-card-group>
   <dds-card-group-link-item href="https://example.com">
   </dds-card-group-link-item>
</dds-card-group>
```

### Testing Instructions
- Visit the `Tab Extended` layout component - `/?path=/story/components-tabs-extended--default`
- Navigate to the "Knobs" tab and set `Orientation` to `vertical`
- Ensure that the "Vertical" tabs component matches with the design specs for all breakpoints.


proeung:
@shixiedesign @photodow While testing the assembling of the Tabs Extended to look like this Geo home layout component (see attached). I noticed some follow-up work that we need to do on the WC package side (see my "note" section - https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/6826#issue-704987352). Let me know what you both think.

proeung:
@emyarod Thanks for the review! I just pushed up a commit that should fix the hover issue that you mentioned above.

@shixiedesign or @photodow Any thoughts about what I mentioned here - https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/6826#issuecomment-893770697?

shixiedesign:
Hi Putra! Sorry I'm still a bit distracted from moving. Re your note above, do you mean that the current dds Card group component cannot accept Card link? If so, we should make a fix to Card group, so that the card group can arrange card links on a grid just like regular Cards. If my understanding is correct, please feel free to open an issue. 

Taking a look at the component now.

shixiedesign:
Hey @proeung ! Thanks! It looks great. I'm noticing the specs for it isn't exactly complete and I will patch that hole momentarily... sorry about that, but you did a great job interpreting the states of the component based on existing Tabs extended! I think it looks good, and only a few comments: 

1. `max - lg` breakpoint: content area is not aligned to grid: 
image

2. Currently the first tab is showing Card group in content area. Ideally we use a simple string in content area to indicate it's just an empty slot and can be used for many different components. 
image

3. `md` breakpoint, accordion currently have a bug (in horizontal tabs too). Expecting the accordion horizontal rule to hang both sides: 
image

4. Same comment on `sm` breakpoint
image

5. Motion should be great if we can animate the underlines of the tabs same as how Hybrid cloud library has it. I believe it was something like this: 
tabs-vertical

proeung:
@shixiedesign Thanks for the review! Please see my response below.

> Hi Putra! Sorry I'm still a bit distracted from moving. Re your note above, do you mean that the current dds Card group component cannot accept Card link? If so, we should make a fix to Card group, so that the card group can arrange card links on a grid just like regular Cards. If my understanding is correct, please feel free to open an issue.

No worries! My current understanding is that the current `<dds-card-group>` component does not accept the card-link component. We might want to possibly extend the <dds-card-link> component to the following (see below). I think we can capture this change in a separate ticket but just wanted to bring it up as it's going to be difficult for AEM or other adopters to assemble the band/section outlined here - https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6501.

```
<dds-card-group>
   <dds-card-group-link-item href="https://example.com">
   </dds-card-group-link-item>
</dds-card-group>
```

> max - lg breakpoint: content area is not aligned to grid:

Fixed

> Currently the first tab is showing Card group in content area. Ideally we use a simple string in content area to indicate it's just an empty slot and can be used for many different components.

I adjusted the SB preview to include the simple text sting, however, I think from a library documentation perspective, it's still good to showcase to users (eg. designers or UXers) the capability of this component where you can add other UI components to make a band/component layout. Perhaps, this is something we should consider adding to the design guidelines/docs if there's any.

> md breakpoint, accordion currently have a bug (in horizontal tabs too). Expecting the accordion horizontal rule to hang both sides:

Fixed

> Same comment on sm breakpoint

Fixed

> Motion should be great if we can animate the underlines of the tabs same as how Hybrid cloud library has it. I believe it was something like this (on hover text darken & underline grows. onClick previous blue underline fades out quickly 90ms and the new selected item's underline turns blue).

Happy to add this animation, however, I just wanted to note that the current underlines are using the `border-bottom` property and if we want to add this left-to-right animation, we'll need to refactor the current code that's being used within the tab items for both orientations (Horizontal and Vertical), unless you just want to scope this to just the vertical tabs. Also, we'll need to adjust the hover direction on RTL as well. Anywho, let me know what you think.

kennylam (approved):
LGTM thanks @proeung! Looks like you just need to update the unit test.

shixiedesign:
@proeung Ha fantastic point for RTL getting the other dir animation. Let's avoid too much refactor and I can add motion to a future ticket! 

Okay, I just open a ticket for Card group modification. Please review if the description is accurate. https://app.zenhub.com/workspaces/carbon-for-ibmcom-5d449f3642eb1962336cbe52/issues/carbon-design-system/carbon-for-ibm-dotcom/6876

> Perhaps, this is something we should consider adding to the design guidelines/docs if there's any.

Yes this example will be captured in the website!

shixiedesign:
Also gonna tag @skuruvil to review this one moving forward.

proeung:
@shixiedesign The 3 items that you've noticed should be fixed now. Please take a look when the deploy preview is ready.

skuruvil:
@proeung couple of updates

1. Tabs label should wrap to 2 lines at 3 columns (not at 4 columns)
Screenshot 2021-08-17 at 9 29 50 PM

2. Based on the discussion with Shixie we have agreed to update the space between the labels to 24px

Screenshot 2021-08-17 at 9 33 03 PM

3. We have updated the type tokens for labels 
Tab label selected : `$productive-heading-02`, `$text-01 `
Tab label enabled: `$body-short-02`, `$text-02 `
Tab label hover: `$body-short-02`, `$text-01 `
Here is the updated visual spec [](https://ibm.ent.box.com/file/847598629410 )

Note: **We may need to do these updates for Tab Extended Horizontal too**

Screenshot 2021-08-17 at 10 03 38 PM

Thank you

skuruvil:
@proeung looks great to me. I don't have any further changes, Thank you

proeung:
@skuruvil Wonderful!

@jeffchew @annawen1 @ariellalgilmore Can I get a code review so that we can get this PR merged down?

ariellalgilmore (approved):
LGTM!

Issue #6750, opened 2021-07-29
[Masthead RTL] - issue with overflow in react


ariellalgilmore:
## Detailed description

> Describe in detail the issue you're having.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

bug

> Is this issue related to a specific component?

masthead

> What did you expect to happen? What happened instead? What would you like to
> see changed?

rtl overflow does not work
<img width="1237" alt="Screen Shot 2021-07-29 at 6 41 09 AM" src="https://user-images.githubusercontent.com/20210594/127502510-4eece63a-ec7b-4792-905b-62aff1642685.png">

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

- Screenshots or code
- Notes

PR #6828, opened 2021-08-05
fix(masthead): react rtl overflow

annawen1:
### Related Ticket(s)

[Masthead RTL] - issue with overflow in react #6750

### Description

Problem: 
Masthead overflow functionality was not working in RTL. 

Solution:
Detect when the page is RTL and swap the overflow logic.

overflow rtl

### Changelog

**Changes**

- add overflow logic for `paginateRight()`, `paginateLeft()`, and for the tabbing in rtl


packages/react/src/components/Masthead/HeaderNavContainer.js
```diff
@@ -30,75 +30,140 @@ const HeaderNavContainer = ({ children }) => {
   const [io, setIO] = useState(null);
   const [position, setPosition] = useState(0);
   const buttonSize = 48; // 40px(width) + 8px(gradient)
+  const pageIsRTL = document.dir === 'rtl';
```
jeffchew, line 33: @annawen1 this is causing errors in NextJS

image

packages/react/src/components/Masthead/HeaderNavContainer.js
```diff
@@ -155,25 +221,49 @@ const HeaderNavContainer = ({ children }) => {
...
-            position <=
-            buttonSize
-        ) {
-          paginateLeft();
+      if (pageIsRTL) {
+        if (event.shiftKey) {
+          if (
+            document.activeElement.parentElement.previousSibling &&
+            document.activeElement.parentElement.previousSibling.offsetLeft +
+              document.activeElement.parentElement.previousSibling.offsetWidth -
```
emyarod, line 229: just one question, would we be able to use optional chaining here and also on L238/L240?

annawen1: sure, I can add that

jeffchew (approved):
LGTM!

Issue #6831, opened 2021-08-06
[Cloud Masthead] Slight page shift when the Mega Menu is opened on machines where scrollbar is set to visible


proeung:
## Detailed description

- On machines (Mac OS/Windows) where the scrollbar is set to visible, there's a slight page shift when the mega menu dropdown is opened.

https://user-images.githubusercontent.com/54281166/128551883-736e8566-987b-4fa4-8847-0f5d250c16ff.mp4

- This is due to this logic where we're setting the `margin-right` value (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/packages/web-components/src/components/masthead/megamenu-top-nav-menu.ts#L111), so that the profile and search icons don't shift on the right (see attached) when the `dds-masthead` component position is `position: fixed`. In the Cloud test page, the `dds-masthead` component position is set to `position: static` to account for the L1. This causes the buttons and icons to shift as it is adding the `margin-right` to both the `body` and `masthead` element.

![Aug-06-2021 11-02-04](https://user-images.githubusercontent.com/1815714/128549947-22f84914-fd97-41ac-8389-b321c21d1df7.gif)

## Steps to reproduce the issue

1. Visit this Drupal preview page (VPN requires) - https://pr1481-1n16qxtlhmeiub0flqa89rbjxqvpxst8.w3-origin-007522.intranet.ibm.com/cloud/watson-assistant
2. Click on the "Products" mega menu item

## Purposed solution
- Add a "fixed" position to `dds-masthead` when the Mega Menu dropdown is opened and scope the changes to just `<dds-cloud-masthead-container>`. This will hopefully override the `position: static` that has been set in the Cloud page.

PR #6836, opened 2021-08-09
fix(cloud-masthead): set position of masthead to fixed when expanded

annawen1:
### Related Ticket(s)

[Cloud Masthead] Slight page shift when the Mega Menu is opened on machines where scrollbar is set to visible + Adjust minor icon RTL issue #6831

### Description

Cloud masthead shifts when megamenu is expanded. This is due to the combination of styles implemented from the cloud side to set the masthead `position: static` to accommodate the L1 and the `margin-right` added when the scrollbar is locked on megamenu expand. This PR sets the masthead - scoped to cloud - to `position: fixed` when megamenu is expanded. 

### Changelog

**Changed**

- set `position:fixed` to the cloud-mastead on megamenu expand


jeffchew (approved):
LGTM!

Issue #6342, opened 2021-06-10
[Masthead with L1] Sticky behavior 


oliviaflory:
## Detailed description

> Describe in detail the issue you're having.

### 1. Web components sticky behavior does not match spec
If the masthead has an L1, the L1 should be the only sticky element. The React dotcom shell shows the correct scroll behavior, where as in web components the L0 and L1 are both sticky.

**React**
L0 scrolls out of view when use scrolls down the page
[Dotcom shell](https://ibmdotcom-react-canary.mybluemix.net/?path=/story/components-dotcom-shell--with-l-1)
![masthead-react-Jun-10-2021 16-02-31.gif](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/8b2f8ca9-0f79-45c8-b9cd-d76576f7f9bb)

**Web components**
Both L0 and L1 are sticky, it should only be L1
[Dotcom shell](https://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-dotcom-shell--with-l-1)
![masthead-webcomp-Jun-10-2021 16-02-52.gif](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/b64db65b-e4b1-4790-95d8-deeb1974d326)

### 2. Scroll up behavior
Neither ~React~ (React will be addressed in #6391) or Web components are following the scroll up behavior stated in the specs:
`Masthead L0 remains hidden if L1 is present, except reveal L0 on scroll up.`
[Masthead L1 specs](https://ibm.ent.box.com/folder/118448193520?s=o0v38tot6t0s9tnr3y3estp4cnxar6vo)
> Is this a feature request (new component, new icon), a bug, or a general
> issue?

> Is this issue related to a specific component?

Masthead L1, best seen in Dotcom shell with L1 story to demonstrate scroll

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Feature parody between React & WC, also update scroll up behavior

> What browser are you working in?

> What version of Carbon for IBM.com are you using?
[React canary](https://ibmdotcom-react-canary.mybluemix.net/?path=/story/components-dotcom-shell--with-l-1)
[WC canary](https://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-dotcom-shell--with-l-1)

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. Step one
2. Step two
[...]

PR #6838, opened 2021-08-10
fix(masthead): sticky behavior when l1 is present

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6342

### Description

If the masthead has an L1, the L1 should be the only sticky element.

### Changelog

- Also fixes `Masthead L0 remains hidden if L1 is present, except reveal L0 on scroll up ` which now happens at any scroll position, not just at the top

### DIFF
--- a/packages/web-components/src/components/dotcom-shell/dotcom-shell-composite.ts
+++ b/packages/web-components/src/components/dotcom-shell/dotcom-shell-composite.ts
@@ -137,16 +137,26 @@ class DDSDotcomShellComposite extends LitElement {
    * Scrolls the masthead in/out of view depending on scroll direction if toc is present
    */
   private _handleIntersect = () => {
-    // masthead hide l0 on scroll if l1 is present
-
+    /**
+     * Sets sticky masthead. If both L0 and L1 are present, L1 will be sticky.
+     *
+     */
     if (this._masthead && this._masthead.querySelector(`${ddsPrefix}-masthead-l1`)) {
+      /**
+       * L0 will hide/show only in the top 25% of the viewport.
+       *
+       */
+      const hideTopnavThreshold = 0.25;
       const mastheadl1 = this._masthead.querySelector(`${ddsPrefix}-masthead-l1`) as HTMLElement;
       this._masthead!.style.transition = 'none';
+
       if (window.scrollY < this._lastScrollPosition) {
         this._masthead.style.top = `0`;
-      } else {
+      } else if (this._lastScrollPosition > window.innerHeight * hideTopnavThreshold) {
         this._masthead.style.top = `-${mastheadl1.getBoundingClientRect().height}px`;
       }
+
+      this._lastScrollPosition = window.scrollY;
     }
 
     if (this._tableOfContentsInnerBar && !this._localeModal?.hasAttribute('open')) {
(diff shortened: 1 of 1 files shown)


oliviaflory:
Hey @NSDrowned this is looking good, super exciting to see!

Desktop looks great, and I was expecting to see the same scroll behavior on the medium and small breakpoints as well because the L0 + L1 + TOC start to take up a lot of screen and feels like a 3 bar sandwich. 

**Desktop**
L1 desktop Aug-11-2021 09-32-14


**Medium & small**
L1 mobile Aug-11-2021 09-32-56

annawen1:
Hmm want to make sure the expected behavior in mobile is that the L1 will take precedence over the TOC? @oliviaflory 

https://user-images.githubusercontent.com/54281166/129735916-ea59bb28-6145-4904-b70f-3215a4a4ccc7.mov

NSDrowned:
@annawen1, that doesn't look right. This is what's needs to happen:

image

Can you try again? I had to reload the preview thanks to my browser cache to get the latest version.

annawen1:
> Can you try again? I had to reload the preview thanks to my browser cache to get the latest version.

Tried again in incognito, behaving as you explained. Thanks!

oliviaflory:
Hey @NSDrowned apologies for the delay.

Thank you for applying the logic to the medium and small breakpoints as well! 

I may have not been really clear on this before, because we have the Dotcom shell + L1 as the only place we can see the masthead logic, I forgot to address what should happen when the TOC is also present. I believe this logic is in the functional specs

1. if L0 + L1 = L1 sticky
2. if L0 + L1 + TOC = TOC sticky

It looks like the L1 and TOC are both sticky in the Dotcom shell story, and I worry that change might negatively affect some of the AEM templates. Would it be possible to fix that? 

**Current/expected**
if L0 + L1 + TOC = TOC sticky
TOC current Aug-18-2021 13-27-08


**Preview**
TOC + L1
TOC L1 Aug-18-2021 13-28-44

shixiedesign:
Confirming that the specs / expected state is only 1 thing sticky at any time, both desktop and mobile. Some additional details in this issue https://app.zenhub.com/workspaces/carbon-for-ibmcom-5d449f3642eb1962336cbe52/issues/carbon-design-system/carbon-for-ibm-dotcom-website/971

NSDrowned:
@oliviaflory ready for another review 👍

oliviaflory:
Thanks @NSDrowned this looks great when I look at on the browser in inspect. Let me know what you think about the below:

One thing I noticed when I tried it on my phone (iPhone 12 pro) on Safari and Chrome:
- Select an item from the TOC
- TOC scrolls you down to that section and the L0+L1 appears, TOC disappears
- When you start to scroll, the TOC appears again.

https://user-images.githubusercontent.com/45692622/130482871-569fc260-2d46-4468-901b-48420a17f3e3.mp4


I would have expected it to behave more like what we are seeing in inspect:
- Select an item from TOC
- TOC scrolls you down the section (Lo +L1 hide), TOC is at top of page
- TOC stays as you scroll

L1-TOC--Aug-23-2021 09-27-51

PR #6839, opened 2021-08-10
feat(card-link-copy): adding optional copy

mwanberg:
### Related Ticket(s)

DDS ticket (epic): https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6501
HC Jira ticket: https://jsw.ibm.com/browse/HC-2115

### Description

This PR adds an option to include copy text beneath the card title. It mimics what the [original card component](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/86e62a820d41f43fdd5ce7cb674b128e454c2b1a/packages/web-components/src/components/card/__stories__/card.stories.ts#L46) is doing for its optional copy.

This PR also adds a corresponding knob and updates the default demo text to match the newer comps.

Screen Shot 2021-08-09 at 9 25 50 PM

### Changelog

**New**

- N/A

**Changed**

- Added optional copy to `<dds-card-link>`

**Removed**

- N/A


packages/web-components/src/components/card-link/__stories__/card-link.stories.ts
```diff
@@ -17,10 +17,15 @@ import textNullable from '../../../../.storybook/knob-text-nullable';
...
+  const { disabled, href, copy } = parameters?.props?.CardLink ?? {};
   return html`
     <dds-card-link ?disabled=${disabled} href=${ifNonNull(href || undefined)}>
-      Explore AI use cases in all industries
+      IBM developer
+      ${copy
+        ? html`
+            <p>${copy}</p>
+          `
+        : ``}
```
proeung, line 28: I think we need to define a `copy` slot within this HTML return that allows for easy styling if needed.

`<slot name="copy"></slot>`

With this slot, you can include the following markup to the Storybook preview.

`<dds-card-link-copy slot="copy">text</dds-card-link-copy>`

@annawen1 or @ariellalgilmore Let me know if you both agree.

shixiedesign:
Hi! This looks pretty good, only 2 nitpicks: 
1. Copy please leave as empty by default. We want to default to Card link without optional copy.
2. Please use this text for Card heading "Explore AI uses cases in all industries". We don't have this done through out the library yet, but we'd like to gear towards realistic placeholder content, and CTAs should have action oriented text.

packages/web-components/src/components/card-link/__stories__/card-link.stories.ts
ariellalgilmore: Thanks @proeung and @mwanberg! Since DDSCardLink is extending DDSCard I think we actually want to keep the copy like @mwanberg implemented, but I think it would make sense to actually change the `heading` to use the `heading slot` and apply the heading styles appropriately there. Currently both the copy and heading are being rendered within the `copy div`

cc: @annawen1 

image

annawen1: I agree with @ariellalgilmore, we should utilize the `heading` slot

proeung: Makes sense. Thank you both! 

@mwanberg I would leverage the existing card heading component and adjust the type styling to use `expressive-heading-02`  within the `_card-link.scss` partial as defined in the design specs (see link -  https://ibm.ent.box.com/file/657497664059?s=pkjve2gmmldfi9dinak8ihahkct9hb9f).

mwanberg:
@proeung @shixiedesign @ariellalgilmore @annawen1 Thanks for reviewing!

1. I've moved the heading text so it lives in its own slot now
2. Updated the heading styling so it wouldn't inherit the default styling of the base `card` component
3. Removed the default knob text for the copy
4. Changed the default heading text to "Explore AI uses cases in all industries"
5. Added a knob for the heading text

Let me know if anything else is needed!

packages/web-components/src/components/card-link/__stories__/card-link.stories.ts
```diff
@@ -43,7 +48,9 @@ export default {
     knobs: {
       CardLink: ({ groupId }) => ({
         disabled: boolean('Disabled (disabled):', false, groupId),
-        href: textNullable('Card Href (href):', 'https://example.com', groupId),
+        href: textNullable('Card href (href):', 'https://example.com', groupId),
+        heading: textNullable('Card heading (heading):', 'Explore AI uses cases in all industries', groupId),
```
ariellalgilmore, line 52:
```suggestion
        heading: textNullable('Card heading (heading):', 'Explore AI use cases in all industries', groupId),
```

ariellalgilmore:
Thanks for the updates @mwanberg!! Looks good! Just one small thing the disabled state is no longer "greying out" the heading

Screen Shot 2021-08-12 at 9 12 54 AM

mwanberg:
@ariellalgilmore Good catch on the disabled state! I've fixed that plus the typo you found.

proeung:
@mwanberg Thanks for the update! Everything looks good from a code perspective, however, I think there are some typography changes that might be needed for the `copy` so that it's in alignment with Saijan's design specs (see - https://ibm.ent.box.com/file/827717793884). I believe we're using `$body-long-02` for the copy area. I'll defer the design QA of this PR to @shixiedesign. 

Screen Shot 2021-08-12 at 12 49 05 PM

packages/styles/scss/components/card-link/_card-link.scss
```diff
@@ -32,9 +33,13 @@
       }
     }
   }
+  ::slotted([slot='heading']),
   .#{$prefix}--card__CTA.#{$prefix}--card .#{$prefix}--card__copy {
```
proeung, line 37: Looking at this design specs doc, it looks like we're using `$body-long-02` for the copy, unless this doc is out of date. cc: @shixiedesign 

https://ibm.ent.box.com/file/827717793884

packages/web-components/src/components/card-link/__stories__/README.stories.mdx
```diff
@@ -33,7 +33,8 @@ import '@carbon/ibmdotcom-web-components/es/components/card-link/index.js';
 
 ```html
 <dds-card-link href="https://example.com">
-  Copy text
+  <dds-card-heading>Heading text</dds-card-heading>
```
proeung, line 36: Since we're now referencing `<dds-card-heading>`, let's ensure that the [index.ts](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/packages/web-components/src/components/card-link/index.ts) of this component is importing from `import '../card/card-heading;`, so all of the dependent components are imported in the build process.

proeung (changes requested):

shixiedesign (approved):
Looks good! Thank you!

proeung (approved):
LGTM!

jeffchew (approved):
LGTM!

PR #6840, opened 2021-08-10
chore(deps): remove unused dependency

kennylam:
### Description

Removes `is-port-reachable` since it is no longer being used.

### Changelog

**Removed**

- `is-port-reachable` package from both React and web components


jeffchew (approved):
Thanks @kennylam !

Issue #6844, opened 2021-08-10
[CTASection]: passing Onclick to the cta does not work


annawen1:
## Detailed description

> I'm trying to use the CTA Section component, with CTA as a button. The buttons work properly except for the onClick property, that seems to be ignored. This is how I'm calling the component:

<CTASection {...props.contentBlockProps} cta={{ style: "button", type: "local", buttons: [{ type: "local", copy: "CTA 1", onClick: () => console.log(true) }] }} />

https://www.ibm.com/standards/carbon/react/?path=/story/components-cta-section--default

I used a similar approach in the LeadSpace component and it works fine.
Could you help me figured out this issue, please?

Inspecting the CTASection component code, on the "@carbon/ibmdotcom-react" package, seems that the prop type for "onClick" is not there

> General

> CTA Section component

> The onClick should work as on the cta on leadspace

> Google Chrome

annawen1:
looks like the onClick is being overwritten here: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/packages/react/src/components/CTA/ButtonCTA.js#L79

annawen1:
This issue was originally reported by @KIQsco working on the the think 2021 summit page

PR #6863, opened 2021-08-11
fix(CTA): allow onClick for buttonCTA to be passed

annawen1:
### Related Ticket(s)

[CTASection]: passing Onclick to the cta does not work #6844

### Description

The `button.onClick` was being overwritten, so adopters trying to pass in an onClick function were not able to. Solution is to check if an `onClick` function was passed in before setting it.

### Changelog

**Changed**

- check if `onClick` function has been passed before setting the jump function


jeffchew (approved):
LGTM!

kennylam (approved):

PR #6864, opened 2021-08-11
fix(masthead): remove cloud masthead position fixing code

annawen1:
### Related Ticket(s)

### Description

Reverting changes made in: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/6836

Setting the position to fix solves part of the issue, but when there is a banner above the masthead, the masthead is hidden underneath the banner. Reverting the changes for no until there is a more solid fix.

### Changelog

**Removed**

- cloud-masthead position fixing code


emyarod (approved):
looks good to me!

Issue #6794, opened 2021-08-03
Content Group Banner | Not showing the Divider line between title and CTA for content group banner in safari - DDS Consulting


druckerk:
Brief Description: The Divider line between title and cta for the content group banner is not showing up in safari. 

## Detailed description

> Describe in detail the issue you're having.

The divider line between the title and cta for the content group banner will appear in chrome and firefox browsers. However, when opening the links in safari the divider line does not appear. 

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

This is a general issue when open the component in Safari

> Is this issue related to a specific component?

This issue is related to the Content Group Banner component.

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Actual result: 
- Not showing the Divider line between title and CTA for content group banner in safari
- https://dev-publish.np-roks.cms.cis.ibm.net/qa/abhinav-testpage/learn-landing-page-20-july
- https://www.ibm.com/standards/carbon/web-components/?path=/story/components-content-group-banner--default

Expected Result:
- Show show the Divider line between title and CTA for content group banner in safari as it is showing in Chrome

> What browser are you working in?
- Chrome (works as expected)
- FireFox (works as expected)
- Safari (Does not work)

> What version of Carbon for IBM.com are you using?
- ibmdotcom-styles: 1.21.0
- ibmdotcom-we-compnents: 1.6.0

## Steps to reproduce the issue

1. Copy the links provided above
2. Paste the links into Safari
3. Then paste the links to chrome or firefox
4. Then review Safari and chrome/firefox side by side and you will notice the divider is missing in firefox.

## Additional information

Safari Screenshot: 
![Screen Shot 2021-08-03 at 3 10 46 PM](https://user-images.githubusercontent.com/88403358/128079577-31a9a3d1-e5a1-412d-9216-ac691411a465.png)
![Screen Shot 2021-08-03 at 3 10 39 PM](https://user-images.githubusercontent.com/88403358/128079579-8fd3e6a5-3cd2-450e-955a-2600d641f0e5.png)

Chrome Screenshot: 
![Screen Shot 2021-08-03 at 3 10 17 PM](https://user-images.githubusercontent.com/88403358/128079630-17221bdd-7a85-4356-93ef-8c31aa8269eb.png)
[...]

PR #6865, opened 2021-08-11
fix(content-group-banner-6794): fix before pseudo-element bug on safari

proeung:
### Related Ticket(s)

GH Issue: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6794
HC JIRA Ticket: https://jsw.ibm.com/browse/HC-2130

### Description

- This PR removes the use of `::before` pseudo-element on components/slots that was previously used to create the divided line because Safari can't recognize the pseudo-elements like `::before` and `::after` in the Shadow Dom.
- 
### Testing Instruction

- Visit the `Content group banner` in Safari- `/?path=/story/components-content-group-banner--default`
- Ensure that the divider line between title and CTA is showing up correctly in Safari on larger breakpoints.

### Expected Display

Screen Shot 2021-08-11 at 3 23 00 PM


ariellalgilmore:
Thanks @proeung! Looks like Percy caught a change that the divider should be aligned to the edge of the column in the max breakpoint vs in the middle for xl and lg breakpoints

Screen Shot 2021-08-12 at 8 45 02 AM

shixiedesign:
Hi! Yes thanks for the fix - like Ari mentioned, I think the only thing is the text links didn't start on the column: 
**Current:** 
image

Expecting as spec image Ari posted above ☝️

proeung:
@ariellalgilmore @shixiedesign Great catch! I just pushed up a commit that should fix the spacing issue.

shixiedesign (approved):
👍 ⚡

ariellalgilmore (approved):
LGTM! Thank you!

Issue #6884, opened 2021-08-13
React: LightboxMedia viewer - On FireFox browser, lightbox media modal is not closing on click of close icon.


Praveen-sr:
## Detailed description

Issue; 
On **Firefox** browser, lightbox media modal is not closing on click of close icon.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/50964c07-4231-4f56-922b-a4329ef3b797)

PR #6885, opened 2021-08-13
fix(lightbox-media): check for kaltura onClose

kennylam:
### Related Ticket(s)

#6884 

### Description

Fixes an issue when Kaltura callback is called every time `onClose`, even when video isn't being used.

### Changelog

**Changed**

- send Kaltura notifications only when widget is detected

### DIFF
--- a/packages/react/src/components/LightboxMediaViewer/LightboxMediaViewer.js
+++ b/packages/react/src/components/LightboxMediaViewer/LightboxMediaViewer.js
@@ -142,13 +142,11 @@ const LightboxMediaViewer = ({ media, onClose, ...modalProps }) => {
    * Stop video on modal close
    */
   function closeModal() {
-    if (onClose?.() !== false) {
-      if (root.kWidget) {
-        root.kWidget.addReadyCallback(videoId => {
-          const kdp = document.getElementById(videoId);
-          kdp.sendNotification('doStop');
-        });
-      }
+    if (onClose?.() !== false && root.kWidget) {
+      root.kWidget.addReadyCallback(videoId => {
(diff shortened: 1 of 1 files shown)


packages/react/src/components/LightboxMediaViewer/LightboxMediaViewer.js
```diff
@@ -142,7 +142,7 @@ const LightboxMediaViewer = ({ media, onClose, ...modalProps }) => {
    * Stop video on modal close
    */
   function closeModal() {
-    if (onClose?.() !== false) {
+    if (onClose?.() !== false && root.kWidget) {
       root.kWidget.addReadyCallback(videoId => {
```
emyarod, line 146: would this also work as expected when Kaltura isn't instantiated?

```suggestion
    if (onClose?.() !== false) {
      root.kWidget?.addReadyCallback(videoId => {
```

emyarod (approved):
looks good to me! default lightbox media viewer can be dismissed on Firefox macOS and Win10 again

packages/react/src/components/LightboxMediaViewer/LightboxMediaViewer.js
kennylam: @emyarod It would, yeah. `onClose` works regardless since it's inherited from `ExpressiveModal`. That condition is just to stop the video when the modal closes so it's not needed for the default version.

emyarod: got it, makes sense!

ariellalgilmore (approved):
LGTM! Tested on ff and got it to close

PR #6900, opened 2021-08-17
feat(card-group): change same height util to apply to whole group instead of per row

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5868

### Description

Changed card group to apply `sameHeight()` to all cards equally instead of per row


shixiedesign (approved):
Awesome! Thanks for the fix!

jeffchew:
@IgnacioBecerra can you review this change? You were the last one to touch the sameHeight logic here based on the history so you might have the best context, but offhand it looked ok to me.

IgnacioBecerra (approved):
LGTM, thanks @NSDrowned!

Issue #6904, opened 2021-08-17
[CTA Block] Using wrong top padding above heading-05


kirbyevans:
## Detailed description

The CTA Block component is using the wrong top padding above the heading-05 title. It is currently using 48px on all breakpoints. Instead, it should be using a top padding of 64px on max, xlg,and lg and use 32px on md and sm breakpoints.

## Steps to reproduce the issue

[Storybook](https://www.ibm.com/standards/carbon/web-components/?path=/story/components-cta-block--default)
[Design specs](https://ibm.box.com/s/ck4ludd5kj6rgt4uf19btrwor18nsrq4)

## Additional information

**What is currently on Storybook**
![image](https://user-images.githubusercontent.com/77290329/129738572-88c12c45-d017-45e9-aaad-723286371ec2.png)

**Expected on max, xlg, and lg**
![image](https://user-images.githubusercontent.com/77290329/129738622-4c7d3931-1b3e-497f-b7cb-7bed36eb7665.png)

**Expected on md and sm**
![image](https://user-images.githubusercontent.com/77290329/129738747-c3e6a96c-4440-439a-8c24-3185746e8502.png)

shixiedesign:
All good except CTA section is not mis-aligned. The top margin of CTA section's section heading should also be 64 in here. 

![image](https://user-images.githubusercontent.com/15144993/129782627-9d6fcefb-fdaf-4967-aab4-26cb1886559f.png)

When a Content section is used with Content block, the top spacing should follow the larger one like in this case: 

![image](https://user-images.githubusercontent.com/15144993/129782943-946ad2e2-f813-4574-9e5a-79c1d931017e.png)

Though there are quite a lot of spacing bugs in this component and I can open a more thorough issue later if that's better. Lemme know.

RobertaJHahn:
@emyarod I think this can be closed.  Is that correct?

emyarod:
@RobertaJHahn yes this should be resolved by #6907

RobertaJHahn:
closing,  based on comments above.

PR #6907, opened 2021-08-17
fix(cta-block): adjust header padding values to spec

emyarod:
### Related Ticket(s)

#6904

### Description

This PR reduces the padding above the CTA Block header on sm and md sizes and increases the padding on lg+

sm, md breakpoints:

image

lg, xl, max:

image

### Changelog

**Changed**

- move padding rule for CTA block heading to `styles` package and adjust values according to spec


shixiedesign:
All good except CTA section is not mis-aligned. The top margin of CTA section's section heading should also be 64 in here. 

image

When a Content section is used with Content block, the top spacing should follow the larger one like in this case: 

image

Though there are quite a lot of spacing bugs in this component and I can open a more thorough issue later if that's better. Lemme know.

emyarod:
@shixiedesign CTA blocks in CTA sections should now have the 64px spacing again!

shixiedesign (approved):
Visually it's looking expected. Thanks!

jeffchew (approved):

Issue #6912, opened 2021-08-18
Missing Storybook prop table for Filter panel component


emyarod:
## Detailed description

[Originally reported in Slack](https://cognitive-app.slack.com/archives/CL83LMKSA/p1629243736014500)

When viewing the Filter panel component story docs, the prop table is missing: https://www.ibm.com/standards/carbon/web-components/?path=/docs/components-filter-panel--default#props

![image](https://user-images.githubusercontent.com/8265238/129896590-a7a8a62b-57f7-49a2-a27d-c9a3efd46a91.png)

PR #6913, opened 2021-08-18
fix(filter-panel): use correct element name

emyarod:
### Related Ticket(s)

#6912 

### Description

This PR sets the correct element names for the filter panel and filter modal footer which resolves a docgen issue for Storybook prop tables. The story should now display the prop table as expected

image

### Changelog

**Changed**

- update filter panel and filter modal footer JSDoc TS declarations


jeffchew (approved):
good catch!

Issue #6205, opened 2021-05-26
[Leaving IBM] without expressive theme, updates needed


oliviaflory:
## Detailed description

> Describe in detail the issue you're having.

Without the expressive theme, some items within the Leaving IBM component got smaller

- [ ] Update label to use $label-02 token
- [ ] Update link to use `Link large` variant from Carbon core

<img width="630" alt="Screen Shot 2021-05-26 at 2 43 42 PM" src="https://user-images.githubusercontent.com/45692622/119714619-15011700-be31-11eb-89d1-401b657aebfc.png">

## NOTE TO DEVELOPER

**Branch should be off of the `feat/expressive-sunset` branch and targeted back to this branch when opening a PR!**

## Additional information

[PR removing expressive](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/6127)
[Percy](https://percy.io/538fc19a/Carbon-for-IBM.com-Web-Components/builds/10505563/search/preview/590974692?browser=chrome&searchParam=leaving&viewLayout=overlay&viewMode=new&width=1280)
[Web components](https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/6127/index.html?path=/story/components-leaving-ibm--default)
[React wrapper](https://ibmdotcom-web-components-react.s3-web.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/6127/index.html?path=/story/components-leaving-ibm--default)

PR #6928, opened 2021-08-19
fix(leaving-ibm): token updates

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6205

### Description

- [X] Update label to use $label-02 token
- [x] Update link to use `Link large` variant from Carbon core


packages/web-components/src/components/leaving-ibm/leaving-ibm.scss
```diff
@@ -30,19 +30,23 @@
 :host(#{$dds-prefix}-leaving-ibm-modal-body) {
   @extend :host(#{$prefix}-modal-body);
 
+  #{$prefix}-link {
+    @include carbon--type-style('body-long-02', true);
+  }
```
photodow, line 35: I don't think you need to override it. The component has a setting `size="lg"`. Is there a reason we can't use that instead of overwriting the `CSS`?

photodow: Also, it looks like `body-short-01` from the original component is getting the higher specificity.

packages/web-components/src/components/leaving-ibm/leaving-ibm.scss
```diff
@@ -30,19 +30,23 @@
...
   }
 }
 
 :host(#{$dds-prefix}-leaving-ibm-modal-supplemental) {
   display: block;
   margin-top: $spacing-07;
   color: $text-02;
-  @include carbon--type-style('label-01');
+  @include carbon--type-style('label-02');
 }
```
photodow, line 52: Looks like this was here previously, but any idea why these values are duplicated under slightly different selectors? Isn't there only one label?

photodow (commented):
Did the React not build?

packages/web-components/src/components/leaving-ibm/leaving-ibm.scss
NSDrowned: @photodow I have no idea why it was there but I tested it and the first one doesn't seem to be doing anything. I removed it.

packages/web-components/src/components/leaving-ibm/leaving-ibm.scss
NSDrowned: @photodow for some reason when I tested `size="lg"`, it didn't show any difference. I thought it was broken so I added the override. I just retested and it works. I updated the code. I guess my local build had a hiccup.

NSDrowned:
@photodow checking what happened to the react wrapper

NSDrowned:
@photodow so the duplicated style block was for the react wrapper. I restructured it so we don't have to use it anymore.

photodow (approved):

jeffchew (approved):
LGTM!

Issue #6936, opened 2021-08-20
[Tabs extended media] Extra padding between tabs and media


RichKummer:
## Detailed description

When checking the Tabs extended media React wrapper, noted that there was additional padding between the Tabs component and the media component on desktop. According to design specs, there should be 64px padding between both Checked in Chrome and Safari and had the same issue.

At a glance, this could be bottom padding/margin coming from the Tabs component.

## Steps to reproduce the issue

[Web Components v.1.8.1](http://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-tabs-extended-with-media--default)

[Code sandbox](https://webcomponents-codesandbox.s3-web.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/6919/tabs-extended-media/index.html)

Inspect the padding between tabs and media section.

## Additional information

**Visual specs**
<img width="528" alt="Screen Shot 2021-08-20 at 9 13 56 AM" src="https://user-images.githubusercontent.com/15643582/130238663-d0552763-3050-411b-bdfe-cf4a7994947c.png">

**React wrapper**
<img width="1130" alt="Screen Shot 2021-08-19 at 5 00 20 PM" src="https://user-images.githubusercontent.com/15643582/130235434-5f36879c-9eb4-489b-ba7c-6cd06d77d29e.png">

PR #6938, opened 2021-08-20
fix(tabs-extended): adjust scope of vertical tab content spacing rules

emyarod:
### Related Ticket(s)

#6936

### Description

This PR fixes a regression in `<tabs-extended>` styles by keeping the scope of the new `tab-content` spacing rules to the tabs-extended-vertical variant only. `<tabs-extended-media>` should no longer have an additional space between the tabs and the media itself:

image

Reviewers please confirm that these style changes also revert the appearance of all nonvertical extended tabs to be in line with the RC and stable packages while retaining the same spacing for vertical extended tabs proposed in #6614 and #6826!

### Changelog

**Removed**

- redundant vertical extended tabs specific spacing rules


jeffchew (approved):
LGTM!

RichKummer:
Looks good, thanks for fixing this!

PR #6947, opened 2021-08-23
fix(cloud-masthead): prevent masthead shifting on top of page

annawen1:
### Related Ticket(s)

[Cloud Masthead] Slight page shift when the Mega Menu is opened on machines where scrollbar is set to visible #6831

### Description

Problem:
The Cloud Masthead shifts when user is on top of the page. The cloud team has applied extra styling to set the position of the masthead to `static` when on top of page. This is so it doesn't overlap their L1. When user scrolls, the `ibm-masthead-sticky ibm-masthead-sticky-showing` v18 classes are applied to the `body` element. This also sets the position of the masthead back to `fixed`. When expanding the megamenu, a `margin-right` needs to be applied to balance out the disappearing scrollbar when position is `fixed`. This `margin-right` is unnecessary when position is `static`.

Solution:
When megamenu is expanded, detect whether the `body` element has the `ibm-masthead-sticky ibm-masthead-sticky-showing` classes applied before setting the `margin-right`.

### Changelog

**Changed**

- only set the `margin-right` for cloud-masthead when the specific v18 classes are applied.


packages/web-components/src/components/masthead/megamenu-top-nav-menu.ts
```diff
@@ -105,19 +105,33 @@ class DDSMegaMenuTopNavMenu extends DDSTopNavMenu {
...
       if (this.expanded) {
         doc.body.style.marginRight = `${this._scrollBarWidth}px`;
         doc.body.style.overflow = `hidden`;
         forEach(doc.querySelectorAll((this.constructor as typeof DDSMegaMenuTopNavMenu).selectorOverlay), item => {
           (item as DDSMegaMenuOverlay).active = this.expanded;
         });
-        if (masthead) {
+        if (cloudMasthead) {
+          if (doc.body.classList.contains('ibm-masthead-sticky') && doc.body.classList.contains('ibm-masthead-sticky-showing')) {
+            cloudMasthead.style.marginRight = `${this._scrollBarWidth}px`;
```
jeffchew, line 122: @annawen1 this will be interesting to test across all browsers and their varying treatment of the scrollbar!

annawen1: haha we actually already have `margin-right` logic in our global masthead that works fine, just the cloud masthead needs adjustment because the `position` style was changed. Hoping it will work fine for them as well.

jeffchew (approved):
LGTM!

Issue #6953, opened 2021-08-24
[Video player] With Lightbox Media - Audio plays on the background


Praveen-sr:
## Detailed description

Issue: 
Video player With Lightbox Media - When the user does not open the video/ lightbox modal, Audio plays on the background. 

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/60b62f46-4520-4f70-8f0f-e25e3913eaa5)

emyarod:
this issue is no longer present when viewing the story by itself, but in storybook if another video player story is viewed in advance, the autoplay issue appears

PR #6958, opened 2021-08-24
fix(VideoPlayer): restrict lightbox variant autoplay to modal open state

emyarod:
### Related Ticket(s)

#6953

### Description

This PR resolves an issue with the video player with lightbox media viewer, where the video would autoplay on page render (ignoring the open/closed state of the lightbox media viewer modal). The video should now only autoplay after the lightbox modal is launched

### Changelog

**Changed**

- add click handler to `<LightboxMediaViewer>` to autoplay videos only when the lightbox is launched
- fix missing styles in code sandbox example
- update storybook video IDs so that video examples play properly


jeffchew (approved):
LGTM!

Issue #6956, opened 2021-08-24
Masthead Mega menu subjects dismiss the menu when selected on Safari browsers


raximkoron:
## Title line template: [Title]: Masthead Mega menu subjects dismiss the menu when selected on Safari browsers

-->

## Detailed description

> Describe in detail the issue you're having.
When selecting a subject in the masthead mega menu in Safari, the menu dismisses from view rather than displaying the information. If the user re-opens the masthead mega menu at this point, the subject's information is displayed.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?
Bug

> Is this issue related to a specific component?
Masthead Megamenu

> What did you expect to happen? What happened instead? What would you like to
> see changed?
The Masthead Mega Menu will remain open and display the subject's contents when selected.

> What browser are you working in?
Safari 14.1.2

> What version of Carbon for IBM.com are you using?

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?
HC Cloud

## Steps to reproduce the issue

1. Navigate to https://stage.cms.ibm.com/cloud in Safari browser
2. Select Products or Solutions from the mega menu
3. Select a subject at the left-side of the menu (automation, Containers, etc.)
4. Observe that the mega menu dismisses rather than showing the contents of the selected subject

## Additional information

Quicktime is hindering the user experience when screen recording, so the subject doesn't update in the video, but works when screen recording is not taking place.

https://user-images.githubusercontent.com/2856797/130664470-4a3df3d7-c6dd-4e8b-b1fa-4443763d2096.mov

proeung:
The fixes for this issue should be opened against this branch `release/v1.23.2`.

kennylam:
Is the staging URL correct? It's timing out for me.

raximkoron:
Oops, typo, URL should be:
https://stage.cms.ibm.net/cloud

raximkoron:
![IBM_Cloud_-_Российская_Федерация___IBM](https://user-images.githubusercontent.com/2856797/130674013-fa496061-28e5-4df5-9d66-401384ece3d1.jpg)
This fix will be included with the PR as well

PR #6962, opened 2021-08-24
fix(cloud-masthead-6956): Safari text showing issue on cloud-top-nav-name

proeung:
### Related Ticket(s)

- https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6956

### Description
- This PR fixes one of the issues mentioned within this issue - https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6956

### Changelog

**Changed**

- make sure the menu items do not appear behind the Cloud platform name

### DIFF
--- a/packages/web-components/src/components/masthead/cloud/cloud-masthead-composite.ts
+++ b/packages/web-components/src/components/masthead/cloud/cloud-masthead-composite.ts
@@ -117,11 +117,7 @@ class DDSCloudMastheadComposite extends DDSMastheadComposite {
           <dds-cloud-megamenu-tabs value="${sortedMenuItems[0]?.itemKey}">
             ${sortedMenuItems.map(item => {
               return html`
-                <dds-cloud-megamenu-tab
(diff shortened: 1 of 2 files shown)


jeffchew (approved):
LGTM!

emyarod (commented):
setting the tabindex to 0 introduces an unintended side effect where Safari megamenu tabs will have 2 tab stops

https://user-images.githubusercontent.com/8265238/130700115-8185302f-e595-4beb-9693-3be2461cac73.mov

setting it to -1 eliminates the added tab stop and also solves the existing Safari-only issue (but it would break keyboard nav for non-Safari browsers)

I believe I have a fix that would be contained in `carbon-web-components` but further testing would be required

annawen1:
I have removed the `tabindex` fix and will wait for the `carbon-web-components` changes instead

annawen1 (approved):
LGTM

Issue #6939, opened 2021-08-20
[tabs-extended] disabled tab labels style regression in Canary vs RC


emyarod:
[original Slack discussion](https://cognitive-app.slack.com/archives/GBG66T9NK/p1629476763067900)

disabled tabs appear to have a regression in Canary vs RC since the text color differs

Canary

![image](https://user-images.githubusercontent.com/8265238/130271531-ccf33607-5c27-4b91-8519-ace52fff9430.png)

RC

![image](https://user-images.githubusercontent.com/8265238/130271536-0828ccfd-358a-431f-a466-efc84cfa4487.png)

disabled vertical tabs also don't seem to be correct as reported by @ariellalgilmore on Slack

![image](https://user-images.githubusercontent.com/8265238/130271795-e38157cb-0edf-453f-8f3a-f6de3cc2fb65.png)

#### Updated Aug 24: Expecting disabled tab to have text color `$disabled-02` 

![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/2e869199-5dda-4c19-84b1-b9f26f8fd041)

RobertaJHahn:
Moving to To be Triaged pipeline.  Design decision needed.

shixiedesign:
This is an issue. Disabled text is expected to be: `$disabled-02`
![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/f91293f7-1da9-4af6-9b28-af04ca671385)

PR #6963, opened 2021-08-24
fix(tabs-extended): remove disabled label color override

emyarod:
### Related Ticket(s)

#6939

### Description

This PR reverts the extended tabs disabled label color changes to remove a style regression. The disabled tab should now receive the correct color rule

image


kennylam (approved):
LGTM thanks @emyarod!

PR #6964, opened 2021-08-24
fix(footer): Set dds-footer role to contentinfo

### LINKED ISSUE #6890: [Masthead/Footer] Accessibility - The 'role' defined on the element is not valid per WAI-ARIA specification - DDS Consulting
IBM Equal Access Accessibility Checker
The 'role' defined on the element is not valid per WAI-ARIA specification

Elements showing the error 
```
<dds-footer size="" data-autoid="dds--footer" aria-label="footer" role="footer">
```
----
- Hybrid Cloud developer: 
- [HC Jira ticket]
- Dev estimate:  
- DDS consultant:

andy-blum:
### Related Ticket(s)
Fixes #6890 

### Description
This PR sets the `dds-footer` element's role to `contentinfo` overriding any pre-existing roles that may have been set on the element.

### Changelog

**Changed**
- `dds-footer` has `contentinfo` landmark role

### DIFF
--- a/packages/web-components/src/components/footer/footer.ts
+++ b/packages/web-components/src/components/footer/footer.ts
@@ -36,9 +36,7 @@ class DDSFooter extends StableSelectorMixin(LitElement) {
   size = FOOTER_SIZE.REGULAR;
 
   connectedCallback() {
-    if (!this.hasAttribute('role')) {
-      this.setAttribute('role', 'footer');
-    }
+    this.setAttribute('role', 'contentinfo');
 
     if (!this.hasAttribute('aria-label')) {
       this.setAttribute('aria-label', 'footer');
--- a/packages/web-components/tests/snapshots/dds-footer-composite.md
+++ b/packages/web-components/tests/snapshots/dds-footer-composite.md
@@ -9,7 +9,7 @@
   <dds-footer
     aria-label="footer"
     data-autoid="dds--footer"
-    role="footer"
+    role="contentinfo"
     size=""
   >
     <dds-footer-logo data-autoid="dds--footer-logo">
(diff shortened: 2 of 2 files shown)


andy-blum:
One issue I noticed that remains (at least in the dotcom shell story): The `dds-footer` element now having the `contentinfo` role ought to be a top-level landmark on the page. Currently it is a child of an element with `role="main"`
Screen Shot 2021-08-24 at 4 29 33 PM

Additionally, we're using an HTML5 `<main>` element which carries an implicit `role="main"`. A page should not have more than one element with a role of `main`.
Screen Shot 2021-08-24 at 4 29 53 PM

jeffchew:
Thank you @andy-blum !

Can you run `yarn test:unit:updateSnapshot` in the web components package folder to update the snapshot files?

andy-blum:
Yes! I knew what test failed but not how to fix it!

jeffchew (approved):
Thank you!

proeung:
@jeffchew @Praveen-sr Any thought about what @andy-blum mentioned here - https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/6964#issuecomment-904959569?

jeffchew:
@proeung @andy-blum looks like the configuration is on the container level in Storybook, so might be something we will need to globally set:

image

Ideally we would not be setting the `role="main"` if there is already a `dds-dotcom-shell` on the page, if that isn't happening we will need to look into that.

@Praveen-sr any thoughts?

jeffchew:
Looking more into this, the attribute `hasMainTag: true` will probably need to be added to the default dotcom-shell story parameters (`dotcom-shell.stories.ts`)

PR #7001, opened 2021-08-30
chore(snapshot): update snapshot and workflow token

jeffchew:
### Related Ticket(s)

No related issue

### Description

The react snapshot update workflow failed because it didn't have the right permissions to push. This should hopefully fix it, as well as updating the existing snapshots so that it clears.

### Changelog

**Changed**

- Updated snapshot-update.yml
- Updated react snapshot file


emyarod (approved):
looks good to me!

Issue #7003, opened 2021-08-30
Leadspace tags and headline display inline


rastogir:
Steps to reproduce the issue:

- Go to tall, super or medium variant of leadspace
- Reduce headline character count including spaces to 17 characters or less.
- Enable tag group navigational elements via knobs
- Tag group and headline are displayed side by side for lg, xlg and max breakpoints
<img width="1655" alt="Screen Shot 2021-08-30 at 10 09 20 AM" src="https://user-images.githubusercontent.com/11525031/131361854-00346f75-e510-4cd9-8ff2-47694f1bf752.png">
<img width="1301" alt="Screen Shot 2021-08-30 at 10 08 26 AM" src="https://user-images.githubusercontent.com/11525031/131361859-8e44bc10-5575-4c87-ad15-c8d87e323200.png">
<img width="1576" alt="Screen Shot 2021-08-30 at 10 07 40 AM" src="https://user-images.githubusercontent.com/11525031/131361862-c0b0e78c-1a84-4c77-a3cd-90e4a79c0f39.png">

PR #7004, opened 2021-08-30
 fix(leadspace): update styles so nav and heading are not inline

ariellalgilmore:
### Related Ticket(s)

#7003

### Description

When Leadspace character count including spaces to 17 characters and navigational elements are present then the heading and navigation will appear in the same row. 

Fix:
Screen Shot 2021-08-30 at 9 32 10 AM

### Changelog

**Changed**

- style to adjust to nowrap


IgnacioBecerra (approved):
LGTM!

PR #7005, opened 2021-08-30
chore(codesandbox): update leadspace image

kennylam:
### Description

Remove local assets from the `leadspace` web components Codesandbox example as Parcel currently does not support specifying directories into `dist`. This is causing the image to 404 when the page loads.

### Changelog

**Changed**

- change leadspace image to use online service (`picsum`)

**Removed**

- local assets (leadspace image)


ariellalgilmore (approved):
Thank you!!

Issue #6737, opened 2021-07-27
[Card] Card component static not changing color


leticiafontoura:
By removing the `light: true` property for the card static, it remains with the light background, not changing to the grayish one.

ariellalgilmore:
Hi @leticiafontoura, is there a link where we can see this happening? Here is a [codesandbox example](https://codesandbox.io/s/fervent-leaf-q2k2o?file=/index.html) of the static card without the `color-scheme="light"`.

leticiafontoura:
Sorry for my lack of details.

This is for the react component. 
![image (15)](https://user-images.githubusercontent.com/47862337/127228646-efe90feb-5b82-4489-bb28-c5a82070a6e5.png)

```
const card = {
    heading: "Bla Bla",
    copy: "Copsssy",
    border: true,
    cardStatic: true,
    eyebrow: "dr. blabla",
    cta: {
      href: "",
      copy: ""
    },
  }
...
<Card {...card} />
```

RobertaJHahn:
@leticiafontoura  What application are you using this component in?  Do you have a time frame for this bug to be fixed?

leticiafontoura:
@RobertaJHahn no, I don't have a time frame really. the application is just on my computer atm but i'm using the nextJS template

RobertaJHahn:
@leticiafontoura Good to know.  We are currently scheduling work several sprints out, so if there is no urgency, the work will be scheduled as the engineering resources become available.

RobertaJHahn:
@leticiafontoura The dev work is finished and will be included in sprint v1.25.0.  The release date is Sept 28.

cc:  @NSDrowned

PR #7006, opened 2021-08-30
fix(card): removes light background override from static card class

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6737

### Description

In React, `bx--card--static` class was overriding the light background set by `bx--card--light`. Now `cardStatic` and `light` props work separately.


ariellalgilmore (approved):
Thank you @NSDrowned!!

photodow (commented):
Looks like I'm getting some sort of specificity issue in the card group on NextJS.

image

NSDrowned:
> Looks like I'm getting some sort of specificity issue in the card group on NextJS.
> 
> image

I checked the template and the card groups have the G10 theme hardcoded:

image

Is that how it's supposed to be? ❓

NSDrowned:
@photodow fixed the issue (be sure to clear cache):

image

photodow:
Maybe I'm doing something wrong, but here are a couple of things I'm seeing.

- I'm seeing hover states on static cards.

- I'm seeing static card navigating to another page when clicked on.

- NextJS has some theming issues. Looks correct in Storybook at least?

photodow (approved):

Issue #5753, opened 2021-04-09
[Video player] React: image overlay override and caption override enhancements


RobertaJHahn:
#### User Story

> As a `[user role below]`:
Carbon for ibm.com developer

> I need to:
create/change the `Video player`

> so that I can:
provide the ibm.com adopter developers components they can use to build ibm.com web pages

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- **See the Epic (#5642 ) for the Design and Functional specs information**
- Visual QA review is not needed for these changes.
- Prod QA testing issue (#6194 )

#### Acceptance criteria
- [ ] Built as a pure React component/variant
- [ ] Include README for the react component and corresponding styles
- [ ] Add any necessary stable selectors
- [ ] Create codesandbox example under `/packages/react/examples/codesandbox` and include in README
- [ ] Minimum 80% unit test coverage
- [ ] Update the Carbon for ibm.com website component [data file](https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/blob/master/src/data/components.json) to be sure the web site Component Status and Storybook links are correct
- [ ] If a design is provided, the Designer is included as a Reviewer in the Pull Request
- [ ] Provide a direct link to the deploy preview for the designer in the Pull Request description
- [ ] A comment is posted in the Design QA issue, tagging Designer, when development is finished
- [ ] The Storybook link is added to the Design QA issue for their testing
- [ ] A comment is posted in the Prod QA issue, tagging Praveen when development is finished

RobertaJHahn:
Closing as the video image enhancement will be managed as a sub-task under the additional focus status epic  (#5514 )

RobertaJHahn:
PR #6841 was linked in error.  It should have been linked to issuse #5505.

PR #7007, opened 2021-08-30
feat(VideoPlayer): support image and caption overrides

emyarod:
### Related Ticket(s)

#5753

### Description

This PR adds support for custom thumbnails and video captions in the React `VideoPlayer` component. If no custom overrides are provided the values will be taken from the Kaltura API 

image

### Changelog

**New**

- `thumbnail` prop on `VideoPlayer` and `VideoImageOverlay`
- `caption` prop on `VideoPlayer`


packages/react/src/components/VideoPlayer/VideoPlayer.js
```diff
@@ -116,7 +121,7 @@ const VideoPlayer = ({
       </div>
       {showCaption && (
         <div className={`${prefix}--video-player__video-caption`}>
-          {videoData.name} {videoDuration}
+          {videoCaption}
```
ariellalgilmore, line 124: Do we still want the `videoDuration` appened after the custom caption

packages/react/src/components/VideoPlayer/__stories__/VideoPlayer.stories.js
```diff
@@ -48,27 +50,29 @@ Default.story = {
         aspectRatio: 'default',
         videoId: '1_9h94wo6b',
         playingMode: 'inline',
+        caption: text('Caption (caption):', '', groupId),
```
ariellalgilmore, line 53:
```suggestion
        caption: text('Custom caption (caption):', '', groupId),
```

packages/react/src/components/VideoPlayer/VideoPlayer.js
emyarod: I left it since that was in the existing caption but I will defer to design on this one

jeffchew: @emyarod @ariellalgilmore functional specs says it gets appended, so it looks like that's correct: https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/wiki/Video#1-caption

emyarod: thanks for checking! I will leave it as is

IgnacioBecerra (approved):
LGTM, thanks @emyarod!

shixiedesign (approved):
Works great!

PR #7010, opened 2021-08-30
fix(LightboxMediaPlayer): set `autoPlay` based on modal open state

emyarod:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6953

related PR https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/6958

### Description

This PR updates the `autoPlay` prop on lightbox media viewers with embedded video players so that it is dependent on whether or not the lightbox modal is open. The original issue stemmed from how the video player with lightbox media viewer recursively creates `VideoPlayer` components

Reviewers should confirm that the fix from #6958 is preserved and that switching between `VideoPlayer` stories does not trigger autoplay in the background

### Changelog

**Changed**

- `autoPlay` logic in `LightboxMediaViewer`

### DIFF
--- a/packages/react/src/components/LightboxMediaViewer/LightboxMediaViewer.js
+++ b/packages/react/src/components/LightboxMediaViewer/LightboxMediaViewer.js
@@ -105,7 +105,7 @@ const LightboxMediaViewer = ({ media, onClose, ...modalProps }) => {
               <div
                 className={`${prefix}--lightbox-media-viewer__media ${prefix}--no-gutter`}>
                 {media.type === 'video' ? (
-                  <VideoPlayer videoId={media.src} autoPlay />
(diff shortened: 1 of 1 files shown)


packages/react/src/components/LightboxMediaViewer/LightboxMediaViewer.js
```diff
@@ -30,6 +30,8 @@ const LightboxMediaViewer = ({ media, onClose, ...modalProps }) => {
     description: '',
   });
 
+  console.log(modalProps);
```
jeffchew, line 33: @emyarod looks like this console log was missed.

ariellalgilmore (approved):
LGTM :D

Issue #6141, opened 2021-05-20
[CTA] Without Expressive theme, switch to use Carbon core


shixiedesign:
### CTA component without expressive theme should be using core's expressive components: 
From Percy diff

**CTA - Text**
Should use Carbon's `Link (lg)` which should be 16px. 
![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/c95d1d4a-2cae-4e88-ac27-b2443a246f14)

**CTA - Button**
Should use Carbon's `Button (lg expressive)` ([storybook](https://react.carbondesignsystem.com/?path=/story/components-button--expressive-buttons)) which should be 16px label and 48px heigh container. 
![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/f443a55a-532f-4439-bf1f-c2727c2e7aaa)

*Note: no design spec update needed
**CTA - Card** and **CTA - Feature** are fine and unaffected! :tada: 

*Note: these updates might be made through #6179 and #6180 if changes cascade correctly*

## NOTE TO DEVELOPER

**Branch should be off of the `feat/expressive-sunset` branch and targeted back to this branch when opening a PR!**

PR #7022, opened 2021-08-31
fix(cta): updated button to use expressive

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6141

### Description

CTA text already using lg size, updated Button to use expressive

image


oliviaflory:
@NSDrowned Web components looks great!

Seems like Percy caught some upstream changes in the Next JS build, any idea where that could be coming from? Card link text looks smaller and the label sizes look a little more spaced out than expected.

I also didn't see React in this PR, is that separate or would the changes for expressive button be seen in both?

Thanks!

NSDrowned:
@oliviaflory forgot to mention that about react, yes, both are fixed: (React previews are not working I think):

image

image

Let me check what's up with NextJS

NSDrowned:
@oliviaflory I checked the differences reported by Perci and they don't seem related to the button which was the only element updated in this PR.

oliviaflory (approved):
LGTM!

Issue #6975, opened 2021-08-25
[Cloud-masthead]: Adjust tabs to not use title as the identifier for related tab content, use unique ID instead


annawen1:
## Detailed description

> Describe in detail the issue you're having.

The first tab from the `Solutions` menu is not showing any content when using the arabic masthead data. The content is rendered with the `hidden` attribute still applied to it. 

Appears to only be affecting the arabic locales.

**This is occurring due to the titles of the first tabs in both "Products" and "Solutions" being the same. The solution is to not use the title as an ID for the tabs.**

<img width="1525" alt="Screen Shot 2021-08-25 at 9 40 59 AM" src="https://user-images.githubusercontent.com/54281166/130801808-b594deb8-0c97-4d1d-9520-bafdb88931b6.png">

## Steps to reproduce the issue

1. Issue can be viewed at the translations deploy preview: https://ibm-translations-cloud.s3-web.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/417/index.html?cc=ae&lc=ar
2. Open the second menu item from the left in masthead
3. Notice the content for the first tab is not being displayed

proeung:
@annawen1 Thanks for opening this issue! I think adding a `unique ID` instead of using the `title` to connect the tab item with the content would eliminate the special case that we've seen in the Arabic translation.

I'm marking this issue as "Severity 3" and slated for the `v.1.25` release.

PR #7024, opened 2021-08-31
fix(cloud-masthead): set unique id for megamenu tabs instead of using title

annawen1:
### Related Ticket(s)

[Cloud-masthead]: Adjust tabs to not use title as the identifier for related tab content, use unique ID instead #6975

### Description

Ran into an issue with the cloud megamenu tabs when the same tab title is used in two different megamenus. This causes an issue because the current logic uses the title as the identifier of the tabs. Solution is to use a unique id for the tabs instead so no issues occur if the same title is used in multiple megamenus.

**BEFORE:**
Screen Shot 2021-08-31 at 1 09 35 PM

**AFTER:**
Screen Shot 2021-08-31 at 1 06 59 PM

### Changelog

**Changed**

- set a unique id using the parent menu item key + the tab key

### DIFF
--- a/packages/web-components/src/components/masthead/cloud/cloud-masthead-composite.ts
+++ b/packages/web-components/src/components/masthead/cloud/cloud-masthead-composite.ts
@@ -89,7 +89,7 @@ class DDSCloudMastheadComposite extends DDSMastheadComposite {
    *  Render MegaMenu content
    *
    * @param sections menu section data object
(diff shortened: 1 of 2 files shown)


jeffchew (approved):
LGTM!

IgnacioBecerra (approved):
LGTM! Thanks @annawen1

Issue #6929, opened 2021-08-19
[Lead space - short] Web Component: Develop new variation


RobertaJHahn:
#### User Story

> As a `[user role below]`:
Carbon for ibm.com developer

> I need to:
create/change the `Lead space - short`

> so that I can:
provide the ibm.com adopter developers components they can use to build ibm.com web pages

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- Create with Shadow DOM and Custom Elements standards
- **See the Epic (#6873 ) for the Design, Functional specs and comments**
- **Design spec:** https://ibm.box.com/s/mnqx7jzi3vo7ssqaf8oyd3kb8x9mv3p8
- **Functional spec:** https://ibm.box.com/s/b0xcd5fncm2424b2dyxlz28u60bmwqsj
- Web Components Visual QA testing issue (#6930)
- Web Components Prod QA testing issue (#6931)

#### Acceptance criteria
- [ ] Include README for the web component and corresponding styles
- [ ] Create Web Components styles in styles package
- [ ] Do not create knobs in Storybook that include JSON objects
- [ ] Break out Storybook stories into multiple variation stories, if applicable
- [ ] Create codesandbox example under `/packages/web-components/examples/codesandbox` and include in README
- [ ] Minimum 80% unit test coverage
- [ ] Update the Carbon for ibm.com website component [data file](https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/blob/master/src/data/components.json) to be sure the web site Component Status and Storybook links are correct
- [ ] Use the [Visual QA checklist](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/wiki/Definition-of-done-(Visual-QA-checklist)) to verify design quality
- [ ] If a design is provided, the Designer is included as a Reviewer in the Pull Request
- [ ] Provide a direct link to the deploy preview for the designer in the Pull Request description
- [ ] A comment is posted in the Prod QA issue, tagging Praveen when development is finished

jeffchew:
Fix Central is requesting this is available by the `v1.25.0` release.

PR #7026, opened 2021-08-31
feat(leadspace): add short variation

ariellalgilmore:
### Related Ticket(s)

#6929

### Description

Create leadspace short variation. 

### Changelog

**New**

- short size
- storybook with short variation and for react-wrapper story

**Changed**

- cleaned up unnessary classes and styles
- cleaned warning in leadspace search


IgnacioBecerra:
I saw a small issue on the `Short with image` variation where in the lg breakpoint and above, the image seems to get slightly cut off. Wondering if the use of `height: fit-content` in `dds-leadspace-image` or some kind of similar property might help, as seen in the After image below, but would like feedback from designers to see their thoughts on this. (cc: @RichKummer)

Before (current):
Screen Shot 2021-08-31 at 1 43 15 PM


After (with `fit-content`):
Screen Shot 2021-08-31 at 1 45 05 PM

kennylam:
> Wondering if the use of `height: fit-content`

`height: auto` would also work.

ariellalgilmore:
Thanks @IgnacioBecerra and @kennylam updated the height!

RichKummer (approved):
The added `height-auto` solved the issue on `max`. The behavior on `md` where the image drops is intended, I'll update the design specs in sketch to match.

Issue #5997, opened 2021-05-07
[Table of contents] horizontal - On theme g90 & g100, overflow icon is not visible for the user


Praveen-sr:
## Detailed description

Issue:
Table of contents - horizontal - On theme g90 & g100, overflow icon is not visible for the user. 
On hover on overflow, its seen for the user.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/35c87e60-81fb-4f17-b528-639d4359d22d)

PR #7027, opened 2021-08-31
fix(table-of-contents): added text color token to navigation icon

BrunnoM7:
### Related Ticket(s)

[Table of contents] horizontal - On theme g90 & g100, overflow icon is not visible for the user #5997

### Description

Updated the fill color of caret icon on Styles package assuming this will also be used in React as well.

### Changelog

**New**

- {{new thing}}

**Changed**

- {{changed thing}}

**Removed**

- {{removed thing}}


packages/styles/scss/components/tableofcontents/_tableofcontents.scss
```diff
@@ -308,7 +308,7 @@ $hover-transition-timing: 95ms;
     > {
       svg {
         position: absolute;
-        fill: currentColor;
+        fill: $text-01;
```
ariellalgilmore, line 311: Thanks @BrunnoM7 !!! just looked at the specs and it seems like we should be using `$icon-02` here. 
Screen Shot 2021-08-31 at 2 17 20 PM

BrunnoM7: Thanks @ariellalgilmore, I just made a new commit with the right color token.

ariellalgilmore (approved):
Thank you @BrunnoM7!

PR #7030, opened 2021-09-01
fix(masthead): remove aria-selected from menu items

### LINKED ISSUE #6888: [Masthead/Footer] Accessibility - The WAI-ARIA attribute 'aria-selected' is not valid for the element <a> with WAI-ARIA role 'menuitem' - DDS Consulting
----
- Hybrid Cloud developer: 
- HC Jira ticket:
- DDS consultant: 

----
IBM Equal Access Accessibility Checker
The WAI-ARIA attribute 'aria-selected' is not valid for the element <a> with WAI-ARIA role 'menuitem'

Element that is showing this error :- 
```
<a aria-selected="false" aria-expanded="false" class="bx--header__menu-item bx--header__menu-title" aria-haspopup="menu" href="javascript:void 0" tabindex="0" part="trigger" role="menuitem">
```

andy-blum:
### Related Ticket(s)
Fixes #6888 

### Description
Changes `aria-selected` attribute to `selected` to eliminate a WAI-ARIA violation

### Changelog

**Changed**

- replace "aria-selected" attribute with "selected" attribute on dds-megamenu-top-nav-menu links


packages/styles/scss/components/masthead/_masthead.scss
```diff
@@ -314,7 +314,7 @@ $search-transition-timing: 95ms;
 
       &:focus,
       &:active {
-        &[aria-selected='true'] {
+        &[selected='true'] {
```
annawen1, line 317: @andy-blum thanks! Looking at the Percy snapshots, the React Masthead L0 and L1 also uses the `aria-selected` attribute. Could we adjust the React version as well?

Screen Shot 2021-09-01 at 9 02 39 AM

annawen1: or it may make sense to switch both React and Web components to use a classname instead for styling

packages/react/src/components/Masthead/MastheadTopNav.js
```diff
@@ -64,7 +64,7 @@ const MastheadTopNav = ({ navigation, ...topNavProps }) => {
     } else {
       return (
         <HeaderMenuItem
-          aria-selected={selected ? 'true' : 'false'}
+          selected={selected ? 'true' : 'false'}
```
annawen1, line 67: looks like it should be `data-selected`?

andy-blum: Yeah 😩

annawen1 (commented):
Almost there @andy-blum! I think we also need to change the `aria-selected` to `data-selected` in `top-nav-item.ts`

Screen Shot 2021-09-02 at 4 19 12 PM

annawen1 (changes requested):
believe we just need to update the files I pointed out

annawen1 (approved):
LGTM! Both React and Web-components working as expected! Thanks!

Issue #7032, opened 2021-09-01
[Cloud-Masthead]: Cookie Authentication checking method is not working & Overlapping issue on v18 L1 Nav


proeung:
### Describe the issue
- The current `<dds-cloud-masthead-container>` is using the cookies to check whether the login status of `could.ibm.com` returns `true` or `false`, however, this method has stopped working (see - https://www.ibm.com/cloud?bust=foo)
- Also, there's an overlapping issue seen on a page where the v18 L1 Nav is visible (see - https://www.ibm.com/ru-ru/cloud/watson-natural-language-classifier?bust=1qwsdvfanskds).
  - Solution: Looks like we can adjust the z-index of `dds-cloud-masthead-container`.

<img width="1438" alt="Screen Shot 2021-08-31 at 6 27 15 PM" src="https://user-images.githubusercontent.com/1815714/131599731-4b5e5264-1a58-48a9-a1ff-b09d8f23823e.png">

----
- Hybrid Cloud developer:   Putra Bonaccorsi
- HC Jira ticket:
- DDS consultant:  Anna Wen

PR #7037, opened 2021-09-01
fix(cloud-masthead): adjust the cookie to PROD ID & overlapping issue on v18 L1 Nav

proeung:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7032

### Description

- This PR addresses the two bugs that we've discovered after the launch of the new Cloud Masthead on the Drupal PROD environment (see - https://www.ibm.com/cloud?bust=foo).
- See this issue for more detail - https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7032

### Changelog

**New**

- {{new thing}}

**Changed**

- {{changed thing}}

**Removed**

- {{removed thing}}

### DIFF
--- a/packages/services/src/services/CloudAccountAuth/CloudAccountAuth.js
+++ b/packages/services/src/services/CloudAccountAuth/CloudAccountAuth.js
@@ -13,7 +13,7 @@ import Cookies from 'js-cookie';
  * @type {string}
  * @private
  */
-const _cookieName = 'com.ibm.cloud.iam.LoggedIn.manual';
+const _cookieName = 'com.ibm.cloud.iam.LoggedIn.prod';
 
 class CloudAccountAuthAPI {
   /**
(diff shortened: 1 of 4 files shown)


packages/services/src/services/CloudAccountAuth/CloudAccountAuth.js
```diff
@@ -13,7 +13,7 @@ import Cookies from 'js-cookie';
  * @type {string}
  * @private
  */
-const _cookieName = 'com.ibm.cloud.iam.LoggedIn.manual';
+const _cookieName = 'com.ibm.cloud.iam.LoggedIn.prod';
```
jeffchew, line 16: @proeung something to consider for the future, perhaps make this configurable via environment variable, and the default is the prod cookie name?

annawen1 (approved):
LGTM!

packages/web-components/src/components/masthead/cloud/cloud-masthead-composite.ts
```diff
@@ -249,6 +249,8 @@ class DDSCloudMastheadComposite extends DDSMastheadComposite {
     }
     this._loadTranslation?.(language, dataEndpoint).catch(() => {}); // The error is logged in the Redux store
     this._loadUserStatus?.(this.authMethod);
+
+    this.style.zIndex = '999';
```
annawen1, line 253: this is a temp fix until we figure out why we can't set styles to the `cloud-masthead-container` with stylesheets

packages/services/src/services/CloudAccountAuth/CloudAccountAuth.js
proeung: Yes! I agree 💯 

I'll create a feature request issue to track this improvement.

Issue #6764, opened 2021-07-29
[Masthead] React: Change component to include logic that highlights the current and parent menu items within the Masthead dropdown + Mega Menu based on the URL


RobertaJHahn:
#### User Story

> As a `[user role below]`:
Carbon for ibm.com developer

> I need to:
create/change the `Masthead`

> so that I can:
provide the ibm.com adopter developers components they can use to build ibm.com web pages

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- **See the Epic for the Design and Functional specs information**
- No visual changes and visual QA testing required.
- The manual option stays in place.  The automatic mapping is added.
- React Prod QA testing issue (#6766 )
- Incremented func spec number to **1.2.5** https://ibm.ent.box.com/file/781809973704?s=89qwjyjy6kwbi69qnz4nzatwfv3mh2kc

![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/c62cd242-5ea8-452d-a04d-ec49cc8a21b0)
![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/99737657-691f-43c2-9da3-5546e18185e4)

#### Acceptance criteria
- [ ] Built as a pure React component/variant
- [ ] Include README for the react component and corresponding styles
- [ ] Add any necessary stable selectors
- [ ] Create codesandbox example under `/packages/react/examples/codesandbox` and include in README
- [ ] Minimum 80% unit test coverage
- [ ] Update the Carbon for ibm.com website component [data file](https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/blob/master/src/data/components.json) to be sure the web site Component Status and Storybook links are correct
- [ ] A comment is posted in the Prod QA issue, tagging Praveen when development is finished

PR #7040, opened 2021-09-01
feat(masthead): masthead L0 selected when child item is current page

kennylam:
### Related Ticket(s)

#6764 

### Description

React masthead L0 parent menu now shows selected state when the current page is a child item. Example can be seen [here](https://codesandbox.io/s/github/kennylam/masthead-submenu-csb) (must open into external window).

Untitled-2

### Changelog

**New**

- L0 parent menu shows selected state when current page is a child item

### DIFF
--- a/packages/react/src/components/Masthead/MastheadTopNav.js
+++ b/packages/react/src/components/Masthead/MastheadTopNav.js
@@ -33,7 +33,7 @@ const MastheadTopNav = ({ navigation, ...topNavProps }) => {
    * @returns {*} Top masthead navigation
    */
   const mastheadLinks = navigation.map((link, i) => {
-    const subMenuSelected = hasCurrentUrl(link, root.location.href);
+    const subMenuSelected = _hasCurrentUrl(link, root.location.href);
     const autoid = `${stablePrefix}--masthead-${topNavProps.navType}__l0-nav${i}`;
     const selected =
       link.titleEnglish === topNavProps.selectedMenuItem || subMenuSelected;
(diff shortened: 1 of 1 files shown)


packages/react/src/components/Masthead/MastheadTopNav.js
```diff
@@ -126,6 +129,29 @@ function renderNav(link, autoid) {
...
 }
 
+/**
+ * Checks if the current url exists in the list of links
+ *
+ * @param {object} obj The navigation list to search
+ * @param {string} target The URL to search for
+ * @returns {boolean} Whether or not the URL is found in the navigation list
+ */
+function hasCurrentUrl(obj, target) {
```
jeffchew, line 139: Super minor thing, wondering if we should make this a private method?

kennylam: Oh yeah sure, good call.

packages/react/src/components/Masthead/MastheadTopNav.js
```diff
@@ -126,6 +129,30 @@ function renderNav(link, autoid) {
...
+ * @param {object} obj The navigation list to search
+ * @param {string} target The URL to search for
+ * @private
+ * @returns {boolean} Whether or not the URL is found in the navigation list
+ */
+const _hasCurrentUrl = (obj, target) => {
+  const findUrl = obj => {
+    /* eslint-disable no-unused-vars */
+    for (const [key, val] of Object.entries(obj)) {
+      if (val === target) {
```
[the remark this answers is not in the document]
kennylam: I did a few variations of testing already and as long as the URL matches what's in the translation file, I didn't experience any false negatives (tested a few dozen of the URLs in the megamenu). The only thing we'll need to watch out for are query params and such. So far they're ok, but if we notice that it gets tripped up from query params (both in the translation file and the super long tracking tags that get appended) we can probably split those out and compare against the URL with no params.

Issue #6025, opened 2021-05-11
[Masthead/Footer] Accessibility - unique labels for navigation elements


kyjak:
Multiple elements with "navigation" role do not have unique labels
Each element with "navigation" role must have a unique label that describes its purpose
please note when adding labels, these need to be translated to all DOCS languages, we can help with that,
and include the label in our translation request.

`<nav class="bx--legal-nav">`

kyjak:
you can use our dev https://ibmdocs-dev.mybluemix.net/docs/en
and the tool to check errors is "IBM Equal Access Accessibility Checker", plugin for chrome

jeffchew:
@Praveen-sr can you review?

jeffchew:
SDMA squad has also reported this, but for web components. Added label to the issue.

Praveen-sr:
Ram's comments:
Duplicate of [#6023](https://app.zenhub.com/workspaces/carbon-for-ibmcom-5d449f3642eb1962336cbe52/issues/carbon-design-system/carbon-for-ibm-dotcom/6024)

Footer default - Landmark region ( On JAWS FF & iPhone Voiceover)
Issue: Found 2 navigation region and one complimentary region with in the footer section . Which is not needed.
Expected: Avoid resentence or give the first as primary navigational links and second one as secondary navigation.
@jeffchew @RobertaJHahn

PR #7042, opened 2021-09-01
fix(footer): update footer a11y labels

kennylam:
### Related Ticket(s)

#6025 

### Description

This updates missing/inconsistent accessibility labels in the footer.

### Changelog

**Changed**

- updates various accessibility labels in the `footer` component


packages/react/src/components/Footer/FooterNav.js
```diff
@@ -26,7 +26,8 @@ const FooterNav = ({ groups }) => {
   return (
     <nav
       data-autoid={`${stablePrefix}--footer-nav`}
-      className={`${prefix}--footer-nav`}>
+      className={`${prefix}--footer-nav`}
+      aria-label="Footer navigation">
```
jeffchew, line 30: Do we have translations for these?

kennylam: No, from what Ram told me the role will be read by the screen reader. This is just to suppress the validation error.

jeffchew: Good to know, thanks!

Issue #7050, opened 2021-09-02
Web component: Dotcom shell - Locale selector modal is not operable on windows machine.  


Praveen-sr:
Describe the issue: 

Issue:
Dotcom shell - Locale selector modal is not operable on windows 10 machine.  
On Firefox, version 91
chrome, version 92

Note: Working fine on Mac, Safari.

Please provide any additional notes:

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/e0b816fd-c1e8-4033-8df6-12cbfeb83128)

PR #7053, opened 2021-09-02
fix(dotcom-shell): ensure locale modal opens with scrollbars on

IgnacioBecerra:
### Related Ticket(s)
#7050

### Description
This bug occurs when scrollbars are always active. Make sure to activate this setting on MacOS for testing purposes.

 Opening the modal causes the scrollbar to disappear and resize the page, which called the scroll handler, closing the modal immediately as its considered an "out of modal" action/click event. In other words, the modal closed immediately as it opened. 

This PR ensures that this resize change only affects the code when it's actually scrolling.

### Changelog

**Changed**

- moved a variable update to within the conditional that uses its value

### DIFF
--- a/packages/web-components/src/components/dotcom-shell/dotcom-shell-composite.ts
+++ b/packages/web-components/src/components/dotcom-shell/dotcom-shell-composite.ts
@@ -169,13 +169,13 @@ class DDSDotcomShellComposite extends LitElement {
           this._masthead!.style.top = '0';
         }
       }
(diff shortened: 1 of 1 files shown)


emyarod (approved):
looks good to me, modal appears as expected now on Windows 10 and macOS with scroll bars enabled (Firefox 93+ and Chrome 93+)

ariellalgilmore (approved):
wooohooo

Issue #5857, opened 2021-04-21
[Card Section Simple] Title section and cards don't align to the 16 column grid


kirbyevans:
## Detailed description

> It doesn't appear that the Section title and cards are aligning to the 16 column grid

> This appears to be a bug.

> Card section simple component

> I expected the content on the page to align to the 16 column grid. It appears that there is extra padding applied to the content that creates a misalignment with the 16 column grid.

> Seeing the issue in Chrome on a mac.

> Issue is being seen on Storybook. https://ibmdotcom-web-components.mybluemix.net/?path=/story/components-card-section-simple--default

> AEM team.

## Steps to reproduce the issue

1. Go here: https://ibmdotcom-web-components.mybluemix.net/?path=/story/components-card-section-simple--default
2. Initiate carbon chrome plugin to inspect grid and content

## Additional information

- ![image](https://user-images.githubusercontent.com/77290329/115594069-51c27700-a2a3-11eb-89ab-5508c292d672.png)
- For more background or context please reach out to rrastog@us.ibm.com.
**- Engineering task:  Remove the demo class from Storybook.  The Carbon grid class remains.  Adjust the presentation layout using the Carbon grid.**

photodow:
Is this related to #5813?

RobertaJHahn:
Add to the Team Huddle Thursday, 4/29

RobertaJHahn:
@IgnacioBecerra Do you know whether arora-yash is going to finish this work?  Do you have to take this work on the get it finished?

IgnacioBecerra:
Update on this issue, @NSDrowned and I will be taking over the work for this issue as Yash is currently sick and hasn't been able to work on this.

cc: @RobertaJHahn

PR #7062, opened 2021-09-02
fix(card-section-simple): aligning component in storybook

IgnacioBecerra:
### Related Ticket(s)
#5857

### Description
In the current Storybook example in Card Section Simple, it still has the old demo classes. This PR ensures that it only uses the `bx--grid` classes only so it aligns to the grid properly.

### Changelog
**Changed**

- Card Section Simple story now uses `bx--grid` classes

**Removed**

- removed demo classes from the Card Section Simple story in Web Components


jeffchew (approved):
LGTM!

PR #7071, opened 2021-09-07
fix(image-with-caption): update expressive modal prop to expressive-size

ariellalgilmore:
### Related Ticket(s)

#7073 

### Description

Image with caption modal not displaying properly in large breakpoints

### Changelog

**Changed**

- updated prop from `size` to `expressive-size`


jeffchew (approved):
LGTM!

Issue #6878, opened 2021-08-12
[Web-components] LightboxMediaViewer (Embedded with VideoPlayer): Allow for title & description override


annawen1:
### The problem
The title and description in the lightbox-media-viewer with video player always renders the title and description generated by the API. Adopters may sometimes need to adjust the copy to fit their needs.

<img width="1719" alt="Screen Shot 2021-08-12 at 3 57 24 PM" src="https://user-images.githubusercontent.com/54281166/129261341-54711d6b-f2e4-41c4-9525-c20164ef9ed5.png">

### The solution
Allow for override of the title and description. This has already been done for the React version.

### Additional information
- Functional Specs: https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/wiki/Lightbox-media-viewer#22-heading

PR #7083, opened 2021-09-08
feat(lightbox-media-viewer): title & description override

guilhermelMoraes:
### Related Ticket(s)

[Web-components] LightboxMediaViewer (Embedded with VideoPlayer): Allow for title & description override #6878

### Description

 - CTA's now can provide a custom `title` and `description` for video;
 - Video `playing-mode="lightbox"` now uses custom caption as the custom title;
 - Video `playing-mode="lightbox"` allows the user to provide a custom `description`.


packages/web-components/src/components/video-player/video-player-composite.ts
```diff
@@ -191,6 +198,7 @@ class DDSVideoPlayerComposite extends HybridRenderMixin(HostListenerMixin(LitEle
         duration="${ifNonNull(duration)}"
         ?hide-caption=${hideCaption}
         name="${ifNonNull(caption || name)}"
+        video-description="${ifNonNull(customVideoDescription)}"
```
annawen1, line 201: I see `video-description` is a new attribute being passed into `video-player`,  but looking at the `video-player` component, where is the attribute being used?

annawen1 (approved):
LGTM!

Issue #5731, opened 2021-04-08
[Content group pictograms] Storybooks - pictogram size issue


DragosRistici:
## Both Storybooks (React and Web Components)

 General issue: Make sure the pictograms are 80x80 on both Storybooks, as in the specs file.

<img width="653" alt="Screenshot 2021-04-08 at 15 49 36" src="https://user-images.githubusercontent.com/48327647/114029284-07dc9880-9882-11eb-88a9-c53de7035f94.png">

<img width="816" alt="Screenshot 2021-04-08 at 15 50 51" src="https://user-images.githubusercontent.com/48327647/114029470-35c1dd00-9882-11eb-889c-5c288572bdd5.png">

PR #7086, opened 2021-09-08
fix(content-group-pictogram): icon size fix and alignment

NSDrowned:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5731

### Description

Make sure the pictograms are 80x80 on both Storybooks, as in the specs file

### Changelog

**Changed**

- Fixed SVG size for web components
- Fixed mobile alignment for svg and content in web components


jeffchew:
Is size supposed to be adjusted for all stories?

image

image

RichKummer:
@jeffchew The original ticket only calls out the Storybooks, but I think ideally we would have the SVG size updated in the example pages.

NSDrowned:
@jeffchew, @RichKummer I'll fix the storybook variations 👍

NSDrowned:
@RichKummer, SVGs have been updated 👍

RichKummer:
It looks like the SVGs are solid on the storybook links, but I am still seeing the NextJS and HTML examples with smaller SVGs. It looks like the container for it is 80x80 but the SVG shows up smaller inside the container.
NextJS-HTML

NSDrowned:
@RichKummer, the nextjs template was using outdated icons, I updated the icon versions and the problem has been fixed:

image

RichKummer:
@NSDrowned I'm seeing it updated in the NextJS template (thank you!). I'm assuming it's the same issue on the HTML template, but that one doesn't look like it was updated. Is that a quick fix we can make to ensure we're consistent?

NSDrowned:
@RichKummer you're right, let me fix that right away

NSDrowned:
@RichKummer icons should have the correct size now 👍

RichKummer (approved):
LGTM! Thanks for making the last change.

ariellalgilmore (approved):
LGTM!

PR #7098, opened 2021-09-09
fix(card-in-card): fixed transition animation speed on card image

andysherman2121:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom-nextjs-template/issues/39

### Description

The image on card in card wasn't receiving a transition delay so the hover between the card and the image were out of sync.


packages/styles/scss/components/card-in-card/_card-in-card.scss
```diff
@@ -98,6 +102,7 @@
       opacity: 0;
       z-index: 1;
       background-color: $inverse-02;
+      transition: all 150ms cubic-bezier(0.2, 0, 0.38, 0.9) 0s;
```
photodow, line 105: We should be leveraging tokens from [Carbon motion guidelines](https://www.carbondesignsystem.com/guidelines/motion/overview/) for our transitions. Also, any chance we could get more selective with our transitions? Helps a little with debugging, but also performance to my understanding.

```suggestion
      transition: opacity $moderate-01 motion('standard', 'productive');
```

packages/styles/scss/components/card-in-card/_card-in-card.scss
```diff
@@ -44,6 +44,10 @@
     }
 
     .#{$prefix}--card__wrapper {
+      &:hover {
+        transition: all 150ms cubic-bezier(0.2, 0, 0.38, 0.9) 0s;
+      }
```
photodow, line 49: Is this needed? It appears to still work once removed, and you have the transition at the `bx--card` level still. Unless there is something I'm missing here?

```suggestion
```

packages/styles/scss/components/card-in-card/_card-in-card.scss
```diff
@@ -98,6 +102,7 @@
       opacity: 0;
       z-index: 1;
       background-color: $inverse-02;
+      transition: all 150ms cubic-bezier(0.2, 0, 0.38, 0.9) 0s;
     }
 
     .#{$prefix}--image__img {
```
photodow, line 108: Is there a good way we could apply it to the card image as well? I also noticed we're using different techniques. Card in card uses opacity on a psuedo element while card image uses filter. Just thinking out loud we do something like this?

```scss
/* image component */
[data-autoid*=dds--image] {
    filter: brightness(1);
    transition: opacity $moderate-01 motion('standard', 'productive');
}

/* card component */
.bx--card:hover [data-autoid*=dds--image] {
    filter: brightness(.8);
}
```

Feel free to poke holes, but at a minimum could we fix card with image? It appears to be a similar, if not the same issue.

photodow (approved):
Just a couple of comments for you to consider before merging. Looks like card in card was working at least. Thanks!

packages/styles/scss/components/card-in-card/_card-in-card.scss
jeffchew: Agreed, targeting `all` is known to reduce performance.

packages/styles/scss/components/card-in-card/_card-in-card.scss
andysherman2121: yeah doesn't look like this is necessary anymore

PR #7101, opened 2021-09-09
fix(masthead): add check for focus elem in Masthead SideNavMenuSection to prevent type error

annawen1:
### Related Ticket(s)

### Description

Type error was occurring when user hovered over the mobile nav items in smaller breakpoints. This was happening because there is a `transitionend` listener needed to provide focus for tabbing through the menu. Add a check for the `focusElement` to prevent error.

### Changelog

**Changed**

- add check for `focusElement` before setting focus

### DIFF
--- a/packages/react/src/components/carbon-components-react/UIShell/SideNavMenuSection.js
+++ b/packages/react/src/components/carbon-components-react/UIShell/SideNavMenuSection.js
@@ -41,12 +41,11 @@ const SideNavMenuSection = ({
           'transitionend',
           function focus(event) {
             if (
-              focusElement &&
-              (event.propertyName === 'left' ||
-                event.propertyName === 'background-color' ||
-                event.propertyName === 'color')
+              event.propertyName === 'left' ||
+              event.propertyName === 'background-color' ||
+              event.propertyName === 'color'
             ) {
-              focusElement.focus();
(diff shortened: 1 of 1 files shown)


jeffchew (approved):
thank you @annawen1 !

packages/react/src/components/carbon-components-react/UIShell/SideNavMenuSection.js
```diff
@@ -41,9 +41,10 @@ const SideNavMenuSection = ({
...
             if (
-              event.propertyName === 'left' ||
-              event.propertyName === 'background-color' ||
-              event.propertyName === 'color'
+              focusElement &&
+              (event.propertyName === 'left' ||
+                event.propertyName === 'background-color' ||
+                event.propertyName === 'color')
             ) {
               focusElement.focus();
```
emyarod, line 49:
```suggestion
              focusElement?.focus();
```

would this work as well?

annawen1: ah yes! 🙏 Thank you @emyarod, I forget about the optional chaining sometimes - looks so much cleaner!

kennylam (approved):

Issue #6319, opened 2021-06-09
[Callout with Media][Web components | HTML test page] Solution page - text color is same as background color


Praveen-sr:
## Detailed description

Issue:
Dotcom Shell & Solution page - In CallOut Media, text color is same as background color. 

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/5678a065-99d9-4102-bafc-a56c1a241854)

PR #7127, opened 2021-09-13
fix(callout-with-media): update copy color

emyarod:
### Related Ticket(s)

#6319

### Description

This PR updates the copy color token in callout with media so that it is visible on the `inverse-02` background in the web components HTML test app

image


jeffchew (approved):
LGTM!

Issue #6417, opened 2021-06-18
[Feature card] Web Components:  Feature card large & medium wrong stable selector


photodow:
## Detailed description

In Web Components our large & medium size feature card has the wrong stable selector. It is supposed to be:
`dds--feature-card` (medium) and `dds--feature-card-large` (large)

Correctly in react:  https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/packages/react/src/components/FeatureCard/FeatureCard.js#L41-L43

emyarod:
maybe relevant: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/packages/web-components/docs/stable-selectors.md#data-autoid-support-for-partial-backward-compatibility

I can update the feature card stable selector to be `dds--feature-card` but it won't be dynamic based on the card size. I believe if a stable selector for the web component is required, checking the tag name and the value of the `size` attribute should help determine whether or not it is a medium or large feature card

RobertaJHahn:
@jeffchew I am not sure who should determine the correct implementation strategy for this bug.  Can you take a look and let me know if we have to include in one of the engineering meetings?

cc:  @emyarod

jeffchew:
Speaking with @emyarod , the feature card appears to be the only component with dynamic auto-id. In React, the `feature-card-large` selector isn't even documented in Storybook docs, though is unknown whether if teams are actively using that stable selector.

This might be a huddle topic to discuss first, to see if we should stick with static stable selectors and do something about this particular case.

PR #7146, opened 2021-09-14
fix(feature-card): update stable selector

emyarod:
### Related Ticket(s)

#6417

### Description

This PR updates the feature card stable selector to be `dds--feature-card` rather than `dds--card`

image

question for the reviewers: should the React stable selector be updated? I'm not sure if the dynamic stable selector is intentional https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/packages/react/src/components/FeatureCard/FeatureCard.js#L39-L42

if so, then this I don't believe this behavior is replicable in web components, although it's not as significant in web components to have the stable selector be a class if I understand correctly

### Changelog

**Changed**

- feature card stable selector


jeffchew:
React has the functionality but doesn't appear to be documented: https://www.ibm.com/standards/carbon/react/?path=/docs/components-feature-card--default

Stable selectors are meant for pretty much any way to reliably reference the component on the page in things like A/B tests, analytics, e2e tests, etc. One example at least I was thinking of for analytics is finding a way for Google Analytics to determine how many (and which) components are being used across all of IBM.com. This type of tracking can help determine if our components are (or aren't) being used on the site and can help with decisions of decommissioning the ones that aren't being used often (or at all).

jeffchew (approved):
I think we're good with setting it as `dds--feature-card` for now, I'm wondering if it needs a code restructure if we want a different stable selector (or maybe even a different custom element name) for feature card large to handle this.

emyarod:
@jeffchew yeah it looks like feature card is the only React component with a dynamic stable selector (depending on the value of `props.size`). all other components seem to have a static stable selector

Issue #6424, opened 2021-06-19
[Feature card - medium] Styles bugs


shixiedesign:
Some styling issue with both React & Web components: 

### 1. Web component: arrow size between `lg--xlg` are too big.  

**Currently 32px**
![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/16f2f7ff-e7fc-40de-a64b-3402504dc748)
**Expecting 24px**
![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/14f84111-09b1-4fcd-b18e-15a0d4963736)

### 2. Both Web comp & React: card hover effect is broken 

**Current**

![ad9cb669-8b02-4eab-b633-b561220423db](https://user-images.githubusercontent.com/8265238/133324123-017c764d-862e-471b-b7a6-e082ec62aa71.gif)

**Expecting**

Expecting the image or video poster image to darken on light themes, and lighten on dark themes. This will match the card text portion background's hover effect.

| White & Gray 10 themes | Gray 90 and Gray 100 themes |
| --- | --- |
| `darken: 8%` | `lighten: 8%` |

Transition easing & duration should match the current text portion's background change: 

Transition duration: 150ms
Transition easing: `cubic-bezier(02, 0, 0.38, 0.9)`

### 3. Both react & web comp:`sm` breakpoint arrow should be bottom pinned, currently floating far from bottom edge of card. See gif above 👆

shixiedesign:
@shixiedesign testing

PR #7149, opened 2021-09-14
fix(feature-card): resolve styling issues

emyarod:
### Related Ticket(s)

#6424

### Description

* Web component: arrow size between `lg--xlg` should be 24px

image

* hover transition should be as expected across the card

* image overlay should also darken on light themes and lighten on dark themes now

reviewers should confirm that the removed styles in the web component stylesheet do not cause a regression


jeffchew (approved):
LGTM!

shixiedesign:
1. CTA - Feature, WC, is showing the arrow using a wrong color. Is this outside the scope of this PR? 
image

2. Hover state animation should be in scope since it's captured in the bug ticket. I'm still seeing the "flicker" on hover. Happy to jump on a call to explain this. It's subtle but hard to unsee. 
Sep-16-2021 10-19-00

emyarod:
@shixiedesign issues should all be resolved on feature card now!

for the Feature CTA, I updated the icon color and size, and the hover interaction should also be fixed as well. but if the icon alignment issue also applies, I think I should separate that since it is affected by Card/CTA base styles whereas the standalone feature card isn't (and should be visually correct)

shixiedesign (approved):
Fantastic! Looks great. I realize these issues are also present in Feature card - large. I will make a new issue though. Thanks for the fix!

shixiedesign:
@jeffchew @emyarod Percy results are looking weird, all changed snapshots are "new snapshots". Any clue if Percy is having a bad day?

jeffchew:
> @jeffchew @emyarod Percy results are looking weird, all changed snapshots are "new snapshots". Any clue if Percy is having a bad day?
> image

I was just looking at this too. The snapshot names changed from pipes to forward slashes for some reason...

PR #7157, opened 2021-09-15
fix(cta-block): clean up spacing

### LINKED ISSUE #5779: [CTA block] spacing clean up - DDS consulting
#### Overview
Restore original design.  Reference the spec below. Specifically the spacing and grid alignment need to be fixed.

Some spacing adjustments needed: 

----
- Hybrid Cloud developer:   John kaeser
- HC Jira ticket:  https://jsw.ibm.com/browse/HC-2172
- DDS consultant:   Julio Marroquin

jkaeser:
### Related Ticket(s)

Resolves #5779

### Description

This PR adjusts some spacing rules for the `dds-cta-block` element and its intended subcomponents so they better align to the design spec.

_Before:_
Screen Shot 2021-09-15 at 12 19 14 PM

_After:_
Screen Shot 2021-09-15 at 12 18 54 PM

### Changelog

**Changed**

- reduce spacing between `dds-cta-block`'s `copy` and `action` slots
- align `dds-cta-block`'s intended subcomponents to Carbon grid

### DIFF
--- a/packages/web-components/src/components/cta-block/cta-block.scss
+++ b/packages/web-components/src/components/cta-block/cta-block.scss
@@ -28,7 +28,6 @@
   }
 
   ::slotted([slot='action']) {
-    margin-top: $carbon--layout-01;
     margin-bottom: $carbon--layout-04;
   }
 
@@ -45,6 +44,7 @@
 
     display: inline-block;
     max-width: 640px;
+    padding-left: 0;
 
     @include carbon--breakpoint('sm') {
       width: 100%;
@@ -53,8 +53,11 @@
     @include carbon--breakpoint('md') {
       width: 90%;
     }
+  }
 
-    padding-left: 0;
+  ::slotted(#{$dds-prefix}-button-group) {
+    // <dds-button-group-item> elements have top margins
+    margin-top: -#{$carbon--layout-02};
   }
 
   ::slotted(#{$dds-prefix}-link-list) {
--- a/packages/web-components/src/components/cta-block/cta-block.scss
+++ b/packages/web-components/src/components/cta-block/cta-block.scss
@@ -11,10 +11,9 @@
 :host(#{$dds-prefix}-cta-block) {
   @include cta-block--pattern;
 
+  display: block;
+
   .#{$prefix}--content-layout {
-    @include carbon--breakpoint('md') {
-      padding-left: 0;
-    }
     @include carbon--breakpoint('sm') {
       padding-left: $carbon--layout-01;
     }
@@ -27,8 +26,6 @@
   &.#{$prefix}--cta-block__border {
     border-bottom: 1px solid $ui-04;
   }
-
-  display: block;
 
   ::slotted([slot='action']) {
     margin-top: $carbon--layout-01;


packages/web-components/src/components/cta-block/cta-block.scss
```diff
@@ -56,8 +53,11 @@
     @include carbon--breakpoint('md') {
       width: 90%;
     }
+  }
 
-    padding-left: 0;
+  ::slotted(#{$dds-prefix}-button-group) {
+    // <dds-button-group-item> elements have top margins
+    margin-top: -#{$carbon--layout-02};
```
jkaeser, line 60: I'm not in love with this solution. Is there a way to apply style rules to deeply nested subcomponents? I.e., the following doesn't work, but is there a way to achieve its goal?

```
dds-cta-block dds-button-group[slot="action"] > dds-button-group-item {
  margin-top: 0;
}
```

packages/web-components/src/components/cta-block/cta-block.scss
```diff
@@ -11,10 +11,9 @@
 :host(#{$dds-prefix}-cta-block) {
   @include cta-block--pattern;
 
+  display: block;
```
jkaeser, line 14: Code cleanup. Top level rules should appear before nested ones.

shixiedesign (approved):
Visually the issue is fixed! I cannot speak to your comments on the code though. Let's wait for a dev review. Thank you!

jkaeser:
Thank you for the review, @shixiedesign! Would you mind taking a look at my comment on the related issue (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5779#issuecomment-919200662)?

jeffchew (commented):
This looks like this had an impact on all button groups, was this the intent?

@shixiedesign can you take a look? Both react and web components percy diffs were impacted.

packages/styles/scss/components/buttongroup/_buttongroup.scss
```diff
@@ -34,7 +34,7 @@
   .#{$prefix}--buttongroup-item,
   :host(#{$dds-prefix}-button-group-item),
   :host(#{$dds-prefix}-button-cta) {
-    margin-top: $carbon--layout-01;
+    margin-bottom: $carbon--layout-01;
```
jeffchew, line 37: @jkaeser this looks like this changes all instances of button group across all components, not sure if this was the intent?

jeffchew (changes requested):

packages/styles/scss/components/buttongroup/_buttongroup.scss
jkaeser: @jeffchew The intent here was to leverage Percy to see how many components are affected by this change, then make a decision about how to move forward. The existing styles contribute to the spacing issues reported in #5779.

jkaeser: Particularly, the challenge I'm facing here relates to how two deeply nested web components each have margin rules that combine to create a gap that is too large, and the `<dds-cta-block>` component's CSS cannot reach into their shadow roots to override those margins.

@NSDrowned and I discussed, and the two options we came up with were:
1. Extend the `<dds-button-group-item>` into something like `<dds-cta-button-group-item>` so that I can control it's CSS directly.
2. Change the global styling and see how many regressions it causes.

Option 1 feels a little heavy handed, so we agreed to try option 2 first.

jeffchew: Adding @guilhermelMoraes here, as @NSDrowned became a father yesterday!

shixiedesign:
Hi @jkaeser i just answered your question on spacing here: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5779#issuecomment-921716957

I really should have documented the bug more specifically, sorry! But taking a look, changing button group as a whole is too risky right now. It impacts a lot of components. I'd say if fixing the spacing between copy and button group is too hard otherwise, we can temporarily leave it as is (a bit bigger than expected). This does make me wonder if button group is looking correct. Here's a visual of what's going on: 

image

If there's erronous padding on button group, we might need a separate epic to refactor button group and fix regression on all the components touched.

jkaeser:
@shixiedesign IMHO taking a more considered look at the button group spacing would be the best decision in the long term. The margin-top on the buttons exists to create vertical space between buttons when they wrap to multiple rows. The fact that it also creates the correct amount of space between components is a side effect, and as this issue demonstrates, that side effect is not always desirable.

For now, I will revert the button margin change and leave the negative margin solution in as a temporary fix. I'll also go ahead and make any necessary changes to address https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5779#issuecomment-921716957.

Thanks, everyone!

kennylam (approved):
LGTM thanks @jkaeser!

shixiedesign (approved):
Thank you! LGTM!

PR #7188, opened 2021-09-17
fix(tabs-extended): Add keyboard navigation

### LINKED ISSUE #5816: Web component: Tabs extended &  Tabs extended with media - Keyboard tab focus issues. - DDS consulting
## Detailed description

Issue:
For keyboard only users
1.Each tabs are receiving the tab focus.
2.We are not able to navigate between the tabs with right/left arrow keys.

Expected:
The only selected tab should receive the tab focus and rest of the tabs should be navigable with right/left arrow keys.

## Additional information
-  Engineering info:  
     - HC engineer:
     - JC Jira Ticket: https://jsw.ibm.com/browse/HC-2191
     - DDS consulting engineer:   Anna Wen

andy-blum:
### Related Ticket(s)

Fixes #5816

### Description

Adds keyboard navigation that follows [W3 guidance](https://www.w3.org/TR/wai-aria-practices-1.1/examples/tabs/tabs-1/tabs.html#kbd_label)

- right arrow goes to next non-disabled tab
- left arrow goes to previous non-disabled tab
- left and right arrows wrap between first and last tabs
- home goes to first non-disabled tab
- end goes to last non-disabled tab
- tab gives focus to tabpanel

### Changelog

**New**

- Adds keyboard navigation to dds-tab


andy-blum:
Realized as I was writing the PR that I didn't account for left/right arrow movements in RTL. Will work on that facet monday morning.

packages/web-components/src/components/tabs-extended/tabs-extended.ts
```diff
@@ -32,23 +32,25 @@ class DDSTabsExtended extends StableSelectorMixin(LitElement) {
...
 
   /**
    * Defines the active tab index.
    */
   @internalProperty()
-  private _activeTab: Number = 0;
+  private _activeTab: number = 0;
+
+  @internalProperty()
+  private _isLTR: boolean = false;
```
annawen1, line 44: should we be using `_isRTL` instead? I believe most pages will be `LTR` so making the default value to false doesn't make sense to me

annawen1: Also testing out the WC RTL deploy preview, it seems to be focusing on the opposite element? When I'm focused on the `First Tab` of `tabs-extended` and I pressed the left arrow, I expect the focus to be on the `Second Tab` but instead it focuses on the `Fourth Tab`

andy-blum: 🤦 I started it as isRTL, then switched to isLTR to make the code more easily readable. Thoughts on making this property default to `true` instead?

Also, the property sets in `updated`, but I wasn't sure if maybe it would be better in `connectedCallback`? We probably don't anticipate much switching between RTL/LTR after page load, do we?

annawen1: ah ok, yeah let's make the `isLTR` default to `true` instead and I'm fine with it in the `updated()` since yeah I don't think there will be much switching between RTL and LTR.

Also, is the RTL navigation I mentioned in the comment above the expected behavior?

andy-blum: @annawen1 can you double check the WC RTL deploy preview? Maybe it hadn't finished updating with new code, but it's working for me now.

annawen1: @andy-blum oh yep, opened it in incognito and it works as expected now

annawen1 (approved):
LGTM! Thanks @andy-blum!

andysherman2121 (approved):
Nice work! @andy-blum

Issue #6182, opened 2021-05-25
[Web components] update docs for ButtonGroup


diegomarcuz:
## Detailed description

> Describe in detail the issue you're having.

I would use the `@click` event on [buttongroup](http://ibmdotcom-web-components.mybluemix.net/?path=/docs/components-button-group--default) web component, but it is missing this indication on its docs.

PR #7195, opened 2021-09-20
fix(button-group): added documentation

andysherman2121:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6182

### Description

Button group was missing documentation for how to add an @click to button group items.


packages/web-components/src/components/button-group/__stories__/README.stories.mdx
```diff
@@ -49,6 +49,13 @@ import 'carbon-web-components/es/icons/arrow--right/20.js';
 </dds-button-group>
 ```
 
+### @click example
+
+```html
+<dds-button-group>
+  <dds-button-group-item @click=${this._handleClick}></dds-button-group-item>
```
jeffchew, line 56: Might want others' opinions as well, but wondering if the example method should be something like `${myCustomClickHandler}` rather than `${this_handleClick}`?

jeffchew (approved):
Thanks @andysherman2121 !

Issue #7181, opened 2021-09-16
[Cloud Masthead]: API authentication method is not working on Cloud Docs dev application - DDS Consulting


proeung:
### The problem
- The Cloud Docs/Knowledge Center team that's adopting our `<dds-cloud-masthead-container>` is having trouble getting the API authentication method to work correctly within their dev environments.
- `devtest` - https://dev.console.test.cloud.ibm.com/docs | `devprod` https://dev.console.cloud.ibm.com/docs
- Below is the markup that they've implemented.

```
<dds-cloud-masthead-container
  platform="Cloud"
  auth-method="api"
  data-endpoint="/common/carbon-for-ibm-dotcom/translations/cloud-masthead"
></dds-cloud-masthead-container>
```
- This issue is seen on the latest and canary versions.

### Additional information
- API - (see [link](https://cloud.ibm.com/api/v6/selected-account?profile=true))
- Docs/Console authentication fetching method - see [link](https://github.ibm.com/Bluemix/docs-components-react/blob/master/src/components/DocsHeader/DocsHeader.js#L97)

----
### HC Ticket Tracking
- **Ticket:** https://jsw.ibm.com/browse/HC-2196
- **Assignee:** @huntermacd 
- **DDS Engineer:** @annawen1

PR #7199, opened 2021-09-20
fix(cloud-masthead): API authentication method is not working on Cloud Docs dev application

huntermacd:
### Related Ticket(s)

https://jsw.ibm.com/browse/HC-2196

GH issue: #7181 

### Description

The Cloud Docs/Knowledge Center team that's adopting our `<dds-cloud-masthead-container>` is having trouble getting the API authentication method to work correctly within their dev environments.

### Changelog

**New**

- n/a

**Changed**

- updated Cloud masthead state interface
- updated API response check to use status code because Axios does not provide an `ok` key like native fetch does

**Removed**

- n/a


annawen1 (approved):
LGTM!

Issue #7129, opened 2021-09-13
[Footer] Resolve a11y issues identified by storybook in web components


jeffchew:
Describe the issue: 

> What did you expect to happen? What happened instead? 

Address all accessibility violations identified in storybook:

![image](https://user-images.githubusercontent.com/1641214/133140677-11216672-4ab6-4575-9077-7c26276ea092.png)

> Component(s) impacted? Please include screenshots, or code and relevant URLs.

Footer in web components

PR #7214, opened 2021-09-21
fix(footer): addressing accessibility issues

guilhermelMoraes:
### Related Ticket(s)

[Footer] Resolve a11y issues identified by storybook in web components #7129

### Description

**NOTE**
In order to resolve this a11y issue below, I've set the `hasMainTag` to `true` in the `footer.stories.ts`, as this removes the conflict between the landmarks - the one from the footer itself and the other set by the story container.

Screen Shot 2021-09-21 at 12 58 07

The a11y checker is complaining about the `dds-legal-nav-item` as direct children of a ul element. Turns out that the `dds-footer-nav-group` is inside a ul tag as well, and the a11y checker isn't complaining about it. Both the `dds-legal-nav-item` and the `dds-footer-nav-group` has `role="listitem"`, so I'm not really sure what's causing this.
Tried a few approaches to get rid of this serious accessibility issue but none of them worked. 

Screen Shot 2021-09-28 at 16 23 28


jeffchew (approved):
LGTM!

IgnacioBecerra (approved):

Issue #7211, opened 2021-09-21
[Cloud-Masthead]: Calculate the offset top-position of the Mobile Left-Nav component when there's a local banner


proeung:
### Description

- When visiting local pages within Drupal, you're prompted with a local banner that sits on top of the `<dds-cloud-masthead-container>` component.
- The issue is related to the mobile slide-out menu where the top offset position is not currently defined, thus users are not able to close the menu out.
- We need to come up with a solution that calculates the top off-set position of the masthead and apply the value to the top position of the left-nav.

### Component(s) impacted

Cloud Masthead

### Browser

Chrome, Safari, Firefox, Microsoft Edge

### Carbon for IBM.com version

v1.24

### Severity

Severity 1 = The design is broken in a critical way that blocks users from completing tasks or damages the brand. Affects major functionality, no workaround.

### Application/website

Drupal CMS

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://www.ibm.com/sg-en/cloud/application-modernization

### Steps to reproduce the issue (if applicable)

- Visit one page of the local pages within the `/cloud` web segment in Drupal (eg. https://www.ibm.com/sg-en/cloud/application-modernization) on smaller breakpoints.
- See video screenshot

https://user-images.githubusercontent.com/1815714/134199100-7882cfb6-8cad-45d5-991d-ec963aa72145.mov

### Release date (if applicable)

Patch to v.1.25, if possible.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #7215, opened 2021-09-21
fix(left-nav): calculate offsetTop to set top of left-nav

annawen1:
### Related Ticket(s)

[Cloud-Masthead]: Calculate the offset top-position of the Mobile Left-Nav component when there's a local banner #7211

### Description

Calculate the offsetTop of the masthead, and set the `top` positioning of the `left-nav` and `left-nav-overlay`.

Fix looks something like this when there is a banner above:
Screen Shot 2021-09-21 at 3 46 15 PM

### Changelog

**New**

- get `offsetTop` from masthead and set `top` styling of `left-nav` and `left-nav-overlay`

### DIFF
--- a/packages/web-components/src/components/masthead/left-nav.ts
+++ b/packages/web-components/src/components/masthead/left-nav.ts
@@ -136,6 +136,12 @@ class DDSLeftNav extends StableSelectorMixin(BXSideNav) {
       if (expanded) {
         this._hFocusWrap = focuswrap(this.shadowRoot!, [startSentinelNode, endSentinelNode]);
         doc.body.style.overflow = `hidden`;
+
+        // TODO: remove this logic once masthead can account for banners.
+        // calculate top offset to display left-nav correctly when page has banner above masthead
+        const masthead: HTMLElement | null = doc?.querySelector('dds-masthead');
+        const mastheadTopOffset = masthead?.offsetTop;
+        this.style.top = `${mastheadTopOffset}px`;
       } else {
         const { selectorMenuSections, selectorFirstMenuSection } = this.constructor as typeof DDSLeftNav;
         doc.body.style.overflow = `auto`;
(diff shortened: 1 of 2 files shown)


packages/web-components/src/components/masthead/left-nav.ts
```diff
@@ -136,6 +136,12 @@ class DDSLeftNav extends StableSelectorMixin(BXSideNav) {
       if (expanded) {
         this._hFocusWrap = focuswrap(this.shadowRoot!, [startSentinelNode, endSentinelNode]);
         doc.body.style.overflow = `hidden`;
+
+        // TODO: remove this logic once masthead can account for banners.
+        // calculate top offset to display left-nav correctly when page has banner above masthead
+        const masthead: HTMLElement | null = doc?.querySelector('dds-masthead');
+        const mastheadTopOffset = masthead?.offsetTop;
+        this.style.top = `${mastheadTopOffset}px`;
```
proeung, line 144: I think this top calculation works great on page load, however, I think we need to bind this to a window resize event when users decided to click on the "Stay here" button, which closes the locale banner, thus the top offset position value will need to be updated.

Build_and_modernize_-_Singapore___IBM

proeung (changes requested):
Looks good overall. I just have a minor comment about a use case where the banner goes away.

packages/web-components/src/components/masthead/left-nav.ts
proeung: This is an edge case and might not be needed for this patch.

proeung (approved):

packages/web-components/src/components/masthead/left-nav.ts
annawen1: tried using `resizeObserver` to detect body height change when the banner disappears but it doesn't seem to detect the change locally. Hoping that https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7216 will help to incorporate the masthead with some awareness of components above and that it will resolve the edge case that we're seeing here

PR #7223, opened 2021-09-22
fix(masthead): fix focus wrapping and menu-section paging

### LINKED ISSUE #6625: [Masthead] Mobile - On tab key from close icon focus is skipping platform ('IBM cloud') and jumping to the 'Product and Solution' - DDS consulting
## Detailed description

Issue:
Hitting tab key from close icon focus is skipping platform 'IBM cloud' and jumping to the 'Product and Solution'

Expected:
Tab focus should go the the IBM cloud first then to the product and solution.

## Additional information
-  Engineering info:  
     - HC engineer: @andy-blum 
     - JC Jira Ticket: https://jsw.ibm.com/browse/HC-2188
     - DDS consulting engineer:   Anna Wen

andy-blum:
### Related Ticket(s)
Fixes #6625 

### Description
This PR updates the left-nav, left-nav-menu-section, and left-nav-name components to ensure a consistent paging and focus wrapping experience

### Changelog

**New**

- New `platform-id` slot in `left-nav` for `left-nav-name` component

**Changed**

- Pagination logic between `left-nav-menu-section` components


packages/web-components/src/components/masthead/left-nav-menu-section.ts
```diff
@@ -166,17 +181,21 @@ class DDSLeftNavMenuSection extends HostListenerMixin(FocusMixin(LitElement)) {
...
+          document.addEventListener(
+            'transitionend',
+            () => {
+              (tabbable as HTMLElement).focus();
+            },
+            { once: true }
+          );
         }
       } else {
+        console.log(`Contracting ${sectionId}`);
```
annawen1, line 197: oh rogue `console.log` left behind!

packages/web-components/src/components/masthead/left-nav.ts
```diff
@@ -225,9 +222,16 @@ class DDSLeftNav extends StableSelectorMixin(BXSideNav) {
...
+      <div class="${prefix}--side-nav__wrapper">
+        <a id="start-sentinel" class="${prefix}--visually-hidden" href="javascript:void 0" role="navigation"></a>
+        <div class="${prefix}--side-nav__platform-name">
+          <slot name="platform-id"></slot>
+        </div>
+        <div class="${prefix}--side-nav__menu-sections">
+          <slot></slot>
+        </div>
+        <a id="end-sentinel" class="${prefix}--visually-hidden" href="javascript:void 0" role="navigation"></a>
+      </div>
```
proeung, line 234: @andy-blum Looks like some of the refactoring work here is causing a regression where the `slot="platform-id"` is being added to the `<dds-top-nav-name>` component and causing the platform text value to not be placed in the correct slot of the shadow root (see Percy result - https://percy.io/538fc19a/Carbon-for-IBM.com-Web-Components/builds/12861003/changed/726831191?browser=chrome&browser_ids=16&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1280&widths=320%2C1280).

Can you take a look?

proeung (changes requested):

packages/web-components/src/components/masthead/left-nav.ts
annawen1: yeah the `dds-cloud-top-nav-name` component extends `left-nav-name` so adding the slot affects the top-nav-name too. Instead of setting a default value for the `slot` attribute in `left-nav-name`, we could can just pass it in `cloud-masthead-composite.ts` like so:

```
 <dds-left-nav-name slot="platform-id" href="${ifNonNull(platformAltUrl)}">${platform}</dds-left-nav-name>
```

That will put the `dds-left-nav-name` into the `platform-id` slot within `left-nav` without setting a default value for it if there is no `slot` value passed in. I would also make sure to do the same in `masthead-composite.ts` so that the global masthead is aligned with these changes as well!

andy-blum: How about conditionally setting the slot based on it's placement?

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/7223/files?file-filters%5B%5D=.ts#diff-8c78a6780e0e06ea091179bf82f49ee0b97c487f62b65371910d9ec73ed1c8d4R28-R40

annawen1: yeah that would work too! 👍

annawen1 (approved):
LGTM!

proeung (approved):

Issue #6700, opened 2021-07-21
[Content item] expose "sub-component" as simple component in storybook


shixiedesign:
### Design spec

**Content item [is here](https://ibm.box.com/s/xfm17x7dwbe4z7kjk482b8hbpjasft7o).**  
From Dragos via #6190

![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/1ad7d8d1-b835-4008-9df3-9a6099e41837)

### Functional spec

https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/wiki/Content-item

### Proposed storybook knobs

**Content item**
Heading (textarea)
Add media (select)
	– Image
	– Video
Description (textarea)
CTA copy (textarea)
CTA type (select)

#### Additional Information
- Visual QA testing issue (#6790)
- Prod QA testing issue (#6791)

### Acceptance criteria 
- [ ] Content group added to storybook & Readme added to storybook
- [ ] Create a code sandbox example of Content group
- [ ] Follow visual design specifications
- [ ] Provide a direct link to the deploy preview for the designer in the Pull Request description
- [ ] A comment is posted in the Visual QA issue, tagging Designer, when development is finished
- [ ] The Storybook link is added to the Design QA issue for their testing
- [ ] A comment is posted in the Prod QA issue, tagging Praveen when development is finished

PR #7233, opened 2021-09-23
fix(content-item): expose component in storybook

ariellalgilmore:
### Related Ticket(s)

#6700 

### Description

Expose `content-item` in storybook

### Changelog

**New**

- storybook, readme and codesandbox examples


kennylam (approved):
LGTM thanks @ariellalgilmore!

photodow (commented):
- [ ] Web components version isn't sitting on the grid correctly.

image

- [ ] Why is the video so much bigger than the image? Might be pre-existing issue.

Personally I feel like you could have a drop down for the CTA similar to what you did with the media, but I see Shixie called out the two options there so let's not worry about it unless Shixie sees this and agrees. 😄 

A couple small things that might have existed before, but looking good overall though. I think the only thing that would block this is the grid issue above.

photodow (approved):
Minus some small preexisting issues the exposing of the component itself looks good. Thanks!

Issue #7216, opened 2021-09-21
[Universal banner]: Spec for how Masthead may appear below banner (desktop, mobile)


annawen1:
### Overview
This problem was seen when the Cloud MH was introduced on the [v19a page and the yellow translation banner appears](https://www.ibm.com/sg-en/cloud/application-modernization).

- This behavior should be resolved in the Universal banner designs and func specs.  This FR has been linked to the Universal Banner epic (#6547) and included in sprint 21-10 for the designer working on the functional specs in that sprint to revisit the designs to be sure that this problem is addressed.  

### The problem

The current masthead doesn't seem to account for banners above and overlaps them. An issue was brought up with when the cloud-masthead is used with the locale banner: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7211

Currently the Cloud pages override the `position: fixed` style of the masthead by setting to `position: static` when banner is seen, and setting it back to `position: fixed` when banner is scrolled out of view. The Carbon THINK banner Ryan had created uses similar logic to override our masthead positioning.

<img width="1677" alt="Screen Shot 2021-09-21 at 4 04 31 PM" src="https://user-images.githubusercontent.com/54281166/134240123-e122633b-9e1c-4159-8299-9f4c2f610608.png">

### The solution

Detect when there is a banner above the masthead and correct the position to static. When banner is scrolled out of view, the masthead can be set to `fixed`. Not sure if this needs to be separated into a wrapper component or can be part of the masthead?

Ensure that the scrolling behavior of the masthead maintains the expected behavior when there is an L1 / TOC / L1 & TOC.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

proeung:
@annawen1 Thanks for opening this feature request issue! Once, these changes are implemented in the WC package, we (HC) can clean up the extra styling/logic that is added on the Drupal side.

shixiedesign:
![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/ae50610c-0e32-4c23-9563-620928c71a06)

[...]

PR #7238, opened 2021-09-23
fix(cloud-masthead): query for cloud masthead container to set top offset

annawen1:
### Related Ticket(s)

[Cloud-Masthead]: Calculate the offset top-position of the Mobile Left-Nav component when there's a local banner #7211

### Description

When testing on the Cloud Drupal test page, the query for the `dds-masthead` top offset returns `0`, use `dds-cloud-masthead-container` instead to retrieve top offset

### Changelog

**Changed**

- retrieve top offset from `dds-cloud-masthead-container`


packages/web-components/src/components/masthead/left-nav-overlay.ts
```diff
@@ -31,7 +31,8 @@ class DDSLeftNavOverlay extends LitElement {
     if (changedProperties.has('active')) {
       const doc = this.getRootNode() as Document;
       if (this.active) {
-        const masthead: HTMLElement | null = doc?.querySelector('dds-masthead');
+        // TODO: remove when masthead can account for banners above
+        const masthead: HTMLElement | null = doc?.querySelector('dds-cloud-masthead-container');
         const mastheadTopOffset = masthead?.offsetTop;
         this.style.top = `${mastheadTopOffset}px`;
```
jeffchew, line 37: Just curious, should this line go off only if the const `masthead` above exists?

packages/web-components/src/components/masthead/left-nav.ts
```diff
@@ -139,7 +139,7 @@ class DDSLeftNav extends StableSelectorMixin(BXSideNav) {
 
         // TODO: remove this logic once masthead can account for banners.
         // calculate top offset to display left-nav correctly when page has banner above masthead
-        const masthead: HTMLElement | null = doc?.querySelector('dds-masthead');
+        const masthead: HTMLElement | null = doc?.querySelector('dds-cloud-masthead-container');
         const mastheadTopOffset = masthead?.offsetTop;
         this.style.top = `${mastheadTopOffset}px`;
```
jeffchew, line 144: Same comment as above, `mastheadTopOffset` would be undefined or null I think if `dds-cloud-masthead-container` isn't on the page, right?

packages/web-components/src/components/masthead/left-nav-overlay.ts
annawen1: yeah I checked to see in the regular masthead - if `mastheadTopOffset` is null, the `top` inline styling doesn't get applied

annawen1: I could add a conditional in there so it doesn't try to set at all I suppose

packages/web-components/src/components/masthead/left-nav-overlay.ts
```diff
@@ -31,9 +31,12 @@ class DDSLeftNavOverlay extends LitElement {
     if (changedProperties.has('active')) {
       const doc = this.getRootNode() as Document;
       if (this.active) {
-        const masthead: HTMLElement | null = doc?.querySelector('dds-masthead');
+        // TODO: remove when masthead can account for banners above
+        const masthead: HTMLElement | null = doc?.querySelector('dds-cloud-masthead-container');
```
proeung, line 35: I'm assuming that we'll have a follow-up PR that accounts for the offset top position of the `dds-masthead-container` if it exists. Just wanted to bring this up in case AEM implemented a banner treatment above the masthead (eg. Think banner, locale, etc).

jeffchew (approved):
LGTM!

proeung (approved):
Looks good overall. I think we'll get the proper offsetTop value this time now that we're checking against `<dds-cloud-masthead-container>`. Thanks for taking a look into this, @annawen1!

packages/web-components/src/components/masthead/left-nav-overlay.ts
annawen1: yeah that should be addressed in this feature request: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7216

jeffchew:
Going to close this in favor of https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/7247.

PR #7247, opened 2021-09-23
fix(cloud-masthead): set masthead position to fixed on menu open

annawen1:
### Related Ticket(s)

### Description
To unblock users on the cloud pages with locale banners, set the position of the masthead to `fixed` when menu is opened.

Previous attempts to set offset of the menu haven't been successful, so this will be a temporary workaround

Sep-23-2021 17-44-34

### Changelog

**Changed**

- query for the `cloud-masthead-container` masthead and set position to `fixed` on `expanded`

**Removed**

- remove top offset calculation logic for the `left-nav-overlay`


jeffchew (approved):
LGTM!

kennylam (approved):

PR #7217, opened 2021-09-21
fix(cloud-masthead): ensure megamenu bottom graident sticks to bottom

annawen1:
### Related Ticket(s)

#7232

### Description

Cloud megamenu has a gradient at the bottom of the megamenu that user sees when the megamenu is scrollable. When user scrolls however, the gradient is stuck in the same position. Seems the megamenu needs a set `bottom` value

Screen Shot 2021-09-21 at 4 15 25 PM

FIX:
Screen Shot 2021-09-21 at 4 21 52 PM

### Changelog

**Removed**

- `bottom: auto`

PR #7253, opened 2021-09-27
fix(cloud-masthead): ensure megamenu closes when user clicks below it

annawen1:
### Related Ticket(s)

### Description

The `bottom: auto` was removed so the gradient would be positioned correctly, but it affected the megamenu's closing in this PR: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/7217

This PR reverses those changes, and sets the positioning of the gradient pseudo element to `fixed` instead.

### Changelog

**Changed**

- add `bottom: auto` back to megamenu
- change the gradient pseudo element to fixed so it stays at the bottom of the megamenu


jeffchew (approved):
LGTM!

Issue #6808, opened 2021-08-04
[Styles] Create print-friendly styles for identified components


jeffchew:
### The problem
Details are in the epic #6607 .

### The solution
Should make use of the print media query and apply print friendly styles per recommendations in #6607.

### Additional information
 - The downstream test applications would need to be updated first before this can be properly tested: #6732 
 - Can apply these styles in the styles package so that this can be applied to both React and Web Components
 - Testing can happen in the deploy previews of NextJS Test app and Web Components HTML App once all relevant components are added to those downstream applications

jeffchew:
@RobertaJHahn dev issue created and ready for planning poker.

PR #7260, opened 2021-09-27
feat(print-styles): adding print styles to components

IgnacioBecerra:
### Related Ticket(s)
#6808

### Description
Added print styles for 
- Masthead (hide)
- Footer (hide)
- Table of Contents (show as list)
- Tabs Extended + w/Media (show all tabs)
- Back to top (hide)

### Changelog

**New**

- added styles for some components to hide/show when printing


RichKummer:
`Masthead` and `Back to Top` look like they are working as intended. 

1. In Web Components, the `Footer` links still appear in the print view (though the background disappears). This also happens in the WC code sandbox. These link should be hidden like in React version.
Screen Shot 2021-10-01 at 10 28 18 AM


2. `Tabs extended media` should display content from each tab one after the other. In the Gray100 example page, the currently selected tab is displayed, but the non-selected tabs are only printed with the title in list format. Each tab should display full content as if it were selected. It looks like `Tabs extended media` is working properly in the sandbox.
Screen Shot 2021-10-01 at 10 31 19 AM

3. In the React and WC storybooks, it looks like `Tabs extended media` is displaying the second tab as open, but the print function only prints the first page. Is this something we can fix on our end?
Screen Shot 2021-10-01 at 10 32 54 AM

4. In the React storybook, the vertical `Table of contents` is overlapping the content of the page if the user has scrolled past the start of the section. The ToC should display above the content it is leading into. Also in WC storybook
Screen Shot 2021-09-30 at 6 37 32 PM

IgnacioBecerra:
@RichKummer Addressed the comments in the last commit! Seemed like the style that made the Footer disappear on print was being overridden by something else. 

As for the Tabs Extended with media, it was only showing the first tab on the Storybook pages since it was trying to print the Storybook page itself, not the page content. If you open the story in a new tab, you'll be able to see the component printing out as it should in multiple pages. As for the example in the NextJS pages, it may be that since it's a React wrapper, the styles aren't updating as it should within that repo just yet (i.e. would work once this PR gets merged and the repo gets updated to canary). I think that this must be it since it's working on the sandbox like you also saw.

RichKummer (approved):
Thanks for fixing in your last commit! I noticed that when we try to print the vertical ToC with an image above it, the image is enlarged to take up nearly the full space of the printed page. Can we lock the size of the image to reflect the digital version? Otherwise that is a greater amount of space/ink used.
Screen Shot 2021-10-05 at 9 51 39 AM

jeffchew (approved):
LGTM!

PR #7266, opened 2021-09-28
feat(feature-section): introducing react wrapper

IgnacioBecerra:
### Related Ticket(s)
#6290

### Description
Introducing the React wrapper version of Feature Section.

### Changelog

**New**

- Feature Section React wrapper


packages/react/src/components/FeatureSection/__stories__/FeatureSection.stories.js
```diff
@@ -0,0 +1,3 @@
+# Feature section
```
jeffchew, line 1: Hey @IgnacioBecerra , this is throwing prettier errors.

kennylam: Looks like you swapped the `.js` and `.mdx` content, @IgnacioBecerra.

IgnacioBecerra: Ah good eye, the script had the wrong file content when generating the react wrapper for these two files. Updated with the correct file contents.

skuruvil:
@IgnacioBecerra 
I think the image aspect ratios are to be swapped between medium and small. In medium breakpoint, it should be 4 :3, and in small 1:1

IgnacioBecerra:
@skuruvil Thanks for the review! I'm changing the aspect ratio to 4x3 for all breakpoints from medium (i.e. md -> xlg) and changing small to 1x1. Let me know if that's what you'd like or if we only have to adjust these two breakpoints (i.e. leave 1x1 for lg/xlg).

skuruvil:
@IgnacioBecerra We only have to adjust these two breakpoints (i.e. 1x1 for small / lg / xlg / max and 4x3 for only medium).

skuruvil:
@IgnacioBecerra the aspect ratios look good now. thanks.  However, I can see some minor grid issues, especially for the CTA card. The same issue I could find in the WC as well. Not sure it’s just for my browser. 

Screenshot 2021-10-04 at 11 12 44 PM

Issue #7128, opened 2021-09-13
[Masthead] Resolve a11y issues identified by storybook in web components - DDS Consulting


jeffchew:
Describe the issue: 

> What did you expect to happen? What happened instead? 

Address all accessibility violations identified in storybook:

https://www.ibm.com/standards/carbon/web-components/?path=/story/components-masthead--default

![image](https://user-images.githubusercontent.com/1641214/133140387-80513950-bc95-4078-8b7f-9d32cf864e72.png)

> Component(s) impacted? Please include screenshots, or code and relevant URLs.

Masthead (web components)

## Additional information
-  Engineering info:  
     - HC engineer: @huntermacd 
     - JC Jira Ticket: https://jsw.ibm.com/browse/HC-2202
     - DDS consulting engineer:  @kennylam

PR #7285, opened 2021-09-29
(masthead): Resolve a11y issues identified by storybook in web components

huntermacd:
### Related Ticket(s)

JIRA: https://jsw.ibm.com/browse/HC-2202
GH: #7128 

### Description

Address all accessibility violations identified in storybook.

### Changelog

**Changed**

- change masthead profile to `role='button'` rather than `listitem` or `menuitem` because it isn't within a list or a menu
- change `header` element to plain `div` in masthead to avoid nesting landmarks (`header` was nested within an element with `role='main'`)
- add aria labels to each type of `nav` element in top nav component to distinguish one from the other
- remove the `role='menubar'` attribute from top-nav because it contains no menuitems.


packages/web-components/src/components/masthead/top-nav-menu.ts
```diff
@@ -21,28 +21,12 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
...
    */
   @property({ type: Boolean, reflect: true })
   active = false;
 
-  updated(changedProperties) {
-    super.updated(changedProperties);
-    if (changedProperties.has('active')) {
-      const { active, _triggerNode: triggerNode } = this;
-      if (triggerNode) {
-        triggerNode.setAttribute('data-selected', String(Boolean(active)));
```
annawen1, line 41: The `aria-selected` attribute was already addressed here: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/7030. The attribute was changed to `data-selected` as seen in the code being removed. There's corresponding css styles for the `data-selected` attribute so that it works for both react and WC.

Unless there's an issue with `data-selected`, I think we should keep it and remove the `:host(#{$dds-prefix}-megamenu-top-nav-menu)[active] a::after` styles as it would be redundant.

packages/web-components/src/components/masthead/masthead.scss
```diff
@@ -499,3 +499,15 @@
 :host(#{$dds-prefix}-masthead-search-container) {
   display: contents;
 }
+
+:host(#{$dds-prefix}-megamenu-top-nav-menu)[active] a::after {
```
annawen1, line 503: If there is no issue with `data-selected` attribute, I think we can remove this as the `aria-selected` issue has already been addressed in a previous PR.

annawen1 (changes requested):
Looking good, just some comments on the `selected` attribute

packages/web-components/src/components/masthead/top-nav-menu.ts
huntermacd: Ah, right you are. I hadn't updated my branch from upstream in a little while, so I didn't see that this was already taken care of! Fixed in the latest commit.

annawen1 (approved):
LGTM! Thanks @huntermacd!

jeffchew:
Thank you @huntermacd ! Can you update the web components snapshot files?

Issue #6702, opened 2021-07-21
[Content block] expose this "sub-components" as simple components in storybook


shixiedesign:

jeffchew:
@RobertaJHahn this should be added to the next engineeering refinement for sizing.

PR #7294, opened 2021-09-30
fix(content-block): expose story

ariellalgilmore:
### Related Ticket(s)

#6702 

### Description

Expose `content-block` component in storybook

### Changelog

**New**

- story, codesandbox, readme


photodow (commented):
So we have some major bugs to fix on this one. Everything from spacing, to type, to dependencies are an issue, but that's one of the reasons we're exposing all of these components to better support them and QA them.

As far as the exposing this one goes I have 2 asks...

1. Either remove the CTA card link in content group simple, or change it to a text link. That way we don't have two card links butting up against each other. Might be best to just remove it for now. 

2. Can we either get an additional story, or even a just toggle switch to demo its aside?

photodow (commented):
- [x] The heading in web components is not updating when I edit it.

- [ ] And I'm guessing this issue is unrelated, but maybe we should pick another component to demo because when the aside is added with the content group with cards the whole layout shifts incorrectly. Unless maybe we can disabled aside if content group with cards is there?

photodow (approved):
I'll state it again, we got some major things to fix here. I'll approve it, but I also noticed the border bottom doesn't work with the aside in react. Is that a known problem?

Issue #6761, opened 2021-07-29
[Locale Modal] Regions are not disabled when there are no locales configured


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

When there is a limited number of locales/languages set up, it doesn't disable the regions when there are no countries/anguages defined:

![image](https://user-images.githubusercontent.com/1641214/127556814-68a45d22-e078-4464-bd69-38f38307ec5b.png)

![image](https://user-images.githubusercontent.com/1641214/127556841-362d3c39-a7e0-4878-b432-b7acaabf6f60.png)

> Is this issue related to a specific component?

Locale Modal

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Should disable the regions if there are no locales/languages available

## Steps to reproduce the issue

1. Example deploy preview of web components template: https://web-components-template.s3-web.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/6/en-us/index.html
2. Open the locale modal

PR #7302, opened 2021-09-30
fix(locale-modal-composite): set region item as invalid if locale not configured

emyarod:
### Related Ticket(s)

#6761

### Description

This PR marks region items as invalid in the locale modal if a locale is not configured in the given region

image

~~The storybook preview locales have been reviewed for testing but will be restored prior to merging~~

EU locales have been commented out for future testing


jeffchew (approved):
@emyarod looks good to me!

I'm wondering if it's worth keeping one region (or two) disabled moving forward, so we can visually (and functionally) test that that feature is working fine? Perhaps we can omit the European region. It's probably good to be able to test non-Latin languages like Arabic, Japanese and Korean.

image

emyarod:
sounds good, updated to remove EU locales in the storybook!

Issue #7168, opened 2021-09-16
[Services] Change the Locale API and the Translation Service API to support a language only translation feature


RobertaJHahn:
### The problem
- The `cloud.ibm.com/docs` application is currently loading just the language code within its `?locale= parameter` (eg. https://cloud.ibm.com/docs?locale=es) where other content ecosystems like Drupal is using both `lc` and `cc`.
- This is a feature that has not yet been accounted for in the Masthead component data fetching script.

### The solution
- Change the Locale API to fetch the Locale or LC code if only LC is provided.
- Change the Translation Service API to pull the correct translation file if only LC is provided.

### Additional information
- Translation API:  https://ibmdotcom-services.mybluemix.net/TranslationAPI.html
- Locale API:  https://ibmdotcom-services.mybluemix.net/LocaleAPI.html
- Reach out to Putra and Mark Kulube when the code is ready to be tested.  They have to test this new function with the IBM Docs team.

PR #7306, opened 2021-10-01
feat(services): translation and locale to handle `lc` only locales

annawen1:
### Related Ticket(s)

[Services] Change the Locale API and the Translation Service API to support a language only translation feature #7168

### Description

For adopters that only use `lc` for translations instead of `lc-cc` combo, this PR adjusts the locale and translation service to accommodate `lc` only locales. 

In the `Locale` service:
1. the DDO is first checked if the `language` field contains the `-` character (this is assuming that the language field either contains `lc-cc` or `lc`).
2. if not, the `lc` is returned for the `Translation` service to use.
3. if the DDO does not exist, check the `lang` html attribute for the `-` character.

In the `Translation` service:
1. form the session storage key with either `lc` or both `lc` and `cc` from the `Locale` service
2. make the translation api call formed with  either `lc` or `lccc` depending on object from `Locale` service

Can test by adding the `&lc=${lc}` (ie. `&lc=es`) parameter to the end of the cloud-masthead deploy preview: https://ibmdotcom-web-components-experimental.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/7306/ifra[…]ud-masthead--default&lc=es

**Note**
The above testing will only work for the cloud-masthead because there are `lc` only cloud translation files set up. We will need to set `lc` only translation & countrylist files to support the global masthead.

### Changelog

**Changed**

- Locale service: if the `language` field in the `DDO` only provides the `lc`, pass only the `lc` to translation service
- Translation service: call the api with only the `lc` in the endpoint
- Updated Locale API unit tests
- SearchTypeahead API call to not include `cc` parameter in endpoint if there is no `cc` provided in DDO or `lang` html element
- `preview-html` storybook files to support `lc` only DDO language and `lang` html attribute


jeffchew (approved):
This one was a doozy to review, but looks good to me!

annawen1:
> Code looks good overall!
> 
> QQ: Is there a way to test out the fetching of the translation using this WC test app with the Cloud Masthead (see [link](https://ibmdotcom-web-components-html-test-upstream.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/7306/cloud/index.html))? If not, we'll need to merge these changes down and have @markkulube test it out within the Docs local application.

I've edited the storybook `preview-html` files to allow for `lc` only setting. You can view it working by adding `&lc=es` query at the end of the cloud deploy preview url like so: https://ibmdotcom-web-components-experimental.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/7306/ifra[…]ud-masthead--default&lc=es

This will set the Digital Data Object and `lang` html attribute in storybook to be `lc` only

proeung:
@annawen1 Thanks, @annawen1! You're awesome! I think we are good to mark this PR with the "Ready to merge" label unless @markkulube has anything else that he'd like to add.

Issue #7061, opened 2021-09-02
[Cloud Masthead]:  setting `has-contact` boolean to false doesn't remove the "Contact us" CTA or set to display none - DDs Consulting


proeung:
### The problem

- Our adopters (Cloud Docs) are having trouble using the `has-contact` prop to set remove the "Contact us" CTA from the DOM.
- Users with no contact widget should be able to adopt the `<dds-cloud-masthead-container>` and sets `has-contact=false` to conditionally remove the CTA.

```
<dds-cloud-masthead-container
  platform="Cloud"
  auth-method="api"
  data-endpoint="/common/carbon-for-ibm-dotcom/translations/cloud-masthead"
></dds-cloud-masthead-container>
```

### Example
![image (68)](https://user-images.githubusercontent.com/1815714/131920431-2221242b-85cf-41cc-8f74-bcb20fbf7e2a.png)

----
HC engineer: Putra Bonaccorsi
JC Jira Ticket: https://jsw.ibm.com/browse/HC-2179
DDS consulting engineer: @annawen1

PR #7322, opened 2021-10-05
fix(cloud-masthead): adjust has-contact attribute to accept string

proeung:
### Related Ticket(s)

GH Issue: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7061
HC JIRA Ticket: https://jsw.ibm.com/browse/HC-2179

### Description

- This is a follow-up PR for an issue that was closed, however, still an issue for adopters using the `has-contact="false"` to optionally hide the "Contact us" button.
- This PR sets the `has-contact` prop to accept a string.

### Testing Instruction

- Pull down the changes of this PR in a test preview and sets the following markup with the `has-contact` attribute sets to `false`.
- Ensure that the "Contact us" button is removed when the attribute is set to `false`. If the `has-contact` is not defined or set to `true`, the "Contact us" button should be set visible.

```
<dds-cloud-masthead-container
  platform="Cloud"
  auth-method="API"
  has-contact="false"
  data-endpoint="/common/carbon-for-ibm-dotcom/translations/cloud-masthead"
></dds-cloud-masthead-container>
```

### Expected Display

Screen Shot 2021-09-09 at 3 20 57 PM

### DIFF
--- a/packages/web-components/src/components/masthead/cloud/cloud-masthead-composite.ts
+++ b/packages/web-components/src/components/masthead/cloud/cloud-masthead-composite.ts
@@ -64,8 +64,8 @@ class DDSCloudMastheadComposite extends DDSMastheadComposite {
   /**
    * `true` if Contact us should be shown.
    */
-  @property({ type: Boolean, reflect: true, attribute: 'has-contact' })
-  hasContact = true;
+  @property({ type: String, reflect: true, attribute: 'has-contact' })
+  hasContact = 'true';
 
   /**
    * The profile items for unauthenticated state.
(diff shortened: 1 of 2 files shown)


packages/web-components/src/components/masthead/cloud/cloud-masthead-composite.ts
```diff
@@ -282,6 +282,8 @@ class DDSCloudMastheadComposite extends DDSMastheadComposite {
         platformAltUrl = platformUrl[formattedLang].url || platformUrl;
       }
     }
+    console.log(hasContact);
```
annawen1, line 285: console.log left behind!

annawen1 (commented):
some small things, but looks good!

Issue #7213, opened 2021-09-21
[Button group] Remove extra padding and restore spacing


shixiedesign:
Currently Button group contains an extra padding on the top. This is not expected and is causing spacing styling to be less straight forward. We need to systematically remove that extra padding and restore the styles of all components using Button group.

![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/2241fc95-6855-4f99-b2c0-e50030ab1f46)

Please note, removing the padding should **NOT** affect: 
1. on mobile, buttons stack and 16px apart 
![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/2f612db4-22d7-488f-9bfe-ff3f022e56c3)
2.  all the component currently using Button group needs to have their styles maintained... this might be a lot of work. Please check Percy for unexpected diffs.

PR #7330, opened 2021-10-06
fix(button-group): refactor button group spacing rules

jkaeser:
### Related Ticket(s)

Resolves https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7213

### Description

`<dds-button-group-item>` elements previously had top margins in order to create vertical spacing when they stack at mobile viewport widths. This created problems when integrating them into other components, however, so now the buttons only create internal spacing.

### Changelog

**Changed**

- refactor `<dds-button-group`> and `<dds-button-group-item`> components to only apply internal spacing
- add spacing rules where appropriate to prevent regressions


jeffchew (approved):
LGTM!

Issue #5990, opened 2021-05-06
[Content item horizontal] Storybook grid issue


ariellalgilmore:
## Detailed description

> Describe in detail the issue you're having.

Content item Horizontal default does not align to grid

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

bug

> Is this issue related to a specific component?

content item horizontal default

> What did you expect to happen? What happened instead? What would you like to
> see changed?

not aligned to grid look at specs

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

- Screenshots or code
- Notes

shixiedesign:
Still an issue. I just updated the description with latest.

andy-blum:
@andysherman2121 there are some alignment issues remaining in #7264 that weren't addressed in #7318. Due to the heavily inter-connected nature of `content-item-horizontal` and the thumbnail variation, I recommend those fixes be done as a part of this issue.

PR #7361, opened 2021-10-08
fix(content-item-horizontal): fixed grid issues

andysherman2121:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5990

### Description

Content item horizontal wasn't aligning with the grid for the default and with thumbnail variation. This PR adds CSS grid to fix the issues.


andysherman2121:
> The _with Media_ and _with Thumbnail_ versions look great. For default, I'm seeing about 4px between the body copy and the grid: Screen Shot 2021-10-08 at 3 15 46 PM
> 
> `sm` and `md` align properly, it's just off by a small amount on the larger screensizes.

@RichKummer just fixed! thanks for the review Rich. Let me know if it looks good now

RichKummer:
> @andysherman2121 @RichKummer checking the diffs, it is stacking in the dotcom shell even at 1280px. Is this correct?
> 
> image

It shouldn't be stacking until `sm` for the default Content item horizontal - I'm seeing the same thing.

andysherman2121:
@RichKummer just added the fix! React breakpoints should be good now

jeffchew:
@andysherman2121 looks like your snapshots are out of date.

jeffchew:
@shixiedesign I noticed the same thing, I think the branch was out of date and was diffing an old version of the branch to `master`. I did an update so the deploy previews and percy runs should kick off again.

cc: @andysherman2121

shixiedesign:
@andysherman2121 still seeing extra padding-right on copy of Content item horizontal -- thumbnail, but that's all! 

**Current:**
image
**Spec**
image

Issue #7369, opened 2021-10-12
Web component & React: Table of contents - horizontal - TOC vertical and horizontal navigation menus are displayed for the user.  


Praveen-sr:
### Description

Issue:
On Table of contents - horizontal,  both TOC vertical and horizontal navigation menus are displayed for the user.  

Expected:
Only horizontal TOC navigation menus must be displayed.

### Component(s) impacted

![image](https://user-images.githubusercontent.com/60002341/136963619-9456b9f4-36f1-4160-86eb-a8934114e5d2.png)

![image](https://user-images.githubusercontent.com/60002341/136963637-1f6d8c28-b258-4ef0-9102-98c15075908b.png)

### Browser

Chrome, Safari, Firefox

### Carbon for IBM.com version

1.26

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

Web component & react

### Package

@carbon/ibmdotcom-react, @carbon/ibmdotcom-web-components

### CodeSandbox example

NA

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

Praveen-sr:
@RobertaJHahn  Re tested the above issue on Web component & React, now working as expected.

![image](https://user-images.githubusercontent.com/60002341/137130245-a407afc8-0bcf-408e-9c54-f7bb4f7c43a2.png)

PR #7370, opened 2021-10-12
fix(table-of-contents): add default TOC styles

kennylam:
### Related Ticket(s)

#7369

### Description

Currently `table-of-contents` print styles are visible at all times. This fixes to show only when printing.

### Changelog

**New**

- default print styles to hidden when not printing

image


kennylam:
cc: @Praveen-sr

Issue #7375, opened 2021-10-13
React: Feature card-large - On hover, user is not able to see the content properly


Praveen-sr:
### Description

Issue: 
Feature card-large - On hover, user is not able to see the content properly

On both white theme and g90, g100

Note: Working fine on Web component version.

### Component(s) impacted

![image](https://user-images.githubusercontent.com/60002341/137134560-6e4938ee-7670-4ef4-ac4b-028f2bf77ad5.png)

### Browser

Chrome, Safari, Firefox

### Carbon for IBM.com version

1.26

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

React

### Package

@carbon/ibmdotcom-react

### CodeSandbox example

NA

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #7377, opened 2021-10-13
fix(feature-card): fix hover styles for large variation

kennylam:
### Related Ticket(s)

#7375 

### Description

Fixes hover styles for React `FeatureCard - large`.

### Changelog

**Changed**

- fix hover styles for `FeatureCard - large`


kennylam:
> Separate thing, I noticed the footer content changed. I thought we were using mock data for there, maybe we are actually pulling live footer content?

@jeffchew It looks like some of the stories are using mock data while others aren't. I'll create an issue to get them consistent.

Issue #6574, opened 2021-07-07
React: Dotcom shell - Callout quote, above the link no spacing 


Praveen-sr:
## Detailed description

Issue:
Callout quote, no spacing between line and the link
 
Note: Working fine on the actual component.

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/3a4a6011-8ccf-4411-820b-f26c82b1dec9)

guilhermelMoraes:
@Praveen-sr tested this issue locally and on canary for web components and react, and couldn't reproduce. Would you please make sure that this is still a thing?

PR #7387, opened 2021-10-13
fix(toc): removing wrong spacing for horizontal rule

guilhermelMoraes:
### Related Ticket(s)

[Callout quote] On DotcomShell story, above the link no spacing #6574

### Description

This margin was interfering with all `hr` elements inside the TOC. Plus I didn't find anything related to this being necessary for any of the visual specs.


oliviaflory:
Hey @guilhermelMoraes the spacing on the Callout quote looks good.

I do see a change in the horizontal rule padding under the lead space block on the Dotcom shell stores that seems pretty significant. Was that margin top also controlling the padding there? Should we open a separate issue?
Screen Shot 2021-10-14 at 11 46 42 AM

Issue #6229, opened 2021-06-02
Web component: Card Static - Clicking next the link, CTA link is getting focus


Praveen-sr:
## Detailed description

Issue:
 Clicking next the link, CTA link is getting focus

## Additional information

![image.png](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/06caf969-6525-44d5-a7e8-148b0beab031)

proeung:
@Praveen-sr I just checked PROD and [Canary](https://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-card--static) Storybook preview and I'm not able to replicate the focus issue that you're seeing. Can you take a look again?

I believe the issue mentioned here might have been fixed in the [PR](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/6215). @RobertaJHahn @jeffchew I'm going to move this issue to the "Done" pipeline, please review and close out this issue if applicable.

Issue #6255, opened 2021-06-03
[Card] Eyebrow text container issues - no DDS consultant


DragosRistici:
###  Engineering info:
- HC engineer: @andy-blum 
- HC Jira Ticket: https://jsw.ibm.com/browse/HC-2251
- DDS consulting engineer: N/A

-----------

#### Overview
- All card have this issue.

1. Web Components  Card – Static Eyebrow container does not have the width rule of card container - 32px implemented. Also, the same is for the text CTA. This also provokes the error in #6229. 
<img width="515" alt="Screenshot 2021-06-03 at 16 25 36" src="https://user-images.githubusercontent.com/48327647/120652772-bff47f00-c488-11eb-9672-82a83c8a6f0b.png">

2. The Tag group container has an extra 8px on bottom that adds to the 16px before the card paragraph: 

What it is now:
<img width="878" alt="Screenshot 2021-06-03 at 16 37 48" src="https://user-images.githubusercontent.com/48327647/120654123-0c8c8a00-c48a-11eb-9d30-d53f4f5eecb8.png">
What it is expected:
<img width="821" alt="Screenshot 2021-06-03 at 16 33 48" src="https://user-images.githubusercontent.com/48327647/120653575-7ce6db80-c489-11eb-9cc9-3d080c1349e3.png">

PR #7400, opened 2021-10-15
fix(card): update card styles to match design specs

andy-blum:
### Related Ticket(s)

Fixes: #6229 #6255 

### Description

- Updates styles of card component to match specs
- Modernizes spacing techniques of card copy wrapper
- Reduces width of CTA link host element to prevent focus-on-adjacent-click

### Changelog

**Changed**

- Updated card styles
- Updated tag-group styles
- Fixed bug where clicking next to a card link


andy-blum:
@oliviaflory The side effects percy found should be addressed now.

Issue #6558, opened 2021-07-02
[Styles]: revert to using keywords in `carbon--breakpoint-between` mixin


kennylam:
Due to a breaking change in `@carbon/layout`, all values used in the `carbon--breakpoint-between` mixin must be numerical. Using keywords will break certain application builds. A temporary fix was put in place to bypass this issue.

A [PR](https://github.com/carbon-design-system/carbon/pull/9102) has been created in the Carbon repository addressing this issue. Once available, the changes in #6557 should be reverted.

RobertaJHahn:
Scheduled in 8/30 engineering refinement meeting.  Priority needed.

RobertaJHahn:
Carbon has berged the fix.  Need to revert the hard coded values.

PR #7418, opened 2021-10-18
fix(styles): revert - temporarily hard-code breakpoint values (#6557)

emyarod:
This reverts commit 6615d5e86cdda0c6311fd2afc42e622e420dac78.

### Related Ticket(s)

#6558

### Description

Since the upstream fix is now available we can revert #6557 as described in the original PR

### Changelog

**Changed**

- replace hard coded breakpoint values with keywords


jeffchew:
Thanks @emyarod , changing the subject so that our changelog script can pick up for our logs.

Issue #6820, opened 2021-08-05
[Leadspace with search] Add the sticky search scroll functionality


IgnacioBecerra:
### The problem
Currently the Leadspace with search doesn't have the scroll functionality in which the search bar sticks at the top upon scroll down.

### The solution
Implement the feature.

### Additional information
 - Some work has already been done when having worked on the original Leadspace with search component, just need to finish the implementation as it was too complex to include originally
 - [Leadspace functional specs](https://ibm.ent.box.com/file/814160426264?s=r2wmgutqhybbmey96zv8rvjjufsw6ib4)

RobertaJHahn:
Confirming whether the sizing of 3 is OK with Ignacio.  Once sized, the issue can be moved to Pending Scheduling pipeline.

PR #7421, opened 2021-10-19
feat(leadspace-with-search): added sticky scroll functionality

IgnacioBecerra:
### Related Ticket(s)
#6820 

### Description
Added the Leadspace with search scroll functionality where the search container is fixed on top of the page.

This feature must be tested in the `Dotcom Shell` story `With leadspace with search`.

### Changelog

**New**

- `leadspace-with-search` now has an optional behavior to allow the component to remain stickied at the top of the page upon scrolling down to the content


IgnacioBecerra:
@emyarod Thanks for the review! I had tried using `translateY` first but it resulted in some weird behavior. Some previous sticky/scroll logic has used `top` before, but I think this could be revisited in the future so any sticky logic can be consolidated in one place.

Issue #5434, opened 2021-03-10
[Quote] text rendering - no DDS consultant


oliviaflory:
###  Engineering info:
- HC engineer: @jkaeser 
- HC Jira Ticket: https://jsw.ibm.com/browse/HC-2253
- DDS consulting engineer: N/A

-----------

## Detailed description

> Describe in detail the issue you're having.

The quote text is rendering thinner than expected within the Quote and Call out Quote component on Chrome, and heavier than expected on Safari.

The font weight is supposed to be light, but there may be a font smoothing added to the style that is making it even lighter in Chrome?

| Browser | Example | 
| --- | --- | 
| Chrome | ![Screen Shot 2021-03-10 at 1.34.00 PM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/32eef436-560e-4d33-a998-1e9a359008c4)  | 
| Safari | ![Screen Shot 2021-03-10 at 1.50.48 PM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/bdc89831-804c-4603-a281-7e9d4f376f06)   |  
| Expectation | ![Screen Shot 2021-03-10 at 1.33.07 PM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/80357a7e-0606-4c70-a642-999503f5eb40) | 

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

bug or general, maybe an investigation question

> Is this issue related to a specific component?

[Quote](https://ibmdotcom-web-components.mybluemix.net/?path=/story/components-quote--default)
[Call out quote](https://ibmdotcom-web-components.mybluemix.net/?path=/story/components-callout-quote--default)

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Font lighter weight than expected

> What browser are you working in?

Chrome Version 89.0.4389.82 (Official Build) (x86_64)
Safari Version 14.0.3 (16610.4.3.1.4)

OS Big Sur 11.2.1

> What version of Carbon for IBM.com are you using?

Latest Web components

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

- Screenshots or code
- Notes

shixiedesign:
## Question
- is Carbon setting font smoothing? 
[...]

PR #7437, opened 2021-10-19
fix(quote): Load Plex Serif font into Storybook previews

jkaeser:
### Related Ticket(s)

Resolves https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5434

### Description

Previous to this PR, the serif versions of the Plex font were not loaded in Storybook component `<iframe>`s. This forced any component that was supposed to use them, including quote components, to use fallback fonts instead.

Now that we're loading the correct font, all quote fonts should appear identically in all browsers. I personally tested the latest versions of Chrome, Firefox, and Safari on macOS 11.4 (Big Sur).

### Screenshots

**Quote**
- Before: 
Screen Shot 2021-10-19 at 10 01 02 AM

- After: 
Screen Shot 2021-10-19 at 10 00 53 AM

**Callout Quote**
- Before:   
Screen Shot 2021-10-19 at 10 01 42 AM

- After:
Screen Shot 2021-10-19 at 9 59 27 AM

### Changelog

**New**

- IBM Plex Serif fonts are loaded in Storybook component iframes.

**Changed**

- Plex fonts loaded in Storybook component iframes are pulled from the `@carbon/type` package instead of `carbon-components`.


andy-blum:
@jkaeser Did the serif font load from the CDN when we tested in codepen yesterday? If not @jeffchew should the CDN package be modified to include the serif font for the quote component?

PR #7476, opened 2021-10-21
chore(telemetry): remove carbon telemetry package

jeffchew:
### Related Ticket(s)

no related issue

### Description

A security issue was brought up in the Carbon Telemetry package (see https://github.com/carbon-design-system/carbon-charts/issues/1193). Rather than patching, the decision is to remove the telemetry package entirely as it is currently not providing any value at this time.

### Changelog

**Removed**

- `@carbon/telemetry`


jeffchew:
> Hmm... @jeffchew why remove telemetry?
> 
> Seems like this is just a vulnerable dependency that needs to be updated.
> 
> Happens all the time with most repos. Why not allow the team some time to push a patch?

Our statistics for some packages also look incorrect, which was brought up awhile ago but hasn't been addressed yet. We currently do not trust the reporting so going to remove for now.

jeffchew:
> @jeffchew statistics being incorrect... can you let us know what appears to be incorrect here so we can investigate? https://github.ibm.com/carbon/ibm/issues
> 
> If we address this dependency vulnerability would you be open to accepting a `@carbon/telemetry` patch instead of removing telemetry?

Sure, I will open an issue there.

jeffchew:
Closing for now, giving Carbon team time to address the underlying issues.

Issue #7425, opened 2021-10-19
[Test Scenario]: Footer-Short -language-only


Praveen-sr:
### Component and test scenario title

Footer-Short -language-only

### Test scenario steps

a. Check IBM logo and language_selector (dropdown) is available and clickable
b. Click on language_selector, drop down opens and user must be able to select language/geo
Note: Footer navigation links should not be present

PR #7527, opened 2021-10-27
test(footer): add e2e tests for footer short language only

emyarod:
### Related Ticket(s)

#7425

### Description

This PR adds e2e tests on the short language only footer variant for the following scenarios:

a. Check IBM logo and language_selector (dropdown) is available and clickable
b. Click on language_selector, drop down opens and user must be able to select language/geo
Note: Footer navigation links should not be present


packages/react/tests/e2e-storybook/cypress/integration/Footer/Footer.e2e.js
```diff
@@ -0,0 +1,98 @@
...
+
+    const languageSelector = cy.get(
+      '[data-autoid="dds--language-selector__select"]'
+    );
+    languageSelector.select('Arabic / عربية');
+    languageSelector.should('have.value', 'ar');
+
+    cy.screenshot();
+    cy.percySnapshot(
+      'dds-footer | Short language only (desktop language selector)',
```
annawen1, line 92: Small thing here with the snapshot name, let's change it to `Footer` instead of `dds-footer` for the React tests.

emyarod: updated the names!

Issue #6511, opened 2021-06-30
Web component: Content item horizontal with Media - Visual order and tab order are not matching.


Praveen-sr:
## Detailed description

Issue:
Visual order and tab order are not matching.
First focus is going to the link and then its come to the video player.

Expected:
Here focus first should go to the video player then to the Learn more and Microservices and containers link.

proeung:
@Praveen-sr Can you provide a video screenshot of the tabbing issue you're seeing?

Praveen-sr:
@proeung 
![CiH_Med2.gif](https://images.zenhubusercontent.com/5e21a7b470118a498429a20e/17725b75-17aa-4b44-9e18-73cac2279872)

PR #7528, opened 2021-10-27
fix(content-item-horizontal): use accessible focus order

jkaeser:
### Related Ticket(s)

Resolves https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6511

### Description

The content-item-horizontal-media component allows adopters to configure which columns the media and text content appear in at desktop viewport widths. This reordering was done with CSS, which created opportunities for the tabbing order to go out of logical sequence. Additionally, the design stipulates that the media content should always appear below the text content at mobile viewport widths.

To fix this, I added conditions to the render method to ensure the markup is printed in the correct logical order. I also added an event listener that triggers a re-render when the screen is resized across the `lg` Carbon breakpoint. This caused some render issues with the footer content which have been (temporarily?) resolved by using [the `shouldUpdate` lifecycle method](https://lit-element.polymer-project.org/guide/lifecycle#shouldupdate).

### Changelog

**Changed**

- Component markup is printed in an order that matches the logical tabbing order.
- Updated CSS to prevent visual regressions.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -3,7 +3,7 @@ description:
   Something isn't working as expected? Here is the right place to report.
 title: '[YOUR TITLE]: Brief description'
 assignees:
-  - jeffchew,ljcarot,shixiedesign,RobertaJHahn,photodow,andysherman2121,annawen1
+  - jeffchew,ljcarot,shixiedesign,RobertaJHahn,RichKummer,kennylam,BrunnoM7
 labels: ['bug']
 body:
   - type: markdown
--- a/.github/ISSUE_TEMPLATE/contribution_request.yaml
+++ b/.github/ISSUE_TEMPLATE/contribution_request.yaml
@@ -3,7 +3,7 @@ description:
   Contribute things large and small — code, design, ideas, and guidance.
 title: '[YOUR TITLE]: Brief description'
 assignees:
-  - jeffchew,ljcarot,shixiedesign,RobertaJHahn,photodow,andysherman2121,annawen1
+  - jeffchew,ljcarot,shixiedesign,RobertaJHahn,RichKummer,kennylam,BrunnoM7
 labels: ['contribution']
 body:
   - type: markdown
--- a/.github/ISSUE_TEMPLATE/feature_request.yaml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yaml
@@ -3,7 +3,7 @@ description:
   Suggest a new idea for the project.
 title: '[YOUR TITLE]: Brief description'
 assignees:
-  - jeffchew,ljcarot,shixiedesign,RobertaJHahn,photodow,andysherman2121,annawen1
+  - jeffchew,ljcarot,shixiedesign,RobertaJHahn,RichKummer,kennylam,BrunnoM7
 labels: ['Feature request']
 body:
   - type: markdown
--- a/.github/ISSUE_TEMPLATE/question.yaml
+++ b/.github/ISSUE_TEMPLATE/question.yaml
@@ -3,7 +3,7 @@ description:
   Usage question or discussion about Carbon for IBM.com.
 title: '[YOUR TITLE]: Brief description'
 assignees:
-  - jeffchew,ljcarot,shixiedesign,RobertaJHahn,photodow,andysherman2121,annawen1
+  - jeffchew,ljcarot,shixiedesign,RobertaJHahn,RichKummer,kennylam,BrunnoM7
 labels: ['question']
 body:
   - type: markdown


packages/web-components/src/components/content-item-horizontal/content-item-horizontal-media.ts
```diff
@@ -18,30 +20,86 @@ import { MEDIA_ALIGN } from './defs';
 const { prefix } = settings;
 const { stablePrefix: ddsPrefix } = ddsSettings;
 
+// The "lg" Carbon breakpoint.
+// @TODO See if this can be pulled from a Carbon package.
+const breakpoint: number = 1056;
```
jkaeser, line 25: Is there a package I can pull this number from?

ariellalgilmore: you can see how the breakpoints are being included here: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/02ab728998f8e898300d9f15dc17e5b703a093eb/packages/web-components/src/components/card-group/card-group.ts#L15

ariellalgilmore:
> @jkaeser and I worked around the render issue mentioned in the description by adding `shouldUpdate` life cycle method. Wanted to get some more engineering opinions on this. This work around fixes the issue we were having but it ensures that if there ever was a `footer` than it would never be hidden in the future, if they were to try and hide it for some reason cc @ariellalgilmore @IgnacioBecerra @annawen1

checked the functional specs https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/wiki/Content-item-horizontal#5-link-list and the `footer` or cta portion should be optional. Maybe we could do something like Content Block does here: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/7a302582c1cba065a57341e37ec65c42e5fa5c7d/packages/web-components/src/components/content-block/content-block.ts#L141-L152

jkaeser:
> checked the functional specs https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/wiki/Content-item-horizontal#5-link-list and the `footer` or cta portion should be optional. Maybe we could do something like Content Block does here:
> 
> https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/7a302582c1cba065a57341e37ec65c42e5fa5c7d/packages/web-components/src/components/content-block/content-block.ts#L141-L152

@ariellalgilmore Thanks for responding! This component extends `content-item`, which does the same thing: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/7a302582c1cba065a57341e37ec65c42e5fa5c7d/packages/web-components/src/components/content-item/content-item.ts#L73-L83

The issue we're observing is that the `hasFooter` check resolves to "false" when the new `window:resize` listener triggers an update when sizing the viewport down, even if there is footer content. As a result that footer content becomes hidden.

This is the order of events given a `content-item-horizontal` component that has slotted footer content:
1. The resize listener changes the value of the `_isOneColumn` property, triggering a component update.
2. This triggers a slot change event on the footer slot, and `_handleSlotChange` fires. This is the point of failure where the footer content appears to be empty, and `_hasFooter` is set to `false`, triggering a second update.
3. The component renders with the `hidden` bound attribute set in the markup, hiding the footer content.

I imagine there's something clever we can do with LitElement's lifecycle methods to reduce the number of updates and avoid this issue. Does anyone have any ideas?

jkaeser:
Here's a recording of what I'm seeing:

https://user-images.githubusercontent.com/12532727/139739527-0e7a99b8-7ec1-4f7c-86dc-4208dd7a88e5.mov

jkaeser:
Notably, I don't see that issue on Firefox 91.0.2.

shixiedesign (approved):
See comment above

shixiedesign:
The Percy Web component test diffs could be false positives. But please check, there are a few failures on Percy React e2e test affecting masthead: 

Nov-03-2021 10-12-47
Nov-03-2021 10-13-08

jeffchew:
> The Percy Web component test diffs could be false positives. But please check, there are a few failures on Percy React e2e test affecting masthead:
> 
> 1. Overflow scrolling caret disappeared
> 
> Nov-03-2021 10-12-47
> 
> 1. Search typeahead highlighting
> 
> Nov-03-2021 10-13-08

Good catches, I think this PR is a bit behind, we did a bunch of e2e test stabilization in the last week. I'll update the branch.

proeung:
**Update:** It looks like the accessibility order issue has been fixed and implemented correctly, however, the changes in this PR introduced a regression bug that causes the footer to disappear when the screen sizes are reduced from medium to small breakpoints. @andysherman2121 is looking for a solution that would address this regression concern. He'll be pushing up commits so that we can get this PR/issue https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6511 to the finishing line.

andysherman2121 (approved):
Nice work! bug looks fixed now

RichKummer (approved):

PR #7531, opened 2021-10-28
fix(search-with-typeahead): remove gap between search action buttons

emyarod:
### Description

the search with typeahead has a gap between the search button and clear input button when the input field is populated

React

image

Web components

image

This PR resolves the issue by turning the action button container into a flexbox

image

This should be visible on the standalone search w/typeahead, masthead with search, as well as the dotcom shell with search

### Changelog

**Changed**

- search action button container display property


RichKummer (approved):
Went through all the breakpoints and the gap isn't appearing. TY!

emyarod:
I have read the CLA Document and I hereby sign the CLA

Issue #7295, opened 2021-09-30
[Card component] React Wrapper:  React Wrapper: Update the storybook and documentation with the web component changes


RobertaJHahn:
#### Overview
The web component version of this component has changed.  A react wrapper version exists.  Review Storybook and any documentation for any required updates to align both components.

#### Additional Information
- **See the Epic (#7281) for the Design and Functional specs information**
- [React Wrapper Storybook link](https://www.ibm.com/standards/carbon/web-components/react/)
- [Web Components Storybook link](https://www.ibm.com/standards/carbon/web-components/)
- **Final Designs Box folder link:  https://ibm.ent.box.com/folder/95514001950?s=hgvu48r1fv2ga2h0yn3qshrtxjc71vch**
- Web Component dev issue (#7293 )
- React Visual QA testing issue (#7299 )
- React Prod QA testing issue (#7301 )

#### Acceptance Criteria
- [ ] Use the [Visual QA checklist](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/wiki/Definition-of-done-(Visual-QA-checklist)) to verify design quality
- [ ] If a design is provided, the Designer is included as a Reviewer in the Pull Request
- [ ] Provide a direct link to the deploy preview for the designer in the Pull Request description
- [ ] A comment is posted in the Visual QA issue, tagging Designer, when development is finished
- [ ] The Storybook link is added to the Visual QA issue for their testing
- [ ] A comment is posted in the Prod QA issue, tagging Praveen when development is finished

RobertaJHahn:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Carbon-for-IBMcom-5d449f3642eb1962336cbe52/issues/carbon-design-system/carbon-for-ibm-dotcom/7295?planning-poker) with ZenHub @andysherman2121 @annawen1 @emyarod

RobertaJHahn:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Carbon-for-IBMcom-5d449f3642eb1962336cbe52/issues/carbon-design-system/carbon-for-ibm-dotcom/7295?planning-poker) with ZenHub @jeffchew

RobertaJHahn:
[...]

PR #7538, opened 2021-10-28
feat(card): create react wrapper story and codesandbox

ariellalgilmore:
### Related Ticket(s)

#7295 

### Description

Created react wrapper story, README, and codesandbox example

### Changelog

**New**

- created react wrapper story for Card including static and pictogram
- created codesandbox example
- added back `important!` to margin styles, so they don't get overwritten by carbon's tag


jeffchew:
@ariellalgilmore prop tables aren't showing up, can you add for Card and its sub-components?

image

jeffchew (approved):
Going to let this through for now, looks like an upgrade to Storybook 6 might fix this based on what @ariellalgilmore mentioned, which we will try to do later this year.

Issue #7546, opened 2021-10-29
[Lit]: Include Lit as a core dependency for web components


jeffchew:
### The problem

Related issue for `carbon-web-components`:
https://github.com/carbon-design-system/carbon-web-components/issues/723

End users currently need to install lit-element and lit-html manually as part of their application setup. Many other web component design system teams includes as part of their projects. In addition, upgrading Lit (especially major versions) currently would require a major version release to handle.

### The solution

Include lit-element and lit-html as core dependencies to the project.

### Application/website

Carbon for IBM.com

### Business priority

Medium Priority = upcoming release but is not pressing

### What time frame would this ideally be needed by (if applicable)

Ideally can bring in soon, so teams using the library can start removing lit-element/lit-html from their application setup.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

jeffchew:
Will need to update the web components template once this is published in production.

PR #7556, opened 2021-11-01
chore(lit): add lit-html/lit-element as core dependencies

jeffchew:
### Related Ticket(s)

Refs https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7546

### Description

This moved `lit-element` and `lit-html` into the main `dependencies` for `Carbon for IBM.com Web Components`. Documentation has been added to highlight this.

This PR also alphabetizes the scripts on the root `package.json`.

### Changelog

**Changed**

- Web Components `package.json` dependencies for `lit-element`/`lit-html`
- Root `'package.json` alphabetized scripts


ariellalgilmore (approved):
LGTM! Thanks for adding this!

jeffchew:
Holding on this for a moment, looking into the Cypress errors in the CSB examples:

```The `classMap` directive must be used in the `class` attribute and must be the only part in the attribute.```

jeffchew:
I'm wondering if the error seen is something legitimate? In `lit-html` version `1.3.0` (which the project is now set to use), it has some documentation about it:

https://github.com/lit/lit/blob/f43b811405be32ce6caf82e80d25cb6170eeb7dc/packages/lit-html/src/directives/class-map.ts#L38

https://lit.dev/docs/templates/custom-directives/#limiting-a-directive-to-one-expression-type

Looking back at `1.2.0` of `lit-html`, which is what the csb examples were using before, here is the error which also seems to be part of `class-map.ts`:

https://github.com/lit/lit/blob/d4a0b986639a36c6997dd0ffe1d3eb79f15dfd23/src/directives/class-map.ts#L68

jeffchew:
Digging more into this, downgrading `lit-html` and `lit-element` doesn't appear to be an option because it seems to also will want to downgrade typescript to below v4 for the project to compile properly. 

I'm trying to understand how we're using `classMap`, but offhand seems like it shouldn't be an issue with the way we're using it?

jeffchew:
Thank you @abdonrd for the support on this! Looks like the dependency update and resolving resolution on `lit-html` is the fix for the integration build issues!

Issue #7424, opened 2021-10-19
[Test Scenario]: Footer - Short


Praveen-sr:
### Component and test scenario title

Footer - Short

### Test scenario steps

a. Check IBM logo and Select geo area button is available and clickable
b. Click on select geo area, it should open locale modal with 4 geos (Americas, Asia Pacific, Europe & Middle east and Africa)
c. On click of any geo's, the specific locations and languages must be displayed
d. Check the search edit field, user can search with keywords for locations and languages Note: Footer navigation links should not be present 
e. Check all footer legal navigation links are loading and clickable

Note: Footer navigation links must not be displayed.

PR #7561, opened 2021-11-01
test(footer): add e2e tests for footer short

BrunnoM7:
### Related Ticket(s)

[Test Scenario]: Footer - Short #7424

### Description

e2e test cases for the footer short covering the scenarios:
- Check IBM logo and Select geo area button is available and clickable
- Click on select geo area, it should open locale modal with 4 geos (Americas, Asia Pacific, Europe & Middle east and Africa)
- On click of any geo's, the specific locations and languages must be displayed
- Check the search edit field, user can search with keywords for locations and languages Note: Footer navigation links should not be present
- Check all footer legal navigation links are loading and clickable

### Changelog

**New**

- e2e test cases for footer short for react and web components


annawen1 (commented):
LGTM so far @BrunnoM7, just have to fix the conflicts

packages/react/tests/e2e-storybook/cypress/integration/Footer/Footer.e2e.js
```diff
@@ -114,82 +122,123 @@ describe('Footer | default (desktop)', () => {
...
-    });
+    cy.wait(500);
 
-    cy.get(
-      '.bx--locale-modal__locales:not(.bx--locale-modal__locales-hidden) > div'
-    )
-      .first()
-      .then(e => {
-        expect(e.text()).to.equal('Mexico');
+    cy.get('.bx--locale-modal [data-region="mea"]')
```
jeffchew, line 168: I think the only thing I'd ask is if there are stable selectors ([data-autoid="*"]) that are available and can be used instead of classes? Those would probably be more reliable long-term.

BrunnoM7: Hey Jeff, I removed some of the class selectors. But I leave some especially in the search results since they don't have stable selectors. Let me know if those are to be removed as well.

packages/react/tests/e2e-storybook/cypress/integration/Footer/Footer.e2e.js
```diff
@@ -165,17 +168,19 @@ describe('Footer | Short (desktop)', () => {
...
     cy.wait(500);
 
-    cy.get('.bx--locale-modal__filter')
-      .find('.bx--locale-modal__list li a')
+    cy.get('[data-autoid="dds--locale-modal"]')
+      .find('ul li a[href][data-region]')
       .each($locale => {
         if (!$locale.attr('data-region') === 'mea') {
-          $locale.should('have.class', 'bx--locale-modal__locales-hidden');
+          $locale.should('have.css', 'display', 'none');
```
jeffchew, line 183: Getting others' opinions, wondering if the assertion should be this instead:

```$locale.should('not.be.visible');```

cc: @annawen1 @ariellalgilmore

annawen1: yeah I like that instead of checking the css!

ariellalgilmore: agree! probably want to avoid relying on css when possible

BrunnoM7: Thanks! I updated the assertion to not rely on css.

jeffchew (approved):
LGTM!

PR #7576, opened 2021-11-02
test(e2e): various false positive fixes for Masthead and Quote

jeffchew:
### Related Ticket(s)

No related issue

### Description

We were noticing some false positives in the percy snapshots, which is probably due to elements not properly loading before the snapshots are being taken. For the tests in `Masthead`,  `CalloutQuote`, and `Quote`, additional selector gets were added in order to ensure that the proper elements are in place before taking a snapshot.

### Changelog

**Changed**

- e2e tests for `Masthead`, `CalloutQuote`, and `Quote`


IgnacioBecerra (approved):
LGTM, thanks! The `waitUntil` function seems to be proving its worth 👍

annawen1 (approved):
Nice!

Issue #6391, opened 2021-06-16
[Masthead with L1] Scroll up behavior


oliviaflory:
## Detailed description

> Describe in detail the issue you're having.

### Scroll up behavior
React Masthead with L1 is not following the scroll up behavior stated in the specs:
`Masthead L0 remains hidden if L1 is present, except reveal L0 on scroll up.`
[Masthead L1 specs](https://ibm.ent.box.com/folder/118448193520?s=o0v38tot6t0s9tnr3y3estp4cnxar6vo)
> Is this a feature request (new component, new icon), a bug, or a general
> issue?

> Is this issue related to a specific component?

Masthead L1, best seen in Dotcom shell with L1 story to demonstrate scroll

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Feature parody between React & WC, also update scroll up behavior

> What browser are you working in?

> What version of Carbon for IBM.com are you using?
[React canary](https://ibmdotcom-react-canary.mybluemix.net/?path=/story/components-dotcom-shell--with-l-1)

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

- Screenshots or code
- Notes

PR #7548, opened 2021-10-29
fix(table-of-contents): bump navbar z-index

emyarod:
### Related Ticket(s)

related PR #5476

### Description

When scrolling through a page with interactive media and a horizontal table of contents component, the media is interactive while the ToC is over the media element. This can be found in the "Dotcom shell with ToC horizontal" story

https://user-images.githubusercontent.com/8265238/139480676-48f0ee9d-981f-4c8e-801c-9db858ebb19c.mov

This PR increases the z-index of the ToC navbar so that it is greater than interactive pseudoelements

e.g. [video player](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/6ff8ac10534f2e89c40e4f2c301da481f9d76d00/packages/styles/scss/components/video-player/_video-player.scss#L69), [feature card](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/6ff8ac10534f2e89c40e4f2c301da481f9d76d00/packages/styles/scss/components/feature-card/_feature-card.scss#L119), etc

### Changelog

**Changed**

- table of contents navbar z-index

PR #7591, opened 2021-11-03
fix(masthead): display L0 nav with L1 when scrolling up

emyarod:
### Related Ticket(s)

#6391

related https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/97

### Description

This PR updates the Dotcom Shell Composite web component and React Masthead behavior to show the L0 nav when scrolling up if an L1 nav is also present

In the Dotcom Shell Composite web component, this is done by changing the Table of Contents component's `stickyOffset` **property** on scroll if an L1 is present. The `stickyOffset` **attribute** is not changed though, so one point of discussion for the reviewers would be whether or not we also want the attribute to reflect the property. If we do, then I believe we'll need to store the original attribute value the shell composite for calculations

In the React Masthead component, this is done by setting the sticky masthead classes based on scroll direction if an L1 is present. The actual offset values are on the application side so the storybook content component needs to know whether or not there is an L1 to pass the correct offset value to `<TableOfContents />`

https://user-images.githubusercontent.com/8265238/140102486-23fbb3b2-5ec5-4634-b158-ba9d89ad0a94.mov

### Changelog

**New**

- `withL1` prop on `<Content />`

**Changed**

- Dotcom Shell Composite masthead and table of contents position styles
- React Masthead sticky styles logic


oliviaflory:
Hey @emyarod overall this looks really good! Adding some observations that may or may not be in scope, we can break out to other issues if needed.

React
1. On Desktop, using the TOC to scroll back up, the headline is cut off Desktop. It doesn't do that in the prod storybook, but the L0 also doesn't come into view in prod today. 
Screen Shot 2021-11-04 at 3 38 23 PM


Desktop React scroll upNov-04-2021 15-39-27


Pretty sure this isn't in scope because it's in prod as well. Happening in both React and Web components

On medium or small breakpoint, if I interact with the TOC while it's over the Feature card, I am taken to the feature card's link. Same thing happens over the Call out with media, the video plays rather than opening the native drop down with the TOC.
TOC-callout--Nov-04-2021 15-56-00

emyarod:
> React
>
> 1. On Desktop, using the TOC to scroll back up, the headline is cut off Desktop. It doesn't do that in the prod storybook, but the L0 also doesn't come into view in prod today.

@oliviaflory this should be resolved now!

> Pretty sure this isn't in scope because it's in prod as well. Happening in both React and Web components
> 
> On medium or small breakpoint, if I interact with the TOC while it's over the Feature card, I am taken to the feature card's link. Same thing happens over the Call out with media, the video plays rather than opening the native drop down with the TOC.

I fixed this already in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/7548! it hasn't been released yet but it should be working correctly in Canary

oliviaflory:
@emyarod thank you for the update!

1. React: The cut off headline is definitely fixed on Desktop breakpoints 👍 
~I do see a bit of variation in cut off vs visible on the medium and small breakpoints where the TOC changes to the select. Showing the variations here with two scroll down and one scroll up, maybe the variation depends on how each band/section is set up within the Dotcom shell story?~   Scratch all that, the TOC should be the only sticky element in the medium and small breakpoints when the TOC is present on a page. Lmk if this is not in scope, we could address in another issue. 

2. 
> I fixed this already in #7548! it hasn't been released yet but it should be working correctly in Canary

Yay!

3. Additionally, now when using the TOC the 'active' item and the headline are disconnected. If I click on the TOC item "Lobortis elementum nibh tellus" I scroll to the correct content block, but the active TOC item is the section above "Tincidunt ornare massa". I hope that makes sense, including a GIF as well.

Not sure if that is in scope for this PR? 

TOC-item--Nov-09-2021 15-45-53


4. Web components: I see a bit of variation in cut off vs visible on the medium and small breakpoints where the TOC changes to the select. And a bit of disconnect of which item/section is active.
Maybe the variation depends on how each band/section is set up within the Dotcom shell story and the variation in spacing between section? If this isn't in scope LMK, seems like the shell might need a bit of cleaning up.
Screen Shot 2021-11-09 at 4 10 18 PM
Screen Shot 2021-11-09 at 4 10 33 PM

emyarod:
@oliviaflory I believe I addressed all of the points in your last review! but if 1 and 4 still persist then they should probably be tracked separately since it's slightly out of the scope of this issue

packages/styles/scss/components/masthead/_masthead.scss
```diff
@@ -213,25 +213,26 @@ $search-transition-timing: 95ms;
...
-  .#{$prefix}--masthead--sticky__l1 + .#{$prefix}--dotcom-shell {
-    @include carbon--breakpoint-up(800px) {
-      .#{$prefix}--tableofcontents__sidebar {
-        top: 98px;
-      }
-    }
+  .#{$prefix}--masthead--sticky__l1
+    + .#{$prefix}--dotcom-shell
+    .#{$prefix}--tableofcontents__sidebar {
+    top: 98px;
```
IgnacioBecerra, line 226: Might be better using the `rem()` function in the stylesheet to keep things consistent across styles?

emyarod: @IgnacioBecerra should I convert the existing styles there to `rem` as well?

PR #7592, opened 2021-11-03
fix(browserstack): various browserstack e2e test fixes

jeffchew:
### Related Ticket(s)

No related issue

### Description

This took some doing, but this should address all of the e2e test failures on Browserstack. Here is the rundown:

1) Had to remove the importing of `carbon-components` in the `footer.e2e.js` tests. Browserstack couldn't resolve that package when pulling in the integration folder. When adding it as a separate dependency in Browserstack itself, it threw an error when trying to run the Carbon Telemetry postinstall script, and thus preventing all of the Windows runs to even fire:

image

2) Added an exception to ignore the `Request aborted` error happening in Firefox only. This seemed to be the only way to get around this until we can figure out a better way to handle this.

3) Activated video capture in Browserstack. So far, it seems like this only works in Chrome, but looking at Cypress v8.7.0 changelog, and there is a fix to have the videos run in other browsers. Browserstack is currently running Cypress v8.5.0.

4) Masthead scroll tests were failing in React as it wasn't targeting the correct element for testing. the detection to check if the left carat is visible was also switched using a more reliable check.

5) Added an intercept for the Masthead search typeahead API. It is now always returning test results, which should help as it was before hitting live data which isn't good for tests like these.

### Changelog

**Changed**

- See above

### DIFF
--- a/packages/react/tests/e2e-storybook/browserstack.json
+++ b/packages/react/tests/e2e-storybook/browserstack.json
@@ -35,7 +35,6 @@
     "npm_dependencies": {
       "@percy/cypress": "^3.1.0",
       "@percy/dom": "^1.0.0-beta.55",
-      "carbon-components": "^10.47.0",
       "cypress-wait-until": "^1.7.2"
     },
     "package_config_options": {},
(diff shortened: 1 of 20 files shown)


andysherman2121 (approved):
nice work!

Issue #7587, opened 2021-11-03
Web component: Masthead - Search edit field close icon is not working


Praveen-sr:
### Description

Issue: 
On Mobile:
Masthead search edit field close icon does not clear the keyword and does not close edit field.
Once the user enters keyword and clicks on close icon, user again opens the search field then close icon is missing.

Note: Working fine on react version.

### Component(s) impacted

![Mast3](https://user-images.githubusercontent.com/60002341/140065306-c6682f29-8b99-4934-9d2b-eef52bfb712d.gif)

### Browser

Chrome, Safari, Firefox

### Carbon for IBM.com version

1.27

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

Web component

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

NA

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

jeffchew:
@Praveen-sr was monitoring e2e tests after the hotfix was merged, it looks like it might have introduced another bug where clocking away from the search box will close it out, even when there are results showing. Can you verify as part of your testing?

![Kapture 2021-11-03 at 20 19 44](https://user-images.githubusercontent.com/1641214/140236931-d22b2ac9-74d9-459b-927e-cc93f079d039.gif)

cc: @IgnacioBecerra

Praveen-sr:
@jeffchew I verified on web component:
Yes, clicking away from the search box will close it out, even when there are results showing.

![Search_close](https://user-images.githubusercontent.com/60002341/140280804-9c8433cc-53b5-4b24-965d-291a4376f7d2.gif)
cc @RobertaJHahn 


Expected: 

![Search_close_expected](https://user-images.githubusercontent.com/60002341/140280979-3754f5d4-0075-4bb1-a4ec-7712b3932bb5.gif)

PR #7200, opened 2021-09-20
fix(search-with-typeahead): Prevent search form closure on focusout

andy-blum:
### Related Ticket(s)
fixes #6321
reverts #7069

### Description

This PR removes the logic in the components `_handleFocusOut` method. The method & super call remain to close the dropdown on `focusout`, but closing the form altogether no longer happens. Since the form closure behavior is removed, we can also remove the span tag added in #7069 to prevent safari's a-typical behavior from closing/re-opening the search bar.

### Changelog

**Changed**

- `dds-search-with-typeahead` no longer closes on focusout

PR #7596, opened 2021-11-04
fix(search-with-typeahead): remove focusout code

IgnacioBecerra:
### Related Ticket(s)
#7587 

### Description
Removes the `handleFocusOut` code that used to be in the `dds-masthead-search` component since it causes a cypress test to fail, as well as any search toggle related issue.

### Changelog
**Removed**

- removes `handleFocusOut` code from `dds-search-with-typeahead` so its reverted to how it was since #7200


jeffchew (approved):
Strangely the tests run fine now, LGTM!

PR #7603, opened 2021-11-04
chore(lit): update lit dependencies

abdonrd:
Update Lit dependencies in the same way as `carbon-web-components`.

Also move the `resolutions` to the web-components package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,6 @@
   },
   "resolutions": {
     "kind-of": ">=6.0.3",
-    "lit-html": "^1.4.0",
     "minimist": ">=1.2.2",
     "node-notifier": ">=8.0.1"
   }
--- a/packages/web-components/package.json
+++ b/packages/web-components/package.json
@@ -86,6 +86,7 @@
     "wca": "web-component-analyzer analyze src --outFile custom-elements.json"
   },
   "resolutions": {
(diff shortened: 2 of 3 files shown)


jeffchew:
Hey @abdonrd , thank you for the PR!

The resolutions actually only seem to work on the root of the project, which is why we mainly put them in there. If you notice, the downstream test app is actually blowing up like we saw since it seems to be pulling in an older `lit-html` again that doesn't play nicely with `lit-element`:

image


Also, I like to keep the dependencies locked (especially core dependencies) so there aren't any random surprises down the line.

abdonrd:
> Hey @abdonrd , thank you for the PR!
> 
> The resolutions actually only seem to work on the root of the project, which is why we mainly put them in there. If you notice, the downstream test app is actually blowing up like we saw since it seems to be pulling in an older `lit-html` again that doesn't play nicely with `lit-element`:

Oh! That's right! https://github.com/yarnpkg/yarn/issues/5039

abdonrd:
> image
> 
> Also, I like to keep the dependencies locked (especially core dependencies) so there aren't any random surprises down the line.

Regarding this... I think it would be better to try to leave the range (with `^`). Since if an application uses Carbon components + other components made in Lit... Otherwise they share the same range, there could be different versions of Lit.

PR #7607, opened 2021-11-05
fix(cla-assistant): updated kodiakhq name for allowlist

jeffchew:
### Related Ticket(s)

No related issue

### Description

There looks to be an issue trying to get `kodiakhq` to be ignored in the CLA:

image

In addition, signing the CLA was unclear before, made updates so that it should be clearer and is now working.

### Changelog

**Changed**

- `cla.yml` allowlist change for `kodiakhq[bot]`
- Added better instructions for users to sign properly
- Changed the message to the default signature comment


emyarod (approved):
looks good to me!

PR #7608, opened 2021-11-05
feat(leadspace & button group): add print friendly styles

proeung:
### Related Ticket(s)

DDS Issue: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7413
HC JIRA Ticket: https://jsw.ibm.com/browse/HC-2297

### Description

- This PR adds print-friendly styles for the Button Group and Leadspace components and variations based on the design specs (see below).
  - Lead space: https://ibm.ent.box.com/file/879963299367
  - Button group: https://ibm.ent.box.com/file/879964531443

### Changelog

**Changed**

-  Add `href` prop to `renderInner()` event so that we can showcase the button `href` for print mode type and hide it on normal media types.

### Expected designs
Screen Shot 2021-11-05 at 12 24 14 PM
Screen Shot 2021-11-05 at 12 24 23 PM

### Testing
- Visit the "Button group" and "Leadspace" components within the Storybook preview.
- Inspect the page using Chrome DevTools and open the rendering tab.
- Scroll down to the Emulate CSS Media Type and selected "Print"
- From here you should be able to view the print style and compare it with the design specs to ensure the styling are looking as expected.


packages/web-components/src/components/button/button.ts
```diff
@@ -94,6 +94,7 @@ class DDSButtonExpressive extends FocusMixin(StableSelectorMixin(LitElement)) {
     const { _handleSlotChange: handleSlotChange } = this;
     return html`
       <slot @slotchange="${handleSlotChange}"></slot>
+      <p class="${prefix}--btn--hidden" aria-hidden="true">${'(' + this.href + ')' /* eslint-disable-line prefer-template */}</p>
```
IgnacioBecerra, line 97:
```suggestion
      <p class="${prefix}--btn--hidden" aria-hidden="true">(${this.href})</p>
```

Concatenation like this may not be needed after all since this is a big string after all, the parentheses will be a normal part of the markup and then the href will be placed inside properly.

RichKummer:
Thanks for this @proeung! This is looking good, the one issue I noticed is that the URL appears in parentheses when the CTA copy should end with a colon and then followed by the URL
Design specs
Screen Shot 2021-11-08 at 9 42 23 AM
:

Current build
Screen Shot 2021-11-08 at 9 57 34 AM

proeung:
@RichKummer Looking at the current design specs, it looks to me like we're using both parentheses and colon (see attached), so if I'm understanding what you're saying above correctly, there should NOT be any parentheses, but a colon right next to the CTA text? Is that correct?

Screen Shot 2021-11-08 at 10 18 51 AM

RichKummer:
> @RichKummer Looking at the current design specs, it looks to me like we're using both parentheses and colon (see attached), so if I'm understanding what you're saying above correctly, there should NOT be any parentheses, but a colon right next to the CTA text? Is that correct?
> 
> Screen Shot 2021-11-08 at 10 18 51 AM

Yes, that's correct. The button should appear as **CTA copy:** url

Ex. **Bring AI to your business:** https://www.ibm.com/services/artificial-intelligence?lnk=hpmsc_bups)

I'll update the Design specs.

proeung:
@RichKummer Thanks for verifying the correct treatment for the CTA print style. I just pushed up a commit that removes the parentheses and adds the colon.

IgnacioBecerra (approved):
LGTM, thanks @proeung!

proeung:

RichKummer (approved):
Looks great! Thanks for making the last round of changes 👍

jeffchew:
@annawen1 just checking if you accepted the DCO yet?

annawen1:

PR #7630, opened 2021-11-09
fix(percy): adding waitForTimeout for false positive stories in Safari

jeffchew:
### Related Ticket(s)

No related issue

### Description

Noticing that the `card group with images` and `leaving ibm` stories were showing false positives consistently in Safari. This is an attempt to have these stories skipped in Safari only (but still run in Chrome).

### Changelog

**Changed**

- add `percy skip` to `card group with images` and `leaving ibm` for Safari


jeffchew:
Closing this, will do further investigation offline as this approach doesn't appear to be working.

PR #7639, opened 2021-11-10
chore(dependabot): add config and update dependabot offline mirror flow

jeffchew:
### Related Ticket(s)

No related issue

### Description

This introduces a dependabot configuration for the repo (should have added one awhile ago).

Let me know if you think the versioning strategy is too aggressive. It would be good for the dev dependencies if it has problems upgrading but we'd probably have to keep an eye on the ones that would cause any breaking changes.

### Changelog

**New**

- `dependabot.yml` config in `.github`

**Changed**

- fixes to the dependabot offline mirror management workflow

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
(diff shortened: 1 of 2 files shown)


IgnacioBecerra (approved):

jeffchew:
Going to actually close this out, will switch to use renovate since it handles yarn offline mirror / monorepos better than dependabot.

Issue #7139, opened 2021-09-14
[Card - logo] Web Component: Develop a card variation dedicated to logo / app icons - DDS consulting


RobertaJHahn:
**Engineering info:**  
- HC engineer: @proeung 
- HC JIRA Ticket:  https://jsw.ibm.com/browse/HC-2312
- DDS consulting engineer: @IgnacioBecerra 

----

#### User Story

> As a `[user role below]`:
Carbon for ibm.com developer

> I need to:
create/change the `Card - logo`

> so that I can:
provide the ibm.com adopter developers components they can use to build ibm.com web pages

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- Create with Shadow DOM and Custom Elements standards
- **See the Epic (#6592 ) for the Design and Functional specs information**
- [Design and functional spec Box folder](https://ibm.ent.box.com/folder/95514001950s=hgvu48r1fv2ga2h0yn3qshrtxjc71vch)
- Web Components Visual QA testing issue (#7142)
- Web Components Prod QA testing issue (#7144)

#### Acceptance criteria
- [ ] Include README for the web component and corresponding styles
- [ ] Create Web Components styles in styles package
- [ ] Do not create knobs in Storybook that include JSON objects
- [ ] Break out Storybook stories into multiple variation stories, if applicable
- [ ] Create codesandbox example under `/packages/web-components/examples/codesandbox` and include in README
- [ ] Minimum 80% unit test coverage
- [ ] Update the Carbon for ibm.com website component [data file](https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/blob/master/src/data/components.json) to be sure the web site Component Status and Storybook links are correct
- [ ] Use the [Visual QA checklist](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/wiki/Definition-of-done-(Visual-QA-checklist)) to verify design quality
- [ ] If a design is provided, the Designer is included as a Reviewer in the Pull Request
- [ ] Provide a direct link to the deploy preview for the designer in the Pull Request description
- [ ] A comment is posted in the Prod QA issue, tagging Praveen when development is finished

proeung:
@RobertaJHahn Is this feature on the agenda for the 09/27 "Engineering Epic and Story refinement" meeting? If so, I'd like to attend to review the design and functional specs of this logo/app icon variation within the Card component.

RobertaJHahn:
[...]

PR #7647, opened 2021-11-11
feat(card): add logo variant

proeung:
### Related Ticket(s)

DDS Consulting Issue: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7139
HC JIRA Ticket: https://jsw.ibm.com/browse/HC-2312

### Description
- This PR adds a new variant within the `Card` component called `Card logo`. 

### Changelog

**New Components/Props**

- `<dds-image-logo>` - this component extends from `<dds-image>` and is needed because of styling requirements.

**Removed**

- N/A

### Testing Instructions

- Visit the `Card` UI component with `Logo` variant - `?path=/story/components-card--logo`
- Ensure that the preview matches with the design specs 
  - [Design and functional spec Box folder](https://ibm.ent.box.com/folder/95514001950?s=hgvu48r1fv2ga2h0yn3qshrtxjc71vch)


ariellalgilmore:
Thanks @proeung! Not opposed to creating `dds-image-logo`, but wondering if we just want to make this within the card directory such as `dds-card-logo` or a `logo` prop in `dds-image` ?

cc: @IgnacioBecerra

proeung:
@ariellalgilmore Thanks for the review! What you're suggesting makes sense as well. I created `dds-image-logo` based on my discussion with @IgnacioBecerra and the potential future usage of that component, but I'm open to adding a `logo` prop in `dds-image` or creating a `dds-card-logo` within the `/card` directory if these implementations make more sense. Let me know which direction I should take.

IgnacioBecerra:
@proeung FYI we're checking with the other devs to get their thoughts on this as well, will keep you updated 👍

IgnacioBecerra:
@proeung We agreed that it was better to extend the component since the adding a prop/attribute should mainly affect that specific component (like `hidden`, `disabled`) instead of changing in favor of another component. So your current changes are good, just a little thing, could you move the `dds-image-logo` file to the `card` folder? Just so we can have all the files needed for Card in one place, especially to be added to the `index.ts` file 😄

IgnacioBecerra (approved):
LGTM, thanks @proeung!

proeung:
@IgnacioBecerra No prob!

Looks like we just need design QA approval for this PR. @skuruvil Can you help take a look? Thanks!

skuruvil:
@proeung Here are some of my observations. Please let me know if you have any questions.

1. `Text link CTA`  is optional to use **or have the entire card area be interactive**. I think we don't have an `Icon CTA `variant. We have defined the interactive states as well in the design spec (Since we have the logo/image, interactive states applied only to the border but not to the background)

Screenshot 2021-11-16 at 8 54 17 PM


2. We have one more optional variant with a `heading`.  I can see the card heading option in the knobs but it's not reflecting on the preview. 

Screenshot 2021-11-16 at 8 56 43 PM

3. Right now the logo/image size seems to be bigger, it is supposed to be 160px x 80px fixed in all the breakpoints

Screenshot 2021-11-16 at 8 57 50 PM


4. The space between the logo and the eyebrow is 24px in design spec, right now it seems like 16px.


Screenshot 2021-11-16 at 9 00 03 PM

5. @small we are not adding 32px padding to the text width but it will be full width

Screenshot 2021-11-16 at 9 01 05 PM

6. The minimum aspect ratio we have set as 4:3 in design spec, in the preview it seems like 1:1

Screenshot 2021-11-16 at 9 03 51 PM

proeung:
@skuruvil I just pushed up a commit that should address all of the design QA feedback that you've listed above. Can you take a look again?

skuruvil:
@proeung Everything looks great to me, thanks. I found a small issue. The bottom margin seems to be more in the preview. In design spec it's 32pix total bottom margin however in the SB preview it seems like 32px + 16px padding. I am thinking that could be the one reason that the 4:3 minimum ratio is still not achieved, there is a slight difference in the height.  

Screenshot 2021-11-19 at 3 40 03 PM

skuruvil:
@proeung looks perfect to me. Thanks

skuruvil:

proeung:
recheck

jeffchew:
@proeung looks like CI checks are failing for web components, it's having a hard time finding the image being imported:

image

packages/styles/scss/components/image/_image.scss
```diff
@@ -37,6 +38,19 @@
     top: calc(50% - 32px);
     right: calc(50% - 32px);
   }
+
+  // dds-image-logo style.
+  :host(#{$dds-prefix}-image-logo) {
```
jeffchew, line 43: Sorry for the last minute, an integration test is failing and seems to be coming from here:

image

jeffchew (changes requested):
Noted an integration test failing with the `Image` component.

packages/styles/scss/components/image/_image.scss
proeung: @jeffchew I just added the missing `dds` prefix variable (see - https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/7647/commits/c8ccf4150b98dc0a40b8afe0f29abb1118c8a2b3). Hopefully that should fix the build error.

jeffchew:
@proeung thanks for changing!

It looks like there are merge conflicts now with @ariellalgilmore 's latest PR merged in.

jeffchew (approved):
Thank you for the update, glad we caught that integration issue!

proeung:
@jeffchew @ariellalgilmore Just resolved the merge conflicts. Can you take a look again?

ariellalgilmore:
Looks good! Wondering though if we can update the knobs to have the same format though as the other stories now? Also, @skuruvil do you think the logo variant should be it's own story or maybe just an additional knob under "Add image:" then we could add "Add logo:"

proeung:
@ariellalgilmore My understanding is that "Logo" is its own variant of the Card component since it has different design specs and the mouse over effect is slightly different from your regular card or static variants (see [link](https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/7647/index.html?path=/story/components-card--logo)).

For the purpose of moving these changes forward in the `v.128` release for AEM adoption, I think we can capture any follow-up enhancements for the settings of the knobs in a different issue.

ariellalgilmore:
@proeung okay thanks! Yeah let's get this one in

ariellalgilmore (approved):

jeffchew:
Waiting for the React CSB examples job to run and if it looks good, will get this one merged in!

Issue #7562, opened 2021-11-02
[Test Scenario]: Link list – End of section


shixiedesign:
### Component and test scenario title

Link list – End of section

### Test scenario steps

1. Check all link list items are showing the correct states: hover, click, focus, and active
2. Check knobs CTA type = [ local, jump, external, download, video ] and nothing else
3. Check when knob CTA type = local, items uses an arrow right icon, and takes user to example page
4. Check when knob CTA type = jump, items uses an arrow down icon, and do nothing on click
4. Check when knob CTA type = external, items uses a launch icon, and takes user to example page in new tab
5. Check when knob CTA type = download, items uses a download icon, and downloads the annual report.
6. Check when knob CTA type = video, items uses a play icon, adds video duration to item heading, and on click plays test video
3. Check link list design on mobile & desktop are unchanged

PR #7648, opened 2021-11-11
test(link-list): add e2e storybook cypress tests

ariellalgilmore:
### Related Ticket(s)

#7562 

### Description

Added link-list e2e testing for react and web components.

Updated web component link-list styling for hover state background. Also, added styles so when changing CTA knobs on web components the arrow wouldn't collapse next to the left text.

### Changelog

**New**

- tests for link-list
- checking each CTA type
- styles for hover,active, and focus states

**Changed**

- added display: flex for web components
- changed hover color
- react storybook download to use a pdf link


packages/react/tests/e2e-storybook/cypress/integration/LinkList/LinkList.e2e.js
```diff
@@ -0,0 +1,147 @@
...
+/**
+ * Sets the correct path
+ *
+ * @type {string}
+ * @private
+ */
+const _pathEndOfSection =
+  '/iframe.html?id=components-link-list--end-of-section';
+
+const getCssPropertyForRule = (rule, prop, sheets) => {
```
annawen1, line 17: This is great, I wonder if we can pull this out into a util folder or something so other tests can use this

ariellalgilmore: ooh yeah makes sense. i'll change that

annawen1 (approved):
LGTM! Might want to think about pulling out the cssproperty method into a folder somewhere since I'm sure other tests will use that

Issue #7269, opened 2021-09-29
[Content item horizontal with featured media] Web Component: Develop new featured media variant - DDS consulting


RobertaJHahn:
**Engineering info:**
- HC engineer:@jkaeser
- HC Jira Ticket: https://jsw.ibm.com/browse/HC-2298
- DDS consulting engineer:  Anna Wen

----

#### User Story

> As a `[user role below]`:
Carbon for ibm.com developer

> I need to:
create/change the `Content item horizontal with featured media`

> so that I can:
provide the ibm.com adopter developers components they can use to build ibm.com web pages

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- Create with Shadow DOM and Custom Elements standards
- **See the Epic (#7256) for the Design and Functional specs information**
- **Content item horizontal with media specs**
[PDF](https://ibm.ent.box.com/file/869427591424?s=2h7ultdzzh4vbn9g31w8kijhexufvvut)
[Sketch](https://ibm.ent.box.com/file/869453525289?s=ruwgj206fa3uhb7pgyyu0dmeugdgtlw0)
- Web Components Visual QA testing issue (#7273)
- Web Components Prod QA testing issue (#7276)

----**Requested new component for "Featured media"**
![Content item horizontal with featured media@2x](https://user-images.githubusercontent.com/77290329/134953367-36429182-2e2a-4a14-bcc0-33ae59e8d786.png)
![Content item horizontal with featured media (320px_sm)@2x](https://user-images.githubusercontent.com/77290329/134953486-bfd7ec86-421c-43ba-acd3-9ec68beef301.png)
----

#### Acceptance criteria
- [ ] Include README for the web component and corresponding styles
- [ ] Create Web Components styles in styles package
- [ ] Do not create knobs in Storybook that include JSON objects
- [ ] Break out Storybook stories into multiple variation stories, if applicable
- [ ] Create codesandbox example under `/packages/web-components/examples/codesandbox` and include in README
- [ ] Minimum 80% unit test coverage
- [ ] Update the Carbon for ibm.com website component [data file](https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/blob/master/src/data/components.json) to be sure the web site Component Status and Storybook links are correct
[...]

PR #7650, opened 2021-11-11
feat(content-item-horizontal): create featured media variant

jkaeser:
### Related Ticket(s)

Part of https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7269

### Description

This PR adds the `Content item horizontal with featured media` variant to the `Content item horizontal` component

### Changelog

**New**

- adds the `<dds-content-item-horizontal-media-featured>` web component, along with related tests and documentation


photodow:
- [ ] Are we using [Image with caption](https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/wiki/Image-with-caption) here? I think the [functional specs](https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/wiki/content-item-horizontal#3-media) calls out image with caption, and not just image.
- [ ] Let's remove the lists in the copy.
- [ ] Let's remove the last two sentences in main paragraph of copy.
- [ ] I don't understand the chosen knobs in content item horizontal across the board, but I see you've just copied and pasted here. Should consider breaking away from the others here, or update the other knobs? Here are a couple things to consider...
     - remove 3 knobs for each CTA. If we have any knob here it should be a toggle, or a number value or something.
     - image alt text isn't needed for the storybook demo I don't think.
     - if we have things like heading, eyebrow, we should have an area to update the copy as well.
     - some of the knob names could be updated. A couple of them are redundant like "Heading (heading):"

jkaeser:
Thanks for the review, @photodow! I made some changes to the Featured Media variant story based on your feedback.

1. Used `<dds-image-with-caption>` instead of `<dds-image>`.
2. Removed the lists and sentences you requested.
3. Added a knob for the body copy (I did this for all variants). 
4. Removed the alt text knob and the CTA href and type knobs. The CTA copy knobs seem valuable for folks to test how short/long CTAs appear in this component's context, so I left those alone.
5. Adjusted redundant knob labels.

Please take a look and let me know if you'd like to see any additional changes.

photodow:
Great improvements. Thanks! Here are just a couple small maybe even nit picky things.

- [ ] in storybook turn off the lightbox option for the image with caption.
- [ ] remove label and just call it eyebrow.
- [ ] If you could also put the eyebrow above the heading if possible that'd be awesome.
- [ ] weird formatting issue with the copy
- [ ] copy in storybook is not updating, is that just me? If that's going to be a problem I guess we can remove it then. It looked so nice and complete though.
- [ ] if we can get copy working and we don't have to remove it than we can just call it "copy" and remove the "body" from the label.

Thanks!

jkaeser:
@photodow Unfortunately, it looks like there's a bug in either `DDSContentItemCopy` or `DDSMarkdown` that prevents the copy from updating. `DDSContentItemHorizontalMediaCopy` inherits from these, and it is not updating when the knobs send it new content. I've removed the "Copy" knob for now, since fixing this would deserve its own ticket.

I was able to make the other changes you requested, though. Back to you for further review!

photodow (approved):
A lot cleaner. Thanks for working on this!

annawen1 (approved):
Code-wise LGTM! Thanks @jkaeser!

Issue #7575, opened 2021-11-02
Safari baseline: [Feature card - medium & large] - No DDS consulting


shixiedesign:
**Engineering info:**  
- HC engineer: @jkaeser 
- HC JIRA Ticket:  https://jsw.ibm.com/browse/HC-2310
- DDS consulting engineer: No consultant

----

1. Arrow became blue 60 which is not accessible. Verified it's actually a bug in Safari storybook
2. Arrow position changed. Moved right. 

Note these bugs exist in desktop as well, both Feature card medium, and large. 

![feature card.gif](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/ca542f2d-8484-4e6b-8bd1-05cc9f2aa0ba)

PR #7653, opened 2021-11-12
fix(feature-card): correct icon display in Safari

jkaeser:
### Related Ticket(s)

Resolves https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7575

### Description

The Feature Card's footer was inheriting an unnecessary CSS class that, in Safari only, caused an unwanted selector to take precedence. I've removed this class from the footer to resolve the browser inconsistencies.

### Changelog

**Changed**

- Removed an unnecessary class from `<dds-feature-card-footer>` to ensure correct appearance in Safari


photodow (approved):
Any reason we have new snapshots under the end to end percy? Otherwise this looks good. Thanks!

IgnacioBecerra (approved):
LGTM! 
@photodow Not sure why those extra snapshots are there, but they're coming from the e2e we've done in the past few sprints, I'd say we can ignore them.

Issue #6473, opened 2021-06-24
[Masthead L1] Web Component: Change the active menu item to be visible on page load always no matter where it is in the menu


RobertaJHahn:
#### User Story

> As a `[user role below]`:
Carbon for ibm.com developer

> I need to:
create/change the `Masthead L1`

> so that I can:
provide the ibm.com adopter developers components they can use to build ibm.com web pages

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- Create with Shadow DOM and Custom Elements standards
- **See the Epic for the Design and Functional specs information**
- Be sure that this feature also works in RTL
- Web Components Visual QA testing issue (#6503 )
- Web Components Prod QA testing issue (#6516 )

#### Acceptance criteria
- [ ] Include README for the web component and corresponding styles
- [ ] Create Web Components styles in styles package
- [ ] Do not create knobs in Storybook that include JSON objects
- [ ] Break out Storybook stories into multiple variation stories, if applicable
- [ ] Create codesandbox example under `/packages/web-components/examples/codesandbox` and include in README
- [ ] Minimum 80% unit test coverage
- [ ] Update the Carbon for ibm.com website component [data file](https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/blob/master/src/data/components.json) to be sure the web site Component Status and Storybook links are correct
- [ ] Use the [Visual QA checklist](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/wiki/Definition-of-done-(Visual-QA-checklist)) to verify design quality
- [ ] If a design is provided, the Designer is included as a Reviewer in the Pull Request
- [ ] Provide a direct link to the deploy preview for the designer in the Pull Request description
- [ ] A comment is posted in the Prod QA issue, tagging Praveen when development is finished

RobertaJHahn:
@emyarod Hey team! Can you please add your [planning poker estimates with ZenHub](https://app.zenhub.com/workspaces/Carbon-for-IBMcom-5d449f3642eb1962336cbe52/issues/carbon-design-system/carbon-for-ibm-dotcom/6473?planning-poker)

PR #7655, opened 2021-11-12
feat(masthead): ensure active l1 item is seen on load

IgnacioBecerra:
### Related Ticket(s)
#6473

### Description
This ensures that the selected L1 item will be scrolled to if its out of view upon load. This change also works on the RTL version.

In order to test this out, go to the Masthead L1 story, and under `selected menu item` type `Sed cursus ante dapibus diam`, and open the story in a new tab so the active element updates. You will be able to see the new change upon load. 

### Changelog

**New**

- selected item on the L1 menu will be scrolled to if its out of view


annawen1 (approved):
Testing in the deploy preview, it looks good to me!

Issue #7433, opened 2021-10-19
[Test Scenario]: Pictogram-item


Praveen-sr:
### Component and test scenario title

Pictogram-item

### Test scenario steps

a. Check pictogram item and content is loaded 
b. Check the Link with icon is loaded and clickable
c. Verify Pictogram item can be customized (desktop, touch, pattern)
d. Customize theme from Carbon theme and verify gray 10, gray 90 and gray 100 theme.

PR #7661, opened 2021-11-15
docs(pictogram-item): add select pictogram knob

emyarod:
### Related Ticket(s)

unblocks #7433

### Description

This PR adds a pictogram selector knob to the web components storybook to match the React storybook.

The SVGs are currently hard coded as a temporary stop gap to unblock #7433 tests, but we should probably import the pictograms directly from the upstream Carbon package. We have a couple of options on how to do this (this can be an engineering huddle discussion):

* add `@carbon/pictograms` as a dependency, add `svg-inline-loader` for webpack, and import the SVGs directly inline

or

* enhance [svgResultCarbonIcon](https://github.com/carbon-design-system/carbon-web-components/blob/master/tools/svg-result-carbon-icon.js) in `carbon-web-components` so that we can import pictogram helpers similar to the icon helpers we currently use (leaning towards this so that we can use them in other components in the library)

Either of these can be implemented in the future to avoid hard coding pictograms

### Changelog

**New**

- pictogram selector in Pictogram Item web component storybook


jeffchew (approved):
LGTM!

Issue #6409, opened 2021-06-17
[Masthead] React storybook demo clean up


oliviaflory:
## Detailed description

Masthead storybook clean up, Web components issue: #6407 
**_Content updates are documented in the web components issue #6407_** 

> Describe in detail the issue you're having.

### Default
Knobs:
- [x] no platform knob, default global nav should not enable a platform name
- [x] remove "Other" tab

### With custom navigation
- [x] no platform knob, default global nav should not enable a platform name
- [x] Label should be sentence case: `With custom navigation` (not initial caps)

### Search Open Onload
- [x] label should be in sentence case, not initial caps `Search open onload`

### With Platform
- [x] label should be in sentence case, not initial caps `With platform`

### With Alternate Logo And Tool Tip
- [x] Label should be sentence case: `With alternate logo and tool tip` (not initial caps)

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

storybook clean up

> Is this issue related to a specific component?

Masthead demo & variants

> What did you expect to happen? What happened instead? What would you like to
> see changed?

> What browser are you working in?

Chrome

> What version of Carbon for IBM.com are you using?

[React](https://ibmdotcom-react-canary.mybluemix.net/?path=/story/components-masthead--default)

RobertaJHahn:
engineering needs the content from design before they can make the changes.  Olivia will talk to Shixie about when to schedule the design work.  She may be able to work on it this sprint.

oliviaflory:
@RobertaJHahn content updates are documented in the web components issue #6407. @shixiedesign signed off on the updates in that issue so this is ready for scheduling.

PR #7664, opened 2021-11-15
fix(masthead): clean up storybook demos

IgnacioBecerra:
### Related Ticket(s)
#6409

### Description
Removed the interactive `useMockData` knob as well as the `platform` knob for the Default and Custom Navigation story.

### Changelog
**Removed**

- removed Masthead knobs for `useMockData` and `platform` for certain stories.


ariellalgilmore (approved):
LGTM!

photodow:
Thanks for this @IgnacioBecerra. Here are a couple things I noticed...

- [ ] on load in react is two words and one word in web components. Looks like olivia suggested it as one word.
- [ ] Is Olivia suggesting we remove the L1 story? Or not touch it?
- [ ] platform name appears to still be there in react.
- [ ] Do we need the repetitive knob labels? (e.g. `(selected-menu-item)`)

IgnacioBecerra:
@photodow thanks for the review! Going by the issue, I don't think there's anything to clean in the L1 story. 

As for the platform, seems like we're only getting rid of the knob, not the story itself?

And as for the repeated knob labels, as those are also currently in Web Components, I believe we have them so adopters know which attributes to use to modify the Masthead accordingly. I can remove them from React in this PR if desired.

photodow (approved):

ariellalgilmore:
@IgnacioBecerra looks like just need to update snapshots

Issue #5651, opened 2021-03-29
[Card] storybook updates


oliviaflory:
### Card - default storybook entry:

The pink box highlights the knobs that need update.

![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/8c076af5-5401-4b88-825a-b5fb71231566)

### Card - pictogram storybook entry: 

![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/b4d9565a-6f31-4542-89bc-545ac5ac9ee6)

### Card - static storybook entry: 

Content on the card should change to below to be accurate to the intended usage of this card: 

| Field | Content | 
| --- | --- |
| Eyebrow text | SPSS Statistics |
| Heading | Free trial |
| Copy | Enjoy full SPSS Statistics capabilities including all add-ons. All trial registrants are restricted to one free trial per computer per year. |

![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/089d6e20-8fb3-4a64-8b9a-f4667b1a600f)

## Additional information
- https://ibmdotcom-web-components.mybluemix.net/?path=/story/components-card--default
- https://ibmdotcom-react.mybluemix.net/?path=/story/components-card--default

shixiedesign:
Noticed recently in canary many knobs under `Card - default` disappeared. Can we restore the knobs so I can test the cases where some optional texts are absent? Left side is Web components, right side React. Reference Olivia's issue description above for the expected list of knobs: 

![image](https://user-images.githubusercontent.com/15144993/115052518-062b5a00-9eac-11eb-87f0-c8fe4b7ba379.png)

shixiedesign:
This is still an issue to address. Design need to look through the issue description to make sure it's still up to date? One thing I noticed we might need to add is `Outlined card`.

PR #7665, opened 2021-11-15
fix(card): update storybook knobs

ariellalgilmore:
### Related Ticket(s)

#5651

### Description

Update Web components, React, and React wrapper stories. Changed knobs to be updated to design.

### Changelog

**Changed**

- cleaned up card storybook


photodow:
This is absolutely fantastic. Thank you for this! Here are a couple things I noticed.

Card default
- [ ] Small thing, but can we use the same image across the libraries?
- [ ] Outline doesn't appear to be working correctly in the React wrapper.
- [ ] Also seeing some link color difference in the React wrapper inverse?
- [ ] Can we keep the order of knobs the same? React looks like outline card got moved to the top.
- [ ] I wonder what Shixie or Olivia would think, but I wonder if we could combine outline and inverse into like a style dropdown? I don't think we want to demo those two together? Just thinking out loud here.

Card pictogram
- [ ] How are we missing this in React? 😱
- [ ] Isn't copy supposed to be present on both top or bottom? (have we not made that update yet?)
- [ ] same suggestion on styles inverse and outline

Card static
- [ ] Image has a hover state
- [ ] The card in react in general has a hover state
- [ ] Looks like the issue is asking for a dynamic CTA toggle. When on the input copy field comes back? Wonder if we should add this to all variations actually.
- [ ] Are we missing tags in react?

photodow (approved):
- [ ] Card static still has hover states in React
- [ ] Tags appear to be missing in the React version
- [ ] Might be good to still make the CTA copy dynamic across all card stories, or at least the default. similar to how you did it in static.

Other wise looks great.

jeffchew (approved):
LGTM!

jeffchew:
@ariellalgilmore did changes to storybook impact the e2e tests?

image

Issue #7580, opened 2021-11-03
Safari baseline: [CTA - Card link, Card section simple, card group, card section, card in card] arrow position moved up - No DDS consulting


shixiedesign:
**Engineering info:**  
- HC engineer: @proeung 
- HC JIRA Ticket:  https://jsw.ibm.com/browse/HC-2311
- DDS consulting engineer: No consultant

----

Similar problem observed in
- CTA - Card link
- Card section simple: Default
- Card section simple: With CTA
- Card section images: Default
- Card section offset: Default
- Card in card: Default
- Card group: With card links

![Nov-03-2021 02-53-26.gif](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/ad9a2307-340c-405f-b9ec-6a72771997f1)

PR #7672, opened 2021-11-16
fix(card-cta-footer): adjust arrow position inconsistency seen in safari

proeung:
### Related Ticket(s)

DDS Consulting Issue: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7580
HC JIRA Ticket: https://jsw.ibm.com/browse/HC-2311

### Description
- This PR fixes the arrow position inconsistency seen in Safari for the following components.
  - CTA - Card link
  - Card section simple: Default
  - Card section simple: With CTA
  - Card section images: Default
  - Card section offset: Default
  - Card in card: Default
  - Card group: With card links


annawen1 (approved):
LGTM code-wise!

shixiedesign:
Still seeing around 2 px difference, on icon position and card height. If all implementation is correct, and we didn't miss anything, I don't think it's necessary to fix that difference, but want to point it out in case this is something unintended. 

Safari icon is 2px higher than Chrome, and overall card is shorter on Safari as well: 

image

Nov-17-2021 16-19-20

proeung:
@shixiedesign I checked the standalone components (eg. CTA - Card link) in Percy and I'm not seeing the 2px difference, however, upon checking the Card Group component, I noticed that there's a height calculation is being added to the `<dds-card-cta-footer>` within the Card group's `setSameHeight` function. 

Safari seems to output a different height value from Chrome. I'll take a look deeper look at what's causing this, but do you know why we're defining the same height for the card footer?

Screen Shot 2021-11-17 at 5 28 05 PM

shixiedesign:
Hi @proeung yes the 2px only happens in the card groups, and I'm not sure why same height util is applied to footer. Could it be related to when tags are added? (See below second magenta line above tags) It's also possible that the same height on footer is simply not necessary to achieve that: 

image

proeung:
@shixiedesign Looking at the `card-group.ts`, I don't see the `sameHeight()` being applied to the Tag Group. Right now, this function is being defined for `heading`, `eyebrow`, and `CTA footer`. 

In order to fix this height calculation inconsistency, we'll need to modify the `sameHeight` function within the Utilities package, unless we decided that applying this height to the CTA footer is not needed for the Card Group. @annawen1 Let me know what you think as well.

proeung:
@shixiedesign @annawen1 Since I haven't gotten a response regarding the question above, I went ahead and modified the `sameHeight.js` utility function to use `clientHeight` because `offsetHeight` doesn't offer a standardized value across browsers (see - https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/7672/commits/47264df240d8c02ab911062a2d64c593cbe83a3c#diff-66d8a0f970b5f541fef9c17bc32a11b6aace0fdb82237247925f98ac07266e13R40).

With this change, the height value is somewhat the same, but with a `1px` difference in Safari vs Chrome. Let me know what you all think. cc: @jeffchew

shixiedesign:
Thanks Putra! I think that's a good idea, safer than removing it from footer. I'll go ahead and approve. Thanks!

shixiedesign (approved):
🥇

proeung:
@shixiedesign Thanks! 👍

Issue #5644, opened 2021-03-26
[Table of contents] storybook updates


oliviaflory:
## Update ToC storybook

<img width="899" alt="Screen Shot 2021-03-26 at 4 27 48 PM" src="https://user-images.githubusercontent.com/45692622/112689172-4ae54780-8e50-11eb-836d-24a77a18293e.png">

### Variations
Only two variations of Table of contents:
- Vertical (default)
- Horizontal
- Remove "With heading content" as its own story, it should be a knob under Vertical (see below)

### Vertical knobs
**1. With heading content**
- remove image example
- use link list like in the design specs
 | ![Screen Shot 2021-03-15 at 3.34.11 PM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/e016bf3b-0e8d-4d12-bfdc-8b37981aa141) | ![Screen Shot 2021-03-15 at 3.43.04 PM.png](https://images.zenhubusercontent.com/5cacfd2d41f23a1e95e068dc/fdccb286-4304-4057-a44c-fb7e3c15cc55) |
- rule is added by default when heading content is included, see functional specs
![image (1)](https://user-images.githubusercontent.com/45692622/112689554-d8289c00-8e50-11eb-95de-8491dda58ff9.png)
- this can be a checkbox

**2. Number of items** 
- Default to 5
- Dropdown option of 4-8

**3. Content**
_It would be nice if we could allow users to include their own labels/content into the demo_

### Horizontal knobs
**1. Number of items** 
- Default to 5
- Dropdown option of 4-8

**2. Content**
_It would be nice if we could allow users to include their own labels/content into the demo_

## Additional information
React only has vertical as of now (March 26), TBD if horizontal variation is added as a React wrapper or React.
- If added as a wrapper, could React storybook link to React wrapper?

React also has options for `Dynamic items` and `Manually define menu items` 
- not sure if those options are built into Web components
- Raph was looking into if we needed to have both options and if we should show in storybook?

#### Acceptance criteria
 - [ ] Answer/confirm questions in Additional information to determine if they need to be included in the stories 
- [ ] Web components storybook is updated to reflect above changes
 - [ ] React storybook is updated to reflect above changes

emyarod:
[...]

PR #7676, opened 2021-11-16
docs(table-of-contents): add dynamic copy knobs

emyarod:
### Related Ticket(s)

#5644

### Description

This PR updates the web components, React wrapper, and React storybook Table of Contents stories according to match the expected content described in #5644:

* only "Default" (vertical) and "Horizontal" stories remain
* "With heading content" knob in default story toggles a link list header instead of an image
image

* the number of items is limited from 4-8 and the copy can be customized. This does not work properly in the web components storybook UI so you will need to view the story in a separate tab. This will work in the React storybook after #7687 is merged

### Changelog

**New**

- dynamic content knobs (the table of contents component does not update when its child elements change though, need to view story in new tab for updates)
- "Horizontal" story in React
- "Default" story in React

**Changed**

- "Default" story in web components and react wrapper

**Removed**

- obsolete React stories
- "With heading content" stories in all storybooks

### DIFF
--- a/packages/react/src/components/LinkList/__stories__/LinkList.stories.js
+++ b/packages/react/src/components/LinkList/__stories__/LinkList.stories.js
@@ -239,7 +239,7 @@ EndOfSection.story = {
     knobs: {
       LinkList: ({ groupId }) => ({
         heading: text('Heading (heading):', 'Tutorials', groupId),
-        items: items,
+        items,
       }),
     },
     propsSet: {
--- a/packages/react/src/components/TableOfContents/__stories__/TableOfContents.stories.js
+++ b/packages/react/src/components/TableOfContents/__stories__/TableOfContents.stories.js
@@ -105,6 +105,7 @@ Horizontal.story = {
   name: 'Horizontal',
   parameters: {
     ...readme.parameters,
+    knobs: null,
     percy: {
       skip: true,
     },
--- a/packages/react/src/components/TableOfContents/__stories__/TableOfContents.stories.js
+++ b/packages/react/src/components/TableOfContents/__stories__/TableOfContents.stories.js
@@ -126,3 +126,28 @@ WithHeadingContent.story = {
     },
   },
 };
+
+export const Horizontal = () => (
+  <p>
+    This component is maintained in{' '}
+    <code>@carbon/ibmdotcom-web-components</code> library with a{' '}
+    <a
+      className="bx--link"
+      target="_blank"
+      href="https://www.ibm.com/standards/carbon/web-components/react/?path=/story/components-table-of-contents--horizontal">
+      React wrapper
+    </a>
+    .
+  </p>
+);
+
+Horizontal.story = {
+  name: 'Horizontal',
+  parameters: {
+    ...readme.parameters,
+    percy: {
+      skip: true,
+    },
+    proxy: true,
+  },
+};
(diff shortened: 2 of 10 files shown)


photodow:
Here is a just a couple high level ones I'm seeing.

- [ ] React has a additional spacing



- [ ] The number input doesn't appear to be working in any of the libraries. Am I doing something wrong? It's also not clear what this number of items is supposed to change.
- [ ] The issue appears to be asking for a dropdown for the number of items.
- [ ] The content doesn't appear to update in the web components. Is this the storybook issue? Should we maybe ignore it for web components?
- [ ] In React the headings don't update the link text in the table of contents?
- [ ] I don't think we want the heading toggle in the horizontal stories.

emyarod:
@photodow the React version is directly pulling in the React link list. should the styles be changed there as well or does it only apply to the header content in ToC?

image


as for the knobs, this is a bug in storybook from what I can tell (documented in the issue comments). To see your changes you will have to open the story in a new tab (in the upper right corner of the story)

image

emyarod:
@photodow the React version is directly pulling in the React link list. should the styles be changed there as well or does it only apply to the header content in ToC?

image


as for the knobs, this is a bug in the web components storybook from what I can tell (documented in the issue comments). To see your changes you will have to open the story in a new tab (in the upper right corner of the story)

image

in the React storybook this should be resolved by https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/7687

photodow:
@emyarod the components should work the same in React, vs Web Components. So the fact that we are seeing different spacing in one is a bit of a problem. Whether that spacing is applied to the component or it's parent is mostly up to y'all (or has been in the past) I definitely have my opinions, and the specs outline design's thoughts as well. The issue here is more of the inconsistency, and I don't feel comfortable pushing out something like this to production as an example of how adopters and designers should use our components.

emyarod:
@photodow yeah I think the web components and React version should be consistent, just want to make sure before I made the changes since it looks like the React Link List got extra padding from this PR https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/1688. I think it's fine to revert this, right?

photodow:
- [ ] Something is still not quite right with that spacing.

###### Web components
image

###### React
image

- [ ] React wrapper has number input instead of drop down
- [ ] It looks like the `with heading content` is still a knob in the horizontal stories across the board.
   - [ ] Does the react horizontal need any knobs?

emyarod:
@photodow I believe the spacing issue is resolved now and all of the corrections have been made for the knobs in all 3 storybooks!

photodow (approved):
I'm going to go ahead and approve this with one caveat I noticed.

There is a problem where it looks like if I open the canvas in a new tab after making changes it won't load the page, but if I load it with no changes applied it works. I'm wondering if the URL is too long with the amount of content we have in there?

emyarod:
I think it might just be an issue with the deploy preview builds but I'll keep an eye on it as it rebuilds

jeffchew (approved):
LGTM!

jeffchew:
@emyarod can you update the snapshot file?

jeffchew:
@emyarod the TOC e2e tests are failing, can you take a look? 

image

emyarod:
it should be resolved now @jeffchew!

jeffchew:
@emyarod snapshots look out of date, can you update?

Issue #7678, opened 2021-11-17
DotcomShell component L1 not working


amanda-mendes-1998:
## Issue description
I added a mastheadL1Data  prop object to my DotcomShell and actually, it works and renders if the dev server is already loading when I add it but if I refresh my page or restart localhost I receive an error and nothing renders

## Screenshots
![image](https://user-images.githubusercontent.com/80002122/142198252-6f279717-c651-4149-94f2-a149d138f435.png)
![image](https://user-images.githubusercontent.com/80002122/142198278-e6a89bb1-e299-4631-b919-8ede52d31032.png)

## Steps to reproduce the issue (if applicable)
- Add the DotcomShell component with the mastheadL1Data and the error should happen when you start localhost

PR #7681, opened 2021-11-17
fix(Masthead): add optional chaining to root.location to remove error in NextJS when rendering server-side

annawen1:
### Related Ticket(s)

DotcomShell component L1 not working #7678

### Description

When setting up DotcomShell with L1 in NextJS template, adopter is running into an error specifically with ` const currentUrlPath = root.location.href;`. Seems like NextJS is trying to load it server-side and throws an error that `root.location` is undefined. Adding optional chaining resolves this issue.

### Changelog

**Changed**

- add optional chaining to `currentUrlPath`


jeffchew (approved):
LGTM!

PR #7687, opened 2021-11-17
fix(TableOfContents): sync props and state for dynamically added items

emyarod:
### Related Ticket(s)

#7676

### Description

This PR updates the React TableOfContents so that when items are dynamically added by the ToC component a rerender is triggered. Previously when dynamically populating the ToC, it would remain in the same state as the initial render

this can be tested in the TableOfContents story from #7676 by adding additional items in the story knobs https://ibmdotcom-react.s3.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/7676/index.html

### Changelog

**New**

- NPE checks on `validateMenuItems`

**Changed**

- watch changes in `children` to update when ToC items are dynamically added/removed


jeffchew (approved):
LGTM!

PR #7690, opened 2021-11-17
feat(masthead): scroll to active menu item on L0

IgnacioBecerra:
### Related Ticket(s)
#6473 

### Description
Upon consulting design team, this PR is a follow up to #7655 to add the same scroll to active item functionality to the L0. 

In order to test this out, go to the default Masthead story, and under selected menu item type `Explore more`, and open the story in a new tab so the active element updates. You will be able to see the new change upon load. It may be necessary to shrink the browser width so the L0 is small enough to need the left/right caret pagination functions.

### Changelog

**New**

- added scroll to active element in the L0 menu


ariellalgilmore (approved):
LGTM! Thank you!

jeffchew:
Looks like snapshot files are still out of sync @IgnacioBecerra

jeffchew:
@IgnacioBecerra I think that some of the e2e tests might have been impacted by changes, can you take a look? 
image

PR #7714, opened 2021-11-19
fix(digitalData): add `globalInit` function to `dotcom-shell-composite`

jeffchew:
### Related Ticket(s)

No related issue

### Description

When running tests on web component pages using the `dotcom-shell`, I noticed that the digitalData object was not properly setting the version. This PR will add the `globalInit` function to fire on the `dotcom-shell-composite`. Right now, it is only configured to run on `masthead-composite` and `footer-composite`. 

I want to test on the downstream test app and see if the change has any effect.

### Changelog

**Changed**

- add `globalInit` in `firstUpdated` lifecycle method in `dotcom-shell-composite.ts`


kennylam (approved):
LGTM thanks @jeffchew!

Issue #7605, opened 2021-11-05
[Carousel]: Carousel scrolls to empty page if clicked a second time - DDS consulting


dratzenboeck-ibm:
**Engineering info:**  
- HC engineer: @jkaeser 
- HC JIRA Ticket: https://jsw.ibm.com/browse/HC-2321
- DDS consulting engineer:  Pending DDS resourcing review

----

### Description

When clicking a jump link or a video link inside the carousel you stay on the same page. After closing the video / scrolling back to the carousel and clicking another link, the carousel scrolls to an empty page. It seems to scroll to a page 0, because the page indicator says "0/3" for example. 

### Component(s) impacted

https://www.ibm.com/standards/carbon/web-components/?path=/story/components-carousel--cards-with-images

https://user-images.githubusercontent.com/85555103/140490627-97d9f445-28a5-4fa3-a708-de72e7832124.mp4

I tried around a bit and it seems to be caused by using the `<dds-video-cta-container>` inside the carousel, but this was suggested to us after this issue https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6339 

### Browser

Chrome

### Carbon for IBM.com version

1.26.0

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

Afterburner Adobe CMS

### Package

@carbon/ibmdotcom-web-components, @carbon/ibmdotcom-styles

### CodeSandbox example

https://codesandbox.io/s/nostalgic-liskov-k2o16?file=/index.html

### Steps to reproduce the issue (if applicable)

1. Click on a video cta or jump cta inside the carousel
2. Go back to the same carousel
3. Click on any other link
4. The link will not work as expected and only scroll the carousel to page 0

_No response_

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

emyarod:
related #6372

PR #7733, opened 2021-11-22
fix(carousel): prevent scrolling to empty page & ensure scrolling to focused item's page

jkaeser:
### Related Ticket(s)

Resolves https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7605

### Description

A bug in the focus handler logic forced the carousel to attempt to scroll to page -1 under certain circumstances. This page doesn't exist, of course, and resulted in the carousel scrolling to a blank page.

While I was working on this handler, I found another bug that prevented the first and last carousel items from being scrolled to when they were not in the active carousel page but received focus. Fixing this was a simple matter of 

### Changelog
**Added**
- Added a Video Card Carousel story.

**Changed**
- Updated `DDS Carousel` focus handler logic to correctly identify index of carousel item that has received focus.

**Removed**
- Removed unnecessary conditional that prevented desirable scroll behavior.


emyarod (approved):
looks good to me!

RichKummer (approved):
Looks like the bug is fixed. TY!

andysherman2121 (approved):
LGTM

Issue #7328, opened 2021-10-06
[Layout]: Deprecation of `Layout`


jeffchew:
### User story

> As a `[user role below]`:

IBM.com developer

> I need to:

be aware that `Layout` and its classes are deprecated

> so that I can:

take action to remove in my application

### Additional information

- Issues blocking this: #7325 #7327 
- Documentation for `Layout` should point to the new mixins for React and Web Components
- Deprecation message should be added to the `Layout` component itself

### Acceptance criteria

- [x] Deprecation notice added to the `Layout` component in React
- [x] Documentation for `Layout` in React and Web Components updated to reference the mixins
- [x] Added deprecation message to `breaking-changes.md` for React and Web Components
- [x] Reach out to AEM team to see if they are still using `Layout`

PR #7747, opened 2021-11-23
chore(Layout): deprecate Layout and update Layout story/readme

annawen1:
### Related Ticket(s)

[Layout]: Deprecation of Layout #7328

### Description

This PR deprecates the React `<Layout />` component. The Layout storybook has been updated to use the layout style classes and mixins instead. 

### Changelog

**New**

- added deprecation message to `breaking-changes.md`

**Changed**

- noticed we were setting the layout according to 12-col instead of 16-col, so set that to corrected that in the `layout` scss file
- removed `<Layout />` usage in Layout React storybook, using the classes instead.
- edited README to display layout classes to use instead


jeffchew:
Thanks @annawen1 ! Looks like the snapshot files are out of date.

jeffchew (approved):
LGTM!

PR #7762, opened 2021-11-24
chore(percy): skip cards with video in percy diffs

jeffchew:
### Related Ticket(s)

No related issue

### Description

This will skip the `Carousel | Cards with Video` story in Percy diff checking, as the async data is causing false positives in reports.

### Changelog

**Changed**

- Add `percy: skip` to the `Carousel | Cards with Video` story


ariellalgilmore (approved):

emyarod (approved):
looks good to me!

PR #7763, opened 2021-11-24
test(e2e): overhaul of snapshot testing for web components and react

jeffchew:
### Related Ticket(s)

No related issue

### Description

Noticed in percy snapshots in our e2e tests that there were duplicate names being set for the snapshots:

image

~~This PR should fix the existing duplicate names, though we should perhaps by practice change all tests to use the variable `titlePath` values so that we won't run into this in the future.~~

I ended up creating a new global method called `cy.takeSnapshots`, and then applied it across all tests. Hopefully this will help to normalize how we do screenshot/snapshot testing going forward.

### Changelog

**Changed**

- Added `cy.takeSnapshots` method, can switch between `desktop` and `mobile` snapshots, and pass additional options
- Replaced all instances of `cy.screenshot()` and `cy.percySnapshot` with `cy.takeSnapshots`
- Replaced all theme tests with `cy.carbonThemesScreenshot()`
- Suppress table of contents tests for scroll due to unstable tests

### DIFF
--- a/packages/react/tests/e2e-storybook/cypress-canary.json
+++ b/packages/react/tests/e2e-storybook/cypress-canary.json
@@ -5,6 +5,6 @@
   "nodeVersion": "system",
   "includeShadowDom": true,
   "testFiles": "**/*.e2e.js",
-  "pageLoadTimeout": 50000,
-  "defaultCommandTimeout": 50000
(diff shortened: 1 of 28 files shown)


ariellalgilmore (approved):
Looks great!! Thank you @jeffchew!

Issue #7713, opened 2021-11-19
[CTA Section with Content Items]: Bottom Padding is not 160px - No DDS consulting


EAVilleda:
###  Engineering info:
- HC engineer: @jkaeser 
- HC Jira Ticket: https://jsw.ibm.com/browse/HC-2327
- DDS Designer: @DragosRistici 

----

### Description

The bottom spacing of the content sections that contain the CTA section component have a bottom spacing of 136px. The bottom spacing should be 160px. However, this spacing does match what is on Storybook. Can the bottom spacing be updated to have 160px? 

### Component(s) impacted

CTA Section with content items

### Browser

Chrome

### Carbon for IBM.com version

v1.26.0

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

AEM (Marketing)

### Package

@carbon/ibmdotcom-web-components, @carbon/ibmdotcom-styles

### CodeSandbox example

https://codesandbox.io/s/github/carbon-design-system/carbon-for-ibm-dotcom/tree/master/packages/web-components/examples/codesandbox/components/cta-section

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

proeung:
There are spacing bugs related to the CTA Section layout component that was mentioned within this ticket (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7482). The engineer that's picking up this issue should account for the discrepancies mentioned in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7482 as well.

Issue #7482, opened 2021-10-21
[CTA section with content items]: Padding in between rows is incorrect - no DDS consulting


jwitkowski79:
###  Engineering info:
- HC engineer: @jkaeser 
- HC Jira Ticket: https://jsw.ibm.com/browse/HC-2327
- DDS Designer: @DragosRistici 

----

### Description

In [storybook](https://www.ibm.com/standards/carbon/web-components/?path=/story/components-cta-section--with-content-items), when more than 3 items are added to the CTA section with content items, the top and bottom padding between the rows is displaying incorrectly. According to the [specs](https://ibm.ent.box.com/folder/133551296345?s=g33w79p1lg33nxdkbf0o36dtyybaxzie), there should be 64px between each row on the max, xlg, lg, and md breakpoints See attached screenshot for specs and screenshot of bug
![Components___CTA_section_-_With_content_items_⋅_Storybook](https://user-images.githubusercontent.com/2040791/138338135-09cb4792-1665-4976-92d6-2b221bee8e84.png)
![Components___CTA_section_-_With_content_items_specs](https://user-images.githubusercontent.com/2040791/138338137-722b6b9c-df6b-41ad-98fc-e3806ce079c8.jpg)

### Component(s) impacted

[CTA Section with content items](https://www.ibm.com/standards/carbon/web-components/?path=/story/components-cta-section--with-content-items). Need to add more than 3 content items in order to see the bug

### Browser

Chrome

### Carbon for IBM.com version

v1.11

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

AEM Product Template Team

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://www.ibm.com/standards/carbon/web-components/?path=/story/components-cta-section--with-content-items

### Steps to reproduce the issue (if applicable)

Visit [Storybook](https://www.ibm.com/standards/carbon/web-components/?path=/story/components-cta-section--with-content-items). Add more than 3 content items to the CTA section with content items to see padding issue

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #7764, opened 2021-11-26
fix(cta-section): use correct bottom spacing and spacing between item rows

jkaeser:
### Related Ticket(s)

- Resolves https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7482
- Resolves https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7713

### Description

Incorrect layout and spacing tokens were used for certain parts of the `<dds-cta-block>` component, visible in the "CTA section with content items" story. This PR's main goal was to correct those spacing issues, but it also fixes a bug that was introduced in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/commit/e2fb14d8541b688a14b9dbf69cf6cf21f848ff3d where the horizontal rules between item rows went missing.

### Changelog

**Added**

- Re-established horizontal rules between `<dds-cta-block-item-row>` rows beyond the `md` breakpoint.

**Changed**

- Adjusted space between children of `<dds-cta-block-item-row>` elements to match latest designs.
- Adjusted `<dds-cta-block>` bottom spacing to match latest designs.


shixiedesign:
Hi @jkaeser thanks for the fix! One clarification. Although the issue is reported in CTA section, we should try to make the fix to the root of the problem which is CTA block. I noticed the issue is fixed in CTA section **but still existing in CTA block**. CTA block's content items are missing the horizontal rule as well as extra spacing. 

Issue 7713 Bottom Padding is not 160px is fixed ✅ 
Issue 7482 [CTA section with content items]: Padding in between rows is incorrect 👈 still needs some fixes

image

jkaeser:
Hi @shixiedesign, and thank you for the feedback! I updated the **CTA Block** story content to match the style modifications I previously made.

---

@jeffchew I'm making a somewhat significant change to the markup pattern in order to align this component to the design specs. I figure it's worth discussing to give folks an opportunity to raise objections, though I'm not entirely sure who to include in the conversation.

### Previous to this PR
Each row is manually curated to include up to three items, and `no-border` is set on the last row. This caused the third item in each row to be isolated at the `md` breakpoint:

```html
<dds-cta-block-item-row>
  <dds-cta-block-item>Item 1</dds-cta-block-item>
  <dds-cta-block-item>Item 2</dds-cta-block-item>
  <dds-cta-block-item>Item 3</dds-cta-block-item>
</dds-cta-block-item-row>
<dds-cta-block-item-row no-border>
  <dds-cta-block-item>Item 4</dds-cta-block-item>
  <dds-cta-block-item>Item 5</dds-cta-block-item>
  <dds-cta-block-item>Item 6</dds-cta-block-item>
</dds-cta-block-item-row>
```

Screen Shot 2021-12-06 at 10 41 19 AM

### After this PR
All items are placed in one row element, and the styles handle arrangement and borders. Items meet the design specs at the `md` breakpoint:

```html
<dds-cta-block-item-row>
  <dds-cta-block-item>Item 1</dds-cta-block-item>
  <dds-cta-block-item>Item 2</dds-cta-block-item>
  <dds-cta-block-item>Item 3</dds-cta-block-item>
  <dds-cta-block-item>Item 4</dds-cta-block-item>
  <dds-cta-block-item>Item 5</dds-cta-block-item>
  <dds-cta-block-item>Item 6</dds-cta-block-item>
</dds-cta-block-item-row>
```

Screen Shot 2021-12-06 at 10 41 29 AM

shixiedesign (approved):
Thank you!

jeffchew:
^^ @emyarod @IgnacioBecerra would you be able to weigh in?

packages/web-components/src/components/cta-block/cta-block.scss
```diff
@@ -103,24 +107,37 @@
...
+      content: '';
+      background-color: $ui-background;
+      height: 3px;
+      position: absolute;
+      bottom: -1px;
+      left: 0;
+      right: 0;
+      pointer-events: none;
     }
   }
```
emyarod, line 125: just wondering, when is this pseudoelement visible? it seems like it's always hidden when it gets rendered to the DOM

image

emyarod (commented):
just have question about the positioning rules on the cta block item row!

packages/web-components/src/components/cta-block/cta-block.scss
jkaeser: There should only be borders between each row, and this pseudo element serves that goal by hiding the last row's items' bottom borders. There is a darker bottom border in the design specs, but my understanding is that comes from `.bx--cta-block__border`, which can be disabled by adding the `no-border` attribute to `<dds-cta-block>`. I just pushed a commit adding a knob to the CTA Block story so this can be seen.

emyarod (approved):
looks good to me!

PR #7769, opened 2021-11-29
fix(carousel): add null check for carousel same height method

jeffchew:
### Related Ticket(s)

No related issue

### Description

We were occasionally getting exception errors on a NextJS page that loaded in the carousel, which seems to be stemming from the internal `_setSameHeight` method:

image

image

image

Adding a null check in this method will hopefully fix the issues here.

### Changelog

**Changed**

- Added optional chaining in `_setSameHeight` method in `carousel.ts`


IgnacioBecerra (approved):
nice fix @jeffchew, thanks for adding the null check!

PR #7748, opened 2021-11-24
test(card): add default e2e storybook tests

ariellalgilmore:
### Related Ticket(s)

#7449

### Description

Added Card default tests. Checking to make sure all the text renders properly and clickable, all the different storybook knobs (tags, image, outline, inverse) work properly, and checks the different theming changes the card.

### Changelog

**New**

- card e2e tests
- testing for eyebrow, heading, copy, footer text
- entire card is clickable
- tags appear 
- image appear
- outline classes applied with correct css
- inverse classes applied with correct css

**Changed**

- removed card-static test files to condense to one card e2e test

Issue #7450, opened 2021-10-20
[Test Scenario]: Card - pictogram


Praveen-sr:
### Component and test scenario title

Card - pictogram

### Test scenario steps

a. Check card, pictogram is loaded and card is clickable
b. Verify pictogram placement (Top/bottom)  from knobs
c. Verify on bottom placement, Tags and copy text is visible.
d. Customize theme from Carbon theme and verify gray 10, gray 90 and gray 100 theme.

PR #7775, opened 2021-11-29
test(card-pictogram): added e2e tests

IgnacioBecerra:
### Related Ticket(s)
#7450

### Description
Created e2e tests to check for pictogram placement, check if copy will display upon hover, and theme snapshots. 
Note: the pictogram variation is Web Components only, so there's no React version.

To keep all Card related tests together in one file, #7748 should be merged first so we can incorporate these new tests into the single `Card` file in this PR to avoid any big merge conflicts.

### Changelog

**New**

- new e2e tests for card pictogram


emyarod (approved):
looks good to me!

Issue #6407, opened 2021-06-17
[Masthead] Web components storybook demo clean up


oliviaflory:
## Detailed description

> Describe in detail the issue you're having.

Masthead storybook clean up, React issue #6409 

### Default
Knobs:
- [x] no platform knob, default global nav should not enable a platform name

Storybook example is good as-is, update knobs per above
![masthead-global](https://user-images.githubusercontent.com/45692622/124154336-666d7900-da63-11eb-8ac0-4bf4ac7190a7.jpg)

### ~With custom navigation~ hide/remove story
- [x] We are proposing to hide this story, in most cases, IA guild does not recommend teams have custom navigation within the L0. But we do not want to deprecate yet as it would be a breaking change.
- [ ] ~no platform knob, default global nav should not enable a platform name~
- [ ] ~Label should be sentence case: `With custom navigation` (not initial caps)~
- [ ] ~Custom navigation should **not** have mega menu styling, remove all mega menu styling and use dropdown. May have to change data sets?~

![Screen Shot 2021-06-17 at 3 50 31 PM](https://user-images.githubusercontent.com/45692622/122464757-3a42fa00-cf85-11eb-8ef7-452c471f37c1.png)
![Screen Shot 2021-06-17 at 3 51 02 PM](https://user-images.githubusercontent.com/45692622/122464759-3adb9080-cf85-11eb-9e5d-e4776009036f.png)

### Search Open Onload
- [x] label should be in sentence case, not initial caps `Search open onload`
- [x] hide platform knob

Storybook demo proposal:
![masthead-Search-open-onload](https://user-images.githubusercontent.com/45692622/124154819-dbd94980-da63-11eb-8411-21efd8cc2e2c.jpg)

### With Platform
- [x] label should be in sentence case, not initial caps `With platform`
- [ ] ~All nav items within the With platform story should **not** have mega menu styling, remove all mega menu styling and use dropdown. May have to swap data sets?~
- [x] Remove all links from With platform story other than the Platform
- [x] Update "Cloud" to "Platform"
![Screen Shot 2021-06-17 at 4 01 35 PM](https://user-images.githubusercontent.com/45692622/122464901-5fd00380-cf85-11eb-80b3-5924dca29c80.png)

Story book demo proposal: 
![masthead-with-platform](https://user-images.githubusercontent.com/45692622/124155222-425e6780-da64-11eb-9dd4-48e089eb8b1a.jpg)

### With L1
[...]

PR #7779, opened 2021-11-30
docs(masthead): storybook demo cleanup

emyarod:
### Related Ticket(s)

#6407

### Description

This PR updates all of the masthead stories according to the changes outlined in #6407.

For the "With custom navigation" story, should we just remove it instead of hiding it? Given that it is a storybook-only change I don't think it would be a breaking change even though the ticket states otherwise, right?


packages/web-components/src/components/masthead/__stories__/masthead.stories.ts
```diff
@@ -200,16 +205,14 @@ export const withPlatform = ({ parameters }) => {
           ></dds-masthead-composite>
         `
       : html`
-          <dds-masthead-container
-            platform="${ifNonNull(platformData.name)}"
+          <dds-masthead-composite
```
IgnacioBecerra, line 208: Just wondering the reason behind the change here?

IgnacioBecerra (commented):
Overall looks fine, I also think removing the story would be better since it's not a breaking change.

packages/web-components/src/components/masthead/__stories__/masthead.stories.ts
emyarod: this was to fulfill the "Remove all links from With platform story other than the Platform" requirement in the original issue. From what I could tell, the container will always render the links https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/packages/web-components/src/components/masthead/masthead-container.ts#L86

kennylam (approved):
LGTM thanks @emyarod!

oliviaflory:
Hey @emyarod 
I saw your note about deleting the `With custom navigation` story, makes sense that it's not a breaking change being only in storybook! 

This looks really good, I noticed two things:

1. There's a bit of an odd thing happening with the recent update to the Global nav label from Services & Consulting to Consulting & Services. The Consulting & Services only shows as selected if you type "Services & Consulting", but not if you type "Consulting & Services". 
 
I'm not 100% sure why that is, but guessing it has something to do with that label originally being "Services & Consulting"
Screen Shot 2021-12-07 at 3 48 41 PM
Screen Shot 2021-12-07 at 3 48 56 PM

2. I missed requesting to hide the Platform knob on the `With alternate logo and tooltip` story. Is it difficult to remove that too?
Screen Shot 2021-12-07 at 4 18 04 PM

I saw Shixie asked in the original issue if updating the e2e tests would be part of this, let us know if we need a separate issue to cover that 🙂

emyarod:
opened https://github.ibm.com/webstandards/ibm-dotcom-translations/pull/544 to address the selected menu highlight mismatch

emyarod:
@oliviaflory the platform knob should now be gone from that story, and the issue with the highlighted/selected menu will be resolved in the previous PR I linked. I will update the e2e tests in this PR once all of the visual changes are confirmed to be ok!

oliviaflory (commented):
Looks good @emyarod, thank you!

There are a few unexpected visual difs in the React e2e Percy test, but I don't believe anything in this PR would have caused that.

annawen1 (approved):
LGTM!

annawen1:
@emyarod looks like the e2e tests need to be adjusted, can you take a look?

ariellalgilmore:
Looks like it's still failing now for some reason on the masthead default selected menu item test. It passes though for the L1...

emyarod:
yeah that's really strange, even when I view the cypress storybook locally the menu item is properly highlighted for both of the stories with and without L1. I'm not sure why it fails when running headlessly though

emyarod:
I'm not able to approve the percy builds but the changes look good to me

Issue #7439, opened 2021-10-19
[CTA section] Content item should not allow Button as a CTA type


shixiedesign:
Expected content item types are only Text, Stat, Picto, Media. No buttons in this component.
Design spec: https://ibm.box.com/s/g33w79p1lg33nxdkbf0o36dtyybaxzie

![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/99b62f33-b767-46e0-b13c-1bd2db859021)

PR #7782, opened 2021-11-30
fix(cta-section): ensure buttons are not allowed as cta types

IgnacioBecerra:
### Related Ticket(s)
#7439

### Description
Currently the CTA Section story allowed Buttons as CTA types for the content items. With this new PR, I removed the ability to choose Buttons from the knob, as well as ensure that any slotted Button Group does not render if within a `cta-block-item` element.

### Changelog

**New**

- `ButtonGroup` will not rendered if used inside a `cta-block-item` element

**Removed**
- buttons option in CTA type knob for `CTASection`


packages/web-components/src/components/cta-block/cta-block-item.ts
```diff
@@ -60,6 +60,11 @@ class DDSCTABlockItem extends StableSelectorMixin(DDSContentItem) {
     const { name } = target as HTMLSlotElement;
     const hasContent = (target as HTMLSlotElement)
       .assignedNodes()
+      .filter(elem =>
+        (elem as HTMLElement).matches !== undefined
+          ? !(elem as HTMLElement).matches((this.constructor as typeof DDSCTABlockItem).selectorButtonGroup)
+          : false
+      )
```
emyarod, line 67:
```suggestion
      .filter(elem => !(elem as HTMLElement).matches?.((this.constructor as typeof DDSCTABlockItem).selectorButtonGroup))
```

instead of checking if `elem.matches()` is defined would we be able to make use of optional chaining somehow?

emyarod (commented):
looks good to me, just one question about the filtering logic!

packages/web-components/src/components/cta-block/cta-block-item.ts
IgnacioBecerra: You're right, that method works and looks nicer as well. I've noticed some files still use the current logic, so I might open a new PR updating those as well. Thanks @emyarod!

ariellalgilmore (approved):
LGTM!

PR #7786, opened 2021-11-30
chore(card): update wrapper story and docs with card logo

IgnacioBecerra:
### Related Ticket(s)
#7140

### Description
Updated the React Wrapper's Card story to include the Card logo variation, as well as update the docs to include an example of it.

### Changelog

**New**

- new card logo story in react wrapper storybook


emyarod (approved):
looks good to me!

packages/web-components/src/components/card/__stories__/card.stories.react.tsx
```diff
@@ -187,6 +189,83 @@ Static.story = {
...
+          copy,
+          tagGroup,
+          cta,
+          ctaCopy,
+          outlinedCard,
+        };
+      },
+    },
+  },
+};
```
ariellalgilmore, line 225: looks like this is a duplicate from above and might be causing the static [deploy preview](https://ibmdotcom-web-components-react.s3-web.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/7786/index.html?path=/story/components-card--static) for react wrapper to look like this: 
Screen Shot 2021-12-02 at 10 32 06 AM

shixiedesign (approved):
👍

PR #7787, opened 2021-11-30
chore(card): cleanup card e2e storybook tests

ariellalgilmore:
### Related Ticket(s)

none 

### Description

Cleanup card e2e tests into one file for web components. Also, call different tests within each describe and just change the `path` depending on which knob being tested. 

### Changelog

**Changed**

- clean up react and web components e2e tests


packages/web-components/tests/e2e-storybook/cypress/integration/card/card.e2e.js
```diff
@@ -7,6 +7,13 @@
 
 'use strict';
 
+import getCssPropertyForRule from '../../utils/get-css-property-for-rule';
+import ddsSettings from '@carbon/ibmdotcom-utilities/es/utilities/settings/settings.js';
+import settings from 'carbon-components/es/globals/js/settings';
+
+const { stablePrefix: ddsPrefix } = ddsSettings;
+const { prefix } = settings;
```
jeffchew, line 15: These imports will fail in browserstack, hardcode the prefixes for now. I'm looking into setting up global variables in Cypress which we can use instead.

jeffchew (approved):
LGTM!

PR #7652, opened 2021-11-12
feat(KalturaAPI): services listeners-exposure and extra metrics

tweenn:
### Related Ticket(s)

[#3716 - [AnalyticsAPI] Add Video Progress and Video seek tag generation using the ibmstats.event function](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/3716)
[#7488 - Kaltura Player Metrics: For CTA clicks (& other events) - DDS Consulting](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7488)

### Description

First I want to point out that there shouldn't be a breaking change in this PR - except by the error addressed in 34cc4387c83bd83c3883bfa29e800458fb4f3d49 (more on changelogs)

Second, this PR still need improvement and a review by a DBDM expert (more on changelogs)

Finally, this PR addresses 3 major changes in the kaltura-service - and analytics-video portion:

1. Make IBM Metrics firing optional - but turned on by default
2. Expose the readyCallback for custom manipulations of the player and their event-listeners
3. Add extra event listeners for IBM Metrics - like Mediacenter's cta-clicks and cta-impressions, and quartile and seek events from [#3716](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/3716)

It also has some unit-testing and coverage bumping, a bug fix in the analytics file (found while writing tests).

And I tried my best to make each commit be as semantic as possible with the code implemented - meaning the changelog may feel redundant.
 
### Changelog

* 3ac850d1ae768ec855915f3dc4c2efde5122c53a
Added namespaces to the player event-listeners.
This will keep IBM listeners in a safe position if a developer try to remove/unbind listeners, as like using a `kdp.removeListener('seek')` will remove all global/generic `'seek'` listeners, but will not remove the `'seek.ibm'` listener.

* ff7b5211cea4f4d38d8b8988e46edf4c74b2bd47
Added a param to `embedMedia` to make the IBM Metrics listeners optional - turned on by default
And fixed a documentation error where flashvars was set to boolean, when it was actually an object

* 34cc4387c83bd83c3883bfa29e800458fb4f3d49
Added a param to `embedMedia` that is a function that receives the `KDP` object in order to expose the `readyCallback`.

* ad9d2a1704aa21085e156964bcbdb819cc4dde92
[...]

Issue #7488, opened 2021-10-22
Kaltura Player Metrics: For CTA clicks (& other events) - DDS Consulting


tweenn:
### The problem

As we spoke in the office hours, C4IDC don't have a listener to track call-to-action (CTA) clicks.

I'm creating this feature-request as we agreed in the call in order to track and better discuss how to correctly add as optional, but turned on by default, the ibm metrics events for the already existing events and newer ones, like the CTA clicks and other metrics like https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/3716

Plus to this, add a way to expose the kaltura-player listeners in order to allow future customizations of the player - as seen in the audio player.

### The solution

In KalturaPlayerAPI.embedMedia, have two extra parameters.
- "turnOnMetrics" or "activateMetrics"
bool = true
Add all default metric-events from IBM if true

- "customReadyCallback" or any other appropriate nomeclature
undefined or function
If function defined, execute said function with the KDP (kaltura digital player) as the unique parameter, so any one can use "anything that is kdp.something" that exists in the [Kaltura Player Documentation](http://player.kaltura.com/docs/api).
Like manually creating their own CTA system or creating a controller that abstract the player (like C4IDC Audio Player Component)

### Application/website

https://mediacenter.ibm.com

### Business priority

Medium Priority = upcoming release but is not pressing

### What time frame would this ideally be needed by (if applicable)

Best to have it as soon as possible, but not in a hurry for pressing a PR acceptance for yesterday.

### Examples

[CTA event listener in v18 as an example](https://github.ibm.com/webstandards/v18/blob/dev/src/js/sources/ibm/common/widgets/videoplayer-kaltura.js#L613)

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #7789, opened 2021-12-01
feat(KalturaAPI): added ibm namespace to event-listeners

tweenn:
### Related Ticket(s)

[#7488 - Kaltura Player Metrics: For CTA clicks (& other events) - DDS Consulting](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7488)
[#7652 - feat(KalturaAPI): services listeners-exposure and extra metrics](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/7652)

### Description

I'm opening this PR as a follow-up from [#7652](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/7652) that had an issue with extra metrics for quartiles and seek (drag and drop the scrubber).

Said that, I want to point out that there shouldn't be a breaking change in this PR - except by the error addressed in 388d15e3b4047a8a9d2d4df3b5a844c1c2f23a05 (more on changelogs)

And this PR addresses 3 (major but not breaking) changes in the kaltura-service - and analytics-video portion:

1. Make IBM Metrics firing optional - but turned on by default
2. Expose the readyCallback for custom manipulations of the player and their event-listeners
3. Add extra event listener for Mediacenter's cta-clicks and cta-impressions

It also has some unit-testing and coverage bumping, a bug fix in the analytics file (found while writing tests).

And I tried my best to make each commit be as semantic as possible with the code implemented - meaning the changelog may feel redundant.
 
### Changelog

* 52dc06812bb236bef1e45b6ebd7d4c257b44d78d
Added namespaces to the player event-listeners.
This will keep IBM listeners in a safe position if a developer try to remove/unbind listeners, as like using a `kdp.removeListener('seek')` will remove all global/generic `'seek'` listeners, but will not remove the `'seek.ibm'` listener.

* 3c7c1835be59637153e73804e68683ebadb81d06
Added a param to `embedMedia` to make the IBM Metrics listeners optional - turned on by default
And fixed a documentation error where flashvars was set to boolean, when it was actually an object

* 1ddf75a61a2e884cd3a369143d65f17bea74d721
Added a param to `embedMedia` that is a function that receives the `KDP` object in order to expose the `readyCallback`.

* 388d15e3b4047a8a9d2d4df3b5a844c1c2f23a05
The `AnalyticsAPI.videoPlayerStats` prop `executionPath` was expecting a `params.videoId`, that the old `VideoPlayer` service sent, but since the deprecation of that service in favor of `KalturaPlayer`, this param was changed/renamed to `mediaId` - and thus the metrics being fired right now are basically sending `{ executionPath: undefined }` to DBDM

* 69ab26a0c98ff7460f21adc6038ca7dd59db03e7
Added Mediacenter's CTA-CLICK listener and custom metric data to `KalturaPlayer` and the appropriate behaviour over the `AnalyticsAPI.videoPlayerStats` to add custom metrics if available.

* 74f2f2ce9d2c0620f9a5f47a3875d60063ece097 & 8c05d04434fda0fb14d716fd003f04d2662bbb10
Test and Coverage bumping - because I didn't like the values I saw 🤣 

* fb7109de74fbe0590141346a43d89621a0cbea26
Add retrocompatibility for the fix done in 388d15e3b4047a8a9d2d4df3b5a844c1c2f23a05 - which would insert a breaking change otherwise - plus I've added a test case in Analytics's test scenarios for this change AND cta-clicks

-----------


tweenn:
@jeffchew asked in the previous PR #7652 [here](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/7652#pullrequestreview-813071455) the following:
> Not sure if we wanted to include as part of this PR or perhaps a separate PR, if I recall we were discussing also creating standard event listeners that any application could also subscribe to if they wanted to monitor these behaviors (e.g. append their own custom tracking triggers), which would be triggered by these Kaltura listeners. I'd be fine with adding this in a separate PR though.

The `readyCallback` method under the `kWidget.embed` will be executed as soon as the player iframe is created, which you can use to set listeners or other player-driven configuration.
So, instead of adding some custom events, I thought on executing a function inside the readyCallback that passes the `kdp` object would allow customization of such execution context without limiting the consumer to some specific events - and that's what commit [#1ddf75a61a2e884cd3a369143d65f17bea74d721](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/commit/1ddf75a61a2e884cd3a369143d65f17bea74d721) is about.

Since the global kWidget object allow the removal of any listener associated to a player, kaltura defines as a best practice the addition of a namespace to each listener so a global removal wouldn't affect our code - which is the change made in [#52dc06812bb236bef1e45b6ebd7d4c257b44d78d](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/7789/commits/52dc06812bb236bef1e45b6ebd7d4c257b44d78d).

And finally, for those consumers who don't want IBM analytics over a Gartner (or other partners) video, I've added a bool (that by default if true) to define if the namespaced events we have for metrics should be added or not to the player.
PS: Gartner videos that IBM buy the rights to stream are kaltura based

In short, we have some scenarios covered like:
* "I don't want IBM's analytics listeners that are on by default, here's my own events"
* "Here's my own events, which will remove one specific listener from IBM and nothing more"

And by the way - this changes allow a great level of refactoring to the audio player component - that sets a custom readyCallback through the global `kWidget.addReadyCallback()` method which is not very memory-friendly

packages/services/src/services/KalturaPlayer/KalturaPlayer.js
```diff
@@ -234,17 +243,33 @@ class KalturaPlayerAPI {
...
+                fireEvent({
+                  playerState: 101,
+                  kdp,
+                  mediaId,
+                  customMetricsData,
+                });
+              });
+            }
+
+            customReadyCallback(kdp);
```
jeffchew, line 272: 👍

packages/services/src/services/KalturaPlayer/KalturaPlayer.js
```diff
@@ -177,7 +177,10 @@ class KalturaPlayerAPI {
    *
    * @param {string} mediaId  The mediaId we're embedding the placeholder for.
    * @param {string} targetId The targetId the ID where we're putting the placeholder.
-   * @param {boolean} flashvars Determine any extra param or plugin for the player.
+   * @param {object} flashvars Determine any extra param or plugin for the player.
+   * @param {boolean} useIbmMetrics Whether or not should IBM Metrics events be fired.
+   * @param {Function} customReadyCallback Determine any extra functions that should be executed
```
jeffchew, line 182: Minor thing, but can add a note that the player object is passed in the custom method in order to add additional custom functionality (e.g. listeners, etc). Is there a link we can send users to for a full API that they can tap into?

packages/services/src/services/Analytics/Analytics.js
```diff
@@ -267,14 +267,26 @@ class AnalyticsAPI {
...
+      eventData.playerStateLabel = data.customMetricsData.playerStateLabel;
+    }
+
+    if (data?.customMetricsData?.driverId) {
+      eventData.driverId = data.customMetricsData.driverId;
+    }
+
+    if (data?.customMetricsData?.targetURL) {
+      eventData.targetURL = data.customMetricsData.targetURL;
+    }
```
emyarod, line 288:
```suggestion
      driverId: data?.customMetricsData?.driverId,
      executionPath: data.videoId || data.mediaId,
      execPathReturnCode: playerState,
      eventVidStatus: data.playerState,
      eventVidTimeStamp: currentTime,
      eventVidLength: duration,
      eventVidPlayed: percentWatched + '%',
      playerStateLabel: data?.customMetricsData?.playerStateLabel,
      targetURL: data?.customMetricsData?.targetURL,
    };
```

would we be able to move the custom metrics data logic into the `eventData` declaration like this?

tweenn: @emyarod - As far as this was implemented on v18 (in last year) and as far as I understood, the 3 params (playerStateLabel, driverId and targetURL) were to be used exclusively with this event firing, not on all of then.

And by declaring the property on object initiating means that it will be fired as `undefined`, while having it under the above if statement, the object will have the property added only if the param is passed.

jeffchew (approved):
The comments I had were pretty minor and think we can just merge as-is, let me know if there was anything else you wanted to have addressed @emyarod

Issue #7553, opened 2021-11-01
[Card - pictogram] Web Component: Change to include mobile accessibility/hover state improvements - DDS consulting


RobertaJHahn:
###  Engineering info:
- HC engineer: @proeung 
- HC Jira Ticket: https://jsw.ibm.com/browse/HC-2328
- DDS consulting engineer: Pending DDS resourcing review
- DDS Designer: Saijan Kuruvila 
- DDS engineering sizing:  5

----

#### User Story

> As a `[user role below]`:
Carbon for ibm.com developer

> I need to:
create/change the `Card - pictogram`

> so that I can:
provide the ibm.com adopter developers components they can use to build ibm.com web pages

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- Create with Shadow DOM and Custom Elements standards
- **See the Epic (#6739 ) for the Design and Functional specs information**
- [Design and func specs Box folder](https://ibm.ent.box.com/folder/95514001950?s=hgvu48r1fv2ga2h0yn3qshrtxjc71vch)
- This work includes the **animation** change on the desktop, motion specs are included in the Box folder, [specific subfolder link here](https://ibm.box.com/s/ykv2ukr3twyfx2nrro3rn7ge4iqoz48x).  
- Web Components Visual QA testing issue (#7554 )
- Web Components Prod QA testing issue (#7555 )

#### Acceptance criteria
- [ ] Include README for the web component and corresponding styles
- [ ] Create Web Components styles in styles package
- [ ] Do not create knobs in Storybook that include JSON objects
- [ ] Break out Storybook stories into multiple variation stories, if applicable
- [ ] Create codesandbox example under `/packages/web-components/examples/codesandbox` and include in README
- [ ] Minimum 80% unit test coverage
- [ ] Update the Carbon for ibm.com website component [data file](https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/blob/master/src/data/components.json) to be sure the web site Component Status and Storybook links are correct
- [ ] Use the [Visual QA checklist](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/wiki/Definition-of-done-(Visual-QA-checklist)) to verify design quality
- [ ] If a design is provided, the Designer is included as a Reviewer in the Pull Request
- [ ] Provide a direct link to the deploy preview for the designer in the Pull Request description
- [ ] A comment is posted in the Prod QA issue, tagging Praveen when development is finished

[...]

Issue #6739, opened 2021-07-27
[Card - pictogram] Design updates to improve mobile accessibility 


shixiedesign:
### Problem

[**Card - pictogram** component ](https://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-card--pictogram) has a few issues: 
- When pictogram is bottom, a hover interaction reveals additional body copy. On mobile (no hover state), these additional body copy is never readable. We should have a new mobile design so the body copy is always revealed
- When the body copy and card headline are long, hover interaction makes the card grow taller. This is probably because the card size is not calculated with those elements present at first. Might need a refactor? 

### Task 
1. Look into additional of the designs for mobile design 
2. Create logic for how the content edge cases should be handled

### Additional information
- Design spec: https://ibm.ent.box.com/folder/114042319288
- Functional spec: https://ibm.box.com/s/hgvu48r1fv2ga2h0yn3qshrtxjc71vch

RobertaJHahn:
React version is behind a feature flag.  React component will not be updated.  No issues needed.
https://ibmdotcom-react-experimental.mybluemix.net/iframe.html?id=components-card--default&knob-Show%20pictogram%20(hides%20CTA)_Card=true&knob-Card%20Eyebrow_Card=Eyebrow%20text&knob-Card%20Heading_Card=Lorem%20ipsum%20dolor%20sit%20amet&knob-Card%20body%20copy_Card=&knob-CTA%20copy_Card=Card%20CTA%20text

RobertaJHahn:
Jeff will include agenda item in the engineering Huddle to talk about deprecating the Pure React version of Card - pictogram component.

shixiedesign:
Re: deprecating Pure react card pictogram @jeffchew @RobertaJHahn 
1. I believe IBM Docs is using currently, so should we clear it with them first? 
2. if we can get the Web component card pictogram fixed first, that would make it easier for deprecation first I hope

skuruvil:
Design spec: https://ibm.ent.box.com/folder/114042319288
Functional spec: https://ibm.box.com/s/hgvu48r1fv2ga2h0yn3qshrtxjc71vch

oliviaflory:
This epic is complete

PR #7795, opened 2021-12-01
feat(card-pictogram): add design updates to improve mobile accessibility & hover interactions

proeung:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7553

### Description

- This PR adds design updates to improve the visibility of card copy content on the mobile breakpoints as well as adding motion/hover interactions on larger breakpoints.
- See EPIC for more detail - https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6739
- Link to animation/hover interaction specs - https://ibm.box.com/s/ykv2ukr3twyfx2nrro3rn7ge4iqoz48x

### Changelog

**New**

- `.#{$prefix}--card__motion` - a new class for applying the card pictogram hover interactions.

**Changed**

- Move Card Pictogram related styles from `packages/web-components/src/components/card/card.scss` to `packages/styles/scss/components/card/index.scss`, so that these changes are available for use in other packages.

https://user-images.githubusercontent.com/1815714/144290175-1ebdb608-a1aa-4cc1-9454-764f6046f7b0.mov


RichKummer:
Hey @proeung , this is looking great! One issue I found is that when the window changes breakpoints from `md` to `lg`, the body text animation runs over the header.

**Video**
https://user-images.githubusercontent.com/15643582/144436203-01d52b93-4989-45d6-98fa-076a96b744d9.mov

Ideally, when going from `md` to `lg`, the card just changes to the initial state for desktop (only heading and pictogram) without an animation running over it.

emyarod (commented):
on a side note, should we allow the card height to grow as the viewport width shrinks?

image

proeung:
@RichKummer @emyarod This PR is ready for a re-review.

RichKummer (approved):

IgnacioBecerra (approved):
LGTM, the hover animation looks really nice! thanks @proeung!

jeffchew (changes requested):
@proeung can you double check the e2e test? This one in particular is failing:

image

shixiedesign:
Hey @proeung this is beautifully done! Sorry to nitpick, just a few small issues and all on focus state: 

1. When card is on focus, we should trigger the hover effect as well so keyboard users have a chance to read the card copy. 

Dec-03-2021 08-38-12

2. Focus outline when Card is "Outlined" is 3 px while default is 2px. Expecting 2px
image


3. Focus outline when Card is "Inversed" should be 2px blue + 1px white in order to avoid the outline disappearing on white theme: 

Current: 
Dec-03-2021 08-44-11

Expected - according to default card's specs, focus outline should be as below.
image


I also have question if the screen reader is able to read the copy.

proeung:
@shixiedesign No worries! I just pushed up a commit that shows the copy on `focus` and addresses border styling issues that you've noticed on the `outlined` and `Inverse` styles.

@jeffchew The errors seen within the e2e test should be fixed as well.

shixiedesign:
Hi @proeung Noticed another style change. There should be a padding right of 2rem on `md—max` breakpoint on the heading. `sm` don't need this padding.

**Current:**
image
**Expected:**
image

 I'm around before 3pm ET and feel free to when you are done! Thanks so much for making all these fixes 🌹

skuruvil:
@shixiedesign @proeung Sorry, couple of other things I have noticed

1. It seems like we have not applied a minimum 32px bottom/top margin for the heading. I think similar to body copy card should grow taller to accommodate lengthier heading too. 
Screenshot 2021-12-06 at 4 23 54 PM

2. Position of a pictogram at the bottom is supposed to be the default variant
3.  @sm breakpoint text may take 100% of the container width, right now I think we have applied 32px padding at the right side. But that is something I can see across all the card variants (default, static, etc...)
Screenshot 2021-12-06 at 4 35 17 PM

jeffchew:
Thanks @proeung , looks like other tests are failing now though:

image

proeung:
@jeffchew Those errors are popping up because the border is no longer being applied to the card host component, but within `.bx--card` to address the focus styling issue that @shixiedesign mentioned here (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/7795#issuecomment-985533253). I just pushed up a commit and that should fix the error log.

----------

@shixiedesign @skuruvil The long heading overlapping issue is a result of the text being positioned as `absolute` to get the nice transition in and out. I pushed up an example of the hover motion work with the heading and copy positioned as `relative` and the height set to 0, however, you don't get that smooth transition out effect. Below are screenshots of the differences as well as the pros and cons of each implementation. Take a look and let me know what you guys think.

### **Heading with `absolute` position**

**Pros**
- Creates nice transition in and out.
- Restrict heading copy to just 2 to 3 lines, we can add line-clamp.

**Cons**
- Heading text is overlapped when the character length goes to 5 lines unless we grab the height of the card heading and apply it to the card height. This requires additional dev time.

https://user-images.githubusercontent.com/1815714/144922119-8d543c3a-0139-4691-ae00-00faebe86502.mov


### **Heading and copy using height 0**

**Pros**
- Heading can be set to 5 or more lines without overlapping the icon.
- Height of the card grows based on the length of the heading text.

**Cons**
- No transitions out
- Card with a heading that spans 5 or more lines and with a short copy, there will be a change in the card heigh, unless used within the Card Group component which is where the sameHeight() function is beingappiled.

https://user-images.githubusercontent.com/1815714/144922146-f6df88ba-0b00-4dc9-94b8-f3cd9827eea8.mov

shixiedesign:
Thanks @proeung for demo'ing the two approaches. I've got a weak spot for beautiful animation so my vote is definitely the first one. Also, I noticed the second approach added a bug - if copy is long, height of card grow on hover/focus abruptly. 

Dec-07-2021 01-53-36

shixiedesign (approved):
Looks good! Thank you.

jeffchew:
Hey @proeung , looks like in React it still can't find the border:

image

proeung:
@jeffchew Looks like React 2e2 tests came back clear now.

jeffchew (approved):
Thank you for fixing @proeung !

PR #7832, opened 2021-12-07
fix(ContentBlockHorizontal): add border-bottom option & knob

### LINKED ISSUE #7640: [Content block horizontal] Add ability to hide horizontal rule in dds-content-group-horizontal element  - DDS Consulting
**Engineering info:**  
- HC engineer: @andy-blum 
- HC JIRA Ticket: https://jsw.ibm.com/browse/HC-2340
- DDS consulting engineer:

----

### The problem

image

We need to be able hide horizontal rule in the ~dds-content-group-horizontal~ [Content block horizontal](https://www.ibm.com/standards/carbon/web-components/?path=/story/components-content-block-horizontal--default) element. 

### The solution

The solution would be an attribute we can pass in the element to hide the horizontal rule. Similar to how no-border is used in the dds-cta-block element.

### Application/website

AEM (Marketing)

### Business priority

_No response_

### Examples

_No response_

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/.github/CODE_OF_CONDUCT.md)

andy-blum:
### Related Ticket(s)

Fixes #7640 

### Description

Corrects an error where ContentBlockHorizontal hardcoded the "border" value to `true`.

### Changelog

**Changed**

- Fixes ContentBlockHorizontal to respect `border` prop
- Adds knob to ContentBlockHorizontal storybook for `border` prop


proeung (changes requested):
@andy-blum Looks like the ` border` attribute for conditionally showing and hiding the border/hr line is only being set on the React package side. We'll need to introduce a similar solution on the Web Components side where instead of returning `<dds-hr>` (see [link](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/packages/web-components/src/components/content-block-horizontal/content-block-horizontal.ts#L28)) in the `<content-block-horizontal>` component we're setting the border styling on the host level. @ariellalgilmore or @IgnacioBecerra Let me know if you agree.

packages/react/src/components/ContentBlockHorizontal/ContentBlockHorizontal.js
```diff
@@ -17,12 +17,12 @@ const { prefix } = settings;
 /**
  * ContentBlockHorizontal pattern.
  */
-const ContentBlockHorizontal = ({ heading, items }) => {
+const ContentBlockHorizontal = ({ heading, items, border }) => {
```
emyarod, line 20: should `border` have a default value to maintain continuity with the currently released version?

emyarod (commented):
looks good to me, just one question about the default `border` value

CI tests should pass once the snapshots are updated!

packages/react/src/components/ContentBlockHorizontal/ContentBlockHorizontal.js
andy-blum: While setting `border == true` would be the backwards-compatible solution, the functional specs specify that the default value be `false`

https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/wiki/Layout#4-border

emyarod: got it, thanks for clarifying!

proeung (approved):
LGTM! Thanks for addressing my feedback!

annawen1 (approved):
LGTM! Thanks @andy-blum!

Issue #7831, opened 2021-12-07
[LOCALE MODAL]: geographic card is clickable while list is empty and disabled icon is shown 


Fezi32:
### Description

When the Locale Modal is used and the one or more geographic areas has an empty list, the disabled icon is shown, but the geographic area card is clickable. 

**How it's working right now:**
User can click on the geographic area even if the list behind it is empty and the disabled icon is displayed.

**How it should work**:
User can't click on the geographic area, if a list is empty and the disabled icon is shown.

### Component(s) impacted

Locale Modale (https://www.ibm.com/standards/carbon/react/?path=/story/components-locale-modal--default):
<img width="655" alt="Locale-modal" src="https://user-images.githubusercontent.com/4804449/145063926-d4cfb8ac-0c19-4ba9-87b1-ddebdc8e24c9.png">

### Browser

Chrome

### Carbon for IBM.com version

1.29.0

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

IBM project

### Package

@carbon/ibmdotcom-react

### CodeSandbox example

https://www.ibm.com/standards/carbon/react/?path=/story/components-locale-modal--default

### Steps to reproduce the issue (if applicable)

1. Add custom `localeData` with at least one empty list for any geographic area
2. Open the Locale Modal in your application
3. Click on the "disabled" geographic area card
4. You should now see an empty list

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

Fezi32:
Thanks guys ❤️

PR #7840, opened 2021-12-08
fix(localemodal): disable region button when no countries present

kennylam:
### Related Ticket(s)

#7831 

### Description

Fixes a bug in `LocaleModal` where the region button is still clickable even when no countries of that region are available.

**Note**: `Europe` has been disabled for testing purposes.

Dec-08-2021 16-38-07

### Changelog

**Changed**

- update event handler for region buttons with no countries/languages available for localization
- sort country list in `preview.html` by region


emyarod (approved):
looks good to me! the empty region is disabled properly now

jeffchew:
@kennylam hey for web components, the tile doesn't look disabled:

image

How it looks in React:

image

emyarod (commented):
after rechecking on narrower viewports, it looks like there's a little weird hover effect on the disabled tile in web components

https://user-images.githubusercontent.com/8265238/145442736-d9d4af62-c592-474d-a044-18951553b208.mov

kennylam:
@jeffchew @emyarod Both of those are existing issues that should probably be addressed separately:

- the `disabled` styles for web components `card-link` do not match the React (and Carbon) version
- the disabled icon shifting position on hover can be seen [here](https://www.ibm.com/standards/carbon/web-components/?path=/story/components-locale-modal--default)

I'll open tickets for both of these issues.

emyarod (approved):
sounds good! then I think we can merge after restoring the removed locales for testing

kennylam:
> sounds good! then I think we can merge after restoring the removed locales for testing

I left Europe disabled to match the web components version made in [this PR](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/7302#pullrequestreview-769307374). Should we enable them again?

jeffchew:
> > sounds good! then I think we can merge after restoring the removed locales for testing
> 
> I left Europe disabled to match the web components version made in [this PR](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/7302#pullrequestreview-769307374). Should we enable them again?

No let's keep them both to match where Europe is disabled.

Issue #7754, opened 2021-11-24
[Test Scenario] CTA - Button: DDS Consulting


Praveen-sr:
**Engineering info:**  
- HC engineer: @jkaeser 
- HC JIRA Ticket: https://jsw.ibm.com/browse/HC-2334
- DDS consulting engineer:

----

### Component and test scenario title

CTA - Button

### Test scenario steps

a. Check 2 buttons, with copy text is loaded
b. Verify the buttons are clickable
c. Change CTA type on knobs section:
- Verify Jump option
- Verify External option
- Verify Download option
- Verify Video option
Note: Copy text and link href can be customized from knobs section.

Issue #7753, opened 2021-11-24
[Test Scenario] CTA - Text: DDS Consulting


Praveen-sr:
**Engineering info:**  
- HC engineer: @jkaeser 
- HC JIRA Ticket: https://jsw.ibm.com/browse/HC-2333
- DDS consulting engineer:

----

### Component and test scenario title

CTA - Text

### Test scenario steps

a. Check links, with heading and CTA link is loaded
b. Verify the link is clickable
c. Change CTA type on knobs section:
- Verify Jump option
- Verify External option
- Verify Download option
- Verify Video option

PR #7844, opened 2021-12-09
test(cta): add e2e tests for text and button variants

jkaeser:
### Related Ticket(s)

- https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7753
- https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7754

### Description

Adds storybook test coverage for Text and Button CTA component variants in the web components and react packages.

### Changelog

**New**

- Adds e2e-storybook tests for `dds-text-cta` web component
- Adds e2e-storybook tests for `TextCTA` react component
- Adds e2e-storybook tests for `dds-button-cta` web component
- Adds e2e-storybook tests for `ButtonCTA` React component


emyarod (commented):
it looks like there are undefined variables causing some tests to fail, should there be a null check before destructuring them?

packages/web-components/tests/e2e-storybook/cypress/integration/cta/_tests.js
```diff
@@ -0,0 +1,105 @@
...
+        })
+        .visit(`${paths.copy}${customCopyInput}`)
+        .get(selector)
+        .should(([copy]) => {
+          customCopyOutput = copy.innerText.trim();
+
+          expect(customCopyOutput).to.be.eq(customCopyInput);
+          expect(customCopyOutput).to.not.eq(defaultCopy);
+        });
+    });
```
jkaeser, line 58: > it looks like there are undefined variables causing some tests to fail, should there be a null check before destructuring them?

@emyarod Hmm, I don't know if adding a null check is the way to go since that would invalidate the test (it _should_ fail if the element isn't found). Notably, this test function is run against desktop and mobile, but it only failed on mobile in this latest run:

Screen Shot 2021-12-09 at 1 44 27 PM

I haven't been able to replicate the failure locally, either. Is it possible this is an issue with the CI environment?

jkaeser: Turns out it was inconsistently failing locally, too, and I simply hadn't run the test enough times to get a failure. Pushed a commit that seems to make things more consistent. Let's see what the CI tests have to say about it 🤞

emyarod (approved):
looks good to me!

proeung:
@jkaeser I'm seeing some e2e errors. Can you take a look?

Screen Shot 2021-12-14 at 3 22 15 PM

jeffchew:
> > @jkaeser I'm seeing some e2e errors. Can you take a look?
> 
> Hmm it's strange that it's working fine in Chrome but not Firefox...I'm going to re-run the e2e tests to see if it isn't a load issue.

@proeung the latest run seemed ok, so I think it might be the tests throwing some false positives at us. We are still working through those (there are a couple of culprits that keep sporadically keep coming up)...

proeung:
@jeffchew Good to know! We'll make sure to re-run the test with the latest changes so that we can avoid these types of false positives.

jeffchew:
> @jeffchew Good to know! We'll make sure to re-run the test with the latest changes so that we can avoid these types of false positives.

Also don't be afraid to reach out to ask if something is a false positive or not. I might actually create some issues so we can tackle this and prevent these from popping up in the future.

PR #7845, opened 2021-12-09
fix(leadspace): allow background color to extend to full viewport width

kennylam:
### Related Ticket(s)

#7790 

### Description

This change allows the leadspace theme background color to appear in `>=lg` breakpoints.

https://user-images.githubusercontent.com/909118/145420196-acf90526-6b4c-447e-b3eb-464c96bf952d.mp4

### Changelog

**New**

- CSS to allow leadspace to go full-width of the viewport allowing the theme background color to show


emyarod (commented):
this may be a storybook only issue but it looks like there are horizontal scrollbars when viewing the stories now (sometimes they only appear after changing the viewport width). the overflow doesn't happen when viewing the full stories but in the storybook view it is a slight change from the existing stories

image

image

shixiedesign:
Discussed with Kenny on Slack. This PR might need a sandbox in order to test the mixing of themes on 1 page. Waiting to see that.

shixiedesign (approved):
Reviewed using https://lcnw3.sse.codesandbox.io/
Thank you!

kennylam:
> this may be a storybook only issue

Good catch @emyarod, fixed now. Thanks!

Issue #7847, opened 2021-12-09
Broken sandbox code on L0 component


estevaocma:
### Description

https://ibmdotcom-web-components-canary.mybluemix.net/?path=/docs/components-masthead--with-custom-navigation

### Component(s) impacted

Masthead

### Carbon for IBM.com version

v 1.14.0

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

https://ibmdotcom-web-components-canary.mybluemix.net/?path=/docs/components-masthead--with-custom-navigation

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://codesandbox.io/s/github/carbon-design-system/carbon-for-ibm-dotcom/tree/master/packages/web-components/examples/codesandbox/components/masthead

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #7848, opened 2021-12-09
docs(masthead): add missing sandbox dependency

emyarod:
### Related Ticket(s)

#7847

### Description

This PR adds a missing dependency to the masthead sandbox example so that it displays without any build errors

The working sandbox should be viewable at https://codesandbox.io/s/github/emyarod/carbon-for-ibm-dotcom/tree/7847-docs/masthead-sandbox-example/packages/web-components/examples/codesandbox/components/masthead

### DIFF
--- a/packages/web-components/examples/codesandbox/components/masthead/package.json
+++ b/packages/web-components/examples/codesandbox/components/masthead/package.json
@@ -16,6 +16,7 @@
   "devDependencies": {
     "@babel/core": "^7.0.0-0",
     "carbon-components": "^10.36.0",
+    "lit-element": "^2.5.1",
     "parcel-bundler": "1.12.3",
     "rimraf": "^3.0.2",
     "sass": "^1.32.13"
--- a/packages/web-components/examples/codesandbox/components/masthead/src/index.js
+++ b/packages/web-components/examples/codesandbox/components/masthead/src/index.js
@@ -10,6 +10,13 @@
 import '@carbon/ibmdotcom-web-components/es/components/masthead/masthead-container.js';
 import './index.scss';
 
+/**
+ * the `lit-element` package is not required in `package.json`, but is included
+ * in this example for CodeSandbox compatibility
+ *
+ * https://github.com/codesandbox/codesandbox-client/issues/4456
+ */
+
 window.digitalData = {
   page: {
     pageInfo: {


kennylam (approved):
Good catch thanks @emyarod!

packages/web-components/examples/codesandbox/components/masthead/package.json
```diff
@@ -16,6 +16,7 @@
   "devDependencies": {
     "@babel/core": "^7.0.0-0",
     "carbon-components": "^10.36.0",
+    "lit-element": "^2.5.1",
```
jeffchew, line 19: @emyarod should this be added? I thought we removed `lit-element` and `lit-html` as dev dependencies.

emyarod: @jeffchew yeah I'm not sure, but it was the only way I could get the code sandboxes to load properly after users reported not being able to use them

jeffchew:
@emyarod we should actually look into this because ideally our adopters shouldn't need to do this.

packages/web-components/examples/codesandbox/components/masthead/package.json
jeffchew: @emyarod we should actually look into this because ideally our adopters shouldn't need to do this.

jeffchew: Checking another random PR, building the codesandbox example actually seems to build fine: 
https://webcomponents-codesandbox.s3-web.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/7827/masthead/index.html

I'm wondering if it's a CSB specific issue?

emyarod: yeah I thought it might be related to CSB's bundler but I'm not 100% sure on that

jeffchew: And funny enough, this actually sounds more related to this which was opened by @tay1orjones:
https://github.com/codesandbox/codesandbox-client/issues/4456

jeffchew: @emyarod since it sounds like a CSB  issue, perhaps we can leave this in but leave a note somewhere in the code sandbox example in a pretty obvious place that `lit-element` is actually not required but added to get around an issue in CSB itself

emyarod: @jeffchew I added a note in `index.js`, would that be enough or should we add a README to that example instead?

I imagine some other code sandboxes may be affected by this issue but I'm not sure which ones exactly yet

packages/web-components/examples/codesandbox/components/masthead/src/index.js
```diff
@@ -10,6 +10,13 @@
 import '@carbon/ibmdotcom-web-components/es/components/masthead/masthead-container.js';
 import './index.scss';
 
+/**
+ * the `lit-element` package is not required in `package.json`, but is included
+ * in this example for CodeSandbox compatibility
+ *
+ * https://github.com/codesandbox/codesandbox-client/issues/4456
+ */
```
jeffchew, line 18: 👍

packages/web-components/examples/codesandbox/components/masthead/package.json
jeffchew: I think should be fine where you put it, thank you!

Issue #7474, opened 2021-10-21
[Test Scenario]: Lightbox-media-viewer--embedded-video-player


Praveen-sr:
### Component and test scenario title

Lightbox-media-viewer--embedded-video-player

### Test scenario steps

a. Check the modal is loaded with Video, heading and copy. 
b. Verify the video is left aligned and heading and copy text is on right side
c. Verify Video player, play, pause, mute and full screen settings.
d. Customize theme from Carbon theme and verify gray 10, gray 90 and gray 100 theme.
e. Verify Close icon, modal should be closed.

PR #7870, opened 2021-12-13
test(lightbox-media-viewer): add embedded video player e2e tests

emyarod:
### Related Ticket(s)

#7474

### Description

This PR adds storybook e2e tests for the lightbox media viewer - embedded video player variant.

> ### Test scenario steps
> 
> a. Check the modal is loaded with Video, heading and copy. 
> b. Verify the video is left aligned and heading and copy text is on right side
> c. Verify Video player, play, pause, mute and full screen settings.
> d. Customize theme from Carbon theme and verify gray 10, gray 90 and gray 100 theme.
> e. Verify Close icon, modal should be closed.


ariellalgilmore (commented):
LGTM! I am wondering though if we can try to compact some of this where we have the same checks for default and video paths such as aligning the elements and checking if the `title` and `desc` are not empty

jeffchew:
> LGTM! I am wondering though if we can try to compact some of this where we have the same checks for default and video paths such as aligning the elements and checking if the `title` and `desc` are not empty

Do you mean updating so we have shared methods that are then triggered across multiple tests?

jeffchew:
@ariellalgilmore @emyarod think this is something we can merge in or more changes are needed?

ariellalgilmore:
> > LGTM! I am wondering though if we can try to compact some of this where we have the same checks for default and video paths such as aligning the elements and checking if the `title` and `desc` are not empty
> 
> Do you mean updating so we have shared methods that are then triggered across multiple tests?

I was hoping we could make methods so that the default and video tests could use the same methods instead of repeating. Let me know what you think @emyarod

emyarod:
@ariellalgilmore I just made added this https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/7870/commits/a82588df6242d96370be0f9102748995ff193cf2 is that closer to what you meant?

ariellalgilmore (approved):
LGTM!

Issue #7481, opened 2021-10-21
[Card]: Card copy doesn't properly convert markdown to HTML


emyarod:
### Description

When passing markdown into the Card copy, the markdown should be converted to formatted HTML

![image](https://user-images.githubusercontent.com/8265238/138337795-14236c22-91a1-4e14-aff5-aa3426bcbc16.png)

### Component(s) impacted

Card

### Carbon for IBM.com version

canary

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

-

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-card--default

### Steps to reproduce the issue (if applicable)

Pass markdown into the Card copy and observe that it is not formatted as expected

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

Issue #7469, opened 2021-10-20
[Card] Ampersand not being rendered on card copy prop


leticiafontoura:
### Description

I'm using the CardGroup component and after going through a lot of files on node_modules, I discovered that here, on [this ](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/0a41071880c17eb485d4261eaf771ed22b3b8b90/packages/utilities/src/utilities/markdownToHtml/markdownToHtml.js#L91)line, the current version of "marked" is removing ampersand from the text. I realized this by letting it just `const convertedMarkdown = DOMPurify.sanitize(converted)` and the problem was solved: ampersand was displayed. 

Not sure if changing marked's version would help but...

### Component(s) impacted

CardGroup > cards copy prop

### Browser

Chrome

### Carbon for IBM.com version

1.22.0

### Severity

Severity 1 = The design is broken in a critical way that blocks users from completing tasks or damages the brand. Affects major functionality, no workaround.

### Application/website

https://wwwpreview.ibm.com/resources/watson-health/gaining-insights-healthcare-data-webinar/

### Package

@carbon/ibmdotcom-react

### CodeSandbox example

N/A

### Steps to reproduce the issue (if applicable)

Try updating "marked" version.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #7871, opened 2021-12-13
fix(card): ensure copy can render markdown

IgnacioBecerra:
### Related Ticket(s)
#7481

### Description
This PR ensures that the copy elements in Card and Content Section can render markdown properly. The approach used in Card queries the slotted `p` tag and applies the markdown utility function. While a simpler method may be creating a `card-copy` component that extends `DDSMarkdown`, I refrained from doing so to avoid a breaking change for adopters using Card. Feel free to let me know what you think about doing that approach instead and simply mentioning it in the release notes.

Note: 
Testing markdown in Storybook seems to only work once (i.e. inputting `*test*`, deleting it, and trying the same thing will result in the text being wrapped in asterisks without formatting. 

However, if an adopter uses markdown in the copy of the Card component, it will render properly.

Screen Shot 2021-12-13 at 3 10 55 PM

### Changelog

**Changed**

- Card copy can now use markdown
- Content section copy now extends `DDSMarkdown` component


emyarod (approved):
> creating a `card-copy` component that extends `DDSMarkdown`

I think this would be fine given that the component was supposed to have this functionality already. Since the concern is around it being a breaking change it should be ok to make this update in v2

looks good to me, just needs a few CI test updates

emyarod (commented):
one thing I noticed, characters like `#` and `&` won't render properly in the card copy with the markdown parser

related https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7469

IgnacioBecerra:
@emyarod I was able to get the `#` to render, but yeah the ampersand was getting removed. It seemed to have been getting stripped using the regex expression after the `&` became `&amp` when parsed through `marked()`, so I removed that part from the expression so it can remain there.

Screen Shot 2021-12-14 at 2 20 14 PM
Screen Shot 2021-12-14 at 2 19 43 PM

emyarod (approved):
looks good to me! should we create a separate issue for the storybook rendering problems?

jeffchew:
@IgnacioBecerra just double checking if the font size change  in the link list was intentional here? or perhaps this is coming from an outdated baseline?

image

IgnacioBecerra:
@jeffchew yeah that change isn't coming from this PR, the actual font size isn't changing with these changes

jeffchew:
@IgnacioBecerra there seems to be something going on with the card group too, I don't see this on latest canary and double checked that it's happening on your fork:

image

IgnacioBecerra:
@jeffchew Found out what was happening -- once there is a `dds-content-block-copy` in the page, the `renderer` function it uses to convert its text to `<dds-content-block-paragraph` overrides the default one in the `markdownToHtml` utility, which affects the rest of the components using it.

In other words, any use of the utility after using `content-block-copy` will always be wrapped with at `<dds-content-block-paragraph`, which includes the `p` tags in Card that we’re modifying in this PR. A way to circumvent the big text in the cards is to just remove the styles from the rendered `content-block-paragraph` element. 

As mentioned in an earlier comment, creating a `dds-card-copy` component that extends `ddd-markdown` will solve the problem properly as it will have its own renderer, but since it will be a breaking change we could have this feature for v2 next year.

jeffchew:
@IgnacioBecerra thank you for updating, the diffs look much better!

There's still a slight diff for this, wondering if we need a design eye to take a glance at this?

image

jeffchew (approved):
Thanks @IgnacioBecerra !

PR #7879, opened 2021-12-14
test(cdn): introduce e2e tests for cdn artifacts

jeffchew:
### Related Ticket(s)

No related issue

### Description

After getting a request from an adopter to help with their usage of the CDN artifacts and noticing it throwing exceptions, it'd be good to get some testing into place so we are aware of any load issues in the CDN aftifacts as we are deploying regularly on EI server.

There is no way for us to regularly purge akamai automatically based on our limited access, so we would need to do a manual purge as needed. With this, we still would need a heads up whenever there are any issues on production, so this test that runs nightly will assist with that.

This test introduces e2e tests that run against all CDN artifacts on canary, next, latest, and the latest specific version. It will throw errors if there are any exceptions from these.

Right now, there is an error showing for Canary, which we can fix separately from this test harness (and is actually a deployment fix and not a code fix).

image

image

### Changelog

**New**

- e2e tests for CDN artifacts

**Changed**

- Updated kitchen sink test page to include all current CDN artifacts


emyarod (approved):
looks good to me!

PR #7888, opened 2021-12-15
chore(react-wrapper): update to use prod url

jeffchew:
### Related Ticket(s)

No related issue

### Description

Noticed that the lower environment was being used for the React wrapper links. This updates to use the prod url instead.

### Changelog

**Changed**

- Multiple files updated to point to `www.ibm.com/standards/carbon/web-components/react` for react wrapper links


emyarod (approved):
looks good to me!

Issue #7794, opened 2021-12-01
[Masthead]: Only one item should be underlined - DDS Consulting 


EAVilleda:
**Engineering info:**  
- HC engineer: @jkaeser 
- HC JIRA Ticket: https://jsw.ibm.com/browse/HC-2339
- DDS consulting engineer: @annawen1 - @huntermacd 

----

### Description

On the live site, we have issues where two items are underlined in the masthead. Example: https://www.ibm.com/topics/supply-chain-management
![image](https://user-images.githubusercontent.com/47429622/144280069-4c52cb71-f765-4066-8c05-f6b81d54bb7a.png)

Anna Wen explained that a new feature was implemented to detect the selected/active menu item based on the url. This is why "Learn" is underlined, but “Supply Chain” should not be underlined in this case.

### Component(s) impacted

Masthead https://www.ibm.com/standards/carbon/web-components/?path=/story/components-masthead--default

### Carbon for IBM.com version

v1.26.0

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

AEM (Marketing)

### Package

@carbon/ibmdotcom-web-components, @carbon/ibmdotcom-styles

### CodeSandbox example

No codesandbox avaiable

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

proeung:
This bug issue is related to the work that went into this EPIC (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6638) & https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6763.

proeung:
The highlighting style on the platform name (eg. `Supply Chain`) within the L1 nav component is set by default. We'll need to adjust the logic for selecting the current nav item so that the platform name doesn’t get highlighted as well.

proeung:
@RobertaJHahn @jeffchew I'm adding Anna as the consulting engineer for this issue as she worked on #6763.

PR #7890, opened 2021-12-16
fix(masthead): account for platform name in active item logic

jkaeser:
### Related Ticket(s)
Resolves https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7794.

### Description

There are currently two ways menu items can be set to "active" and receive a blue underline:
1. Manually: Set the `selected-menu-item` property on the `<dds-masthead-composite>` element
2. Automatically: A menu item has a URL that is identical the the current page's URL.

Additionally, there is a concept of a "platform" supported by the optional `platform` and `platformUrl` properties. When these properties are set, a bolded link appears first in the masthead's L1 nav, and the tricky part is that this link also has logic for whether or not it is set to active. This logic was aware of the manually set active menu items, but it did not account for the automatically set ones.

### Changelog

**Changed**

- Adds `_hasAutoSelectedItems` internal property to track whether automatically selected menu items are present or not.
- Uses `_hasAutoSelectedItems` internal property as a factor when determining if platform link should receive active styles.


emyarod (commented):
looks good to me! is this something we should add as a knob to storybook or include in the docs?

proeung:
@emyarod Great point! I wonder if it's useful to add a knob that allows you to insert the URL of the SB page URL for menu items like “Link 2”. The "selected menu item" will need to conditionally set to empty if this knob has a value (see attached). This will be good for testing purposes, but I don't want it to hold us back from getting a solution out in the next release since this is a bug that the AEM adopters are seeing on PROD.

Screen Shot 2021-12-16 at 2 30 21 PM

emyarod:
@proeung yep understood, if we could track the effort for storybook in a separate issue that would be great! as for this ticket I think a note about it in the docs/props table should be fine for now

annawen1 (approved):
LGTM! Thank you @jkaeser for the fix!

proeung (approved):
LGTM!

------

@annawen1 @emyarod Thank you both for the review!

Issue #6338, opened 2021-06-10
[Card group] Storybook refinement


shixiedesign:
### Update Card group to only contain following entries, and demo the other variations through knobs: 

Left side image is current state screenshot. Right side text describes what we expect.

```
Card group
|--- Default
|--- With card in card
```

![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/790543ac-ab85-43fa-97d5-2e8047311ced)

### Expected knobs within each story

Left side image is current state screenshot. Right side text describes what we expect.

**Card group - default**

![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/049437f5-7b7f-448e-9da1-3dce87209743)

**Card group - with card in card**

![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/a1a2d321-e64f-4053-8ae1-feece965c4fc)

### Additional note on follow on work
- after the storybook entry is updated, create `test: e2e` stories to test for Percy to capture snapshot when the knobs are changed

shixiedesign:
Design need to revisited to make sure issue description is accurate.

PR #7924, opened 2021-12-30
fix(card-group): update storybook

ariellalgilmore:
### Related Ticket(s)

#6338 

### Description

Update card-group story to use knobs instead of having different stories.

Screen Shot 2021-12-30 at 8 16 26 AM

Note: 
- cards-per-row knob does not work with narrow grid mode
- same height with tag group with video and images are not aligning

### Changelog

**New**

- prop to card-group `pictograms` so heading for pictogram cards stay aligned to bottom

**Changed**

- knob to choose card types
- knob to add media (alternates images and videos
- min 2 and max 6 to amount of cards
- added `outline` to `grid mode` options, so user can only select one at a time


annawen1 (approved):
LGTM!

packages/web-components/src/components/card-group/__stories__/card-group.stories.react.tsx
```diff
@@ -48,184 +47,229 @@ const phraseArray = [
...
+      <DDSCardEyebrow>Topic</DDSCardEyebrow>
+      {tagGroup ? tagGroupContent : ''}
+      <DDSCardCTAFooter cta-type="video" href="1_9h94wo6b" slot="footer"></DDSCardCTAFooter>
     </DDSCardGroupItem>
   );
+
   count = count > 3 ? 0 : count + 1;
-  return defaultCardGroupItem;
+  // eslint-disable-next-line no-nested-ternary
+  return media ? (index % 2 === 0 ? defaultCardGroupItem : videoCardGroupItem) : defaultCardGroupItem;
```
emyarod, line 95:
```suggestion
  return media && index % 2 ? videoCardGroupItem : defaultCardGroupItem;
```

would this be equivalent without nesting ternaries?

emyarod (commented):
looks good to me, just wondering would it be worth extracting out the `gridMode` logic to avoid nesting ternaries as well?

RichKummer:
Took a quick look before the last commit: **Card group - Default** looks good, but it looks like there's a missing checkbox for **Card group - with card in card:**
* Missing an _Outlined cards_ checkbox.

Screen Shot 2022-01-04 at 2 26 19 PM

* It also looks like the default Number of cards should be 5, but currently defaults to 3.

RichKummer (approved):
Looks good! TY for the last change!

PR #7996, opened 2022-01-10
chore(deps): upgrade @percy/storybook to v4

jeffchew:
### Related Ticket(s)

No related issue

### Description

This is a test to see how percy diffs look when upgrading to `@percy/storybook` v4. 

### Changelog

**Changed**

- Upgraded `@percy/storybook` to v4


jeffchew:
Closing this PR for now, reached out to the Browserstack Percy team the findings. Noted that enable javascript was turned on in the percy config file, and that there are a number of missing snapshots in both Web Components and React.

Robdel12:
Hey @jeffchew! After some some digging the skipped stories are because of the Percy story config. For example:
https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/d1f36c5328c599695456cacd1535791b9f4bb45a/packages/react/src/components/TabsExtended/__stories__/TabsExtended.stories.js#L16 or https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/08eea4f1d7d36dbcc07739544f3acac2c7248cc5/packages/react/src/components/FeatureSection/__stories__/FeatureSection.stories.js#L17

Issue #7599, opened 2021-11-04
[IBM Accessibility]: Add Karma Accessibility Checker in Web Components


jeffchew:
### The problem

Currently the accessibility checker is not configured for web components. This is something that the accessibility wishes to have in place in addition to the standard WCAG testing we are currently doing.

### The solution

Need to integrate the automated checker: https://www.npmjs.com/package/karma-accessibility-checker

This can be added to the existing workflow here:
https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/7c36799d67b9268d52fe30a8a7f39203677ec83a/.github/workflows/ci-check.yml#L44

### Application/website

Carbon for IBM.com

### Business priority

Low Priority = release date is not dependent on fix or not upcoming

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #7997, opened 2022-01-10
feat(karma-accessibility-checker): add to web components

ariellalgilmore:
### Related Ticket(s)

#7599 

### Description

Adds the karma accessibility checker to web components.

In order to render each of the stories and obtain props, each wc story added a `propsSet` value.

### Changelog

**New**

- {{new thing}}

**Changed**

- {{changed thing}}

**Removed**

- {{removed thing}}


packages/web-components/tests/utils/achecker-compliance.js
```diff
@@ -0,0 +1,126 @@
...
+                          XPathResult.ANY_TYPE,
+                          null
+                        )
+                        .iterateNext();
+                    return !filterFuncs.some(filterFunc => filterFunc(issue, elem));
+                  }
+            );
+            if (issues.length > 0) {
+              if (process.env.AAT_VERBOSE) {
+                console.error(
```
jeffchew, line 82: @ariellalgilmore there is a warning here about using console, can you add an eslint ignore for this line?

packages/web-components/tests/karma.conf.js
```diff
@@ -43,21 +43,25 @@ module.exports = function setupKarma(config) {
...
+      'src/polyfills/index.ts',
+      'tests/utils/snapshot.js',
+      'tests/snapshots/**/*.md',
+    ].concat(specs.length > 0 ? specs : ['tests/a11y/karma-test-shim.js']),
 
     preprocessors: {
-      'src/**/*.[jt]s': ['webpack', 'sourcemap'], // For generatoring coverage report for untested files
-      'tests/karma-test-shim.js': ['webpack', 'sourcemap'],
+      'src/**/*.[jt]s': ['webpack', 'sourcemap'], // For generating coverage report for untested files
+      'tests/a11y/**/*.js': ['webpack', 'sourcemap'],
```
jeffchew, line 64: Does this run as part of the regular functional unit tests? Have you noticed any test failures when running?

ariellalgilmore: yes it runs with `yarn test:unit`.

and yes lots of failures when runing 

Screen Shot 2022-01-10 at 12 59 02 PM.

A lot of errors seem to be missing roles 
Screen Shot 2022-01-10 at 1 00 06 PM

43 failed, 102 sucess.

jeffchew: > yes it runs with `yarn test:unit`.
> 
> and yes lots of failures when runing
> 
> A lot of errors seem to be missing roles Screen Shot 2022-01-10 at 1 00 06 PM
> 
> 43 failed, 102 sucess.

I wonder if we should separate out the functional unit tests from the a11y tests and run as a separate job? We can probably see if we can keep the same config file, but perhaps pass a flag if we can so that it will run the a11y test configs only in that case?

And when we do that, we can pass the tests regardless of status for now (kinda like how we're doing it with React right now) until we fix the failures then we can enforce it when we're ready?

jeffchew: Or if there are enough diffs, we can have a separate `karma.conf.js` specifically for `aChecker` for the a11y specific job...

jeffchew (approved):
Looks good to me, thank you @ariellalgilmore and @emyarod !

Issue #7822, opened 2021-12-06
[Masthead: search]: add grouped results in search typeahead


kennylam:
### The problem

Grouped search results feature is not present in web components `masthead`.

### The solution

Add grouped search results in masthead search. See https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/4845. Example usage (in React) can be seen [here](https://codesandbox.io/s/nice-worker-d5qpi). There is a compile error when viewing in Codesandbox, however it will work locally. In Codesandbox go `File` > `Export to zip` to download the example.

### Application/website

Carbon for IBM.com

### Business priority

Medium Priority = upcoming release but is not pressing

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- Create with Shadow DOM and Custom Elements standards
- **See the Epic (#7267 ) for the Design and Functional specs information**
- Example:  React version [example](https://codesandbox.io/s/nice-worker-d5qpi?file=/src/App.js).
- Prod QA and Visual QA will be done after the scoped search issue (#7627 ) is finished.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

#### Acceptance criteria
- [ ] Include README for the web component and corresponding styles
- [ ] Create Web Components styles in styles package
- [ ] Do not create knobs in Storybook that include JSON objects
- [ ] Break out Storybook stories into multiple variation stories, if applicable
- [ ] Create codesandbox example under `/packages/web-components/examples/codesandbox` and include in README
- [ ] Minimum 80% unit test coverage
- [ ] Update the Carbon for ibm.com website component [data file](https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/blob/master/src/data/components.json) to be sure the web site Component Status and Storybook links are correct
- [ ] Use the [Visual QA checklist](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/wiki/Definition-of-done-(Visual-QA-checklist)) to verify design quality
- [ ] Update the e2e test scripts, if needed. 
- [ ] Confirm e2e test results have no major errors.

PR #8004, opened 2022-01-12
feat(masthead): add grouped search support

IgnacioBecerra:
### Related Ticket(s)
#7822

### Description
This PR tests the feature under the `With custom search API` story under Masthead. It hangs for a few seconds (due to the API itself) but the result suggestions appear underneath afterwards.

### Changelog

**New**

- add support for grouped search


packages/web-components/src/components/search-with-typeahead/search-with-typeahead.ts
```diff
@@ -248,9 +260,20 @@ class DDSSearchWithTypeahead extends HostListenerMixin(StableSelectorMixin(BXDro
   @HostListener('focusout')
   // @ts-ignore: The decorator refers to this method but TS thinks this method is not referred to
   protected _handleFocusOut(event: FocusEvent) {
-    super._handleFocusOut(event);
+    //super._handleFocusOut(event);
```
kennylam, line 263: Can remove.

packages/web-components/src/components/masthead/__stories__/README.stories.mdx
```diff
@@ -111,6 +111,69 @@ To set a custom endpoint for the translation service to fetch from, set either t
 <dds-masthead-container data-endpoint="/custom-endpoint"></dds-masthead-container>
 ```
 
+## Using custom search with typeahead API
+
+Using a custom search API than the default provided from IBM Search is supported, as well as separating any existing groups into its separate sections.
```
kennylam, line 116:
```suggestion
Using a search API other than the default provided from IBM Search is supported, as well as separating any existing groups into its separate sections.
```

packages/web-components/src/components/masthead/__stories__/README.stories.mdx
```diff
@@ -111,6 +111,69 @@ To set a custom endpoint for the translation service to fetch from, set either t
...
+
+First, pass in the `custom-typeahead-api` boolean into the Masthead.
```
kennylam, line 118:
```suggestion
First, add the `custom-typeahead-api` attribute to the masthead and set to `true`.
```

kennylam (commented):
This is great, thanks @IgnacioBecerra! Some small comments. Also, it doesn't seem clear that custom API and multiple search sections are separate features, which they are. Meaning it is entirely possible for a custom API to return the default, single-sectioned results. Can those features be separated in the docs and more clearly defined? Thanks!

oliviaflory:
@IgnacioBecerra This is very cool! I did notice a few things I'd love to get your thoughts on:

PS. I am not able to see the React example in CodeSandbox (I get an error) So I'm not able to confirm if these behaviors are existing in the React version.
Screen Shot 2022-01-13 at 9 16 34 AM


### 1. Results are changing 
I saw you mentioned the results hang a bit, they show up ~10 seconds after I type something, however I noticed the results are changing ~20 seconds after populating. This is a bit strange because sometimes the results change to something that has nothing to do with what I typed in the search field. I tried it with `Wat` `Watson` and `Cloud` (wondering if the partial word is doing something strange). Because they're 20+ seconds I have to attach as videos.

Example: Search `Watson` after 20 seconds results switch to `WebSphere`
https://user-images.githubusercontent.com/45692622/149346800-2ff01200-56ae-434f-bd4b-7c15d38886aa.mp4

Example: Search `Cloud` theres a mix of results that don't have Cloud, then end result does end up with `Cloud` but I noticed it updates some of the results towards the end of the video

https://user-images.githubusercontent.com/45692622/149347127-a4bc25cc-62f9-48ed-b271-f9c4320ab04f.mp4


 ### 2. Product pages interactive state
I wasn't expecting the Product pages title to have a hover state because I didn't think it was a link, can we remove that since it's not a link to anywhere? It does skip on keyboard navigation 👍 
Screen Shot 2022-01-13 at 9 01 27 AM

### 3. First section not highlighting matching results
When searching `Clo` the first section of the grouped type ahead results are not highlighting the `clo`. Is this expected?
Screen Shot 2022-01-13 at 11 31 39 AM

### 4. The delay
You mention that `It hangs for a few seconds (due to the API itself) but the result suggestions appear underneath afterwards.` I realize we might not have much control over this if the issues is in the API itself, but is it hanging for 10 seconds for you like it is for me (see videos above)? Just seems like a lot time and doesn't seem quite as nice as the non-grouped search. Just curious what you think!

IgnacioBecerra:
@oliviaflory @kennylam Addressed your feedback! Unfortunately the API being slow is the cause of the delay issues we see in the Storybook example, but with a fast API the results should be the same speed as our regular search.

oliviaflory:
@IgnacioBecerra I think everything looks good from a design perspective. 

Last bit I just want to comment on: In Safari on my iPhone I am able to scroll past the final result and see a white bit of space, and there is no horizontal rule on the last result. Because we don't see any indication of the "end" it sort of looks like we are missing results. 

When I use the Chrome app on my iPhone I am not able to scroll past the final result and we see a final divider line + just a bit of white background.  

Screen shot:

| Safari | Chrome |
| --- | --- |
| IMG_5969 | IMG-5970 |


Video example:
https://user-images.githubusercontent.com/45692622/149594509-63cacdfa-92c8-4d25-8018-eba733a61a46.mp4

oliviaflory:
@IgnacioBecerra I'm having a little trouble scrolling in Safari on my iPhone. I can definitely get there after scrolling a few times, so I can approve, but wondering if we can make it more smooth somehow, what do you think?

https://user-images.githubusercontent.com/45692622/150397391-1edbe897-7598-424f-b245-a136e6def29e.mp4

oliviaflory (approved):
Thanks @IgnacioBecerra !

packages/web-components/src/components/masthead/__stories__/masthead.stories.ts
```diff
@@ -87,6 +102,55 @@ export const Default = ({ parameters }) => {
...
+    <style>
+      ${styles}
+    </style>
+    <dds-masthead-composite
+      platform="${ifNonNull(platform)}"
+      .platformUrl="${ifNonNull(platformData.url)}"
+      selected-menu-item="${ifNonNull(selectedMenuItem)}"
+      user-status="${ifNonNull(userStatus)}"
+      searchPlaceholder="${ifNonNull(searchPlaceholder)}"
+      .authenticatedProfileItems="${ifNonNull(authenticatedProfileItems)}"
```
ariellalgilmore, line 124: Looks great! Just one minor thing. Can we add `navLinks` to the story
Screen Shot 2022-01-24 at 12 06 15 PM



```suggestion
      .navLinks="${navLinks}"
      .authenticatedProfileItems="${ifNonNull(authenticatedProfileItems)}"
```

ariellalgilmore (approved):
LGTM!

jeffchew (approved):
And adding e2e tests with a new feature, love it!

PR #8014, opened 2022-01-13
fix(yarn): rebuild yarn.lock to fix mismatched versions

kennylam:
### Description

This rebuilds `yarn.lock` to fix potential mismatched versions from various dependency updates which have introduced breaking changes to Storybook, specifically `addon-docs` in the React package (story docs tab throws an error).

### Changelog

**Changed**

- rebuild `yarn.lock`
- add `jsx-a11y/role-supports-aria-props` to `MastheadLeftNav.js`. This linting error appeared after `yarn.lock` was rebuilt probably due to minor `eslint` updates. The linting error refers to `aria-haspopup` being used on a role link. Added comment to suppress error for now and will address in separate issue.


kennylam:
Closing in favor of https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/8019

Issue #7899, opened 2021-12-17
Web component: Card - Pictogram - Screen reader does not announce content change that happened on focus.


Praveen-sr:
### Description

Issue: 
When we navigate with tab/shift tab, On focus mentioned content is displayed However Screen reader does not announce anything hence the user would not be aware of the content change that happened on focus.

Expected result:
1. provide button role for the element that triggers the content on focus
2. Provide appropriate label
3. Associate the content using aria-describedby

### Component(s) impacted

![image](https://user-images.githubusercontent.com/60002341/146518624-a2983c32-4344-4357-80d2-212a6ab55515.png)

### Browser

Firefox

### Carbon for IBM.com version

1.28

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

Web component

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

NA

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #8031, opened 2022-01-13
fix(card): announce content change on keyboard focus

emyarod:
### Related Ticket(s)

#7899

### Description

This PR updates the pictogram card content area so that screen readers will announce content changes on focus. The `aria-describedby` attribute will also point to the card copy container element


IgnacioBecerra (approved):
Tested in JAWS and Mac VoiceOver, works as expected 👍

Issue #8034, opened 2022-01-14
[Leadspace]: Component becomes scrollable


IgnacioBecerra:
### Description

Noticed that sometimes the Leadspace with Image can be scrollable. It can be appreciated better in this NextJS test page:
https://ibmdotcom-nextjs-test.mybluemix.net/example-page-c.html

https://user-images.githubusercontent.com/24970122/149581625-80e953cb-85c1-4cc8-abb6-90fc85ced5db.mov

### Component(s) impacted

Leadspace with Image

### Browser

Chrome

### Carbon for IBM.com version

canary

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

N/A

### Package

@carbon/ibmdotcom-react, @carbon/ibmdotcom-web-components

### CodeSandbox example

https://ibmdotcom-nextjs-test.mybluemix.net/example-page-c.html

### Steps to reproduce the issue (if applicable)

1. Place cursor on leadspace with image on a scrollable/long page
2. scroll down
3. Leadspace will move up within its own container instead of scrolling down to the rest of the page

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #8035, opened 2022-01-14
fix(leadspace): fix scrollable component

IgnacioBecerra:
### Related Ticket(s)
#8034

### Description
The Leadspace currently has a bug where, if the image is big enough, it can be scrolled within its own component. This means that if a user tries to scroll a bug when the cursor is on the leadspace, the leadspace will out of view. See the issue for a video with the bug.

You can see this in any Leadspace story containing an image (e.g. Leadspace Super with Image), though in the deploy previews here it will be fixed and unscrollable.

### Changelog

**Changed**

- `overflow-x` to `overflow` to account for the `y` overflow


emyarod (approved):
looks good to me!

RichKummer (approved):
Double checked it, and looks like the scroll is fixed. Great catch!

PR #8060, opened 2022-01-18
fix(github-actions): move slack webhook url to under slack action

jeffchew:
### Related Ticket(s)

No related issue

### Description

The slack webhook as a global environment variable was getting burned into the React artifacts when building to Github Pages. This exposes the secret which isn't desired.

This PR moves the environment variable specifically under the Slack action in the workflow to avoid this.

### Changelog

**Changed**

- `deploy-canary.yml`
- `deploy-next.yml`


emyarod (approved):
looks good to me!

PR #8062, opened 2022-01-18
fix(github-actions): restricts e2e tests on release branches for next

jeffchew:
### Related Ticket(s)

No related issue

### Description

Noticed that the browserstack e2e tests were failing in absence of a release branch. This adds a condition to only run if a release branch exists.

### Changelog

**Changed**

- add `startsWith` function check to the if statement in `e2e-tests-browserstack-next.yml`


ariellalgilmore (approved):
LGTM!

Issue #7798, opened 2021-12-01
[CTA Section w/ Link List]: heading,copy, and background has no color token applied


annawen1:
### Description

The CTA Section with link list appears fine in Storybook, however the `<dds-content-section-heading>` and the `dds-content-block-copy>` are displaying with blue text instead of white. It is blue admittedly because I'm wrapping it in an `<a>` element, but the issue still remains that there are no defined color tokens for the heading and copy (in the dark theme, the text is not visible as it remains black).

<img width="1440" alt="Screen Shot 2021-12-01 at 5 38 57 PM" src="https://user-images.githubusercontent.com/54281166/144325840-b2ec4f04-446b-454f-9d10-60296fb4da1e.png">

This bugs appeared with setting up the carbon for ibm.com web components tutorial step 2. Deploy preview in this PR shows the issues: https://github.com/carbon-design-system/carbon-for-ibm-dotcom-tutorial/pull/2

Issues:
1. I had to add `slot="heading"` to `<dds-content-section-heading>` in order for it to appear correctly in the layout (this isn't done for the storybook code or in the documentation strangely)
2. the heading & copy text are blue
3. I had to set the `background-color` of the component in the application scss file in order for the theming to work, we should ideally set the background color in the component style sheet instead.

### Component(s) impacted

CTA Section

### Browser

Chrome

### Carbon for IBM.com version

v1.12.0

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

Carbon for IBM.com WC Tutorial 

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://github.com/carbon-design-system/carbon-for-ibm-dotcom-tutorial/pull/2

### Steps to reproduce the issue (if applicable)

Go to https://github.com/carbon-design-system/carbon-for-ibm-dotcom-tutorial/pull/2 deploy preview

For further troubleshooting, you can checkout my branch/view the code in github

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #8079, opened 2022-01-19
fix(styles): missing text color for content block and section

ariellalgilmore:
### Related Ticket(s)

#7798

### Description

When themeing CTA Section in the tutorial the content-section-heading and content-block-copy are missing text color and are not visible

Screen Shot 2022-01-19 at 12 19 43 PM

### Changelog

**New**

- add missing text color


IgnacioBecerra (approved):
Thanks @ariellalgilmore!

Issue #8078, opened 2022-01-19
[Locale Modal]: Web Components - Locale Modal disappears when tapping into geographic area


annawen1:
### Description

When user opens the locale modal and taps into a geographic area, the modal suddenly closes.

This is on Safari, iPhone 12

![locale-modal](https://user-images.githubusercontent.com/54281166/150201085-48973929-b575-416c-bf90-661f13e6b74a.gif)

### Component(s) impacted

Locale Modal

### Browser

Safari

### Carbon for IBM.com version

1.30.0-rc.0

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

https://ibmdotcom-web-components-html-staging.mybluemix.net/example-page-b/index.html

### Package

@carbon/ibmdotcom-web-components

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #8083, opened 2022-01-19
fix(dotcom-shell): stop intersection logic if modal is open

IgnacioBecerra:
### Related Ticket(s)
#8078

### Description
There was some previous logic to avoid this bug from happening, but some check may have slipped through that made it reoccur. Instead of adding a check to every conditional, this PR only adds a simple check if the locale modal is open, and stops any code that might make the browser "scroll" in the `handleIntersection` function and would make the modal disappear. 

This needs to be tested in an actual phone (or Browserstack).

### Changelog
**Changed**

- changed location of the conditional check for the local modal being open to stop it from disappearing


jeffchew:
@IgnacioBecerra looks like some prettier issues, can you take a look?

annawen1 (approved):
Testing on iPhone 12, looks good!

RichKummer (approved):
Checked on iPhone 11, looks good!

Issue #7627, opened 2021-11-08
[Masthead, search] Web Component: Change the component to include the scoped search feature


RobertaJHahn:
#### User Story

> As a `[user role below]`:
Carbon for ibm.com developer

> I need to:
create/change the `Masthead, search`

> so that I can:
provide the ibm.com adopter developers components they can use to build ibm.com web pages

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- Create with Shadow DOM and Custom Elements standards
- **See the Epic (#7267 ) for the Design and Functional specs information**
- This issue is blocked by the issue (#TBD) to include the Grouped results (included in the React version)
- The widget will allow self service style scoping of search results by taking in explicit and implicit parameters in order of priority
- Scoping can be done via different types of parameters, so we need a param for scoping "type" and one for "value"
- Explicit scoping parameters overrule the implicit ones, so a user can set the type and value explicitly and that will be what is used
- For explicit scoping types/values, the masthead search widget will not keep an allow list on the allowed scoping values for each type, and maybe even which explicit types will be allowed.  The widget should be agnostic and the explicit scoping type/value communication is between the page owner and the search app
- Web Components Visual QA testing issue (#7628)
- Web Components Prod QA testing issue (#7629)

#### Acceptance criteria
- [ ] Include README for the web component and corresponding styles
- [ ] Create Web Components styles in styles package
- [ ] Do not create knobs in Storybook that include JSON objects
- [ ] Break out Storybook stories into multiple variation stories, if applicable
- [ ] Create codesandbox example under `/packages/web-components/examples/codesandbox` and include in README
- [ ] Minimum 80% unit test coverage
- [ ] Update the Carbon for ibm.com website component [data file](https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/blob/master/src/data/components.json) to be sure the web site Component Status and Storybook links are correct
- [ ] Use the [Visual QA checklist](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/wiki/Definition-of-done-(Visual-QA-checklist)) to verify design quality
- [ ] Update the e2e test scripts, if needed. 
- [ ] Confirm e2e test results have no major errors.
- [ ] If a design is provided, the Designer is included as a Reviewer in the Pull Request
- [ ] Provide a direct link to the deploy preview for the designer in the Pull Request description
[...]

PR #8096, opened 2022-01-21
feat(search-with-typeahead): add scoped search support

IgnacioBecerra:
### Related Ticket(s)
#7627

### Description
This PR changes the SearchWithTypeahead API slightly so it's able to retrieve scoped results from different categories within IBM, such as PartnerWorld or Developer as examples. This can be tested under the `Masthead - with scoped results` in WC Storybook. Change the scope by clicking on the filter select button while the results are displayed to see how they change between scopes.

### Changelog

**New**

- scoped search in search with typeahead

**Changed**

- added scope to `SearchWithTypeahead` API query if requested


kennylam (commented):
Looks good @IgnacioBecerra! Just a few comments.

RichKummer:
Hey I took a look and for the most part it's looking really good! I noticed that in the design specs, the comments say that the browser / device default would be used for the selector. Is that what it's using right now? I just want to confirm, the specs used the spinner (no need to change if the dropdown is right).

**Design specs**
Screen Shot 2022-01-27 at 6 23 21 PM

**iPhone 11**
IMG_1995 2

Also, when I tested it on my iphone/tablet, the selector looked fine, but when I tried the `sm` and `md` sizes in browser in Safari, the selector didn't populate with any text. I'm not sure a lot of people will hit that issue but it caught my eye as it was going through:

**Chrome - Macbook**
Screen Shot 2022-01-27 at 5 53 53 PM

**Safari - Macbook**
Screen Shot 2022-01-27 at 5 53 33 PM

IgnacioBecerra:
@RichKummer It seems that the wheel has been changed in the latest iOS, but the functionality remains. As for the Safari thing you noticed, I was able to fix it in the latest commit 👍  Nice job catching that!

kennylam (approved):
LGTM thanks @IgnacioBecerra!

RichKummer (approved):
Thanks for fixing the Safari issue, LGTM!

Issue #8098, opened 2022-01-21
[tabs-extended]: tooltip not appearing when there is not enough text


ariellalgilmore:
### Description

The text is being truncated but the tooltip isn't being set yet 
![image](https://user-images.githubusercontent.com/20210594/150555239-fef1f960-6cda-49fd-83c8-19aefebd3247.png)

### Component(s) impacted

tabs-extended

### Carbon for IBM.com version

latest

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

privacy

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://codesandbox.io/s/eloquent-easley-kjfzp?file=/src/index.js

### Steps to reproduce the issue (if applicable)

hover over tabs-extended with truncated text and you don't see a tooltip.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #8099, opened 2022-01-21
fix(tabs-extended): tooltip calculation

ariellalgilmore:
### Related Ticket(s)

#8098

### Description

Tooltip calculation needed updates. If there wasn't enough text the text would get truncated, but no tooltip would appear.

Screen Shot 2022-01-21 at 8 22 00 AM

### Changelog

**New**

- tooltip calculation
- added min amount of text for tooltip to appear in storybook on second tab


kennylam (approved):
Thanks @ariellalgilmore!

Issue #8105, opened 2022-01-21
[Card static]: React wrapper – card static has no content, knobs not working


oliviaflory:
### Description

The Card static component has no content displayed in Storybook and none of the knobs are working in storybook.

**v1.15.0 rc 0**
<img width="1345" alt="Screen Shot 2022-01-21 at 3 43 23 PM" src="https://user-images.githubusercontent.com/45692622/150598041-cea501e9-ce3d-4479-b877-9f99cfb351a6.png">

**Production v1.14.1** 
<img width="1351" alt="Screen Shot 2022-01-21 at 3 44 14 PM" src="https://user-images.githubusercontent.com/45692622/150598062-68c296be-91dc-4355-bddf-05681817becc.png">

### Component(s) impacted

Card static

### Browser

Chrome

### Carbon for IBM.com version

React wrapper v1.15.0 rc 0

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

Carbon for IBM.com

### Package

@carbon/ibmdotcom-react

### CodeSandbox example

https://carbon-design-system.github.io/carbon-for-ibm-dotcom/next/web-components-react/?path=/story/components-card--static

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #8106, opened 2022-01-21
fix(static-card): react wrapper story not showing

ariellalgilmore:
### Related Ticket(s)

#8105 

### Description

update storybook variable for static card. so static card renders properly


jeffchew (approved):
LGTM!

PR #8142, opened 2022-01-25
fix(masthead): fix undefined error in locale call

jeffchew:
### Related Ticket(s)

No related issue

### Description

There are a number of `Uncaught (in promise) TypeError: Cannot convert undefined or null to object` errors happening in the masthead/footer. This is an attempt to fix this.

Need to run the integration tests to see if this fixes the issues.

### Changelog

**Changed**

- Roll back `axios` to `~0.21.4`


ariellalgilmore (approved):
LGTM!

Issue #8085, opened 2022-01-19
[Card-CTA]: Aria-label reads out "Duration" even if there's no video to play


IgnacioBecerra:
**Engineering info:**  
- Innovation Team engineer: @mwanberg 
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2372
- DPO consulting engineer: 

----

### Description

Using VoiceOver, it reads out the Card's content fine, but it also reads out `"DURATION"`, which should only be read if there is a video attached to the Card.

Originally seen in `Card Group`'s `Card-in-card` story, but this actually comes from being extended from `CardCTA`.

This might be where the issue comes from:
https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/cta/card-cta.ts#L146

<img width="904" alt="Screen Shot 2022-01-19 at 3 03 59 PM" src="https://user-images.githubusercontent.com/24970122/150232471-4b0c0991-2a16-4c76-b75c-c86f4515f6d6.png">

### Component(s) impacted

Card-in-card
CardCTA

### Browser

Chrome

### Carbon for IBM.com version

1.30.0-rc.0

### Severity

Severity 4 = The problem is not visible to or noticeable to an average user. Affects minor functionality, no workaround needed.

### Application/website

N/A

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

N/A

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #8145, opened 2022-01-25
fix(card-cta): adjusting the conditional aria-label video duration

mwanberg:
### Related Ticket(s)

Github: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8085

### Description

Using VoiceOver, it read out the Card's content fine, but it also read out "DURATION", which should only be read if there is a video attached to the Card.

150232471-4b0c0991-2a16-4c76-b75c-c86f4515f6d6

I updated the logic to have the duration text be empty by default and only populated when there's a video.

### Changelog

**Changed**

- `ariaDuration` logic in `card-cta.ts`


ariellalgilmore (approved):
LGTM! Thanks @mwanberg

proeung (approved):
LGTM as well!

andy-blum (approved):

Issue #7898, opened 2021-12-16
[Footer]: e2e test failure: sporadic errors with footer short - DPO Consulting


jeffchew:
**Engineering info:**  
- Innovation Team engineer: @andy-blum 
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2363
- DPO consulting engineer: @IgnacioBecerra 

----

### Description

The following appears to be happening on certain tests for the footer in Browserstack:

![image](https://user-images.githubusercontent.com/1641214/146444808-17110631-e280-4037-a7fa-3ce2b92a6327.png)
![image](https://user-images.githubusercontent.com/1641214/146444526-b492cbc2-e1b7-4bbb-9848-3da269a10b9f.png)

### Component(s) impacted

Footer

### Browser

Chrome, Microsoft Edge

### Carbon for IBM.com version

Canary

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

Browserstack Automate

### Package

@carbon/ibmdotcom-react

### CodeSandbox example

https://automate.browserstack.com/dashboard/v2/builds/06deea8134cab157ea53ad6b92ff518c9329dcf0

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

andy-blum:
@jeffchew I don't know what's causing these errors, but I'm confident there's nothing wrong with the way the test is written.

In both cases, the tests properly asserted all 4 links that are loading from the storybook data [here](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/react/src/components/Footer/__data__/footer-thin.json), and that data hasn't changed since February 2020.

```js
cy.get('[data-autoid="dds--footer-legal-nav"]')
      .find('li a')
      .each($link => {
        const url = $link.prop('href');
        expect(url).not.to.be.empty;
      });
```

The way the test fails it seems to think there's a fifth element matching the selector `[data-autoid="dds--footer-legal-nav"] li a`, but that shouldn't be present at all.

IgnacioBecerra:
[...]

PR #8147, opened 2022-01-25
fix(e2e): mock ibm-common.js in e2e tests

jeffchew:
### Related Ticket(s)

Refs https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7898

### Description

The footer tests were still failing for the footer in Browserstack automate. This is an attempt to approach the fix in a different way, and instead to block the cookie preferences (ibm-common.js) from loading entirely. 

This also introduces the approach for globally mocking network calls rather than doing it in a per-test approach.

### Changelog

**New**

- `ibm-common.js` fixture in e2e tests

**Changed**

- moved cypress fixtures in global support index.js


packages/react/tests/e2e-storybook/cypress/integration/CTA/Feature.e2e.js
```diff
@@ -172,7 +172,7 @@ describe('dds-feature-cta | (desktop)', () => {
   });
 
   it('Should load and be fully clickable', _tests.checkBlockLink);
-  it(
+  xit(
```
emyarod, line 175: should we add a note about this omitted test?

emyarod (approved):
looks good to me, just one question about a disabled test

packages/react/tests/e2e-storybook/cypress/integration/CTA/Feature.e2e.js
jeffchew: Ahh just saw your note after the PR was merged. I'll open up a PR with a note.

jeffchew: Actually let me create an issue for this too.

Issue #8033, opened 2022-01-14
[Cloud-Masthead]: add services endpoint to determine auth status for Cloud Console - DPO Consulting


proeung:
**Engineering info:**  
- Innovation Team engineer: @proeung 
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2377
- DPO consulting engineer: @annawen1 

-----

### The problem

- It was brought to our attention that the Cloud console team has completed a feature to detect and personalize based on if a person was authenticated/logged-on on via the `cloud.ibm.com` console.
- This detection method was based on a request to personalize `/cloud` pages with users' live login state, which can now be accomplished.
- @kjwhitla Can you let me know if this is correct from a business requirement perspective?

### The solution

Steps to verify the personalization functionality:

- Goto an IBM.com marketing page, like https://ibm.com/cloud
- Open the developer console
- type `_dl.ddo.user.segment.isCloudLoggedOn` (previously made a typo and omitted the _ in this command)
- The console output should read `true` or `false`
- Either log in or log out of `cloud.ibm.com` (change your state)
- Reload the https://ibm.com/cloud page
- Repeat step 3 and see the status updated

We’ll need to adjust our `CloudAccountAuth` script within the Services package of the Carbon for ibm.com.

### Application/website

AEM, Drupal

### Business priority

Medium Priority = upcoming release but is not pressing

### What time frame would this ideally be needed by (if applicable)

The timeframe of this work has not been determined yet.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

proeung:
Reopening this ticket as the solution added to this PR doesn't seem to be working on PROD. https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/8149

PR to revert the changes - https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/8792

proeung:
@annawen1 Just wanted to revisit this issue. Do you have any recommendations on how to best test this change (see above).

annawen1:
[...]

PR #8149, opened 2022-01-26
fix(cloud-masthead): replace cookie auth with window object status

proeung:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8033

### Description

- It was brought to our attention that the Cloud console team has completed a feature to detect and personalize based on if a person was authenticated/logged-on on via the `cloud.ibm.com` console.
- This detection method was based on a request to personalize `/cloud` pages with users' live login state, which can now be accomplished.
- Instead of using the cookie checking method to verify if a user has logged into `cloud.ibm.com`, we're now using the status object coming from the client-side.
- @kjwhitla Can you let me know if this is correct from a business requirement perspective?

### Changelog

**Changed**

- Replace the `checkCookie()` method with `checkPersonalization()`.

**Note:** Let's hold off on merging this PR down until we get confirmation from @kjwhitla.


packages/services/src/services/CloudAccountAuth/__tests__/CloudAccountAuth.test.js
```diff
@@ -14,7 +14,7 @@ describe('CloudAccountAuth cookie utility', () => {
       value: 'com.ibm.cloud.iam.LoggedIn.prod=1',
```
annawen1, line 14: probably need to update the mocking of the cookie in these two tests. Should be similar to the `DDO.test.js` tests

```
import root from 'window-or-global';

describe('CloudAccountAuth personalization utility', () => {
    beforeEach(function() {
        root._dl.ddo.user.segment.isCloudLoggedOn = true;
    });

 .....
 
}

File: packages/services-store/src/actions/cloudAccountAuthAPI.ts [hunk 2]
```diff
@@ -59,7 +59,9 @@ export function loadUserStatus(
 ): ThunkAction<Promise<UserStatus>, { cloudAccountAuthAPI: CloudAccountAuthAPIState }, void, CloudAccountAuthAPIActions> {
   return async dispatch => {
     const promiseStatus: Promise<UserStatus> =
-      authMethod === 'cookie' ? CloudAccountAuthAPI.checkCookie() : CloudAccountAuthAPI.checkAPI();
+      authMethod === 'cookie' || authMethod === 'personalization'
```
huntermacd, line 62: I'm guessing 'cookie' is left here so any instances of the masthead won't break when we merge. How big of a concern is that? Is it conceivable we could update all instances of the masthead to use 'personalization' instead of 'cookie' once we merge this so we can remove all cookie-related code?

packages/services/src/services/CloudAccountAuth/CloudAccountAuth.js
```diff
@@ -1,35 +1,33 @@
...
+ * Copyright IBM Corp. 2021, 2022
  *
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
  */
 import axios from 'axios';
-import Cookies from 'js-cookie';
+import root from 'window-or-global';
 
 /**
```
huntermacd, line 10: We can remove this cookie-related documentation.

packages/services-store/src/actions/cloudAccountAuthAPI.ts
proeung: @huntermacd That's correct! I left `cookie` here because it's being used on multiple applications and platforms so simply removing this value will have an adverse effect on those content ecosystems as we can't easily update the `authMethod` value across different adopters.

I think keeping it as it's, for now, should be fine and we can announce the use of the personalization method once we review this work with stakeholders and the seal of approval to merge these changes down.

packages/services/src/services/CloudAccountAuth/CloudAccountAuth.js
proeung: @huntermacd Where is this cookie-related documentation that you're referencing?

packages/services/src/services/CloudAccountAuth/__tests__/CloudAccountAuth.test.js
proeung: @annawen1 I've tried the suggestion that you've noted above, but it's failing. Maybe there's something that I'm missing?

```
import CloudAccountAuthAPI from '../CloudAccountAuth';
import root from 'window-or-global';

describe('CloudAccountAuth personalization utility', () => {
  it('should fetch the personalization window boolean and return the authenticated string', () => {
    root._dl.ddo.user.segment.isCloudLoggedOn = true;

    const loginStatus = CloudAccountAuthAPI.checkPersonalization();
    expect(loginStatus).toStrictEqual({ user: 'authenticated' });
  });

  it('should fetch the personalization window boolean and return the anonymous string', () => {
    root._dl.ddo.user.segment.isCloudLoggedOn = false;

    const loginStatus = CloudAccountAuthAPI.checkPersonalization();
    expect(loginStatus).toStrictEqual({ user: 'anonymous' });
  });
});
```

packages/services/src/services/CloudAccountAuth/CloudAccountAuth.js
annawen1: maybe this part? https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/8149/files#diff-f125cbf578fa155879790dfaa7d41016c75a0d15b871448d44cf53cc27b5a6ceR11

packages/services/src/services/CloudAccountAuth/__tests__/CloudAccountAuth.test.js
proeung: Resolved by https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/8149/commits/4a03d29ec62562d5a51f1539c42754dc5757b843

proeung:
@huntermacd @annawen1 Got confirmation from @kjwhitla via slack that we're good to replace the cookie method with this new window object status but ensuring that the authentication check doesn't affect existing ecosystems using the "cookie" value (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/8149#discussion_r794873020).

When you have a chance this week, can you both review this new authentication method and ensure that everything is working as expected. See the code sandbox using changes from this branch - https://codesandbox.io/s/github/proeung/carbon-for-ibm-dotcom/tree/fix/cloud-masthead-8033/packages/web-components/examples/codesandbox/components/cloud-masthead

packages/services/src/services/CloudAccountAuth/CloudAccountAuth.js
huntermacd: Yep, those lines.

huntermacd (changes requested):
Looks good to me! Just remove those cookie doc lines and you've got a :shipit: from me.

annawen1:
Speaking with Juan Leon, we should be using `digitalData.user.segment.isCloudLoggedOn` instead

Screen Shot 2022-02-04 at 11 03 32 AM

packages/services/src/services/CloudAccountAuth/__tests__/CloudAccountAuth.test.js
```diff
@@ -1,30 +1,41 @@
...
-describe('CloudAccountAuth cookie utility', () => {
-  it('should fetch the CloudAccountAuth cookie and return the authenticated string', () => {
-    Object.defineProperty(window.document, 'cookie', {
-      writable: true,
-      value: 'com.ibm.cloud.iam.LoggedIn.prod=1',
-    });
+describe('CloudAccountAuth personalization utility', () => {
+  it('should fetch the personalization window boolean and return the authenticated string', () => {
+    root.digitalData = {
+      ddo: {
```
annawen1, line 14: no need for the `ddo` here! we're reading from `digitalData.user.segment.isCloudLoggedOn`

annawen1 (approved):
LGTM!

huntermacd (approved):
:shipit:

Issue #8016, opened 2022-01-13
[Test Scenario]: Storybook Docs pages


jeffchew:
### Component and test scenario title

All Storybook docs tabs

### Test scenario steps

1. Open docs page for each storybook component
2. Ensure no exceptions are being thrown

PR #8150, opened 2022-01-26
test(storybook): add e2e test for storybook docs

annawen1:
### Related Ticket(s)

[Test Scenario]: Storybook Docs pages #8016

### Description

This PR adds e2e tests for the storybook docs tab - making sure there are no errors thrown.

### Changelog

**New**

- tests for React and WC storybook docs


ariellalgilmore (approved):
LGTM!

packages/web-components/tests/e2e-storybook/cypress/integration/storybook-docs/storybook-docs.e2e.js
```diff
@@ -0,0 +1,43 @@
...
+     */
+    const components = cy.get('a[id*="explorercomponents"]');
+
+    components.each($component => {
+      const docsPath = $component.prop('href');
+
+      if (docsPath) {
+        const url = docsPath.split('?')[1].replace('path=/story', 'path=/docs');
+
+        cy.visit(`?${url}`);
```
jeffchew, line 25: @annawen1 after messing around with Cypress a bunch, running a `cy.visit` multiple times in a single `it` test can cause some pretty bad memory issues. You should do an iteration and create multiple `it` tests instead.

jeffchew (changes requested):

packages/web-components/tests/e2e-storybook/cypress/integration/storybook-docs/storybook-docs.e2e.js
jeffchew: Case in point, Github Actions is pretty unhappy with this test:

image

annawen1: @jeffchew I ended up creating two tests, one to scrape all components from storybook and writing to the `fixtures/components.json` and then another one to import from that fixtures file to generate the tests. Looks like it's all passing! 🎆

jeffchew:
@annawen1 looks good! I think we can omit the snapshots though. They don't looks like they would be too useful anyway:

image

Checking if there are any errors coming might be enough I think.

annawen1:
> @annawen1 looks good! I think we can omit the snapshots though. They don't look like they would be too useful anyway:
> 
> image
> 
> Checking if there are any errors coming might be enough I think.


ok! I removed the snapshot commands from the tests!

jeffchew (approved):
LGTM!

PR #8153, opened 2022-01-27
fix(filter-panel): return to original heading after clearing filters

### LINKED ISSUE #8020: [Filter Panel]: Inconsistent Heading - DPO Consulting
**Engineering info:**  
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2376
- DPO consulting engineer: @IgnacioBecerra 

----

### Description

When I set the filter panel heading with `<dds-filter-panel-heading slot="heading">Programs by Category</dds-filter-panel-heading>` there is some inconsistency. 
After filtering it changes to "Filter (<nr. of selected filters>)". After clearing the filters it is "Filter" and won't change back to "Programs by Category". On my project this happens only in the modal view, but on the code sandbox it happens also in the default view. 
Also using a longer text for the heading does not look good in mobile view - the modal button icon becomes smaller:
image
 
It might make sense to not use a custom heading for the modal button and always use "Filter". Or allow to override it using an html attribute, because it should be translatable. 

### Component(s) impacted

https://www.ibm.com/standards/carbon/web-components/?path=/story/components-filter-panel--default

### Browser

Chrome

1.29.1

### Severity

### Application/website

IBM Afterburner Adobe CMS

### Package

@carbon/ibmdotcom-web-components, @carbon/ibmdotcom-styles

### CodeSandbox example

https://codesandbox.io/s/awesome-resonance-7wkge?file=/index.html

1. Have a different heading for the filter panel than "Filter"
2. Apply some filters using checkboxes
3. Clear the Filters
4. The heading is not the same as on start

_No response_

### Code of Conduct

jkaeser:
### Related Ticket(s)

Resolves #8020

### Description

The mechanism for rendering the header assumed the header would always be "Filter". It is a simple change to let this return the original header once no filters are set.

### Changelog

**Changed**

- Modified `renderStatus()` method to return the original header if no filters are set.


jkaeser:
Given the header is customized to "My Header" and I select a filter, should the header be "Filter (1)" or "My Header (1)"? Currently we show "Filter (1)".

IgnacioBecerra (approved):
LGTM!

proeung (approved):
LGTM as well!

jkaeser:
Thanks for the reviews, @proeung and @IgnacioBecerra!

Do you have any thoughts on my previous comment? It seems to me we'd be better off appending the filter count to the customized header, but I don't see anything about it in the specs.

proeung:
@oliviaflory Maybe you can help take a look? See @jkaeser's comment - https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/8153#issuecomment-1023324549

jeffchew:
@jkaeser looks like there are some prettier formatting issues, can you take a look?

image

PR #8156, opened 2022-01-28
fix(cta): add imports to be able to render cdn files

ariellalgilmore:
### Related Ticket(s)

none

### Description

not able to use certain cta types because they are missing in the `cta/index.js` file

### Changelog

**New**

- added 
```
import './link-list-item-card-cta';
import './link-list-item-cta';
import './video-cta-container';
```


proeung (approved):
LGTM! Thanks for adding these missing CTA types to the `cta/index.js` file!

jeffchew (approved):
LGTM!

Issue #8009, opened 2022-01-12
[a11y]: 'role' defined on the element is not valid per ARIA specification


ariellalgilmore:
**Engineering info:**  
- Innovation Team engineer: @proeung 
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2383
- DPO consulting engineer: N/A

-----

### Description

When running `yarn test:a11y` multiple stories are recieving an example of the following 'role' not defined error: 

<img width="785" alt="Screen Shot 2022-01-12 at 12 07 29 PM" src="https://user-images.githubusercontent.com/20210594/149213920-8e18ece9-30c4-48d4-b57a-c91a4a54564f.png">

Example output of all accessibility errors:
https://github.com/carbon-design-system/carbon-for-ibm-dotcom/runs/4779749308?check_suite_focus=true#step:7:291

### Component(s) impacted

- Card section images
  - `<dds-card-section-image>` 
- Card section simple, WithCTA
  - `<dds-card-section-simple>` 
- Content secion
  - `<dds-content-section>` 
- CTA block | With Content items
  - `<dds-cta-block-item-row>` 
- CTA section simple, withLinkList
  - `<dds-cta-section>` 
- Dotcom shell (all)
  - `<dds-cta-block-item-row>` 
- Link list section
  - `<dds-link-list-section>` 

### Carbon for IBM.com version

Canary

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

Carbon for IBM.com web components

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/runs/4779749308?check_suite_focus=true#step:7:2065

### Steps to reproduce the issue (if applicable)

View the actions tab and look under `Run a11y compliance tests` for web components

To run locally go to `packages/web-components` and `yarn test:a11y`

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #8157, opened 2022-01-28
fix(a11y): remove invalid role attribute to be a11y compliance

proeung:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8009

### Description

- This PR addresses the `'role' defined on the element is not valid per ARIA specification` error seen when running the `test:a11y` test for the following components.

  - `<dds-card-section-image>`
  - `<dds-card-section-simple>`
  - `<dds-content-section>`
  - `<cta-block-item-row>`

### Changelog

**Removed**

- Remove callback functions that add the role attribute.


ariellalgilmore (approved):
LGTM! Don't see it failing in the a11y test anymore

PR #8175, opened 2022-01-31
chore(deps): remove tsickle

jeffchew:
### Related Ticket(s)

No related issue

### Description

It doesn't appear that `tsickle` is being used, and rather than having it upgraded by renovatebot, this is a PR to remove it instead.

### Changelog

**Removed**

- `tsickle`


emyarod (approved):
looks good to me!

Issue #7804, opened 2021-12-02
[Content section]: aligning children to content-section heading


ariellalgilmore:
### Description

Expected children items to align correctly with the `content-section-heading`
<img width="1140" alt="Screen Shot 2021-12-02 at 9 11 42 AM" src="https://user-images.githubusercontent.com/20210594/144470306-8a1cbda5-4dc8-40f4-aa1d-a4f494318960.png">

### Component(s) impacted

Content Section with children (content block, content item, content group) need all to align to top properly when using `content-section`

### Browser

Chrome

### Carbon for IBM.com version

v1.12.0

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

Carbon for IBM.com WC Tutorial and broken in storybook

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://github.com/carbon-design-system/carbon-for-ibm-dotcom-tutorial/tree/web-components-step-3

### Steps to reproduce the issue (if applicable)

go to https://carbon-design-system.github.io/carbon-for-ibm-dotcom-tutorial/step-3/ to see issue. Code can be found here: https://github.com/carbon-design-system/carbon-for-ibm-dotcom-tutorial/tree/web-components-step-3. Also in storybook here: http://ibmdotcom-web-components-canary.mybluemix.net/?path=/story/components-content-section--default

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

kennylam:
@ariellalgilmore Can this not be achieved with CSS?

PR #8182, opened 2022-02-01
fix(content-section): update child elements alignment to the top

ariellalgilmore:
### Related Ticket(s)

#7804 

### Description

When a child element is added to content-section we need to remove the top of either padding or margin, so the element aligns to the content-section heading. 

Note: there are multiple style differences from the design. This PR though is just trying to get the headings aligned. When trying to fix the other styles (ie text and spacing), it was causing a lot of percy diffs. Thinking we should tackle one at a time. Will create an issue with the style discrepancies. 

### Changelog

**New**

- wc: slotchange
- react useEffect
- added cta-section styles, so that those headings still align


ariellalgilmore:
> @ariellalgilmore Can this not be achieved with CSS?

yes it totally can! went down a rabbit hole with this one. will update!

emyarod (approved):
looks good to me!

PR #8190, opened 2022-02-01
test(footer): change number of links to check for in chrome

IgnacioBecerra:
### Related Ticket(s)
None.

### Description
The Footer tests in Chome Big Sur seem to always render the proper 38 links, whereas other browsers render `Follow IBM` links twice. This PR ensures that the check for 38 only happens in Chrome.

### Changelog

**Changed**

- number of links checked in footer tests


emyarod (approved):
looks good to me! maybe we can rename the tests to no longer reference 41  as the total number of links too

Issue #7999, opened 2022-01-11
[NextJS Test]: Consistent false positives with Table of Contents


jeffchew:
### Description

We are consistently seeing false positive Percy reports for the highlight indicator of the Table of Contents on the NextJS test page:

![image](https://user-images.githubusercontent.com/1641214/148958465-b0483e74-d75c-4f00-8bb2-a56a77be1624.png)

### Component(s) impacted

Table of Contents

### Browser

Chrome

### Carbon for IBM.com version

Canary

### Severity

Severity 4 = The problem is not visible to or noticeable to an average user. Affects minor functionality, no workaround needed.

### Application/website

Carbon for IBM.com NextJS Test Application

### Package

@carbon/ibmdotcom-react

### CodeSandbox example

https://percy.io/538fc19a/Carbon-for-IBM.com-NextJS-Test-App-Upstream/builds/14936838/changed/844277403?browser=chrome&browser_ids=20&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1280&widths=375%2C1280

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

jeffchew:
Putting this back in pending scheduling, tried a few things and still seeing this issue.

Issue #7941, opened 2022-01-06
[Card in Card]: Clicking smaller card with a new tab link opens twice within Firefox


EAVilleda:
**Engineering info:**  
- Innovation Team engineer: @proeung 
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2369
- DPO consulting engineer: N/A

----

### Description

When the link is set with target="_blank", clicking on the small card within the bigger card will open the new tab twice.

![image](https://user-images.githubusercontent.com/47429622/148430921-d6356a28-8872-4752-b5b4-c76f56574ed9.png)

### Component(s) impacted

Card Group with Card in Card
Card in Card

### Browser

Chrome

### Carbon for IBM.com version

v1.29.1

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

AEM (Marketing)

### Package

@carbon/ibmdotcom-web-components, @carbon/ibmdotcom-styles

### CodeSandbox example

No codesandbox avaiable

### Steps to reproduce the issue (if applicable)

https://www.ibm.com/standards/carbon/web-components/iframe.html?id=components-card-in-card--default
Set the target="_blank" on dds-card-in-card tag.
Click on the smaller card within the bigger card.
Two new tabs open. 

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

jeffchew:
Hi @EAVilleda , we followed the steps and couldn't reproduce the issue. Are there additional details you can provide to reproduce the issue?

proeung:
@EAVilleda The Card in Card component extends the `DDSCardCTA` component (see - https://www.ibm.com/standards/carbon/web-components/?path=/story/components-cta--card), therefore the external/new tab behavior should be the same as what you're seeing within the CTA Card preview.

Are you manually adding the `target="_blank"` within the shadow DOM, if so I wonder if that causes the link to open twice? Let us know if there's something that we're missing here.

proeung:
[...]

PR #8191, opened 2022-02-02
fix(card-in-card): add null check for card-in-card-image

jeffchew:
### Related Ticket(s)

Related to https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7941, https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7999

### Description

This will hopefully fix the occasional exceptions occurring around the `card-in-card` component within the NextJS test page.

### Changelog

**Changed**

- Add null check for the `dds-card-in-card-image` element


emyarod (approved):
looks good to me!

PR #8201, opened 2022-02-03
test(masthead): mock grouped and scoped search suggestions

IgnacioBecerra:
### Related Ticket(s)
None.

### Description
This PR mocks search suggestion results for both the grouped and scoped search API to avoid relying on live data for tests.

### Changelog

**New**

- added `scoped-typeahead.json` and `grouped-typeahead.json` Cypress fixtures

**Changed**

- intercepted grouped/scoped typeahead API calls in the tests to use the mock fixtures instead


emyarod (approved):
looks good to me!

Issue #8017, opened 2022-01-13
[Masthead]: unable remove profile from masthead


annawen1:
**Engineering info:**  
- Innovation Team engineer: @proeung 
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2391
- DPO consulting engineer: N/A

-----

### Description

When using the web-components masthead and trying to disable the profile login, documented method does not work

`<dds-masthead-container has-profile={false}></dds-masthead-container>` 

### Component(s) impacted

dds-masthead

### Carbon for IBM.com version

latest

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

N/A

### Package

@carbon/ibmdotcom-web-components

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #8202, opened 2022-02-03
fix(masthead): change hasProfile to accept a string instead of boolean

proeung:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8017

### Description

- This PR addresses the issue for adopters using the `has-profile="false"` to optionally hide the "Profile" button.

### Changelog

**Changed**

- This PR sets the `has-profile` prop to accept a string instead of a boolean.

### Testing Instruction

- Spin up a sandbox example of the Masthead component - https://codesandbox.io/s/github/preoung/carbon-for-ibm-dotcom/tree/fix/masthead-8017/packages/web-components/examples/codesandbox/components/masthead
- Set `has-profile` attribute sets to `false`.
- Ensure that the "Profile" button is removed when the attribute is set to `false`. If the `has-profile` is not defined or set to `true`, the "Contact us" button should be set visible.
```
<dds-masthead-container
  has-contact="false"
></dds-masthead-container>
```

### DIFF
--- a/packages/web-components/src/components/masthead/__stories__/cloud-masthead.stories.ts
+++ b/packages/web-components/src/components/masthead/__stories__/cloud-masthead.stories.ts
@@ -1,7 +1,7 @@
 /**
  * @license
  *
- * Copyright IBM Corp. 2021
+ * Copyright IBM Corp. 2021, 2022
  *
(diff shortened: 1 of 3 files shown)


annawen1 (approved):
This looks good to me!

PR #8229, opened 2022-02-07
chore(deps): remove `event-target-shim`

jeffchew:
### Related Ticket(s)

No related issue

### Description

Renovatebot tried to upgrade this, but noticed that this isn't being used anywhere in the library. This removes this dependency from the repo.

### Changelog

**Removed**

- `event-target-shim`


emyarod (approved):
looks good to me!

PR #8234, opened 2022-02-08
chore(yarn): upgrade yarn to 1.22.17

jeffchew:
### Related Ticket(s)

No related issue

### Description

We are seeing connectivity issues with `yarn install` on our CI, this is an attempt to see if this might help at all.

### Changelog

**Changed**

- Upgrade yarn to `1.22.17`


.yarnrc
```diff
@@ -2,8 +2,8 @@
 # yarn lockfile v1
 
 
-version v1.19.0
-lastUpdateCheck 1597082923381
+version "1.22.10"
```
emyarod, line 5: is this supposed to match up with the release version in `yarn-path`?

emyarod (commented):
just 1 question about the version

.yarnrc
jeffchew: hmm yeah, this was auto generated and I set the policy to use 1.22.17. Not sure why it says 1.22.10...

emyarod: gotcha, I guess it should be fine since it was autogenerated and still the same minor version

emyarod (approved):
looks good to me if the version mismatch is just from yarn's autogenerating!

Issue #6814, opened 2021-08-05
[Universal banner] Web Component: Develop the Universal banner component


RobertaJHahn:
#### User Story

> As a `[user role below]`:
Carbon for ibm.com developer

> I need to:
create/change the `Universal banner`

> so that I can:
provide the ibm.com adopter developers components they can use to build ibm.com web pages

#### Additional information

- Story within Storybook with corresponding knobs
- Utilize Carbon
- Create with Shadow DOM and Custom Elements standards
- **See the Epic for the Design and Functional specs information**
- [Final design and design specs](https://ibm.ent.box.com/folder/142303458357?s=fas5qv5zrscudr63y85g28v0k657p83u)
- [Functional specs](https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/wiki/Universal-banner)
- The background color should be able to change.
- Include MH change for the behavior when Banner is present.  Func spec (#7620)
- Include a Knob in the dotcom shell storybook to allow for the Universal Banner.  
- Web Components Visual QA testing issue (#6816 )
- Web Components Prod QA testing issue (#6821 )

#### Acceptance criteria
- [ ] Include README for the web component and corresponding styles
- [ ] Create Web Components styles in styles package
- [ ] Do not create knobs in Storybook that include JSON objects
- [ ] Break out Storybook stories into multiple variation stories, if applicable
- [ ] Create codesandbox example under `/packages/web-components/examples/codesandbox` and include in README
- [ ] Minimum 80% unit test coverage
- [ ] Update the Carbon for ibm.com website component [data file](https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/blob/master/src/data/components.json) to be sure the web site Component Status and Storybook links are correct
- [ ] Use the [Visual QA checklist](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/wiki/Definition-of-done-(Visual-QA-checklist)) to verify design quality
- [ ] Update the e2e test scripts, if needed. 
- [ ] Confirm e2e test results have no major errors.
- [ ] If a design is provided, the Designer is included as a Reviewer in the Pull Request
- [ ] Provide a direct link to the deploy preview for the designer in the Pull Request description
- [ ] A comment is posted in the Prod QA issue, tagging Praveen when development is finished

IgnacioBecerra:
@jeffchew Moving this one back to Sprint 3 since it got approved on Friday to merge

PR #8237, opened 2022-02-08
feat(universal-banner): introducing component

IgnacioBecerra:
### Related Ticket(s)
#6814

### Description
Introducing the Universal Banner component, along with some scroll interactions between the Masthead, L1, and ToC mobile.

### Changelog

**New**

- new `universal-banner` component along with its child components

**Changed**

- scroll interaction algorithm in `dotcom-shell-composite`


RichKummer (commented):
@IgnacioBecerra this is looking really good! I noticed a few things combing through it:

1. **Spacing** - found a few places where spacing for the content is off

* At `max`, when there is no image, the start of the heading and copy should have more spacing on the left - should be 40px at max, currently it's 32px:

Screen Shot 2022-02-09 at 9 51 23 AM

* At `sm`, with or without an image, there's an extra 16px spacing between the heading and the CTA, and the copy and CTA (if copy appears):
Screen Shot 2022-02-09 at 9 58 27 AM

* When a 4-column image is in, there should be additional spacing between the heading and copy at `xlg` and `lg` (currently 8px, should be 16px)

2. **Button** - Button had a few grid issues at certain breakpoints, depending on whether an image was in or not.

* At `xlg`, it looks like the button is slightly off the grid with no image. The other breakpoints look good, this one for some reason is a little off:
Screen Shot 2022-02-09 at 1 15 34 PM

* When a 4-column image is in, the button at `max` looks off the grid. Same when using the 8-column image:
Screen Shot 2022-02-09 at 1 30 30 PM

3. **Type tokens**
* When the 8-column image is used, the heading and copy should use `body-short-02`, with the heading using semibold for `lg` through `max`. Currently, the heading is using `productive-heading-02` and the copy is using `expressive-heading-01`

Screen Shot 2022-02-09 at 1 57 09 PM
Screen Shot 2022-02-09 at 1 57 31 PM

4. **Minor updates**
* Update the default CTA copy to "Register for Think. Free"

RichKummer:
@IgnacioBecerra, checked out the latest - looks like a lot of the issues were fixed (yay!), just seeing a few remaining ones: 

1. I'm still seeing the spacing between heading and copy at `xlg` and `lg` as 8-px when it should be 16px when using the 4-column image. Same issue when there is no image.
2. At `xlg` and `lg` for all versions, I'm seeing the button text still slightly off the start of the column, but the other breakpoints look good. You had mentioned yesterday using the middle of the gutter, but the button padding is `15px` - I'm not sure if that accounts for the difference.

RichKummer:
@IgnacioBecerra seeing the spacing issue's fixed, and I think this is super close! I did see some last issues with the button on the grid

* At `xlg` with no image, the button with wrapped text looks like it's not aligning the text to the grid
Screen Shot 2022-02-11 at 10 10 08 AM

* At `xlg` and `lg` with the 8-column image (but not when using the 4-column), the button text looks like it's slightly off the grid
Screen Shot 2022-02-11 at 10 10 28 AM
Screen Shot 2022-02-11 at 10 10 37 AM

* I created a new image for the `lg` breakpoint for the 4-column image, I'll send that over Slack

RichKummer (approved):
Ran through it and looks like you fixed the grid issues 🎉🎉🎉. Everything looks good!

PR #8239, opened 2022-02-09
test(footer): ensure nav test passes

IgnacioBecerra:
### Related Ticket(s)
None.

### Description
Seems that even with the browser check, the navigation links in the Footer still seem to vary between 38 and 41 at times. This PR ensures that either of those values will be accepted so the tests can always pass, 

### Changelog

**Changed**

- using a different way to test for the nav links in the footer


emyarod (approved):
looks good to me!

jeffchew:
@IgnacioBecerra I think there's something going on with the mock data itself actually, also we didn't have a proper wait before loading the test. I'm opening up another PR that might help with some of the stabilization too.

PR #8241, opened 2022-02-09
fix(footer): stabilization of e2e tests

jeffchew:
### Related Ticket(s)

No related issue

### Description

The footer e2e tests were for some reason switching between 38 and 41 nav links in some of the tests. This seems to be coming from the fact that the `Follow IBM` links are getting duplicated.

This PR has some stabilization fixes, so will see if this addresses it.

### Changelog

**Changed**

- Separated out footer e2e tests into variant files
- Updated translation mock fixture data
- Inject mock translation and countrylist data into sessionStorage to ensure mocking
- Switch test to check for 38 links instead of 41
- Add `cy.waitUntil` for all footer e2e tests


emyarod (approved):
looks good to me!

PR #8244, opened 2022-02-09
fix(footer): change cypress intercept for translation to use raw data

jeffchew:
### Related Ticket(s)

No related issue

### Description

This is a followup to the cypress e2e tests for Footer. In some cases the sessionStorage isn't being used and the `cy.intercept` is caught instead. In this case, it should use the raw response from the translation repo.

### Changelog

**Changed**

- Changed fixture for translation API to use raw response data

### DIFF
--- a/packages/react/tests/e2e-storybook/cypress/support/index.js
+++ b/packages/react/tests/e2e-storybook/cypress/support/index.js
@@ -20,7 +20,7 @@ beforeEach(() => {
   // Mock the translation file
   cy.intercept(
     'https://1.www.s81c.com/common/carbon-for-ibm-dotcom/translations/masthead-footer/usen.json',
-    { fixture: 'translation.json' }
+    { fixture: 'translation-raw.json' }
   );
 
   // Mock the user status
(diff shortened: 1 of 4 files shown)


packages/react/tests/e2e-storybook/cypress/support/index.js
```diff
@@ -20,7 +20,7 @@ beforeEach(() => {
   // Mock the translation file
   cy.intercept(
     'https://1.www.s81c.com/common/carbon-for-ibm-dotcom/translations/masthead-footer/usen.json',
-    { fixture: 'translation.json' }
+    { fixture: 'translation-raw.json' }
```
emyarod, line 23:
```suggestion
    { fixture: 'translations-raw.json' }
```

packages/web-components/tests/e2e-storybook/cypress/support/index.js
```diff
@@ -27,7 +27,7 @@ beforeEach(() => {
 
   // Mock the translation file
   cy.intercept('https://1.www.s81c.com/common/carbon-for-ibm-dotcom/translations/masthead-footer/usen.json', {
-    fixture: 'translation.json',
+    fixture: 'translation-raw.json',
```
emyarod, line 30:
```suggestion
    fixture: 'translations-raw.json',
```

emyarod (approved):
looks good to me, just needs matching file names I believe

jeffchew:
Something is not jiving with the React tests, investigating...

image

jeffchew:
Turns out that Cypress didn't like the setting of `sessionStorage`, so that was removed and now fully relies on the raw translation fixture.

IgnacioBecerra (approved):

PR #8250, opened 2022-02-11
fix(cloud-masthead): update the services and utilities to use canary

annawen1:
### Description

We were getting an error that the `checkPersonalization()` method from `CloudAccountAuth` service was not defined. Updating the package versions for `services` and `utilities` resolves this issue.

### Changelog

**Changed**

- add optional chaining for cloud auth check for `digitalData.user.segment.isCloudLoggedOn`
- update services and utilities packages


jeffchew (approved):
Thanks @annawen1 !

proeung (approved):
LGTM!

PR #8251, opened 2022-02-11
fix(filter-panel): remove unneeded code

IgnacioBecerra:
### Related Ticket(s)
None.

### Description
Not sure if this will fix the wrapper issue but we'll see. This getter didn't have a header comment and we had ran into an issue before where the ci-checks were failing because another getter was lacking it. Long shot, we'll see if the wrapper renders.

### Changelog

**Removed**

- removed unneeded code


emyarod (approved):
looks good to me!

jeffchew:
@IgnacioBecerra do we need this one still?

cc: @emyarod

Issue #7541, opened 2021-10-29
[BUG: Card group]: Grey border in Safari


KarMartin:
**Engineering info:**  
- Innovation Team engineer: @jkaeser 
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2403
- DPO consulting engineer: 

-----

### Description

**Problem description**
The component "Card Group" does not displayer properly in Safari. 
It can be seen in variants "With card in card", "With card in card and image cards" and "With card Links".

CURRENT: (View in Safari (15))
![image](https://user-images.githubusercontent.com/93377090/139412228-e52a5cc7-41a1-4d12-90fd-e787e5b16a44.png)

EXPECTED: (View in Chrome (94) & Firefox (93))
![image](https://user-images.githubusercontent.com/93377090/139412397-d58c1b87-55ac-4526-96fc-98dba80c7a63.png)

**System used to reproduce**
- Macbook pro
- MacOS 11.6
- Safari 15
- Carbon for IBM.com Web Components v1.11.0: https://www.ibm.com/standards/carbon/web-components/?path=/story/components-card-group--with-card-in-card-and-image-cards

### Component(s) impacted

**Affected components:**
card-group--with-card-in-card
https://www.ibm.com/standards/carbon/web-components/?path=/story/components-card-group--with-card-in-card

card-group--with-card-in-card-and-image-cards
https://www.ibm.com/standards/carbon/web-components/?path=/story/components-card-group--with-card-in-card-and-image-cards

card-group--with-card-links
https://www.ibm.com/standards/carbon/web-components/?path=/story/components-card-group--with-card-links

Inspection:
![image](https://user-images.githubusercontent.com/93377090/139413503-9bd8fb8d-1473-412a-b81a-223c2a91be1b.png)

`<dds-card-group-card-link-item href="https://example.com" pattern-background="" data-autoid="dds--card-group-card-link-item" size="" color-scheme="" pictogram-placement="top" cta-type="">
    <dds-card-link-heading role="heading" aria-level="3" data-autoid="dds--card-link-heading" slot="heading">IBM Developer</dds-card-link-heading>
    <p>Learn, code and connect with your community</p>
    <dds-card-cta-footer slot="footer" data-autoid="dds--card-cta-footer" icon-placement="right" color-scheme="" cta-type="" parent-href="https://example.com" href="https://example.com">
[...]

Issue #8299, opened 2022-02-16
[Card group - default]: Extra border on Gray90 & Gray100


RichKummer:
### Description

In React wrapper: On the darker themes, a light border appears on the edge of the 3rd card in the row. This also shows up for `Card group - with card in card` in Gray90 & Gray100.

<img width="1571" alt="Screen Shot 2022-02-16 at 8 38 00 AM" src="https://user-images.githubusercontent.com/15643582/154276439-cd18881f-ecd4-47c5-b0fa-6ec7b8c0dbc3.png">

### Component(s) impacted

Card group

### Browser

Firefox

### Carbon for IBM.com version

v1.31.0-rc.1

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

C4IBM.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://codesandbox.io/s/github/carbon-design-system/carbon-for-ibm-dotcom/tree/main/packages/web-components/examples/codesandbox/components-react/card-group

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

proeung:
This issue looks to be similar to the one outlined in this one (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8335).

proeung:
Closing this issue in favor of this one (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8335), which will address some of the border regression bugs seen within the Card Group component.

PR #6934, opened 2021-08-19
feat(card-group): modify to allow Card link design pattern

proeung:
### Related Ticket(s)

DDS Consulting Issue: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6876
HC JIRA Ticket: https://jsw.ibm.com/browse/HC-2138

### Description
- This PR is a follow-up based on the discussion mentioned in this PR - https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/6826

### Design & Functional Specs
- https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/wiki/Tabs-extended#22-orientation 
- https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/wiki/Card-link

Screen Shot 2021-08-19 at 5 24 54 PM
Screen Shot 2021-08-19 at 5 25 00 PM

### Changelog

**New Components/Props**

- `pattern-background` - this prop/attribute is used when the card group item has the same background color as the pattern container.
- `<dds-card-link-heading>` - a component extended from `<dds-card-heading>` that captures the proper expressive heading styling for the `Card Link UI` component.

**Removed**

- N/A

### HTML (with Card Link)

```html
<dds-card-group grid-mode="narrow">
  <dds-card-group-item href="https://example.com" pattern-background>
    <dds-card-link-heading slot="heading">IBM Developer - Card 01</dds-card-link-heading>
    <p>Learn, code and connect with your community</p>
    <dds-card-cta-footer slot="footer">
      <svg slot="icon" focusable="false" preserveAspectRatio="xMidYMid meet" xmlns="http://www.w3.org/2000/svg" fill="currentColor" aria-hidden="true" width="20" height="20" viewBox="0 0 20 20">
        <path d="M11.8 2.8L10.8 3.8 16.2 9.3 1 9.3 1 10.7 16.2 10.7 10.8 16.2 11.8 17.2 19 10z"></path>
      </svg>
    </dds-card-cta-footer>
  </dds-card-group-item>
</dds-card-group>
```

### Testing Instructions
- Visit the `Card Group with Card Links` layout component - `/?path=/story/components-card-group--with-card-links`
- Ensure that the `Card links` look good within the `Card group` layout component in all breakpoints.
- Visit the `Tab Extended with Card Group Card Links` layout component - `/?path=/story/components-tabs-extended--with-card-group-card-links`
- Ensure that the "Vertical" tabs component matches with the design specs for all breakpoints.

PR #8320, opened 2022-02-16
fix(card-group): prevent extra top "borders" in Safari

jkaeser:
### Related Ticket(s)

Resolves https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7541

### Description

Once again, Safari doesn't handle web component selector specificity correctly, or at least in the same way as other major browsers. In this case, a background color on `<dds-card-group-item>` elements was not being overridden, and those element's `padding-top` made a chunk of the dark gray background visible.

I side-stepped this issue by modifying the way we create even spacing between grid items. Instead of only setting a `column-gap` value, we now use the full `gap` property for internal spacing and set a `padding-top` on the `<dds-card-group>` to create external spacing above the card group.

### Changelog

**Changed**

- Refactor `card-group` internal and external spacing to prevent Safari bug.

### DIFF
--- a/packages/styles/scss/components/card-group/_card-group.scss
+++ b/packages/styles/scss/components/card-group/_card-group.scss
@@ -1,5 +1,5 @@
 /**
- * Copyright IBM Corp. 2016, 2021
+ * Copyright IBM Corp. 2016, 2022
  *
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
@@ -175,6 +175,16 @@
   // Add Grid Mode Narrow (16px gutter) style.
   :host(#{$dds-prefix}-card-group)[grid-mode='narrow'],
   .#{$prefix}--card-group--narrow {
+    @include carbon--breakpoint('sm') {
+      padding-top: $carbon--spacing-03;
+      gap: $carbon--spacing-03;
+    }
+
+    @include carbon--breakpoint('md') {
+      padding-top: $carbon--spacing-05;
+      gap: $carbon--spacing-05;
+    }
+
     @include carbon--breakpoint('lg') {
       grid-template-columns: repeat(
         3,
@@ -182,20 +192,11 @@
       );
     }
 
-    @include carbon--breakpoint('sm') {
-      column-gap: $carbon--spacing-05;
-    }
-
     ::slotted(#{$dds-prefix}-card-group-item),
     ::slotted(#{$dds-prefix}-card-group-card-link-item),
     .#{$prefix}--card-group__cards__col {
       border: 0;
-      background-color: inherit;
-      padding: $carbon--spacing-03 0 0 0;
-
-      @include carbon--breakpoint('md') {
-        padding: $carbon--spacing-05 0 0 0;
-      }
+      padding: 0;
     }
   }
 


proeung (changes requested):
@jkaeser Thanks for opening a PR for this issue.

I'm noticing a couple of style regression bugs that are happening within a couple of the `Card Group` variants. Can you take a look?

**Card Group with Card Link (Type)**
Screen Shot 2022-02-16 at 3 45 10 PM

I'm seeing a regression in the styling where the border-left is being added and border-bottom is currently set to `2px` instead of `1px`. This might not be related to the changes that you're making here, but I just want to quickly bring this bug up as it doesn't align with the current design specs (see [link](https://ibm.ent.box.com/file/827717793884)) or the changes coming from the PR that I worked on (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/6934).

--------------

**Card Group with Card in Card**
https://user-images.githubusercontent.com/1815714/154355052-b35b35a5-5416-4910-9717-33b8e4a3e1c6.mov

I'm seeing an extra `1px` border-left being added to each card when the knob switches from `Narrow 16px` to `Clappsed 1px` and back to `Narrow 16px`. See video screenshot above.

@oliviaflory @RichKummer Let me know if you guys noticed anything else from a design QA perspective.

jkaeser:
Thanks for the review, @proeung! I fixed both of those regressions you pointed out. Notably, the first is already in production – https://www.ibm.com/standards/carbon/web-components/iframe.html?id=components-card-group--default&knob-Card%20type:_CardGroup=Card%20link&knob-Number%20of%20cards:_CardGroup=5&knob-Cards%20per%20row:_CardGroup=dds-ce-demo-devenv--cards-in-row-3&knob-Grid%20mode:_CardGroup=collapsed&knob-Offset:_CardGroup=0&knob-Number%20of%20cards_CardGroup=5.

oliviaflory:
Thanks for catching those regressions @proeung, and thank you @jkaeser for the fixes !


@jkaeser this is looking good on all browsers and breakpoints, I did notice one thing that's a little off:

The last card on the right side seems to have an extra white (or lighter in the dark themes) 1px border on the right side.

**PR**
Screen Shot 2022-02-16 at 6 44 41 PM


**Zoom in**
Screen Shot 2022-02-16 at 6 45 53 PM

**Production** 
Screen Shot 2022-02-16 at 6 46 17 PM

jkaeser:
@oliviaflory Thank you for your review as well! I've fixed the issue you were seeing, which I believe is the same one reported in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8299.

proeung:
@jkaeser Looks good! Thanks for addressing my feedback!

----------

@oliviaflory When the card group is set to `Outlined card 1px`, I'm noticing that we don't have a full border around the card, similar to what's being depicted in the design specs. Is this a recent design change? I'm seeing this as well within the `next` tag release (https://carbon-design-system.github.io/carbon-for-ibm-dotcom/next/web-components/?path=/story/components-card-group--default). Let me know if this is a regression bug. If so, we can maybe address it in this PR or open a follow-up issue to track the work.

component--card-group--specs--max copy
Screen Shot 2022-02-17 at 10 20 25 AM

oliviaflory:
@jkaeser thank you for the fix!

---

@proeung I noticed this issue with the outlined cards too 😞 but I am also seeing it in Production so I don't believe it's a regression bug. We must have missed it at some point!

I'm trying to be conscious of scope creep..if it's an easy fix to do in this PR that is amazing. But if it makes more sense to fix in a follow on issue that works for me because it is present in Production.

> @oliviaflory When the card group is set to `Outlined card 1px`, I'm noticing that we don't have a full border around the card, similar to what's being depicted in the design specs. Is this a recent design change? I'm seeing this as well within the `next` tag release (https://carbon-design-system.github.io/carbon-for-ibm-dotcom/next/web-components/?path=/story/components-card-group--default). Let me know if this is a regression bug. If so, we can maybe address it in this PR or open a follow-up issue to track the work.
> 
> component--card-group--specs--max copy Screen Shot 2022-02-17 at 10 20 25 AM

jkaeser:
> When the card group is set to Outlined card 1px, I'm noticing that we don't have a full border around the card, similar to what's being depicted in the design specs.

@proeung @oliviaflory I think it's best to create a separate issue for this regression. It may not end up being a lot of code changes, but it's going to require some careful thought and testing.

oliviaflory (approved):
@jkaeser thank you, visual changes approved!

proeung (approved):
@jkaeser Sounds good! I'll open a new issue for the border regression issue that we've discussed above. In the meantime, this PR looks good from a code and visual perspective.

proeung:
Looks like we need to review and approved the changes in the Percy visual diffs (see [link](https://percy.io/538fc19a/Carbon-for-IBM.com-Web-Components/builds/15830245/changed/893252912?browser=chrome&browser_ids=20&subcategories=changes_requested%2Cunreviewed&viewLayout=overlay&viewMode=new&width=320&widths=320%2C1280)), in order for the changes in this PR to be ready for merge. @oliviaflory or @IgnacioBecerra can you help take a look? I don't have permission to approve the visual diffs.

jkaeser:
> Looks like we need to review and approved the changes in the Percy visual diffs (see [link](https://percy.io/538fc19a/Carbon-for-IBM.com-Web-Components/builds/15830245/changed/893252912?browser=chrome&browser_ids=20&subcategories=changes_requested%2Cunreviewed&viewLayout=overlay&viewMode=new&width=320&widths=320%2C1280)), in order for the changes in this PR to be ready for merge. @oliviaflory or @IgnacioBecerra can you help take a look? I don't have permission to approve the visual diffs.

This caught some regressions that I'll need to fix before we merge this.

jkaeser:
@oliviaflory @proeung I had to revert the changes that fixed the extra border on the right side in order to get Percy tests to pass.

The extra border only appears if the Carbon theme changes after the card group has already rendered. It's a pre-existing issue that feels like an edge case, so perhaps it's best to create a separate ticket to resolve it. Thoughts?

proeung:
> @oliviaflory @proeung I had to revert the changes that fixed the extra border on the right side in order to get Percy tests to pass.

@jkaeser Yeah, I don't think we need to revert the commit for the fix that you've added for the extra border issue in order to mark the Percy visual diffs as approved. Regardless, I have created an issue (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8335) to track the border regression issue that we discovered above. Feel free to adjust, if I'm missing anything.

In the meantime, looks like everything checked out in this PR and we can go ahead and merge down the fixes that you have here.

Issue #8322, opened 2022-02-16
[Masthead]: Errors when data lacks a menuSections 


IgnacioBecerra:
### Description

When passing in translation data, the React Masthead fails if it lacks `menuSections` in the JSON object.

```
{
        "title": "Let's Create",
        "titleEnglish": "Let's Create",
        "url": "https://www.ibm.com/lets-create?lnk=hpmlc",
        "hasMenu": false,
        "hasMegapanel": false
},
```

One solution is to simply add an empty `menuSections` array to the object in the data files:

```
{
        "title": "Let's Create",
        "titleEnglish": "Let's Create",
        "url": "https://www.ibm.com/lets-create?lnk=hpmlc",
        "hasMenu": false,
        "hasMegapanel": false,
       "menuSections": []
},
```

However, this issue needs to be addressed within the code by adding null checks when needed for this object.

### Component(s) impacted

Masthead
Dotcom Shell

### Browser

Chrome

### Carbon for IBM.com version

canary

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

N/A

### Package

@carbon/ibmdotcom-react

### CodeSandbox example

N/A

### Steps to reproduce the issue (if applicable)

1. Remove `menuSections` from the translation files, it will fail immediately.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #8324, opened 2022-02-16
fix(masthead): added menusection null checks

IgnacioBecerra:
### Related Ticket(s)
#8322

### Description
When the`menuSection` object is not found, the React masthead will fail to render and will cause an error that might crash the website.

### Changelog

**New**

- more null checks for the `menuSection` object


packages/react/src/components/Masthead/Masthead.js
```diff
@@ -256,7 +256,7 @@ const Masthead = ({
       if (!matchFound) {
         if (sections.url === currentUrlPath) {
           matchFound = true;
-        } else if (sections?.menuSections[0]) {
+        } else if (sections?.menuSections && sections?.menuSections[0]) {
```
emyarod, line 259: I believe we can use optional chaining here too right?

```suggestion
        } else if (sections?.menuSections?.[0]) {
```

packages/react/src/components/Masthead/MastheadLeftNav.js
```diff
@@ -124,14 +124,18 @@ const MastheadLeftNav = ({
         rest.hasL1Data ? 'l1' : 'l0'
       }-nav${i}`;
 
-      const menuItems = link.menuSections[0]?.menuItems;
+      const menuItems = link.menuSections && link.menuSections[0]?.menuItems;
```
emyarod, line 127:
```suggestion
      const menuItems = link.menuSections?.[0]?.menuItems;
```

packages/react/src/components/Masthead/MastheadLeftNav.js
```diff
@@ -269,10 +273,11 @@ function _renderLevel1Submenus(
     let highlightedItems = [];
     const items = [];
 
-    menu.sections[0].menuItems.forEach(item => {
-      if (item.highlighted) return highlightedItems.push(item);
-      return items.push(item);
-    });
+    menu.sections &&
+      menu.sections[0].menuItems.forEach(item => {
```
emyarod, line 277:
```suggestion
    menu.sections?.[0].menuItems.forEach(item => {
```

packages/react/src/components/Masthead/MastheadLeftNav.js
```diff
@@ -286,7 +291,7 @@ function _renderLevel1Submenus(
             menuState.level0 === menu.parentKey && menuState.level1 >= 0,
         })}
         id={`panel__(${menu.parentKey},-1)`}
-        heading={menu.sections[0]?.heading}
+        heading={menu.sections && menu.sections[0]?.heading}
```
emyarod, line 294:
```suggestion
        heading={menu.sections?.[0]?.heading}
```

emyarod (approved):
looks good to me! if I'm not mistaken I think we can use optional chaining for some more of these checks

IgnacioBecerra:
@emyarod Weird, that's actually the first thing I tried and it had given me a different error. Could be that I was still missing other null checks at the time, but regardless the last commit now uses optional chaining. Thanks!

Issue #8247, opened 2022-02-10
[Carousel]: mixture of dds-card and dds-card-cta causing heading alignment issue - DPO Consulting


proeung:
**Engineering info:**  
- Innovation Team engineer: @proeung 
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2406
- AEM JIRA Ticket: https://jsw.ibm.com/browse/ADCMS-1338
- DPO consulting engineer: @IgnacioBecerra 

-----

### Description

- The AEM team is noticing an alignment issue when there's a mixture of `<dds-card>` and `<dds-card-cta>` within the `<dds-carousel>` component.
- If one of the first two cards has a heading with three lines, then the copy text of all cards align again, however, if a video card has the longest heading, then they do not align.
- This might be an issue related to the sameHeight function is not calculating the video heading compared to the regular headings.

<img width="1624" alt="image-2022-02-09-17-15-11-703" src="https://user-images.githubusercontent.com/1815714/153505293-b13e3282-ed6a-471b-9527-dc2348df12ce.png">

### Component(s) impacted

dds-carousel

### Browser

Chrome, Safari, Firefox, Microsoft Edge

### Carbon for IBM.com version

v1.30, Canary

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

AEM

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://codesandbox.io/s/gifted-blackburn-uk21f?file=/index.html

### Steps to reproduce the issue (if applicable)

See CodeSandbox link to view the issue.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

jeffchew:
Should create a test scenario issue for this.

proeung:
@jeffchew Yep! I'll create a new issue for adding a new SB variant that includes a mixture of `dds-card` and `dds-card-cta` so that we can use it in the e2e testing of this component. We'll account for this work in a separate ticket.

proeung:
[...]

PR #8336, opened 2022-02-18
fix(carousel): apply same height function to heading used in dds-card-cta

proeung:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8247

### Description

- This PR addresses the alignment issue seen when there's a mixture of `<dds-card>` and `<dds-card-cta>` within the `<dds-carousel>` component.

### Changelog

**New**

- Add a new SB variant called Carousel > Cards with Media - https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/8336/index.html?path=/story/components-carousel--cards-with-media

**Removed**

- Removed `<dds-card-heading>` from `CardsWithVideos` as the card heading should be coming from the Mediacenter/Kaltura because this is the content usage when displaying a card with a video button.


IgnacioBecerra:
@proeung Looking good upon resize, but at first render it looks a bit off:

Screen Shot 2022-02-21 at 12 39 43 PM
 
This probably happens due to the max height being applied from the other regular cards before the tallest title (from the video API) comes through. Maybe we could add `setTimeout` or something that will trigger once the video titles come through that will retrigger the `sameHeight` function.

packages/web-components/src/components/cta/card-cta.ts
```diff
@@ -47,6 +47,13 @@ class DDSCardCTA extends VideoCTAMixin(CTAMixin(DDSCard)) {
       return super._renderHeading();
     }
     const caption = formatVideoCaptionInEffect({ name: videoName });
+
+    this.dispatchEvent(
+      new CustomEvent('video-updated', {
```
IgnacioBecerra, line 52: Works great! Let's just update the custom event name to something like `dds-card-cta-video-title-updated` and use getters instead of a string, both here and in the `carousel.ts` file like we did for the other issue.

proeung: Great suggestion! Please see the changes here (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/8336/commits/a057f6519ec806c246f94e2dc6176f084c5afb0b).

IgnacioBecerra (approved):

kennylam (approved):
LGTM thanks @proeung!

Issue #8254, opened 2022-02-11
[Internal Vendor Packages]: Update services and utilities to be referenced in internally built vendor packages


jeffchew:
### User story

> As a `[user role below]`:

Carbon for IBM.com developer

> I need to:

ensure that changes in upstream packages within the repo are used instead of published packages

> so that I can:

not run into issues applying changes in upstream packages to downstream packages in the monorepo

### Additional information

- We have run into issues in the past regarding referencing newer changes in `utilities` and `services` in packages like `web-components` and `react`.
- We are already using this method for `services-store` in `web components`

### Acceptance criteria

- [ ] `services` package uses internal vendor build of `utilities`
- [ ] `web-components`,`react`, `services-store` uses internal vendor build of `utilities` and `services`

PR #8363, opened 2022-02-23
chore(vendor): using vendor directories for services and utilities

IgnacioBecerra:
### Related Ticket(s)
#8254

### Description
This PR ensures that all the calls to `utility` and `services` functions in our components are using the internal vendor build of these packages.

Everything works properly, however, in React, the following components give an error relating to the import of `marked` like we have been seeing recently:
* Card
* Contentblock
* Content group
* Content item
* Content item horizontal
* Content section

Note that they work properly with the new vendor path for web components.

### Changelog

**New**

- added gulp tasks to create the vendor build for the `services` and `utilities` packages

**Changed**

- now using vendor paths for the `utilities` and `services` packages in `react`, `web-components`, and `services` packages


kennylam (approved):
Nice one, thanks @IgnacioBecerra!

Issue #8332, opened 2022-02-17
[Link with icon]: Visited link in mobile Safari


RichKummer:
### Description

On mobile Safari, the visited link toggle in Storybook only changes the color of the text - the icon remains blue when it should also turn the same color as the visited text in the CTA.

![IMG_2039](https://user-images.githubusercontent.com/15643582/154545829-93f6db37-f0c8-4302-843c-2f7dfc08ce30.PNG)

### Component(s) impacted

Link with icon

### Browser

Safari

### Carbon for IBM.com version

v1.31.0-rc.1

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

C4IBM.com

### Package

@carbon/ibmdotcom-react

### CodeSandbox example

https://carbon-design-system.github.io/carbon-for-ibm-dotcom/next/react/?path=/story/components-link-with-icon--default

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

kennylam:
This has already been resolved in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/8367.

PR #8367, opened 2022-02-24
fix(styles): add visited styles to link with icon

kennylam:
### Related Ticket(s)

#8332 

### Description

This adds `visited` styles to the `link-with-icon` component.

### Changelog

**New**

- `visited` styles added to `link-with-icon`


RichKummer (approved):
Tested it out on Safari and mobile and looks good. Thanks @kennylam !

Issue #8368, opened 2022-02-24
[Link with icon]: incorrect font size when hovering over `disabled` link


kennylam:
### Description

Both web components and React versions of `link with icon` have the incorrect hover styles when the link is disabled.

![Feb-24-2022 09-15-05](https://user-images.githubusercontent.com/909118/155544927-3d82154a-ebda-4402-868e-42569d3ecf13.gif)

### Component(s) impacted

`link with icon`

### Carbon for IBM.com version

latest

### Severity

Severity 4 = The problem is not visible to or noticeable to an average user. Affects minor functionality, no workaround needed.

### Application/website

Carbon for IBM.com

### Package

@carbon/ibmdotcom-styles

### CodeSandbox example

can be seen in latest [Storybook](https://carbon-design-system.github.io/carbon-for-ibm-dotcom/canary/web-components/iframe.html?id=components-link-with-icon--default&knob-Link%20text%20(unnamed%20slot)_LinkWithIcon=Link%20text&knob-Disabled%20(disabled)_LinkWithIcon=true&knob-Link%20href%20(href)_LinkWithIcon=https://github.com/carbon-design-system/carbon-web-components&knob-Icon%20Position%20(icon-placement):_LinkWithIcon=right)

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #8369, opened 2022-02-24
fix(link-with-icon): fix disabled styles

kennylam:
### Related Ticket(s)

#8368 

### Description

Fixes `link with icon` component hover styles when it is disabled.

### Changelog

**New**

- add correct font size when disabled and hovering

**Removed**

- unused css


RichKummer (approved):
Seeing the correct hover for the disabled state!

Issue #8335, opened 2022-02-18
[Card Group]: Cards no longer have 1px border & Extra border on Gray90 & Gray100 (Regression)


proeung:
**Engineering info:**  
- Innovation Team engineer: @jkaeser 
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2415
- DPO consulting engineer: 

-----

### Description

- When the card group is set to Outlined card 1px, I'm noticing that we don't have a full border around the card, similar to what's being depicted in the design specs (see attached). 
- After talking to @oliviaflory, this looks to be a regression issue that's present in Production.
- See discussion from https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/8320#issuecomment-1043070740

![](https://user-images.githubusercontent.com/1815714/154513734-1dfcf227-a7d1-4df6-bb60-fe2b7f098ec9.png)
![](https://user-images.githubusercontent.com/1815714/154513740-83ef9fca-4f09-41ee-8963-a88aed3537a3.png)

- In React wrapper: On the darker themes, a light border appears on the edge of the 3rd card in the row. This also shows up for `Card group - with card in card` in Gray90 & Gray100. See - https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8299

<img width="1571" alt="Screen Shot 2022-02-16 at 8 38 00 AM" src="https://user-images.githubusercontent.com/15643582/154276439-cd18881f-ecd4-47c5-b0fa-6ec7b8c0dbc3.png">

### Component(s) impacted

Card Group

### Browser

Chrome, Safari, Firefox, Microsoft Edge

### Carbon for IBM.com version

v1.31.0

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede usability or functionality. Affects minor functionality, has a workaround.

### Application/website

AEM

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://carbon-design-system.github.io/carbon-for-ibm-dotcom/next/web-components/?path=/story/components-card-group--default

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

Issue #8301, opened 2022-02-16
[Card section offset]: Card border that disappears on screen resize


RichKummer:
**Engineering info:**  
- Innovation Team engineer: @jkaeser 
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2416
- DPO consulting engineer: 

-----

### Description

When the `Card section offset` component is loaded on Storybook - React wrapper, there is a border that appears on the edge of the right-most cards. The border disappears when I resize the browser window, but reappears if I change the theme.

https://user-images.githubusercontent.com/15643582/154314857-6f5c31f9-61bc-4a32-97ad-12b1c03065c0.mov

### Component(s) impacted

Card section offset

### Browser

Firefox

### Carbon for IBM.com version

v1.31.0-rc.1

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

C4IBM.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://codesandbox.io/s/github/carbon-design-system/carbon-for-ibm-dotcom/tree/main/packages/web-components/examples/codesandbox/components-react/card-section-offset

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #8373, opened 2022-02-24
fix(card-group): fix grid mode knob and prevent incorrect border colors

jkaeser:
### Related Ticket(s)

Resolves https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8335 and https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8301

### Description

This PR solves two problems: 
1. The default `DDSCardGroup` story's "grid mode" knob was not actually updating the grid mode, which resulted in what looked like a visual regression with the "outlined" mode.
2. There were circumstances in Storybook where changing the Carbon theme caused the wrong color border to be used on some grid items.

Additionally, it tidies up some related `DDSCardGroup` code.

### Changelog

**Changed**

- Fixes attribute name to make "grid mode" knob work.
- Prevents incorrect border colors on `DDSCardGroup` grid items.
- Update methods to clean up empty card items when appropriate.

### DIFF
--- a/packages/styles/scss/components/tabs-extended/_tabs-extended.scss
+++ b/packages/styles/scss/components/tabs-extended/_tabs-extended.scss
@@ -64,10 +64,29 @@
         height: auto;
         overflow: visible;
         white-space: normal;
+        background: none;
+        border: initial;
+        border-bottom: carbon--rem(2px) solid $ui-03;
+        display: inline-flex;
+        width: $spacing-13;
+        text-align: left;
 
         &[hasTooltip] {
           @include tooltip--trigger('definition', top);
           @include tooltip--placement('definition', 'bottom', 'start');
+
+          &::before,
+          &::after {
+            display: none;
+          }
+
+          &:hover,
+          &:focus {
+            &::before,
+            &::after {
+              display: flex;
+            }
+          }
 
           &:focus {
             @include focus-outline('outline');
@@ -129,9 +148,6 @@
 
         .#{$prefix}--tabs__nav-link {
           width: auto;
-        }
-
-        a.#{$prefix}--tabs__nav-link {
           padding: 0;
           border-bottom: carbon--rem(2px) solid transparent;
         }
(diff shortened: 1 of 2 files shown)


packages/web-components/src/components/card-group/card-group.ts
```diff
@@ -149,8 +150,14 @@ class DDSCardGroup extends StableSelectorMixin(LitElement) {
       if (!this.pictograms) {
         this._setSameHeight();
       }
-      if (this.gridMode !== GRID_MODE.NARROW) {
+
+      if (gridMode !== GRID_MODE.NARROW) {
         this._fillLastRowWithEmptyCards(columns);
+      } else {
+        this._removeEmptyCards();
+      }
```
jkaeser, line 158: Previously, you could get leftover empty cards when switching from a non-narrow grid mode to the narrow one. This didn't cause any issues (yet!), but there was an easy opportunity to reuse some existing code to clean up after ourselves.

jkaeser:
Putting this back in draft mode since I see now the reason we can't use CSS grid `gap` 😢 It's imperative that `<dds-card-group>` elements have a transparent background to work within elements such as `<dds-card-section-offset>`.

oliviaflory:
@jkaeser this looks really great! I ran through all the knobs and the stories look good.

One strange thing happened a few times where a background color would appear in the first card position (top right) when using the `Offset = 1` and `Grid mode = Outlined cards` 
Screen Shot 2022-03-01 at 5 20 54 PM

I took a few videos attempting to replicated, but I'm not sure if there is a magic sequence, might just be a weird storybook thing. Did you happen to notice this at all?

**Chrome**
https://user-images.githubusercontent.com/45692622/156264900-1c19d43b-c43c-4d01-9b5e-966520be7819.mov


**Firefox**
https://user-images.githubusercontent.com/45692622/156264919-7cda8576-a325-4ab7-891c-d496e091ab02.mov

jkaeser:
> One strange thing happened a few times where a background color would appear in the first card position (top right) when using the Offset = 1 and Grid mode = Outlined cards

@oliviaflory Good catch! I can consistently replicate when I set the offset to 1 and then change to the "Outlined cards (1px border)" grid mode. Looking into a fix now.

jkaeser:
> @oliviaflory Good catch! I can consistently replicate when I set the offset to 1 and then change to the "Outlined cards (1px border)" grid mode. Looking into a fix now.

@oliviaflory This issue should be fixed now. Can you take another look when you have a chance? Thanks!

oliviaflory (approved):
LGTM, thank you @jkaeser 🎉

proeung (approved):
LGTM!

PR #8378, opened 2022-02-24
chore(gh-actions): add timeout to e2e tests

kennylam:
### Description

This adds a timeout to the `e2e-tests` GitHub jobs for both React and web components. These jobs can hang for hours and clog the pipeline. At a quick glance, these jobs can finish as quickly as 14 minutes to as long as 35 minutes (or longer, depending on the amount of jobs in the queue). Initial timeout value is set to 60, which can be adjusted as needed.

### Changelog

**New**

- add `timeout-minutes` to end-to-end Github jobs


jeffchew (approved):
LGTM!

Issue #8348, opened 2022-02-21
[Content Block Horizontal border attribute]: The content block horizontal web component is always setting border attribute as true by default.


druckerk:
**Engineering info:**  
- Innovation Team engineer: @proeung 
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2419
- DPO consulting engineer:

-----

### Description

**Expected Results:**
When a user uses the <dds-content-block-horizontal> component, the default for this component should be showing in the inspect html as <dds-content-block-horizontal data-autoid="dds--content-block-horizontal"></dds-content-block-horizontal>. Then when a user adds the attribute "border" to the <dds-content-block-horizontal border> component, then the component should be appearing as <dds-content-block-horizontal border data-autoid="dds--content-block-horizontal"></dds-content-block-horizontal> in the inspect html. 

**Actual Results:** 
When a user uses the <dds-content-block-horizontal> component, the current default is showing in the inspect html as <dds-content-block-horizontal data-autoid="dds--content-block-horizontal" border></dds-content-block-horizontal>. Then when the user adds the attribute "border" to the <dds-content-block-horizontal border> component, the component will appear as <dds-content-block-horizontal border data-autoid="dds--content-block-horizontal"></dds-content-block-horizontal> in the inspect html. 

### Component(s) impacted

Content Block Horizontal

Storybook:
https://www.ibm.com/standards/carbon/web-components/?path=/docs/components-content-block-horizontal--default

### Browser

Chrome

### Carbon for IBM.com version

ibmdotcom-web-components: 1.15.0 / ibmdotcom-styles: 1.30.0

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

AEM

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://codesandbox.io/s/red-cookies-5v9qnw?file=/dist/index.html

### Steps to reproduce the issue (if applicable)

In the code sandbox:

1. Test the <dds-content-block-horizontal> element without the "border" attribute.
2. Once the build completes, then inspect the HTML from the browser and you will see the "border" attribute appear to the right of the data-autoid attribute in the element. 
3.  Then go back to the code add the "border" attribute to the element. 
4. Once the rebuild is completed, then inspect the HTML from the browser and you will see the "border" attribute has moved to the left of data-autoid attribute in the element. 

[...]

PR #8382, opened 2022-02-24
fix(content-block-horizontal): ensure border attribute returns correctly

proeung:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8348

### Description

- This PR ensures that when the `border` attribute/prop is defined in the `<dds-content-block-horizontal>` component that the border-bottom renders and is hidden when it's not defined. 
- See the issue for more detail of the current behavior seen by the AEM adopters - https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8348

### Changelog

**Changed**

- Ensure that the `border` attribute is defined
- Set border default prop to `false`.

### Testing

- Visit the `Content Block Horizontal` layout component code sandbox
- Add `border` to the `<dds-content-block-horizontal>` element (eg. `<dds-content-block-horizontal border>` and ensure that the border-bottom renders correctly.
- If the `border` attribute is not being used, ensure that the border-bottom does not get rendered.


ariellalgilmore (approved):
LGTM!

kennylam (approved):
Thanks @proeung!

Issue #8383, opened 2022-02-24
[CI]: CDN builds failing


kennylam:
### Description

Current CDN builds are failing for the `build:components` job.

### Component(s) impacted

`@ibmdotcom/utilities`

![image](https://user-images.githubusercontent.com/909118/155619138-9fdc6ded-c397-4b33-be16-7e1b8d6f04ce.png)

### Carbon for IBM.com version

latest

### Severity

Severity 1 = The design is broken in a critical way that blocks users from completing tasks or damages the brand. Affects major functionality, no workaround.

### Application/website

Carbon for IBM.com

### Package

@carbon/ibmdotcom-utilities

### CodeSandbox example

N/A

### Steps to reproduce the issue (if applicable)

```
yarn
yarn build
cd packages/web-components
yarn build:components
```

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

jeffchew:
@kennylam moved this into `to do this sprint` and added to `sprint 4`.

PR #8384, opened 2022-02-25
fix(rollup): fix to named exports declaration for marked dependency

jeffchew:
### Related Ticket(s)

Refs https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8383

### Description

The namedExport declaration specific to `marked` needed to be very specific in order for the commonjs plugin in rollup to register the import correctly. This will hopefully be the fix.

### Changelog

**Changed**

- Updated `namedExports` declaration in rollup config for web components


kennylam (approved):
Great catch thanks @jeffchew!

Issue #7815, opened 2021-12-06
[Tabs & Tabs extended with media]: Not on grid


shixiedesign:
**Engineering info:**  
- Innovation Team engineer: @mwanberg 
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2387
- DPO consulting engineer: N/A

-----

### Description

1. Tabs extended media: Contents of the component are not aligned to Carbon grid. 

**Current Canary web components**
![image](https://user-images.githubusercontent.com/15144993/144798385-d6124585-b78e-41e9-bffd-59e5fb3c9ca7.png)

**Expecting**
![image](https://user-images.githubusercontent.com/15144993/144798590-8d961cc2-1831-4fbf-a085-4e13a63182d4.png)

2. For Tabs extended, the individual tabs should be 2 columns wide. 

**Current:** 
![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/52368160-9d5e-49cf-bd6d-0c490eb83840)

**Expecting:**
![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/6d77531e-9c89-455f-a572-634aaf66ed0b)

### Component(s) impacted

Tabs extended
Tabs extended with media

### Browser

Chrome

### Carbon for IBM.com version

canary

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

NA

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

NA

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

proeung:
Design specs - https://ibm.ent.box.com/folder/133803243554?s=3sun8lo3gu2b4uaa4oxgpxrqit1m7t6i

PR #8365, opened 2022-02-23
fix(tabs-extended): change anchor-based tabs to button-based tabs

andy-blum:
### Related Ticket(s)

Fixes #8357 

### Description

Fixes an issue where iOS on tablet couldn't use tabs.

### Changelog

**Changed**

- changes `<a>` elements to `<button>` elements in tabs so iOS devices getting the desktop-style layout can navigate with voiceover

PR #8394, opened 2022-02-25
fix(Tabs & Tabs extended with media): adjust the layout to align with the defined grid

mwanberg:
### Related Ticket(s)

[[Tabs & Tabs extended with media]: Not on grid](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7815)

### Description

1. Tabs extended media: Contents of the component are not aligned to Carbon grid. 

**Current Canary web components:**
image

**Expecting:**
image

**Updated:**
Screen Shot 2022-03-02 at 9 12 59 PM

2. For Tabs extended, the individual tabs should be 2 columns wide. 

**Current:** 

**Expecting:**

**Updated:**
Screen Shot 2022-03-02 at 9 12 29 PM

### Notes

As of this first pass, there are still some differences between the specs and the results, notably:

1. Tabs Extended with Media
    - The tab contents' image and text are not aligned to the grid
    - Some of the vertical margins of the tab contents are different
2. Tabs Extended
    - The tabs' text is not bottom aligned when they have multiple lines
    - The tab contents' text and cta block have some left misalignment due to inherited styles from the component it's using, Content Block Media

I am unsure if the scope of this ticket includes addressing the above.


proeung:
@mwanberg Thanks for opening this draft PR. I just looked at the WC deploy preview and I noticed that there's some `16px` spacing between each tab, which I don't believe is part of the design specs are in the ticket that Shixie has outlined here (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/7815) (see attached). Can you take a look?

Screen Shot 2022-02-28 at 5 24 21 PM

Also, do you mind outlining the style targeting issue that you're having? You should be able to target elements within the shadow root by referencing the slotted items.

mwanberg:
@proeung Thanks for pointing out the tab width issue - that should be resolved now.

Here is what I am encountering with the tab / accordion contents styling:

1) For Tabs Extended, the content of the tab (accordion in mobile) is a separate component, `dds-content-item`, which has its own styling.
2) For Tabs Extended with Media, the content of the tab is `dds-content-item-horizontal-media`, which also has its own styling.

Am I mistaken in that it will be very challenging to affect the styling of those elements without extending them in the Tabs Extended / with Media?

Example of `dds-content-item` margins on Tabs Extended:

Screen Shot 2022-03-01 at 3 03 00 PM

Example of `content-item-horizontal` grid styles on Tabs Extended with Media:

Screen Shot 2022-03-01 at 3 05 50 PM

proeung:
@mwanberg I see what you're trying to do here now. If you want to remove the margin around the `dds-content-item` component within the `dds-tab`, you can target it through the slotted return. Something like `:host(dds-content-block-media-content) ::slotted([slot='dds-content-item'])`.

Screen Shot 2022-03-02 at 12 48 58 PM

In this case, where you want the text in the tab content to align with the text of the tab itself, it might be better to remove the extra padding on `:host(dds-tabs-extended) .bx--tab-content` (see attached).

Screen Shot 2022-03-02 at 12 56 23 PM

mwanberg:
> @mwanberg I see what you're trying to do here now. If you want to remove the margin around the `dds-content-item` component within the `dds-tab`, you can target it through the slotted return. Something like `:host(dds-content-block-media-content) ::slotted([slot='dds-content-item'])`.

Do you mean I should target it in the Tabs Extended styles or in the Content Block Media styles? I attempted to put that selector in the Tabs Extended styles and got no results. Perhaps I am misunderstanding your suggestion.

> In this case, where you want the text in the tab content to align with the text of the tab itself, it might be better to remove the extra padding on `:host(dds-tabs-extended) .bx--tab-content` (see attached).

I actually do not know how it should be aligned. In the [Tabs Extended specs](https://ibm.ent.box.com/file/916036051163), nothing is defined for the desktop width contents. However, the mobile spec does have contents (see page 14), and the text there doesn't have the added margins (unlike in the original Content Block Media component), so I am assuming that it would be the same for the desktop version (to have no margins around the text).

For both the Tabs Extended and Tabs Extended with Media, the specs have them using what are essentially Content Block Media and Content Item Horizontal (with Media) respectively, but with somewhat different grid alignments and margins than the originals.

Do you think it is within scope of this ticket to attempt to align the tab contents (of both or either Tabs Extended or Tabs Extended with Media) to the original specs?

Are these tabs components supposed to be agnostic in terms of what goes in the tab contents? Or are those tab contents supposed to for sure look a certain way? If the former, then I would guess that updates are not needed. If the latter, then it becomes a scope and effort question.

I will update the description of this PR to reflect the original requests, what I've updated, and what might remain.

andy-blum:
Heads up: there might be conflicts with this PR: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/8365

oliviaflory:
Hi @mwanberg adding my perspective, and I hope this helps to clarify a few things!

> I actually do not know how it should be aligned. In the [Tabs Extended specs](https://ibm.ent.box.com/file/916036051163), nothing is defined for the desktop width contents. However, the mobile spec does have contents (see page 14), and the text there doesn't have the added margins (unlike in the original Content Block Media component), so I am assuming that it would be the same for the desktop version (to have no margins around the text).

I'm very sorry that this was confusing! `Tabs extended` was originally designed as `Tabs extended with media` and we decided to build the tabs out separately so `Tabs extended` could take on many forms of content. 

You are correct in assuming the tabs extended content should not have any margins around the text in the desktop version. Image example below, and I will make a new issue for us to update the design specs to make this more clear for future reference!
Screen Shot 2022-03-04 at 3 57 13 PM



> For both the Tabs Extended and Tabs Extended with Media, the specs have them using what are essentially Content Block Media and Content Item Horizontal (with Media) respectively, but with somewhat different grid alignments and margins than the originals.

Unfortunately it looks like we updated the [Content item horizontal with media spec](https://ibm.ent.box.com/folder/112262788897?s=yiv7f37ktudwx1do7d9mp4bak17wl4dh) when the image is left aligned and did not go into Tabs extended with media to update the visual (so you are seeing an old spec and that's why it doesn't look quite right!) 😞   I will make an issue to address this as well.

The intent is that the Tabs extended are agnostic. IMO Tabs extended with media is an example of what type of content you could place into the tabs (Content block with media), the content **_should_** adhere to how it was designed on the grid. Below is the latest content block with media (left) in the Tabs content area. I assume it would just "plop" in and sit nicely 😅 
Screen Shot 2022-03-04 at 4 05 06 PM

I personally agree with @proeung that ensuring both tabs, tabs extended, and the content within the tabs and tabs extended are properly on the grid. However that might make the issue more than 1 story point(?)

mwanberg:
> I personally agree with @proeung that ensuring both tabs, tabs extended, and the content within the tabs and tabs extended are properly on the grid is within the scope of the original issue. However that might make the issue more than 1 story point(?)

Thank you so much for replying and providing some insight, @oliviaflory! That is very helpful.

Two followup questions:

1. For updating the contents of the tabs, should we do that in their original components (e.g. `Content Item Horizontal (with Media)`) or extend those so we can update the styling within Tabs Extended / Tabs Extended with Media?
2. If we need to update the story points, who should do that and to how much?

oliviaflory:
> 1. For updating the contents of the tabs, should we do that in their original components (e.g. Content Item Horizontal (with Media)) or extend those so we can update the styling within Tabs Extended / Tabs Extended with Media?

I believe we'd want to use the Content item horizontal (not extend or change the component to accommodate Tabs extended). But I am double checking with the DDS engineers.

> 2. If we need to update the story points, who should do that and to how much?

Storypoints are usually estimated and set by the developers, so I'd leave it up to you to let us know how much time this took and we can update. It helps us and you document how much time and effort is taken within a sprint

mwanberg:
> I believe we'd want to use the Content item horizontal (not extend or change the component to accommodate Tabs extended). But I am double checking with the DDS engineers.

@oliviaflory great! Please let me know what the consensus is so I know how to finish this PR. I'd like to get it in for this sprint, though I know it's not a super high priority one.

proeung:
@mwanberg This PR (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/8365) has been merged, please move forward with addressing the merge conflict in this PR.

As for your first question in this comment (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/8394#issuecomment-1060758685), please reach out to one of the DPO engineers (@IgnacioBecerra , @ariellalgilmore , or @emyarod ), so that we can move this work forward. Thanks!

mwanberg:
This PR is now ready for review. As discussed in [this Slack thread](https://ibm-marketing-comms.slack.com/archives/G01LKRP5RB6/p1646848391825839), we will not be tackling the layout of the tab contents in this PR. A followup ticket for the contents is here: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8470

proeung (approved):
LGTM from a code presentive!

I'll defer to @oliviaflory on the design QA side to ensure that the changes are expected. https://percy.io/538fc19a/Carbon-for-IBM.com-Web-Components/builds/16409602/changed/923734665?browser=chrome&browser_ids=20&subcategories=unreviewed%2Cchanges_requested&viewLayout=overlay&viewMode=original&width=1280&widths=320%2C1280

oliviaflory:
@mwanberg the shift in the tabs story content does feel like a regression (outlined in pink). Can you remind me if something was changed in the content area or if the 16px shift to the right is due to something in the content item?
Screen Shot 2022-03-10 at 5 20 15 PM


In the current tabs story the content is aligned correctly, with the type aligning to the column.
Screen Shot 2022-03-10 at 5 22 30 PM

---

For tabs extended with media the Percy difs are as expected 👍

mwanberg:
@oliviaflory Hm, if I remove the side padding from Tabs Extended tab content, then that will make the Tabs Extended with Media tab content become misaligned. However, I think that might be better, because it finishes making the tab content area agnostic in terms of what components get plopped in.

So I will make that change and remove the added padding, but you will now see a visual diff change in Tabs Extended with Media, in that content area.

The unnecessary padding, which I am now removing, is outlined in pink. It will now be up to the components that get put in to add their own padding as necessary. Does that sound ok?

Screen Shot 2022-03-11 at 10 24 58 AM

Screen Shot 2022-03-11 at 10 24 02 AM

oliviaflory:
@mwanberg Thanks for explaining! The previews are still building, but I will check back on Percy / Deploy previews. I think this makes sense..just about weighing all the repercussions 👍

oliviaflory (approved):
Thanks for the update @mwanberg

Issue #8032, opened 2022-01-14
[Masthead]: In mobile, Masthead + L1 disappears when scrolling up if already at the top


IgnacioBecerra:
### Description

In React, when using the Masthead + L1 in mobile view, if the user is at the very top of a page and they scroll up, instead of remaining stickied, the Masthead will completely disappear even though it shouldn't. 

https://user-images.githubusercontent.com/45692622/149419486-1c005906-c851-44bc-83f9-415d18448ef3.mp4

https://user-images.githubusercontent.com/24970122/149464371-ccbf75f2-473c-49f7-a86c-793360273f9b.MOV

### Component(s) impacted

Masthead with L1

### Browser

Safari

### Carbon for IBM.com version

canary

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

N/A

### Package

@carbon/ibmdotcom-react

### CodeSandbox example

https://www.ibm.com/standards/carbon/react/iframe.html?id=components-dotcom-shell--with-l-1

### Steps to reproduce the issue (if applicable)

1. Go to the very top of a page containing Masthead + L1
2. Scroll up
3. Masthead should erroneously disappear

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

emyarod:
on Android I'm unable to replicate this on Firefox and Chrome, this could be an iOS/mobile Safari specific bug

PR #8398, opened 2022-02-25
fix(masthead): aligned L1 sticky with wc

IgnacioBecerra:
### Related Ticket(s)
#8032

### Description
The previous Masthead L1 sticky code was only making so that the masthead disappeared immediately the moment the user started scrolling down without the table of contents pushing it up due to the way it was being activated. It also introduced a weird bug in iOS where the Masthead could simply disappear scrolling up. 

With this PR, the method used for sticky behavior aligns with the way it's being done in WC, so it's smoother all around.

### Changelog

**Changed**

- using new method to define sticky behavior with L1

**Removed**

- removed sticky booleans and states inside masthead


ariellalgilmore:
Seeing the entire masthead scroll away in Desktop now..
https://user-images.githubusercontent.com/20210594/156211122-fa3a3186-e966-417d-a95c-d9adc37318a0.mov

ariellalgilmore (approved):
LGTM!

kennylam (approved):
LGTM thanks @IgnacioBecerra!

PR #8437, opened 2022-03-02
chore: downgrade cypress

emyarod:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/8233/
https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/8147/

### Description

This PR downgrades Cypress in the `web-components` package back to v8.1.0 due to out of memory issues present in v8.2+

Firefox is also added to the browsers matrix again for end to end tests


kennylam (approved):
Thanks @emyarod!

PR #8455, opened 2022-03-04
chore: reenable firefox for github actions

emyarod:
### Related Ticket(s)

Do not merge – testing underway.

This PR reenables Firefox for GitHub Actions checks


jeffchew:
@emyarod wondering if we should close this for now? Don't think there's an urgency to jump back on this, perhaps when we turn on parallelization for cypress tests we can introduce?

emyarod:
@jeffchew yep sounds good to me!

Issue #8305, opened 2022-02-16
[CTA section - with content items]: Pictograms have extra spacing - Web Components and React Wrapper


RichKummer:
**Engineering info:**  
- Innovation Team engineer: @mwanberg 
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2412
- DPO consulting engineer:

-----

### Description

In Firefox with React wrapper, the `CTA section` that uses pictogram content items adds spacing below the pictogram that should not be there. 

The spacing between the content item text and pictogram should be 32px.

**Firefox**
<img width="1474" alt="Screen Shot 2022-02-16 at 12 41 45 PM" src="https://user-images.githubusercontent.com/15643582/154324418-990595db-e416-4cc6-9c6e-abc4d55bdf2a.png">

<img width="1476" alt="Screen Shot 2022-02-16 at 12 41 24 PM" src="https://user-images.githubusercontent.com/15643582/154324467-508b0c3b-0151-425d-8f1d-5aa3fa51fedc.png">

**Design specs**

<img width="724" alt="Screen Shot 2022-02-16 at 12 40 18 PM" src="https://user-images.githubusercontent.com/15643582/154324523-1b3773b2-6fce-426f-b7f4-dc32a13e2c8f.png">

### Component(s) impacted

CTA section - with content item (pictogram)

### Browser

Firefox

### Carbon for IBM.com version

v1.31.0-rc.1

### Severity

Severity 4 = The problem is not visible to or noticeable to an average user. Affects minor functionality, no workaround needed.

### Application/website

C4IBM.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://codesandbox.io/s/github/carbon-design-system/carbon-for-ibm-dotcom/tree/main/packages/web-components/examples/codesandbox/components-react/cta-section

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #8457, opened 2022-03-04
fix(cta-section): [CTA section - with content items]: Pictograms have extra spacing

mwanberg:
### Related Ticket(s)
[
[CTA section - with content items]: Pictograms have extra spacing](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8305)

### Description

"The CTA section that uses pictogram content items adds spacing below the pictogram that should not be there. The spacing between the content item text and pictogram should be 32px."

The extra spacing was caused by the SVG itself. I updated the path within the SVG to fill its available space and now the margins should look more correct. This was appearing in all instances of the Web Components component, not just Firefox and the React wrapper.

**Original:** 
Screen Shot 2022-02-16 at 12 41 45 PM

**Design specs:**
Screen Shot 2022-02-16 at 12 40 18 PM

**Updated:**
Screen Shot 2022-03-04 at 1 20 00 PM


oliviaflory:
@mwanberg this looks great, thank you!

I just want to double check with you, the original issue was written for `Web Components – React wrapper`, but I'm only seeing the updates in `Web Components` story book (it was correct to fix in Web components first!) Will we be doing the React wrapper work in a separate issue?

proeung:
@oliviaflory I think we can keep everything (Web Components and WC with React Wrapper) in one issue. @mwanberg Let's adjust the React Wrapper Storybook preview to include the proper SVG sizes - https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/cta-section/__stories__/cta-section.stories.react.tsx#L71

mwanberg:
I was so excited about the overall problem that I forgot to check the original. :) The React wrapper has been updated and should be reflected in its preview now.

proeung (approved):
Looks good on both WC and WC with React Wrapper!

Screen Shot 2022-03-08 at 3 59 36 PM
Screen Shot 2022-03-08 at 3 59 41 PM

kennylam (approved):
LGTM thanks @mwanberg!

oliviaflory (approved):
LGTM, thank you @mwanberg !

Issue #8307, opened 2022-02-16
[Filter panel]: Keyboard controls for checkboxes


RichKummer:
**Engineering info:**  
- Innovation Team engineer: @jkaeser 
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2423
- DPO consulting engineer: N/A

-----

### Description

Unclear if this is a bug - might be something we can discuss with Mike Gower.

Using keyboard controls with VoiceOver on in Firefox, I can check a box inside the filter panel menu. However, once I've changed a checkbox to active and start to tab again, I'm at the start of the page. This makes it difficult to navigate through the menu and use checkboxes.

https://user-images.githubusercontent.com/15643582/154327136-22bffc86-de7f-4002-91f1-5b760a2c2bd6.mov

### Component(s) impacted

Filter panel

### Browser

Firefox

### Carbon for IBM.com version

v1.31.0-rc.1

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

C4IBM.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://codesandbox.io/s/github/carbon-design-system/carbon-for-ibm-dotcom/tree/main/packages/web-components/examples/codesandbox/components-react/filter-panel

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #8492, opened 2022-03-14
fix(filter-panel): return focus to selected elements

jkaeser:
### Related Ticket(s)

Resolves https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8307

### Description

Filter panel elements (e.g. checkbox, select item) were losing focus when clicked/selected. This was happening because the whole filter panel is re-rendered whenever an element is clicked, and as far as the browser is concerned, the element that had focus disappeared. As a result, it was dropping focus back to the window.

To fix this, the composite now tracks a matching selector for any element gets selected. It uses this selector to find that element and reapply focus to it after any re-renders.

### Changelog

**New**

- Adds `_focusElement` state property to `DDSFilterPanelComposite` elements. This property tracks which filter element was last selected and is used to reapply focus to that element after re-renders.

**Changed**

- Renames `DDSFilterPanelComposite` events and event listeners to better match those events' purposes.


IgnacioBecerra (commented):
LGTM, really like the mobile/desktop specific `querySelector`s! Thanks @jkaeser!

IgnacioBecerra (approved):

kennylam (approved):
LGTM thanks @jkaeser!

RichKummer (approved):
Tested in FF, and this is working perfectly. TY @jkaeser !

Issue #8334, opened 2022-02-17
[Masthead]: searchNoRedirect prop no longer working in 1.30.0


PatrikHysky:
### Description

`<Masthead` `searchNoRedirect>`

It works correctly in 1.24.0 but no longer works in 1.30.0 upon pressing the Enter key while inside the search box. It does not block the default redirect action as it should and performs the redirect anyway. Using identical code. This breaks my application.

### Component(s) impacted

Masthead

### Browser

Chrome, Safari, Firefox, Microsoft Edge

### Carbon for IBM.com version

1.30.0

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

Redbooks

### Package

@carbon/ibmdotcom-react

### CodeSandbox example

https://codesandbox.io/s/github/carbon-design-system/carbon-for-ibm-dotcom/tree/main/packages/react/examples/codesandbox/components/Masthead?file=/src/App.js

### Steps to reproduce the issue (if applicable)

Use component `<Masthead searchNoRedirect />`

Type a keyword into search box. Press ENTER. This should result in no action as the searchNoRedirect prop should be blocking the default redirect. Instead, it redirects.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

gdelory:
Hi, any idea when this will be released?

PR #8516, opened 2022-03-16
fix(masthead-search): react searchNoRedirect prop doesn't work on enter

ariellalgilmore:
### Related Ticket(s)

#8334

### Description

people using the `searchNoRedirect` prop and press enter within the search box are still being redirected. The `onKeyDown` was removed in this [PR](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/7365), but it is actually necessary when the `searchNoRedirect` is used

### Changelog

**New**

- added back the `onKeyDown` function


emyarod (approved):
looks good to me! not sure if we want to include a `break` and default case in the switch statement for best practices, but right now the switch is fine as is

IgnacioBecerra (approved):

Issue #8330, opened 2022-02-17
[Test Scenario]: Carousel - (Default, with images, with videos, with Media)


proeung:
**Engineering info:**  
- Innovation Team engineer: @mwanberg 
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2409
- DPO consulting engineer: @IgnacioBecerra 

-----

### Component and test scenario title

Carousel - (Default, with images, with videos, and with mixture of images and videos)

### Test scenario steps

**Carousel - Default**

1. Check the carousel cards and ensure that the content is loaded and the whole card clickable. 
2. Ensure that the `Next` and `Previous` buttons advance the carousel.
3. Check the content (heading, copy, and arrow link) on the carousel

**Carousel - With Images**

1. Ensure that the images are rendered correctly.
2. Ensure that the body text has the same alignment.

**Carousel - With Videoes**

1. Ensure that the video thumbs and play buttons are rendered correctly.
2. Ensure that the body text has the same alignment.
3. Ensure that the duration text value is visible in the CTA area.

**Carousel - With a mixture of images and videos (New SB Variant)**

1. Ensure that the body text has the same alignment. (see https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8247)

proeung:
Adding `sev 3` to this e2e ticket as we want to make sure we have test coverage for the Carousel component which is being used on multiple templates within AEM.

PR #7807, opened 2021-12-02
test(e2e): multiple fixes for browserstack automate

jeffchew:
### Related Ticket(s)

No related issue

### Description

This fixes a number of issues with browserstack and how it handles tests. React seems fine, still sorting through web components but can probably do a followup PR to do additional fixes if needed.

### Changelog

**Changed**

- Multiple fixes across e2e tests

PR #8538, opened 2022-03-17
test(carousel): add testing for carousel default, with images, with videos, with media

mwanberg:
### Related Ticket(s)

[[Test Scenario]: Carousel - (Default, with images, with videos, with Media)](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8330)

### Description

The original requirements can be found in the ticket. I erred a little on the verbose side, and gave each variant a full suite of tests applicable to its unique content. If I should pare back and if any are redundant, let me know.

#### Test scenario steps

Note: Accessibility and theme compliance tests were applied to all.

**Carousel - Default**

1. Check the content (heading, copy, and arrow link) are appearing on the carousel cards
2. Check the headers to make sure they all have the same height and are not set to 0px tall, providing an even vertical height for all the copy text
3. Check that the whole card is clickable
4. Ensure that the `Next` and `Previous` buttons move the carousel

**Carousel - With Images**

1. Check the content (heading, copy, and arrow link) are appearing on the carousel cards
2. Ensure that the images are rendered
2. Check the headers to make sure they all have the same height and are not set to 0px tall, providing an even vertical height for all the copy text
4. Check that the whole card is clickable
5. Ensure that the `Next` and `Previous` buttons move the carousel

**Carousel - With Videoes**

1. Check the content (heading, copy, and arrow link) are appearing on the carousel cards
2. Ensure that the video thumbs and play buttons are rendered correctly
3. Ensure that the duration text value is visible in the CTA area
2. Check the headers to make sure they all have the same height and are not set to 0px tall, providing an even vertical height for all the copy text
4. Check that the whole card is clickable
5. Ensure that the `Next` and `Previous` buttons move the carousel

**Carousel - With a mixture of images and videos (New SB Variant)**

1. Check the content (heading, copy, and arrow link) are appearing on the carousel cards
2. Ensure that the images are rendered
3. Ensure that the video thumbs and play buttons are rendered correctly
4. Ensure that the duration text value is visible in the CTA area
2. Check the headers to make sure they all have the same height and are not set to 0px tall, providing an even vertical height for all the copy text
5. Check that the whole card is clickable
6. Ensure that the `Next` and `Previous` buttons move the carousel

### Changelog

**New**

- Test for the Carousel component


andy-blum:
Looks like it's just a formatting error remaining. `yarn format` should fix it. All [the tests you wrote](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/runs/5581094446?check_suite_focus=true#step:8:382) passed.

packages/web-components/tests/e2e-storybook/cypress/integration/carousel/carousel.e2e.js
```diff
@@ -0,0 +1,343 @@
...
+
+'use strict';
+
+/**
+ * Sets the correct path
+ *
+ * @param default - Path to default variant
+ * @param withImages - path to variant with images
+ * @param withCta - path to variant with videos
+ * @param withCta - path to variant with both images and videos
```
andy-blum, line 16: Good work documenting the code, but lets make sure the @param names match the keys in the object

packages/web-components/tests/e2e-storybook/cypress/integration/carousel/carousel.e2e.js
```diff
@@ -0,0 +1,343 @@
...
+ * @private
+ */
+const _paths = {
+  default: '/iframe.html?id=components-carousel--default',
+  withImages: 'iframe.html?id=components-carousel--cards-with-images',
```
andy-blum, line 21: The tests passed, so this might just be nitpicky, but can we make sure the paths all follow the same pattern (either start with leading slash or dont)

packages/web-components/tests/e2e-storybook/cypress/integration/carousel/carousel.e2e.js
```diff
@@ -0,0 +1,343 @@
...
+ * @private
+ */
+const _selectorBase = `[data-autoid="dds--carousel"]`;
+
+/**
+ * Defines the card element selectors.
+ *
+ * @type {Object.<string>}
+ * @private
+ */
```
andy-blum, line 39: Could be good to add @param keys to this (like the _paths documentation above)

packages/web-components/tests/e2e-storybook/cypress/integration/carousel/carousel.e2e.js
```diff
@@ -0,0 +1,343 @@
...
+              const url = $link.prop('href');
+              expect(url).not.to.be.empty;
+
+              const win = $link[0].ownerDocument.defaultView;
+              const after = win.getComputedStyle($link[0], ':after');
+              const positionValue = after.getPropertyValue('position');
+              const insetValue = after.getPropertyValue('inset');
+
+              expect(positionValue).to.eq('absolute');
+              if (Cypress.browser.name !== 'firefox') {
```
andy-blum, line 184: Why are we excluding firefox here?

packages/web-components/tests/e2e-storybook/cypress/integration/carousel/carousel.e2e.js
```diff
@@ -0,0 +1,343 @@
...
+
+        if ($carousel.find(_selectors.videoFooter).length > 0) {
+          cy.get(_selectors.videoFooter)
+            .shadow()
+            .find('a.bx--card__footer')
+            .then($link => {
+              const url = $link.prop('href');
+              expect(url).not.to.be.empty;
+
+              const win = $link[0].ownerDocument.defaultView;
```
andy-blum, line 178: cy.window() again

packages/web-components/tests/e2e-storybook/cypress/integration/carousel/carousel.e2e.js
```diff
@@ -0,0 +1,343 @@
...
+  },
+  checkScroll: () => {
+    it('should scroll forward when Next button is clicked and back when the Previous button is clicked', () => {
+      cy.get(_selectors.buttonNext).click();
+
+      cy.takeSnapshots();
+
+      cy.get(_selectors.buttonPrevious).click();
+
+      cy.takeSnapshots();
```
andy-blum, line 200: Nit: I think you ought to be able to chain the takeSnapshots into these and only need to call on `cy` once.

Not Nit: Does the scroll happen instantly or is it using smoothScroll behavior? If the scroll takes time, we might want to throw a `.wait(1000)` in here to make sure the page has time to scroll before we take snapshots

packages/web-components/tests/e2e-storybook/cypress/integration/carousel/carousel.e2e.js
```diff
@@ -0,0 +1,343 @@
...
+    cy.injectAxe();
+  });
+
+  it('should check a11y', _tests.checkA11y);
+  it('should render correctly in all themes', _tests.screenshotThemes);
+
+  _tests.checkTextRenders();
+  _tests.checkSameHeight();
+  _tests.checkClickableCard();
+  _tests.checkScroll();
```
andy-blum, line 218: You're sort of mixing styles here, I think we probably want all the `it('does a thing', function())` syntax to be down here inside the `describe` function

andy-blum (changes requested):
Looks like a solid foundation here. Mainly just opinions and nitpicks in the comments. Let me know if you have questions/need more info!

packages/web-components/tests/e2e-storybook/cypress/integration/carousel/carousel.e2e.js
```diff
@@ -0,0 +1,343 @@
...
+  checkSameHeight: () => {
+    it('should have headings all the same height', () => {
+      cy.get(_selectors.heading)
+        .wait(1000)
+        .then($heading => {
+          let headingHeightNum = $heading.first().height();
+          let headingHeight = headingHeightNum + 'px';
+
+          expect($heading).to.not.have.css('height', '0px');
+          expect($heading).to.have.css('height', headingHeight);
```
jkaeser, line 143: I don't know for sure if this is testing what you think. Since `$heading` is a jQuery object with multiple elements in it, I have a suspicion this only tests the first element, in which case this test only proves that the first element has the same height as itself.

Also, you might consider testing for the headers' computed heights rather than checking for a specific CSS style. This has the benefit make the test implementation agnostic.

packages/web-components/tests/e2e-storybook/cypress/integration/carousel/carousel.e2e.js
```diff
@@ -0,0 +1,343 @@
...
+
+describe('dds-carousel | default (desktop)', () => {
+  beforeEach(() => {
+    cy.viewport(1280, 720);
+    cy.visit(`${_paths.default}`);
+    cy.injectAxe();
+  });
+
+  it('should check a11y', _tests.checkA11y);
+  it('should render correctly in all themes', _tests.screenshotThemes);
```
jkaeser, line 213: Can you handle these tests in the same way as the ones below? You should be able to move the `it('should...)` parts into the functions in your `_tests` object.

jkaeser (changes requested):
@mwanberg This is looking great! I left a few notes for your consideration. Mostly little nitpicks and clean ups, and a couple suggestions that should make the tests more robust.

packages/web-components/tests/e2e-storybook/cypress/integration/carousel/carousel.e2e.js
```diff
@@ -0,0 +1,341 @@
...
+      cy.get(_selectors.videoFooter)
+        .find('span.bx--card__cta__copy')
+        .then($duration => {
+          expect($duration).not.to.be.empty;
+        });
+    });
+  },
+  checkSameHeight: () => {
+    it('should have headings all the same height', () => {
+      cy.get(_selectors.heading).wait(1000).then($heading => {
```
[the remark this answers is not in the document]
jkaeser: Oops! That GitHub degradation time left this in here.

packages/web-components/tests/e2e-storybook/cypress/integration/carousel/carousel.e2e.js
```diff
@@ -0,0 +1,343 @@
...
+        .find('span.bx--card__cta__copy')
+        .then($duration => {
+          expect($duration).not.to.be.empty;
+        });
+    });
+  },
+  checkSameHeight: () => {
+    it('should have headings all the same height', () => {
+      cy.get(_selectors.heading)
+        .wait(1000)
```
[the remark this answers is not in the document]
andy-blum: I had a similar question - I'm assuming it's to give the setSameHeight function time to run and re-render, yes?

packages/web-components/tests/e2e-storybook/cypress/integration/carousel/carousel.e2e.js
andy-blum: Note: this is counter to what I suggested above. EIther way is fine, really, but consistency is the key to this.

jkaeser: Oh, you're right. FWIW I suggested this approach so you don't have to copy the test description in each grouping.

proeung:
@mwanberg Can you take a look at the feedback that @andy-blum and @jkaeser have given?

packages/web-components/tests/e2e-storybook/cypress/integration/carousel/carousel.e2e.js
```diff
@@ -0,0 +1,343 @@
...
+      cy.get(_selectorBase).then($carousel => {
+        if ($carousel.find(_selectors.footer).length > 0) {
+          cy.get(_selectors.footer)
+            .shadow()
+            .find('a.bx--card__footer')
+            .then($link => {
+              const url = $link.prop('href');
+              expect(url).not.to.be.empty;
+
+              const win = $link[0].ownerDocument.defaultView;
```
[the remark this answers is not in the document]
mwanberg: @andy-blum I stole this section directly from the existing Card e2e test: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/b4f02244b83539b2e30cc0240c55348395db9051/packages/web-components/tests/e2e-storybook/cypress/integration/card/card.e2e.js#L105

Happy to update it if needed, but figured I'd save time and use what's already working.

packages/web-components/tests/e2e-storybook/cypress/integration/carousel/carousel.e2e.js
mwanberg: As mentioned above, this code is a copy of very similar code elsewhere. Looking at the git history, it seems that was added in this PR: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/7807 It seemed prudent to keep it in, though if it's now unneeded I can remove it.

packages/web-components/tests/e2e-storybook/cypress/integration/carousel/carousel.e2e.js
mwanberg: Good points. I chained them and added a wait.

packages/web-components/tests/e2e-storybook/cypress/integration/carousel/carousel.e2e.js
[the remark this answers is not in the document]
mwanberg: Yes, it was breaking in certain circumstances without the wait. However, see below for more to this story...

packages/web-components/tests/e2e-storybook/cypress/integration/carousel/carousel.e2e.js
```diff
@@ -0,0 +1,343 @@
...
+    cy.checkAxeA11y();
+  },
+  screenshotThemes: () => {
+    cy.carbonThemesScreenshot({
+      capture: 'viewport',
+    });
+  },
+  checkTextRenders: () => {
+    it('should render card text and arrow icon', () => {
+      cy.get(_selectors.heading).then($heading => {
```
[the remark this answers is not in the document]
mwanberg: Done! Though I had to add a wait to account for sameHeight delaying the rendering.

packages/web-components/tests/e2e-storybook/cypress/integration/carousel/carousel.e2e.js
mwanberg: I made it all consistent. 👍

packages/web-components/tests/e2e-storybook/cypress/integration/carousel/carousel.e2e.js
mwanberg: @jkaeser You were very, very right. I tested by removing the `sameHeight` code from the `carousel.ts` file and sure enough, it passed with the code I originally had. So I added an `each()` to loop through them and check each one's height against the first. It now breaks appropriately when `sameHeight` is missing.

_However_, now it is passing on desktop, but not in mobile. 🤔 Even when I add a long `wait()`, for some reason the `sameHeight` is not getting applied. I could use you and @andy-blum 's suggestions on this. Of note, I'm seeing that the `sameHeight` isn't applying correctly in the scroll check, either (you can see it in the Percy screenshots).

packages/web-components/tests/e2e-storybook/cypress/integration/carousel/carousel.e2e.js
andy-blum: I would remove this for now until someone tells you to add it back in.

packages/web-components/tests/e2e-storybook/cypress/integration/carousel/carousel.e2e.js
andy-blum: IIRC the sameHeight code only causes the same height on objects that are horizontally next to each other. If the elements stack vertically, they likely shouldn't get the height adjustments.

Are there any elements here that are still next to each other at the mobile width?

mwanberg: Nope, at mobile the cards are not visible next to each other except when sliding. I'll add a check for the viewport size.

jkaeser (approved):

andy-blum (approved):

packages/web-components/tests/e2e-storybook/cypress/integration/carousel/carousel.e2e.js
```diff
@@ -0,0 +1,349 @@
...
+  _tests.checkTextRenders();
+  _tests.checkSameHeight();
+  _tests.checkClickableCard();
+  _tests.checkScroll();
+});
+
+describe('dds-carousel | default (mobile)', () => {
+  beforeEach(() => {
+    cy.viewport(320, 720);
+    cy.visit(`/${_paths.default}`);
```
kennylam, line 241:
```suggestion
    cy.visit(`${_paths.default}`);
```

kennylam (commented):
Thanks @mwanberg! Just one small fix and it looks like it's good to go!

mwanberg:
> Thanks @mwanberg! Just one small fix and it looks like it's good to go!

Done!

kennylam (approved):
LGTM thanks @mwanberg!

Issue #8552, opened 2022-03-17
[Table of contents]: Mobile select has no text


IgnacioBecerra:
### Description

In Mobile view for Web Components Table of Contents, the `<select>` element is populated, however when choosing an option, although the functionality remains, the selected option vanishes right away. It also appears as if the option was never selected, since the currently "selected" option remains the default one.

https://user-images.githubusercontent.com/24970122/158906681-ffd721e7-8932-4616-b669-e61aff10c8a4.mov

### Component(s) impacted

Table of Contents

### Browser

Chrome

### Carbon for IBM.com version

canary

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

N/A

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://carbon-design-system.github.io/carbon-for-ibm-dotcom/canary/web-components/?path=/story/components-table-of-contents--default

### Steps to reproduce the issue (if applicable)

1. Open ToC story in mobile view
2. Select an option
3. Text vanishes

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

Issue #8271, opened 2022-02-14
[TOC - Default]: Web Components - TOC links navigate to a point where section titles are obscured in Mobile Chrome


QALosey:
**Engineering info:**  
- Innovation Team engineer: @andy-blum 
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2402
- DPO consulting engineer: 

-----

### Description

Expected Results: 
  When viewing the TOC - Default component on Mobile Chrome, selecting a TOC link will direct the user to the relevant section of the page with the section headline displayed just below the page header.
 
Actual Results: 
  After selecting a TOC link in Mobile Chrome, the user is directed to the relevant section of the page, but at a point where the section headline is partially obscured by the page header.

### Component(s) impacted

Table of Contents - Default
![TOC - Android](https://user-images.githubusercontent.com/57540897/153952863-1f107831-1a56-4daf-9f59-fbc6a60e248d.jpg)

### Browser

Chrome

### Carbon for IBM.com version

v1.16.0-rc.1

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

https://carbon-design-system.github.io/carbon-for-ibm-dotcom/next/web-components/

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://carbon-design-system.github.io/carbon-for-ibm-dotcom/next/web-components/?path=/story/components-table-of-contents--default

### Steps to reproduce the issue (if applicable)

1. Open https://carbon-design-system.github.io/carbon-for-ibm-dotcom/next/web-components/?path=/story/components-table-of-contents--default on an Android device in Chrome browser
2. Select 'Canvas' to view the TOC web component
3. Open the drop down and select a content item
4.  Observe that the user is navigate to the relevant section of the page, but the section headline is partially obscured

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

proeung:
@QALosey Thanks for opening this bug issue!

[...]

PR #8553, opened 2022-03-17
fix(table-of-contents): fixing mobile issues

IgnacioBecerra:
### Related Ticket(s)
#8271
#8552 

### Description
Original issue was to add a bit of padding so the mobile ToC was able to show the title upon selection, but there was another issue where the `<select>` element wasn't updating its text and selection, so I resolved that issue in this PR as well.

### Changelog

**New**

- added a mobile check to add a bit more padding upon scrolling to the selection to ensure it can be read

**Changed**

- using `getAttribute('name')` as the `name` declaration has been deprecated and thus not working properly
- updated the `retriggerHarvest` function to follow the new structure after a previous refactor


ariellalgilmore (approved):
LGTM! Wondering if we need to open a different issue though for this blue focus state happening on mobile?
IMG_2CA9BBDFC3FF-1

Issue #8294, opened 2022-02-15
[Layout]: Column background and spacing


RichKummer:
### Description

In the Layout story, when changing to any of the gray themes, there's a white background behind the Column content that makes the content harder to read. Also, there seems to be spacing taken away from the sides and top that is not present in the current WC build.

**Layout in React - Gray90**
<img width="1403" alt="Screen Shot 2022-02-15 at 3 52 28 PM" src="https://user-images.githubusercontent.com/15643582/154147728-b099da23-d900-4171-aff3-975a61809e78.png">

**Layout in WC - Gray90**
<img width="1437" alt="Screen Shot 2022-02-15 at 3 52 18 PM" src="https://user-images.githubusercontent.com/15643582/154147796-681ff3b6-9c2f-455d-9a3d-06d75b287130.png">

### Component(s) impacted

Layout

### Browser

Firefox

### Carbon for IBM.com version

v1.31.0-rc.1

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

C4IBM.com

### Package

@carbon/ibmdotcom-react

### CodeSandbox example

https://codesandbox.io/s/github/carbon-design-system/carbon-for-ibm-dotcom/tree/main/packages/react/examples/codesandbox/components/Layout

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #8554, opened 2022-03-17
fix(layout): removed unneeded white background inline styles

IgnacioBecerra:
### Related Ticket(s)
#8294

### Description
In React, there was an inline style for a white background in one of the elements, but it seems to be unneeded nowadays. The component can be properly seen in different themes with the proper colored backgrounds.

### Changelog

**Removed**

- white background inline styles in the component story


annawen1 (approved):
LGTM!

Issue #8467, opened 2022-03-08
[Leaving IBM]: links missing functionality to trigger modal


kennylam:
**Engineering info:**  
- Innovation Team engineer: @jkaeser 
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2433
- DPO consulting engineer: N/A

-----

### Description

Currently the only way to trigger the `leaving-ibm` modal is manually, on a per-link basis. All links should have a means to easily open the modal, as well as replacing the `href` attribute to where the page should be redirected. Documentation for this must also be included.

Possible solution – `leaving-ibm` component scans page for links with an attribute, e.g. `type="external`, and binds an event listener to them which would toggle the modal and update the `href` on click.

### Component(s) impacted

`ibmdotcom/web-components/leaving-ibm`

### Carbon for IBM.com version

v1.31.1

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

Carbon for IBM.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://codesandbox.io/s/github/carbon-design-system/carbon-for-ibm-dotcom/tree/main/packages/web-components/examples/codesandbox/components/leaving-ibm

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #8583, opened 2022-03-22
fix(leaving-ibm): provide attribute to trigger modal

jkaeser:
### Related Ticket(s)

Resolves https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8467

### Description

Currently, adopters have to write their own code to trigger the `DDSLeavingIBM` element when an external link is clicked. This PR makes it so that adding the `data-leaving-ibm` attribute to links and link-like components will trigger the modal when clicked.

#### Example
```html
<dds-leaving-ibm-container></dds-leaving-ibm-container>

...

<bx-link href="https://www.example.com" data-leaving-ibm>
  Clicking will open the Leaving IBM modal.
</bx-link>
```

### Changelog

**New**

- Configures `DDSLeavingIBM` elements to activate when any link with the `data-leaving-ibm` attribute is clicked.
- Adds Storybook stories to demonstrate external link behavior with various components.


packages/web-components/src/components/leaving-ibm/leaving-ibm-composite.ts
```diff
@@ -74,6 +77,36 @@ class DDSLeavingIbmComposite extends ModalRenderMixin(LitElement) {
...
+      const { target } = event;
+      const linkTarget =
+        target instanceof HTMLAnchorElement
+          ? target
+          : (event.composedPath().find(pathTarget => {
+              return pathTarget instanceof HTMLAnchorElement;
+            }) as HTMLAnchorElement | undefined);
+
+      if (linkTarget) {
+        const linkIsExternal = linkTarget.hasAttribute('data-external');
```
jkaeser, line 92: Does anyone have any opinions on the name of this attribute? `'data-external'` feels a bit too generic. Maybe `'data-leaving-ibm'`?

packages/web-components/src/components/leaving-ibm/leaving-ibm-composite.ts
```diff
@@ -74,6 +77,36 @@ class DDSLeavingIbmComposite extends ModalRenderMixin(LitElement) {
...
+        const targetIsExternal = target instanceof Element && target.hasAttribute('data-external');
```
jkaeser, line 93: Does anyone have any opinions on the name of this attribute? `'data-external'` feels a bit too generic. Maybe `'data-leaving-ibm'`?

annawen1: 🤔 hmmm yeah `data-leaving-ibm` is good with me! That way it's clear it's connected to the leaving ibm modal

proeung (commented):
Looks good, but curious as to why these variants have been added. Not sure if we need to show the different CTA types here, but I'll defer to @oliviaflory and @kennylam on this.

Screen Shot 2022-03-23 at 4 32 51 PM

jkaeser:
> Looks good, but curious as to why these variants have been added.

@proeung I added this selection of elements to demonstrate the feature and provide an end-to-end testing surface.

proeung:
@oliviaflory Thanks for getting back to us. What you've outlined for the `Leaving IBM modal (Default)` story looks good, however, these requirements were not outlined in the original ticket (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8467), thus we'll need to expand the SOW of the issue to account for these different knob types.

@jkaeser Let's get these knobs added to this PR as best as you can. Any follow-up items (time-consuming) can be addressed in a separate issue.

jkaeser:
@oliviaflory Thanks for your feedback! I've made the changes you requested to the stories.

The way this work was done, as long as the `<dds-leaving-ibm-container>` is also added somewhere in the document, any component that you can put a `href` attribute on can be made to trigger the modal by giving it the `data-leaving-ibm` attribute. In other words, there shouldn't be any additional updates required to support all the CTA types you listed.

oliviaflory:
> The way this work was done, as long as the <dds-leaving-ibm-container> is also added somewhere in the document, any component that you can put a href attribute on can be made to trigger the modal by giving it the data-leaving-ibm attribute. In other words, there shouldn't be any additional updates required to support all the CTA types you listed.

Awesome, thanks for updating those knobs @jkaeser I'll approve in a few minutes!

cc @kennylam 
1. Do we need to add an e2e test to make sure Percy is still capturing the Leaving IBM modal, now that you see the trigger in storybook?
2. Do we need a separate story for React wrapper update or does that happen by default?

oliviaflory (approved):
LGTM!

kennylam:
> 1. Do we need to add an e2e test to make sure Percy is still capturing the Leaving IBM modal, now that you see the trigger in storybook?

@oliviaflory Yes that's a good call.

@proeung @jkaeser I didn't add any additional link/cta types in the original issue because this component only supports one. That's what prompted Olivia to reach out to me when we saw other link types being used. If those are required then we should put them in a separate feature request.

proeung (approved):
@jkaeser LGTM!

@kennylam Sounds good! Let's capture the e2e test and the React wrapper items that @oliviaflory mentioned above in a separate ticket so that we move forward with the changes in this PR.

kennylam (commented):
Thanks for this @jkaeser it looks great! Only thing i see is that `data-leaving-ibm` is missing from the `@property` declarations. This prevents it from appearing in the [props](https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/8583/index.html?path=/docs/components-leaving-ibm--default#props) table in the documentation. It would basically be just like the `open` property.

jkaeser:
> Only thing i see is that data-leaving-ibm is missing from the @property declarations. This prevents it from appearing in the [props](https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/8583/index.html?path=/docs/components-leaving-ibm--default#props) table in the documentation. It would basically be just like the open property.

@kennylam I'm not sure it make sense to add this to the props documentation. It's an attribute this component looks for, not one that is set on the component itself. I documented how this is meant to work in [the HTML section](https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/8583/index.html?path=/docs/components-leaving-ibm--default#html) (though looking at that now I should change the `bx-link` element to a `dds-link-with-icon` element...).

Thoughts?

kennylam (approved):
> It's an attribute this component looks for, not one that is set on the component itself.

Hm, right, this is a good point. Thanks @jkaeser!

Issue #6710, opened 2021-07-22
[Web Components] Failed to execute 'define' on 'CustomElementRegistry': this name has already been used with this registry


jeffchew:
## Detailed description

> Describe in detail the issue you're having.

The following errors seems to sporadically occur with Web Components CDN, and some ES import code sandbox examples:

```
Uncaught DOMException: Failed to execute 'define' on 'CustomElementRegistry': this name has already been used with this registry
```

This also seems to consistently happen with the component `dds-content-block-card-static`. The Codesandbox example is also blank, which might be a different issue (ES version, CDN version is fine).

Content Block Card Static Example: (https://webcomponents-codesandbox.s3-web.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/6681/content-block-card-static/index.html)

> Is this issue related to a specific component?

Seems to be random, but noticeable more with the content block components.

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Should try to remove these errors if possible. There is no way to unregister a custom element once it is defined, so the other way is to do defense detection so that a custom element cannot be declared more than once.

Tried the following approach: https://stackoverflow.com/questions/54211849/uncaught-domexception-failed-to-execute-define-on-customelementregistry-th. This works for web components, but unfortunately will break the react wrapper generation as I believe it relies on the custom-element.json that is generated by the `@customElement` declaration.

## Additional information

- This might pose an issue when we start to do more feature flag development, and injection of canary versions of components that live alongside production components (this was requested by the Adobe Target team).

PR #8608, opened 2022-03-28
fix(rollup): use non-hash filenames for CDN chunk files

jeffchew:
### Related Ticket(s)

Refs https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6710, https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8609

### Description

Whenever there was a CDN publish of a tag version of Carbon for IBM.com Web Components, it would be briefly out of sync with akamai cache as the changes propagated globally. This would cause issues where custom elements have already been defined when trying to load in differing but similar artifacts:

image

This change will update so that the chunk file names no longer use a hash, but an incremental number instead in order to have consistency in the file names as Akamai does its job to propagate changes to users.

### Changelog

**Changed**

- Introduce `chunkFileNames` custom configuration for rollup config
- Moved RTL output to subfolder
- Updated corresponding documentation for RTL artifacts
- Fixes to codesandbox examples to point to `latest` version instead of `canary`
- Added universal banner to CDN tests


jeffchew:
Triggered the codesandbox examples and e2e test runs with the `test: e2e` label.

jeffchew:
17 components seem to not be happy in the e2e integration tests:

image

image

kennylam (approved):
LGTM thanks @jeffchew!

Issue #8606, opened 2022-03-28
[Table of Contents]: Focus on selected heading on mobile


IgnacioBecerra:
### Description

On mobile, upon selecting an option on the ToC, the heading has a focused outline. Could be related to the new iOS update?
![image](https://user-images.githubusercontent.com/24970122/160472843-9fff1942-f4df-4b6f-a8b8-153136a0133b.png)

### Component(s) impacted

Table of Contents

### Browser

Safari

### Carbon for IBM.com version

canary

### Severity

Severity 4 = The problem is not visible to or noticeable to an average user. Affects minor functionality, no workaround needed.

### Application/website

N/A

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

N/A

### Steps to reproduce the issue (if applicable)

1. Select an option on the ToC in mobile
2. Notice the focus outline

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

oliviaflory:
Confirm that screen readers announce the heading if the focus is taken away (visually hidden).

ariellalgilmore:
confirmed!

PR #8659, opened 2022-03-29
fix(TOC): safari focus

ariellalgilmore:
### Related Ticket(s)
#8606

### Description

In safari when navigating to elements in TOC the default blue focus would show. I removed this and tested that VO still works properly

### Changelog

**New**

- no outline focus styles


kennylam (approved):
LGTM thanks @ariellalgilmore!

Issue #8473, opened 2022-03-11
[Accessibility issue with Masthead Search]: 2 violations found via IBM Accessibility Checker


Daynssjal:
### Description

Hello,
2 violations were reported regarding Masthead search component which we are using in our project: IBM Documentation.
These violations were found after scanning our page through IBM Equal Access Accessibility Checker chrome extension.
I then run the checker on the standalone dotcom component Masthead with open search, and saw the same 2 violations.

Screenshot attached.
<img width="958" alt="accessibility-violation-dotcom-masthead" src="https://user-images.githubusercontent.com/33804471/157833352-5855fed6-d74c-4d74-9cfb-6337d80e2a0d.PNG">

### Component(s) impacted

Masthead

### Browser

Chrome

### Carbon for IBM.com version

1.30.0

### Severity

Severity 1 = The design is broken in a critical way that blocks users from completing tasks or damages the brand. Affects major functionality, no workaround.

### Application/website

IBM documentation

### Package

@carbon/ibmdotcom-react, @carbon/ibmdotcom-styles

### CodeSandbox example

https://www.ibm.com/standards/carbon/react/iframe.html?id=components-masthead--search-open-onload 

### Steps to reproduce the issue (if applicable)

1. Go to Chrome webstore
2. Look for: IBM Equal Access Accessibility Checker  extension
3. Install the extension and restart browser
4. open the following link https://www.ibm.com/standards/carbon/react/iframe.html?id=components-masthead--search-open-onload  
5. open chrome devtools and choose Accessibility Assessment (next to console/networ/sources/performance)
6. scan the page
7. see the 2 violations 

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

jeffchew:
@IgnacioBecerra @ariellalgilmore does the work here fix what we're seeing here? https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8353

kennylam:
@Daynssjal What version of the package are you using? Testing it on our latest and can't reproduce the error.

Daynssjal:
Hi @kennylam, 1.30.0

ingridsamakova:
[...]

PR #8662, opened 2022-03-30
fix(masthead): fix search combobox `aria-expanded` value

kennylam:
### Related Ticket(s)

#8473

### Description

Currently the search combobox `aria-expanded` value is based on whether or not the suggestions list is open. This is incorrect, as the value should come from the search input (combobox) itself. This would cause the value to be `false` unless the suggestions list was visible.

Note: making this change in React introduced a new accessibility violation, which I believe is a false positive. The combobox popup (suggestions list) has no focus or child elements, and `aria-activedescendant` is only set when the list populates and a selection is highlighted. I'll follow up with the a11y team about this.

image

### Changelog

**Changed**

- change `aria-expanded` value to be based on search input state, not suggestions list

**Removed**

- unused React Storybook DOM element


annawen1 (approved):
LGTM!

PR #8610, opened 2022-03-28
chore(release): merge release/v1.32.0 branch

emyarod:
This PR merges the changes from the `release/v1.32.0` branch to `main`

PR #8671, opened 2022-04-01
chore: merge v1.32.1 release branch

emyarod:


jeffchew:
@emyarod this probably should be a manual merge, otherwise the tag would get lost with the deleted branch?

emyarod:
@jeffchew wouldn't this be similar to https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/8610?

jeffchew:
> @jeffchew wouldn't this be similar to #8610?

Ok, maybe this is why the latest builds have been failing since 1.32.0?

emyarod:
I see, so would this just be a direct push to `main`?

jeffchew:
> I see, so would this just be a direct push to `main`?

Possibly, either that or we enable a merge commit (not a squash commit) for just this PR?

jeffchew:
> > I see, so would this just be a direct push to `main`?
> 
> Possibly, either that or we enable a merge commit (not a squash commit) for just this PR?

I take that back, the tag is still on the release branch itself.

emyarod:
ok I see so a direct push to `main` branch would be the only option then?

jeffchew:
Yes I think so...I'm wondering what we do with the `1.32.0` tag. I see that the branch is still around and hasn't been deleted yet. Perhaps we just re-tag the squash merge with those tags once we delete the branch...

jeffchew (approved):
LGTM!

Issue #8285, opened 2022-02-15
[card-in-card]: (react wrapper) image not coming through


ariellalgilmore:
### Description

Expected there to be an image where there is a large white space

<img width="371" alt="Screen Shot 2022-02-15 at 9 33 48 AM" src="https://user-images.githubusercontent.com/20210594/154106836-7e88b741-1753-4b32-b38d-02e7d4cf54be.png">

### Component(s) impacted

card-in-card

### Browser

Chrome

### Carbon for IBM.com version

1.31.0-rc.1

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

carbon for ibm.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://carbon-design-system.github.io/carbon-for-ibm-dotcom/next/web-components-react/?path=/story/components-card-section-offset--default

### Steps to reproduce the issue (if applicable)

1. go to react wrapper story: https://carbon-design-system.github.io/carbon-for-ibm-dotcom/next/web-components-react/?path=/story/components-card-section-offset--default
2. should see image in the background

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

kennylam:
Fixed in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/8673.

Issue #8308, opened 2022-02-16
[Lead space search - with image]: Image background not appearing


RichKummer:
**Engineering info:**  
- Innovation Team engineer: @mwanberg 
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2414
- DPO consulting engineer: @IgnacioBecerra 

-----

### Description

In Firefox using React wrapper - an image should appear inside the `Lead space search - with image`, but no image appears in the background.

<img width="1478" alt="Screen Shot 2022-02-16 at 1 03 12 PM" src="https://user-images.githubusercontent.com/15643582/154327687-20e6f359-14c8-42a7-aa31-d86e9c6e0ac3.png">

### Component(s) impacted

Lead space search - with image

### Browser

Firefox

### Carbon for IBM.com version

v1.31.0-rc.1

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

C4IBM.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://carbon-design-system.github.io/carbon-for-ibm-dotcom/next/web-components-react/?path=/story/components-lead-space-search--with-image

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

andy-blum:
I'm not sure this is a bug so much as it's a feature that's not available in the react wrapper.

The leadspace-with-search component is utilizing the `<dds-background-media>` component which is not a part of the react-wrapper stories, and the only other react-wrapper story to use background-media is the card-section-offset where the image also fails to load.

The leadspace component variations within react-wrapper are still utilizing the old setup of dds-leadspace-image.

Are we expecting `<dds-background-media>` to work in this environment at this time or not?

proeung:
@kennylam Please take a look at @andy-blum's comment. I'd like to get your input on what direction we should take for the react wrapper version of this leadspace component as well.

kennylam:
@proeung This definitely is a bug and should be fixed. Agree on removing the severity since this is more of a 3 to me.

PR #8673, opened 2022-04-02
fix(background-media): add check for default-src

mwanberg:
### Related Ticket(s)

#8308 [[Lead space search - with image]: Image background not appearing](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8308)

### Description

The original ticket pointed out that in the "React wrapper - an image should appear inside the `Lead space search - with image`, but no image appears in the background." While investigating, I found that the `Card section offset` was also missing its background image.

This happened because of some changes in the `Background media` component.

The fix should really be credited to @IgnacioBecerra as he was the one who finally tracked down the problem. :)

### DIFF
--- a/packages/web-components/src/components/background-media/background-media.ts
+++ b/packages/web-components/src/components/background-media/background-media.ts
@@ -167,6 +167,10 @@ class DDSBackgroundMedia extends DDSImage {
     if (this.parentElement instanceof DDSLeadSpace) {
       this.gradientHidden = true;
     }
+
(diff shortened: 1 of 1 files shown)


kennylam (approved):
LGTM thanks @mwanberg! Also fixes https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8285.

proeung (approved):
LGTM as well!

Screen Shot 2022-04-04 at 12 58 58 PM

Issue #8420, opened 2022-02-28
[Pricing table] Implement table header fixed scroll behavior


proeung:
**Engineering info:**  
- Innovation Team engineer: @jkaeser 
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2425
- DPO consulting engineer: @IgnacioBecerra 

-----

#### User Story
> As a `[user role below]`:
Carbon for ibm.com developer

> I need to:
create/change the `Pricing table`

> so that I can:
provide the ibm.com adopter developers components they can use to build ibm.com web pages

#### Additional information
- Story within Storybook with corresponding knobs
- Utilize Carbon
- Create with Shadow DOM and Custom Elements standards
- The React version is created as a wrapper to the Web Components version
- **See the Epic (#8389) for the Design and Functional specs information**
- Visual QA testing issue (#8391)
- Prod QA testing issue (#8392)

#### Acceptance criteria
- [ ] Include README for the web component and corresponding styles
- [ ] Create Web Components styles in styles package
- [ ] Do not create knobs in Storybook that include JSON objects
- [ ] Break out Storybook stories into multiple variation stories, if applicable
- [ ] Create codesandbox example under `/packages/web-components/examples/codesandbox` and include in README
- [ ] Minimum 80% unit test coverage
- [ ] add the trigger to create the React Wrapper
 - [ ] create the React storybook stories and related documentation
- [ ] Update the Carbon for ibm.com website component [data file](https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/blob/master/src/data/components.json) to be sure the web site Component Status and Storybook links are correct
- [ ] Use the [Visual QA checklist](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/wiki/Definition-of-done-(Visual-QA-checklist)) to verify design quality
- [ ] Update the e2e test scripts, if needed. 
- [ ] Confirm e2e test results have no major errors.
- [ ] A comment is posted in the Prod QA issue (#8392) when development is finished.  
- [ ] The `dev complete` label is added to the Prod QA issue (#8392).

proeung:
Marking this issue as "on-hold" pending additional design exploration from @jwitkowski79 and @RichKummer (see - https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/8824). This feature might not be able to get in for the `v1.35.0` code freeze period.

PR #8674, opened 2022-04-04
feat(pricing-table): create pricing table component

jkaeser:
### Related Ticket(s)

Resolves https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8393

### Description

Adds Pricing Table web component.

### Changelog

**New**

- Adds `DDSPricingTable` and supporting web components.
- Creates Storybook stories and documentation for Pricing Table.

**Changed**

- Adds class names to selectors in `DDSStructuredList`'s stylesheet.
- Modifies `structured-list-colspan` mixin to accommodate `--default-cols` CSS custom property.
- Adds overflow behavior to Structured List (which is inherited by Pricing Table) to accommodate wide tables.


packages/web-components/src/components/pricing-table/pricing-table.scss
```diff
@@ -0,0 +1,11 @@
+//
+// Copyright IBM Corp. 2020, 2022
+//
+// This source code is licensed under the Apache-2.0 license found in the
+// LICENSE file in the root directory of this source tree.
+//
+
+// @import '@carbon/ibmdotcom-styles/scss/components/pricing-table/pricing-table';
+@import '../../../../styles/scss/components/pricing-table/pricing-table';
```
jkaeser, line 9: Temporary while in active development.

jkaeser:
While this is still a WIP, I understand we want to get #8393 in this week if at all possible, and I think things are in a good enough place to do some preliminary design QA if you have the time, @oliviaflory. Specifically, I could use guidance for:

1. **Column widths:** I noticed in the design that columns' widths change depending on how many there are in the table. Could you map out how these should be sized?
2. **Tag link and tooltip toggle interactions:** Each of these can be interacted with, but I didn't find specs for their interaction states (e.g. on hover). Do you have any thoughts on these?

Some of the more glaring style issues I'm still working on:
- Mobile styling issues
- Group header widths at 6 columns
- lining up tags, whether or not the markup is present

jkaeser:
@oliviaflory @jwitkowski79 One more question! The design token specified for the table highlight label text, `text-01`, does not work for the dark themes. Is there another token I can use instead?

Screen Shot 2022-04-05 at 4 23 42 PM

oliviaflory:
> @oliviaflory @jwitkowski79 One more question! The design token specified for the table highlight label text, text-01, does not work for the dark themes. Is there another token I can use instead?

Good catch @jkaeser the text-01 shouldn't work on the light theme either so I'm surprised we didn't see that there too.
light themes #text-01 = #161616 which wouldn't work on top of the dark highlight.

Tokens should not be changing between themes, all components should be speced with a single theme, so I'm curious what we have that text set to in the white theme that's allowing it to work?
Screen Shot 2022-04-05 at 5 28 03 PM


We should update the specs to follow a single theme:
background: `$ui-05`
text: `$inverse-01`

packages/web-components/src/components/pricing-table/pricing-table.ts
```diff
@@ -0,0 +1,98 @@
...
+    const wrapper = this.shadowRoot?.getElementById(DDSStructuredList.wrapperId) || this;
+    (async () => {
+      return this.querySelector(`${ddsPrefix}-pricing-table-highlight-label`);
+    })()
+      .then(value => {
+        wrapper.style.marginTop = `${value?.getBoundingClientRect().height}px`;
+      })
+      .catch(() => {
+        wrapper.style.marginTop = '';
+      });
```
andy-blum, line 71: I don't think any part of this is asynchronous - should this just be if/else logic?

packages/web-components/src/components/structured-list/structured-list.scss
```diff
@@ -1,5 +1,5 @@
 //
-// Copyright IBM Corp. 2020, 2021
+// Copyright IBM Corp. 2020, 2022
```
andy-blum, line 2: thanks 😬

packages/styles/scss/components/pricing-table/_pricing-table.scss
```diff
@@ -0,0 +1,225 @@
...
+    &::after {
+      content: '';
+      display: block;
+      position: absolute;
+      right: 0;
+      bottom: 0;
+      width: 100%;
+      height: 1px;
+      background-color: $ui-03;
+      box-shadow: 50vw 0 $ui-03, 100vw 0 $ui-03, 150vw 0 $ui-03, 200vw 0 $ui-03,
```
andy-blum, line 52: Did you not find a way to get the sass function working?

packages/web-components/src/components/pricing-table/__stories__/pricing-table.stories.ts
```diff
@@ -0,0 +1,281 @@
...
+      ${renderHead(columnCount, heading)}
+      <dds-pricing-table-body>
+        ${renderBodyRow(columnCount, 1, CELL_TYPES.ICON, false)} ${renderBodyRow(columnCount, 2, CELL_TYPES.EMPTY, false)}
+        ${renderBodyRow(columnCount, 3, CELL_TYPES.TEXT, false)}
+      </dds-pricing-table-body>
+    </dds-pricing-table>
+  `;
+};
+
+export const WithSubheaders = ({ parameters }) => {
```
andy-blum, line 208: We need to title-case these variations in storyboo
Screen Shot 2022-04-05 at 5 46 11 PM
k

andy-blum (changes requested):
Seems like a great piece of work here, though I didn't stress test it too much. Thanks for backporting some of those style issues to the structured-list component. A few changes/questions below

packages/styles/scss/components/pricing-table/_pricing-table.scss
jkaeser: I couldn't figure it out quickly, so I shifted priority to getting this PR into a reviewable state. I'll look into it more today.

packages/web-components/src/components/pricing-table/__stories__/pricing-table.stories.ts
```diff
@@ -0,0 +1,287 @@
...
+        >
+        <dds-pricing-table-header-cell-description>
+          Lorem ipsum dolor sit amet consectetur.
+          <bx-unordered-list>
+            <bx-list-item>Lorem ipsum dolor</bx-list-item>
+            <bx-list-item>sit amet</bx-list-item>
+            <bx-list-item>consectetur retention adispiscing elit sed do eiusm Eiusmod tempor</bx-list-item>
+          </bx-unordered-list>
+        </dds-pricing-table-header-cell-description>
+        <dds-pricing-table-header-cell-cta href="https://www.carbondesignsystem.com/all-about-carbon/what-is-carbon/"
```
proeung, line 41: I know that @jwitkowski79's design spec doesn't include the CTA icon, however, since this component is extending from `DDSButtonCTA` and having seen the node structure of the CTA button in AEM, we'll need to add `cta-type` which will render different SVGs depending on the selection (see - https://carbon-design-system.github.io/carbon-for-ibm-dotcom/canary/web-components/?path=/story/components-cta--button). I think in this storybook preview, we can just add `cta-type="local"`.

proeung (changes requested):
@jkaeser Good work putting this component together! I see that there was a lot of work and attention to detail that has been put into this PR! I just have a couple of comments, but overall this is looking solid.

---------------------

@dratzenboeck-ibm @EAVilleda Can you help review this PR to ensure that we're not missing any dependencies (eg. slot styling, etc) needed on the AEM side?

packages/web-components/examples/codesandbox/components/pricing-table/cdn.html
```diff
@@ -0,0 +1,38 @@
...
+    }
+  </style>
+  <script type="module"
+          src="https://1.www.s81c.com/common/carbon-for-ibm-dotcom/tag/v1/latest/pricing-table.min.js"></script>
+</head>
+<body>
+<div class="bx--grid">
+  <div class="bx--row">
+    <div class="bx--col-sm-4 bx--col-lg-12">
+      <dds-pricing-table></dds-pricing-table>
```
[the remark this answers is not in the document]
jkaeser: Done!

packages/web-components/src/components/pricing-table/__stories__/pricing-table.stories.ts
jkaeser: Thank you for clarifying! I added that attribute to `<dds-pricing-table-header-cell-cta>` elements in the documentation and each of the examples.

dratzenboeck-ibm:
> @dratzenboeck-ibm @EAVilleda Can you help review this PR to ensure that we're not missing any dependencies (eg. slot styling, etc) needed on the AEM side?

I can't really tell what is needed in detail before starting with the AEM implementation. Regarding slots it would be generally helpful to apply css via the slot name instead of the tag name, so that we are able to wrap the tags with vanilla tags in the author view, as AEM doesn't apply its overlay to custom tags.

jwitkowski79:
@jkaeser Hi! This is looking great! Thanks for all of your hard work on this! Just did a round of design QA. I didn't get a chance yet to test on tablet and mobile but wanted to post what I have so far. I'm on vacation until Monday so hoping @oliviaflory  can step in and help finish the design QA on this component. 

1. Can we switch the info icon to the down chevron? My apologies. This never was updated in the design specs. It should be updated now in [box](https://ibm.ent.box.com/folder/157069618862?s=5qv9zo7jdm4pd7hc7lj5ogqy0czj50ie)
Components___Pricing_Table_-_Default_⋅_Storybook 2

2. Looks like we need to increase the bottom padding of cell data to 16px (see attached screenshot)
Components___Pricing_Table_-_With_subheaders_⋅_Storybook_🔊

3. Add 8px padding between text block and info icon
Components___Pricing_Table_-_Without_Row_Headers_⋅_Storybook (2)

4. I know this is a work in progress but is a sticky header still being implemented on scroll down? See [functional and design specs](https://ibm.ent.box.com/folder/157069618862?s=5qv9zo7jdm4pd7hc7lj5ogqy0czj50ie)

5. Is the variation of the Pricing table with content block still being worked on? Seems like it's missing as an option
component--structured lists-complex--block--specs--max

6. In the CTA, can we turn off the arrow icon by default? We wouldn’t recommend using it in our documentation and don’t want to have it on by default. Also, the arrows are behaving oddly. See screenshot of Safari and Chrome
Components___Pricing_Table_-_Default_⋅_Storybook (1) 2
Components___Pricing_Table_-_Default_⋅_Storybook 3

7. @oliviaflory would love your input on the below grid layout suggestions from @jkaeser to implement
    - 6-8 column table: columns should expand 2 columns
    - 4-5 column table: columns should expand 3 columns
    - 1-3 column table: columns should expand 4 columns

proeung:
@jkaeser Per my conversation (see [Slack thread](https://ibm-marketing-comms.slack.com/archives/G01LKRP5RB6/p1649266357514599?thread_ts=1649263741.454539&cid=G01LKRP5RB6)) with @ljcarot and @kennylam, we're going to mark this component behind a "Feature Flag" for now.

Can you add a commit that adds a new boolean and wraps the storybook component via a feature flag (see - https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/docs/feature-flags.md)? Also, let's call it `DDS_PRICING_TABLE`. See [Cloud Masthead](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/masthead/__stories__/cloud-masthead.stories.ts#L48) as an example.

RichKummer:
Took a look at `md` and `sm`

1. Jen mentioned in #6 that the button is behaving oddly. Just wanted to surface this in `md` when set to 2-columns: 
Screen Shot 2022-04-06 at 3 11 45 PM

2. In the Table heading, the last item in the unordered list has additional padding at the bottom, but this should be removed so that there is 24px padding between the last item and the CTA.
Screen Shot 2022-04-06 at 3 20 14 PM

3. If the tag in the Table heading wraps to a new line, the copy right before the unordered list should maintain a 16px padding between:
Screen Shot 2022-04-06 at 3 32 16 PM

4. When I tried to expand the row for more information on iPad in Safari, the text will appear instantly instead of fading in from the top like on Desktop. When I condense the row, the added content will fade up as expected. This issue only happens when the row is already at the appropriate height - if the row needs to expand the height, the animation will work as intended.

https://user-images.githubusercontent.com/15643582/162058675-4a552dbd-0319-4f95-83ad-e2ca5d0797d0.MP4

RichKummer:
Thanks for this @jkaeser ! Took a look at `md` and `sm`:

1. Jen mentioned in #6 that the button is behaving oddly. Just wanted to surface this in `md` when set to 2-columns: 
Screen Shot 2022-04-06 at 3 11 45 PM

2. In the Table heading, the last item in the unordered list has additional padding at the bottom, but this should be removed so that there is 24px padding between the last item and the CTA.
Screen Shot 2022-04-06 at 3 20 14 PM

3. If the tag in the Table heading wraps to a new line, the copy right before the unordered list should maintain a 16px padding between. Is that right @jwitkowski79 ?
Screen Shot 2022-04-06 at 3 32 16 PM

4. When I tried to expand the row for more information on iPad in Safari, the text will appear instantly instead of fading in from the top like on Desktop. When I condense the row, the added content will fade up as expected. This issue only happens when the row is already at the appropriate height - if the row needs to expand the height, the animation will work as intended.

https://user-images.githubusercontent.com/15643582/162058675-4a552dbd-0319-4f95-83ad-e2ca5d0797d0.MP4


5. When a row is expandable, but the expanded content _does not_ contain a checkmark icon, it looks like there is some additional padding at the top of the row text that should not be there. I didn't see an example in the specs, so wanted to double check that there should not be that added 16px spacing in this instance @jwitkowski79 

Screen Shot 2022-04-06 at 3 55 19 PM

jkaeser:
Thanks for the review, Jen, and sorry to bother on your vacation!

@oliviaflory @RichKummer I've addressed Jen's feedback except for the following:

4. Sticky headers are coming in a followup task. See https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8420.
5. Is this variation something that needs to be demonstrated in Storybook? As I understand it, this represents a composition of two components rather than a feature of the Pricing Table. cc @proeung
6. See Putra's comment: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/8674#discussion_r844000133. @proeung, does the AEM team _need_ this functionality? Currently, the Pricing Table CTAs will respect the `cta-type` attribute, but it sounds like maybe we should disable that in this component to adhere to design specs.

@RichKummer Thank you as well for your review! I've started working through your feedback, and so far I fixed the tag spacing (`#3`).

proeung (approved):
LGTM! Thanks for addressing my feedback!

----------

@kennylam @IgnacioBecerra Can one of you take a look at this feature flag PR? I'd like to get this merge down before the code freeze period. Thanks!

kennylam (approved):
LGTM thanks @jkaeser!

IgnacioBecerra (approved):
LGTM, the component looks pretty good, thanks @jkaeser!

kennylam:
@jkaeser Looks like this needs to get resolved before this check can pass https://github.com/carbon-design-system/carbon-for-ibm-dotcom/runs/5875018446?check_suite_focus=true#step:6:5905

RichKummer:
@jkaeser still seeing a few design issues:

1. The spacing at the bottom of the row should be 16px per Jen's comment - I'm seeing the bottom as 24px
Screen Shot 2022-04-08 at 4 29 07 PM

2. At the `md` breakpoint exactly (672px), the description on the second column moves up to align with the tag. Once I move the browser 1px larger, the text moves to the correct spot - aligning with the descriptive text in the first column.

**672px:**
Screen Shot 2022-04-08 at 4 31 31 PM

**673px:**
Screen Shot 2022-04-08 at 4 34 03 PM

3. The last item in the unordered list in the table header still has 4px bottom spacing - to match the specs, we'd need to remove that spacing so that there is 24px space between the last item and the button
Screen Shot 2022-04-08 at 4 36 49 PM

jkaeser:
Good catches @RichKummer! I just pushed some commits that should fix each of those issues with the exception of the list item spacing. That is baked into the `<bx-list-item>` element from Carbon Web Components, and as such I am unable to override its style. I _could_ adjust the margin above the CTA buttons to be 8px less, but that would cause spacing issues when the last element in the text area isn't a `<bx-list-item>` and doesn't set exactly 8px of bottom margin. How do you prefer I proceed?

RichKummer:
> Good catches @RichKummer! I just pushed some commits that should fix each of those issues with the exception of the list item spacing. That is baked into the `<bx-list-item>` element from Carbon Web Components, and as such I am unable to override its style. I _could_ adjust the margin above the CTA buttons to be 8px less, but that would cause spacing issues when the last element in the text area isn't a `<bx-list-item>` and doesn't set exactly 8px of bottom margin. How do you prefer I proceed?

I'll write a separate issue for this one, I'd rather not make an update here that causes other issues. Thanks @jkaeser, I'll take a look when the build finishes!

RichKummer (approved):

Issue #8658, opened 2022-03-29
Web Components: Video component not changing to custom thumbnail


guilhermealima:
### Description

When trying to add a custom thumbnail to an inline video using the `dds-video-player-container` component, the video kept pulling the default thumbnail image from MediaCenter, even using the `thumbnail-url` property as indicated by the [Carbon documentation](https://www.ibm.com/standards/carbon/web-components/?path=/docs/components-video-player--default).

### Component(s) impacted

`dds-video-player-container`

### Browser

Chrome

### Carbon for IBM.com version

1.16.1

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

https://www.ibm.com/resources/guides/hybrid-cloud-strategy/

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://codesandbox.io/s/video-thumbnail-bug-p7zyhc

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

oliviaflory:
This PR is hardcoding the Kaltura video thumbnail and needs to be undone to allow for custom thumbnails 
https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/7914/files#diff-c16cc0206fb8b5202bfa8ddd7900037a71fb8b30235347e4c33ece727f8d8860R94-R97

PR #8680, opened 2022-04-06
fix(video-player): ensure only automatic thumbnails get updated.

andy-blum:
### Related Ticket(s)

#8658 

### Description

Updates logic to only overwrite thumbnailUrl with a new copy from Kaltura if current value comes from a kaltura hostname and contains the current video's ID

### Changelog

**Changed**

- Fixed bug which prevented use of custom video thumbnails


ariellalgilmore (approved):
Screen Shot 2022-04-06 at 8 17 00 PM
Thanks for the updates!

Issue #8603, opened 2022-03-28
[Pictogram Item]: Allow for changing of Pictogram color


proeung:
### The problem

- In the recent brand expression campaign, there's a desire to showcase the pictograms/SVGs in Blue instead of the default black color.
- See - https://www.ibm.com/analytics/data-fabric/

### The solution

- Expose a `pictogram-color` color prop/attribute that allows Content Editors to pick and choose from the following options.
  - Black (Default)
  - Blue 50

### Application/website

AEM

### Business priority

Medium Priority = upcoming release but is not pressing

### Examples

See - https://www.ibm.com/analytics/data-fabric/

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

proeung:
@oliviaflory I don't have access to the design specs of this pictogram color change, however, I would assume that the color adjustment on dark/Gray 100 theme will follow the same pattern as `CTA Text` component (https://carbon-design-system.github.io/carbon-for-ibm-dotcom/canary/web-components/?path=/story/components-cta--text), where we're using token `$link-01`. Let me know if that's correct.

oliviaflory:
@proeung, yes we will use the `$link-01` token for the pictograms!

PR #8682, opened 2022-04-07
feat(pictogram-item): Allow for changing of Pictogram color

mwanberg:
### Related Ticket(s)

#8603: [Pictogram Item]: Allow for changing of Pictogram color

### Description

Expose a pictogram-color color prop/attribute that allows Content Editors to pick and choose from the following options.
- Black (Default)
- Blue 50

#### To test

1. Open the component at `/?path=/story/components-pictogram-item--default`
2. In the "Knobs" tab, the last knob should be "Pictogram color"
3. Select "Blue 50" and check that the pictogram SVG changes to blue

### DIFF
--- a/packages/web-components/src/components/pictogram-item/__stories__/pictogram-item.stories.ts
+++ b/packages/web-components/src/components/pictogram-item/__stories__/pictogram-item.stories.ts
@@ -48,6 +48,7 @@ const Pattern = html`
     focusable="false"
     preserveAspectRatio="xMidYMid meet"
     xmlns="http://www.w3.org/2000/svg"
+    fill="currentColor"
     data-autoid="dds--pictogram-item__pictogram"
     aria-label="Pictogram description"
     width="64"
@@ -92,6 +93,7 @@ const Touch = html`
     focusable="false"
     preserveAspectRatio="xMidYMid meet"
     xmlns="http://www.w3.org/2000/svg"
+    fill="currentColor"
     data-autoid="dds--pictogram-item__pictogram"
     aria-label="Pictogram description"
     width="64"
@@ -146,8 +148,8 @@ const pictograms = {
 };
 
 const pictogramColors = {
-  'Black (default)': COLOR_OPTIONS.DEFAULT,
-  'Blue 50': COLOR_OPTIONS.BLUE,
+  'Text color (default)': COLOR_OPTIONS.DEFAULT,
+  Blue: COLOR_OPTIONS.BLUE,
 };
 
 export const Default = ({ parameters }) => {
@@ -196,7 +198,7 @@ export default {
           src: selectPictogram(select('Pictogram (required)', pictograms, pictograms.Desktop, groupId)),
           'aria-label': textNullable('Aria-label:', 'Pictogram description', groupId),
         },
-        pictogramColor: select('Pictogram color:', pictogramColors, pictogramColors.black, groupId),
+        pictogramColor: select('Pictogram color:', pictogramColors, COLOR_OPTIONS.DEFAULT, groupId),
       }),
     },
     propsSet: {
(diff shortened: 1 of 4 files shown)


packages/web-components/src/components/pictogram-item/__stories__/pictogram-item.stories.ts
```diff
@@ -190,6 +196,7 @@ export default {
           src: selectPictogram(select('Pictogram (required)', pictograms, pictograms.Desktop, groupId)),
           'aria-label': textNullable('Aria-label:', 'Pictogram description', groupId),
         },
+        pictogramColor: select('Pictogram color:', pictogramColors, pictogramColors.black, groupId),
```
proeung, line 199: I'm getting a `error TS2339: Property 'black' does not exist on type '{ 'Black (default)': COLOR_OPTIONS; 'Blue 50': COLOR_OPTIONS; }'.`, which might be cause `pictogramColors.black` is not a define object value. See the code suggestion below.

Screen Shot 2022-04-07 at 12 12 20 PM


```suggestion
        pictogramColor: select('Pictogram color:', pictogramColors, pictogramColors.COLOR_OPTIONS.DEFAULT, groupId),
```

packages/web-components/src/components/pictogram-item/defs.ts
```diff
@@ -0,0 +1,23 @@
...
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * Pictogram colors.
+ */
+export enum COLOR_OPTIONS {
+  /**
+   * Defaults to black.
```
proeung, line 15: This is a bit minor, but I'm not sure if "black" is a good label, I think we can keep to `Gray 100` or `g10` to align with the Carbon theme. Anywho, I'll defer to @oliviaflory or @RichKummer on this.

RichKummer: I double checked and it looks like the Pictogram is using `text-01` to maintain readability across the themes, so we should call the default option `text-01` to align with that - otherwise "Black" would not make sense when it swaps to a lighter version on Gray 90/100.

RichKummer: Sorry, disregard the other comment - 

I don't think the "Black" works since the color token swaps to a lighter version on Gray 90/100. Naming the default option "Inverse" would clarify that and make it readable. We should probably change "Blue 50" to "Blue" since the the token used is slightly different based on Theme as well.

packages/web-components/src/components/pictogram-item/__stories__/pictogram-item.stories.ts
mwanberg: Good catch. `pictogramColors.COLOR_OPTIONS.DEFAULT` didn't work, but plain `COLOR_OPTIONS.DEFAULT` does, so I used that.

packages/web-components/src/components/pictogram-item/defs.ts
mwanberg: @proeung @RichKummer I changed the description to "text color" since "Inverse" doesn't make sense to me as the default on the lighter schemes. What do you think?

I removed the "50" from blue.

Screen Shot 2022-04-07 at 10 03 01 PM

Screen Shot 2022-04-07 at 10 03 52 PM

proeung (approved):
LGTM!

packages/web-components/src/components/pictogram-item/defs.ts
RichKummer: @mwanberg I noticed the WC build failed on the last commit, does this need to be rerun/merged?

mwanberg: @RichKummer I just merged in the latest from `main`, which has kicked off another build. Let's hope it finishes this time. 🤞🏻

kennylam (approved):
LGTM thanks @mwanberg!

RichKummer (approved):
Looks good, thanks @mwanberg !

Issue #8459, opened 2022-03-04
[a11y-axe-core]: React serious issues


IgnacioBecerra:
### Description

When removing true from the checkAxeA11y cypress custom command, I expect everything to pass: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/react/tests/e2e-storybook/cypress/support/commands.js#L105

<img width="895" alt="Screen Shot 2022-03-04 at 3 57 52 PM" src="https://user-images.githubusercontent.com/24970122/156857566-f1f0d30f-0018-4918-b540-6a7d33c65c80.png">

<img width="818" alt="Screen Shot 2022-03-04 at 3 58 30 PM" src="https://user-images.githubusercontent.com/24970122/156857596-b675851c-2581-4a6f-910f-7b6097000c83.png">

Once all pass change `includeImpacts` to include `moderate` and create a new issue with `moderate` bugs

### Component(s) impacted

Lightbox media viewer
Masthead w/search open on load

### Browser

Chrome

### Carbon for IBM.com version

canary

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

N/A

### Package

@carbon/ibmdotcom-react

### CodeSandbox example

N/A

### Steps to reproduce the issue (if applicable)

1. packages/react
2. change `checkAxeA11y` to remove `true`
3. run `yarn test:e2e-storybook:test:no-percy`

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #8689, opened 2022-04-07
fix(axe-core): update aria attribute errors

kennylam:
### Related Ticket(s)

#8459

### Description

This addresses various aria attributes that were either incorrect or missing that were causing accessibility end-to-end tests to fail.

All tests should now pass in `packages/react` when running
```
yarn test:e2e-storybook:test:no-percy
```

### Changelog

**New**

- add `components.json` to `.gitignore`

**Changed**

- `LightboxMediaViewer` containing element `aria-labelledby` changed to `aria-label`
- masthead search `suggestionsListContainer` is now labeled by the search input
- add `role="dialog"` to `VideoPlayer` container

**Removed**

- remove `true` from tests so they will properly fail when errors occur
- removes Cypress e2e Firefox tests for both React and web components


ariellalgilmore (approved):
LGTM! thanks!

Issue #8694, opened 2022-04-08
[Pricing table]: Subheader bottom spacing - Experimental


RichKummer:
**Engineering info:**  
- Innovation Team engineer: @jkaeser 
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2461
- DPO consulting engineer: N/A

-----

### Description

Follow on to https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/8674

When the Pricing table has subheaders, there is `24px` bottom padding currently. To match the specs, there should be `16px` bottom padding in the subheader rows.

**Current build:** 
<img width="1036" alt="Screen Shot 2022-04-08 at 5 56 42 PM" src="https://user-images.githubusercontent.com/15643582/162537546-42c9e6a0-3601-446e-9722-4ba05cafed21.png">

**Design specs**
<img width="812" alt="Screen Shot 2022-04-08 at 5 58 37 PM" src="https://user-images.githubusercontent.com/15643582/162537564-ac5dbb1b-10a6-4acc-916c-c108856b9583.png">

### Component(s) impacted

Pricing table - with subheaders

### Browser

Chrome

### Carbon for IBM.com version

v1.17.1

### Severity

Severity 4 = The problem is not visible to or noticeable to an average user. Affects minor functionality, no workaround needed.

### Application/website

C4IBM.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://codesandbox.io/s/github/carbon-design-system/carbon-for-ibm-dotcom/tree/master/packages/web-components/examples/codesandbox/components/pricing-table

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #8699, opened 2022-04-12
fix(pricing-table): use correct subheader bottom padding

jkaeser:
### Related Ticket(s)

Resolves https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8694

### Description

Updates the bottom padding applied to cells in the table groups to adhere to the design specs. Also fixes a few minor bugs from the original Pricing Table work that weren't caught in review.

### Changelog

**Changed**

- Updates bottom padding for `<th>` elements within `<dds-pricing-table-group>`s to adhere to design specs.
- Imports Pricing Table styles from the styles package rather than local path traversal.
- Replaces spacing rules with their logical property counterparts.
- Fixes closing tags for body row header cells in CodeSandbox examples.


proeung (approved):
LGTM from a code perspective. Thanks for fixing the closing tags issue in CodeSandbox and doc examples!

RichKummer (approved):
LGTM!

Issue #8717, opened 2022-04-18
Wrong Carbon Version


EAVilleda:
### Description

I’ve noticed something interesting in our AEM instance. If we check the version in the web console using window.digitalData.page.pageInfo.version, the console prints 'Carbon for [ibm.com](http://ibm.com/) v1.31.0'. I checked our package.json, and it’s
    "@carbon/ibmdotcom-styles": "1.32.0",
    "@carbon/ibmdotcom-web-components": "1.17.0",
If I change the values to 1.30.0 and 1.15.0 in the package.json and rebuild the node_modules, the version in web console correctly reflects 1.30.0.

Provided by @jeffchew, Production Storybook is also referencing version 1.31.0. 
![image](https://user-images.githubusercontent.com/47429622/163863744-d2c6667e-352a-423a-9ff3-867de6806535.png)

### Component(s) impacted

All components

### Carbon for IBM.com version

v1.32.0

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

AEM (Marketing)

### Package

@carbon/ibmdotcom-web-components, @carbon/ibmdotcom-styles

### CodeSandbox example

https://www.ibm.com/standards/carbon/web-components/?path=/story/*

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #8734, opened 2022-04-20
fix(vendor): fix to pull direct source for internal vendor packages

jeffchew:
### Related Ticket(s)

Refs https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8717

### Description

The internal vendor packages were pulling in the `node_modules` version, which was causing the issue of the version not getting properly reported in the `digitalData` object through the services package. The configs for vendor builds have been adjusted to now pull the source relatively within the project.

The version of `utilities` was also bumped to the proper version in the `services` packages, in hopes that lerna will now begin to properly bump versions going forward.

### Changelog

**Changed**

- Updated vendor build scripts to resolve via relative path in the monorepo
- Bumped utilities version in services to latest rc version


kennylam (approved):
Thanks @jeffchew!

PR #8737, opened 2022-04-21
fix(masthead): using better selector

IgnacioBecerra:
### Related Ticket(s)
None.

### Description
The React Canary tests were failing recently due to the fact that the previous selector that was using the `data-autoid` over the menu class was selecting the `Let's Create` menu that was implemented kind of recently. For some reason it was only failing in Chrome100 , Big Sur but it was okay in the rest of the browsers. Hopefully this PR makes it so it will always attempt to select the first menu that contains a submenu as the tests intended.

### Changelog

**New**

- {{new thing}}

**Changed**

-`data-autoid="dds--masthead-default__l0-nav0"]` to `.bx--header__submenu` to only specify the first menu that is selectable


ariellalgilmore (approved):
LGTM!

Issue #8139, opened 2022-01-25
[Tabs extended media]: React wrapper – not able to access other tabs with VO instructions


oliviaflory:
### Description

I am not able to select the second, third or fourth tab in Tabs extended or Tabs extended – with media using the keyboard and voice over on Desktop

VO instructs the user to press `control` + `option` + `space` to select a tab and it doesn't work, I see no changes in the selected state (all tabs content are the same).

VO is also reading out the tabs as first tab: "1 of 8"; second tab "3 of 8", which seems a bit strange? Is that the expected behavior and it's indicating a tab has two sections, the heading and the content?

https://user-images.githubusercontent.com/45692622/150992704-43131f18-e298-4f7a-8800-d9b976e8616a.mov

_This seems to be present in Production_ 

### Workaround
I am able to select other tabs by using: `control` + `option` + `shift` + `down arrow`, then `control` + `option` + `space`, then `tab` to navigate into the tab content area.

### Component(s) impacted

Tabs extended
Tabs extended with media

### Browser

Chrome

### Carbon for IBM.com version

1.15.0 rc 1

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

Carbon for IBM.com

### Package

@carbon/ibmdotcom-react

### CodeSandbox example

https://carbon-design-system.github.io/carbon-for-ibm-dotcom/next/web-components-react/?path=/story/components-tabs-extended--default

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

proeung:
@andy-blum Can you take a look at this issue and see if it's related to the one you just opened a PR for (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8357)? Feels like the VO issues that @oliviaflory mentioned here are the same as what you're working on, but let me know if that's not the case.

andy-blum:
The main thing I'd recommend to @oliviaflory (and anyone else testing VoiceOver) is to [do it in _safari_.](https://www.tempertemper.net/blog/getting-started-with-voiceover-on-macos#open-safari)

[...]

PR #8746, opened 2022-04-24
fix(tabs-extended): allow VO to access tab panel content

mwanberg:
### Related Ticket(s)

[[Tabs extended media]: React wrapper – not able to access other tabs with VO instructions](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8139)

### Description

The problems described in the ticket appeared to apply to not just the React wrapper but the original web components as well, so I addressed them as a whole.

Test the following in both the original web components and in the React wrapper:
1. Tabs extended
    1. Vertical
    2. Horizontal
2. Tabs extended - with media

Original issue in quotes:

> I am not able to select the second, third or fourth tab in Tabs extended or Tabs extended – with media using the keyboard and voice over on Desktop
>
> VO instructs the user to press control + option + space to select a tab and it doesn't work, I see no changes in the selected state (all tabs content are the same).

To aid in testing, I updated the stories to have different images in each of the tabs, so now when you click on or select a tab you can visually tell when it has changed.

Previously, when a tab was clicked, the keyboard/VO focus would remain on the tab. I changed the focus to go to the tab panel contents instead of the tab when the tab is clicked or selected by keyboard or in VO. This allows the user to access the tab panel content.

However, unfortunately, I was not able to find a way to get Chrome (or Firefox) to respect the VO `control + option + space`. There are differences between these browsers and Safari in how they interpret VO commands and keyboard presses. The VO commands all work as expected in Safari.

A workaround for Chrome and Firefox is to switch from using pure VO keyboard commands and just use the plain navigation arrows, which have already been mostly set up for use. Doing so, when you are on a tab and press `enter` or `space` or (if on the horizontal variation) the `down arrow`, you will be taken to the tab panel content.

> VO is also reading out the tabs as first tab: "1 of 8"; second tab "3 of 8", which seems a bit strange? Is that the expected behavior and it's indicating a tab has two sections, the heading and the content?

This does not appear to be an issue anymore, at least when in the Storybook previews. It is possible that when on a page with multiple tab sections that the screen reader might get confused, but I haven't tested that.

#### Followup

I think the VO and keyboard navigation might need some more followup and discussion to further refine the experience. For example:

1. When reaching the end of the tabs list, should it re-cycle through the tabs or continue into the selected content?
2. When reaching the end of the tab panel content, should it go back up to the tabs list or continue down the page?
3. If it goes up to the tabs list, should it go to its own tab first?
4. How should the keyboard navigation adjust to a vertical versus horizontal alignment (for `tabs-extended`)?

### DIFF
--- a/packages/web-components/src/components/tabs-extended-media/__stories__/tabs-extended-media.stories.react.tsx
+++ b/packages/web-components/src/components/tabs-extended-media/__stories__/tabs-extended-media.stories.react.tsx
@@ -1,7 +1,7 @@
 /**
  * @license
  *
- * Copyright IBM Corp. 2020, 2021
+ * Copyright IBM Corp. 2020, 2022
  *
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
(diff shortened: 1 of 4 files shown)


andy-blum:
Recommend closing this PR as the issue fixed isn't really an issue. See more information on [this comment on the issue](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8139#issuecomment-1108650344)

proeung:
@kennylam I'm going to close out this PR in the meantime, but feel free to reopen this if you see fit.

PR #8758, opened 2022-04-28
fix(e2e): adding wait times for React storybook e2e percy snapshots

jeffchew:
### Related Ticket(s)

No related issue

### Description

Noticing consistent false positives happening around snapshots in Percy that involve motion/height calculation. This adds some wait time to see if this helps at all.

### Changelog

**Changed**

- Added `cy.wait()` to various React e2e tests


kennylam (approved):
LGTM!

PR #8767, opened 2022-04-29
fix(tabs-extended & table-of-contents): ensure DOM updates are handled

### LINKED ISSUE #8751: [Table of Contents and Tabs Extended]: Does not recognize dynamic changes to DOM elements
**Engineering info:**  
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2466

Patch the fix to the v1.33.0 branch - https://github.com/carbon-design-system/carbon-for-ibm-dotcom/tree/release/v1.33.0

-----

### The problem

After modifying the DOM of dds-tabs-extended or dds-table-contents

The web component does not recognize the change made by Adobe Target or by any other DOM manipulation.

In the case of dds-tabs-extended, this causes the tabs to be grayed out and non-active.

In this case the Element was moved by editing the DOM in the developer tools and the content is moved but the tab is disabled.
https://www.ibm.com/about/secure-your-business

image

Please reach out to me the Marketing Target Team and @stsao

### The solution

The web components need to listen for changes after the initial rendering.

### Application/website

https://www.ibm.com

### Business priority

This is preventing paid media campaigns from going live that have intended traffic in April and May 2022

### Examples

Other issue page: https://www.ibm.com/analytics/data-fabric?at_preview_token=fKU34MaaYbqZigrAPUtFjg&at_preview_index=1_3&at_preview_listed_activities_only=true

### Code of Conduct

jkaeser:
### Related Ticket(s)

Resolves  #8751

#### DDSTabsExtended
There were cases where certain boolean attributes were being parsed as strings (e.g. `"false"`). Since non-empty strings are truthy, the value `"false"` was treated as `true`, and this caused any new tabs injected into the DOM to be  disabled.

This has been fixed by declaring to Lit Element that these attributes should be reflected as booleans. This tells Lit Element to convert the strings `"true"` and `"false"` to the booleans `true` and `false`, respectively.

#### DDSTableOfContents
The `slotchange` event that we typically use to catch updates to slotted DOM nodes has a limitation. From [the MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/slotchange_event):

> Note: the slotchange event doesn't fire if the children of a slotted node change — only if you change (e.g. add or delete) the actual nodes themselves.

This means the event will not fire if one of the immediate descendent DOM nodes are merely updated (e.g. have an attribute added/removed), nor will it fire if any of their children are added, removed, or modified.

Instead, we have to implement a MutationObserver to track changes to those nodes and rebuild the table of contents accordingly.

### Changelog

**Added**
- Added a MutationObserver to `DDSTableOfContents` components that rebuilds the table of contents when it sees changes to descendants of the default slot's children.

**Changed**

- Declare `boolean` type for `DDSTab`'s boolean reactive properties.

### DIFF
--- a/packages/web-components/src/components/table-of-contents/__stories__/content.ts
+++ b/packages/web-components/src/components/table-of-contents/__stories__/content.ts
@@ -22,7 +22,7 @@ export const headings = [
 export const LOREM = `Lorem ipsum dolor sit amet, consectetur adipiscing elit. Cras molestie condimentum consectetur. Nulla tristique lacinia elit, at elementum dui gravida non. Mauris et nisl semper, elementum quam non, lacinia purus. Vivamus aliquam vitae sapien volutpat efficitur. Curabitur sagittis neque facilisis magna posuere consectetur. Praesent fermentum sodales facilisis. Mauris a efficitur sem. Aliquam vehicula sapien libero, a viverra felis scelerisque vel. Vestibulum ante ipsum primis in faucibus orci luctus et ultrices posuere cubilia Curae; Donec fringilla dui tellus, a pretium diam vehicula et. Etiam non vulputate augue. Morbi laoreet diam dapibus sapien pellentesque tristique. Morbi id nibh metus. Integer non scelerisque nisl.`;
 
 const generateCopySections = n =>
-  [...Array(n)].map(
+  Array(n).map(
     () =>
       html`
         <p>${LOREM}</p>
--- a/packages/web-components/src/components/table-of-contents/table-of-contents.ts
+++ b/packages/web-components/src/components/table-of-contents/table-of-contents.ts
@@ -305,13 +305,25 @@ class DDSTableOfContents extends HostListenerMixin(StableSelectorMixin(LitElemen
   };
 
   /**
-   * Handles `slotchange` event on the default `<slot>`.
-   *
-   * @param event The event.
+   * Watches for changes to content in the default slot.
    */
-  private _handleSlotChange(event: Event) {
+  private _contentMutationObserver = new MutationObserver(this._contentObserverCallback.bind(this));
+
+  /**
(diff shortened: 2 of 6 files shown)


packages/web-components/src/components/table-of-contents/__stories__/content.ts
```diff
@@ -22,7 +22,7 @@ export const headings = [
 export const LOREM = `Lorem ipsum dolor sit amet, consectetur adipiscing elit. Cras molestie condimentum consectetur. Nulla tristique lacinia elit, at elementum dui gravida non. Mauris et nisl semper, elementum quam non, lacinia purus. Vivamus aliquam vitae sapien volutpat efficitur. Curabitur sagittis neque facilisis magna posuere consectetur. Praesent fermentum sodales facilisis. Mauris a efficitur sem. Aliquam vehicula sapien libero, a viverra felis scelerisque vel. Vestibulum ante ipsum primis in faucibus orci luctus et ultrices posuere cubilia Curae; Donec fringilla dui tellus, a pretium diam vehicula et. Etiam non vulputate augue. Morbi laoreet diam dapibus sapien pellentesque tristique. Morbi id nibh metus. Integer non scelerisque nisl.`;
 
 const generateCopySections = n =>
-  [...Array(n)].map(
```
jkaeser, line 25: This was redundantly spreading the contents of an array into a new array.

proeung:
@jkaeser Can you adjust the base branch of this new patch PR to https://github.com/carbon-design-system/carbon-for-ibm-dotcom/tree/release/v1.33.1?

cc: @kennylam

proeung (approved):
LGTM!

@kennylam Can I get a review from you as well? Thanks!

kennylam (approved):
Thanks @jkaeser!

Issue #8297, opened 2022-02-15
[CardStatic]: should not change background on hover


ariellalgilmore:
### Description

Static card has no link so should not be able to hover and see background change

![Screen Shot 2022-02-15 at 11 00 33 AM](https://user-images.githubusercontent.com/20210594/154164676-bc5cb0ff-157b-4f85-b551-ef13ea38d041.png)

### Component(s) impacted

CardStatic

### Browser

Chrome

### Carbon for IBM.com version

1.31.0-rc.1

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

carbon for ibm.com

### Package

@carbon/ibmdotcom-react

### CodeSandbox example

https://carbon-design-system.github.io/carbon-for-ibm-dotcom/next/react/?path=/story/components-card--card-static

### Steps to reproduce the issue (if applicable)

1. https://carbon-design-system.github.io/carbon-for-ibm-dotcom/next/react/?path=/story/components-card--card-static
2. hover over card

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #8781, opened 2022-05-02
fix(card): remove hover style for static card

IgnacioBecerra:
### Related Ticket(s)
#8297

### Description
The React Static Card hover styles were still present in the static variation, which shouldn't hover due to the lack of interaction it has. This PR ensures the background doesn't change color upon hover.

### Changelog

**New**

- added specific hover styles to ensure the background color stays the same in static variation


ariellalgilmore (approved):
LGTM!

Issue #8786, opened 2022-05-03
[Dependencies]: Potential vulnerability detected in `axios`


ajohar1:
Hello, I'm an IBM employee and we use Carbon for ibm.com for one of our internal applications. 

We have Whitesouce set up on our GitHub repo and it found a reported vulnerability with axios 0.21.4. This appears to be a dependency of one of the Carbon for ibm.com packages. Is this a known issue and if so, is there an update planned?

kennylam:
Hi @ajohar1, yes, expect `axios` to be updated soon, probably in the next release. We are slowly upgrading all our dependencies and that one is on the list.

PR #8816, opened 2022-05-10
chore(deps): upgrade `axios`

kennylam:
### Related Ticket(s)

#8786 

### Description

Upgrade `axios` to latest version

### Changelog

**Changed**

- upgrade `axios` to latest


ariellalgilmore:
LGTM! Should we also update in the utilities package? https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/3dda922731680e878ae116054bbe3303689c77ff/packages/utilities/package.json#L52

kennylam:
> LGTM! Should we also update in the utilities package?

Doh, good catch!

yarn.lock
```diff
@@ -5757,6 +5757,14 @@ axios@^0.21.1:
...
 
+axios@^0.27.2:
+  version "0.27.2"
+  resolved "https://registry.yarnpkg.com/axios/-/axios-0.27.2.tgz#207658cc8621606e586c85db4b41a750e756d972"
+  integrity sha512-t+yRIyySRTp/wua5xEr+z1q60QmLq8ABsS5O9Me1AsE5dfKqgnCFzwiCZZ/cGNd1lq4/7akDWMxdhVlucjmnOQ==
+  dependencies:
+    follow-redirects "^1.14.9"
+    form-data "^4.0.0"
+
 axios@~0.21.4:
```
ariellalgilmore, line 5768: sorry last thing: shouldn't axios `0.21.4` be removed now?

jeffchew: It looks like utilities is still using the old version:

image

kennylam: For some reason `services-store` is still using the incorrect `utilities` version (which was still using `axios@0.21.4`).  @jeffchew Didn't you push a fix for this recently?

kennylam: @jeffchew Looks to be unrelated to vendor – `services-store` correctly bumps the `services` version but not [`utilties`](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/services-store/package.json#L43). Not sure what's happening, will investigate further. In the meantime I've manually bumped the version to latest.

ariellalgilmore (approved):
LGTM!

kennylam:
> Digging into the `yarn.lock` file,

@jeffchew Yep, good catch! Upgraded to latest, but the sub-dependency is only on `0.25`, but at least the very old version we were using will go away.

PR #8817, opened 2022-05-10
feat(react-wrappers): content-block-media/mixed

ariellalgilmore:
### Related Ticket(s)

#8623 #8624

### Description

Add react wrappers for content-block-media and content-block-mixed. Also updated docs/readme's where for cta's for some were using the prop as `type="` instead of `cta-type="`

### Changelog

**New**

- readme, storybook, codesandbox content-block-media/mixed

**Changed**

- should be using card-link-cta - not card-link


IgnacioBecerra:
Everything looks good, but the heading isn't being updated through the Storybook knob.

This is happening due to the whole markdown feature within some of the text components, but this is how I'm doing it in the other stories to make sure it updates:
https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/8823/files#diff-dc08f86e31a62bc1c7c0fecded8365fd2cae86adc6b40f67924635a15e62b21dR39-R43

kennylam (approved):
LGTM thanks @ariellalgilmore! Just need to address Ignacio's comment.

PR #8867, opened 2022-05-18
fix: update thumbnails dot com

andreancardona:
Updates 2 thumbnails for the carbon platform: https://github.com/carbon-design-system/carbon-platform/issues/736


kennylam (approved):
Thanks @andreancardona!

Issue #8230, opened 2022-02-07
[Back to top]: needs to relocate keyboard focus


oliviaflory:
### Description

Back to top component should relocate the keyboard focus to the top of the page, currently the focus does not reset. Michael Gower said they set the tab index to -1 in Gatsby.

### Component(s) impacted

Back to top, Web components and react wrapper

### Browser

Chrome

### Carbon for IBM.com version

v1.15.0

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

Carbon for IBM.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://www.ibm.com/standards/carbon/web-components/?path=/story/components-back-to-top--default

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #8906, opened 2022-05-24
fix(back-to-top): reset tab order on click

emyarod:
### Related Ticket(s)

#8230

### Description

This PR resets the tab order after the back to top component is clicked. The first focusable element in the document should be focused after clicking the back to top component and the pressing <kbd>Tab</kbd>

https://user-images.githubusercontent.com/8265238/170094268-db4f27f6-a572-4df1-b530-3627121bc40b.mov


ariellalgilmore (approved):
LGTM!

Issue #8690, opened 2022-04-08
[Quote]: Corner bracket placement


RichKummer:
### Description

Seen on both Web Components and React builds:

On `max`, when the `Quote` component has a markType set to "cornerBracket", the opening corner bracket will be overlaying the top of the first word in the quote. 

Web Components:
<img width="1653" alt="Screen Shot 2022-04-06 at 10 17 25 AM" src="https://user-images.githubusercontent.com/15643582/162456422-95c76b51-30d5-49d0-8d13-f4a33051a875.png">

React:
<img width="1716" alt="Screen Shot 2022-04-06 at 10 20 00 AM" src="https://user-images.githubusercontent.com/15643582/162456303-090fdca8-13a6-437d-8eb3-e1ab0061fcc4.png">

Compare with xlg spacing:
<img width="1802" alt="Screen Shot 2022-04-06 at 10 20 30 AM" src="https://user-images.githubusercontent.com/15643582/162456633-742fd7d0-b705-4995-87b4-c66143c72faa.png">

This is also effecting `Callout quote`:
<img width="1008" alt="Screen Shot 2022-04-08 at 10 31 52 AM" src="https://user-images.githubusercontent.com/15643582/162457305-1fb817c0-1408-4659-bab5-fa188f1ef960.png">

### Component(s) impacted

Quote
Callout quote

### Browser

Chrome

### Carbon for IBM.com version

v1.17.1

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

C4IBM.com

### Package

@carbon/ibmdotcom-react, @carbon/ibmdotcom-web-components

### CodeSandbox example

https://codesandbox.io/s/github/carbon-design-system/carbon-for-ibm-dotcom/tree/main/packages/web-components/examples/codesandbox/components/quote

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #8907, opened 2022-05-24
fix(quote): update max breakpoint corner bracket position

emyarod:
### Related Ticket(s)

#8690 

### Description

This PR updates the Quote and Callout Quote corner bracket placement at the `max` breakpoint in both the `web-components` and `react` packages. The other quote variants should be unchanged at all breakpoints

image

image

### DIFF
--- a/packages/styles/scss/components/quote/_quote.scss
+++ b/packages/styles/scss/components/quote/_quote.scss
@@ -78,6 +78,10 @@
     @include carbon--breakpoint('lg') {
       left: -1rem;
     }
+
+    @include carbon--breakpoint('max') {
+      left: -2rem;
+    }
   }
 
   .#{$prefix}--quote__source {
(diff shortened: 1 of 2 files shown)


RichKummer (approved):
`Callout quote` and `Quote` both look good. Thanks @jkaeser !

packages/styles/scss/components/quote/_quote.scss
```diff
@@ -78,6 +78,10 @@
     @include carbon--breakpoint('lg') {
       left: -1rem;
     }
+
+    @include carbon--breakpoint('max') {
+      left: -2rem;
```
kennylam, line 83: @emyarod can we get spacing tokens in this file (wherever possible, I see a few fractions which I guess can be left that way for now).

kennylam (approved):
Thsnks @emyarod!

Issue #8839, opened 2022-05-16
[image]: merge with `image-with-caption`


kennylam:
These components should be merged, with `caption` and `lightbox` as options.

PR #8919, opened 2022-05-25
feat(image): merge image-with-caption with regular image

IgnacioBecerra:
### Related Ticket(s)
#8839

### Description
This PR merges both Image and Image with caption into one single component. The caption and lightbox options can be activated just as in the Image with caption. Also updated documentation to show that these two components have been merged. Also added a deprecation notice for whoever still uses the Image-with-caption component

### Changelog

**Changed**

- merged `image` and `image-with-caption` into one single component (`image`)
- `image-with-caption` now contains a deprecation notice

**Removed**

- `image-with-caption` stories for both WC and React wrapper


kennylam:
@IgnacioBecerra looks like a few component `e2e` tests need to be adjusted to pass the tests.

ariellalgilmore:
Awesome!!! looks like leadspace-block needs to get updated

Screen Shot 2022-05-26 at 8 10 17 AM

kennylam (approved):
LGTM!

PR #8937, opened 2022-06-01
chore(csb): fixes `content-block` CDN url

kennylam:
### Description

Fixes incorrect CDN url for `content-block` Codesandbox example

### Changelog

**Changed**

- fix CDN url


ariellalgilmore (approved):
LGTM! or do we just want to add the import here: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/b6f191530dea482bf07e5411648c7b20fe86d180/packages/web-components/src/components/content-block/index.ts#L15 `../cta/card-link`

PR #8940, opened 2022-06-01
feat(feature-cta): support Kaltura video thumbnails

### LINKED ISSUE #8887: [Feature Card/CTA Large]: Support Video Thumbnail
### The problem

We are currently using our own markup for Feature Card Large component. I attempted to update to the DDS markup for Feature CTA, but I noticed that the Feature CTA does not support video thumbnail.

This is an example screenshot of how we use the Feature Card Large with video:
AEM page: https://ibm.ent.box.com/s/j09224aicjsu0hra7eckd3pul5mz8w7b 
Drupal page (equivalent?): https://ibm.ent.box.com/s/fpss1tm90er4zr8sqxf8557vhyxeykrl

Here is the Feature CTA with video sandbox: https://codesandbox.io/s/determined-sanne-kp8n20?file=/index.html

### The solution

Solution would be to support video thumbnail, similar to how cards support it. 

### Application/website

AEM (Marketing)

### Business priority

Low Priority = release date is not dependent on fix or not upcoming

_No response_

### Examples

_No response_

### Code of Conduct

jkaeser:
### Related Ticket(s)

Resolves #8887

### Description

Adds support in `DDSFeatureCTA` elements for pulling Kaltura video thumbnails. Thumbnails are retrieved via the video ID set in the CTA's `href` and, like similar CTAs, can be suppressed with the `no-poster` attribute. Custom thumbnails can still be set by adding a `DDSImage` element to the `image` slot.

### Changelog

**New**

- `DDSFeatureCTA` elements can pull thumbnails from Kaltura based on video ID.


RichKummer:
`Card group` has an option to include a video instead of an image, and in that design the timing is displayed down by the CTA. It also has two play buttons including the icon, but for consistency I think that's fine. I would adapt the Card design here by moving the timing from the header to beside the link at the bottom right.

Screen Shot 2022-06-02 at 4 35 26 PM

jkaeser:
> Thanks for shifting the icon, @jkaeser ! The only thing I noticed was that on max only, the icon size jumps up to 32px x 32px. The icon should stay the same size as the other breakpoints.

@RichKummer I tweaked the styles so the icons are consistently `20px` wide & tall at all breakpoints. The incorrect sizing was introduced by the fix for https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6424, which indicates the icons should be `24px` at the `xlg` breakpoint. Has this changed?

jkaeser:
> We should drop the timestamp - it becomes inconsistent with other components when it's in the header or when it's with the CTA (and introduces odd sizing). When you get a chance, please revert the size change you made in the last update and remove the timestamp from the CTA.

@RichKummer Done! Any other changes needed here from a design perspective?

RichKummer:
@jkaeser I'm still seeing the timestamp in the header - if you can remove that we should be good on the design end! Percy noticed a few updates, so if you can also merge the updates I think this will be ready and I'll do a last review.

Screen Shot 2022-06-07 at 4 04 59 PM

proeung:
> @jkaeser I'm still seeing the timestamp in the header - if you can remove that we should be good on the design end! Percy noticed a few updates, so if you can also merge the updates I think this will be ready and I'll do a last review.

@RichKummer Thanks for reviewing this PR! My understanding from working with the Tech QA/Globalization team on production is that the timestamp/duration is required when the card or CTA is linking out to a video. From what I’ve seen so far within this library and other examples, the duration can be rendered next to the Icon or in the Card Heading. The screenshot above looks correct to me unless the duration should be placed next to the icon similar to the current treatment that we have in the `Card` component (see attached)? Or perhaps we just need to approve the change within the Percey visual diffs?

Screen Shot 2022-06-07 at 4 48 59 PM
 
cc: @oliviaflory @jwitkowski79

oliviaflory:
Hi all, I'm jumping in here, I think there's a few things that need to be resolved before moving forward so we aren't having @jkaeser make tons of updates that don't fit the guidelines.

> My understanding from working with the Tech QA/Globalization team on production is that the timestamp/duration is required when the card or CTA is linking out to a video. From what I’ve seen so far within this library and other examples, the duration can be rendered next to the Icon or in the Card Heading

@proeung thank for you for reminding us of the time stamp requirement, I had forgotten and suggested as a quick fix to allow this PR in.

Like you mentioned Putra, the other cards render the time stamp next to the CTA icon or the header. There are a few things I think we need to resolve first:

1. From @EAVilleda feature request screen shots, it looks to me like the they're using Feature card large, and it seems like @jkaeser is recommending the changes be made at the CTA feature level which only has the Feature card medium variant from what I can tell in Storybook. **Where is the appropriate place to add the video option?** cc @annawen1 (are you consulting on this issue, would love your input) 

2. > I tweaked the styles so the icons are consistently 20px wide & tall at all breakpoints. The incorrect sizing was introduced by the fix for https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6424, which indicates the icons should be 24px at the xlg breakpoint. Has this changed?

My understanding from that PR and the design specs are as follows:

| Breakpoint | Icon size |
| --- | --- |
| sm | 20px |
| md | 24px |
| lg | 24px |
| xlg | 24px |
| max | 32px | 

So the icon should not be consistently 20px at all breakpoints according to the specs.

3. Because the icon does change and is larger than all other card icon styles, we need design specs to identify what type size should be paired with the icon at each breakpoint. 16px text + 24px or 32px icon is not adhering to IDL guidelines.

- If AEM doesn't want the time stamp next to the CTA, is the header a good enough place?

4. > There's almost certainly some design updates we'll need to consider, and I'd really like to your opinions, @RichKummer and @jwitkowski79. You'll see there are currently two play icons when the Kaltura video thumbnail is used: one over the thumbnail, and one as the card icon in the lower right corner. I left the one over the thumbnail for consistency with other CTAs that have similar functionality, but perhaps the easiest solution is to remove it.

John K brings up a good point that we need design specs and guidance on whether the video thumbnail should have the video overlay, we have that in all our other cards so I'd lean towards yes. 

---
Essentially, I think we need some design specs for this before moving forward. @RichKummer and I do not have the bandwidth to address the specs during this sprint. cc: @jwitkowski79 @EAVilleda is this something your team has the bandwidth to look at?

proeung:
@oliviaflory Thank you for your response. It sounds to me like we're on the same page with the display of the timestamp/duration next to the CTA icon or the header, which is great!

The example that @EAVilleda mentioned in her ticket (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8887) is using `<dds-feature-cta>` component which allows for different CTA type options that AEM will now be adopting instead of [Feature Card Large](https://carbon-design-system.github.io/carbon-for-ibm-dotcom/canary/web-components/?path=/story/components-feature-card--large) component, which we realized is limiting when it comes to having the ability to display different CTA types. With that being said, I do agree that we need to ensure the icon in `<dds-feature-cta>` follows the specs defined for the Feature Card (https://ibm.ent.box.com/file/789229799989?s=wys6fs8f4x716fvzj2ow3vsd65dagra9).

**Here's my proposal in an effort to keep this work moving forward and give @jkaeser clarity on what to adjust.**

- Keep the video stamp time/duration in the card heading as what we have currently (see attached) and approve the diffs in Percy, so that the changes in this PR can be merged down.
 - Screen Shot 2022-06-09 at 1 50 50 PM
- Design (@RichKummer or @jwitkowski79) updates the "Feature Card" design to include the "Play" button on top of the image and other changes - https://ibm.ent.box.com/file/789229799989?s=wys6fs8f4x716fvzj2ow3vsd65dagra9
- Follow up with a new issue/ticket to update the Feature CTA Storybook preview (https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/8940/index.html?path=/story/components-cta--feature) to include an option for the "Large" and "Media" variants.

proeung (approved):
LGTM!

@ariellalgilmore or @IgnacioBecerra Can I get a code review from you?

IgnacioBecerra (approved):
LGTM, thanks @jkaeser!

RichKummer:
It looks like the icon is staying `20 x 20px` at all breakpoints. Is this a fix we could get in before merging? If not I can create a separate ticket.
Screen Shot 2022-06-16 at 7 52 25 PM

Percy's result for React-e2e are giving an build failed error:
Screen Shot 2022-06-16 at 7 47 04 PM

jkaeser:
@RichKummer I reverted all changes I made to the icon sizing in this PR. At this point, any incorrect sizing you're seeing is pre-existing and should be addressed in a separate ticket.

I merged the latest from `main`, and that seems to have cleared up the react e2e test errors. The Percy changes seem unrelated.

RichKummer (approved):
@proeung @jkaeser - cleared out the Percy errors and did one last check, we're good to add `Ready to merge`

Issue #8660, opened 2022-03-29
[footer]: update documentation


ariellalgilmore:
### Description

When trying to create a footer with language only based on the documentation [here](https://www.ibm.com/standards/carbon/web-components/?path=/docs/components-footer--default-language-only), I recieve errors.

can also reference: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/footer/combo-box.ts#L261

### Component(s) impacted

footer

### Carbon for IBM.com version

canary

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

dotcom

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://www.ibm.com/standards/carbon/web-components/?path=/docs/components-footer--default-language-only

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #8945, opened 2022-06-02
fix(footer): update docs and add language selector csb

ariellalgilmore:
### Related Ticket(s)

#8660 

### Description

Update documentation for correct implement of footer with language selector and add a codesandbox example


kennylam (approved):
Thanks @ariellalgilmore!

PR #8948, opened 2022-06-03
fix(cta): reorder csb styles

ariellalgilmore:
### Related Ticket(s)

none
### Description

range error happening when styles are in a different order


kennylam (approved):
LGTM!

PR #8956, opened 2022-06-06
fix(cloud-masthead): replace cookie auth with personalization status

proeung:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8033

### Description

- It was brought to our attention that the Cloud console team has completed a feature to detect and personalize based on if a person was authenticated/logged-on on via the `cloud.ibm.com` console.
- This detection method was based on a request to personalize `/cloud` pages with users' live login state, which can now be accomplished.
- Instead of using the cookie checking method to verify if a user has logged into `cloud.ibm.com`, we're now using the status object coming from the client-side.

### Changelog

**Changed**

- Replace the `checkCookie()` method with `checkPersonalization()`.

### Testing
- Waiting for a CDN preview to be generated and tested on Drupal and other applications.


proeung:
Testing for this PR is blocked until Tugboat previews are back up and running.

proeung:
Closing this PR out for now and will revisit it later once we figured out why the DDO login status is not outputting consistently.

PR #8970, opened 2022-06-09
chore(storybook): clean up `tabs-extended-media` story

kennylam:
### Description

Currently all content within the tabs for this component is identical, making it seem as if perhaps the tab-switching functionality isn't working. Content is now updated reflecting the current tab chosen, as well as removing various knobs that probably shouldn't be available in this story.

The only knobs remaining are those directly related to the properties of the component itself. All copy, href, CTA text, etc, have been removed and can be edited in real-time using the Codesandbox examples.

### Changelog

**Changed**

- update tab content to reflect currently selected tab for web components and react wrapper stories

**Removed**

- unnecessary knobs


ariellalgilmore (approved):
LGTM! the clean up is so much better

PR #8988, opened 2022-06-13
chore: remove sass `/` division deprecation warnings

emyarod:
### Description

This PR changes the deprecated `/` division operator to `calc` in our Sass files. This cleans up the deprecation warning in our test scripts

### Changelog

**Changed**

- `/` CSS division to `calc()`


emyarod:
this can only be removed after migrating off of `node-sass`!

Issue #8962, opened 2022-06-08
[Structured List & Pricing Table]: AEM currently returns `<p>` tag in the list cell causing style rendering issue


proeung:
**Engineering info:**  
- Innovation Team engineer: @proeung 
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2487
- DPO consulting engineer: N/A

-----

### The problem

- The AEM team is in the midst of adopting the Structure List and Pricing Table components and has recently noticed some discrepancies related to the rendering of the text within the structured list and pricing table cells.
- Currently, the team is using AEM's OOTB WYSIWYG to populate data within the text cell area as this content can be a regular paragraph or includes `<ul>`, which results in `<p>` tag that gets wrapped around the text.
- The text in the cell should be using `productive-body-long-01` and color `$text-02` instead of the `<p>` style from the Carbon reset helper stylesheet.

![Screen Shot 2022-06-08 at 09 06 00](https://user-images.githubusercontent.com/1815714/172673429-2d1c2239-4bf9-4a32-bd68-b508263aad25.png)

### The solution

- Figure out a way to import styling needed for the WYSIWYG area that includes proper formatting for `<p>`, `<ul>`, etc as there's no great way to target/force style within the shadow root (eg. `:host(dds-structured-list-cell) ::slotted(p)`).
- This is the same issue that we're currently running into the Accordion version that's managed in Carbon for AEM (https://github.ibm.com/webstandards/carbon-for-aem/pull/40/files#r6224256). @kennylam @annawen1 Feels like this can be something we can bring to the dev office hour call.

### Application/website

AEM

### Business priority

Medium Priority = upcoming release but is not pressing

### Examples

- https://codesandbox.io/s/jovial-hugle-vpqkqg?file=/index.html

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

proeung:
@kennylam During our last issue-sharing meeting, I believe you mentioned that you have a solution to follow up on regarding this issue. Can you let me know how we can approach this aside from the rejected solution in this PR (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/8993)? Thanks!

[...]

PR #8993, opened 2022-06-13
fix(structured-list & pricing-table): allow for rich text formatting

proeung:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8962

### Description
- This PR adds a `handleSlotChange` in the render function that takes the content of the `<dds-structured-list-cell>`  and `<dds-pricing-table-header-cell-description>` components and appends it to the shadow root for flexibility styling using a class name.
- This allows adopters like AEM to add HTML tags that are used within the WYSIWYG (eg. `<p>`,`<ul>`, `<ol>`) and ones that are defined in the design specs.

### Changelog

**New**

- Handles `mutationObserver` event within the `<dds-structured-list-cell>`  and `<dds-pricing-table-header-cell-description>` components.

**Changed**

- `<dds-pricing-table-cell-annotation>` now extends from `<dds-pricing-table-header-cell-description>`
- Update stories to show examples of rich text format vs custom elements from Carbon Core's `<<bx-unordered-list>`

Screen Shot 2022-06-13 at 4 32 43 PM
Screen Shot 2022-06-13 at 4 32 01 PM

### Testing Instructure
- Visit the "Structure List" component - `?path=/story/components-structured-list--default`
- Ensure that the `<p>` and `<ul>` tags are rendering correctly and follows the design specs.
- Visit the "Pricing Table" component - `?path=/story/components-pricing-table--default`
- Ensure that the `<p>` and `<ul>` tags are rendering correctly and follows the design specs.


RichKummer:
Hey @proeung - the rich text looks like it's rendering well for both. I'm not sure if this is related, but I noticed that there should be a `16px` space between the paragraph and the unordered list in the highlighted/feature section at the top of the pricing table, according to the specs. It looks like this spacing appears properly in the structured list but not in the featured area:

**Build**
Screen Shot 2022-06-15 at 10 35 35 AM

**Specs**
Screen Shot 2022-06-15 at 10 38 29 AM

Again, not sure if this is related to the rich text - @jwitkowski79 might know more when she takes a look!

proeung:
@RichKummer Great point! Also, just so you know, the current upstream version of the Pricing Table description cell doesn't have spacing in between the `<p>` and `<ul>` (see - https://carbon-design-system.github.io/carbon-for-ibm-dotcom/canary/web-components-experimental/?path=/story/components-pricing-table--default), however, if this is something that we missed during the review process, I'm happy to add it to this PR. @jwitkowski79 can you confirm?

Screen Shot 2022-06-15 at 10 53 06 AM

jkaeser (changes requested):
@proeung This is a very clever approach, and the end result looks great!

There's an opportunity here to reduce redundant code by merging the new and identical behaviors in `DDSPricingTableHeaderCellDescription` and `DDSStructuredListCell` into a base class like `DDSStructuredListRichText`. We'd extend from this class in the three places we need the same behavior:
1. `DDSStructuredListCell`
2. `DDSPricingTableHeaderCellDescription`
3. `DDSPricingTableCellAnnotation`

I believe you'd also be able to write the styles just once in the Structured List stylesheet if you used a more generic class name for the wrapper (i.e. `.bx--structured-list-rich-text` instead of `.bx--structured-list-cell-body` and `.bx--pricing-table-header-cell-description-body`). This avoids some messes, such as the annotations receiving a `.bx--pricing-table-header-cell-description-body` class, which at best is confusing and at worst would cause future styles meant only for the header cell descriptions to bleed into the annotations.

Overall I think this would be much more maintainable. What do you think?

proeung:
@jwitkowski79 Thanks for confirming the spacing change. I just pushed up a fix and a new WC deploy preview should be ready for your review soon.

-----------

> There's an opportunity here to reduce redundant code by merging the new and identical behaviors in DDSPricingTableHeaderCellDescription and DDSStructuredListCell into a base class like DDSStructuredListRichText. We'd extend from this class in the three places we need the same behavior:
> 
> DDSStructuredListCell
> DDSPricingTableHeaderCellDescription
> DDSPricingTableCellAnnotation
> 
> I believe you'd also be able to write the styles just once in the Structured List stylesheet if you used a more generic class name for the wrapper (i.e. .bx--structured-list-rich-text instead of .bx--structured-list-cell-body and .bx--pricing-table-header-cell-description-body). This avoids some messes, such as the annotations receiving a .bx--pricing-table-header-cell-description-body class, which at best is confusing and at worst would cause future styles meant only for the header cell descriptions to bleed into the annotations.

@jkaeser Thanks for the review! When working on this PR, I contemplated the same modular component approach that you suggested but realized that there's some discussion that we need to follow up with the other DPO's designers and engineers about this. Ideally, I like to add a new component called `<dds-rich-text>` that's independent of Structured List, where this component can be used and extended in the various components where rich text is enabled in AEM. This way this component is part of the system and its styles are managed in the styles package that can be imported/extended in other components. I'll create a feature request ticket for this idea, but I think we can leave the changes in this PR as it is for now so that we're not expanding the SOW.

cc: @kennylam

jkaeser:
>  Ideally, I like to add a new component called `<dds-rich-text>` that's independent of Structured List, where this component can be used and extended in the various components where rich text is enabled in AEM. This way this component is part of the system and its styles are managed in the styles package that can be imported/extended in other components. I'll create a feature request ticket for this idea, but I think we can leave the changes in this PR as it is for now so that we're not expanding the SOW.

@proeung That's a great idea!

I do still think it's worth making the changes I suggested within the scope of this PR. Doing so would make the eventual migration of this functionality to an independent component more straightforward, and it would make sure this implementation is maintainable in the meantime. That being said, I won't hold up this PR from merging if you feel strongly that the effort is not worth it!

proeung:
@kennylam or @jeffchew Any thought about the solution introduced in this PR?

jeffchew:
Hi @proeung , sorry for the delay in response!

We spoke as a team and agreed that this probably should be in `Carbon for AEM` instead. Let us know if you need any assistance with extending anything there from `Carbon for IBM.com`!

proeung:
@jeffchew Thanks for following up on this, however, can you give a reason behind the decision? Both Structured List and Pricing Table components are currently managed in this repo (C4IBM) and for us to expose the styles for the few tags defined in the AEM WYSIWYG (eg. `<p>`,`<ul>`, `<ol>`) in C4AEM, we'll need to extend these two components, which is a big overhead in my opinion.

We can look into adding a separate rich-text component (eg. `caem-rich-text`) in C4AEM, but it's not a guarantee that everything will translate/ensemble well with the other existing `dds-structured-list` and `dds-pricing-table` used in AEM.

Anywho, this feel like a discussion that we might need to have with the larger group especially with the leadership over on the AEM side. I'll set up a time to loop @ljcarot @kjwhitla and @jwdomb as well.

kennylam:
@proeung We don't feel these changes belong in our codebase. This is an effort to fix something that is happening in the AEM authoring environment. If these AEM-specific styles are being added to address something that does not exist in the component itself, then that is where they should be added. Otherwise, the fix should come from the AEM side. Either way it seems as if this work is being done in reverse.

Happy to help continue looking for a solution, but I'm not sure having this discussion with a larger outside group or leadership will change anything? This is not what dictates what makes it into our codebase.

proeung:
Closing this PR out in favor of an alternative solution coming from @kennylam in this issue (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8962).

PR #9004, opened 2022-06-15
fix(lightbox-with-carousel): fix focus wrapping logic & carousel focus movement

### LINKED ISSUE #8994: [Light box media viewer]: QA: Keyboard focus is not navigating beyond 'Close' button in reverse navigation
**Engineering info:**  
- Innovation Team engineer:
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2488

-----

### Description

Page Element:
Close button

Issue Description:
Focus is not moving to the previous element after 'Close' button. It stays in the same element.

Expected:
On pressing Shift + tab key, focus should go to the previous element in the page.

User Impact:
When standard methods of keyboard navigation is not supported, this impact the user ability to interact with the page controls. This impacts Keyboard only users/Screen reader users

### Component(s) impacted

Light box media viewer

### Browser

Firefox

v1.35.0-rc.0

### Severity

### Application/website

Carbon for IBM.com

### Package

### CodeSandbox example

https://carbon-design-system.github.io/carbon-for-ibm-dotcom/next/web-components

_No response_

_No response_

### Code of Conduct

andy-blum:
### Related Ticket(s)

Fixes #8994 

### Description

This PR addresses two accessibility problems.

## Fixing focus wrapping in modals with complex content

This PR refactors the focus-wrapping of the DDSExpressiveModal. Prior to this PR, the expressive modal uses the internal [_focusableElements](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/expressive-modal/expressive-modal.ts#L129) property to hold onto a reference of all focusable elements so that we can wrap focus to the first or last focusable element within the modal. This property is [populated when the component is first updated](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/expressive-modal/expressive-modal.ts#L414-L417) with all descendants and shadowdom elements that match the `selectorTabbable` string

```js
this._focusableElements = [
  ...Array.from((this.shadowRoot?.querySelectorAll(selectorCloseButton) as NodeListOf<HTMLElement>) || []),
  ...Array.from(this.querySelectorAll(selectorTabbableForModal) as NodeListOf<HTMLElement>),
];
```
The problem with this approach, is that it's only querying the slotted elements of the modal and the modal's own shadowroot. Any focusable elements within slotted shadowroots aren't accounted for, and thus aren't a part of the modal's focus wrapping mechanism. This PR addresses that by changing the private, internal `_focusableElements` property into a getter method, and adding a public `focusableElements` getter method to the modal and any component that might be placed within a modal (right now it's just the carousel).

In this way, each component is responsible for keeping track of it's own focusable elements, and the modal is responsible for asking for those when it needs them. When the modal deems it necessary to wrap focus, it asks the elements it knows about to report their focusable elements, removes any duplicates, sorts the array of items, and then passes that polished list to the `tryFocusElems()` function. This approach _does_ cause re-querying each time the getter is run, but that's necessary for the carousel as the navigation buttons get disabled based on the status of the carousel.

## Fixing focus handling on carousel items that are off-screen

This PR adds some logic to handle focus movements in carousels. Prior to this fix, focusable elements in carousel slides off-screen could still gain focus and move the carousel slides. This creates a difference in how the page is percieved between sighted mouse users and visually-impaired keyboard users. Now, when the carousel's `start` property changes and slides are moved, the slides that are off-screen are given the [`inert`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert) attribute. This attribute is a boolean value that, when present, makes the browser "ignore" user input events for the element, including focus events and events from assistive technologies. This prevents focus from moving to slides that are off-screen, keeping parity between mouse and keyboard navigation. 

Right now, the inert fix would only impact a subset of users, as it's not fully implemented across all browsers, but it's a progressive enhancement feature that won't impact users that _don't_ have support for it at all. If desired, there's also a [polyfill](https://github.com/WICG/inert) that could bring inert functionality to most, if not all, browsers we currently support

### Changelog

**Changed**

- Updated focus-wrap logic within DDSExpressiveModal
- Updated carousel with inert attribute on hidden carousel items
- Added focusableItems getter to DDSCarousel for DDSExpressiveModal compatibility


RichKummer (approved):
Glad this got fixed so quickly! Thanks @andy-blum !

andy-blum:
@IgnacioBecerra @kennylam Do you have any thoughts on adding the [polyfill for `[inert]`](https://github.com/WICG/inert)?

andy-blum:
@kennylam `[aria-hidden]` hides the content from assistive tech, but `[inert]` also prevents focus & other user interactions on the element & its subtree.

Here's a codepen demo: https://codepen.io/andy-blum/pen/QWQoeqp

kennylam:
@andy-blum Right, that part I get. I was more asking what the functional difference would be apart from disabling focus with `inert`. In that case let's add the polyfill, sounds like a good solution.

proeung:
@kennylam Can this polyfill be added in a separate issue or would you like us to add it as part of this PR?

kennylam:
> Can this polyfill be added in a separate issue or would you like us to add it as part of this PR?

@proeung The only major browser which doesn't support this yet is FF (only supported in nightlies atm). I think we should probably include it now since that's a tier 1 browser. What do you think?

proeung:
@kennylam I completely agree! Tier 1 browser support is a top priority. @andy-blum Let's move forward with adding the polyfill in this PR.

Issue #9013, opened 2022-06-17
Required on <DDSTab> by default highlight for any Tab field !


reshmpk1:
### The problem

In this DDSTab component, the first Tab value will be highlighted by default.
But we wanted to make this Tab selection after redirecting from another page, which the ‘Third Tab’ should be highlight. So I wrote this script to make that third tab active and the first tab inactive.
but this scripting is not working. Need to provide the control from the component side. 

### The solution

```
document.getElementById("tab-link-0-default").setAttribute("aria-selected", "false");
document.getElementById("tab-link-3-default").setAttribute("aria-selected", "true");

```

### Application/website

https://www.ibm.com/privacy/portal/us-en

### Business priority

High Priority = pressing release

### Examples

We are looking as below given:
![image](https://user-images.githubusercontent.com/29655795/174257154-65a35cbd-bd85-4cc2-ada8-91490482fa9b.png)

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #9029, opened 2022-06-21
feat(tabs): introducing optional different default tab on load

IgnacioBecerra:
### Related Ticket(s)
#9013

### Description
This PR introduces an attribute to `DDSTabsExtended`, `active-tab` which takes in the index of the desired tab to be selected. With this feature, the component will be able to set an active default tab different than the first one.

### Changelog

**New**

- `DDSTabsExtended` now has a `active-tab` attribute to select the desired tab on load


ariellalgilmore (approved):
LGTM! Documentation looks great too!

Issue #9044, opened 2022-06-28
Why TypeScript as peer dependency?


abdonrd:
### Summary

Why TypeScript as peer dependency?

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/c2e99ca87ef13274577df9ae7b861a0e47f6f82d/packages/web-components/package.json#L265-L267

We have an issue with that:
<img width="609" alt="Screen Shot 2022-06-28 at 18 17 47" src="https://user-images.githubusercontent.com/1007051/176239762-e18f29c2-516d-4270-96a0-24051a1fc494.png">

### Name and organization

IBM

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

abdonrd:
Maybe there is something I don't know, but just in case I send the PR:

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/9045

kennylam:
Hi @abdonrd, it would be for the exact reason you are showing in your terminal output – to alert users that a specific version of Typescript is needed to use this package at runtime. It looks like you are using a different version in your project?

abdonrd:
> Hi @abdonrd, it would be for the exact reason you are showing in your terminal output – to alert users that a specific version of Typescript is needed to use this package at runtime. It looks like you are using a different version in your project?

Mmm but why? TypeScript it's only used for build the package before publishing to npm.
When you download the npm package, you only use the JavaScript files.

abdonrd:
From another conversation:

To import a component such as:
[@carbon/ibmdotcom-web-components/es/components/button/index.js](https://unpkg.com/browse/@carbon/ibmdotcom-web-components@1.21.0/es/components/button/index.js)
[@carbon/ibmdotcom-web-components/es/components/button/button.js](https://unpkg.com/browse/@carbon/ibmdotcom-web-components@1.21.0/es/components/button/button.js)

You don’t need TypeScript... it’s a simple JavaScript file.

PR #9045, opened 2022-06-28
Remove TS as peer dependency

abdonrd:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/9044

### Description

Remove TS as peer dependency
Maybe there is something I don't know, but just in case I send the PR

### Changelog

**Removed**

- Remove TS as peer dependency

### DIFF
--- a/packages/web-components/package.json
+++ b/packages/web-components/package.json
@@ -263,9 +263,6 @@
     "web-component-analyzer": "^1.0.0",
     "webpack": "^4.46.0"
   },
-  "peerDependencies": {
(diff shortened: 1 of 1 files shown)


annawen1:
Closing this PR as the ticket has been closed.

From the slack conversation, the peer dependency is in line with all other carbon design system packages and is intended. The package can be installed using the `--legacy-peer-deps` flag or as you mentioned npm's [overrides](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides) field.

Issue #9065, opened 2022-07-07
[masthead]: megamenu items not displaying correctly


kennylam:
### Description

Web components `masthead` megamenu items are all stacking in one large column.

![image](https://user-images.githubusercontent.com/909118/177762279-0332c82e-f8fe-41b0-9480-6093adadef1f.png)

### Component(s) impacted

`masthead`

### Browser

Chrome

### Carbon for IBM.com version

v1.35.0

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

Carbon for IBM.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://www.ibm.com/standards/carbon/react/?path=/story/components-masthead--default

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

annawen1:
Fix has been merged in previous release

PR #9066, opened 2022-07-07
fix(masthead): fixes megamenu item styles

kennylam:
### Related Ticket(s)

#9065 

### Description

Megamenu child items are being displayed in one, long, single column. This fixes that.

### Changelog

**Removed**

- `.bx--masthead__megamenu__category-group-shield` as child to `::slotted` selector


annawen1 (approved):
Alright, the masthead and cloud-masthead look good!

PR #9078, opened 2022-07-08
fix(pricing-table): enable video CTA in pricing table header cell

### LINKED ISSUE #9050: [Pricing-Table]: Allow for flexible placement of slot CTA so that <dds-video-cta-container> can be used
**Engineering info:**  
- Innovation Team engineer: @andy-blum 
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2502

-----

### Description

- The AEM team is currently integrating the `Pricing Table` component into the CMS and noticed that when wrapping the using the `<dds-video-cta-container>` component to get the video lightbox media viewer, the `<dds-pricing-table-header-cell-cta>`.
- This is happening because the `slot="cta"` is being loaded to the `<dds-pricing-table-header-cell-cta>` and not on the parent level when the CTA type is equal to `video`.

From AEM dev

> The tag does allow us to use "video" as a "cta-type" attribute and the icon will appear in button. However, no action will occur when clicking on the button. I did try wrapping the around the tag. But, the button will not display because the dds-video-cta-container only has a slot in its shadow root and dds-pricing-table-header-cell-cta has "slot='cta'" automatically being applied to it. See screenshot below:

**See PR thread:** https://github.ibm.com/CHQ-Marketing-Products-and-Platforms/adobe-cms/pull/1368#issuecomment-45628527

Screen Shot 2022-06-29 at 9 58 58 AM

### Component(s) impacted

Pricing-Table - `<dds-pricing-table-header-cell-cta>`

### Browser

Chrome, Safari, Firefox, Microsoft Edge

v1.35.0

### Severity

### Application/website

AEM

### Package

### CodeSandbox example

https://codesandbox.io/s/lucid-satoshi-sqvbri?file=/index.html:3986-4304

_No response_

_No response_

### Code of Conduct

andy-blum:
### Related Ticket(s)

Fixes #9050

### Description

Adds an internal property of `ctaType` to the pricing-table-header-cell. When this value is set to 'video' by a slotchange event, the slot is wrapped in a `<dds-video-cta-container>` element to allow the popup video player.

### Changelog

**Changed**

- fixes an issue where `video` CTAs didn't open popup player

### DIFF
--- a/packages/web-components/src/components/pricing-table/pricing-table-header-cell.ts
+++ b/packages/web-components/src/components/pricing-table/pricing-table-header-cell.ts
@@ -23,6 +23,31 @@ class DDSPricingTableHeaderCell extends StableSelectorMixin(DDSStructuredListHea
   @property({ reflect: true })
   type: PRICING_TABLE_HEADER_CELL_TYPES = PRICING_TABLE_HEADER_CELL_TYPES.COMPLEX;
 
+  @property()
+  private _ctaType?: String;
+
+  private _handleCTASlotChange = ({ target }) => {
+    const cta = target.assignedNodes()[0];
+
+    if (cta) {
+      this._ctaType = cta.getAttribute('cta-type').toLowerCase();
+    }
+  };
+
+  renderCTASlot() {
+    const { _ctaType: ctaType } = this;
+
+    return ctaType === 'video'
+      ? html`
+          <dds-video-cta-container>
+            <slot name="cta" @slotchange=${this._handleCTASlotChange}></slot>
+          </dds-video-cta-container>
+        `
+      : html`
+          <slot name="cta" @slotchange=${this._handleCTASlotChange}></slot>
+        `;
+  }
+
   render() {
     const { type } = this;
     const { tagWrapperSelector } = this.constructor as typeof DDSPricingTableHeaderCell;
@@ -45,7 +70,7 @@ class DDSPricingTableHeaderCell extends StableSelectorMixin(DDSStructuredListHea
               </div>
             </div>
             <div>
-              <slot name="cta"></slot>
+              ${this.renderCTASlot()}
             </div>
           </div>
         `
--- a/packages/web-components/src/components/pricing-table/__stories__/pricing-table.stories.ts
+++ b/packages/web-components/src/components/pricing-table/__stories__/pricing-table.stories.ts
@@ -58,11 +58,7 @@ const renderHeaderCell = (iterator: number): TemplateResult => {
             <bx-list-item>consectetur retention adispiscing elit sed do eiusm Eiusmod tempor</bx-list-item>
           </bx-unordered-list>
         </dds-pricing-table-header-cell-description>
-        <dds-pricing-table-header-cell-cta
-          cta-type="local"
-          href="https://www.carbondesignsystem.com/all-about-carbon/what-is-carbon/"
-          >Call to action</dds-pricing-table-header-cell-cta
-        >
+        <dds-pricing-table-header-cell-cta cta-type="video" href="1_9h94wo6b">Call to action</dds-pricing-table-header-cell-cta>
       </dds-pricing-table-header-cell>
     `,
     html`


packages/web-components/src/components/pricing-table/pricing-table-header-cell.ts
```diff
@@ -23,6 +23,31 @@ class DDSPricingTableHeaderCell extends StableSelectorMixin(DDSStructuredListHea
...
+
+    if (cta) {
+      this._ctaType = cta.getAttribute('cta-type').toLowerCase();
+    }
+  };
+
+  renderCTASlot() {
+    const { _ctaType: ctaType } = this;
+
+    return ctaType === 'video'
```
jkaeser, line 40: We should probably replace this with the `enum` defined by (I believe) the CTA component. It will look something like this instead:
```
import CTA_TYPE from 'path-to-cta/defs.ts';

...
    
return ctaType === CTA_TYPE.VIDEO
```

I'll work on that.

packages/web-components/src/components/pricing-table/pricing-table-header-cell.ts
```diff
@@ -23,6 +24,31 @@ class DDSPricingTableHeaderCell extends StableSelectorMixin(DDSStructuredListHea
...
+  };
+
+  renderCTASlot() {
+    const { _ctaType: ctaType } = this;
+
+    return ctaType === CTA_TYPE.VIDEO
+      ? html`
+          <dds-video-cta-container>
+            <slot name="cta" @slotchange=${this._handleCTASlotChange}></slot>
+          </dds-video-cta-container>
```
proeung, line 45: I appreciate the rendering of the `<dds-video-cta-container>` component here so that the adopter doesn't have to define it on the markup side, however, this is a slight divergent from the way the `<dds-video-cta-container>` usage has been documented on other components where we specifically asked adopters to wrap this video player component around the CTA component (see - https://carbon-design-system.github.io/carbon-for-ibm-dotcom/canary/web-components/?path=/docs/components-cta--text#using-a-video-player), when video type is available.

I'll defer to @kennylam and @ariellalgilmore on this, but we might need to adjust this and add a note in the "Docs" of the Pricing Table component for the following markup/HTML used to keep it consistent or explore a different direction.

```
<dds-video-cta-container>
    <dds-pricing-table-header-cell-cta cta-type="video" href="1_9h94wo6b" slot="cta">Call to action</dds-pricing-table-header-cell-cta>
<dds-video-cta-container>
```

proeung (changes requested):
Looks good, just have a comment about the inclusion of the `<dds-video-cta-container>` component.

packages/web-components/src/components/pricing-table/pricing-table-header-cell.ts
kennylam: @proeung This is a good point. However unless you see a use-case for the `composite` version of the player, where authors can alter the intended presentation of this component, I don't see a problem bundling the player inside (unless you really want to maintain that level of consistency, which I also totally get).

proeung: @kennylam I don't have much of a preference, just wanted to point out the inconsistency.

Perhaps, it's probably best to check in with @KDarch928 since he's the one that integrated the Pricing Tables in AEM and noticed this issue. Kayle, are you currently wrapping `<dds-video-cta-container>` in the component template of `<dds-pricing-table-header-cell-cta>` already, or is it handled through a different atom component where the ES import for that component is also being defined?

druckerk: @proeung 

We do have the ```<dds-video-cta-container>``` tag wrapping around the ```<dds-pricing-table-header-cell-cta>``` tag. However, the ```<dds-video-cta-container>``` wrapper will only appear around the ```<dds-pricing-table-header-cell-cta>``` tag when the "video" link type is selected. If it makes it easier to leave the video tag wrapped around the header cell cta tag no matter the link type selection, then we can update our code easily. (see AEM code snippet below) 

```
    <dds-video-cta-container data-sly-unwrap="${linkType != 'VIDEO'}">
        <dds-pricing-table-header-cell-cta data-sly-test.valid="${(wcmmode.edit && !showPlaceholder) || href || text}"
                                           class="${linkType == 'BLOG' || linkType == 'DOWNLOAD_PDF' ? 'bx__pricing-table__button--cta-type' : ''}"
                                           cta-type="${linkType.toLowerCase}"
                                           target="${target}"
                                           href="${href}"
                                           kind="tertiary"
                                           data-autoid="dds--pricing-table-header-cell-cta">
            <sly data-sly-test="${text}">
                ${text}
            </sly>
            <span slot="icon" class="${iconClass} bx--card__cta dds-ce--cta__icon" data-sly-test="${linkType == 'BLOG' || linkType == 'DOWNLOAD_PDF'}"></span>
        </dds-pricing-table-header-cell-cta>
    </dds-video-cta-container>
```

CC: @kennylam

druckerk: @proeung  
As long as all other link types work properly, then then the easy way to approach this would be to wrap the ```<dds-video-cta-container>``` around the ```<dds-pricing-table-header-cell-cta>```.  

Screen Shot 2022-07-13 at 4 55 54 PM

oliviaflory:
@andy-blum @proeung I am wondering what the use case is for including a video in the pricing table header. 

My understanding is that the CTA would direct the user to purchase a product or to a trial, and I don't see why we would allow a video. I spoke briefly with Jen, and she and Aaron don't know of any need to include a video in this instance either. I worry this will not be a good experience for our users. Would you be able to provide any additional context?

andy-blum:
> @andy-blum @proeung I am wondering what the use case is for including a video in the pricing table header. 
> 
> My understanding is that the CTA would direct the user to purchase. I worry this will not be a good experience for our users. Would you be able to provide any additional context?

I would argue this is a content guidance issue and that the video cat ought to still work, even if it's against recommendation

oliviaflory:
> I would argue this is a content guidance issue and that the video cta ought to still work, even if it's against recommendation

I have to counter that argument, we have to strike the right balance of good design/UX within components and relying on content guidance. 

1. There is no use case for this feature. The AEM designers and content strategists have confirmed this.
2. We shouldn’t be adding unnecessary features to components
3. We shouldn’t be adding features that are against recommendation, especially ones we would actively tell an author/designer/dev not to do.
4. The UX of launching a video from the Pricing table is a bad user experience, and we shouldn’t be adding features that aren’t going to be good for the user or for other teams to consume.

jkaeser:
> > I would argue this is a content guidance issue and that the video cta ought to still work, even if it's against recommendation
> 
> I have to counter that argument, we have to strike the right balance of good design/UX within components and relying on content guidance.
> 
> 1. There is no use case for this feature. The AEM designers and content strategists have confirmed this.
> 2. We shouldn’t be adding unnecessary features to components
> 3. We shouldn’t be adding features that are against recommendation, especially ones we would actively tell an author/designer/dev not to do.
> 4. The UX of launching a video from the Pricing table is a bad user experience, and we shouldn’t be adding features that aren’t going to be good for the user or for other teams to consume.

Hi, @oliviaflory! I can see both sides here. From the dev perspective, this isn't exactly implementing a feature but rather maintaining consistent behaviors for CTAs wherever they appear in the code base. Looking at this from a UX perspective, though, it's sensible to disable the video CTA option if it never makes sense to use it here.

@druckerk Is it possible to disable the video CTA type option for this component in the AEM content editing interface? If so, we can go ahead with technically supporting video CTAs (in the event this component is ever used for something other than pricing tables, despite the name) while making sure they can't actually be used on IBM.com.

What do folks think of this potential compromise? cc @andy-blum @proeung

jwitkowski79:
Hi! So just making a note the @oliviaflory and I did discuss this and we brought it back to the AEM content strategy team where Aaron Apple noted that we'd never want to offer the option of a video for a CTA type within the pricing table. I'd hate for this to show up as an option, even if we write content guidance around it because it creates a bad user experience. I think we'd want to avoid adding any functionality that's against recommendation because if it's there, authors will try to use it regardless. From a code perspective, if we have to include the video functionality for consistency, I guess the second best thing would be to disable it in AEM.

druckerk:
@jwitkowski79 - We currently have the "video" cta type option disabled in the pricing table; the only cta type options available are "link", "PDF", and "Blog". One of the reasons we were trying to implement this option was to provide consistency with all our cta components already being utilized in AEM. However, If there are no plans to utilize the ```<dds-pricing-table-header-cell-cta>``` tag in any future components where it will require the "video" cta type, then we can just leave this option disabled.  

CC: @proeung, @jkaeser , @andy-blum

proeung:
@druckerk Thanks for providing some context from the AEM field content model side. Yes, let's move forward with disabling the "video" CTA type for the field group that's managing the display of the Pricing Table Header Cell CTA.

@andy-blum As for this PR, I think we can still move forward with addressing the functionality side. Let's go ahead and remove the `<dds-video-cta-container>` (see [link](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/9078/files#diff-ea4a8cde71efad08d182c949d9055ada189af1ee9abc6337612adba508db4ae6R43)) in the render slot so that we're keeping consistent with the usage of this component within the existing documentation. Also, please fix the merge conflict.

andy-blum:
@proeung if we're removing the usage of `<dds-video-cta-container>` in the component's shadowDOM, then this entire PR is no longer needed. Should we just close the PR & issue instead?

proeung:
@andy-blum I see that the `<dds-video-cta-container>` needed to be in the component’s shadowDOM in order for the Video Lightbox to initialize correctly. The suggested markup outlined here (https://codesandbox.io/s/lucid-satoshi-sqvbri?file=/index.html:4221-4265) wouldn’t work. Let's just proceed with resolving the merge conflict in the PR then.

oliviaflory:
Hey all, I apologize for the delay in getting back on this PR. 

I agree with @andy-blum with removing the usage of the video player this PR is no longer needed. 

I discussed with the AEM designers, Carbon, and Carbon for IBM.com designers and there is no need for a video in the Pricing table in AEM or in Carbon for IBM.com codebase.

Issue #8814, opened 2022-05-09
[e2e testing]: Set up parallelization of Cypress e2e testing on Github Actions


jeffchew:
### The problem

All of the tests are running in serial on a single runner, which is taking a long time for tests to complete. 

### The solution

Set up parallelization of tests across multiple Github runners:
https://docs.cypress.io/guides/continuous-integration/github-actions#Parallelization

### Application/website

Carbon for IBM.com

### Business priority

Medium Priority = upcoming release but is not pressing

### What time frame would this ideally be needed by (if applicable)

Ideally sooner, this will help PRs get checked and merged faster if we can shrink the time it takes to run these tests. Right now web components are running over an hour!

### Examples

https://docs.cypress.io/guides/continuous-integration/github-actions#Parallelization

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #9080, opened 2022-07-09
feat(tests): enable parallelization

IgnacioBecerra:
### Related Ticket(s)
#9093 
#8814 

### Description
This PR introduces the parallelization feature for Cypress testing. 
The work for this PR consisted in creating a IBM Cloud CodeEngine instance out of an image provided by the sorry-cypress service, especifically the `sorry-cypress-director` service. The Director instance is the one who controls what tests each machine should run, coming from the same spec file run.

In order to achieve parallelization among the newly created machine matrix, we are now using the `cy2` dependency, which intercepts the parallelization call made to Cypress Dashboard and redirects all the data to our CE instance. 

However, in order to speed things up and ensure each machine doesn't need to install and build every time, we separate the workflow into separate pieces:

1. Install job
   - Install dependencies 
   - Cache dependencies
   - Build storybook
   - Upload `storybook` artifacts
2. Run tests
   - Retrieve cached dependencies
   - Retrieve `storybook` artifacts
   - Run tests
   - If one test fails, create a `failure.txt` file and upload as artifact
3. Check if all tests passed
   - Wait until all tests finish
   - Retrieve `failure.txt` artifact
   - If said file exists, fail this job

The caching and artifact uploading ensures that each of the test jobs don't need to install/rebuild everything all over again. With this setup, that should only happen once, speeding up the rest of the jobs.

### Changelog

**New**

- introduced parallelization 

**Changed**

- `e2e-tests.yml` file


packages/react/tests/e2e-storybook/cypress/fixtures/components.json
```diff
@@ -0,0 +1,242 @@
+[
```
kennylam, line 1: Pretty sure this file should be added to `.gitignore`?

packages/web-components/tests/e2e-storybook/cypress/fixtures/components.json
```diff
@@ -0,0 +1,254 @@
+[
```
kennylam, line 1: Pretty sure this file should be added to `.gitignore`?

kennylam (commented):
Nice one @IgnacioBecerra! Do you think this would be ready to go as-is, or is more optimization/testing needed on CodeEngine before we merge?

IgnacioBecerra:
@kennylam Due to the way parallelization occurs, those files need to be already existing since the Storybook doc test needs it right away. If it doesn't exist yet the test will fail right away. Figured that since we already have the other fixtures there and we haven't had a new component in a while it should be fine.

As for merging this, I want to keep running the tests for just a bit longer in this PR to ensure things are still working/passing before having them run in all of our PRs.

m4olivei:
Hey there :wave: I came across this PR working in the `feat/masthead-v2` family of feature branches where I was cleaning up tests there to sync with changes. I started looking at the timing, specifically the `e2e-tests` job, and ended up finding this PR.

I'm wondering where this work is at? Also wondering, if we split the caching and parallizing pieces, could we get caching as an easy win and split paralliation to a follow up? Just a thought, not knowing much about the work involved here, and being very new to Github actions (coming from CircleCI).

kennylam:
@m4olivei We've had to put this work on hold for a bit, but it's definitely something we want to get done in the future. Your suggestions are great and hopefully we can implement something along those lines once we streamline the process and see the net benefits from our first real world tests.

IgnacioBecerra:
Closing in favor of #10314

PR #9095, opened 2022-07-13
fix(styles): removes invalid selectors

kennylam:
### Description

This fixes various invalid CSS selectors.

### Changelog

**Changed**

- update styles to remove invalid slotted selectors


packages/styles/scss/components/content-block-segmented/_content-block-segmented.scss
```diff
@@ -63,7 +63,10 @@
   :host(#{$dds-prefix}-content-block-segmented)
     .#{$prefix}--content-block__children
     .#{$prefix}--content-block-segmented__media
-    ::slotted(:not([slot])),
+    ::slotted(:not([slot]):last-of-type) {
```
annawen1, line 66: for this, do we also need to include this as well? some of the percy snapshots for the HTML template look off in terms of spacing - wondering if it's just a false positive?

```
.#{$prefix}--content-block__children
    .#{$prefix}--content-block-segmented__media
    ::slotted(:not([slot])) {
    display: block;
    margin-top: $carbon--layout-05;
    margin-bottom: $carbon--layout-05;
}
 ```

kennylam: Doh, good catch!

IgnacioBecerra (approved):
No storybook percy changes, LGTM!

PR #9130, opened 2022-07-19
chore(react): update Codesandbox examples

kennylam:
### Description

This update replaces `parcel-bundler` with `react-scripts` for all React Codesandbox examples. Additionally, it removes unused dependencies, files, and code. With the deprecation of `parcel-bundler`, the React Codesandbox examples builds were all failing, causing [this](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/runs/7397025337?check_suite_focus=true) action to fail.

**Note:** The `react-scripts` version being used is one major version behind and released ~1 year ago. There is an internal dependency error in the latest version (5.0.1) which is still open as a [bug](https://github.com/facebook/create-react-app/issues/11773). None of the solutions presented in this discussion resolve the issue in our examples. Once this bug is resolved in the latest version, the examples can be updated to use them.

image

### Changelog

**New**

- adds Browserstack entries to `package.json`

**Changed**

- replace `parcel-bundler` with `react-scripts`

**Removed**

- unused dependencies, files, and imports


ariellalgilmore:
I tested a few locally and they all seem to be working, but none of them are coming through here: https://react-codesandbox.s3-web.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/9130/index.html?

kennylam:
> I tested a few locally and they all seem to be working, but none of them are coming through here

@ariellalgilmore Fixed!

ariellalgilmore (approved):
Great! LGTM!

Issue #9124, opened 2022-07-19
[react]: `expressive-modal` story does not render correctly


kennylam:
all - unable to load story `components-expressive-modal--default` (console)

PR #9140, opened 2022-07-20
fix(expressive-modal): render footer button content correctly

emyarod:
### Related Ticket(s)

none

### Description

This PR updates the expressive modal React wrapper story so that the modal footer button content renders correctly

discovered while investigating #9124

image


ariellalgilmore (approved):
LGTM!

PR #9143, opened 2022-07-20
chore(deps): upgrade react storybook to use webpack 5

kennylam:
### Description

This upgrades the React package to use `webpack@5` for Storybook. Also addresses an issue where knobs were being added to stories when clicking from one to the next. This fix address the issue of incorrect knobs appearing in the wrong stories, but the number shown in the `knobs` tab would still be incorrect (it will add the number of knobs from the previous story into the current one). To address this, the number has been removed from the tabs to avoid confusion.

### Changelog

**New**

- upgrade `webpack` to version 5

**Changed**

- remove number `(x)` of knobs from stories from showing in `knobs` tab
- upgrade various Webpack-related deps

**Removed**

- unnecessary deps


ariellalgilmore:
deploy previews are failing because of webpack 5 :(

ariellalgilmore:
Looks great!!! just noticed this weird thing though in the react wrapper stories. The font, images, and some styles aren't coming through

Example:
Screen Shot 2022-07-21 at 4 17 16 PM

kennylam:
> Looks great!!! just noticed this weird thing though in the react wrapper stories. The font, images, and some styles aren't coming through

@ariellagilmore Yep, I noticed that too. Happens during `build-storybook:react` only. Running Storybook only does not cause this issue. Still working on it.

Issue #8911, opened 2022-05-25
[lightbox-media-viewer - with carousel]: QA: Screen reader is not identifying the slide headings


Gopi916:
**Engineering info:**  
- Innovation Team engineer: 
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2520
- DPO consulting engineer: N/A

-----

### Description

Page Element:
Headings in the carousel slide

Issue Description:
Screen reader does not identify the heading tag for the slide headings. they are announced only as text

Expected:
"Use Heading Level 1 for the main content of the page. Heading Level's should be in logical order, Starting from H1, H2, H3, H4, H5 and H6 respectively.
don't use Headings for font, For increasing font & 
style use CSS."
Do not skip Heading levels

User Impact:
If the screen reader does not identify page headings, users are unable to skim through various sections in the page and understand the content.

### Component(s) impacted

lightbox-media-viewer - with carousel

### Browser

Safari, Firefox

### Carbon for IBM.com version

canary

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

Carbon for IBM.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/8757/iframe.html?id=components-lightbox-media-viewer--with-carousel

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [ ] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

gopinathan-n:
@andy-blum @kennylam Screen reader is announcing the headings in each slide now. Issue is fixed. Tested in Web components carousel default [https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/9149/index.html?path=/story/overview-getting-started--page](url)

Issue #8915, opened 2022-05-25
[lightbox-media-viewer - with carousel]: QA: The label "previous page"/"next page" does not seem appropriate. 


Gopi916:
### Description

Page Element:
Label for Previous/next buttons

Issue Description:
Since there is no page navigation, The label "previous page"/"next page" does not seem appropriate. 

Expected:
The label can be given as "previous slide"/"next slide" respectively

User Impact:
If the label is incorrect or inadequately descriptive, it affects the screen reader user’s ability to understand the purpose of the control.   

### Component(s) impacted

lightbox-media-viewer - with carousel

### Browser

Safari, Firefox

### Carbon for IBM.com version

canary

### Severity

Severity 4 = The problem is not visible to or noticeable to an average user. Affects minor functionality, no workaround needed.

### Application/website

Carbon for IBM.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/8757/iframe.html?id=components-lightbox-media-viewer--with-carousel

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [ ] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

gopinathan-n:
Now Screen reader announces the label as "previous"/"next". Issue is fixed

Issue #8914, opened 2022-05-25
[lightbox-media-viewer - with carousel]: QA: Landmark regions for Carousel 


Gopi916:
### Description

Page Element:
Landmark regions for Carousel 

Issue Description:
To provide more context to the screen reader user, the carousel which includes the slides and the Carousel controls could be given in a labelled landmark region.

Expected:
Provide the carousel including all the slides within a region role and aria label. E.g. role=”region” aria-label=”XYZ Carousel”

User Impact:
When the page content is not covered within a labelled landmark region, screen reader users are unable to skim through the page and understand its structure.

### Component(s) impacted

lightbox-media-viewer - with carousel

### Browser

Safari, Firefox

### Carbon for IBM.com version

canary

### Severity

Severity 4 = The problem is not visible to or noticeable to an average user. Affects minor functionality, no workaround needed.

### Application/website

Carbon for IBM.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/8757/iframe.html?id=components-lightbox-media-viewer--with-carousel

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [ ] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

gopinathan-n:
@andy-blum @kennylam Issue is not fixed. The carousel section can be given within a labelled region role. This will help the user to understand the semantics of the carousel and also the beginning and ending boundary of the carousel section. Expected to provide the carousel section in region role with the aria-label="XYZ carousel" where "XYZ" is the name of the carousel

Rasubra8:
Now the previous/next buttons are covered within a navigation region. Instead of this the entire carousel section starting from the H3 till the next button can be given within the region role with the aria-label="XYZ carousel.

Issue #8913, opened 2022-05-25
[lightbox-media-viewer - with carousel]: QA: Screen reader not announcing the change in slide numbers


Gopi916:
**Engineering info:**  
- Innovation Team engineer: 
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2521
- DPO consulting engineer: N/A

-----

### Description

Page Element:
Previous and Next buttons in the carousel

Issue Description:
When we activate the Previous/Next buttons, visually the slides are changing. But Screen reader does not announce anything. Hence the user is not aware of this behavior.

Expected:
Screen reader should announce the change instantaneously when the previous/next buttons are activated. . For example: 1/5... 2/5... etc.

User Impact:
If the screen reader fails to identify a dynamic content change, it affects users who are blind as they may not know about the change in the page.

### Component(s) impacted

lightbox-media-viewer - with carousel

### Browser

Safari, Firefox

### Carbon for IBM.com version

canary

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

Carbon for IBM.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/8757/iframe.html?id=components-lightbox-media-viewer--with-carousel

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [ ] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

gopinathan-n:
@andy-blum @kennylam Issue is fixed in Firefox_JAWS.

Issue is not fixed in Voiceover. In iPhone_Voiceover, when we activate previous/next buttons, Voiceover announces the aria-live twice. e.g. "slide 1 of 5 showing 0 items. slide 1 of 5 showing 1 items". This is happening for all the slides.

gopinathan-n:
Screen Reader is expected to announce the aria-live once with the correct value

Rasubra8:
Now the following issue is observed in both jaws FF and Vo iPhone Safari,
[...]

PR #9149, opened 2022-07-21
fix(lightbox-carousel): accessibility & QA improvements to lightbox carousels

andy-blum:
## Related Ticket(s)

Closes: #8911 #8913 #8915

Also related: #8912 #8914

## Description

This PR makes several accessibility improvements to the DDSCarousel and DDSLightboxMediaViewer components, as well as the lightbox carousel story

### #8911 Update Lightbox to use semantic headings ✅
Lightbox titles now render title text within `<h2 style="all: inherit;">` tags. This gives them semantic `h2` headers but does not bring along the user agent/carbon styles for the element.

### #8912 Add label for modal dialog ✅
While elements with `[role="dialog"]` are required to have a label for WCAG, that label should be relevant to the content within the dialog. Because the component cannot know what content (or even what language) is in it, all we can do is make space for content entry. This already exists within the component in the form of an `aria-labelledby` attribute pointing to [a div that accepts slotted content](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/expressive-modal/expressive-modal.ts#L358). This PR adds a visually hidden modal title to the lightbox carousel story

### #8913 Announce carousel navigation to screen reader ✅ 
The carousel's `renderStatus()` method now renders a visually-hidden span with `aria-live="polite"` and uses a new `formatAnnouncement` property function to format that announcement into a string that accounts for single slides or slide groups. This property is written adjacent to the `formatStatus` property that allows localizing the pagination information.

### #8914 Create new landmark region for carousels ✅ 
_~No changes were made for this ticket.~_

~8914 suggests, "to provide more context to the screen reader user, the carousel which includes the slides and the Carousel controls could be given in a labelled landmark region." While this would be good, wouldn't adding a landmark region require a label that describes _what is in_ the carousel, not just that it _is_ a carousel? It would not be difficult to add a fallback label if this is not the case.~

The carousel now has an additional `<div>` wrapping the scrolling contents and the navigation. This div has `role="region"` and an `aria-labelledby` attribute that points to a new div with a `<slot name="title"><span class="bx--visually-hidden">Carousel</span></slot>`. This allows authors to provide a title relevant to the carousel's contents with the generic fallback label of "carousel"

### #8915 Update carousel's next/previous button labels ✅
The carousel allows authors to specify the labels of the previous & next buttons, but we also provide a default. Previously this value was "next page", but the word `page` can be confusing. We now provide a default fallback that looks at the carousel's `pageSize` and returns either "next slide" or "next slide group" if more than one slide is visible.

### [No Ticket] Refactor carousel item's interactivity based on inert
When the `[inert]` attribute was added to the carousel, it was added in a way that wasn't implemented fully in safari/firefox, and didn't account for carousels with multiple visible values well. That's been updated to now use an intersection observer, and it will mark carousel items more less than 75% visible as `inert` and `aria-hidden`.

### Changelog

**Changed**

- Accessibility improvements to lightbox media viewer component
- Accessibility improvements to lightbox carousel story
- Accessibility improvements to carousel component


packages/web-components/src/components/lightbox-media-viewer/__stories__/lightbox-media-viewer.stories.ts
```diff
@@ -129,6 +129,7 @@ export const WithCarousel = ({ parameters }) => {
       @dds-expressive-modal-beingclosed="${handleBeforeClose}"
       @dds-expressive-modal-closed="${onClose}"
     >
+      <span slot="header" class="bx--visually-hidden">Lightbox modal with carousel</span>
```
proeung, line 132: Minor comment: I think we should note this hidden content for accessibility in the Docs tab. We need to let adopters know that this contextual label should still be accessible to screenreaders while being removed from the interface for sighted users.

kennylam:
@proeung @andy-blum We had a review QA about the a11y issues with this component. For users who rely on screen readers, the labels for previous and next are not really an issue. Simply having `previous`/`next` is enough, and in fact `next slide group` is more confusing. The main point is the visually hidden element with `aria-live`. The most informative would be an announcement to the effect of 'Slide _x_ of _y_ with _n_ items'. It looks like this announcement is already part of the PR, so QA is going to review the deploy preview and share any comments with us.

One thing to note is the visually hidden element will need to exist for `sm`, `md`, and `>=lg` breakpoints, as the number of slides and items displayed will be different for all 3.

packages/web-components/src/components/carousel/carousel.ts
```diff
@@ -330,15 +371,24 @@ class DDSCarousel extends HostListenerMixin(StableSelectorMixin(LitElement)) {
...
-    const { start, pageSize, formatStatus, _total: total } = this;
-    // Copes with the condition where `start % pageSize` is non-zero
-    const pagesBefore = Math.ceil(start / pageSize);
-    const pagesSince = Math.ceil((total - start) / pageSize);
-    return formatStatus({ currentPage: Math.ceil(start / pageSize) + 1, pages: pagesBefore + pagesSince });
+    const { formatStatus, formatAnnouncement, _getStatus: status } = this;
+
+    return html`
+      ${formatStatus(status)}
+      <span class="bx--visually-hidden" aria-live="polite">${formatAnnouncement(status)}</span>
```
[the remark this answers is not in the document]
andy-blum: https://github.com/carbon-design-system/carbon/blob/main/packages/styles/scss/utilities/_visually-hidden.scss

packages/web-components/src/components/lightbox-media-viewer/__stories__/lightbox-media-viewer.stories.ts
proeung: Addressed in this commit - https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/9149/commits/ce744633ee1a7bcf38352fae145b8c2e7c09c5e9

proeung (approved):
LGTM! Thanks for addressing my feedback.

@kennylam Can we get a code review/QA from your team? Thanks!

andy-blum:
☝️ @kennylam I can hop in tomorrow for the QA time at 9

andy-blum:
@kennylam @Gopi916 The most recent commit updates the aria-live announcement logic. It utilizes the intersection observer to count the number of visible items and update the announcement text after a brief 100ms delay to ensure the slide transition is fully complete.

If the screen reader issue experience in firefox is a render lifecycle race condition, this fix should hopefully also address that.

proeung:
@kennylam Looks like this PR is at a good place, but could probably benefit from a code review from your team. Can you take a look and let us know if these changes can be marked as "Ready to merge"? Thanks!

andy-blum:
@Gopi916 I've updated the PR to wrap carousels in `<div role="region" aria-labeledby="carousel-title">` and added a slot for authors to add carousel title text. Additionally, I think the iOS voiceover text _should_ be fixed. I updated the delay from 100ms to 400ms, and it's now using the design token from `@carbon/motion`.

andy-blum:
Updated PR summary to indicate issues that have been closed by @Gopi916 and to update the solution description for #8914

andy-blum:
@Gopi916 Can we get testing for #8913 and #8914 on this PR?

andy-blum:
Per conversation last week with @Rasubra8 and @kennylam this has passed accessibility review.

@oliviaflory would you mind taking a look to make sure everything is still aligned with the designs?

packages/styles/scss/components/lightbox-media-viewer/_lightbox-media-viewer.scss
```diff
@@ -136,6 +136,11 @@
         @include carbon--make-col-ready;
         @include carbon--make-col(12, 16);
 
+        width: auto;
+        flex: none;
```
jkaeser, line 140: This is constraining video media width in Safari.

Screen Shot 2022-10-13 at 1 00 05 PM

jkaeser (changes requested):
@andy-blum All the a11y updates look great. Really excellent work. 

I noticed one visual regression on Safari that should be looked into.

oliviaflory:
@andy-blum thanks for fixing the carousel controls! A few additional comments:

### 1. Unexpected image/video scaling
The image or video is scaling when the browser has a shorter height, see exaggerated examples below. @jwitkowski79 and I were expecting the image to stay the same width as the default lightbox media viewer.

**Carousel**
Screen Shot 2022-10-14 at 3 51 05 PM

**Default**
Screen Shot 2022-10-14 at 3 51 08 PM


### 2. No fade on mobile
On my iPhone 12 Pro in Safari and Chrome I don't see the gradient fade that appears in the desktop experience. This is helpful for users to know there is more content below. Right now you only see tiny specs of the additional copy. Let me know if this isn't a part of the Default lightbox and we can open a feature enhancement!
image_123986672


**Default with gradient fade**
Screen Shot 2022-10-14 at 4 28 46 PM

### 3. Difficult to read text with fade
Lastly, it's difficult to read the last line of text with the gradient fade. We might want to account for extra padding at the bottom so the user can scroll the text enough past the gradient to have the text fully visible.
Screen Shot 2022-10-14 at 4 30 43 PM

andy-blum:
Just merged in the carbon-web-components addition and had to do a bunch of merge conflicts. There may be some regressions at this point.

packages/styles/scss/components/lightbox-media-viewer/_lightbox-media-viewer.scss
m4olivei: I can't replicate this issue in Desktop Safari (version 16.2). It behvaes the same as Chrome. Tested at this link: https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/9149/iframe.html?id=components-lightbox-media-viewer--with-carousel&viewMode=story

Is this still reproducible for you @jkaeser ?

jkaeser: @m4olivei Nope, I can no longer reproduce. Checked on Safari 16.1.

oliviaflory (approved):
Looks good from design perspective! thanks @andy-blum

m4olivei:
Thanks @ariellalgilmore !

I've adjusted things around the gradient. Due to the change we made affecing the scoll area at desktop, where the whole body of the lightbox content scrolls together (image / video and title / description), I also needed to change the gradient at desktop such that it spans the whole width of the scrolling area. This is an implementation limitation, I tried at some length to affect only the description, to no avail. In the end, to me this fits the intent of the gratient which by my interpretation is to signal additional contents beyond the visibile area of the scroll container. Curious to get your thoughts!

ariellalgilmore:
I believe it was design intention to not have the media scroll in the larger breakpoints, so wondering what @oliviaflory and @RichKummer think. Maybe some background on why this was changed for the Accessibility fixes would be helpful...

m4olivei:
We arrived at scrolling the media together with the description with consideration for video. When a video is used in the lightbox carousel, the user needs to be able to scroll to the media controls once playback is initiated. Otherwise, the the scroll was clipped, the user wouldn't be able to access any of the media controls. Taken together with @oliviaflory 's request to avoid scaling the video to fit the available height (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/9149#issuecomment-1279434723), the logical conclusion was to scroll the media and description together, while pinning the carousel controls to the bottom of the lightbox so they remain accessible regardless of available height.

m4olivei:
Friendly bump for @ariellalgilmore / @oliviaflory / @RichKummer . Are we OK with the implementation here as is, or is more iteration necessary? Thanks.

ariellalgilmore:
just noticing now the tests for react/web components are failing as well @m4olivei

m4olivei:
> just noticing now the tests for react/web components are failing as well @m4olivei

Thanks for pointing that out. I'll work on that.

oliviaflory (approved):
looks good from design perspective

m4olivei:
Great thanks @oliviaflory ! I also fixed the test failures we were seeing. So as far as I know (still new to the process), we're ready to merge?

oliviaflory:
@m4olivei hmm it looks like the Test app deploy previews are failing, is that normal @ariellalgilmore ?

ariellalgilmore:
i think @kennylam mentioned he was looking at the test app failures, but shouldn't be an issue with this PR

kennylam:
> @m4olivei hmm it looks like the Test app deploy previews are failing, is that normal @ariellalgilmore ?

Yep we're looking into that but it doesn't impact the work in this PR.

ariellalgilmore:
seems like just changes requested from @jkaeser need to be approved

packages/styles/scss/components/lightbox-media-viewer/_lightbox-media-viewer.scss
m4olivei: Hey @jkaeser looks like the review is needing your approval from this review you left a ways back.

PR #9168, opened 2022-07-26
fix(card-group): apply empty card styles in extended components

### LINKED ISSUE #9147: [card-group]: empty card styles fail to apply when extending component
**Engineering info:**  
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/HC-2522

-----

### Contribution type

Components

### Contribution type (other)

_No response_

### Intent

We've extended `dds-card-group` in C4AEM as `caem-card-group`, and the automatically added empty cards are visible.

Screen_Shot_2022-07-18_at_12_25_37

The styles that make these "invisible" are failing to apply since the selector is more specific than it needs to be: 

`dds-card-group dds-card-group-item[empty] { ... }`

### Solution

Make the selector less specific. Simply removing `dds-card-group` from this selector fixes the problem whilst not negatively impacting `dds-card-group`. This might also be an opportunity to move the styles from the `web-components` package to the `styles` package so they can be extended if need be.

### IBM Organization or Company you are part of

Innovation Team, working on Carbon for AEM.

### Code of Conduct

jkaeser:
### Related Ticket(s)

Resolves #9147

### Description

The CSS that hides the extra cards added to make the grid work was scoped to only ever apply to empty cards within `<dds-card-group>` elements, and those styles could not be extended (e.g. in Sass) since they were bundled up in the `DDSCardGroup` class. This made it impossible to hide the extra cards in components that want to extend from this class, such as our `CAEMCardGroup`, without copying and pasting CSS from the C4IBM source code.

This PR makes changes that should have no visual impact on the DDS components but will allow downstream inheritors to receive styles that are important to this component's appearance.

### Changelog

**Changed**

- Moves CSS styles that handle hiding the empty cards into the `ibmdotcom-styles` package so they can be extended if necessary.
- Removes unnecessary specificity from CSS rules so that empty cards in extending components are targeted by styles that hide them.


kennylam (approved):
LGTM!

Issue #9198, opened 2022-08-01
[react/wc]: expand storybook folders on load


ariellalgilmore:
Components folder shows collapsed 

<img width="1417" alt="Screen Shot 2022-08-01 at 9 26 09 AM" src="https://user-images.githubusercontent.com/20210594/182197689-7d51d08e-2cf9-4818-9ced-d417a7162c29.png">

PR #9202, opened 2022-08-02
fix(react/wc): storybook cleanup folders open onload

ariellalgilmore:
### Related Ticket(s)

#9198

### Description

Storybook cleanup. Now `Components` folder in the sidebar should be open when loaded.

### Changelog

**Changed**

- lots of cleanup


kennylam (approved):
Great stuff @ariellalgilmore!

Issue #9162, opened 2022-07-25
[Pricing Table]: QA: User is unable to scroll to the bottom content of the page.


Gopi916:
**Engineering info:**  
- Innovation Team engineer: 
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/ADCMS-2458
- DPO consulting engineer: N/A

-----

### Description

Page Element: Bottom region of the page

Issue: User is unable to scroll to the bottom content of the page.

### Component(s) impacted

Pricing Table

### Browser

Firefox

### Carbon for IBM.com version

v1.36.0-rc.0

### Severity

Severity 1 = The design is broken in a critical way that blocks users from completing tasks or damages the brand. Affects major functionality, no workaround.

### Application/website

Carbon for IBM.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://carbon-design-system.github.io/carbon-for-ibm-dotcom/canary/web-components-experimental

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [ ] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

proeung:
@Gopi916 Can you provide steps to reproduce this issue or a video screenshot of the issue you're seeing within the Storybook preview? So far, we're not seeing this scrolling issue within the SB preview or on the AEM side. Moving the severity label down to a lower number for now.

cc: @jkaeser

jkaeser:
> @Gopi916 Can you provide steps to reproduce this issue or a video screenshot of the issue you're seeing within the Storybook preview? So far, we're not seeing this scrolling issue within the SB preview or on the AEM side. Moving the severity label down to a lower number for now.

@proeung You can see this when you navigate the to the [Default Story](https://carbon-design-system.github.io/carbon-for-ibm-dotcom/canary/web-components-experimental/?path=/story/components-pricing-table--default) and attempt to scroll down. The fixed header transition engages, but due to the way browser's automatically scroll when rendered content disappears.

https://user-images.githubusercontent.com/12532727/180861788-b315345c-d3f1-4044-8940-ec4190ccbdbb.mov

proeung:
[...]

PR #9205, opened 2022-08-05
fix(pricing-table): accessibility QA updates

### LINKED ISSUE #9159: [Pricing Table]: QA: Entire page is not visible to the user at 400% zoom
**Engineering info:**  
- Innovation Team engineer: 
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/ADCMS-2457

-----

### Description

Page element:
Entire page

Issue : When the page is zoomed to 400%, entire page is not visible to the user.

Expected: Entire page should be visible even when the page is zoomed to 400%.

User Impact: When the page is not visible on 400% zoom, user with low vision will not be able to see the entire page content.

### Component(s) impacted

Pricing Table

### Browser

Firefox

v1.36.0-rc.0

### Severity

### Application/website

Carbon for IBM.com

### Package

### CodeSandbox example

https://carbon-design-system.github.io/carbon-for-ibm-dotcom/canary/web-components-experimental

_No response_

_No response_

### Code of Conduct

jkaeser:
### Related Ticket(s)

Resolves #9159, #9162 

#### Ensure header cells are fully visible regardless of zoom (#9159)
The pricing table header is capable of being taller than the viewport. When this is the case, users are not able to view the entirety of the header's contents before it transitions to its sticky state, which hides much of the cell content. This scenario is uncommon at the default viewport zoom level, but it becomes progressively more likely as the browser is zoomed in. Our Storybook examples exhibit the problem at around 400% zoom.

The best solution I could come up with is to disable the sticky functionality when the header is taller than the viewport. This seems better than the alternative, which is delaying the sticky transition until the user reaches the bottom of the header row. In any scenario where the browser window is not tall enough to display the entirety of the header, the value of having a sticky header is vastly diminished. For example, here's what a sticky header at 400% zoom looks like on my 13" Macbook:

Screen Shot 2022-08-08 at 5 02 24 PM

Compare that to the version of the component where sticky header has been disabled:

Screen Shot 2022-08-08 at 5 03 31 PM

IMHO the non-sticky version seems much more usable. Note that this will affect short browser windows at default zoom levels:

With sticky (current behavior): 

https://user-images.githubusercontent.com/12532727/183514780-f53ef46b-708f-48be-85a3-41220daf1c02.mov

Without sticky (changes introduced in this PR): 

https://user-images.githubusercontent.com/12532727/183514841-7461550b-d55e-46e5-b8bb-a31111640aad.mov

#### Ensure Storybook contents are fully visible on scroll (#9162)

Scrolling down in a Pricing Table story causes the sticky header transition to start, but since there is no other content, it immediately bounces back into its initial position. I fixed this by adding an arbitrary 400px of space beneath the Pricing Tables in Storybook so that users can fully scroll down.

#### Account for new cells in sticky transitions

Adding new columns in Storybook then scrolling down revealed the new columns' header cells were not animating closed during the sticky transition. I've updated the logic to be aware of the new cells and include them in the animation.

#### Prevent unnecessary intersection observer resets

While debugging, I noticed the intersection observers were being reset during each frame of the sticky header animation. I added logic to prevent this unnecessary computation that could cause unexpected behaviors.

### Changelog

**Changed**

- `DDSPricingTable`s' sticky header behavior will no longer trigger when the table header is taller than the viewport.
- Prevents resetting `DDSPricingTable` intersection observers unnecessarily.
- Adds space at the bottom of Pricing Table stories so users can scroll down without sticky header pulling their viewport back up.
- `DDSPricingTableHeaderCell`s added after initial render are animated during sticky header transitions.


kennylam:
@jkaeser I like this solve! I guess we just need design's blessing?

jkaeser:
> @jkaeser I like this solve! I guess we just need design's blessing?

Thanks, @kennylam! I added some members of the design team as reviewers.

jkaeser:
> Deploy preview created for package `"Web Components (experimental)"`: https://ibmdotcom-web-components-experimental.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/9205/index.html

@gopinathan-n Please test using this link.

RichKummer (approved):
LGTM! Thanks @jkaeser !

gopinathan-n:
@jkaeser Verified in deploy preview https://ibmdotcom-web-components-experimental.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/9205/index.html. User is able to scroll to the bottom at 100% and 400% as well.
Issue is fixed. 
One new observation is seen. Some blank spaces are at the bottom of the page.

Issue #9197, opened 2022-08-01
[percy]: snapshots all coming through white


ariellalgilmore:
When running `yarn visual-snapshot` all the diffs show a white page.

React is coming through properly, but the story names are using `/` instead of `|` which is causing differences.

Web components is showing white pages:

<img width="1435" alt="Screen Shot 2022-08-01 at 9 14 44 AM" src="https://user-images.githubusercontent.com/20210594/182194206-54e9af4b-68d4-4dda-8a63-4f312365766f.png">

ariellalgilmore:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Carbon-for-IBMcom-5d449f3642eb1962336cbe52/issues/carbon-design-system/carbon-for-ibm-dotcom/9197?planning-poker) with ZenHub @kennylam @emyarod @annawen1 @IgnacioBecerra

PR #9226, opened 2022-08-10
fix(react): update percy storybook 6

ariellalgilmore:
### Related Ticket(s)

#9197 [react]
### Description

updated percy and added a name field for stories, so we could keep the storybook folder system ths same.
### Changelog

**New**

- {{new thing}}

**Changed**

- updated percy
- added names to match to old percy snapshots so we can see if anything changed
- add groupId string to card static, so knobs wouldn't effect default card

**Removed**

- {{removed thing}}


annawen1 (approved):
LGTM!

Issue #8691, opened 2022-04-08
Tabs Extended media containing a Title heading section


pscheufens:
### The problem

We are using Tabs Extended Media now instead of Tabs Extended and noticed that in the Media version there is another div for a title element causing spacing issues. But this is not the case for Tabs Extended. 
Codebox shows the behaviour where the bx--heading resides inside the dds-tabs-extended-media. And we only have this tag in our markup. Changing it to dds-tabs-extended removes the heading area in the shadow root but we lose media support for all viewports.

### The solution

Would it possible to get rid of this bx--heading div? Since it resides inside the shadow root we cannot access it with css or markup. 

### Application/website

AEM

### Business priority

Medium Priority = upcoming release but is not pressing

### Examples

![screenshot_spacing](https://user-images.githubusercontent.com/29569154/162508110-b727e5e2-c7c1-4457-8754-6eff543b4cfc.png)
![screenshot_spacing2](https://user-images.githubusercontent.com/29569154/162508128-5fb1d97c-34ef-4c21-97ed-da67275de79c.png)

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

proeung:
@RichKummer Once you have determined the solution from the design side, please go ahead and assign this issue to my queue so that we can plan for this work in the next sprint.

oliviaflory:
@pscheufens Could you share the design you are working from?

I'd be happy to make the heading optional and keep the existing grid alignment. This would allow the use of the Horizontal table of contents in that first four columns, and ensure the tabs extended is aligned to any other Carbon for iBM.com components used on the page.

However I would not advise the Tabs extended to span the full 16 columns of the grid and would like to review the overall design if that is the case.

EAVilleda:
@oliviaflory We don't need the Tabs extended with media to extend the full 16 columns. We want it to be the same as Tabs extended.
https://www.ibm.com/standards/carbon/web-components/?path=/story/components-tabs-extended--default
https://www.ibm.com/standards/carbon/web-components/?path=/story/components-tabs-extended-with-media--default

[...]

PR #9265, opened 2022-08-17
feat(tabs-extended-with-media): make section heading optional

annawen1:
### Related Ticket(s)

Tabs Extended media containing a Title heading section #8691

### Description

Make the section heading of `tabs-extended-with-media` optional by adding a new `section-heading` property. `section-heading` will be disabled by default.

DESIGN PROPOSAL:
Screen Shot 2022-08-17 at 12 19 45 PM

STORYBOOK WITH CHANGES:
Screen Shot 2022-08-17 at 12 14 21 PM

Screen Shot 2022-08-17 at 12 14 13 PM

### Changelog

**New**

- `section-heading` property to `tabs-extended-with-media`.


oliviaflory:
@annawen1 this looks great! 

Question for you:
I wasn't expecting the Content item horizontal to stretch when the Section heading is turned off and the tabs extend to the first column
**With section heading** image is ~5 columns wide
Screen Shot 2022-08-17 at 4 58 58 PM

**Without section heading** image is a little more than 6 columns wide
Screen Shot 2022-08-17 at 4 59 21 PM


I was expecting the tabs and content item horizontal to shift over to the first column, but maintain their width. 
**Example:**
Screen Shot 2022-08-17 at 5 05 52 PM


Is what I showed above possible or is that not how the tabs container(?) works and whatever is in the container will span the full width

ariellalgilmore:
Looks good! Can we also adjust the react wrapper story to match? 

And just want to clarify, I forgot if we discussed setting the default to having the section heading there? or will we just need to remember to notify adopters that this could be almost a "breaking change" if there section heading disappears

annawen1:
> Looks good! Can we also adjust the react wrapper story to match?
> 
> And just want to clarify, I forgot if we discussed setting the default to having the section heading there? or will we just need to remember to notify adopters that this could be almost a "breaking change" if there section heading disappears

yeah, I'll adjust the react wrapper story as well. 

Good point about the default setting of section heading - if I remember correctly, we wanted the section-heading to be off/not displayed by default?  But if we do want to avoid breaking changes, we should switch to having the section heading displayed on default instead - what do you think @oliviaflory @kennylam?

oliviaflory:
I'm ok with it being a non breaking change, and keeping the section heading displayed as the default. I don't want angry people who expect it to be there and have it go away, I think AEM will be happy as long as it's optional. 

I will defer to @kennylam for final input though

annawen1:
> @annawen1 this looks great!
> 
> Question for you: I wasn't expecting the Content item horizontal to stretch when the Section heading is turned off and the tabs extend to the first column  .....
> Is what I showed above possible or is that not how the tabs container(?) works and whatever is in the container will span the full width

@oliviaflory from what I can tell, yeah the tabs container spans full width of whatever the adopter drops it into - I've adjusted the storybook though to adjust the grid depending on whether the section heading is disabled or not so that the tabs container maintains the same width in the story

ariellalgilmore (approved):
LGTM! Just need to update snapshots again :D

kennylam (approved):
LGTM! Yes, I think it should be off by default otherwise it would be a breaking change.

PR #9269, opened 2022-08-18
chore(github): add `publish-alpha` workflow

kennylam:
### Description

Adds GitHub workflow to publish `alpha` tag for NPM package from the `feat/masthead-v2` branch. This workflow can be modified to target the `c4dotcom-v2` feature branch when it is created.

### Changelog

**New**

- GH workflow `publish-alpha`

### DIFF
--- a/.github/workflows/publish-alpha.yml
+++ b/.github/workflows/publish-alpha.yml
@@ -30,18 +30,6 @@ jobs:
         run: echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_API_KEY }}" >> .npmrc
       - name: Run Lerna
         run: yarn lerna publish --alpha minor --preid alpha.${{github.run_id}} --dist-tag alpha --force-publish --no-push --no-git-tag-version --yes
-      - name: Deploy NextJS Test
-        uses: peter-evans/repository-dispatch@v1
-        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-alpha.yml
```diff
@@ -0,0 +1,48 @@
...
+        run: echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_API_KEY }}" >> .npmrc
+      - name: Run Lerna
+        run: yarn lerna publish --alpha minor --preid alpha.${{github.run_id}} --dist-tag alpha --force-publish --no-push --no-git-tag-version --yes
+      - name: Deploy NextJS Test
+        uses: peter-evans/repository-dispatch@v1
+        with:
+          repository: carbon-design-system/carbon-for-ibm-dotcom-nextjs-test
+          token: ${{ secrets.GH_DISPATCH_TOKEN }}
+          event-type: deploy-alpha
+      - name: Deploy Web Components HTML Test
```
jeffchew, line 39: I don't think we need this dispatch call for the alphas.

kennylam: Ah, yes. Good call.

jeffchew (approved):
LGTM!

Issue #9271, opened 2022-08-18
[Masthead v2.0.0]: Remove the suppression of the L0 menu items when L1 is present


proeung:
**Engineering info:**  
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/ADCMS-2538
- DPO consulting engineer: N/A

-----

### Description
- For the launch of the Masthead v2, the team is looking to bring back the menu items within the L0/Masthead when it's being stacked next to the L1 Nav.

### Additional information
- **Overview and scope of the engineering work should be included in the linked epic (#9208)**

### Design Specs
- Desktop - https://www.figma.com/file/35JfN0Sg4ltIIjM7jBDKMk/Navigation?node-id=1970%3A110816
- Mobile - https://www.figma.com/file/35JfN0Sg4ltIIjM7jBDKMk/Navigation?node-id=1970%3A111059

### Acceptance criteria
 - [ ] L0/Masthead menu items will be visible next to the L1 Nav.

proeung:
**Notes from the 08/23 "Navigation Stakeholder & Dev Weekly" meeting:**

- The visibility of the L1 Menu items within the hamburger menu on mobile is not a blocker for the end of the Q3 deadline.
- We're going to be following up with a new solution for displaying the L1 Nav on mobile using the accordion behavior that @JaniceChristineM has been working on (see [Figma Link](https://www.figma.com/file/35JfN0Sg4ltIIjM7jBDKMk/Navigation?node-id=2383%3A136325)). This work will most likely go out as part of phase 2 work, which includes changes in the L1 Nav.
- The changes in this PR to remove the suppression of the L0 menu can continue on and be made available.

PR #9274, opened 2022-08-19
feat(masthead-v2): remove the suppression of L0 menu when L1 is enabled

proeung:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/9271

### Description

- This PR removes the suppression of the L0/Masthead menu item when it's being stacked next to the L1 Nav.

### Changelog

**Removed**

- Removal of the `!l1Data` check

**Results**

- Menu items within the Global Masthead are now visible next to the L1 Nav on desktop, however, on mobile, the menu disappears and the L1 Nav items are now part of the Hamburger menu slide out (see attached).
- This is the current behavior of the Masthead + L1 and we need to determine if this is acceptable behavior for the first rollout of the Masthead v2 or do we need to come up with a different solution, which requires a change in the SOW, where we need to implement the new mobile design of the L1, which factors into the end of Q3 deadline.
- Further decisions will need to be discussed with stakeholders and designers.

**Masthead L0 + L1 on Desktop**
Screen Shot 2022-08-19 at 10 48 48 AM

**Masthead L0 + L1 on Mobile**
Screen Shot 2022-08-19 at 10 49 12 AM


oliviaflory:
@proeung this is great to see what will happen when both menus are exposed! We were just discussing this on the design working session and if we needed to address a solution for the September launch. Will you be sharing this with Joseph, Keegan and the other stakeholders?

proeung:
@oliviaflory Love it! We're on the same wavelength! 👍 I'm going to touch base with @kjwhitla to get his input on this and see if there's a solution that makes sense for the end of the Q3 launch.

proeung:
**Notes from the 08/23 "Navigation Stakeholder & Dev Weekly" meeting:**

- The visibility of the L1 Menu items within the hamburger menu on mobile is not a blocker for the end of the Q3 deadline.
- We're going to be following up with a new solution for displaying the L1 Nav on mobile using the accordion behavior that @JaniceChristineM has been working on (see [Figma Link](https://www.figma.com/file/35JfN0Sg4ltIIjM7jBDKMk/Navigation?node-id=2383%3A136325)). This work will most likely go out as part of phase 2 work, which includes changes in the L1 Nav.
- The changes in this PR to remove the suppression of the L0 menu can continue on and be made available.

emyarod (approved):
looks good to me!

jkaeser (approved):
LGTM!

Issue #9212, opened 2022-08-05
[Masthead v2]: Setup a Storybook preview of the L0/Masthead v2 and wrap under a feature flag


proeung:

proeung:
The work for this issue will be accounted for in this PR - https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/9275

Issue #9211, opened 2022-08-05
[Masthead v2.0.0]: Implement changes to the utility menu (top right-hand side)


proeung:
**Engineering info:**  
- Innovation Team JIRA Ticket: https://jsw.ibm.com/browse/ADCMS-2479
- DPO consulting engineer: @annawen1 

-----

### Description
- Apply the new design specs/token changes to the Header bar. See below for the list of areas to be adjusted.
  - search
  - login
  - contact us (chat)

### Additional information
- **Overview and scope of the engineering work should be included in the linked epic (#9208)**

### Design Specs
- Desktop - https://www.figma.com/file/35JfN0Sg4ltIIjM7jBDKMk/Navigation?node-id=1970%3A110816
- Mobile - https://www.figma.com/file/35JfN0Sg4ltIIjM7jBDKMk/Navigation?node-id=1970%3A111059

**New Design**
![Screen Shot 2022-08-09 at 3 03 22 PM](https://user-images.githubusercontent.com/1815714/183740535-061274a0-bb44-49dd-8830-1ee2d25df528.png)

**Current `<dds-cloud-masthead-container>` using data from the L0/Global**
![Screen Shot 2022-08-09 at 1 53 22 PM](https://user-images.githubusercontent.com/1815714/183725192-78628fe9-a391-44a0-90c4-999748d93c53.png)

### Acceptance criteria
 - [ ] Changes should align with the specs outlined by the design team

proeung:
@andy-blum Let's sync up when you're ready to work on this ticket.

PR #9275, opened 2022-08-19
feat(masthead-v2): Implement changes to the utility menu

proeung:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/9211 and https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/9212

### Description

- This PR applies the new design specs/token and UI changes to the utility menu that takes in elements from the Cloud Masthead design and merges them into Global Masthead, which will eventually be Masthead V2.

### Changelog

**New**

- `<dds-masthead-contact>` - contact/chat UI button used to launch the "Let's chat" module.
- `has-contact` - a prop/attribute that allows adopters to optionally add the contact/chat to the masthead utility menu

Screen Shot 2022-08-19 at 11 06 11 AM

###  To-do List
- [x] Move the `auth-method` prop to the Global Masthead composite that allows adopters to define `cookie`, `cloud-docs-api`, `profile-api`  for the Cloud authentication checking method.
  - Default value will be set to `profile-api` 
- ~Import the `cloudAccountAuthAPI` from the services package and add check the status in the return.~
- ~Expose these different authentication checking methods in a knob so that we can verify the regular Global Profile API and the ones coming from Cloud~
- [ ] Ensure proper documentation exists for authentication methods.
- [X] Move/copy the `ctaButtons` from the `cloud-masthead-composite.ts` to `masthead-composite.ts`
- [X] Copy any styles from the `cloud-masthead.scss` over to `masthead.scss`


proeung:
@andy-blum Please see the "To-do List" above for the remaining work for this PR.

andy-blum:
Following the updates to the Translate API in the services-store package (thanks, @annawen1!), the data-endpoint can now be set/modified in the knob for each masthead story. You may need to hard-reload one time to clear out cached resources.

Additionally, because of the way the user authenticated status knob works, the user status _state_ will always reset to "unauthenticated" when the dataendpoint is updated. At this point in time, that knob's _UI_ does not update.

Screen Shot 2022-08-23 at 10 55 25 AM

Screen Shot 2022-08-23 at 10 54 57 AM

proeung (commented):
@andy-blum The addition of the `endpoint` knob and optional (ghost-style) button looks good, however, I'm not seeing any reference/import to the `CloudAccountAuth` service which checks for the `cookie` and `API` auth methods.

We'll need to copy/take the auth methods from `CloudAccountAuth.js ` (https://github.com/huntermacd/carbon-for-ibm-dotcom/blob/master/packages/services/src/services/CloudAccountAuth/CloudAccountAuth.js#L10-L62) and consolidate that with the global `profile.js` (https://github.com/huntermacd/carbon-for-ibm-dotcom/blob/master/packages/services/src/services/Profile/Profile.js). @annawen1 Let me know if you agree.

Also, with the interest of getting some baseline changes (eg. SB preview, etc) into the `feat/masthead-v2` branch for demoing purposes, we can revisit the proper way to get all of the different auth methods (Profile, Cloud Cookie, Cloud API) in place. I'll follow up with a separate ticket and PR for this work.

annawen1:
> @andy-blum The addition of the `endpoint` knob and optional (ghost-style) button looks good, however, I'm not seeing any reference/import to the `CloudAccountAuth` service which checks for the `cookie` and `API` auth methods.
> 
> We'll need to copy/take the auth methods from `CloudAccountAuth.js ` (https://github.com/huntermacd/carbon-for-ibm-dotcom/blob/master/packages/services/src/services/CloudAccountAuth/CloudAccountAuth.js#L10-L62) and consolidate that with the global `profile.js` (https://github.com/huntermacd/carbon-for-ibm-dotcom/blob/master/packages/services/src/services/Profile/Profile.js). @annawen1 Let me know if you agree.
> 
> Also, with the interest of getting some baseline changes (eg. SB preview, etc) into the `feat/masthead-v2` branch for demoing purposes, we can revisit the proper way to get all of the different auth methods (Profile, Cloud Cookie, Cloud API) in place. I'll follow up with a separate ticket and PR for this work.

I think if we consolidate the `CloudAccountAuth.js` and `ProfileAPI`, that may require changing/creating new files in the services-store package. I agree we can handle that in a new PR so that those auth changes don't make this into a gigantic PR. One thing I want to explore keeping the cloud and profile services separate as is and just handle the logic of requesting either method in the composite file itself or at least making minimal changes to the type files so we don't have to go through creating new types/actions/reducers in the services-store.

proeung:
@annawen1 Sounds good! Here's the follow-up ticket (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/9291), which I have you added a consultant. If you can help review this PR that would be great as I'd love to get the baseline changes (eg. SB preview, etc) into the `feat/masthead-v2` branch.

Also, do you know if this CI Percy check needed to be approved before the PR can be merged down? The visual changes are in alignment with the V2 update work.

Screen Shot 2022-08-23 at 5 00 14 PM

annawen1 (approved):
LGTM!

annawen1:
> @annawen1 Sounds good! Here's the follow-up ticket (#9291), which I have you added a consultant. If you can help review this PR that would be great as I'd love to get the baseline changes (eg. SB preview, etc) into the `feat/masthead-v2` branch.
> 
> Also, do you know if this CI Percy check needed to be approved before the PR can be merged down? The visual changes are in alignment with the V2 update work.
> 
> Screen Shot 2022-08-23 at 5 00 14 PM

I went ahead and approved that required percy job, so there shouldn't be any issues merging it down now

packages/web-components/src/components/masthead/masthead.scss
```diff
@@ -43,6 +44,38 @@
   }
 }
 
+:host(#{$dds-prefix}-masthead-button-cta) {
+  @extend :host(#{$dds-prefix}-button-cta);
+
+  margin: 0;
+  padding-right: 0;
```
jkaeser, line 51: Nit:
```suggestion
  padding-inline-end: 0;
```

packages/web-components/src/components/masthead/masthead-contact.ts
```diff
@@ -0,0 +1,53 @@
...
+   * The `aria-label` attribute for the trigger button.
+   */
+  @property({ attribute: 'trigger-label' })
+  triggerLabel = 'Contact';
+
+  render() {
+    const { triggerLabel, _handleClick: handleClick } = this;
+    return html`
+      <a
+        tabindex="0"
```
jkaeser, line 38: I realize this is probably copied from the `DDSMastheadProfile` render method, but do we need `tabindex` on an `<a>`?

jkaeser (commented):
@andy-blum Looks great! As usual I left a few notes for perusal, but I didn't find any blocking issues.

packages/web-components/src/components/masthead/masthead-composite.ts
```diff
@@ -797,6 +810,18 @@ class DDSMastheadComposite extends LitElement {
...
+   * `true` if Contact us should be shown.
+   */
+  @property({ type: String, reflect: true, attribute: 'has-contact' })
+  hasContact = 'true';
+
+  /**
+   * The selected authentication method, either 'cookie' or 'api'.
+   */
+  @property({ attribute: 'auth-method' })
+  authMethod = 'profile-api';
```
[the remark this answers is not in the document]
andy-blum: Yeah, this seems like a good addition to the follow up. I've copied your text above to that issue.

PR #9277, opened 2022-08-22
fix(tabs-extended): move aria-selected to li elem with tab role

annawen1:
### Related Ticket(s)

[Tab Extended]: QA: Screen reader does not identify the currently selected tab #9230

### Description

Screen reader was not picking up the `aria-selected` attribute on the tab buttons. According to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-selected#description, the `aria-selected` attribute should be applied to elements with the `role="tab"` attribute, which in this case is the `li` element wrapping the `button`. 

Moving the `aria-selected` attribute to the `li` should hopefully work to have the screen reader pick it up.

Screen Shot 2022-08-22 at 11 22 52 AM

### Changelog

**Changed**

- move `aria-selected` to `li` element wrapping tabs-extended button

### DIFF
--- a/packages/web-components/src/components/tabs-extended/tabs-extended.ts
+++ b/packages/web-components/src/components/tabs-extended/tabs-extended.ts
@@ -223,14 +223,19 @@ class DDSTabsExtended extends StableSelectorMixin(LitElement) {
               'bx--tabs__nav-item--disabled': disabled,
             });
             return html`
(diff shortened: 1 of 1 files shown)


annawen1:
Reached out to Gopi to confirm if this fix works

Rasubra8:
@annawen1 
Issue is fixed. Now SR announces the first tab in the tablist as "selected".

Issue #8855, opened 2022-05-17
[Footer]: QA: Select geographic modal and language and location selection modal: No label for this modal dialog


Gopi916:
### Description

Page Element:
Select geographic modal and language and location selection modal

Issue:
There is no label for this modal dialog

Expected:
Provide appropriate label
The label can be given as Select geographic area for the first modal and language and location selection for the second one.

User Impact:
If the dialog does not have an appropriate label, screen reader users find it difficult to understand its purpose.

### Component(s) impacted

Footer (Default)

### Browser

Firefox

### Carbon for IBM.com version

v1.34.0-rc.0

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

Carbon for IBM.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://www.ibm.com/standards/carbon/web-components/?path=/story/

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [ ] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #9309, opened 2022-08-29
fix(locale-modal): add title to regions list

emyarod:
### Related Ticket(s)

#8855

### Description

Screen Shot 2022-08-29 at 1 58 45 PM

This PR adds a title attribute to the `dds-regions` component of the locale modal so that the header title is read by the screen reader even after the component auto focuses on the first region button

### Changelog

**New**

- `title` attribute on `dds-regions`


ariellalgilmore (approved):
LGTM! should we have Ram or Gopi check the ddeploy preview before merging?

emyarod:
> LGTM! should we have Ram or Gopi check the ddeploy preview before merging?

yes I think we should!

Rasubra8:
@emyarod 
The Issue is partially fixed. 
Screen reader announces extraneous text when we check the dialog label by pressing the key "insert+T". Along with the label, SR announces the following text. Note: The same issue is found in the second modal that we get after selecting a region. 
"Close modal United States — English Select geographic area
dialog" Please refer the screen shot attached for jaws output. 
Expected result:
Remove the extraneous announcement. And as mentioned earlier, the following label can be given for the dialog.
The label can be given as Select geographic area for the first modal and language and location selection for the second one that is displayed after selecting the region
Dialog 2 jaws output
Dialog 1 jaws output

emyarod:
@Rasubra8 I believe the fix for this may introduce a breaking change so this may need to be delayed until the next major version release

PR #9359, opened 2022-09-12
fix(masthead-v2): fix masthead-contact & add DDO for testing

proeung:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/9360

### Description

- This PR fixes the issue where the "Contact" CTA is not initializing the contact module's Let's Talk module when the `has-contact` attribute is set to `true`.

### Changelog

**Changed**

- Update the `Masthead` component CodeSandbox to include a test `DDO` and `cm-app.min.js` script so that we can verify that the behavior is working as expected.

https://user-images.githubusercontent.com/1815714/189696785-e00a7997-7c33-4322-ad6a-008eefacda45.mov

### Testing Instructions
- Visit the `Masthead` codesandbox app - https://webcomponents-codesandbox.s3-web.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/9359/masthead/index.html
- Click on the "Contact" CTA and ensure that the "Let's talk" widget opens.


jkaeser (commented):
This looks good from a code perspective. Waiting for the CodeSandbox example to do functional testing.

proeung:
CodeSandbox example deploy preview is ready for testing - https://webcomponents-codesandbox.s3-web.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/9359/masthead/index.html

cc: @jwitkowski79 @oliviaflory

jwitkowski79 (approved):
@proeung This looks good and is working as expected. Would love if @oliviaflory could also take a peek at it

oliviaflory:
@jkaeser @proeung mostly there!

I like that when you have a mega menu open and then select the contact us button, the mega menu collapses/closed. 
I think the reverse should also be true. When I select the Contact us button first, the contact module opens, and if I then select a mega menu the chat module stays open and obstructs part of the mega menu content

contact module mega menu Sep-13-2022 17-49-10


Questions that we should run by the accessibility team:
1. Should clicking the chat button again (after the chat module is opened) trigger the chat module to close? Asking because otherwise the user has to move their mouse to the contact module to close, and all other navigation items (mega menus and profile menu) close the menus on a second click. 
2. Should the focus state move to the Chat module once the user clicks it for keyboard users? 
   - When testing with the screen reader, there is no announcement that the chat module has opened
   - After you select the Contact us button, a keyboard user has to tab through the _entire_ page before getting to the contact module. This doesn't seem like a great experience (hence the suggestion to move the focus state). Shift tab would go back to the contact us button.

proeung:
@oliviaflory Thanks for your review! What you've outlined makes sense and is great for usability, however, it's out of scope from this issue (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/9360) and one that I'm not sure if it's feasible for Phase 1. 

For this early Oct launch, we're just trying to prioritize some of the big line item bugs seen not just within the components, but other bugs that arise in the platform integrations (eg. AEM, Drupal, etc) and this one will most likely go into Phase 2 work. I've created a new ticket to capture your feedback along with the accessibility concerns, which we can revisit for Phase 2. Also, as far as I'm aware the global masthead will be launched without the contact button enabled.

GH issue: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/9369
JIRA ticket: https://jsw.ibm.com/browse/ADCMS-2602

Issue #9364, opened 2022-09-12
[Masthead]: cc & lang params improperly set in the search form


jc-palacios:
### Description

We are setting the locale in both, `<html lang>` and `digitalData.page.pageInfo.language`, and still, the search form in our page is using `cc="us"` & `lang="en"`

### Component(s) impacted

Search box in the Masthead:

<img width="1440" alt="image" src="https://user-images.githubusercontent.com/86446190/189731253-529f5a2d-babe-442e-a2ac-d336a042f92f.png">

### Carbon for IBM.com version

1.19.0 (Production) & 1.38.0 (Development)

### Severity

Severity 1 = The design is broken in a critical way that blocks users from completing tasks or damages the brand. Affects major functionality, no workaround.

### Application/website

https://www.ibm.com/fr-fr/products

### Package

@carbon/ibmdotcom-react

### CodeSandbox example

https://www.ibm.com/standards/carbon/react/?path=/story/components-masthead--default

### Steps to reproduce the issue (if applicable)

1. Visit the Products homage page in any locale other than us-en, for example: https://www.ibm.com/fr-fr/products
2. Expand the search form in the masthead (L0), by clicking on the magnifying glass, do any search. You should be redirected to the MH search results page.
3. We are expecting to see results on the locale we were before, like fr-fr, but instead, we're seing results for us-en (`lang=en&cc=us`)

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #9372, opened 2022-09-14
fix(masthead): search locale fix

IgnacioBecerra:
### Related Ticket(s)
#9364

### Description
This PR ensures that any placeholderText can be replaced by the data coming from the TranslationsAPI, as before it just used the default English value no matter what. This PR also fixes the problem where the React Masthead Search would only redirect to `us-en` locale instead of the proper one.

### Changelog

**Changed**

- how the default placeholder text gets replaced
- the React masthead search now uses the proper locale when redirecting


kennylam (approved):
LGTM!

Issue #8977, opened 2022-06-10
[Content block mixed]: Link list icon misaligned in React wrapper


RichKummer:
### Description

In mobile Safari using React wrapper, the icon is still centered in the `Link list` when used in `Content block mixed`. This should match the update here: https://app.zenhub.com/workspace/o/carbon-design-system/carbon-for-ibm-dotcom/issues/8827

<img width="470" alt="Screen Shot 2022-06-10 at 2 03 53 PM" src="https://user-images.githubusercontent.com/15643582/173125311-f4fc2ad3-6856-4062-9981-b4d7e9c5520e.png">

### Component(s) impacted

Content block mixed
Link list

### Browser

Safari

### Carbon for IBM.com version

v1.20.0-rc.0

### Severity

Severity 4 = The problem is not visible to or noticeable to an average user. Affects minor functionality, no workaround needed.

### Application/website

C4IBM.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://codesandbox.io/s/github/carbon-design-system/carbon-for-ibm-dotcom/tree/main/packages/web-components/examples/codesandbox/components-react/content-block-mixed

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #9405, opened 2022-09-21
fix(link-list): ensure icon stays on top

IgnacioBecerra:
### Related Ticket(s)
#8977

### Description
There was a small leftover where the Link list icon doesn't stay at the top when there is more than two lines. 

### Changelog

**Changed**

- removed `padding-top` of `card-cta-footer` when slotted within `list-list-item-card-cta` at `md` or less breakpoint


kennylam (approved):
LGTM!

PR #9408, opened 2022-09-23
feat(cta-block): react wrapper

ariellalgilmore:
### Related Ticket(s)

#8633 

### Description

react wrapper with storybook, readme, and codesandbox

### Changelog

**New**

- story wrapper, codesandbox, readme


IgnacioBecerra:
@ariellalgilmore Looking good! Just a couple things:

1. The heading knob doesn't seem to work, but I'm linking a workaround 
2. Seems like the border isn't showing up when enabled?
3. For With Content items, the Buttons content item type aren't showing up
4. With Link list and tabs, the border is always enabled

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/9407/files#diff-4ac2bb3580c248c5605af672a5a16fc781257362cc62c555ce8f45d4788325d0R41-R45

ariellalgilmore:
> 1. https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/9407/files#diff-4ac2bb3580c248c5605af672a5a16fc781257362cc62c555ce8f45d4788325d0R41-R45

Thanks @IgnacioBecerra !! will update shortly, but one thing it seems like we are actually purposely filtering out so users shouldn't use a button-group here: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/7782
Wondering if we had the option to show that it doesn't work? lol cc: @oliviaflory @RichKummer

IgnacioBecerra (approved):
LGTM code-wise, thanks @ariellalgilmore! Let's see what design says about removing the button from the knobs 👍

ariellalgilmore:
> LGTM code-wise, thanks @ariellalgilmore! Let's see what design says about removing the button from the knobs 👍

thanks for checking again! just updated to removing the button-group option for content-items and adding it instead for the cta option of CTA Block

RichKummer (approved):
LGTM! Thanks for removing the button option for the `Content items` @ariellalgilmore !

Issue #9391, opened 2022-09-19
[Filter Panel]: QA: Screen reader announces all the sub-options as "expanded/collapsed" in Technologies and Business needs


gopinathan-n:
### Description

Page Element:
Sub options present in the following filter options: 
Technologies
Business needs

Issue Description:
Screen reader announces all the sub-options in the mentioned filters as "expanded/collapsed". Since all do not behave like expand /collapse, this announcement would confuse the user. 

Expected:
Remove aria-expanded attribute for the options within the filters that do not "expand/collapse". aria-expanded attribute should be present only for the elements that behave like expand/collapse e.g. Artificial intelligence, etc.

User Impact:
When the state of an element i.e. disabled, selected, expanded, currently active etc is not conveyed for the screen reader user, the blind user can be mislead. this impacts the user interaction.

### Component(s) impacted

Filter Panel

### Browser

Safari, Firefox

### Carbon for IBM.com version

v1.38.0-rc.0

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

Carbon for IBM.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

Web component Canary- https://carbon-design-system.github.io/carbon-for-ibm-dotcom/canary/web-components Web component Staging- https://carbon-design-system.github.io/carbon-for-ibm-dotcom/next/web-components

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [ ] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #9435, opened 2022-09-28
fix(filter-panel-input-select): aria expanded

ariellalgilmore:
### Related Ticket(s)

#9391 
### Description

Removes `aria-expanded` attribute for the options within the filter panel that do not "expand/collapse". aria-expanded attribute should be present only for the elements that behave like expand/collapse e.g. Artificial intelligence, etc.

### Changelog

**Changed**

- checks which filter panel input has items and adds `aria-expanded` to only the ones with items


IgnacioBecerra (approved):

gopinathan-n:
Now the unnecessary aria-expanded attributes are removed

Issue #9279, opened 2022-08-23
[Audio Player]: QA: Screen reader announces incorrect label for Playback Rate Speed and Closed captions Menu


gopinathan-n:
### Description

Page Element: Playback Rate Speed and Closed captions Menu

Issue Description:
Before we activate the "Play" button, These menus are announced as "disabled". After activating the "Play" button, these are announced as "open and close list of options".

Expected:
Provide appropriate label. E.G. 
1. Playback Rate Speed
2. Closed captions

User Impact:
If the label is incorrect or inadequately descriptive, it affects the screen reader user’s ability to understand the purpose of the control. 

### Component(s) impacted

Audio Player

### Browser

Safari, Firefox

### Carbon for IBM.com version

v1.37.0-rc.0

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

Carbon for IBM.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://carbon-design-system.github.io/carbon-for-ibm-dotcom/canary/react-experimental/?path=/story/components-audio-player--default

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [ ] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #9452, opened 2022-10-05
fix(audio-player): allow localized labels for rewind/forward/caption buttons

kennylam:
### Related Ticket(s)

#9279 

### Description

This change allows authors to include localized labels for the rewind, fast-forward, and closed captions buttons.

### Changelog

**New**

- props to allow user-defined labels for rewind, forward, and closed captions buttons


RichKummer:
It sounds like the label read out is right for `Forward` and `Rewind`, but I am seeing that the `Closed captioning` button is announced as "open and close list of options" when using VoiceOver

kennylam:
> It sounds like the label read out is right for `Forward` and `Rewind`, but I am seeing that the `Closed captioning` button is announced as "open and close list of options" when using VoiceOver.

@RichKummer this behavior is coming from the core Carbon component. The title displays correctly (on hover) but screen reader does not pick it up. We'll probably need to see if an issue is open on their end, and if not we can create one.

RichKummer (approved):

Issue #8331, opened 2022-02-17
[Content block segmented]: Text CTA alignment


RichKummer:
### Description

In mobile Safari using a text link style, the text link will align with the browser border without any spacing on the left side of the screen.

![IMG_2037](https://user-images.githubusercontent.com/15643582/154530690-ee1d9788-4bbc-4142-99c0-7d1bb5978523.PNG)

**Content block simple**
![IMG_2038](https://user-images.githubusercontent.com/15643582/154531202-29f58bf0-be20-4229-80b4-b816b87e5de9.PNG)

### Component(s) impacted

Content block segmented
Content block simple

### Browser

Safari

### Carbon for IBM.com version

v1.31.0-rc.1

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

C4IBM.com

### Package

@carbon/ibmdotcom-react

### CodeSandbox example

https://codesandbox.io/s/github/carbon-design-system/carbon-for-ibm-dotcom/tree/main/packages/react/examples/codesandbox/components/ContentBlockSegmented

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #9453, opened 2022-10-05
fix(content-block): fix cta alignment for mobile viewport

kennylam:
### Related Ticket(s)

#8331 

### Description

Content block CTA should align to the left of the viewport in mobile only when `cta-type=card`.

before:
image

after:
image

### Changelog

**Changed**

- update `content-block` CTA alignment for mobile viewport

### DIFF
--- a/packages/styles/scss/internal/content-block/_content-block.scss
+++ b/packages/styles/scss/internal/content-block/_content-block.scss
@@ -237,9 +237,6 @@
 
     @include carbon--breakpoint('sm') {
       .#{$prefix}--card__CTA {
-        // @include hang;
-        // padding-left: 0;
-        // padding-right: 0;
         margin-left: -($carbon--spacing-05);
       }
     }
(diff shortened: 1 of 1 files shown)


packages/styles/scss/internal/content-block/_content-block.scss
```diff
@@ -233,19 +233,23 @@
...
-    max-width: calc(320px + 1rem);
+    max-width: calc(320px + #{$carbon--layout-05});
 
     @include carbon--breakpoint('sm') {
-      padding-left: 0;
-      padding-right: 0;
+      .#{$prefix}--card__CTA {
+        // @include hang;
+        // padding-left: 0;
+        // padding-right: 0;
```
ariellalgilmore, line 242: should we remove these comments?
```suggestion
```

kennylam: @ariellalgilmore Oops, yes good catch!

Issue #8853, opened 2022-05-17
[Masthead]: Screen reader announces the landmark label as "Nav LTR" which is not meaningful


Gopi916:
### Description

Page Element:
nav-ltr navigation region

Issue:
Screen reader announces the landmark label as "Nav LTR" which is not meaningful.

Expected:
As of now, Since there is only one navigation region present here, the label is not required. However If another navigation region gets added in the future , then a unique label may be required for both.

User Impact:
If the label is incorrect or inadequately descriptive, it affects the screen reader user’s ability to understand the purpose of the element.

### Component(s) impacted

Masthead
Default
With custom type ahead
Search open on load
With L1
With alternate logo and tool tip

### Browser

Firefox

### Carbon for IBM.com version

v1.34.0-rc.0

### Severity

Severity 4 = The problem is not visible to or noticeable to an average user. Affects minor functionality, no workaround needed.

### Application/website

Carbon for IBM.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://www.ibm.com/standards/carbon/web-components/?path=/story/

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [ ] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #9480, opened 2022-10-11
fix(masthead): remove unnecessary aria-label

annawen1:
### Related Ticket(s)

[Masthead]: Screen reader announces the landmark label as "Nav LTR" which is not meaningful #8853

### Description

the `top-nav` aria-label is unnecessary and should be removed to not confused the user.

### Changelog

**Removed**

- `aria-label` from `top-nav`


ariellalgilmore (approved):
LGTM!

gopinathan-n:
Now the unnecessary labels are not announced by the screen reader

PR #9563, opened 2022-10-12
feat(card-group): switch to controls

ariellalgilmore:
### Related Ticket(s)

#9490 
### Description

Update card-group to use controls for wc and react wrapper

Note: previously CTA would not be an option if the media toggle was `true` with the new storybook controls, I am not sure how to replicate this.


kennylam (approved):
LGTM!

PR #9568, opened 2022-10-17
fix(storybook): improve feature flags granularity

jkaeser:
### Related Ticket(s)

none

### Description

The way experimental components' feature flags work was changed at some point after updating Storybook. The current behavior is an all-or-nothing approach where the presence of any feature flag as an environment variable reveals every experimental component.

This PR restores the ability to hide specific components by iteratively building up the CSS that hides components from the sidebar list.

### Changelog

**Changed**

- Restores ability to selectively hide/show individual experimental components in Storybook.


ariellalgilmore (approved):
LGTM! Thanks for this update. @jkaeser would you possibly want to make the same changes for the react package?

jkaeser:
> LGTM! Thanks for this update. @jkaeser would you possibly want to make the same changes for the react package?

Sure can, @ariellalgilmore! I'll add that soon.

RichKummer (approved):

Issue #8841, opened 2022-05-16
[cta]: explore removing standalone versions in this component or component itself


kennylam:
Currently the `cta` component has the following variations:
- `text`
- `button`
- `card`
- `card-link`
- `feature`

These components are essentially duplicated in their standalone and `cta` versions. Since `cta` are already included in components which accept them, simply choosing the `type` could be sufficient, using only the `cta` class logic for presentation. This could remove the `cta` component as a standalone component.

Alternatively, the standalone components could be hidden and accessed via `cta` only. In my opinion the first choice makes more sense.

PR #9569, opened 2022-10-18
feat(cta): merging components into one

IgnacioBecerra:
### Related Ticket(s)
 #8841

### Description
This PR merges all the CTA components into one

### Changelog

**Changed**

- CTA is now an individual component that contains the other CTA components within


ariellalgilmore:
@IgnacioBecerra this is looking awesome! 

We will probably need to update this file here too: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/tools/get-rollup-config.js#L108-L110.

And looks like some tests need some updating since the storybook changed!

oliviaflory:
@IgnacioBecerra This looks nice from what I can tell from a design perspective :D 

A few questions:
1. Based on the decision to go with Option 1 stated in #8841: Should the CTA story be removed? Or are we keeping that to show how the different CTAs variations? 
2. Do we need to update the functional specs to state that the "CTA" is no longer pulling from a CTA component?

IgnacioBecerra:
Thanks for the review Olivia!
1. I feel like we should keep the CTA story since it is still a component that adopters can use, and they might not know about it if it's not there. 
2. Hm I don't think I understand this question, all of these CTA component variations are under the same component with this new PR.

oliviaflory:
@IgnacioBecerra I think I may have mis-understood the direction when I first read the issue. All good!

oliviaflory (approved):
LGTM!

IgnacioBecerra:
@oliviaflory Ah, gotcha! Yeah we're gonna need to, we'd only need a single functional spec for the CTA component. The different types should be denoted in them.

PR #9571, opened 2022-10-19
fix(storybook-react-wrapper): props not appearing

ariellalgilmore:
### Related Ticket(s)

N/A
### Description

Components that extend other components were not showing props in the storybook `docs` tab. 

Note: the first time the react wrapper deploy preview ran it failed, but the second time it passed. not sure if it will be finnicky when running on github pages

Screen Shot 2022-10-19 at 7 53 05 AM

### Changelog

**New**

- Added to `babel-create-react-custom-element-type` to check if parentProps exist and if they do copies and pastes props into current component `es` files

### DIFF
--- a/packages/web-components/.storybook/react/main.js
+++ b/packages/web-components/.storybook/react/main.js
@@ -166,6 +166,7 @@ module.exports = {
             ],
           },
         },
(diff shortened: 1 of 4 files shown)


emyarod (approved):
looks good to me!

PR #9570, opened 2022-10-18
feat(masthead-v2): add GA tracking attributes

andy-blum:
### Related Ticket(s)

[ADCMS-2649](https://jsw.ibm.com/browse/ADCMS-2649)

### Description

Adds Google Analytics tracking attributes to the various masthead components

Screen Shot 2022-10-18 at 11 56 22 AM

### Changelog

**Changed**

- Tracking attributes added to masthead link & button elements

PR #9573, opened 2022-10-19
test(masthead-v2): add e2e tests for masthead v2

andy-blum:
### Related Ticket(s)

Followup to #9570 

This PR branches off of the above PR, and should not be added in first.

### Description

- Duplicates & adapts the masthead (v1) e2e tests to cover masthead (v2).
- Adds a desktop/mobile test to check for Google Analytics attributes

### Changelog

**New**

- Add masthead-v2 test coverage


jkaeser (approved):
@andy-blum Looks like you need to run the formatting script. Otherwise LGTM!

RichKummer:
Hey Andy, I took a look at the Percy snapshot and it looks like the PR updated the type token for the menu headers. Sharing here so you can see:

PR 9573 - Masthead type token update

andy-blum:
@RichKummer Those changes are definitely not coming from this PR - we'll need to investigate where that's coming from, but those screenshots _also_ don't match up with the V2 masthead data.

Issue #9582, opened 2022-10-20
Global Masthead Performance Issues: Potentially Degrading User Experience


KeyboardCowboy:
### Description

The Web Components Masthead loads many scripts and there is concern that it is causing a degraded user experience.  We'd like to explore ways to improve load times and/or component weight.

This was brought to light when mobile pages started crashing on iOS and iPadOS the week of October 16.  It turns out there was a significant memory leak in another script that was determined to be the cause, but the masthead was the most recent significant addition to it took the blame.  Still, it brought to light the size of the masthead so we'd like to investigate any performance improvements we could make to help keep the pages as performant as possible.

### Component(s) impacted

Global masthead.

### Browser

Safari

### Carbon for IBM.com version

latest

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

ibm.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

na

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

juanleon9:
We have new information on this issue, the navigation code creates 2/3 of the total number of nodes on IBM pages, which is about 22,000 nodes on both DOM and shadow DOM.

Lighthouse flags pages with DOM trees that:
- Warns when the body element has more than ~800 nodes.
- Errors when the body element has more than ~1,400 nodes.

Attached is a snapshot Lighthouse for https://ww.ibm.com/us-en, which shows 5,255 in the normal DOM, not accounting for the shadow DOM which is over three times larger
<img width="1190" alt="image" src="https://user-images.githubusercontent.com/21178525/197245935-98ed834d-7e79-4b52-a6ba-bc80ac06569f.png">

My recommendation is to backout the new navigation system because of the impact to SEO and paid search and also to overall page performance

KeyboardCowboy:
How many nodes did the previous navigation implement?  Are we net + or - with the new navigation?  Also, can we quantify the impact on page performance with some sort of AB test?  The initial issue wasn't load times, but memory usage where the scrips the continued running after the page loaded ate up over 7GB of system memory.

[...]

PR #9585, opened 2022-10-21
fix(masthead): lazy loading megamenu and side nav

IgnacioBecerra:
### Related Ticket(s)
#9582

### Description
Currently the Masthead loads both the SideNav and Megamenu components at the same time, even if they haven't been opened yet. This causes the amount of nodes to increase considerably right away. In order to mitigate this, a fast solution is to dynamically load the required components *only* when the user has interacted with the menus.

This means that the mobile version wouldn't be loading the Megamenu, and the desktop wouldn't load SideNav until/if required. 

Another optimization that was done ensures that the Megamenu component wouldn't get loaded into the DOM unless a user has clicked on the menu beforehand. Then and only then, the megamenu would be loaded in. Once the menu is closed, the component gets removed from the DOM to ensure the total node count doesn't shoot up.

Using @juanleon9's NodeCount script on the `dds-masthead-container` (which contains both `dds-side-nav` and `dds-megamenu`) these are the following stats:

* Megamenu and SideNav not loaded:
 `NODES 1319 COMMENTS 316 INTERS 671`

* Megamenu loaded in and opened / SideNav not loaded (essentially Desktop view):
  `NODES 2445 COMMENTS 553 INTERS 1224`

The megamenu would go be removed from the DOM right after it closes, so the Masthead will always remain at around `1319`

Seems like the total nodes each menu brings in is about:
* Megamenu: `2744`
* SideNav: `4021`

At least with this PR we're saving about `6765` nodes from populating the page, at least right away. 

### Changelog

**Changed**

- lazy loading `side-nav` and `megamenu` components only when user interacts with them


packages/web-components/src/components/masthead/megamenu-top-nav-menu.ts
```diff
@@ -134,6 +134,14 @@ class DDSMegaMenuTopNavMenu extends DDSTopNavMenu {
         forEach(doc.querySelectorAll((this.constructor as typeof DDSMegaMenuTopNavMenu).selectorOverlay), item => {
           (item as DDSMegaMenuOverlay).active = this.expanded;
         });
+
+        import(`./megamenu-left-navigation`);
+        import('./megamenu-category-link');
+        import('./megamenu-category-group');
+        import('./megamenu-category-group-copy');
+        import('./megamenu-link-with-icon');
+        import('./megamenu-overlay');
```
andy-blum, line 143: We'll probably need to add a way to track once these have been imported, so we don't try to re-import everytime we update the 'expanded' property on a top-nav-menu item

jeffchew:
perhaps there's a babel config for this? 
image

andy-blum:
https://babeljs.io/docs/en/babel-plugin-syntax-dynamic-import

> NOTE: This plugin is included in @babel/preset-env, in [ES2020](https://github.com/tc39/proposals/blob/master/finished-proposals.md)

packages/web-components/src/components/masthead/megamenu-top-nav-menu.ts
IgnacioBecerra: Good idea, I added a boolean check before importing the components.

IgnacioBecerra:
After discussing with @andy-blum, we agreed an even greater optimization would be to render only one megamenu at a time, and remove it from the DOM once it is closed. 

With this new method, we are now down to `NODES 1319 COMMENTS 316 INTERS 671` whenever a megamenu is closed, and only `NODES 2445 COMMENTS 553 INTERS 1224` if a megamenu is open.

andy-blum:
@IgnacioBecerra Can you rebase this onto `feat/masthead-v2` to pull in the v2 story & other updates?

andy-blum:
I had too many problems rebasing this, so I manually duplicated the changes to a new PR, #9602

IgnacioBecerra:
Closing this PR in favor of Andy's remade PR.

PR #9597, opened 2022-10-26
fix(storybook): react wrapper props not rendering

ariellalgilmore:
### Related Ticket(s)

N/A
### Description

Components that extend other components were not showing props in the storybook `docs` tab. 

Screen Shot 2022-10-19 at 7 53 05 AM

### Changelog

**New**

- Added to `babel-create-react-custom-element-type` to check if parentProps exist and if they do copies and pastes props into current component `es` files


packages/web-components/tools/babel-plugin-create-react-custom-element-type.js
```diff
@@ -209,6 +210,69 @@ function createMetadataVisitor(api) {
       }
 
       const metadata = getPropertyMetadata(path);
+
+      // Checks if the current component extends a local dotcom component
+      if (context.parentDescriptorSource && !context.parentDescriptorSource.includes('carbon-web-components')) {
```
emyarod, line 215:
```suggestion
      if (!context?.parentDescriptorSource?.includes('carbon-web-components')) {
```

would something like this work as well?

emyarod (approved):
looks good to me!

packages/web-components/tools/babel-plugin-create-react-custom-element-type.js
ariellalgilmore: ahh i thought it would too, but it then passes if a component does not have a `parentDescriptorSource`. gonna revert back

RichKummer (approved):
LGTM!

PR #9602, opened 2022-10-28
V2 megamenu lazy load

andy-blum:
### Related Ticket(s)
#9582

Replaces #9585 

### Description
Currently the Masthead loads both the SideNav and Megamenu components at the same time, even if they haven't been opened yet. This causes the amount of nodes to increase considerably right away. In order to mitigate this, a fast solution is to dynamically load the required components *only* when the user has interacted with the menus.

This means that the mobile version wouldn't be loading the Megamenu, and the desktop wouldn't load SideNav until/if required. 

Another optimization that was done ensures that the Megamenu component wouldn't get loaded into the DOM unless a user has clicked on the menu beforehand. Then and only then, the megamenu would be loaded in. Once the menu is closed, the component gets removed from the DOM to ensure the total node count doesn't shoot up.

Using @juanleon9's NodeCount script on the `dds-masthead-container` (which contains both `dds-side-nav` and `dds-megamenu`) these are the following stats:

* Megamenu and SideNav not loaded:
 `NODES 1319 COMMENTS 316 INTERS 671`

* Megamenu loaded in and opened / SideNav not loaded (essentially Desktop view):
  `NODES 2445 COMMENTS 553 INTERS 1224`

The megamenu would go be removed from the DOM right after it closes, so the Masthead will always remain at around `1319`

Seems like the total nodes each menu brings in is about:
* Megamenu: `2744`
* SideNav: `4021`

At least with this PR we're saving about `6765` nodes from populating the page, at least right away. 

### Changelog

**Changed**

- lazy loading `side-nav` and `megamenu` components only when user interacts with them


jkaeser:
I'm getting some inconsistent results loading this component fresh with no cache. Sometimes the first tab of the megamenu fails to load in time for render:

https://user-images.githubusercontent.com/12532727/198680176-78ecc8fe-99fc-4fc5-badd-855a1486b974.mov

IgnacioBecerra (approved):

andy-blum:
Video demonstrating node count during masthead/left-nav usage and changes.


https://user-images.githubusercontent.com/25532785/198726269-448631f2-44bf-48a4-8227-aa7b0357e577.mov

andy-blum (commented):
```javascript
        if (!this._importedSideNav) {
          import('./left-nav-cta-item');
          import('./left-nav-name');
          import('./left-nav-menu');
          import('./left-nav-menu-section');
          import('./left-nav-menu-item');
          import('./left-nav-menu-category-heading');
          import('./left-nav-overlay');
          this._importedSideNav = true;
        }
```
@IgnacioBecerra I wanted to do these in parallel, but I couldn't figure out how to make typescript ok with `Promise.allSettled()`. Any ideas on that?

IgnacioBecerra:
@andy-blum Maybe having a separate `megamenuImport.js` file of sorts and only do `await import ('megamenuImports.js')` instead? Though not sure if that would solve the TS problem you're facing. 🤔

jkaeser (approved):
Looks superb. Excellent work, @andy-blum!

Issue #9561, opened 2022-10-12
[Spacing] POC for Spacing logic


oliviaflory:
### User Story
As a Carbon for IBM.com developer I need to create a proof of concept of the spacing logic proposal.

### Additional information
To support the explorations of the Spacing logic epic, we want to create a demo of the spacing logic proposal that the designers shared in the team huddle. This will help to see how our components look with consistent spacing applied throughout a page and will help the designers better define the rules for spacing.
- Figma of [Spacing audit](https://www.figma.com/file/L7aEMOvkO2YpAoxec26WuG/Audit-spacing?node-id=2511%3A2877)
- Figma to [Spacing proposal](https://www.figma.com/file/L7aEMOvkO2YpAoxec26WuG/Audit-spacing?node-id=3901%3A28415) we decided on during Huddle

Suggestion: dump a bunch of our components into a demo page to see how the spacing works.

#### Tasks:
- [ ]

oliviaflory:
## Ideas

### Content block 1
Recreate Content block mixed using modular components:
- **Parent**
- Content block
   - **Children**
   - Content group cards
   - Content group pictograms
   - Content group simple
      - image with caption as a child to content group simple

### Content block 2
Recreate AEMs custom layout:
- Parent
- Content block
   - Child
   - Content item horizontal with featured media
   - Content item horizontal with media
   - Content item horizontal 
   - Card group (5 or 6 cards)

### Content block 3
Recreate AEMs custom layout:
- **Parent**
- Content block
   - **Children**
   - Callout quote
   - Card group with images (3 cards)

oliviaflory:
Review 10/26

### Padding vs margin
Padding = internal / does not collapse
Margin = external / does collapse

Content x are using margin and some are using padding

- Content block
- Content group
- Content item
- Content section

RichKummer:
Looked through the WC storybook and took a snapshot of whether the component overhangs or not. The overhang was most commonly from cards (even if optional), and bottom borders that extended into the gutter.

### Overhang

- Card group
- Card in card
- Card link
- Card
- Carousel (when using cards)
- Content block cards
- Callout quote (currently)
- Callout with media (currently)
- Content block media (Card link)
- Content block mixed (Cards + Card link)
- Content block segmented (Card link)
- Content block simple (Card link optional, also border overhangs)
- Content block (Card link)
- Content group simple (Card link)
- Content group (Card link)
- Content group cards (Card)
[...]

PR #9604, opened 2022-10-28
fix(content-group/block): missing display grid styles

ariellalgilmore:
### Related Ticket(s)

#9561  (found when looking at the spacing issue)
### Description

When testing content-group using CDN's in a codesandbox. 
y 
This first image shows how it appears without `display: grid`. Not properly containing the inner components and causing it to collapse on top of components below.
Screen Shot 2022-10-28 at 9 42 13 AM

This second image shows how it appears with `display: grid`
Screen Shot 2022-10-28 at 9 42 21 AM
### Changelog

This does cause a lot of percy changes, but happy to discuss more about the issue here


kennylam (approved):
LGTM thanks @ariellalgilmore!

PR #9605, opened 2022-10-28
fix(masthead): avoid inline styles for CSP compliance

### LINKED ISSUE #9599: Content Security Policy: remove inline-style from cloud-masthead
### Description

Can inline-style in cloud-masthead get moved to a style sheet?

Cloud platform UIs are required to implement changes for FS Cloud Compliance.

One of the requirements is to set the `csp` header. 

By default `csp` blocks inline styles. The solution is for these to be moved to a style sheet.

The `cloud-masthead` has an instance of inline style that will be blocked by csp (see image).
`<div class="bx--header__nav-content" style="left: -0px">`

image

### Component(s) impacted

The `<dds-cloud-masthead-container>` is affected.

image

### Browser

Chrome

v1.20.0

### Severity

Severity 1 = The design is broken in a critical way that blocks users from completing tasks or damages the brand. Affects major functionality, no workaround.

### Application/website

https://ondeck.console.cloud.ibm.com/docs

### Package

### CodeSandbox example

N/A

_No response_

_No response_

### Code of Conduct

andy-blum:
## Related Ticket(s)

Closes #9599 

## Description

Due to the content-security-policy header, browsers will begin to [block inline scripts and styles](https://content-security-policy.com/examples/allow-inline-style/). To keep the masthead's `<top-nav>` component usable when the number of items would over flow the container, we need a way to handle offset styles _dynamically_ without using inline styles or regular `<style>` elements on the page.

This PR adds a new mixin that creates a `<style>` element with the necessary `nonce` attribute, if available. The mixin provides the below functionality

### On ConnectedCallback

When the component connects, it checks `window.nonce` for the nonce string for this exact pageload, and stores it as `this._nonce` for later use

### Render functions

Components implmenting this mixin should _not_ have their own render function, but should instead use the abstract `renderContents()` method. This mixin renders the dynamic stylesheet along with the returned template from renderContents(). If a component _must_ implement `render()`, it must also run `renderDynamicStyles()` within that render function.

### Style getter/setter methods

- `setStyleBySelector(selectorString, styleProperty, styleValue) { ... }`
  "upserts" a style rule, either creating it for the first time or overriding the style for the given selector
  
- `getStyleBySelector(selectorString, styleProperty) { ... }`

### The Stylesheet

The rendered `<style>` element will not update it's contents, but because we're using CSSOM methods, the styles _do_ apply, and they _do_ cause change on the page. Users should never attempt to set `innerHTML` of the element, as that will cause the element to "reset" and discard all programatic changes made using the setter method above.

### Changelog

**Changed**

- Moved DDSTopNav inline styles to an adoptedStylesheet


andy-blum:
New approach, documented in PR summary.

packages/web-components/src/globals/mixins/csp-compliance.ts
```diff
@@ -0,0 +1,118 @@
...
+        newStyle.classList.add('csp-compliance-styles');
+        newStyle.innerHTML = `
+          :root {
+            --color: darksalmon;
+            background: var(--color);
+          }
+
+          body {
+            padding: 1rem;
+          }
```
IgnacioBecerra, line 50: I think these might be leftover styles we could remove? Unless they have a purpose I'm not seeing 🤔

IgnacioBecerra (commented):
Looking good code-wise, thought I don't see the WC Deploy Preview up, seems like some tests need fixing first.

packages/web-components/src/globals/mixins/csp-compliance.ts
andy-blum: Oops

andy-blum:
Feeling pretty good about this third approach.

ariellalgilmore:
Getting close! the tests are failing though and just double checking it on the deploy preview I see it not functioning totally correct where it gets stuck at this point
https://user-images.githubusercontent.com/20210594/200009522-07b77395-0117-453b-89dd-aa10a183890a.mov

IgnacioBecerra (approved):

Issue #9374, opened 2022-09-16
[Search with Typeahead]: beingredirected event not fired when hitting ENTER while in the search input


jc-palacios:
### Description

As an adopter of the Masthead/Search with Typeahead component, I want to be able to customize the search form, to be able to  "listen" for search inquires and do an on-demand search and present results to users without leaving the current page. Normally, the search component will redirect users to the MH results page; but we are in fact the owners of the MH results page, and we need to keep using the search box to start new searches without redirecting the user each time.

We are implementing the web-components version of the MH component, and subscribing to the `dds-search-with-typeahead-beingredirected` event (where we can cancel its default action and call our api). The problem we're facing at the moment is when the user hits the ENTER key: the event is not triggered as expected and the form gets submitted.

### Component(s) impacted

Search with Typeahead

### Carbon for IBM.com version

v1.21

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

https://www.ibm.com/search

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://www.ibm.com/standards/carbon/web-components/react/iframe.html?id=components-masthead--search-open-onload

### Steps to reproduce the issue (if applicable)

1. Open the MH example page: https://www.ibm.com/standards/carbon/web-components/react/iframe.html?id=components-masthead--search-open-onload
2. Open the Console and paste this code: `document.addEventListener('dds-search-with-typeahead-beingredirected', (e) => { e.preventDefault(); alert('event default prevented'); });`
3. Type something on the search box, then click on the magnifying glass icon on the right, you should see an alert window
4. Discard the alert, now click on the input to show the typeahead and click on any item (alternatively, you can use the keyword to choose an option and hit enter to select it), as with the previous step, you should see the same alert window
5. Finally, discard the alert and click on the input again, update your input and hit ENTER (without selecting any typeahead option).
6. As with the previous 2 steps, we are expecting to see the same alert, since the form is being submitted and we want to cancel that action. Instead, the form gets submitted (user is redirected to the MH search results page)

### Release date (if applicable)

First week of October

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
[...]

PR #9617, opened 2022-11-02
chore(search): allow option to stop redirection with enter

IgnacioBecerra:
### Related Ticket(s)
#9374

### Description
The Search with Typeahead component will now be dispatching an event upon the user hitting the enter key right before being redirected to the search results page. This is done to encompass all possible redirections coming from the component, in order for any adopters to intercept and stop the redirection if desired.

Due to the fact that the component uses a `form`, pressing enter within will always activate the redirection no matter what, using the `action` attribute, unless using an event listener to stop it.

### Changelog

**New**

- `dds-search-with-typeahead-beingredirected` being fired upon hitting enter now


RichKummer (approved):

kennylam (approved):
LGTM!

PR #9638, opened 2022-11-11
fix(videoplayer analytics): extended custom metrics data detection

tweenn:
### Description

This is an update to enhance flexibility on newer player-functionalities analytics-tracking portion.

AKA instead of having to update the player analytics here in carbon every time a new functionality is built on the mediacenter player, like our new was this helpful CTA, the analytics is flexible enough to allow the execPathReturn / playerstate and custometrics data to be auto-detected and sent to DBDM/CDS

### Changelog

**New**

**Changed**

- Services/Analytics/VideoPlayerAnalytics
    - Updated the playerState variable initiator to have either the sent state value or an empty string, instead of an empty string
And if the switch wouldn't have a case available and the value would still be a number, it should be empty
    - Updated the customMetrics data insertion into the analytics object by instead of having one if for each possible property, have an automated detection of properties

**Removed**


kennylam (approved):
LGTM thanks @tweenn!

Issue #9633, opened 2022-11-10
Misspellings in Typeahead on Masthead Search


jmkraus213:
### Description

**Description:**
I am seeing misspelling issues with Typeahead on Masthead Search.

**Steps to reproduce:**
- Open https://www.ibm.com/
- Click in search box and type `ibm `and then a space
- Misspellings appear (see below)

![Screen Shot 2022-11-10 at 12 00 18 PM](https://user-images.githubusercontent.com/79653961/201159140-77c509b5-7ff0-4a69-8b84-d85d75c166fd.png)

We checked on the search side and the API has the correct spelling for misspellings we are seeing.  i.e., there is a space after 'ibm'.  It appears there is a bug in the typeahead renderer. 

### Component(s) impacted

Any URL the Masthead Search box appears on ibm.com

### Browser

Chrome

### Carbon for IBM.com version

1.24.0-alpha.3107886671.0

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

ibm.com/search

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

n/a

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

Issue #8416, opened 2022-02-28
Implement a flexible schemaless config for scoped search widget


ncheaz:
### The problem

The current implementation of the schema is fixed to ESQS scope only, even though the search team has ambitions for other scoping in the future. The idea is to remove any assumptions and logic related to ESQS scopes or any other future coping domains (paths, UT code, DCC, etc...)

### The solution

In order to do that we need to pivot to a more flexible config system that deterministically maps scoped search widget config params to the URL that goes to the search app.

This removes the knowledge of any scoping domain from Carbon, which is a plus as it will support any future domains. In this flexible schemaless approach, we define the interface as documentation that the search team maintains, and the search app will know how to handle the config parameters passed in by the Carbon search widget.

The rules for mapping an attribute value in the config JSON to the query params in the search URL are consistent, deterministic, and predictable.

In this approach, the Carbon team will not have to be involved to make changes when we decide to start implementing more scoping domains.

### Application/website

https://www.ibm.com/search/scoped?q=java

### Business priority

Medium Priority = upcoming release but is not pressing

### What time frame would this ideally be needed by (if applicable)

In the next sprint if possible.  Scope searching is a very high value feature for the search team, and we want to setup the configuration schema one time and not have to maintain deprecated and future versions of it, so we ask that scope search functionality be feature flagged away until this is implemented.

### Examples

This is our recommendation for a schema that meets the requirements.  The rules for mapping an attribute value in the config JSON to the query params in the search URL are consistent, deterministic, and predictable

![image](https://user-images.githubusercontent.com/6854716/156034134-cd2c406d-2e0c-4e61-a83d-2f2cc8623928.png)

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #9659, opened 2022-11-18
feat(search): reintroduce scoped search

IgnacioBecerra:
### Related Ticket(s)
#8416
#9633

### Description
This PR reintroduces the previously removed scoped search functionality while Search changes were being made. Now that the changes have been done, we can reintroduce this feature.

To test, go to the Scoped search masthead story, change the scope with the dropdown menu, look up a query, and submit the search input query. 

You should be seeing a different page signaling that the search has been done in a specific part of IBM

Screen Shot 2022-11-17 at 4 37 37 PM

This is using the following URL that was automatically built through the search with typeahead component.

https://www.ibm.com/search/scoped?lang=en&cc=US&lnk=mhsrch&scope-domain=scope&scope-value=["dw,dwaspera"]&scope-type=dw&scope-label=Desarrolladores&q=blockchain

### Changelog

**New**

- scoped search functionality reintroduced
- e2e scoped search tests

**Changed**
- search component has extra parameters if scoped functionality is present

**Removed**

- removed feature flag for the scoped functionality as it will now be public


kennylam (approved):
Thanks @IgnacioBecerra! Looks like you need to run prettier and remove the monster tabbing in the test file.

PR #9676, opened 2022-11-22
Fix/mobile masthead csp

### LINKED ISSUE #9675: [dds-left-nav]: Make mobile masthead CSP compliant
### The problem

This is a follow up to #9599 to remove inline styles from the mobile version of the masthead. Screenshot from @markkulube 
screenshot_2022-11-18_at_2 42 21_pm

### The solution

We need to implement the CSP compliance mixin to move inline style usage to a nonced stylesheet

### Application/website

ondeck.console.cloud.ibm.com

### Business priority

This needs to be a release patch and merged in before Dec 2.

### Examples

_No response_

### Code of Conduct

andy-blum:
### Related Ticket(s)

Fixes #9675 

### Description

This PR continues the work of moving inline styles out of the masthead to ensure CSP compliance within the cloud console. There are some changes to how the CSP compliance mixin is implemented, but the approach of creating a nonced style element within the shadow root is largely the same.

In addition to components creating an internal stylesheet, they also share a global stylesheet for elements that need to utilize it. The internal stylesheet should always be preferred, but some elements, like the `dds-masthead-container` don't attach a shadowroot, and thus cannot have an "internal" stylesheet.

### Changelog

**New**

- Implemented CSP compliance mixin in additional masthead components

**Changed**

- Updates to CSP compliance mixin

### Testing Instructions

The simplest way to test this is to open the chrome console tab of the devtools and to add a "live expression" for `$$('#main-content [style]')`. This will show any elements within the main content wrapper with an inline style attribute. Click through the masthead in various screen widths and ensure no elements populate that live expression.
Screenshot 2022-11-21 at 9 29 42 PM


packages/utilities/src/utilities/settings/settings.js
```diff
@@ -11,11 +11,11 @@
  * @exports ibmdotcom.settings
  * @type {object} Settings object
  * @property {string} [prefix=dds]
- * Carbon for IBM.com v1.40.0',
+ * Carbon for IBM.com v1.41.1',
  *
  */
 const settings = {
-  version: 'Carbon for IBM.com v1.41.0',
+  version: 'Carbon for IBM.com v1.41.1',
```
annawen1, line 18: hmmm this is coming from the `main` branch, jeff did a test branch release that pushed some changes to `main`. Can you double check to make sure there's no carried no other carried over changes from `main`?

packages/utilities/src/utilities/settings/settings.js
```diff
@@ -11,11 +11,11 @@
  * @exports ibmdotcom.settings
  * @type {object} Settings object
  * @property {string} [prefix=dds]
- * Carbon for IBM.com v1.40.0',
+ * Carbon for IBM.com v1.41.1',
```
annawen1, line 14:
```suggestion
 * Carbon for IBM.com v1.41.0',
```

packages/utilities/src/utilities/settings/settings.js
andy-blum: I just [reverted the publish commit](https://github.com/andy-blum/carbon-for-ibm-dotcom/commit/7f58ec079748ee5585d55255d2aaa4c21724a969) - this should take care of these changes.

packages/utilities/src/utilities/settings/settings.js
```diff
@@ -11,11 +11,11 @@
  * @exports ibmdotcom.settings
  * @type {object} Settings object
  * @property {string} [prefix=dds]
- * Carbon for IBM.com v1.41.1',
+ * Carbon for IBM.com v1.41.0-rc.0',
  *
  */
 const settings = {
-  version: 'Carbon for IBM.com v1.41.1',
+  version: 'Carbon for IBM.com v1.41.0-rc.0',
```
annawen1, line 18: it should be just `Carbon for IBM.com v1.41.0`, we don't update this settings file for every RC released

andy-blum: currently taking bets on how many more commits it'll take me to get this right 😂

annawen1 (approved):
LTGM!

PR #9703, opened 2022-11-30
chore(csp): remove CSPCompliance mixin

andy-blum:
### Related Ticket(s)

#9599 #9675

Followup to PRs #9676 #9605

### Description

This PR removes the CSP compliance mixin. After additional research we discovered the following:
1. The mixin was entirely unnecessary. Setting element styles via `HTMLElement.style.<property>` is, in fact, allowed per CSPs without `unsafe-inline`. [Example codepen](https://codepen.io/andy-blum/pen/ExRRRBj)
2. The mixin as architected wouldn't work anyway, as it requires `unsafe-eval`.

### Changelog

**Removed**

- Removed CSP compliance mixin


annawen1 (approved):
LGTM!

PR #9811, opened 2022-12-07
fix(tabs): basic keyboard navigation

IgnacioBecerra:
### Related Ticket(s)
https://github.com/carbon-design-system/carbon-web-components/issues/970
https://github.com/carbon-design-system/carbon-web-components/issues/707

### Description
This PR ensures that the Tabs and Content Switcher components are able to be work with tab navigation.

### Changelog

**New**

- added tab-index in the tab shadow dom
- added home/end key functionality

**Changed**

- triggering the content switching upon pressing the enter key

**Removed**

- remove unneeded inFocus property in the tab item component


kennylam (approved):
LGTM!

PR #9814, opened 2022-12-08
feat(masthead L1): update data structure

andy-blum:
### Related Ticket(s)

[ADCMS-2723](https://jsw.ibm.com/browse/ADCMS-2723)

### Description

Updates the type definitions in services-store/TranslationAPI.
- Changes current `MastheadL1` to `LegacyMastheadL1`
- Deprecates `LegacyMastheadL1`
- Creates new `MastheadL1` and supporting sub-interfaces

Updates the masthead story's links.ts file to use new `MastheadL1` interface definition.   

### Changelog

**New**

- L1 data structure


packages/services-store/src/types/translateAPI.ts
```diff
@@ -85,13 +85,52 @@ export interface MastheadLink {
...
+  };
+}
+
+export interface L1MenuItem extends BasicLink {
+  submenu?: L1Submenu;
+}
+
+export interface L1Submenu {
+  announcement?: string; // From AEM rich text editor
+  menuSections: L1SubmenuSection[]; // maximum of 3 in outer array | renamed to be agnostic to the design
```
jkaeser, line 115: I wonder if the maximum of three limitation should be enforced in code or part of content guidance. Either way, we can remove my part of this comment.

```suggestion
  menuSections: L1SubmenuSection[]; // maximum of 3 in outer array
```

packages/services-store/src/types/translateAPI.ts
```diff
@@ -85,13 +85,52 @@ export interface MastheadLink {
...
+    login?: BasicLink;
+  };
+}
+
+export interface L1MenuItem extends BasicLink {
+  submenu?: L1Submenu;
+}
+
+export interface L1Submenu {
+  announcement?: string; // From AEM rich text editor
```
jkaeser, line 114: Now that we're working with actual code, can we make this something like `HTMLTemplateResult | HTMLElement | Text`. I considered simply `Node`, but I feel like that's too generic. What do you think?

packages/services-store/src/types/translateAPI.ts
```diff
@@ -85,13 +85,52 @@ export interface MastheadLink {
...
+  footer?: {
+    title: string;
+    url: string;
+  };
+}
+
+export interface L1SubmenuSection {
+  span: 1 | 2 | 3;
+  items: BasicLink | L1SubmenuSectionHeading[];
+}
```
jkaeser, line 125: I'm still not sure about grouping links and heading as `items`. An argument can be made that this design shows two `L1SubmenuSections` in a single column layout, rather than an unstructured collection of headings and links:

Screenshot 2022-12-08 at 10 40 39 AM

In fact, looking at the designs again, I am feeling more and more like this is the paradigm we should be using, as they lay out specific one, two, and three column layouts. In section 4.3.4 they also mention the column span variation is "available only on the 3-column menu".

This leads me to think it would be wiser to establish `columnLayout: 1 | 2 | 3;` in the `L1Submenu` interface and change this to:

```suggestion
export interface L1SubmenuSection {
  span?: 1 | 2;
  items: BasicLink;
  heading: L1SubmenuSectionHeading[];
}
```

where `span` will only be consulted if the parent submenu has `columnLayout: 3`.

jkaeser (changes requested):
@andy-blum Let me know if you want to discuss the changes I'm suggesting further. Might be easier on a call taking a look at the design specs together.

packages/services-store/src/types/translateAPI.ts
andy-blum: I don't know that we want to define the JSON data as HTMLTemplateResult or HTMLElement. I think that value will always be a `string`, even if that string contains markup.

jkaeser: You're right. I was thinking about how these types are enforced in the component code, but that wouldn't work with the JSON that is our actual content source.

jkaeser (approved):
@andy-blum The types look good to me, now! Just gotta fix the CI issues... I think some of the latest type updates aren't reflected in the Stories' link content.

Issue #9696, opened 2022-11-28
[H2 tag component]: content block headings do not render a heading tag


sirkitree:
### Description

The SEO team has noticed pages/components not correctly utilizing heading structures, specifically their H2 tags. 

For example,  in the simple narrative component, it's using <dds-content-block-heading> tag we could update the heading tag to include an H2 tag (see example below) to test how the CSS being applied would be impacted by this change.

What we're looking for is any recommended approaches that could be better than this:
 

Heading Example:
```
<dds-content-block-heading>
    <h2>
         text here
    </h2>
</dds-content-block-heading>
```

Current Sites the SEO team has noticed the issues: 
https://www.ibm.com/aiops
https://www.ibm.com/application-performance-management
https://www.ibm.com/edge-computing
https://www.ibm.com/workflow
https://www.ibm.com/zero-trust

### Component(s) impacted

<dds-content-block-heading>

### Browser

Chrome, Safari, Firefox, Microsoft Edge

### Carbon for IBM.com version

1.39.0

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

IBM.com AEM Sites

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://www.ibm.com/aiops

### Steps to reproduce the issue (if applicable)

- Go to one of the above URLs
- inspect a heading 
- note the lack of a heading (H*) tag

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

sirkitree:
Response from @jeffchew in Slack:
> if rendering as h2's in either the light or shadow doms are what we should look at , if you could open an issue on that would be good so we can review. It might also include an assessment of how this dovetails with other components, or if we approach it as something configurable in the API.

sirkitree:
From the meeting on 12/14

> Heading SEO Issue: Ari, one of the Carbon Devs, might be able to fix this in another task she’s working on. Kenny will let us know how that goes, and if Ari isn’t able to do it, they’d welcome a contribution from us.

@ariellalgilmore do you have an issue you are working on that we could link to here? Is this a duplicate?

[...]

PR #9826, opened 2022-12-14
fix(h1/h2): reveal in lightdom for content-block and leadspace

ariellalgilmore:
### Related Ticket(s)

#9696 

### Description

missing h2 tag in lightdom
Screen Shot 2022-12-14 at 2 34 33 PM

### Changelog

**Changed**

- after stripping the html and h2 tag is applied without the styles 
- the styles do effect the current heading styles


kennylam (approved):
LGTM!

PR #9854, opened 2022-12-22
fix(rollup): update cwc rollup config

kennylam:
### Description

Update `carbon-web-components` rollup config to suppress the following warning:
```
@rollup/plugin-replace: 'preventAssignment' currently defaults to false. It is recommended to set this option to `true`, as the next major version will default this option to `true`.
```

### Changelog

**Changed**

- set `preventAssignment: true`


emyarod (approved):
looks good to me!

Issue #9816, opened 2022-12-08
[React Wrapper]: React-wrapped web-components importing from incorrect file


huntermacd:
### Description

When compiling a Create-React-App bootstrapped app using `@carbon/ibmdotcom-web-components@1.26.0`, importing components as follows: `import DDSCard from '@carbon/ibmdotcom-web-components/es/components-react/card/card';`, throws the following error:

```
ERROR in ./node_modules/@carbon/ibmdotcom-web-components/es/components-react/card/card.js 18:42-58
export 'descriptor' (imported as 'parentDescriptor') was not found in '@carbon/web-components/es/components/link/link.js' (possible exports: LINK_SIZE, default)
```

The web-component should import `descriptor` from `@carbon/web-components/es/components-react/link/link.js` instead.

### Component(s) impacted

All React-wrapped web-components.

### Browser

Chrome

### Carbon for IBM.com version

v1.26.0

### Severity

Severity 1 = The design is broken in a critical way that blocks users from completing tasks or damages the brand. Affects major functionality, no workaround.

### Application/website

Widget Team

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://www.ibm.com/standards/carbon/web-components/react/?path=/story/components-card--default

### Steps to reproduce the issue (if applicable)

- install `@carbon/ibmdotcom-web-components@1.26.0`
- import React-wrapped component
- attempt to compile app with Webpack

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #9870, opened 2022-12-28
fix(react-wrappers): correct import source path to react wrapper component

annawen1:
### Related Ticket(s)

[React Wrapper]: React-wrapped web-components importing from incorrect file #9816

### Description

Correct import source. Babel plugin was looking for old `@carbon/carbon-web-components/...` imports when generating react wrappers. Corrected it to look for `@carbon/web-components/...` so it can correctly swap them to the react wrapper version.

### Changelog

**Changed**

- make sure `@carbon/web-components/...` imports in components are getting properly swapped with react wrapper version


kennylam (approved):
LGTM thanks @annawen1!

PR #9872, opened 2022-12-29
feat(leadspace): add gradient knob to story

### LINKED ISSUE #9631: Leadspace gradient-style-scheme attribute string option request
### The problem

The leadspace web component has the attribute default for ```gradient-style-scheme``` sets to "with-gradient". 

In AEM, we have a dialog option to "Remove Image Gradient" and I'm not able to apply an empty string to the ```gradient-style-scheme```; the string will always default to "with-gradient".  I tried to use another string like "no-gradient" to the attribute, but this did not remove the gradient from the image.  

### The solution

Is there a way we can have a string option like "no-gradient" or "none" that could be applied to the ```gradient-style-scheme``` attribute that removes the gradient from the image? 

### Application/website

AEM

### Business priority

If this update can be part of the next release, that would be greatly appreciated so we can update the leadspace to use carbon tags. Thanks.

### Examples

Example Page from Dev Environment: https://dev-author.np-roks.cms.cis.ibm.net/content/adobe-cms/language-masters/en/kayle-s-test-page/learn-landing-page-test.html?wcmmode=disabled

### Code of Conduct

andy-blum:
### Related Ticket(s)

Closes #9631

### Description

Adds a storybook knob to control the `gradient-style-scheme` attribute values of `'with-gradient'` and `''`

### Changelog

**Changed**

- Added knob to leadspace stories


andy-blum:
Deploy previews haven't built - closing & reopening to try and kick those off.

PR #9881, opened 2023-01-04
fix(bx-dropdown): prevent interactivity while disabled

andy-blum:
### Related Ticket(s)

Closes: #9754

### Description

When disabled bx-dropdown is not interactive

### Changelog

**Changed**

- Fixed bug in bx-dropdown


annawen1 (approved):
LGTM! Thanks!

PR #9883, opened 2023-01-04
fix(github-actions): properly ignore packages in lerna builds

jeffchew:
### Related Ticket(s)

No related issue

### Description

The lerna builds didn't seem to be properly building the vendor files, this is an attempt to use the `--ignore` flag instead (which was previously used).

### Changelog

**Changed**

- `deploy-canary.yml`


annawen1 (approved):
LGTM!

Issue #9831, opened 2022-12-15
[Button group]: Update the button order


oliviaflory:
### The problem

There have been new updates to Carbon button guidance and other research done to suggest that within a button group the `primary` button should come first when left aligned, and the `secondary` or `tertiary` button should be placed on the right.

The Cloud team has had better performance with their new lead space with this approach

Follows Carbon [button group guidelines](https://carbondesignsystem.com/components/button/usage/#button-groups)

This will also allow our adopters to have only one button (the primary) presented when there is only one button needed. Reference #9687 

<img width="782" alt="Screenshot 2022-12-15 at 4 00 24 PM" src="https://user-images.githubusercontent.com/45692622/207970571-e8f35f08-e64a-4b3e-b97c-81fd446f4c70.png">

### The solution

Reverse the order of the current [button group]() to `primary` followed by `tertiary`
<img width="458" alt="Screenshot 2022-12-15 at 4 32 12 PM" src="https://user-images.githubusercontent.com/45692622/207971383-193fcb65-43ed-4ee0-be26-8d947f3665eb.png">

*Note that Carbon does allow the secondary action to lead when the buttons are on the right side, should we allow for both?*
<img width="736" alt="Screenshot 2022-12-15 at 4 28 19 PM" src="https://user-images.githubusercontent.com/45692622/207970770-db011731-2a31-4fae-8ab5-b81df029c86d.png">

### Application/website

Carbon for IBM.com

### Business priority

High Priority = pressing release

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #9884, opened 2023-01-04
feat(button-group): update button type order

annawen1:
### Related Ticket(s)

[Button group]: Update the button order #9831

### Description

Update the `button-group` component to render the primary button on the left instead of right.

Screen Shot 2023-01-04 at 3 58 44 PM

Screen Shot 2023-01-04 at 3 59 12 PM

### Changelog

**Changed**

- set the first button child to be primary instead for both react and web-components `button-group`


kennylam (approved):
LGTM! Looks like you just need to update snapshots.

RichKummer (approved):
LGTM from the design side!

Issue #9815, opened 2022-12-08
[card link] Hover when disabled


RichKummer:
### Description

When `card link` is disabled, it should not have a hover state when a user places their cursor over the `card link`. Currently, the component darkens when the user hovers over it.

See current:
https://user-images.githubusercontent.com/15643582/206571769-9891a708-f1a1-4094-80de-fc811d669ead.mov

### Component(s) impacted

card link

### Browser

Chrome

### Carbon for IBM.com version

v.1.25.0

### Severity

Severity 4 = The problem is not visible to or noticeable to an average user. Affects minor functionality, no workaround needed.

### Application/website

C4IBM.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://codesandbox.io/s/github/carbon-design-system/carbon-for-ibm-dotcom/tree/main/packages/web-components/examples/codesandbox/components/card-link

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

kennylam:
@RichKummer this is the standard behavior for disabled links and is consistent with core Carbon styles. See example [here](https://react.carbondesignsystem.com/?path=/story/components-button--default&args=disabled:true).

oliviaflory:
@kennylam The cursor change is consistent with the core Carbon styles, but I think the darkening of the card background (probably $ui-hover) shouldn't happen if it's disabled.

PR #9892, opened 2023-01-10
fix(styles): update disabled card hover styles

kennylam:
### Related Ticket(s)

#9815

### Description

Removes hover styles from `disabled` card links.

### Changelog

**New**

- add disabled hover styles to card links


RichKummer (approved):
LGTM!

Issue #8834, opened 2022-05-16
[Back-To-Top]: Component doesn't hide itself when page content changes


martin-kocisek-ibm:
### Description

We are dynamically changing content on the page (React),but we use web component because this exact component is not yet available in react. When we jump from one page with back-to-top component shown (so we are scrolled down) we expect back-to-top button to hide, as the new page loaded and we are on top of it. I read the documentation and there is no possibility to refresh component, so we are not able to refresh it even manually when page change occurs.

### Component(s) impacted

Back-To-Top

### Browser

Chrome, Safari, Firefox, Microsoft Edge

### Carbon for IBM.com version

1.32.1

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

https://www.ibm.com/docs/en/

### Package

@carbon/ibmdotcom-react, @carbon/ibmdotcom-web-components, @carbon/ibmdotcom-styles

### CodeSandbox example

no

### Steps to reproduce the issue (if applicable)

1. open the page https://www.ibm.com/docs/en/spectrum-scale/5.1.2?topic=summary-changes
2. scroll down and check the back to top button is displayed
3. click on the item Quick reference in the TOC
4. check the opened page with back to top button which should not be displayed as the page has not enough content for back to top button
5. refresh the page and check the back to top button is not displayed now

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #9900, opened 2023-01-12
fix(back-to-top): trigger manage visibility on body height changes

m4olivei:
### Related Ticket(s)

Closes: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8834

[ADCMS-2857](https://jsw.ibm.com/browse/ADCMS-2857)

### Description

Reproducing this bug was a little tricky within Storybook. What I did was add a story for the Back to Top component called "Default Faux Paged". Initially with the same contents as the "Default" story. At the bottom of the page, there is a button, which when clicked will replace the page contents with other contents, speifically _less_ content, where no Back to Top button would be required at a typical desktop break.

The core of the issue was that, although we have a ResizeObserver to watch changes in the body element height, we don't make use of it to show / hide the Back to Top button when the height changes. I've adjusted the logic and renamed a couple of things, to hopefully make the code more readable.

### Testing instructions

- [ ] Navigate to the Default Faux Paged story under the Back to Top component at a Desktop breakpoint
- [ ] Scroll down the page, until you see the "Next page" button. As you scroll the Back to Top button will show up per usual.
- [ ] Click Next Page. The contents of the page should be replaced.
- [ ] The Back to Top button should dissapear, and not persist. This will depend on your viewport height.
- [ ] Scroll the new content up and down, the Back to Top button should not appear.

### Changelog

**New**

- Back to Top / Default Faux Paged story

**Changed**

- Updated back to top logic to cover body content change scenarios


m4olivei:

m4olivei:
recheck

packages/web-components/src/components/back-to-top/back-to-top.ts
```diff
@@ -90,17 +102,28 @@ class DDSBackToTop extends HostListenerMixin(StableSelectorMixin(LitElement)) {
...
 
-  private _observeResizeBody = (entries) => {
+  /**
+   * ResizeObserver callback function.
+   *
+   * We only observe the body element, therefore entries will typically only
+   * have a single item corresponding to changes in the size of the body
+   * element.
+   */
+  private _observeResizeBody = (entries: ResizeObserverEntry[]) => {
```
andy-blum, line 112: I think we probably ought to rename this function becuase `observeResizeBody` and `observerResizeBody` is just _begging_ for typos/mis-understandings.

I'd suggest `handleBodyResize` here.

packages/web-components/src/components/back-to-top/back-to-top.ts
```diff
@@ -90,17 +102,28 @@ class DDSBackToTop extends HostListenerMixin(StableSelectorMixin(LitElement)) {
...
-   * Show button only when document height is 3x greater than viewport
+   * The back to top button is only necessary when the document height
+   * is 3x greater than the viewport.
    */
-  private _showBackToTop() {
+  private _isBackToTopNeededForPage() {
     // @ts-ignore
-    this._windowHeight = this.ownerDocument.defaultView?.innerHeight;
-    return this._bodyHeight > this._windowHeight * 3;
+    this._viewportHeight = this.ownerDocument.defaultView?.innerHeight;
```
andy-blum, line 125: TIL: [Document.defaultView](https://developer.mozilla.org/en-US/docs/Web/API/Document/defaultView)

andy-blum (changes requested):
Without begin able to actually test this as we wait for the deploy previews to build, I have one change suggestion

packages/web-components/src/components/back-to-top/back-to-top.ts
m4olivei: Good thought. Updated.

packages/web-components/src/components/back-to-top/back-to-top.ts
m4olivei: MDN and I were great friends through this, for me also:

* [ownerDocument](https://developer.mozilla.org/en-US/docs/Web/API/Node/ownerDocument)
* [scrollingElement](https://developer.mozilla.org/en-US/docs/Web/API/Document/scrollingElement)

andy-blum (approved):

ariellalgilmore (commented):
Looks great! thanks for the updates. Is it possible though to have the Faux story instead integrated with the default story to include the updates with the next page and then we can remove the additional Faux story?

ariellalgilmore (approved):
LGTM!

m4olivei:
Thanks for the review @ariellalgilmore !

proeung:
Looks like the Percy React preview needs review, which is blocking the automatic merging. @ariellalgilmore or @annawen1 can you help take a look so that we can get this PR merged?

Screen Shot 2023-01-24 at 11 32 45 AM

PR #9902, opened 2023-01-12
feat(masthead-contact): improve interactions with contact module

### LINKED ISSUE #9369: [Masthead v2.1.0]: Phase 2 - Add improvements to Contact us CTA button
**Engineering info:**  
- JIRA Ticket: https://jsw.ibm.com/browse/ADCMS-2602
- JIRA Epic: https://jsw.ibm.com/browse/AEMSITES-177

-----

### Description
- A follow-up phase 2 task from https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/9359
- From @oliviaflory 
  - I like that when you have a mega menu open and then select the contact us button, the mega menu collapses/closed. I think the reverse should also be true. When I select the Contact us button first, the contact module opens, and if I then select a mega menu the chat module stays open and obstructs part of the mega menu content

contact module mega menu Sep-13-2022 17-49-10

**Questions that we should run by the accessibility team:**

- Should clicking the chat button again (after the chat module is opened) trigger the chat module to close? Asking because otherwise the user has to move their mouse to the contact module to close, and all other navigation items (mega menus and profile menu) close the menus on a second click.
- Should the focus state move to the Chat module once the user clicks it for keyboard users?
- When testing with the screen reader, there is no announcement that the chat module has opened
- After you select the Contact us button, a keyboard user has to tab through the entire page before getting to the contact module. This doesn't seem like a great experience (hence the suggestion to move the focus state). Shift tab would go back to the contact us button.

### Design Specs
- Figma mockup - https://www.figma.com/file/35JfN0Sg4ltIIjM7jBDKMk/Navigation?node-id=2325%3A136628

andy-blum:
### Related Ticket(s)

Closes #9369

[ADCMS-2602](https://jsw.ibm.com/browse/ADCMS-2602)

### Description

- Masthead container adds an event listener to the Document for `cm-app-ready` from the Contact Module ([source](https://github.ibm.com/live-advisor/cm-app/blob/943dfbcac5b8efde05a00c4c5558947495f96a58/js/helpers/documentEvents.js#L7))
- Stores reference to `window.CM_APP` internally to DDSMastheadContainer
- Uses internal reference to CM_APP to close with `.init()` method

### Changelog

**Changed**

- Improved UX between masthead and contact module


packages/web-components/src/components/masthead/masthead-contact.ts
```diff
@@ -45,6 +47,18 @@ class DDSMastheadContact extends DDSMastheadProfile {
     `;
   }
 
+  updated(changedProperties) {
+    if (changedProperties.has('expanded')) {
+      if (!this.expanded) {
+        const mastheadContainer = this.closest(`${ddsPrefix}-masthead-container`) as DDSMastheadContainer;
+
+        if (mastheadContainer.contactModuleApp) {
+          (mastheadContainer.contactModuleApp as CMApp).init();
```
m4olivei, line 56: Maybe worth a comment explaining the workaround and that the team that owns the content module is working on a better API to do this? It's confusing reading a call to `init()` when what we mean here is more like minimize.

packages/web-components/src/components/masthead/masthead-contact.ts
```diff
@@ -45,6 +47,18 @@ class DDSMastheadContact extends DDSMastheadProfile {
     `;
   }
 
+  updated(changedProperties) {
+    if (changedProperties.has('expanded')) {
```
m4olivei, line 51: This condition always evalutes to `true` since we're just checking for `has`, so we can remove it.

packages/web-components/src/components/masthead/masthead-contact.ts
```diff
@@ -45,6 +47,18 @@ class DDSMastheadContact extends DDSMastheadProfile {
...
+      if (!this.expanded) {
```
m4olivei, line 52: Given the outer condition is always `true`, I believe that _may_ result in an unnecessary call to `init()` on the first render here, since `this.expanded` is `flase` on first render. I guess it depends on if the `mastheadContainer.contactModuleApp` is ready yet. Woulud it make sense to change the condition here to `changedProperties.get('expanded') === true && !this.expanded` ?

m4olivei (changes requested):
Behavior looks good, just a couple of small  suggestions for the logic.

packages/web-components/src/components/masthead/masthead-contact.ts
andy-blum: `changedProperties` here is a `Map` of the properties that have changed during this run of the update lifecycle, so this condition is checking if `this.expanded` changed, not if `this` has `expanded`.

We want to make sure `this.expanded` has changed before we implement side effects.

https://lit.dev/docs/components/lifecycle/#updated

packages/web-components/src/components/masthead/masthead-contact.ts
andy-blum: Good call.

packages/web-components/src/components/masthead/masthead-contact.ts
andy-blum: As I noted on the above comment, that value doesn't always eval to `true`. It will evaluate to false if the value of `this.expanded` is not what caused this update lifecycle.

packages/web-components/src/components/masthead/masthead-contact.ts
m4olivei: Ahhh OK gotcha. I misunderstood, sorry about that. Although interestingly, if you set a breakpoint just inside the `updated` method, the first call has all the properties in the `Map` with values of `undefined`. So for the first call at least, this condition will always be `true`.

andy-blum: Correct. `changedProperties` maps to the values that have changed, and each property's values correspond to their _previous_ state.

During initialization, _all_ properties change from `undefined` to whatever their initial state is. There's also a `firstUpdated` method that runs during this first setup after the `connectedCallback` method.

m4olivei: Neat. Thanks for the added context Andy!

m4olivei (approved):
LGTM! Thanks.

proeung (approved):
Code looks good and the behavior is working as expected (see attached).

_Note to self:_ Check to see if there's a plan to migrate the management of the Contact Module script to the Services/Utilities packages in the future.

proeung:
@andy-blum Looks like the `cloud-masthead` e2e test is failing and it's preventing the merge to go through. Can you take a look?

Screen Shot 2023-01-26 at 2 26 16 PM

proeung (approved):

Issue #9936, opened 2023-01-23
[breadcrumb]: / separator not showing on v1.27.0


AlessandroPomponio:
### Description

When using multiple items in the breadcrumb, I'd expect a `/` separator to be present in between the items (as it was on v1.25.0), but on v1.27.0 this does not seem to be the case anymore

### Component(s) impacted

Breadcrumb

Before:
![image](https://user-images.githubusercontent.com/10339005/214085020-0d19fc54-06ad-4bcd-81e2-bb38bdd4075c.png)

After (don't mind the missing suffix):
![image](https://user-images.githubusercontent.com/10339005/214084959-b2b180f7-74ca-447d-8511-0822633729d0.png)

### Browser

Microsoft Edge

### Carbon for IBM.com version

v1.27.0

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

https://registry.st4sd.res.ibm.com/

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

-

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

Issue #9923, opened 2023-01-18
Modal not displaying with @carbon/web-components 1.23.0


pac0rro:
### Description

Today I noticed the change of package from **carbon-web-components** to **@carbon/web-components**.

(I am not sure this is the correct issue page: is not updated for package @carbon/web-components)

I have applied the change in my project's package.json. This is what I found:

- Version 1.22.0 of @carbon/web-components the **Modal appears**
- Version 1.23.0 of @carbon/web-components  **don't shows the Modal component**

I've seen that style.visibility is hidden for "open = true" property.

Thank you

### Component(s) impacted

Modal web component

### Browser

Chrome, Firefox

### Carbon for IBM.com version

1.23.0

### Severity

Severity 1 = The design is broken in a critical way that blocks users from completing tasks or damages the brand. Affects major functionality, no workaround.

### Application/website

Internal website

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

Including the storybook sample re-creates the issue

### Steps to reproduce the issue (if applicable)

Put a button in a page and open the modal that the example shows. (Vue example):

        <button @click="$refs.modalex.open = true">Modal</button>

        <bx-modal id="modalex" ref="modalex">
            <bx-modal-header>
                <bx-modal-close-button></bx-modal-close-button>
                <bx-modal-label>Label (Optional)</bx-modal-label>
                <bx-modal-heading>Modal Title</bx-modal-heading>
            </bx-modal-header>
            <bx-modal-body><p>Modal text description</p></bx-modal-body>
            <bx-modal-footer>
                <bx-modal-footer-button kind="secondary" data-modal-close>Cancel</bx-modal-footer-button>
                <bx-modal-footer-button kind="primary">Save</bx-modal-footer-button>
            </bx-modal-footer>
        </bx-modal>

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #9942, opened 2023-01-25
fix(styles): add missing styles for breadcrumb and modal

annawen1:
### Related Ticket(s)

[breadcrumb]: / separator not showing on v1.27.0 #9936
Modal not displaying with @carbon/web-components 1.23.0 #9923

### Description

Some styles have not been extending from `carbon-components` for `bx-modal` and `bx-breadcrumb` components. Will need to dig further to see why they aren't showing, but in the mean time we can add those styles to ours.

### Changelog

**Changed**

- added styles from `carbon-components` that are not showing for `bx-modal` and `bx-breadcrumb`


ariellalgilmore (approved):
LGTM!

PR #9967, opened 2023-01-30
fix(issue-template): add cwc option

ariellalgilmore:
### Related Ticket(s)

n/a
### Description

add CWC option as a package to choose from in the issue template


annawen1 (approved):
LGTM! Thanks

Issue #9890, opened 2023-01-09
Carbon gray tag is not displaying correctly


sirkitree:
### Description

Environment: Product
 
When I add a Carbon tag to a component and pick the color "gray" in the dropdown, the tag will display as green tag, not gray. This needs to be updated so when gray is selected as the tag color, the tag displays as a gray tag.
 
This bug seems to appear where ever Carbon tags are being used and isn't component specific.

### Component(s) impacted

Components: Several including structured list, CTA section with static card, pricing table, and where ever Carbon tags are being used

Mainly the Tag Group component

### Browser

Chrome, Safari, Firefox, Microsoft Edge

### Carbon for IBM.com version

latest

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

IBM.com AEM Sites

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

?

### Steps to reproduce the issue (if applicable)

https://prod-author.roks.cms.cis.ibm.net/content/adobe-cms/language-masters/en/prod-sandbox/testproduct-overview-jw/-test-product-feature—jw.html?wcmmode=disabled

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

IgnacioBecerra:
@sirkitree Hi there! I'm unable to reproduce the behavior you are mentioning. In the [Storybook page](https://web-components.carbondesignsystem.com/?path=/story/components-tag--default), you can change the color to grey and it does not revert back to green. 

Likewise, in our Tag Group story, you can use individual Carbon tags with different colors, including grey, and they don't default to being green.

That being said, if the `type` isn't set, the tag **will** default to green. 

For example
```
```

PR #9972, opened 2023-01-31
fix(tag-group): ensuring tags can be gray

IgnacioBecerra:
### Related Ticket(s)
#9890

### Description
Ensured that the carbon tag can be gray when used within Tag Group.

### Changelog

**Removed**

- removed tag group condition that would change a tag color to green if gray was chosen


ariellalgilmore:
Should the first tag in this tag group still be green?

[web-components]
Screen Shot 2023-02-01 at 8 13 31 AM

[react-wrappper]
Screen Shot 2023-02-01 at 8 14 57 AM

IgnacioBecerra:
@ariellalgilmore Technically the default color should be green, according to the old specs from Shixie iirc, but wondering what @oliviaflory or @RichKummer might think now?

Should the default color of the regular carbon tag be Grey, or green?

oliviaflory:
@IgnacioBecerra @ariellalgilmore I'm going to attempt to answer without using the word "default" so I am not mixing up what is "default" in the Carbon tag vs "default" in the Card.

In the issue #9890 Jared states
>When I add a Carbon tag to a component and pick the color "gray" in the dropdown, the tag will display as green tag, not gray. This needs to be updated so when gray is selected as the tag color, the tag displays as a gray tag.

>This bug seems to appear where ever Carbon tags are being used and isn't component specific.

I think this is the desired outcome:
- The user/author/developer can opt to use the `gray` Carbon tag within any component (ie Cards, Structured list, Pricing table). (I am assuming they would 
- Within the card component story, they/we do want to show the first tag in the card set to the `green` Carbon tag, which is most commonly used by AEM for tagging a product as "Most popular".

From the way the issue was phrased, it sounded like even if the developer set a tag to `gray` it would display as `green`?

When I look at the [Tag group story](https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/9972/index.html?path=/story/components-tag-group--default&knob-Add%20tags:_Default=true&knob-CTA:_Default=Learn%20more&knob-Card%20style:_Default=none&knob-Eyebrow:_Default=Industry&knob-Heading:_Default=Aerospace%20and%20defence), and change the tag type knob to `Carbon tag` they all are green. Is that part of what Jared is having trouble with?

I hope this is helpful!

annawen1:
@IgnacioBecerra I'm a little confused by the description and changelog for the PR, we're not changing the default color to grey are we? We're just allowing the gray color to show when adopter specifies "gray" in the type

IgnacioBecerra:
@annawen1 Yeah, you're right, that's all we're doing. I'm gonna update the changelog and description.

RichKummer (approved):
Seeing that the theme can be swapped to `gray` and shows the correct colors, LGTM!

ariellalgilmore (approved):
LGTM!

PR #9973, opened 2023-01-31
fix(styles): downgrade sass version as there is an issue with @extend

annawen1:
### Related Ticket(s)

### Description

There seems to be an issue with `@extend` in the recent upgrade of `sass`. Downgrading `sass` version until there is fix

### Changelog

**Changed**

- downgrade sass back to `~1.40.0`

**Removed**

- band-aid fixes that should be resolved with the `sass` downgrade


ariellalgilmore (approved):
LGTM!

Issue #9963, opened 2023-01-27
[Tabs extended media]: Hover resizing in Safari


RichKummer:
### Description

When in Safari, `tabs extended media`'s container will warp and be pushed off grid as the mouse hovers over the component. Only seems to happen in Safari, was not able to replicate in Chrome or Firefox. Does not seem to effect `@sm`, but seeing it `@md` through `@max` breakpoints.

User is unable to click other tabs to navigate to additional content. Seeing in react and wc storybooks.

![tabs-extended-media-bug](https://user-images.githubusercontent.com/15643582/215200006-15d9aa0e-ec33-4861-8ebd-923b7be2f4d8.gif)

**Consulting example**
https://user-images.githubusercontent.com/15643582/215195062-6b25f575-91b2-4a9f-84aa-469af85374b0.mov

**Storybook**
https://user-images.githubusercontent.com/15643582/215194985-b91a1e20-a6ab-4ba1-9f85-b977b38bbd9b.mov

### Component(s) impacted

tabs-extended-media

### Browser

Safari

### Carbon for IBM.com version

v1.27.0

### Severity

Severity 1 = The design is broken in a critical way that blocks users from completing tasks or damages the brand. Affects major functionality, no workaround.

### Application/website

https://www.ibm.com/consulting/?lnk=flathl

### Package

@carbon/ibmdotcom-react, @carbon/ibmdotcom-web-components

### CodeSandbox example

https://codesandbox.io/s/github/carbon-design-system/carbon-for-ibm-dotcom/tree/main/packages/web-components/examples/codesandbox/components/tabs-extended-media

### Steps to reproduce the issue (if applicable)

1. Open Safari
2. Navigate to [IBM Consulting](https://www.ibm.com/consulting/?lnk=flathl)
3. Scroll down to "Strategic partnerships" section
4. Hover over `tabs extended media` component

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #9974, opened 2023-01-31
fix(tabs-extended-media): safari hover

ariellalgilmore:
### Related Ticket(s)

#9963 
### Description

Safari causing tabs extended media to resize on hover

### Changelog

**New**

- added `width: 100%` to keep from tabs-extended-media changing size on hover


IgnacioBecerra (approved):
LGTM, works in Safari properly now, thanks @ariellalgilmore!

Issue #9951, opened 2023-01-26
[Masthead]: Clicking outside the side-nav menu fails to close it after Lit v2 upgrade


IgnacioBecerra:
### Description

After upgrading to using Lit v2, the side nav focus is unable to be opened properly. If the menu is forcefully opened through code manipulation, it won't close if clicking outside the menu, as it used to.

### Component(s) impacted

Masthead

See the last 2-3 seconds in this video where clicking outside fails to work.

https://user-images.githubusercontent.com/24970122/214748294-adbd03b6-a2db-46b1-a13a-4c05703c9547.mov

### Browser

Chrome

### Carbon for IBM.com version

lit-v2 upgrade branch

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

N/A

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://carbon-design-system.github.io/carbon-for-ibm-dotcom/canary/web-components/?path=/story/components-masthead--default

### Steps to reproduce the issue (if applicable)

1. Open the Masthead in mobile viewport to get the side nav button
2. Activate the side nav
3. Keep pressing tab until the focus goes outside the side nav.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #10019, opened 2023-02-02
fix(masthead): left nav click out

ariellalgilmore:
### Related Ticket(s)

#9951 
### Description

with the lit v2 upgrade the masthead left-nav wasn't closing when the user clicked outside of the left-nav

Feb-02-2023 10-50-10
### Changelog

**Changed**

- the `event.target` is returning `dds-masthead-container` instead of `dds-left-nav-overlay`, so instead now using `composedPath` to see if user is clicking on the `left-nav-overlay`


kennylam (approved):
LGTM thanks @ariellalgilmore!

RichKummer (approved):

PR #10028, opened 2023-02-08
docs: remove themeselector from live demo

alisonjoseph:
### Related Ticket(s)

Closes https://github.com/carbon-design-system/carbon-platform/issues/1592

### Changelog

**Removed**

- remove themeSelector from live demo components


francinelucca (approved):
FYI @mattrosno @alisonjoseph web-components does have a theme selector but we can't tap into it from the URL, this is a use case we don't have a solve for so I think removing the option is the way to go here , approved! :) 
image

RichKummer (approved):

Issue #10037, opened 2023-02-11
[Build]: yarn build can fail when switching between branches


m4olivei:
### Description

On some branches, the following artifacts are produced on `yarn build`, but not
cleaned:

* packages/web-components/internal/vendor/@carbon/web-components

When switching to a branch that does not produce the above artifact, problems arise in the build process:

```
$ yarn build
[...]
@carbon/ibmdotcom-web-components: src/internal/vendor/@carbon/web-components/components/checkbox/checkbox.d.ts(315,15): error TS1005: ')' expected.
@carbon/ibmdotcom-web-components: src/internal/vendor/@carbon/web-components/components/checkbox/checkbox.d.ts(315,20): error TS1109: Expression expected.
@carbon/ibmdotcom-web-components: src/internal/vendor/@carbon/web-components/components/checkbox/checkbox.d.ts(315,33): error TS1011: An element access expression should take an argument.
@carbon/ibmdotcom-web-components: src/internal/vendor/@carbon/web-components/components/checkbox/checkbox.d.ts(315,34): error TS1005: ';' expected.
@carbon/ibmdotcom-web-components: src/internal/vendor/@carbon/web-components/components/checkbox/checkbox.d.ts(315,36): error TS1128: Declaration or statement expected.

[...hundreds of errors clipped..]

@carbon/ibmdotcom-web-components: [15:59:46] 'build:modules:types' errored after 5.37 min
@carbon/ibmdotcom-web-components: [15:59:46] Error: TypeScript: Compilation failed
@carbon/ibmdotcom-web-components:     at Output.mightFinish (/home/matt/projects/carbon-for-ibm-dotcom/node_modules/gulp-typescript/release/output.js:130:43)
@carbon/ibmdotcom-web-components:     at /home/matt/projects/carbon-for-ibm-dotcom/node_modules/gulp-typescript/release/output.js:65:22
```

The above errors occur even after a `yarn reset` (which includes a `yarn clean`)

### Component(s) impacted

All components in the `@carbon/web-components` package.

### Carbon for IBM.com version

main and feat/masthead-v2.1

### Severity

Severity 4 = The problem is not visible to or noticeable to an average user. Affects minor functionality, no workaround needed.

### Application/website

https://carbon-design-system.github.io/carbon-for-ibm-dotcom/canary/web-components/?path=/story/overview-getting-started--page

### Package

@carbon/web-components

### CodeSandbox example

[...]

PR #10038, opened 2023-02-11
fix(clean): clean vendor src base dir from web-components package

m4olivei:
### Related Ticket(s)

Resolves https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/10037

### Description

Cleans up the artifacts produced during `yarn build` within `packages/web-components/internal/vendor` to avoid build errors associated with these artifacts encountered during `yarn build` when switching between major feature branches (`main` and `feat/masthead-v2.1`).

### Changelog

**Changed**

- `yarn clean` updated for the web-components package to clean up the vendor src dir artifact produced during build.


jkaeser (approved):
Looks like a good change to me. +1

andy-blum (commented):
I think this is all probably fine, but we should definitely get the go-ahead from @kennylam.

kennylam (approved):
LGTM thanks @m4olivei!

Issue #9607, opened 2022-10-31
[masthead v2]: Improve test coverage


andy-blum:
**Engineering info:**  
- Innovation Team engineer: @m4olivei 
- Innovation Team JIRA Tickets: https://jsw.ibm.com/browse/ADCMS-2948
- Consulting engineer: @andy-blum 

-----

### The problem

The dds-masthead v2 test coverage could use some improving:
- Some of the tests take a long time to run individually. We should look into making them more efficient.
- We currently have no test coverage for lazy-loading/DOM-pruning of subcomponents.

### The solution

Improve e2e test coverage of masthead-v2

### Application/website

C4IBM

### Business priority

Low Priority = release date is not dependent on fix or not upcoming

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #10031, opened 2023-02-09
chore(masthead-v2): update dev branch with main changes

andy-blum:
### Related Ticket(s)

[ADCMS-2944](https://jsw.ibm.com/browse/ADCMS-2944)

### Description

Marges changes from `main` into `masthead-v2-dev`

PR #10053, opened 2023-02-15
Fix e2e tests on feat/masthead-v2.1

m4olivei:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/9607

### Description

Fixes e2e Cypress tests on the `feat/masthead-v2.1` brach. Adjusting for changes that have been made recently.

### Changelog

**Changed**

- e2e Cypress tests fixed to get to a passing state for `feat/masthead-v2.1`
- Add analytics attributes for header links
- Fix CLI option support for unit tests to allow test isolation to an individual test spec


.github/workflows/e2e-tests.yml
```diff
@@ -2,9 +2,9 @@ name: e2e-tests
 
 on:
   push:
-    branches: [ main, release/*, feat/masthead-v2 ]
+    branches: [ main, release/*, feat/masthead-v2, feat/masthead-v2.1 ]
   pull_request:
-    branches: [ main, release/*, feat/masthead-v2 ]
+    branches: [ main, release/*, feat/masthead-v2, feat/masthead-v2.1 ]
```
m4olivei, line 7: Addded so we get the CI check on this PR, not sure if we want to keep it. Not sure about the lifespan of this `feat/masthead-v2.1`, sounds like it's only matter of time before its merged down into `feat/masthead-v2-dev`. I imagine I'll also need to make adjustments here once the [merge conflict resolution happening over here](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10031) makes it to `feat/masthead-v2.1`. We can adjust this once it's closer to ready.

m4olivei:
Yay! Green https://github.com/carbon-design-system/carbon-for-ibm-dotcom/actions/runs/4196030671/jobs/7276426172

m4olivei:
Latest run was a relatively decent result [in terms of timing](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/actions/runs/4198116095/jobs/7281327104). Here's how it stacks up to other runs:

There is still something holding up the `masthead-v2.e2e.js` test and making it take a long time. I _sometimes_ see Cypress hang locally on the [specific test that is taking so long](https://github.com/m4olivei/carbon-for-ibm-dotcom/blob/chore/masthead-tests/packages/web-components/tests/e2e-storybook/cypress/integration/masthead/masthead-v2.e2e.js#L132). This test is making sure all links have the relevant analytics attributes. There are a lot of links to check, all follow the same patterns for each group that gets checked. I think we can shave off most of the time here by just checking the first 4 in each group, which would be representative of the whole and avoid checking all of them, should be a nice compromise.

Also of note is that even if we did get that time down, a lot of the time here goes to install and build steps. There is lots of opportunity there to [leverage caching ](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows). I found this [open PR](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/9080) from @IgnacioBecerra from last year that looked to tackle not only caching but also parallelizing the tests. It's been open for awhile, maybe if we split the two, we could get land caching as one improvement and split paralliation to a follow up. Just a thought.

In the meantime, @andy-blum , feel free to take this PR as is into your work to [consolidate masthead dev branches with upstream work](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10031). 

| Branch  | Latest run | Timing |
| ------------- | ------------- | ------------- |
| `main`  | [run](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/actions/runs/4159598893/jobs/7195799585) | 1h 35m 12s  |
| `feat/masthead-v2`  | [run](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/actions/runs/3566309154/jobs/5992578124) (failing)  | 1h 50m 41s  |
| `feat/masthead-v2-dev`  | [run](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/actions/runs/4118066549/jobs/7110134568)  | 1h 39m 43s  |
| `feat/masthead-v2.1`  | n/a (tests not previously run here)  | n/a  |
| `m4olive:chore/masthead-tests`  | [run](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/actions/runs/4198116095/jobs/7281327104)  | 1h 5m 1s :sparkles:  |

jkaeser (approved):
All these changes look great to me. Thanks for putting this together, @m4olivei!

What do you think about the following plan: we merge this into `feat/masthead-v2.1`, then re-merge that into @andy-blum's https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10031? cc @proeung

proeung (approved):

PR #10055, opened 2023-02-16
fix(button): add button print styles

ariellalgilmore:
### Related Ticket(s)

N/A
### Description

Button-group has specific print styles and in expressive-modal it just uses `button-expressive` causing it to add a `:` without also having the print styles for button-expressive
Screen Shot 2023-02-16 at 8 37 04 AM

after:
Screen Shot 2023-02-16 at 8 41 12 AM

### Changelog

**Changed**

- moved DDSButtonExpressive print styles out of button-group and into button


annawen1 (commented):
Does this affect how the buttons look when printed for folks not using button group - just plain CWC button?

ariellalgilmore:
> Does this affect how the buttons look when printed for folks not using button group - just plain CWC button?

yeah that's what is being fixed. So now plain buttons or button-groups being printed should show the html information

annawen1 (approved):
LGTM!

PR #10070, opened 2023-02-20
feat(masthead-l1): update data, render top-level items

andy-blum:
### Related Ticket(s)

Precursor to [ADCMS-2739](https://jsw.ibm.com/browse/ADCMS-2739), [ADCMS-2911](https://jsw.ibm.com/browse/ADCMS-2911)

### Description

- Updates masthead-composite logic to display L1 data in the L1 region.
- Updates the L1 data to match the figma navigation files
- Splits rendering between two viewport-width-conditional functions to avoid double rendering the menu in the DOM

This will not currently render any dropdowns, but the L1 does switch between a desktop & mobile render method

### Changelog

**Changed**

- Updated masthead-v2 L1 menu to render new data and avoid double rendering in the DOM


packages/web-components/src/components/masthead/__stories__/links.ts
```diff
@@ -397,73 +397,201 @@ const l1Data: LegacyMastheadL1 = {
...
+  actions: {
+    login: {
+      title: 'Log in',
+      url: 'https://example.com',
+    },
+    cta: {
+      title: 'Optional CTA',
+      url: 'https://example.com',
+    },
+  },
```
m4olivei, line 594: Just more curiosity / edification. Does this structure get validated at some point? I see there are Typescript types for it, is Typescript validating it and going to throw errors at some point if it's off?

m4olivei (approved):
This makes sense to me. Looks good.

packages/web-components/src/components/masthead/__stories__/links.ts
proeung: @m4olivei @andy-blum There's already an `eslint` and `prettier` configured within the `ibm-dotcom-translations` (see - https://github.ibm.com/webstandards/ibm-dotcom-translations/blob/master/package.json#L14) that validates these JSON files and part of the existing CI checks.

proeung (approved):
LGTM! @andy-blum Not a blocker, but we'll need to adjust the README portion of the masthead to include the new data types for adopting the L1 Nav v2 (see - https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/47e2d7a2201ccd46d6abc68c58bf0ccd278be917/packages/web-components/src/components/masthead/__stories__/README.stories.mdx#using-l1-nav)

@kennylam @annawen1 As you might already be aware, we're adjusting the data structure of the L1 Nav to accommodate the new design that's part of the Masthead v2 initiative. This means that adopters will need need to switch to the v2 codebase in order to be compatible with the new design.

proeung:
```mermaid
%%{init: { 'theme': 'base', 'gitGraph': {'showBranches': true, 'showCommitLabel':false,'mainBranchName': 'feat/masthead-v2'}} }%%

gitGraph
   commit
   commit
   branch feat/masthead-v2-dev
   checkout feat/masthead-v2-dev
   commit
   branch feat/new-feature
   checkout feat/new-feature
   commit
   commit
   checkout feat/masthead-v2-dev
   commit tag: "hotfix"
   checkout feat/masthead-v2
   merge feat/masthead-v2-dev
   checkout feat/new-feature
   commit
   checkout feat/masthead-v2-dev
   merge feat/new-feature
   commit
   checkout feat/masthead-v2
   merge feat/masthead-v2-dev
```

Also, for the L1 Nav updates, I'm starting to wonder if using the current branching model of pushing everything into `feat/masthead-v2-dev` makes sense, especially since we're making structural changes that might affect the display of the L1 Nav on AEM pages that will eventually adopt the masthead CDN build file that includes changes from `v2.1`/`feat/masthead-v2`, unless as part of that roll out that we recommend that AEM adjust their L1 structure as well. I'm open to suggestions/ideas.

Let's validate this concern. @andy-blum Can you plug in this data (https://codesandbox.io/s/lexu76?file=/src/l1-nav-data.js) and see if the dropdown menu items are still displaying correctly? This is the data set that is used in AEM and just wanted to make sure it's still compatible.

andy-blum:
> unless as part of that roll out that we recommend that AEM adjust their L1 structure as well.

This is what I was anticipating, as we don't want to be on the hook for maintaining multiple versions of the L1 like we currently do for the L0 masthead.

packages/web-components/src/components/masthead/__stories__/links.ts
m4olivei: @proeung I was more asking from a data structure / data types / object shape POV.

jkaeser: @m4olivei Like React's prop-types, right? The components don't do any of that sort of validation, no.

That can change, though! TypeScript can't help us here since we're already executing the compiled JS by the time we get data, right? But after some quick looking, I found there are some tools out there that can help us automatically generate validator functions based on our types. I wonder if this would have performance implications due to the sheer size of the masthead and our expected data payloads...

We could also go the route @andy-blum considered and validate the data generated by the translations repo script. This has the benefits of automatically verifying we aren't sending bad data from the content endpoint, but our test app that gets built as part of CI tooling already helps us verify the data is good, albeit manually.

In either case, this sounds like a great candidate for that list of improvements you've been adding to 🙂

andy-blum: One thing to consider is that right now, we aren't generating L1 data, so it's on the consumers to shape their data correctly.

m4olivei: > Like React's prop-types, right?

I wasn't thinking of that, but yeah something like that would be useful for consumers of the component. I was thinking alont the lines of validation in the translations repo like @andy-blum mentioned. Although, per @andy-blum 's last comment, I realise now that L1 data doesn't live in the translations repo.

> That can change, though! TypeScript can't help us here since we're already executing the compiled JS by the time we get data, right?

Right, makes sense.

> But after some quick looking, I found there are some tools out there that can help us automatically generate validator functions based on our types. I wonder if this would have performance implications due to the sheer size of the masthead and our expected data payloads.

The way React prop types work, they only get used in development and are stripped away in production, so the performance impact is mitigated. Would be interesting if this could be the case for our use case as well.

> We could also go the route @andy-blum considered and validate the data generated by the translations repo script. This has the benefits of automatically verifying we aren't sending bad data from the content endpoint.

That would be neat and give us some warm fuzzies about the output of the script. There should be a way to take our Typescript types and validate the generated translation files against them. Perhaps using [JSON schema](https://json-schema.org/) and a tool that [generates JSON schema from Typescript types](https://github.com/YousefED/typescript-json-schema).

> In either case, this sounds like a great candidate for that list of improvements you've been adding to slightly_smiling_face

Added :).

Issue #9630, opened 2022-11-09
Reveal the HTML elements and move them outside the "shadow-root" while constructing the web components.


mseth24:
### Problem Statement-

1. Adobe Target can edit/modify Standard HTML Components
2. Adobe Target cannot edit any component within the Shadow DOM

Adobe Target works using CSS Selectors. Elements within the "Shadow-root" node are not accessible using CSS selector and hence, not editable in Adobe Target. However, if an  HTML element is revealed using a slot, it becomes accessible using a CSS selector and can be edited in Adobe Target.

### Example Component-
**dds-link-list-item-card**
In the dds-link-list-item-card component, "p" tag  is revealed using a slot, and is editable in Adobe Target using a selector.

**dds-card-cta-footer**
"a" Tag is placed inside the "shadow-root", and hence not accessible/editable in Adobe Target.

Adobe Target cannot directly edit custom web-component (Any DDS component) and it can not see elements inside the "Shadow-root". Hence, updating the first standard parent HTML element is the only way to modify an element that is placed inside the "Shadow-root". 

In certain scenarios, we have to update almost the entire page (for example, to change any menu item in the Table of Content) to make minor edits on the page.

### The solution

Reveal the HTML elements and move them outside the "shadow-root" while constructing the web components. This change will enable Adobe Target to edit the element using a selector easily.

**dds-link-list-item-card**
This element uses both dev style. Few HTML elements are revealed  using slots, whereas a few others are  hidden inside shadow-root.

**Frequently Used Components-**
dds-table-of-contents
dds-link-list-item-card

### Application/website

www.ibm.com

### Business priority

Medium Priority = upcoming release but is not pressing

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

annawen1:
@mseth24 Looking at the possibility of moving elements out of the shadow DOM and into slots, I think the approach is not feasible. Two main reasons:

[...]

PR #10080, opened 2023-02-21
chore(customElementRegistry): add carbonElement decorator

andy-blum:
### Related Ticket(s)

Closes #9940 

### Description

This PR takes a new approach, and offers some advantages of #9941.

While the other PR aims to separate class definitions from their registration in the browser, and should work well for teams or projects that wish to _extend_ C4IBM components, it does nothing to solve the problem for teams that are _double-consuming_ C4IBM components. For example, consider the widget team:

- They are building widgets using the react-wrapper version of the components
- When the react component imports the element, it brings along its `customElements.define()` method
- The widget is placed on a page that is already using the C4IBM CDN links, which _also_ bring along the `customElements.define()` method.
- Both the widget script and the CDN link are rolled up into separate scripts and whichever one runs second will encounter the unexpected error, not know how to handle that problem, and halt execution of the script, likely leaving large pieces of setup functionality unprocessed

This new approach offers a method for elements to _attempt_ to define themselves, but to gracefully handle the failure and to continue executing the remainder of the running script. The new decorator is a simple copy/paste from the LitElement `@customElement` decorator, with the `customElements.define` functions wrapped in a try/catch block.

As of right now, this only impacts one component and is simply a proof-of-concept.

### Changelog

**New**

- Adds a `@carbonElement()` decorator to the carbon-web-components package

**Changed**

- Edit the carbon-web-components package's accordion-item component to use @carbonElement

PR #10087, opened 2023-02-23
feat(stable-selector): dispatch event with shadow DOM

IgnacioBecerra:
### Related Ticket(s)
None.

### Description
Due to AEM having a hard time accessing the shadow DOM, this PR creates a new event that dispatches on component render if any component has a `get-shadow-dom` attribute, containing the shadowDOM as an event detail. This will make it easier to retriever and modify the shadow DOM through the program by listening to the events on load.

```
<dds-card get-shadow-dom></dds-card>
```

in document:

```
  window.addEventListener('dds--card', (e) =>
    console.log((e as CustomEvent).detail.shadowDOM)
   // do stuff with it
  );
```

As an example, you can open the console in the default Card story to see that the shadow DOM element will be retrieved through listening to the event. We could also add more data to be sent through this event, if needed. Likewise, we could limit the elements retrieved through the event by scoping the query instead of the whole shadow DOM (e.g. only retrieving `.bx--card__wrapper` instead).

Another idea is to add event listeners to certain components that would directly modify certain elements, so when a user dispatches events, the component will modify things directly on the component's side. 

```
          exampleCard.dispatchEvent(
            new CustomEvent('edit-heading', {
              bubbles: true,
              composed: true,
              detail: {
                elementTarget: 'dds-card-heading',
               newText: 'this is a new heading'
              },
            })
          );
```

While this could probably be "easier" on adopters, there might not be a way to generalize this behavior, and worst case scenario we'd have to do this on a case by case basis, which we'd have to take into account if there are ever breaking changes.

The first solution might be a more general one, and provides the shadow DOM to the adopters.

### Changelog

**New**

- {{new thing}}

**Changed**

- {{changed thing}}

**Removed**

- {{removed thing}}

### DIFF
--- a/packages/web-components/src/components/card/__stories__/card.stories.ts
+++ b/packages/web-components/src/components/card/__stories__/card.stories.ts
@@ -41,9 +41,14 @@ export const Default = (args) => {
     footer,
     cardStyles,
   } = args?.Card ?? {};
+
+  window.addEventListener('dds--card', (e) =>
+    console.log((e as CustomEvent).detail.shadowDOM)
+  );
   /* eslint-disable no-nested-ternary */
   return html`
     <dds-card
+      get-shadow-dom
       color-scheme=${cardStyles === 'Inverse card'
         ? 'inverse'
         : cardStyles === 'Outlined card'
--- a/packages/web-components/src/globals/mixins/stable-selector.ts
+++ b/packages/web-components/src/globals/mixins/stable-selector.ts
@@ -1,7 +1,7 @@
 /**
  * @license
  *
- * Copyright IBM Corp. 2020, 2022
+ * Copyright IBM Corp. 2020, 2023
  *
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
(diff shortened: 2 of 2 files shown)


andy-blum:
> Due to AEM having a hard time accessing the shadow DOM

I'm curious what you mean by this. Is the AEM software having a hard time accessing this or is it the AEM developers? Why is that information needed?

Also do they need a reference to every _instance_ of the Shadow DOM or just a vague idea of what the render functions put out? If you only need one per class, that kind of information could get wrapped up when the component is defined as part of #10080

jeffchew:
> I'm curious what you mean by this. Is the AEM software having a hard time accessing this or is it the AEM developers? Why is that information needed?
> 
> Also do they need a reference to every _instance_ of the Shadow DOM or just a vague idea of what the render functions put out? If you only need one per class, that kind of information could get wrapped up when the component is defined as part of #10080

This is in reference to a request from the Adobe Target team. This issue was originally brought up and closed:
https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/9630

This request is being brought up again.

Issue #10088, opened 2023-02-23
Date picker (range with calender)  -FROM date gets cleared on it own after selecting range


smitan1394:
### Description

The FROM date goes away (cleared) after selecting range, and moving cursor out of focus.

### Component(s) impacted

Date picker (range with calender) 

### Browser

Firefox

### Carbon for IBM.com version

latest

### Severity

Severity 1 = The design is broken in a critical way that blocks users from completing tasks or damages the brand. Affects major functionality, no workaround.

### Application/website

https://psa-sit.dal2a.ciocloud.nonprod.intranet.ibm.com/university/power/power-cloud-resource-request.html

### Package

@carbon/web-components

### CodeSandbox example

https://codesandbox.io/s/beautiful-gates-8nbj4b?file=/cdn.html

### Steps to reproduce the issue (if applicable)

select range, click outside the component, see FROM date disappear

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

ariellalgilmore:
PR opened to downgrade to 4.6.9. following suit with carbon react: https://github.com/flatpickr/flatpickr/issues/2767

PR #10168, opened 2023-02-27
fix(date-picker): range mode clearing first value

ariellalgilmore:
### Related Ticket(s)

#10088 

### Description

When a user clicked a range of calendar dates the first date would get cleared after clicking the second date.

Before:
Feb-27-2023 10-24-46

After:
Feb-27-2023 10-22-48

### Changelog

**Changed**

- downgraded flatpick to 4.6.9 ( the issue starts to appear in version 4.6.10)


annawen1 (approved):
LGTM!

Issue #10026, opened 2023-02-07
[Masthead]: Side nav focus wrap not working with Lit v2


IgnacioBecerra:
### Description

After upgrading to using Lit v2, the side nav focus wrap seen in the Masthead is broken. It currently uses the `focuswrap` helper function. 

The possible cause behind this is the way Click events are being detected, since the Masthead container is apparently the only event target received when detecting such events, instead of the actual target clicked on.

This also causes a slight side effect where clicking outside of the nav does not close the menu, as it used to.

### Component(s) impacted

Masthead

https://user-images.githubusercontent.com/24970122/214748294-adbd03b6-a2db-46b1-a13a-4c05703c9547.mov

### Browser

Chrome

### Carbon for IBM.com version

lit-v2 upgrade branch

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

N/A

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://carbon-design-system.github.io/carbon-for-ibm-dotcom/canary/web-components/?path=/story/components-masthead--default

### Steps to reproduce the issue (if applicable)

1. Open the Masthead in mobile viewport to get the side nav button
2. Activate the side nav
3. Keep pressing tab until the focus goes outside the side nav.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #10171, opened 2023-02-27
fix(masthead): left nav focuswrap

ariellalgilmore:
### Related Ticket(s)

#10026 
### Description

Similar to the updates made here: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10019

because of the updates to v11 we have to target the event.target differently to get the appropriate value
Feb-27-2023 13-16-57

### Changelog

**Changed**

- updated the else if statement to make sure the right values are getting passed

### DIFF
--- a/packages/web-components/src/components/masthead/left-nav.ts
+++ b/packages/web-components/src/components/masthead/left-nav.ts
@@ -84,7 +84,6 @@ class DDSLeftNav extends StableSelectorMixin(BXSideNav) {
       (event.composedPath()[1] as HTMLElement).tagName ===
       selectorButtonToggle.toUpperCase()
     ) {
-      console.log('ELSE IF');
       const { comparisonResult } = event.detail;
       const {
         selectorExpandedMenuSection,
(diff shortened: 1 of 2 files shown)


packages/web-components/src/components/masthead/left-nav.ts
```diff
@@ -80,7 +80,11 @@ class DDSLeftNav extends StableSelectorMixin(BXSideNav) {
       if (toggle) {
         (toggle as HTMLElement).focus();
       }
-    } else if ((event.target as HTMLElement).matches?.(selectorButtonToggle)) {
+    } else if (
+      (event.composedPath()[1] as HTMLElement).tagName ===
+      selectorButtonToggle.toUpperCase()
+    ) {
+      console.log('ELSE IF');
```
annawen1, line 87: do we need this log?

ariellalgilmore: oops good catch lol

IgnacioBecerra (approved):
@ariellalgilmore LGTM! I'm relieved this didn't need a big refactoring 😌

annawen1:
Noticing when I tap into the next level of the nav, the focus on the back button isn't there so I end up stuck, if it's not an easy  fix maybe we can open up another ticket to track it instead

CURRENT:
https://user-images.githubusercontent.com/54281166/221886112-e144ab78-6bc8-47fa-991f-e988ae229092.mov

DEPLOY PREVIEW:
https://user-images.githubusercontent.com/54281166/221886105-ea59c834-38f6-4d20-b591-a6328509da81.mov

ariellalgilmore:
@annawen1 should be fixed now!

annawen1 (approved):
Just tested on iPhone VO, working nicely!

Issue #9861, opened 2022-12-28
[Lead Space Search]: QA: Search edit combo and the clear button is not defined within in a search region


gopinathan-n:
### Description

Page Element:
The search edit field and the clear button

Issue Description:
The search edit combo and the clear button is not defined within in a search region

Recommend Fix:
The search edit combo and the clear button can be given in search region

User Impact:
When the page content is not covered within a labelled landmark region, screen reader users are unable to skim through the page and understand its structure.

Test Tool:
JAWS FF
Voiceover safari iPhone

### Component(s) impacted

Lead space search
1. Default
2. with image
(web component canary and web component staging)

### Browser

Safari, Firefox

### Carbon for IBM.com version

v1.40.0-rc.0

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

Carbon for IBM.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

Web component Canary https://carbon-design-system.github.io/carbon-for-ibm-dotcom/canary/web-components Web component staging https://carbon-design-system.github.io/carbon-for-ibm-dotcom/next/web-components

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #10173, opened 2023-02-27
fix(search-with-typeahead): add search region for accessibility

annawen1:
### Related Ticket(s)

[Lead Space Search]: QA: Search edit combo and the clear button is not defined within in a search region #9861

### Description

Add "search" role to the `search-with-typeahead` form component 

### Changelog

**New**

- add search `role` to `form` component of `search-with-typeahead`


kennylam (approved):
LGTM!

Issue #9858, opened 2022-12-28
[Lead Space Block]: QA: Heading level is not in an hierarchical order


gopinathan-n:
### Description

Page Element:
heading level order

Issue Description:
Heading level is not in an hierarchical order. found the heading level order as h1,h2 and h4 instead of H1, H2, H3 …

Recommend Fix:
As an accessibility best practice, Heading level should be in hierarchical order. Heading level should not be skipped. 

User Impact:
If page headings are not in the right sequence, screen reader users find it difficult to comprehend the hierarchical structure of the page content. 

Test Tool:
JAWS FF
Voiceover safari iPhone

### Component(s) impacted

Lead space Block
1. Default
2. with video
(web component canary and web component staging)

### Browser

Safari, Firefox

### Carbon for IBM.com version

v1.40.0-rc.0

### Severity

Severity 4 = The problem is not visible to or noticeable to an average user. Affects minor functionality, no workaround needed.

### Application/website

Carbon for IBM.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

Web component Canary https://carbon-design-system.github.io/carbon-for-ibm-dotcom/canary/web-components Web component staging https://carbon-design-system.github.io/carbon-for-ibm-dotcom/next/web-components

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [ ] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #10177, opened 2023-02-28
chore(leadspace-block): adjust heading level for leadspace-block story

annawen1:
### Related Ticket(s)

[Lead Space Block]: QA: Heading level is not in an hierarchical order #9858

### Description

Heading levels are not in order for the Leadspace Block story, currently it is from H1, H2, H4 instead of H1, H2, H3

This PR sets the `aria-level` for the `link-list-heading` so the headings are in hierarchical order instead. Currently by default, the `aria-level` for `link-list-heading` is set to 4.

Screen Shot 2023-02-28 at 11 44 39 AM

### Changelog

**Changed**

- set `aria-level` to `3` for `link-list-heading` in the story


packages/web-components/src/components/leadspace-block/__stories__/leadspace-block.stories.ts
```diff
@@ -38,7 +38,9 @@ const image = html`
 
 const linkList = html`
   <dds-link-list type="end">
-    <dds-link-list-heading>Featured products</dds-link-list-heading>
+    <dds-link-list-heading aria-level="3"
```
IgnacioBecerra, line 41: @annawen1 I'm wondering if it would be better to dynamically setting this within the `leadspace-block` component instead of in the story? I'm assuming the issue will remain if an adopter uses the same layout from our story, so maybe we could loop through the slotted elements of the `leadspace-block`, count how many `aria-heading`s we have, and make sure any wrong levels are adjusted. Or if the `link-list-heading` the only affected component, we could make it always be `aria-level=3` when used within `leadspace-block`.

IgnacioBecerra: Ignore all this if this bug is a Storybook only bug though!

annawen1: Hmmm my concern is adding all that logic can make the component less flexible, but if `link-list` will always be part of `leadspace-block` then setting the `aria-level` within `leadspace-block` makes sense

annawen1: According to the specs, the `link-list` is required so I'll adjust the `aria-level` for `link-list` within the component!

IgnacioBecerra (approved):
LGTM, thanks @annawen1!

Issue #9869, opened 2022-12-28
[Universal banner]: QA: Landmark region should be defined for Universal banner


gopinathan-n:
### Description

Page Element:
Landmark Region for Universal banner

Issue Description:
If it is supposed to be used as banner region, this should be given within A banner role or HTML header tag that is equivalent for banner role.

User Impact:
When the page content is not covered within a labelled landmark region, screen reader users are unable to skim through the page and understand its structure.

Test Tool:
JAWS FF
Voiceover safari iPhone

### Component(s) impacted

Universal banner
(web component canary and web component staging)

### Browser

Safari, Firefox

### Carbon for IBM.com version

v1.40.0-rc.0

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

Carbon for IBM.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

Web component Canary https://carbon-design-system.github.io/carbon-for-ibm-dotcom/canary/web-components Web component staging https://carbon-design-system.github.io/carbon-for-ibm-dotcom/next/web-components

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [ ] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #10179, opened 2023-02-28
fix(universal-banner): add banner role

annawen1:
### Related Ticket(s)

[Universal banner]: QA: Landmark region should be defined for Universal banner #9869

### Description

Set banner `role` for `universal-banner` component for accessibility purposes

### Changelog

**Changed**

- add banner `role` onto `universal-banner`


kennylam (approved):
LGTM!

PR #10186, opened 2023-03-03
chore(e2e): fixes `e2e-integration` tests

kennylam:
### Description

This fixes the `e2e-integration` GH actions that are failing for `react` and `web-components`. With the upgrade to Yarn v3, building the Codesandbox examples needed for e2e tests becomes problematic as they have not been upgraded. The simplest solution is to switch back to an older version of Yarn for the test builds only.

### Changelog

**New**

- Use `execa` for `web-components` build process. Now that `vendor/@carbon/web-components` are bundled with `ibmdotcom-web-components`, the packed file is huge. This exceeds the `maxBuffer` limit in `execSync`, while `execa` has no such issue.

**Changed**

- update `dotcom-shell` and `masthead` e2e tests

**Removed**

- {{removed thing}}

### DIFF
--- a/.github/workflows/e2e-integration.yml
+++ b/.github/workflows/e2e-integration.yml
@@ -25,7 +25,7 @@ jobs:
           node-version: ${{ matrix.node-version }}
           cache: 'yarn'
       - name: Install dependencies
-        run: yarn install
+        run: yarn install --immutable --immutable-cache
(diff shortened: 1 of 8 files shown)


emyarod (approved):
looks good to me!

RichKummer (approved):

PR #10209, opened 2023-03-08
fix(rolloup): point to cwc internal vendor components for custom babel plugin

annawen1:
### Related Ticket(s)

N/A

### Description

Edit roll up config for custom babel plugin that prevents the `customElements.define()` from happening for the cwc components - look for the internal vendor carbon web component components instead to prevent the custom registry error 

### Changelog

**Changed**

- point to cwc internal vendor components for custom babel plugin

### DIFF
--- a/packages/web-components/tools/get-rollup-config.js
+++ b/packages/web-components/tools/get-rollup-config.js
@@ -1,7 +1,7 @@
 /**
  * @license
  *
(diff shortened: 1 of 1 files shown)


ariellalgilmore (approved):
LGTM!

kennylam (approved):

PR #10218, opened 2023-03-10
fix(accordion): misisng animation styles

ariellalgilmore:
### Related Ticket(s)

N/A
### Description

added missing animation styles on open and close

added back the styles in storybook... otherwise they overwrite all slotted `p` styles, but wasn't exactly sure what was broken before? 

- what this pr shows
Screen Shot 2023-03-09 at 7 24 59 PM

- how it shows with previous container styles
Screen Shot 2023-03-09 at 7 26 08 PM

### Changelog

**New**

- {{new thing}}

**Changed**

- {{changed thing}}

**Removed**

- {{removed thing}}

### DIFF
--- a/packages/carbon-web-components/.storybook/_container.scss
+++ b/packages/carbon-web-components/.storybook/_container.scss
@@ -13,16 +13,9 @@
 @use '@carbon/styles/scss/fonts';
 @use '@carbon/styles/scss/theme' as *;
 @use '@carbon/styles/scss/themes' as *;
+@use '@carbon/styles/scss/type' as *;
 
-@import '@carbon/type/scss/type';
-@import '@carbon/type/scss/font-face/mono';
-@import '@carbon/type/scss/font-face/sans';
-@import '@carbon/type/scss/font-face/serif';
-
-@include carbon--type-reset();
-@include carbon--font-face-mono();
-@include carbon--font-face-sans();
-@include carbon--font-face-serif();
+@include reset();
 
 // The default theme is "white" (White)
 :root {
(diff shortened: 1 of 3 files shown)


packages/carbon-web-components/.storybook/_container.scss
```diff
@@ -7,7 +7,22 @@
...
+
+@import '@carbon/type/scss/type';
+@import '@carbon/type/scss/font-face/mono';
+@import '@carbon/type/scss/font-face/sans';
+@import '@carbon/type/scss/font-face/serif';
+
+@include carbon--type-reset();
+@include carbon--font-face-mono();
+@include carbon--font-face-sans();
+@include carbon--font-face-serif();
```
kennylam, line 25:
```suggestion
@use '@carbon/styles/scss/config' as *;
@use '@carbon/styles/scss/utilities/convert' as *;
@use '@carbon/styles/scss/utilities';
@use '@carbon/styles/scss/fonts';
@use '@carbon/styles/scss/theme' as *;
@use '@carbon/styles/scss/themes' as *;
@use '@carbon/styles/scss/type' as *;

@include reset();
```

If importing all the styles is conflicting with component-specific styles, then we should just import type styles from `@carbon/styles`. All the `import`/`include` are from v10. The v10 styles were also interfering with the v11 reset styles, that's why I made that initial change.

ariellalgilmore: that works! thanks @kennylam

kennylam (approved):
LGTM thanks @ariellalgilmore!

PR #10231, opened 2023-03-14
fix(notificaiton/breadcrumb): update styles/remove @extend

ariellalgilmore:
### Related Ticket(s)

N/A
### Description

with update to `sass` we need to import 

### Changelog

**Changed**

- added styles from react package not coming through


annawen1 (approved):
LGTM!

PR #10237, opened 2023-03-16
fix(modal): styles with sass upgrade

ariellalgilmore:
### Related Ticket(s)

N/A
### Description

With the sass upgrade we need to pull in aditional modal styes when modal is open

### Changelog

**Changed**

- opacity: 1;
  visibility: inherit; to modal styles


RichKummer (approved):
Just a heads up that this PR seems to remove the transition animation for when the modal appears/disappears from the screen. I approved in case this is a separate issue!

https://user-images.githubusercontent.com/15643582/225676702-05dcac74-cde7-40a2-bbc8-9c2620b269f2.mov

Issue #10082, opened 2023-02-21
[tile]: sync with @carbon/react version 


IgnacioBecerra:
- Missing Default with Layer
- Missing Clickable with Layer
- Missing Expandable with Interactive
- Missing Expandable with Layer
- Missing Radio with Layer
- Radio functionality needs a rework

PR #10238, opened 2023-03-16
feat(button/tag/tile/toggle): syncing components

IgnacioBecerra:
### Related Ticket(s)
#10082
#10083
#10089
#10124

### Description
This PR syncs the Button, Tag, Tile, and Toggle components. 

In order for the Tag color theme variations, the `scss` file for the Storybook container had to be changed to match with Carbon React's Storybook to simply import `@carbon/styles`. Without this, the themes lack the proper tag color tokens.

### Changelog

**New**

- new tile group selection management process
- new tile features
- button set component
- button tooltip feature
- new tag theme colors
- new small toggle

**Changed**

- component attributes 

**Removed**

- unneeded radio tile group manager
- filter tag, feature moved to regular tag


packages/carbon-web-components/src/components/tag/defs.ts
```diff
@@ -12,9 +12,9 @@
  */
 export enum TAG_SIZE {
   /**
-   * Regular size.
+   * Default size.
    */
-  REGULAR = '',
+  DEFAULT = 'md',
```
ariellalgilmore, line 17:
```suggestion
  MEDIUM = 'md',
```

packages/carbon-web-components/src/components/tag/tag.ts
```diff
@@ -18,21 +22,84 @@ export { TAG_SIZE, TAG_TYPE };
...
+   */
+  @property({ type: Boolean, reflect: true })
+  open = true;
+
   /**
    * The size of the tag.
    */
   @property({ reflect: true })
-  size = TAG_SIZE.REGULAR;
+  size = TAG_SIZE.DEFAULT;
```
ariellalgilmore, line 102:
```suggestion
  size = TAG_SIZE.MEDIUM;
```

packages/carbon-web-components/src/components/tag/tag-story.ts
```diff
@@ -13,21 +13,46 @@ import { boolean, select } from '@storybook/addon-knobs';
...
 import { TAG_SIZE, TAG_TYPE } from './tag';
-import './filter-tag';
+import './index';
 import storyDocs from './tag-story.mdx';
 import { prefix } from '../../globals/settings';
 
-const noop = () => {};
-
 const sizes = {
   'Regular size': null,
```
ariellalgilmore, line 21:
```suggestion
  [`Medium size (${TAG_SIZE.MEDIUM})`]: TAG_SIZE.MEDIUM,
```

packages/carbon-web-components/src/components/tag/tag.ts
```diff
@@ -41,10 +108,36 @@ class BXTag extends LitElement {
...
   render() {
-    return html` <slot></slot> `;
+    const { disabled, filter } = this;
+    return html`
+      <slot></slot>
+
+      ${filter
+        ? html`
+            <button class="${prefix}--tag__close-icon" ?disabled=${disabled}>
+              ${Close16({ 'aria-label': this.title })}
```
ariellalgilmore, line 118: should also be able to render their own icon: 
Screen Shot 2023-03-20 at 12 55 55 PM

ariellalgilmore (commented):
Tag looking good! just missing the ability to renderIcon too

packages/carbon-web-components/src/components/button/button-story.ts
```diff
@@ -9,177 +9,320 @@
...
+  } = args?.[`${prefix}-btn`] ?? {};
   return html`
-    <cds-btn-skeleton
-      ?disabled=${disabled}
-      size=${ifDefined(size)}
-      href=${ifDefined(href || undefined)}
-      ?isExpressive="${isExpressive}"
-      @click=${onClick}>
-    </cds-btn-skeleton>
+    <cds-btn
```
ariellalgilmore, line 222: the react skeleton looks a bit different: https://react.carbondesignsystem.com/?path=/story/components-button--skeleton

packages/carbon-web-components/src/components/button/button-story.ts
```diff
@@ -9,177 +9,320 @@
...
+    </cds-btn>
+  `;
 };
 
 export default {
   title: 'Components/Button',
   parameters: {
     ...storyDocs.parameters,
+    knobs: {
+      [`${prefix}-btn`]: () => ({
```
ariellalgilmore, line 308: some of the knobs are missing from the controls in https://react.carbondesignsystem.com/?path=/story/components-button. if possible can we match not used knobs that are the same in react too?

packages/carbon-web-components/src/components/tile/tile-story.ts
```diff
@@ -184,6 +281,105 @@ export const expandable = (args) => {
...
+  const { colorScheme, expanded, disableChange, onBeforeChange, onChange } =
+    args?.[`${prefix}-expandable-tile`] ?? {};
+  const handleBeforeChanged = (event: CustomEvent) => {
+    onBeforeChange(event);
+    if (disableChange) {
+      event.preventDefault();
+    }
+  };
+  return html`
+    <cds-expandable-tile
```
ariellalgilmore, line 294: weird interaction happening when i click right on the caret button
Mar-20-2023 13-45-57

packages/carbon-web-components/src/components/button/defs.ts
```diff
@@ -1,16 +1,16 @@
...
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
  */
 
 /**
- * Button kinds.
+ * Button type.
  */
-export enum BUTTON_KIND {
+export enum BUTTON_TYPE {
```
annawen1, line 13: Looks like ButtonGroup uses the `BUTTON_KIND` enum, should we update in buttongroup?

Screen Shot 2023-03-20 at 3 35 37 PM

IgnacioBecerra: Whoops, I think I misread some docs somewhere and thought the attribute's name changed to `type`, but it actually remains the same lol. Will change back, thanks for making me notice :)

packages/carbon-web-components/src/components/button/button-story.ts
IgnacioBecerra: I added a couple more knobs that I had missed before, but can you clarify what you mean with the not used knobs?

packages/carbon-web-components/src/components/tag/tag-story.ts
```diff
@@ -13,33 +13,69 @@ import { boolean, select } from '@storybook/addon-knobs';
...
   return html`
     <cds-tag
+      ?filter="${filter}"
+      ?open="${open}"
       size="${ifDefined(size)}"
       type="${ifDefined(type)}"
       title="${ifDefined(title)}"
       ?disabled="${disabled}">
       This is a tag
+      <svg
```
ariellalgilmore, line 61: renderIcon looks good! but it seems like with the react code the filter icon should only render the close icon: https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Tag/Tag.js#L83. Which looks like what you have in the code, so maybe just a storybook update?

packages/carbon-web-components/src/components/button/button-story.ts
ariellalgilmore: hey yeah sorry meant to come back and add! i think now just missing `isExpressive` and maybe `isSelected` for ghost variant

packages/carbon-web-components/src/components/tag/tag-story.ts
IgnacioBecerra: Ohh yeah this was a leftover when i was testing the custom icon haha, thanks for noticing!

packages/carbon-web-components/src/components/button/button-story.ts
IgnacioBecerra: Weird, seems like `isSelected` does nothing? 
https://react.carbondesignsystem.com/?path=/story/components-button--ghost&args=isSelected:true;

ariellalgilmore: https://github.com/carbon-design-system/carbon/blob/9cc6a01fe11636dfb14093a874eb1c0c2b08ca83/packages/react/src/components/Button/Button.js#L97

i think not always the controls they have setup work as expected.... maybe we don't need it as a knob, but definitely need it as a prop. Could we possibly add isExpresive as a knob and then get all of this in!!! lol

ariellalgilmore (approved):
WOOT LGTM!

PR #10243, opened 2023-03-20
chore: sync v11 loading and inline-loading components

emyarod:
### Related Ticket(s)

#10093
#10109

### Description

This PR updates the loading and inline-loading components to match their v11 React counterparts

### Changelog

**New**

- playground stories with assistive text knobs

**Changed**

- `display` style rule for non-overlay loading components


kennylam (commented):
@emyarod Everything looks good except the `small` loader. It's not spinning on a centered axis. I checked the dimensions and everything looks good (container, svg, and svg elements all the same width/height) so I'm not sure what the cause is. If you want to test, setting the loader `height: 19px` fixes it. This is happening in both loaders, `small` version only.

Appears to happen in Windows only.

emyarod:
so far I've only been able to replicate on Windows/Firefox but not Windows/Chrome, need to look into it more

kennylam (approved):
LGTM!

PR #10252, opened 2023-03-22
fix(notice-choice): change js files to ts for notice choice

annawen1:
### Related Ticket(s)

N/A

### Description

rollup config only picks up on the `ts` files when bundling, changing the supplementary files for notice-choice from `js` to `ts` 

### Changelog

**Changed**

- fix typo in codesandbox import
- change `js` files to `ts`


emyarod (approved):
thanks for the fix!

PR #10257, opened 2023-03-23
chore: sync v11 ordered and unordered list components

emyarod:
### Related Ticket(s)

#10104

### Description

This PR updates the ordered-list and unordered-list components to match the Carbon v11 variants. Some spacing may be off due to the font mismatch


kennylam (approved):
Looks good thanks @emyarod!

RichKummer:
Hey @emyarod, noticed that the WC `footer` replaced all of the options with link styling. Not sure if it's unrelated, but wanted to flag!

Screen Shot 2023-03-27 at 12 52 10 PM

emyarod:
good catch @RichKummer, I'll take a look at it!

RichKummer:
@emyarod – checked the styling for both list types and they're in line with react. 

This is more of a storybook issue, but I did notice that in the playground version for both stories, there was a boolean control for **nested** that wasn't present in the knobs here. This might be something to be tackled later, but wanted to call it out

Screen Shot 2023-03-27 at 5 04 38 PM

kennylam:
@RichKummer Lots of the changes done to `@carbon/web-components` will affect/break `@carbon/ibmdotcom-web-components`. These will get fixed in Dotcom V2. No harm in calling them out now but it's not a requirement for the cwc changes. We also are still using knobs, so disregard the controls for now (unless there's actual functionality missing or breaking).

PR #10258, opened 2023-03-24
chore(ci): fixes cwc-v2 failing unit test specs

kennylam:
### Description

This fixes various unit tests in `@carbon/web-components` which were breaking and preventing CI checks from running to completion. This does not imply that all tests are passing - they are not. The failing tests can be updated as their corresponding components are synced with their React counterparts. This fix simply allows the checks to finish.

### Changelog

**New**

- {{new thing}}

**Changed**

- update various unit tests
- update snapshots

**Removed**

- {{removed thing}}

### DIFF
--- a/packages/carbon-web-components/src/components/icon/icon-story.ts
+++ b/packages/carbon-web-components/src/components/icon/icon-story.ts
@@ -49,18 +49,22 @@ withAriaLabel.storyName = 'With aria-label';
 export const withTitle = () => html`
   ${Add16({
     'aria-describedby': 'id-title-1',
+    // @ts-ignore
     children: svg`<title id="id-title-1">add</title>`,
   })}
   ${Add20({
     'aria-describedby': 'id-title-2',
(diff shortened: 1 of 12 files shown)


emyarod (approved):
should we use the `prefix` variable in the tests too?

kennylam:
> should we use the `prefix` variable in the tests too?

That's weird, the last snapshot test I did, it actually did use the new prefix and generated the new spec files with `cds-`. Not sure what happened but yes, let me get that in.

kennylam:
@emyarod Actually after going a little overboard changing every instance of `BX` > `CDS`, I think we should continue doing it the way we are at a component-by-component basis. I just did a global replace and a lot of stuff breaks - styles, tests, etc. This PR is just meant to get the ci-checks running, so I think I'm going to hold off on changing prefixes everywhere.

emyarod:
got it, makes sense to hold off for now!

PR #10266, opened 2023-03-27
chore(gh): adds workflow to auto add issues/PRs to project

kennylam:
### Description

This adds a new GitHub workflow to automatically add new issues and pull requests to the Carbon for IBM.com project. Currently it will add anything newly created labeled. This workflow is just a starting point. In the future, new workflow filters can be created to add to other GitHub projects.

See https://github.com/actions/add-to-project for more use cases.

### Changelog

**New**

- workflow to automatically add new issues to default project

**Changed**

- {{changed thing}}

**Removed**

- {{removed thing}}

### DIFF
--- a/.github/workflows/add-to-project.yml
+++ b/.github/workflows/add-to-project.yml
@@ -1,0 +1,22 @@
+name: Add to Project
+on:
+  issues:
+    types: [opened, labeled]
+  workflow_call:
+    secrets:
+      ADD_TO_PROJECT:
+        required: true
+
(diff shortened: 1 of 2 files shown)


kennylam:
> Look good! Just not totally sure if want to include on types `labeled` since if a new issue was opened then it will already be added to the project?

I had that thought too, but I figured this way, existing issues in the repo that are _not_ in the project will also get added automatically if the labels are adjusted.

Issue #10250, opened 2023-03-22
[Image with Caption]: Button background color doesn't work well with transparent images


jwitkowski79:
**JIRA Ticket:** https://jsw.ibm.com/browse/ADCMS-3105

### Description

The interfaces team is recommending using SVG file types for line art and diagrams. For diagrams, we're recommending not to use a background color in the SVG export unless the design warrants it. When we use the SVG in conjunction with the lightbox media viewer, the SVG gets wrapped in the carbon button element with a background color assigned. This background color can interfere with the accessibility of the diagrams and other line art. 

We're hoping that there's a way to adjust the prop that allows users/to adjust the image with caption background color. That way we'd have more control over the background color when using SVGs or transparent images

Screenshot attached and here's a [test page in AEM](https://prod-author.roks.cms.cis.ibm.net/content/adobe-cms/language-masters/en/prod-sandbox/-test--product-overview-jw.html?wcmmode=disabled)

![87816A98-4EB6-4BE2-B471-750281F5AEA9](https://user-images.githubusercontent.com/2040791/226903483-79849a0a-df0b-4e72-bc6d-3a501368ccc9.png)

### Component(s) impacted

Button element - but only when it's used in conjunction with the lightbox media viewer

### Browser

Chrome

### Carbon for IBM.com version

v1.29.0

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

Interfaces team - Adobe Experience Manager

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://prod-author.roks.cms.cis.ibm.net/content/adobe-cms/language-masters/en/prod-sandbox/-test--product-overview-jw.html?wcmmode=disabled

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #10279, opened 2023-03-28
Fix image lightbox button background

m4olivei:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/10250
https://jsw.ibm.com/browse/ADCMS-3105

### Description

Sets the background style to transparent for the `<button>` elment used in the shadow DOM of the `<dds-image>` and `<dds-image-with-caption>` components, such that when used with a transparent image (SVG, transparent gif/png), the browser's default button grey does not show through.

Note that the original issue was reported for `<dds-image-with-caption>`, but the issue equally effects `<dds-image>` component as well. Note also that the `<dds-image-with-caption>` button was recently deprecated, it's features merged with `<dds-image>`. Use of `<dds-image-with-caption>` should be replaced by `<dds-image>` (see https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8839). The deprecation is documented in the Storybook README for the Image component.

### Testing Instructions

- [ ] Open Storybook to the Image > Default component. 
- [ ] For the "Default image" property, choose "SVG". A transparent SVG is used as the image for the component. 
- [ ] Note the browser default button color does not show through.

### Changelog

**Changed**

- Added `background-color: transparent` to the button styles for the `<dds-image>` component when in lightbox mode.


packages/web-components/src/components/image/__stories__/image.stories.ts
```diff
@@ -24,6 +24,7 @@ import textNullable from '../../../../.storybook/knob-text-nullable';
...
...1pbHk9IklCTSBQbGV4IFNhbnMiIGZvbnQtc2l6ZT0iMTIiIGxldHRlci1zcGFjaW5nPSIwZW0iPjx0c3BhbiB4PSIxMDIwIiB5PSIyNjAuNSI+RE5TPC90c3Bhbj48L3RleHQ+CjxwYXRoIGQ9Ik0xMTQzLjUgNDM2LjVIMTAwOS41VjQ3Ny41SDExNDMuNVY0MzYuNVoiIHN0cm9rZT0iI0E4QThBOCIvPgo8dGV4dCBmaWxsPSIjMTYxNjE2IiB4bWw6c3BhY2U9InByZXNlcnZlIiBzdHlsZT0id2hpdGUtc3BhY2U6IHByZSIgZm9udC1mYW1pbHk9IklCTSBQbGV4IFNhbnMiIGZvbnQtc2l6ZT0iMTIiIGxldHRlci1zcGFjaW5nPSIwZW0iPjx0c3BhbiB4PSIxMDIwIiB5PSI0NjEuNSI+TWFuYWdlbWVudDwvdHNwYW4+PC90ZXh0Pgo8cGF0aCBkPSJNMTY2LjUgNTYwLjVIMTUxLjVWNTc1LjVIMTY2LjVWNTYwLjVaIiBmaWxsPSIjRURGNUZGIiBzdHJva2U9IiMwRTYyRkQiLz4KPHBhdGggZD0iTTAuNSA1NDQuNUgxMTQ0LjUiIHN0cm9rZT0iI0M2QzZDNiIgc3Ryb2tlLWxpbmVjYXA9InNxdWFyZSIvPgo8cGF0aCBkPSJNMTM2LjUgNTI4LjVWMC41IiBzdHJva2U9IiNDNkM2QzYiIHN0cm9rZS1saW5lY2FwPSJzcXVhcmUiLz4KPHRleHQgZmlsbD0iIzE2MTYxNiIgeG1sOnNwYWNlPSJwcmVzZXJ2ZSIgc3R5bGU9IndoaXRlLXNwYWNlOiBwcmUiIGZvbnQtZmFtaWx5PSJJQk0gUGxleCBTYW5zIiBmb250LXNpemU9IjEyIiBsZXR0ZXItc3BhY2luZz0iMGVtIj48dHNwYW4geD0iMTc1IiB5PSI1NzMuNSI+SW5jbHVkZWQgaW4gdGhpcyBhcmNoaXRlY3R1cmU8L3RzcGFuPjwvdGV4dD4KPC9nPgo8ZGVmcz4KPGNsaXBQYXRoIGlkPSJjbGlwMF8yMDM2XzIxNDE5NCI+CjxyZWN0IHdpZHRoPSIxMTQ1IiBoZWlnaHQ9IjU3NiIgZmlsbD0id2hpdGUiLz4KPC9jbGlwUGF0aD4KPC9kZWZzPgo8L3N2Zz4K',
```
m4olivei, line 27: Open to suggestions of a better spot for this SVG. I tried an import, but there isn't a loader setup to handle SVG, so I thought this was an easy way to get an SVG here to test with.

oliviaflory (commented):
@m4olivei The change looks good from a design perspective

If we are going to keep the SVG as an example demo I'd like to change the knob label to SVG (transparent background) or something like that so it's a bit more clear.

I'll also just ping the over devs to get any ideas for a better placement (if we're keeping it)

kennylam:
@m4olivei There actually is a custom [loader](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/7f49e0c909d827a611a9e3861d67841eeb643ee4/packages/web-components/.storybook/main.js#L170-L176) being used, but it's never played nice with assets loaded in Storybook (gotta fix that). In the meantime, I think the best way would be to take the data and put it into its own file, e.g. `image/__stories__/chartSVG.js`?

```js
const foo = 'monster-encoded-chunk-of-text';

export default foo;
```
Then you can just import as usual and use that in `images` object. There are a few stories with external data files for the sake of keeping the stories less cluttered. What do you think?

kennylam (approved):
Thanks @m4olivei!

oliviaflory (approved):
Thanks @m4olivei !

Issue #10118, opened 2023-02-23
[slider]: sync with @carbon/react version


kennylam:

emyarod:
input field affected by box-sizing issue

![image.png](https://images.zenhubusercontent.com/5be364c17610415c19d14847/5e48a285-a346-4f02-b93a-f2a0d43090e1)

PR #10297, opened 2023-04-04
feat(slider): sync slider with @carbon/react

annawen1:
### Related Ticket(s)
[slider]: sync with @carbon/react version #10118

### Description

@carbon/react syncing with `slider` component. 

@carbon/react version: https://react.carbondesignsystem.com/?path=/story/components-slider--playground


RichKummer:
This is looking good from a design perspective! I just found two differences from react:

1. When in `readonly`, the background of the text input should change to transparent (just as the input would for `readonly`)

**React**
Screen Shot 2023-04-04 at 4 24 26 PM

**Current build**
Screen Shot 2023-04-04 at 4 24 17 PM

2. I tried updating the knobs for `min-label` and `max-label`, and it looks like the current build does not add those to the slider like they do with react.

kennylam (approved):
LGTM!

RichKummer (approved):
Thanks for fixing @annawen1 ! LGTM 💪

Issue #10262, opened 2023-03-24
Leadspaces: At the lg breakpoint, the title box is too long


jwitkowski79:
JIRA ticket: https://jsw.ibm.com/browse/ADCMS-3154
 
------ 
### Description

For all leadspaces, at the lg breakpoint, the container of the title is too long. It's currently spanning 10 columns where in the design specs, it shows (8 columns - 32px). Confirmed with @oliviaflory that this is a bug on the component level. Just a note that this poses as an accessibility problem as we're noticing text overlapping background imagery when the imagery is in the "safe area"
![leadspace-bug-lg](https://user-images.githubusercontent.com/2040791/227624263-ace41586-096f-4aad-a5bf-9aa5dc80b85b.png)

**Expected behavor**
At the lg breakpoint, the title container should only extend 8 columns as noted in the design specs
![leadspace-designspecs](https://user-images.githubusercontent.com/2040791/227624814-d3e86fc3-a538-40bf-bed4-3b4041ea4bdb.png)

### Component(s) impacted

All leadspaces at the lg breakpoint

### Browser

Chrome

### Carbon for IBM.com version

1.29.0

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

IBM.com

### Package

@carbon/web-components

### CodeSandbox example

https://www.ibm.com/standards/carbon/web-components/iframe.html?knob-CTA%20type%20(cta-type)=local&knob-Card%20type:=Card%20static&knob-Cards%20per%20row:=dds-ce-demo-devenv--cards-in-row-3&knob-Copy%20text%20(copy)=Lorem%20ipsum%20dolor%20sit%20ametttt&knob-Grid%20mode:=collapsed&knob-Heading%20(heading):=Aliquam%20condimentum%20interdum&knob-Href%20(href):=https://example.com&knob-Number%20of%20cards:=5&knob-Offset:=0&id=components-lead-space--medium-with-image&viewMode=story

### Steps to reproduce the issue (if applicable)

[...]

Issue #10261, opened 2023-03-24
Leadspace with image at md breakpoint - image not stacking below text and CTAs


jwitkowski79:
JIRA ticket: https://jsw.ibm.com/browse/ADCMS-3154
 
------ 
### Description

In all leadspaces with images, at the md breakpoint, the image stays right aligned to the text and does not follow design specs where it stacks below the title, description and CTAs. Checked with @oliviaflory and this is a bug in the web component.
![leadspace-bug](https://user-images.githubusercontent.com/2040791/227621567-8e46014f-338e-498c-9ab4-002015168948.png)

Expected behavior - at the md breakpoint, the image should stand below the text as seen in the design specs
![leadspace-designspec](https://user-images.githubusercontent.com/2040791/227621591-6a628884-ab78-434d-b3a3-82353ed2663a.png)

### Component(s) impacted

All leadspace variations with image 

### Browser

Chrome

### Carbon for IBM.com version

1.20.0

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

IBM.com 

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://www.ibm.com/standards/carbon/web-components/?path=/story/components-lead-space--tall-with-image&knob-CTA%20type%20(cta-type)=local&knob-Card%20type:=Card%20static&knob-Cards%20per%20row:=dds-ce-demo-devenv--cards-in-row-3&knob-Copy%20text%20(copy)=Lorem%20ipsum%20dolor%20sit%20ametttt&knob-Grid%20mode:=collapsed&knob-Heading%20(heading):=Aliquam%20condimentum%20interdum&knob-Href%20(href):=https://example.com&knob-Number%20of%20cards:=5&knob-Offset:=0

### Steps to reproduce the issue (if applicable)

[...]

PR #6154, opened 2021-05-21
fix(leadspace): add super variation and tag group

ariellalgilmore:
### Related Ticket(s)

#5960 #6153 #6160

### Description

Adds tag group to leadspace , fixes height issues and heading text for each variation. 
Screen Shot 2021-05-24 at 8 34 56 AM

### Changelog

**New**

- slot for navigations elements above heading
- handleSlotChange if tag group or breadcrumbs in slot
- add super variation 

**Changed**

- style updates

PR #10303, opened 2023-04-05
fix(leadspace): fix heading width for the leadspace

m4olivei:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/10262
https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/10261

https://jsw.ibm.com/browse/ADCMS-3154

### Description

Fixes a couple of Leadspace issues. First issue it to do with the heading width, it's off according to the design specs. This brings it in line with the spec for both the medium and large breakpoints. Second issue is when the background image shifts from showing on its own line to inline with the heading and description. It now only shifts inline at the large breakpoint.

### Changelog

**Changed**

- Leadspace background now only shifts to inline with headeing and description at the large breakpoint
- Spacing adjustments for the Leadspace title width at medium and large breakpoints per [design specs](https://ibm.ent.box.com/folder/94761007991?s=d9jey3rlfxqmxu059f0x9dtcx7f5ehuh).


packages/styles/scss/components/leadspace/_leadspace.scss
```diff
@@ -188,15 +188,7 @@ $btn-min-width: 26;
 
       ::slotted(#{$dds-prefix}-leadspace-heading),
       .#{$prefix}--leadspace__title {
-        @include content-width;
-
-        @include carbon--breakpoint(md) {
-          @include carbon--make-col(4, 6);
```
m4olivei, line 194: I found these calls to `carbon--make-col` were ineffective on account of the parent (`.bx--leadspace__row`) being changed to a `flex-direction: column` on line 157. `carbon--mak-col` depends on `flex-basis` to set the width and since it gets changed to `flex-direction: column` it doesn't have the intended effect.

proeung: @m4olivei Thanks for pointing this out! I wonder why it was adjusted to `column`. @ariellalgilmore Maybe you have some insight into this?

Ideally, we should stick to using the `carbon--mak-col` mixin which uses `flex-basis` to stay close to the `Carbon @2x grid` width as best as possible.

ariellalgilmore: The background behind the `flex-direction: column` change is here: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/7004

RichKummer (commented):
Hey @m4olivei, thanks for this! Took a look and on `@md` and the image is swapping to below which is great. The width of the heading, copy, and available width for the buttons should be six columns - `32px`. In the current build, it looks like the width is set close to four columns only.

**@md**
Screen Shot 2023-04-07 at 2 42 41 PM

For `@lg`, it looks like the heading width is slightly off. Should be eight columns -32px.

**@lg**
Screen Shot 2023-04-07 at 2 54 04 PM

packages/styles/scss/components/leadspace/_leadspace.scss
m4olivei: Thanks @ariellalgilmore . It feels like we could refactor to remove the `flex-direction: column` there and then get back to using `carbon--make-col`. The original change to make it `flex-direction: column` appears to have come in this PR: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/6154

I think if we want to do that, it would take some thought and more testing to do it without regressions. I'd suggest it's out of scope for this and could get handled in a follow up if there was interest.

m4olivei:
Thanks @RichKummer . I've adjusted the md breakpoint to be 6 columns. The lg breakpoint does get an 8 column width:

image

packages/styles/scss/components/leadspace/_leadspace.scss
ariellalgilmore: thanks! maybe we can add `TODO` statements on those lines, so we can come back to this.

RichKummer (approved):
Thanks for the fix @m4olivei !

kennylam (approved):

PR #10337, opened 2023-04-13
feat(search): sync with carbon react

IgnacioBecerra:
### Related Ticket(s)
[#10097](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/10097)

### Description
This PR synchronizes our Search component with Carbon React's. 

### Changelog

**New**

- search can now be expandable
- search now accepts custom icon to replace magnifier


packages/web-components/src/components/search/search.ts
```diff
@@ -8,7 +8,7 @@
  */
 
 import { customElement } from 'lit/decorators.js';
-import BXSearch from '../../internal/vendor/@carbon/web-components/components/search/search.js';
+import CDSSearch from '../../internal/vendor/@carbon/web-components/components/search/search.js';
```
ariellalgilmore, line 11: i think we might want to keep it as `BXSearch` for now until we update `web-components` to v2 since it's still pulling from `@carbon/web-components: 1.xx`

oliviaflory:
@IgnacioBecerra looking good! 

A few things I noticed:

### 1. Disabled state
- The search icon isn't using the disabled color token
- The close button is still interactive, but it should also be disabled according to the React examples 

Screenshot 2023-04-17 at 12 05 20 PM

### 2. Expandable stories
The search field text is overlapping with the search icon in the Expandable and Expandable with layer stories

Screenshot 2023-04-17 at 12 02 41 PM

### 3. Storybook
- What is the `Name (name)` knob? I didn't see it in React and couldn't tell what it was controlling in WC

IgnacioBecerra:
@oliviaflory Thanks for the review, addressed the comments! 
As for the `name` knob, its property sets the `name` of the actual input element to be used within a form element. You can see this when logging in to a website and your browser tries to autofill the email input with your actual email, because its `name` is `e-mail`, same with other information such as `password` or `address`. 

Since the React storybook doesn't have it, I will remove it from our CWC Storybook as well.

packages/web-components/src/components/search/search.ts
IgnacioBecerra: Completely missed this while I was working on the changes yesterday, but changed it back, good call!

PR #10338, opened 2023-04-13
feat(notifications): sync with @carbon/react v11

annawen1:
### Related Ticket(s)

[notifications]: sync with @carbon/react version #10120

### Description

Sync with @carbon/react v11 - this includes new `actionable` variant 

@carbon/react version here: https://react.carbondesignsystem.com/?path=/story/components-notifications-actionable--playground


oliviaflory:
Looks good @annawen1 !


### Icons
This looks like it might be a bug in React as well, but I believe the warning icons need a black fill in order to pass color contrast. I see it working as expected in the `toast / warning / low contrast` but not some of the other variants (inline and actionable) or warning alts

Screenshot 2023-04-14 at 4 19 52 PM
Screenshot 2023-04-14 at 4 19 58 PM

### Bug in React
WC `inline` variant does have the focus around the close button when you tab in, but React doesn't!

Screenshot 2023-04-14 at 4 21 29 PM


### Storybook (small things that don't fully matter)

1. React uses alphabetical order for all their controls and the order of "kind". Do you think we should too?

Screenshot 2023-04-14 at 4 23 36 PM
Screenshot 2023-04-14 at 4 24 07 PM

2. Playground stories
For some reason the Playground stories have a period after "Subtitle goes here" and also add "Time stamp" in the toast notification, but those both don't appear in the Default stories (which match React)
Screenshot 2023-04-14 at 4 26 45 PM

annawen1:
> Looks good @annawen1 !
> 
> ### Icons
> This looks like it might be a bug in React as well, but I believe the warning icons need a black fill in order to pass color contrast. I see it working as expected in the `toast / warning / low contrast` but not some of the other variants (inline and actionable) or warning alts
> 
> Screenshot 2023-04-14 at 4 19 52 PM Screenshot 2023-04-14 at 4 19 58 PM
> 
> ### Bug in React
> WC `inline` variant does have the focus around the close button when you tab in, but React doesn't!
> 
> Screenshot 2023-04-14 at 4 21 29 PM
> 
> ### Storybook (small things that don't fully matter)
> 1. React uses alphabetical order for all their controls and the order of "kind". Do you think we should too?
> 
> Screenshot 2023-04-14 at 4 23 36 PM Screenshot 2023-04-14 at 4 24 07 PM
> 
> 2. Playground stories
>    For some reason the Playground stories have a period after "Subtitle goes here" and also add "Time stamp" in the toast notification, but those both don't appear in the Default stories (which match React)
> 
> Screenshot 2023-04-14 at 4 26 45 PM

@oliviaflory  I opened an a11y bug with the Carbon core team for the inline notification here: https://github.com/carbon-design-system/carbon/issues/13616 with both of the mentioned issues (close button not focusing & the warning icon fill) 

I added fixes for the warning icon fill - we can remove them once the fix is updated in the Carbon core styles and made the storybook adjustments

oliviaflory (approved):
LGTM, thank you @annawen1

PR #10350, opened 2023-04-17
feat: sync v11 structured list component

emyarod:
### Related Ticket(s)

#10125

### Description

This PR syncs the structured list component with its v11 react counterpart

### Changelog

**New**

- `condensed` structured list variant
- `flush` structured list variant


packages/carbon-web-components/src/components/structured-list/structured-list.ts
```diff
@@ -30,6 +30,18 @@ class BXStructuredList extends FocusMixin(LitElement) {
   @property({ attribute: 'selection-name' })
   selectionName = '';
 
+  /**
+   * Specify if structured list is condensed, default is false
+   */
+  @property({ type: Boolean, reflect: true, attribute: 'condensed' })
```
kennylam, line 36:
```suggestion
  @property({ type: Boolean, reflect: true })
```
I think `attribute` may be redundant here?

packages/carbon-web-components/src/components/structured-list/structured-list.ts
```diff
@@ -30,6 +30,18 @@ class BXStructuredList extends FocusMixin(LitElement) {
...
+  condensed = false;
+
+  /**
+   * Specify if structured list is flush, default is false
+   */
+  @property({ type: Boolean, reflect: true, attribute: 'flush' })
```
kennylam, line 42:
```suggestion
  @property({ type: Boolean, reflect: true })
```
I think `attribute` may be redundant here?

packages/carbon-web-components/src/components/structured-list/structured-list.scss
```diff
@@ -31,6 +31,10 @@ $css--plex: true !default;
   ::slotted(#{$prefix}-structured-list-header-cell-skeleton) {
     @extend .#{$prefix}--structured-list-th;
   }
+
+  ::slotted(#{$prefix}-structured-list-header-cell) {
+    padding-left: 1rem;
```
kennylam, line 36: Let's use spacing tokens for all instances of these units?

kennylam (commented):
@emyarod Nice! Just a few suggestions.

packages/carbon-web-components/src/components/structured-list/structured-list.scss
emyarod: updated and fixed some new spacing issues I found

oliviaflory: @emyarod you may have just fixed some of the spacing things I caught and just posed in my review 😄

emyarod:
@oliviaflory the spacing issues I'm still figuring out how to get around but the remaining issues should be resolved!

packages/carbon-web-components/src/components/structured-list/structured-list.ts
IgnacioBecerra: @emyarod I think you forgot to remove the redundant attributes

packages/carbon-web-components/src/components/structured-list/structured-list.ts
```diff
@@ -52,11 +64,32 @@ class BXStructuredList extends FocusMixin(LitElement) {
...
+  updated(changedProperties) {
+    const attributes = ['condensed', 'flush'];
+    attributes.forEach((attr) => {
+      if (changedProperties.has(attr)) {
+        // Propagate watched attribute to descendants until `:host-context()` gets supported in all major browsers
+        forEach(
+          this.querySelectorAll(
+            (this.constructor as typeof BXStructuredList).selectorRowsWithHeader
+          ),
+          (elem) => {
```
IgnacioBecerra, line 76: Wondering if it might be better to do this in `firstUpdated` instead, or save all the `rowsWithHeader` components to variable/property so we don't have to re-query them every time the component is updated.

packages/carbon-web-components/src/components/structured-list/structured-list.scss
```diff
@@ -47,32 +52,84 @@ $css--plex: true !default;
...
+
+  &:nth-of-type(3n + 3) {
+    width: 15%;
+  }
+
+  span {
+    @include skeleton;
+
+    width: 75%;
+    height: 1rem;
```
IgnacioBecerra, line 104:
```suggestion
    height: $spacing-05;
```
I know this isn't technically spacing, but maybe this should replaced too?

IgnacioBecerra (commented):
Looks good, just two things!

packages/carbon-web-components/src/components/structured-list/structured-list.ts
emyarod: looks like I renamed them instead of removing them 😅

packages/carbon-web-components/src/components/structured-list/structured-list.ts
IgnacioBecerra: Hm, I can see how that would affect it. We can look into that closer in some future release since it's not too performance heavy.

IgnacioBecerra (approved):
LGTM!

oliviaflory:
@emyarod this looks good!

I caught a few things in the selectable story:
### 1. The header column is still taller than it should be in the selectable variant, also in the playground when you activate the `selectable` knob.
May-05-2023 10-53-33


### 2. The hover and selected colors look a little off. Have we updated to v11 tokens or are we still using v10?

| State | element | Token v10 | Token v11 |
| ------ | ------ | ------ | ------ |
| Hover | Text | $text-01    | $text-primary |
| Hover | Row | $hover-row (maybe $hover-ui) | $layer-hover |
| Selected | Text |  $text-01    | $text-primary |
| selected | Row | $selected-ui | $layer-selected |

emyarod:
@oliviaflory I think the color tokens should be correct now! but the spacing issue seems to have reappeared in a different form, need to take a closer look again...

oliviaflory (approved):
LGTM!

PR #10390, opened 2023-04-24
chore(checks): pin node version to 18.15

kennylam:
### Description

A current [bug](https://github.com/nodejs/node/issues/47563) in the latest version of Node is causing CI checks to fail. The bug has been fixed but not yet back-ported, so in the meantime will test pinning the version to `18.15`.

### Changelog

**New**

- {{new thing}}

**Changed**

- pin `ci-check` workflow Node version to `18.15`

**Removed**

- {{removed thing}}


IgnacioBecerra (approved):
LGTM!!!

PR #10400, opened 2023-04-26
chore(mixin): import `carbon-components` mixin

kennylam:
### Description

This imports the `on` helper function from `carbon-components` into the package, as `carbon-components` is no longer a dependency.

### Changelog

**New**

- `on` mixin added

**Changed**

- change `on` references from `carbon-components` to local

**Removed**

- {{removed thing}}


ariellalgilmore (approved):
LGTM!

PR #10402, opened 2023-04-27
fix: add missed import extensions

abdonrd:
### Description

Add missed import extensions.

### Changelog

**Changed**

- Add missed import extensions

### DIFF
--- a/packages/carbon-web-components/examples/codesandbox/data-table/src/demo-data-table.js
+++ b/packages/carbon-web-components/examples/codesandbox/data-table/src/demo-data-table.js
@@ -8,7 +8,7 @@
  */
 
 import debounce from 'lodash-es/debounce';
-import { repeat } from 'lit-html/directives/repeat';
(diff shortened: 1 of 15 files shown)


packages/carbon-web-components/examples/codesandbox/data-table/src/demo-data-table.js
```diff
@@ -8,7 +8,7 @@
  */
 
 import debounce from 'lodash-es/debounce';
-import { repeat } from 'lit-html/directives/repeat';
+import { repeat } from 'lit-html/directives/repeat.js';
```
kennylam, line 11:
```suggestion
import { repeat } from 'lit/directives/repeat.js';
```

abdonrd: Updated with more references!

kennylam (approved):
Thanks @abdonrd!

PR #10407, opened 2023-04-27
chore(storybook): add plex cdn stylesheet

kennylam:
### Description

This adds Plex via CDN, which seems to resolve certain font issues. Also added `fontBase` and `fontCode` to the theme config, which doesn't seem to do anything. Going to leave it for now because it could be something in our config that we need to fix before they are applied.

### Changelog

**New**

- {{new thing}}

**Changed**

- {{changed thing}}

**Removed**

- {{removed thing}}


ariellalgilmore (approved):
LGTM!

emyarod (approved):
looks good to me!

PR #10412, opened 2023-05-01
feat(multi-select): @carbon/react v11 sync

annawen1:
### Related Ticket(s)

[multi-select]: sync with @carbon/react version #10094

### Description

### Changelog

**New**

- {{new thing}}

**Changed**

- {{changed thing}}

**Removed**

- {{removed thing}}


oliviaflory:
@annawen1 this is looking good!

A few things I noticed, all screen shots are WC first, and React second:

### Default: When multi select is opened – focus placement
In React the focus placement is wrapped around the entire menu dropdown, but in WC it remains on the Multiselect input.
- Note the focus is a single pixel in React (not 100% sure why that is)
- Also note this behavior is **not** in the Filterable multi select (seems strange they would behave differently??)
Screenshot 2023-05-09 at 12 11 28 PM
Screenshot 2023-05-09 at 12 11 36 PM

### Read only
Styles are a little off in Read only
- They don't have the tokens listed out on the website for me to grab for you.. if you have trouble finding them LMK!
Screenshot 2023-05-09 at 12 22 54 PM

packages/carbon-web-components/src/components/multi-select/defs.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  /**
+   * selected item jumps to top
+   */
+  TOP = 'top',
+
+  /**
+   * selected item jump to top after reopen dropdown
+   */
+  TOP_AFTER_REOPEN = 'top-after-open',
```
kennylam, line 27:
```suggestion
  TOP_AFTER_REOPEN = 'top-after-reopen',
```

Should be this, right?

packages/carbon-web-components/src/components/multi-select/multi-select.scss
```diff
@@ -90,18 +116,18 @@ $css--plex: true !default;
...
   }
 }
 
-:host(#{$prefix}-multi-select-item[size='xl']) {
-  height: rem(48px);
-
+:host(#{$prefix}-multi-select-item[size='lg']) {
+  height: fit-content;
   .#{$prefix}--list-box__menu-item__option {
+    height: rem(48px);
```
kennylam, line 130:
```suggestion
    height: $spacing-09;
```

packages/carbon-web-components/src/components/multi-select/multi-select.scss
```diff
@@ -90,18 +116,18 @@ $css--plex: true !default;
 }
 
 :host(#{$prefix}-multi-select-item[size='sm']) {
-  height: rem(32px);
-
+  height: fit-content;
   .#{$prefix}--list-box__menu-item__option {
+    height: rem(32px);
```
kennylam, line 121:
```suggestion
    height: $spacing-07;
```

packages/carbon-web-components/src/components/multi-select/multi-select.ts
```diff
@@ -328,6 +374,24 @@ class CDSMultiSelect extends CDSDropdown {
...
+   * Specify the text that should be read for screen readers to clear selection.
+   */
+  @property({ attribute: 'clear-selection-text' })
+  clearSelectionText = 'To clear selection, press Delete or Backspace.';
+
+  /**
+   * Specify the locale of the control. Used for the default compareItems used for sorting the list of items in the control.
+   */
+  @property()
+  locale = 'en';
```
kennylam, line 393: Should this be an attribute so it can be user defined?

annawen1: user should be able to define it, I didn't add that "attribute" field since the attribute name matches

RichKummer (approved):
@annawen1 – looks like Olivia's comments were addressed and LGTM!

PR #10420, opened 2023-05-04
feat(pagination): syncing pagination

IgnacioBecerra:
### Related Ticket(s)

### Description

### Changelog

**New**

- {{new thing}}

**Changed**

- {{changed thing}}

**Removed**

- {{removed thing}}

### DIFF
--- a/packages/carbon-web-components/src/components/select/select.scss
+++ b/packages/carbon-web-components/src/components/select/select.scss
@@ -9,6 +9,7 @@ $css--plex: true !default;
 
 @use '@carbon/styles/scss/config' as *;
 @use '@carbon/styles/scss/components/select';
+@use '@carbon/styles/scss/theme' as *;
 
 :host(#{$prefix}-select) {
   @extend .#{$prefix}--select;
@@ -40,6 +41,27 @@ $css--plex: true !default;
 
     .#{$prefix}--select__invalid-icon--warning {
       right: 2rem;
+    }
+  }
+
+  &[pagination] {
+    .#{$prefix}--label {
+      margin: 0;
+    }
+
+    .#{$prefix}--select-input {
+      padding: 0 2.25rem 0 1rem;
+    }
+
+    .#{$prefix}--select__arrow {
+      top: 50%;
+      transform: translate(-0.5rem, -50%);
+    }
+
+    &[left-select] {
+      .#{$prefix}--select-input {
+        border-right: 1px solid $border-subtle;
+      }
     }
   }
 }
--- a/packages/carbon-web-components/src/components/button/button-story.ts
+++ b/packages/carbon-web-components/src/components/button/button-story.ts
@@ -83,6 +83,7 @@ export const Default = (args) => {
 
   return html`
     <cds-button
+      class-name="thisidasd sias sdsd dsa as ads "
       kind="${ifDefined(kind)}"
       type="${ifDefined(type)}"
       ?disabled="${disabled}"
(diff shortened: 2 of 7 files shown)


oliviaflory:
@IgnacioBecerra overall this looks really good! 

I caught a few style things, in all the screen shots below I posted our CWC first and React example second.


### onClick & focus state of select
The select is changing background color to white in CWC on focus and onClick, but Carbon React doesn't seem to
Screenshot 2023-05-09 at 11 33 11 AM
Screenshot 2023-05-09 at 11 33 21 AM

### Focus state of previous/next buttons
The white and gray 10 themes look like they have a white inner focus that isn't present in React
Screenshot 2023-05-09 at 11 40 34 AM
Screenshot 2023-05-09 at 11 40 41 AM

### Tool tip
- Our tool tip looks like it has some extra bottom padding or maybe just an extra return?

Screenshot 2023-05-09 at 11 33 38 AM
Screenshot 2023-05-09 at 11 33 43 AM

- It also looks like the React version might have a left alignment on the `next` button to prevent it from running off the viewport if the pagination is extended to the edge
Screenshot 2023-05-09 at 11 47 27 AM
Screenshot 2023-05-09 at 11 46 54 AM


### Gray 90 and Gray 100 theme icon color
Our icons are not using the correct color token in the dark themes, it should be $icon-01 (v10) or $icon-primary (v11)
Screenshot 2023-05-09 at 11 36 08 AM
Screenshot 2023-05-09 at 11 36 13 AM


### Size knob
Small thing, but it might be confusing to call it "button size" in our Storybook knob vs React just says "size"
Screenshot 2023-05-09 at 11 39 45 AM


### Playground story
If I enabled the `last page` knob it disables the `Next page` button even if there are more pages left. Idk what the exact expected behavior is because the React playground story seems broken (all the selects and buttons are disabled???) 

It might be worth reaching out to Taylor or Lauren to find out what's going on with the React storybook and what the expected behavior is for this knob?
Screenshot 2023-05-09 at 11 54 58 AM

IgnacioBecerra:
@oliviaflory Man, I'm always so amazed you and Rich are able to notice these tiny things!! 😆 

All the issues have been addressed, and as for the `is-last-page` knob, it's doing what its meant to. It's just a boolean that disables the next page button as it specifies the current page as the last, we used to have it in the earlier Storybook as well: https://web-components.carbondesignsystem.com/?path=/story/components-pagination--default

RichKummer (approved):
Looks like Olivia's comments have been addressed – thanks @IgnacioBecerra ! 🚀

packages/carbon-web-components/src/components/select/select.scss
```diff
@@ -42,4 +43,29 @@ $css--plex: true !default;
...
+      padding: 0 2.25rem 0 1rem;
+
+      &:focus {
+        background-color: transparent;
+      }
+    }
+
+    .#{$prefix}--select__arrow {
+      top: 50%;
+      transform: translate(-0.5rem, -50%);
```
kennylam, line 62:
```suggestion
      transform: translate(-#{$spacing-03}, -50%);
```

packages/carbon-web-components/src/components/select/select.scss
```diff
@@ -42,4 +43,29 @@ $css--plex: true !default;
       right: 2rem;
```
kennylam, line 43:
```suggestion
      right: $spacing-07;
```

kennylam (commented):
@IgnacioBecerra Looks great! Just a few small things.

IgnacioBecerra:
@kennylam Nice catch, addressed the comments!

kennylam (approved):
LGTM!

PR #10450, opened 2023-05-11
feat(image): add 'default-image' ::part

### LINKED ISSUE #10380: [Image]: Enable aspect-ratio on the <dds-image> so that it aligns with the SVG viewbox
**JIRA ticket:** https://jsw.ibm.com/browse/ADCMS-3234

----------

### The problem

- From **Daniel Ratzenböck** - See slack thread (https://ibm-marketing-comms.slack.com/archives/G01LKRP5RB6/p1682066909914929)
  - On AEM we now provide the option for authors to select a crop for SVGs for different breakpoints. Using the `#svgView`  parameter in the source sets. I am then picking the `viewBox` per breakpoint. https://www.w3.org/TR/SVG2/linking.html#SVGFragmentIdentifiersDefinitions.
  - The problem is that if I don't set width and height on the `img` tag the svg will always render with its original aspect ratio. E.g. the SVG is 16:9, but the viewbox is 1:1 - the svg in the image tag will defer the 1:1 viewbox and scale it to 16:9. To bypass that I am setting the `aspect-ratio` css attribute to 1/1  on the `dds-image` .
  - But this works only if there is no caption ("heading" attribute).

### The solution

- Add an `aspect-ratio` prop/attribute on the `<dds-image>` that allows adopters to find the image container aspect ratio (eg. 1:1, 16:9) that aligns with the picking of the viewBox per breakpoint.

### Application/website

AEM

### Business priority

_No response_

### Examples

- Daniel, can you provide screenshots or a test page of what you’re seeing in AEM? Thanks!

### Code of Conduct

andy-blum:
### Related Ticket(s)

Closes #10380

### Description

Adds `part` property with value of `image` to the rendered `` tag.

### Testing Instructions

In the styles tab of the devtools, add a new style targeting `dds-image::part(image)`, and note that style added here impact the shadow root's `` tag.

Would be good to have @dratzenboeck-ibm provide an SVG file with `#svgView` parameter to test with.

### Changelog

**New**

- dds-image::part(image)


dratzenboeck-ibm:
The parameter can be applied to any svg. E.g. for this image from ibm dam https://1.dam.s81c.com/p/0b52088cc2cc2f6d/G2_Leader_Spring_2023.svg#svgView(viewBox(0,0,50,50);preserveAspectRatio(xMidYMid)) . If you put this into a source it should just show part of the image instead of the full one. By changing width and height (e.g. with `viewBox(0,0,50,20)`) you can also change the aspect ratio. 
Though I am not sure if the solution with html attribute is sufficient, because we also have the use case to have different aspect ratios per breakpoint on certain components. Is it a problem to use the default `aspect-ratio` css style and pass it to the `img` tag? Otherwise a custom css style like the `--dds--carousel--page-size` on the [carousel](https://carbon-design-system.github.io/carbon-for-ibm-dotcom/canary/web-components/?path=/docs/components-carousel--default) would also work I think.

andy-blum:
@dratzenboeck-ibm I've updated the component, removing the `aspect-ratio` attribute. Now the component will respond to changes to its CSS aspect-ratio property and pass that value into the internal `img` tag.

andy-blum:
Updated testing instructions

annawen1 (approved):
LGTM

jkaeser (approved):
I wonder if the name is a little confusing. Styles applied to this `::part` will affect whichever image is in use, be that the default `` or whichever `<source>` the browser selects.

Not a big enough deal to block merging, but figured I'd mention it. This solution looks good to me.

andy-blum:
Good point, @jkaeser - I updated the attribute's value.

Issue #10443, opened 2023-05-10
[Table of Contents]: Selecting items using attribute "name" has unintended side-effects


andy-blum:
### Description

The `dds-table-of-contents` component harvests it's link items from it's DOM subtree, finding elements that match `[name]` and that do not have a tagName explicitly blocked by `DDSTableOfContents._tagNamesToAvoid`. ([source](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/table-of-contents/table-of-contents.ts#L178-L188))

This, however causes problems if a form is placed on the page within the TOC's boundaries, as those elements use the `name` attribute provided by the web platform to help with label-field association & identification. See MDN's list of HTML attributes and the elements that support them [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes).

The [_setTargets](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/table-of-contents/table-of-contents.ts#L362-L391) method finds all slotted elements and all elements in the subtrees of slotted elements that match the following criteria:
- Has the `name` attribute
- Has a title 
  - `data-title` attribute with value other than all whitespace
  - `innerText` property with value other than all whitespace
- Has a nodeName that is not explicitly block-listed

The following elements support the `name` attributes per the linked MDN article above:
- button
- form
- fieldset
- iframe
- input
- keygen
- object
- output
- select
- textarea
- map
- meta
- param

Of these, many could have both the `name` attribute and a non-whitespace `innerText` value and be erroneously added to the table of contents.

### Component(s) impacted

Table of Contents

### Carbon for IBM.com version

latest

### Severity

Severity 1 = The design is broken in a critical way that blocks users from completing tasks or damages the brand. Affects major functionality, no workaround.

### Application/website

ibm.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://codesandbox.io/p/sandbox/bold-dust-hxyqxd?file=%2Findex.html%3A203%2C57

### Steps to reproduce the issue (if applicable)

Place one of the name-supporting elements above into the table-of-contents element with a non-whitespace innerText

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
[...]

PR #10469, opened 2023-05-15
feat(table-of-contents): add self-filtering functionality

andy-blum:
### Related Ticket(s)

Related to: #10443 

### Description

This PR adds a small additional filter to the TOC's `_setTargets` method, preventing elements with the`no-toc` attribute from being added to the TOC.

### Changelog

**New**

- Table of Contents filters out elements with `no-toc` attribute


packages/web-components/src/components/table-of-contents/table-of-contents.ts
```diff
@@ -376,9 +376,9 @@ class DDSTableOfContents extends HostListenerMixin(
             elem.innerText.match(notWhiteSpace) ||
             elem.dataset.title?.match(notWhiteSpace);
           const hasNameAttr = elem.matches(selectorTarget);
-          const notExcluded = !tagNamesToAvoid.includes(
-            elem.tagName.toLowerCase()
-          );
+          const notExcluded =
+            !tagNamesToAvoid.includes(elem.tagName.toLowerCase()) &&
+            !elem.hasAttribute('no-toc');
```
FlashGooden, line 381: Ok, so we should be able to just drop this attribute in the `<form />`? 🪄

annawen1 (approved):
LGTM

jkaeser (commented):
@andy-blum Does this block specific elements, or will it prevent the ToC from finding targets within an entire DOM subtree?

jkaeser:
@andy-blum I wonder if it should block subtrees. Then it doesn't matter what issues adopters run into; they will have the power to disable whole chunks of the DOM. This is important when considering DOM output the adopter doesn't necessarily control (such as when using third party code). The current solution implies either:
1. The third party needs to change its output to avoid a visual error when used in conjunction with some framework outside of its control (backwards logic)
2. The adopter needs to write a script to modify third party code (potentially heavy lift)

andy-blum:
I'm fine with that. @carbon-design-system/carbon-for-ibm-com-admins do you have any opinons?

PR #10472, opened 2023-05-15
feat(content-switcher): implement icon-only variant and sync with v11

emyarod:
### Related Ticket(s)

#10108 

### Description

This PR updates the content switcher to match the upstream React version

### Changelog

**New**

- icon-only content switcher variant


oliviaflory:
@emyarod looking good here! A few visual fixes:

### 1. Horizontal shift in icon only
There is the smallest horizontal shift (maybe 1px) when using the arrow keys to navigate the icon only variant
content switcher shift May-31-2023 14-50-57

### 2. Selected state shouldn't show the hairline divider
The selected state in React (top) doesn't show the hairline divider line, but we are seeing it in WC (bottom). Happening in all variants (default & icon only).

Screenshot 2023-05-31 at 2 52 22 PM

### 3. Tool tip size
The first tooltip with "Table of contents" has a larger height than the other tooltips within the content switchers. Would be nice to get them to be the same height.

Screenshot 2023-05-31 at 4 06 55 PM
Screenshot 2023-05-31 at 4 07 00 PM
Screenshot 2023-05-31 at 4 07 04 PM

### 4. Focus state
Looks like the top and bottom stroke of the focus state might be 3px, but it should match the left & right strokes at 2px.
Screenshot 2023-05-31 at 4 11 10 PM

emyarod:
points 1 and 2 should now be resolved, unfortunately I am still unsure of how to address 3 and 4. there may be bigger issues with the icon-only tooltip at play for number 3, and for number 4 I'm confused because the focus outline values are exactly the same as the React version

oliviaflory (approved):
@emyarod Thanks for the update!

I wouldn't worry too much about 4, can definitely be addressed later because probably no one else will notice besides me 😄 

Approving overall and will leave it up to you and the devs to decide what to do for the release

packages/carbon-web-components/src/components/content-switcher/content-switcher.ts
```diff
@@ -12,7 +12,7 @@ import { property, customElement } from 'lit/decorators.js';
 import { prefix } from '../../globals/settings';
 import { forEach, indexOf } from '../../globals/internal/collection-helpers';
 import { NAVIGATION_DIRECTION, CONTENT_SWITCHER_SIZE } from './defs';
-import BXSwitch from './content-switcher-item';
+import CDSSwitch from './content-switcher-item';
```
ariellalgilmore, line 15:
```suggestion
import CDSContentSwitcherItem from './content-switcher-item';
```
Should this be changed?

ariellalgilmore (commented):
Hey! looking good! Just a small comment and if we could move any scoped styles and possibly add the layer import like this pr: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10567

packages/carbon-web-components/src/components/content-switcher/content-switcher.ts
emyarod: I wasn't sure about this because the name in React is still Switch and IconSwitch so I kept it

packages/carbon-web-components/src/components/content-switcher/content-switcher.scss
```diff
@@ -19,6 +20,10 @@ $css--plex: true !default;
   @extend .#{$prefix}--content-switcher;
```
ariellalgilmore, line 20:
```suggestion
  @extend .#{$prefix}--content-switcher;
  
  @include layer-tokens.emit-layer-tokens(1);
```

packages/carbon-web-components/src/components/content-switcher/content-switcher.ts
ariellalgilmore: ahh okay yeah this is a weird one... since we don't have the same switch or iconswitch in wc just contentSwitcherItem

ariellalgilmore (approved):
NICE! LGTM!

PR #10504, opened 2023-05-25
feat(masthead-l1): Removing the has pseudo-class

pjudge:
### Related Ticket(s)

### Description

The `:has` pseudo-class is not supported across all browsers yet (lookin' at you Firefox). This PR removes the usage of `:has` inside the masthead L1 styles, without changing the look on the front end. 

### Changelog

**Changed**

- Added a `classMap` to the L1 dropdown links element to dynamically add a class if the dropdown contains a wide column. 
- Added a similar functionality to the pre-existing headingClasses `classMap` to account for when headings inside dropdowns are not links. 
- Updated the corresponding styles to use the new classes instead of `:has`.

### DIFF
--- a/packages/web-components/src/components/masthead/masthead-l1.ts
+++ b/packages/web-components/src/components/masthead/masthead-l1.ts
@@ -402,7 +402,7 @@ class DDSMastheadL1 extends StableSelectorMixin(LitElement) {
     const normalColumns = menuSections.filter((section) => !(section.span > 1));
     const dropdownClasses = classMap({
       [`${prefix}--masthead__l1-dropdown-links`]: true,
-      [`has-wide`]: hasWideColumn,
+      [`${prefix}--masthead__l1-dropdown--has-column-wide`]: hasWideColumn,
     });
 
     return html`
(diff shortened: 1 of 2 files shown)


kennylam (commented):
@pjudge Thanks for this! The only thing I'd suggest is to use BEM naming conventions, e.g. something like

```css
.#{$prefix}--masthead__l1-dropdown-item--no-link
.#{$prefix}--masthead__l1-dropdown--column-wide
```

packages/web-components/src/components/masthead/masthead-l1.ts
```diff
@@ -752,8 +756,15 @@ class DDSMastheadL1 extends StableSelectorMixin(LitElement) {
...
 
+    let noLink: boolean = false;
+
+    if (!heading.url) {
+      noLink = true;
+    }
+
     const headingClasses = classMap({
       [`${prefix}--masthead__l1-dropdown-heading`]: true,
+      [`no-link`]: noLink,
```
andy-blum, line 767: As @kennylam mentioned, let's BEM-ify this class name.

For the sake of learning here, though:

Readability boost: we could reduce this all to a single line here.

```
const headingClasses = classMap({
      [`${prefix}--masthead__l1-dropdown-heading`]: true,
      "no-link": Boolean(heading.url) === false,
});
```
The `[\`...\`]` version is only necessary if we want to do the string interpolation like in the line above with `${prefix}` (which we will when we make a BEM version)

We can also do simple JS comparisons on the value side of the key/value pair, making the `noLink` variable and it's logic above unnecessary.

packages/web-components/src/components/masthead/masthead-l1.ts
```diff
@@ -400,6 +400,10 @@ class DDSMastheadL1 extends StableSelectorMixin(LitElement) {
 
     const wideColumns = menuSections.filter((section) => section.span > 1);
     const normalColumns = menuSections.filter((section) => !(section.span > 1));
+    const dropdownClasses = classMap({
+      [`${prefix}--masthead__l1-dropdown-links`]: true,
+      [`has-wide`]: hasWideColumn,
```
andy-blum, line 405: See comments at line 767

andy-blum (changes requested):

pjudge:
Thank you @kennylam and @andy-blum ! I have made the updates you suggested. 👍

kennylam (approved):
LGTM!

andy-blum (approved):

andy-blum:
e2e test failure seems unrelated. Re-running.
Screenshot 2023-06-01 at 4 12 38 PM

PR #10505, opened 2023-05-25
fix(ipcInfo): set encoding for ipcinfoCookie

annawen1:
### Related Ticket(s)

### Description

`ipcInfo` cookie was getting encoded twice - once by the `js-cookie` library by default which uses percent-encoding and then second by the `encodedURIComponent()`. This results in extra `25` being populated in the cookie value (ie. `cc%253Dde%253Blc%253Dde`). 

This PR sets the encoding method for the cookie to only use `encodedURIComponent()` to encode once. 

### Changelog

**Changed**

- overwrite default encoding/decoding method of cookie to `encodingURIComponent()/decodingURIComponent()`


kennylam (approved):
LGTM!

PR #10507, opened 2023-05-25
feat(progress-indicator): sync with v11

emyarod:
### Related Ticket(s)

#10092

### Description

This PR updates the progress switcher to match the v11 React component

### Changelog

**New**

- `space-equally` property for nonvertical progress indicator

**Changed**

- update and fix skeleton state
- sync attributes to match React props

**Removed**
- extraneous styles

### DIFF
--- a/packages/carbon-web-components/src/components/progress-indicator/progress-indicator.scss
+++ b/packages/carbon-web-components/src/components/progress-indicator/progress-indicator.scss
@@ -80,7 +80,7 @@ $css--plex: true !default;
   .#{$prefix}--progress-optional {
     margin-top: auto;
     position: initial;
-    margin-left: 2rem;
+    margin-left: $spacing-07;
   }
 
   .#{$prefix}--progress-step-button {
--- a/packages/carbon-web-components/src/components/progress-indicator/progress-indicator.scss
+++ b/packages/carbon-web-components/src/components/progress-indicator/progress-indicator.scss
@@ -48,13 +48,13 @@ $css--plex: true !default;
 :host(#{$prefix}-progress-step[vertical]),
 :host(#{$prefix}-progress-step-skeleton[vertical]) {
   display: list-item;
-  min-height: $spacing-12;
+  min-height: 3.625rem;
   width: initial;
   min-width: initial;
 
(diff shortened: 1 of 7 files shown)


emyarod:
I think I will need a huddle to discuss the plans for the interactive progress indicator in WC vs React

I have (finally) figured out how to update the focus state to use the styling we inherit from the core `styles` package. there MIGHT be some a11y ramifications with adopting the same styling approach but from what I can tell it should be okay

I have also updated the vertical container heights to match the upstream styles. for some reason the spacing values are not tokens but rather hard coded magic numbers, which makes me think they may need to be updated upstream

oliviaflory (approved):
@emyarod thanks for the update! 

It does look good from a visual perspective, other than the ux issue of the interactive vs default. I will leave it up to you and @kennylam to decide if we push the rest of the blockers to be fixed later.

packages/carbon-web-components/src/components/progress-indicator/progress-indicator.scss
```diff
@@ -48,13 +48,13 @@ $css--plex: true !default;
...
   display: list-item;
-  min-height: $spacing-12;
+  min-height: 3.625rem;
   width: initial;
   min-width: initial;
 
   svg {
     display: inline-block;
-    margin: 0.1rem 0.5rem;
+    margin: 0.1rem 0.25rem 0.1rem 0.5rem;
```
kennylam, line 57:
```suggestion
    margin: 0.1rem $spacing-02 0.1rem $spacing-03;
```

packages/carbon-web-components/src/components/progress-indicator/progress-indicator.scss
```diff
@@ -80,7 +80,11 @@ $css--plex: true !default;
   .#{$prefix}--progress-optional {
     margin-top: auto;
     position: initial;
-    margin-left: 2.25rem;
+    margin-left: 2rem;
```
kennylam, line 83:
```suggestion
    margin-left: $spacing-07;
```

kennylam (approved):
LGTM!

PR #10523, opened 2023-05-30
fix(masthead-l1): fix focusout logic

### LINKED ISSUE #10521: [Masthead v2 L1]: Clicking empty space in dropdown causes dropdown to close
### Description

JIRA: https://jsw.ibm.com/browse/ADCMS-3371

Screenshot 2023-05-30 at 10 40 45 AM

### Component(s) impacted

Masthead v2 L1

### Browser

_No response_

masthead-v2-dev

### Severity

### Application/website

ibm.com

### Package

### CodeSandbox example

https://ibm.com/products/turbonomic

_No response_

_No response_

### Code of Conduct

andy-blum:
### Related Ticket(s)

Closes #10521

### Description

Previous logic used only `Node.compareDocumentPosition()` and closed if the `currentTarget` did not contain `relatedTarget`. This forgot to account for `currentTarget === relatedTarget`.

### Changelog

**Changed**

- fixed focusout handler bug in masthead L1

### DIFF
--- a/packages/web-components/src/components/masthead/masthead-l1.ts
+++ b/packages/web-components/src/components/masthead/masthead-l1.ts
@@ -60,19 +60,25 @@ function handleDropdownClose(event: FocusEvent | KeyboardEvent) {
       }
       case 'focusout': {
         const { relatedTarget } = event as FocusEvent;
-        if (
-          !relatedTarget ||
-          !currentTarget ||
-          !(
-            (relatedTarget as Node).compareDocumentPosition(
-              currentTarget as Node
-            ) & 8
-          )
-        ) {
+        const hasUnknownTargets = !relatedTarget || !currentTarget;
+        let focusHasEscaped = false;
+
+        if (!hasUnknownTargets) {
+          const comparison = (relatedTarget as Node).compareDocumentPosition(
+            currentTarget as Node
+          );
+
+          focusHasEscaped =
+            !(comparison & 8) && // relatedTarget is not ancestor of currentTarget
+            relatedTarget !== currentTarget;
+        }
+
+        if (hasUnknownTargets || focusHasEscaped) {
           openDropdowns.forEach((el) => {
             el.classList.remove('is-open');
           });
         }
+
         break;
       }
       default:


packages/web-components/src/components/masthead/masthead-l1.ts
```diff
@@ -60,19 +60,25 @@ function handleDropdownClose(event: FocusEvent | KeyboardEvent) {
       }
       case 'focusout': {
         const { relatedTarget } = event as FocusEvent;
```
pjudge, line 62: is this syntax related to this? https://stackoverflow.com/a/33798750

packages/web-components/src/components/masthead/masthead-l1.ts
```diff
@@ -60,19 +60,25 @@ function handleDropdownClose(event: FocusEvent | KeyboardEvent) {
...
-        if (
-          !relatedTarget ||
-          !currentTarget ||
-          !(
-            (relatedTarget as Node).compareDocumentPosition(
-              currentTarget as Node
-            ) & 8
-          )
-        ) {
+        const hasUnknownTargets = !relatedTarget || !currentTarget;
```
pjudge, line 63: this is saying `hasUnknownTargets` is literally everything except `relatedTarget` or `currentTarget`?

pjudge: Or is this a boolean...?

packages/web-components/src/components/masthead/masthead-l1.ts
andy-blum: Yep, this is [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment). While it's not any more efficient when you need a single item, it's super handy when you need to pull a bunch of stuff out at once. For example:

```
const {
  relatedTarget,
  currentTarget,
  type
} = event;
```

packages/web-components/src/components/masthead/masthead-l1.ts
andy-blum: In [FocusEvents](https://developer.mozilla.org/en-US/docs/Web/API/FocusEvent) there is a `currentTarget` property which points to the element that _triggered_ the event. This is the element that lost focus in `focusout` events and the one that gained focus on `focusin` events.

They also have a special `relatedTarget` property that points to the "other" element in the equation.

In `focusout` events, focus leaves `currentTarget` and goes to `relatedTarget`.

In `focusin` events, focus leaves `relatedTarget` and goes to `currentTarget`.

The `hasUnknownTargets` variable here is a boolean that checks if either is falsy. The most common case for this is someone has navigated away from the page. In this case, focus has left the dropdown and gone to `undefined`.

packages/web-components/src/components/masthead/masthead-l1.ts
```diff
@@ -60,19 +60,25 @@ function handleDropdownClose(event: FocusEvent | KeyboardEvent) {
...
+
+        if (!hasUnknownTargets) {
+          const comparison = (relatedTarget as Node).compareDocumentPosition(
+            currentTarget as Node
+          );
+
+          focusHasEscaped =
+            !(comparison & 8) && // relatedTarget is not ancestor of currentTarget
+            relatedTarget !== currentTarget;
+        }
```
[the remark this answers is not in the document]
andy-blum: `Node.compareDocumentPosition` returns a bitmask, which was a new thing to me in this project as well. See [the article](https://www.lullabot.com/articles/bitmasks-javascript-computer-science-crash-course) I wrote about it.

Essentially, we're checking if the element that has focus is inside the DOM subtree of the `li` we're watching for `focusout`. If focus is still _inside_ the `li` element, we don't want to close.

`focusHasEscaped` will evaluate to true if focus is not in the subtree or in the top-level of the dropdown

Issue #10498, opened 2023-05-22
Autocomplete attribute from bx-input is not working in Firefox


Arunkarthik13:
### Description

In my code autocomplete is on for bx-input and I continuously searching one field in chrome and firefox. In chrome when I'm selecting same field getting previous search but not in Firefox. Please help me to resolve

### Component(s) impacted

The code I'm using:
![bx-input](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/30830178/c71a8f8e-0128-4969-b838-e952c3c850ef)

When I'm inspecting input field in chrome:
![autocomplete_Chrome](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/30830178/81ac6ef2-0f04-40bf-8cbb-3201d6e5adc2)

When I'm inspecting same field in firefox getting below: 
![automcomplete_firefox](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/30830178/c64b494c-77ad-488d-8bbe-f096fb996aa0)

### Browser

Chrome, Firefox

### Carbon for IBM.com version

v1.21.0

### Severity

Severity 1 = The design is broken in a critical way that blocks users from completing tasks or damages the brand. Affects major functionality, no workaround.

### Application/website

IBMLink

### Package

@carbon/web-components

### CodeSandbox example

Code provided above

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

ariellalgilmore:
https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10540

PR #10540, opened 2023-06-05
fix(input): autocomplete property

ariellalgilmore:
### Related Ticket(s)

#10498 
### Description

autocomplete was showing different results in the DOM in firefox and chrome. By removing the `?` it will now show the actual value set by the user instead of being a boolean. 

chrome:
Screenshot 2023-06-05 at 9 27 42 AM
firefox:
Screenshot 2023-06-05 at 9 27 32 AM

if approved will confirm the change gets made in cwc-v2 as well!
### Changelog

**Removed**

- `?` for `autocomplete` when returning `input`


emyarod (approved):
looks good to me!

Issue #10522, opened 2023-05-30
[content-block-cards with-video]: Kaltura thumbnails not sized properly


andy-blum:
### Description

JIRA: https://jsw.ibm.com/browse/ADCMS-3370

card-cta video thumbnails appear blurry

### Component(s) impacted

Card-CTA video type, as seen on https://ibm.com/products/envizi/resources

<img width="1201" alt="Screenshot 2023-05-30 at 11 06 37 AM" src="https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/25532785/9b4a1684-c2eb-47cd-88e4-26c1361e86ca">

---

This may be limited to the card-cta component, as the video-player-container seems unaffected at https://www.ibm.com/products/envizi/scope-1-2-ghg-accounting-reporting

<img width="783" alt="Screenshot 2023-05-30 at 11 08 34 AM" src="https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/25532785/d2908b59-d6f8-4bae-9274-93b0e9586ee4">

### Browser

Chrome

### Carbon for IBM.com version

v1.46.0

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

ibm.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://ibm.com/products/envizi/resources

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

m4olivei:
Fixed and merged to main.

PR #7914, opened 2021-12-21
fix(video): use component width to build video thumbnail

andy-blum:
### Related Ticket(s)

Fixes #7660

### Description

This PR addresses video poster sizes in two different places.

1) In the video-cta-composite component, we use the component's width to build the kaltura thumbnail URL.
2) In the video-player-composite component, that's not possible, so we hardcode a small value of `3` during the page load, and then use an event callback to replace the small image with the full-sized one which uses the component width in the URL

2 has the side effect of loading 2 images, so we minimize the first one by making it a 3px wide image. This gets so pixelated that to the end user it simply looks like a gradient that matches the color scheme of the final poster

Initial load:
Screen Shot 2021-12-21 at 3 52 39 PM

Post-load callback:
Screen Shot 2021-12-21 at 3 52 19 PM

### Changelog

**Changed**

- Video posters are built to the width of the component they'll need to cover

Issue #9698, opened 2022-11-29
Carbon for ibm.com Release Plan 2023 (v1.42.0-v1.54.0)


ljcarot:
**v1.42.0**
Jan 5, 2022: Code freeze
Jan 5 - Jan 13, 2023: Release Candidate(s) and testing
**Jan 17, 2023: Full Release**

**v1.43.0**
Jan 17, 2023: Code freeze
Jan 17 - Jan 27, 2023: Release Candidate(s) and testing
**Jan 31, 2023: Full Release**

**v1.44.0**
Feb 10, 2023: Code freeze
Feb 13 - Feb 24, 2023: Release Candidate(s) and testing
**Feb 28, 2023: Full Release**

**v1.45.0**
Mar 10, 2023: Code freeze
Mar 13 - Mar 24, 2023: Release Candidate(s) and testing
**Mar 28, 2023: Full Release**

**v1.46.0**
Apr 7, 2023: Code freeze
Apr 10 - Apr 21, 2023: Release Candidate(s) and testing
**Apr 25, 2023: Full Release**

**v1.47.0**
May 5, 2023: Code freeze
May 8 - May 19, 2023: Release Candidate(s) and testing
**May 23, 2023: Full Release**

**v1.48.0**
Jun 9, 2023: Code freeze
Jun 12 - Jun 23, 2023: Release Candidate(s) and testing
**Jun 27, 2023: Full Release**

**v1.49.0**
Jul 7, 2023: Code freeze
Jul 10 - Jul 21, 2023: Release Candidate(s) and testing
**Jul 25, 2023: Full Release**

**v1.50.0**
Aug 4, 2023: Code freeze
Aug 7 - Aug 18, 2023: Release Candidate(s) and testing
**Aug 22, 2023: Full Release**

**v1.51.0**
Sep 1, 2023: Code freeze
Sep 4 - Sep 15, 2023: Release Candidate(s) and testing
**Sep 19, 2023: Full Release**

**v1.52.0**
Sep 29, 2023: Code freeze
Oct 2 - Oct 13, 2023: Release Candidate(s) and testing
**Oct 17, 2023: Full Release**

**v1.53.0**
Oct 27, 2023: Code freeze
Oct 30 - Nov 10, 2023: Release Candidate(s) and testing
**Nov 14, 2023: Full Release**

**v1.54.0**
Dec 1, 2023: Code freeze
[...]

PR #10541, opened 2023-06-05
Adjust the handling of videoThumbnailUrl property to avoid small images from Kaltura API

m4olivei:
### Related Ticket(s)

https://jsw.ibm.com/browse/ADCMS-3370
https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/10522

### Description

Adjust the handling of videoThumbnailUrl property on CTA's that use the `VideoCTAMixin`. This should avoid the bug that we are currently seeing in production wher the CTA component `offsetWidth` is not yet computed, resulting in a small thumbnail width that is blurry when stretched to fit.

There isn't a great way to test the specific condition being observed [in production right now](https://ibm.com/products/envizi/resources). This appears to be a race condition, such that when the Kaltura API is leverged to fetch a thumbnail image, the width determined for the videos container via `offsetWidth` is not available and returns 0. The changes here should more reliably set the thumbnail after the `offsetWidth` is available. This borrows the logic from a [similar fix here](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/7914/files#diff-c16cc0206fb8b5202bfa8ddd7900037a71fb8b30235347e4c33ece727f8d8860R233-R235).

### Changelog

**Changed**

- `dds-video-cta-composite` no longer manages the`videoThumbnailUrl`, that handling is now encapsulated in the VideoCTAMixin, on the component where the `videoThumbnailUrl` is actually used.

### Testing

- [ ] Navigate to the Card Group > Default story. Select Add Media from the Knobs. The Video CTA cards should render an image with a width that matches the width of the card.
- [ ] Regression: Navigate to the CTA > Default story. Select CTA Type of "Card" and CTA style of "Video (video)". The Video CTA card should render with an image with a width that matches the width of the card.
- [ ] Regression: Navigate to the Card in Card > Default story. Select "video". The Video CTA card should render with an image with a width tha matches the width of the card.


packages/web-components/src/components/cta/video-cta-composite.ts
```diff
@@ -90,9 +90,12 @@ class DDSVideoCTAComposite extends ModalRenderMixin(
   // @ts-ignore: The decorator refers to this method but TS thinks this method is not referred to
   private async _handleRequestVideoData(event: CustomEvent) {
     const { href, videoName: customVideoName, videoDescription } = event.detail;
+    // Use a default value for thumbnail width in case the element doesn't have
+    // an offsetWidth yet.
+    const thumbnailWidth = (event?.target as HTMLElement)?.offsetWidth || 400;
```
ariellalgilmore, line 95: i wonder if we should use the same logic here: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/7914... set the default to 3 so it's a blurry thumbnail on load?

m4olivei: Hum, maybe. I need to study how it works to eventually replace the default 3px width with the one that's needed. Here it's set once when the components mount and sometimes that so quick the parent element doesn't have a defined width yet. I'll look into it tomorrow and get back. Thanks.

m4olivei: Ok, so couple of things here. I've gone with a slightly different approach, but really just changed where the same thing I was doing here is happening.

* I've moved the determination of the `videoThumbnailUrl` to the more natural place IMO, the `VideoCTAMixin`. The `videoThumbnailUrl` doesn't depend on the async Kaltura API request that's made here. It's put together sync using the local `KalturaPlayerAPI.getThumbnailUrl` method. That also allows us to change it according to the component width when the component updates. This [matches `video-player`](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/video-player/video-player.ts#L276-L278) in terms of when the video thumbnail url is set and should avoid the issue that we are seeing.
* The logic in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/7914 is different than here. There the `<dds-video-player>` is managed by us and is initiated with a `thumbnailUrl` attribute, and [when it's put together](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/video-player/video-player-composite.ts#L295), the component isn't mounted yet, so we don't know what size it will need, so we start with a tiny width of 3px. It then [updates it's thumbnail on `updated`](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/video-player/video-player.ts#L276-L278) when the component has rendered and has an positive `offsetWidth`.
* Here, we're not the ones creating the `<dds-card-group-item>`, that's the user, and it's up to them to initiate with a thumbnail (or not). Whether we get a `video-thumbnail-url` attribute or not, our code sets it using the Kaltura API (this _appears_ to be by design). The user provides a custom thumbnail using the `thumbnail` attribute, which is used first if present. Since we set the video thumbnail url, it's done in `updated`, when we know the `offsetWidth` and can straight away ask for the right size once from Kaltura.

I hope that makes sense. Let me know if anything is unclear.

ariellalgilmore (approved):
LGTM!

ariellalgilmore:
updates make sense to me! seeing some errors in the tests though

Screenshot 2023-06-07 at 8 45 32 AM

m4olivei:
Hum, Github is not seeing my new commits... attempting a Close / Re-open.

m4olivei:
Thanks @ariellalgilmore . Fixed the tests. The Typescript errors actually pointed to further cleanup that I did in the latest commit 🙌

ariellalgilmore:
looking at the cta video and not getting it to come through for [card](https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/10541/iframe.html?knob-CTA%20style%20(cta-style)=card&knob-CTA%20type%20(cta-type)=video&knob-Copy%20(copy):=Lorem%20ipsum%20dolor%20sit%20amet&knob-Content%20link%20href%20(href)=https://www.example.com&knob-Custom%20video%20title=Custom%20video%20title&knob-Custom%20video%20description=This%20is%20a%20custom%20video%20description&knob-Video%20ID%20(href)=1_9h94wo6b&args=&knob-Card%20type:=Card%20-%20default&knob-Add%20media:=true&knob-Number%20of%20cards:=5&knob-Cards%20per%20row:=dds-ce-demo-devenv--cards-in-row-3&knob-Grid%20mode:=collapsed&knob-Offset:=0&id=components-cta--default&viewMode=story)

ariellalgilmore (approved):
thanks for the updates! LGTM!

m4olivei:
Thanks @ariellalgilmore !

m4olivei:
@ariellalgilmore I'm not sure about some of the jobs:

* Looks like Percy is not happy b/c of failed image requests in one case. Not sure why the other two new screenshots are showing up here 🤷. Are those OK to bypass, can you approve them?
* The ci-check / a11y job is also failed on `main`, doesn't seem like it's expected to succeed at this point.
* There is one instance of [e2e-tests-parallel / web-components-tests that is failing](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/actions/runs/5214479287/jobs/9411655673?pr=10541). It's not clear why to me. Is this needed to pass and merge?

Hoping that this can make the [code freeze for tomorrow](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/9698) 🤞

PR #10544, opened 2023-06-05
fix(inputs): update autocomplete

ariellalgilmore:
### Related Ticket(s)

related to pr: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10540
### Description

updated v1 to adjust the autocomplete property for input.... adding the same update to cwc-v2

**Removed**

- `autocomplete` should not be a boolean when being used in `input`


emyarod (approved):
looks good to me!

annawen1 (approved):

Issue #10546, opened 2023-06-06
[GeoLocation service]: API endpoint has been retired, switch to DDO


annawen1:
### Description

The AP endpoint we use to grab the user's geolocation (country) when the DDO language is not available has been retired by the webmaster team. Now we need to grab the geolocation from the DDO which is populated by the IBM analytics script instead. 

### Component(s) impacted

masthead, footer

### Carbon for IBM.com version

s

### Severity

Severity 1 = The design is broken in a critical way that blocks users from completing tasks or damages the brand. Affects major functionality, no workaround.

### Application/website

d

### Package

@carbon/ibmdotcom-services

### CodeSandbox example

d

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #10547, opened 2023-06-06
fix(geolocation): remove geolocation utility, replace with new DDO service method - getLocation

annawen1:
### Related Ticket(s)

[GeoLocation service]: API endpoint has been retired, switch to DDO #10546

### Description

As the API endpoint we were using before to detect user's location (country code) has been retired by the webmaster team, new method is to grab from the DDO (`digitalData.user.location`) populated by the IBM analytics script. 

This PR adds a new method the DDO service, which grabs the location from the DDO in the page and removes all instances of the geolocation utility which no longer works.

### Changelog

**New**

- `getLocation()` method under the DDO service
- have the `Locale` service use the new DDO `getLocation` method instead

**Removed**

- instances of the `geolocation` utility


IgnacioBecerra (approved):
LGTM, thanks @annawen1!

Issue #10499, opened 2023-05-23
[pricing-table]: combine checkmark with copy


andy-blum:
### The problem

Request to allow for checkmark to be combined with text in table cell. Currently it's only possible to have an icon or text.

### The solution

![screenshot_2023-04-25_at_2 09 27_pm](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/25532785/2fbcfa08-8ff9-463d-92b5-15019e8e2ba1)

### Application/website

AEM

### Business priority

Medium Priority = upcoming release but is not pressing

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #10550, opened 2023-06-09
feat(pricing-table): support combined checkmark with copy

m4olivei:
### Related Ticket(s)

https://jsw.ibm.com/browse/ADCMS-3264
https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/10499

### Description

Allows a user of the pricing table to opt in to their slotted text to be shown alongsizde an icon within a pricing table cell.

### Changelog

**New**

- `dds-pricing-table-cell` now supports an `icon-text` attribute that when set will signal slotted text nodes to display next to the selected icon. If the icon attribute is not set, no change.

### Testing

- [ ] Search for "Pricing Table" in Storybook Story
- [ ] Check the "Show text with icon" checkbox
- [ ] Verify that the slotted text displays next to the icon.


packages/styles/scss/components/pricing-table/_pricing-table.scss
```diff
@@ -152,6 +152,16 @@
...
       justify-content: space-between;
     }
+
+    .#{$prefix}--pricing-table-cell-content {
+      display: flex;
+      align-items: start;
+      line-height: 1.25rem;
+
+      svg {
+        margin-right: $spacing-03;
```
jkaeser, line 162: Consider using the [logical property](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_logical_properties_and_values) here. This is especially relevant given our target consumer supports RTL languages.

```suggestion
        margin-inline-end: $spacing-03;
```

packages/web-components/src/components/pricing-table/pricing-table-cell.ts
```diff
@@ -26,6 +26,9 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
 class DDSPricingTableCell extends StableSelectorMixin(
   HostListenerMixin(DDSStructuredListCell)
 ) {
+  @property({ type: Boolean, attribute: 'icon-text', reflect: true })
+  iconText?: boolean = false;
```
jkaeser, line 30: You don't need the `?:` (what is this called??) here since you're setting a default value.
```suggestion
  iconText: boolean = false;
```

packages/web-components/src/components/pricing-table/pricing-table-cell.ts
```diff
@@ -69,6 +72,21 @@ class DDSPricingTableCell extends StableSelectorMixin(
     }
   }
 
+  /**
+   * Override to support having the slotted text side by side with the icon.
+   */
+  protected _renderIcon() {
```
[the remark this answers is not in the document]
jkaeser: 🖐️ My vote is to do this in Structured List Cell and inherit here.

packages/web-components/src/components/pricing-table/pricing-table-cell.ts
andy-blum: "[optional value](https://www.typescripttutorial.net/typescript-tutorial/typescript-optional-parameters/)"

andy-blum: I'm not sure how I feel about using a new parameter for this - would it be better to just assume that text & icon use together warrants displaying both?

m4olivei: I agree that would be way better. The thing is, when I went looking for examples that already exist in the wild, they all seem to have some default text "Cell with icon" that we wouldn't want suddenly appearing :(

https://www.ibm.com/products/instana/pricing
https://www.ibm.com/products/qradar-edr/pricing
https://www.ibm.com/products/planning-analytics/pricing

I suppose if they are adding that default text in a template, we could co-ordinate them removing that when this update gets adopted, but I'm not sure how realistic that would be. That's why I did this as an opt-in.

Open to changing it though, what do you think?

jkaeser: I asked Kayle and Antonio in a comment on the ticket if removing the default text is possible.

I hope it is, because I have fairly strong opinions about avoiding workarounds for AEM's implementation details. Those can always change, and our code could be left with features/behaviors that are unclear to future maintainers or even ourselves in a few months.

What do we think about documenting the decision in a comment should we need to leave this as opt-in?

jkaeser (commented):
Looks great! Will rubber-stamp once we hear back from the AEM folks about the `iconText` boolean.

packages/web-components/src/components/pricing-table/pricing-table-cell.ts
m4olivei: > I asked Kayle and Antonio in a comment on the ticket if removing the default text is possible.
> I hope it is, because I have fairly strong opinions about avoiding workarounds for AEM's implementation details.

Sounds good. Will wait on what they come back with. 🤞 

>  What do we think about documenting the decision in a comment should we need to leave this as opt-in?

Yep, can do.

jkaeser: @m4olivei Good news! The AEM team can remove the default text from existing content, so we're able to avoid the opt-in approach here.

m4olivei: Excellent! I'll update things here accordingly.

jkaeser (changes requested):
@m4olivei Thanks for updating! I noticed a couple issues while checking the [deploy preview](https://ibmdotcom-web-components-experimental.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/10550/index.html?path=/story/components-pricing-table--default&knob-highlighted%20column=2&knob-highlighted%20label=Featured&knob-icon-text=Lorem%20ipsum%20dolor%20sit%20amet&knob-number%20of%20columns=3&knob-section%20heading=Optional%20section%20heading):

1. As I understand it, the text shouldn't receive the same green color as the checkmark.
2. The **icon-text** knob only seems to work in the default story.

jkaeser (approved):

m4olivei:
@kennylam / @ariellalgilmore this has been approved by the Design Team. Off to you for the final nod. Let me know if you have any questions. Thank you!

packages/styles/scss/components/pricing-table/_pricing-table.scss
```diff
@@ -152,6 +152,17 @@
...
+    .#{$prefix}--pricing-table-cell-content {
+      display: flex;
+      align-items: start;
+      line-height: 1.25rem;
+
+      svg {
+        margin-inline-end: $spacing-03;
+        flex-shrink: 0;
+      }
+    }
```
jkaeser, line 165: @m4olivei You know, it just dawned on me – should these styles be moved over to the Structured List stylesheet?

jkaeser: Answered my own question. The wrapper element that has this class only appears in Pricing Table cells, which I believe means these styles are only necessary in this instance.

m4olivei: Heh, fun timing. Was just coming here to say that. I think we could restructure that wrapper element to be part of the parent structured list component, but it gets tricky, b/c there are some cases for the general strucutred list where you really don't want to inherit the flexbox layout.

kennylam (approved):
LGTM!

m4olivei:
Here is a patch PR against the current release: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10595

PR #10552, opened 2023-06-09
feat(masthead-v2): Add selected styles to L1

### LINKED ISSUE #10500: [masthead]: Add "selected" styles to L0/L1 masthead
### The problem

JIRA: https://jsw.ibm.com/browse/ADCMS-3352

One piece of the L0/L1 work that fell through the cracks was styles related to the currently selected page/web segment. We'll need to add those styles back in.

### The solution

Screenshot 2023-05-23 at 2 43 44 PM
Screenshot 2023-05-23 at 2 44 03 PM
Screenshot 2023-05-23 at 2 44 19 PM

### Application/website

ibm.com

### Business priority

_No response_

### Examples

_No response_

### Code of Conduct

jkaeser:
### Related Ticket(s)

Resolves #10500

### Description

Adds "active" menu item styling to L1 menu items.

### Changelog

**New**

- Adds "active" menu item styling to L1 menu items.

**Changed**

- Changes "active" L0 menu item styling if an L1 is present.
- Cleans up `DDSMasthead` Storybook knobs and default data to more accurately represent components.

### DIFF
--- a/packages/web-components/src/components/masthead/masthead-l1.ts
+++ b/packages/web-components/src/components/masthead/masthead-l1.ts
@@ -18,6 +18,7 @@ import {
   queryAll,
 } from 'lit-element';
 import settings from 'carbon-components/es/globals/js/settings.js';
+import root from 'window-or-global';
 import ddsSettings from '../../internal/vendor/@carbon/ibmdotcom-utilities/utilities/settings/settings';
 import StableSelectorMixin from '../../globals/mixins/stable-selector';
 import styles from './masthead-l1.scss';
@@ -102,6 +103,18 @@ class DDSMastheadL1 extends StableSelectorMixin(LitElement) {
    */
   @property()
   l1Data?: MastheadL1;
+
+  /**
+   * The URL path of the current document root.
+   */
+  @property()
+  currentUrlPath?: string = root.location?.href;
+
+  /**
+   * The English title of the selected nav item.
+   */
+  @property({ attribute: 'selected-menu-item' })
+  selectedMenuItem?: string;
 
   /**
    * Whether the current viewport is below 800px or not
@@ -507,8 +520,8 @@ class DDSMastheadL1 extends StableSelectorMixin(LitElement) {
       [`${prefix}--masthead__l1-dropdown-column-wide`]: hasWide && isWide,
     });
     return hasWide
-      ? html` <div class=${classes}>${renderedSections}</div> `
-      : html` ${renderedSections} `;
+      ? html`<div class=${classes}>${renderedSections}</div>`
+      : html`${renderedSections}`;
   }
 
   /**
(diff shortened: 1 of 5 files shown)


packages/web-components/src/components/masthead/masthead-l1.ts
```diff
@@ -103,12 +104,30 @@ class DDSMastheadL1 extends StableSelectorMixin(LitElement) {
   @property()
   l1Data?: MastheadL1;
 
+  /**
+   * The URL path of the current document root.
+   */
+  @property()
+  currentUrlPath?: string = root.location?.href;
```
m4olivei, line 111: Thinking out loud. Would `root.location.href` ever change after this is initialized in a way we would have to be concerned about? Thinking of, for example, a single page app that manages it's own routing client side for instance.

m4olivei (approved):
This is looking great. Nice work. Nothing here blocking, but did want to point out a couple other things along with my comments.

All of the sub-menu items in the L1 are titled "Link item", so when you enter that, you end up (correctly) getting a long blue underline. I wonder if it's worth adding some variability in the titles of the sub-menu items here to have a more accurate depiction of how things would play out in production?
so_many_link_item

Designs didn't specify but wanted to point out that focus + selected gives you a blue top, right, and left border with a thicker gray bottom one. Not sure if that's anything?


focus_plus_selected

packages/web-components/src/components/masthead/masthead-l1.ts
```diff
@@ -739,6 +758,60 @@ class DDSMastheadL1 extends StableSelectorMixin(LitElement) {
...
+
+    // Check for manually set selected item.
+    if (selectedMenuItem) {
+      this.selectedElements = [...menuItems, ...allLinks].filter(
+        (el) =>
+          el.textContent?.trim()?.toLowerCase() ===
+          selectedMenuItem.trim().toLowerCase()
+      );
+    }
+    // Fall back to automated selection based on URL.
```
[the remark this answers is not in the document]
jkaeser: I've never heard anything definitive about this either, but I also imagine most of production uses the automated behavior.

I wouldn't say the attribute is _primarily_ for Storybook. It's an important "escape hatch" if the automatic selection doesn't work for your environment. Or perhaps there's some business case for highlighting a certain section of the menu that doesn't strictly contain the current page.

packages/web-components/src/components/masthead/masthead-l1.ts
m4olivei: Good point, makes sense. Thannks John.

areagan1030:
Since Alycia and I are not on the pending reviewers, dropping a note here that this is approved on our behalf.

jkaeser:
@annawen1 @emyarod This has code and design approvals from folks on our end. Just needs approval from a maintainer and we can ship this 🚀

PR #10555, opened 2023-06-13
10174: Esc to close Cloud Masthead

### LINKED ISSUE #10174: [Cloud-Masthead v1]: a11y Feature/improvement -  support `Esc` key to close the Dropdown in masthead component
### The problem

a11y keyboard accessible Feature/improvement -  support `Esc` key to close the Dropdown in masthead

Docs Microservice uses Carbon masthead component.
User requested below missing support for keyboard accesibility

`Esc key lose its function. We use Esc key to close the drop-down list . but now we have to use tab key +shift key to return to the menu title and use enter key to collapse it`

Related Issue: https://github.ibm.com/Bluemix/Bluemix-doc-framework/issues/4551#issuecomment-53637492

### The solution

support `Esc` key to close the Dropdown in masthead component

### Application/website

Bluemix-doc-framework

### Business priority

_No response_

### Examples

_No response_

### Code of Conduct

pjudge:
### Related Ticket(s)

Closes #10174 

### Description

The Cloud masthead (and other mastheads) simple menus were not closing upon hitting Esc if focus was inside the submenu. This PR fixes that. 

### Changelog

**Changed**

- Adds ability to close Cloud Masthead non-megamenu (and all non-megamenus) with Esc key, just as megamenus already do.


jkaeser (approved):
👏 This looks excellent. I've confirmed the fix works in both the web component packages' deploy previews. Great work!

If you're interested, I have a couple notes that won't functionally change anything but are some opportunities to make this a smidge more idiomatic:

1. Instead of adding an event listener in the `connectedCallback()`, try using the `@HostListener()` decorator above the `_handleKeydownTrigger` method. See the `_handleBlur` method for an example. This way we're consistent with how we're handling event listeners.
2. Since this event listener is being set on the custom element itself, we shouldn't need to set it on the `<a>` tag with the `@keydown` expression anymore.
3. I noticed both the `BXHeaderMenu._trigger` and `DDSMegamenuTopNavMenu._topMenuItem` properties store references to the same thing: the top-level menu item. Seems like an opportunity to reduce some redundancy. Let me know if you want to talk about this one.

packages/carbon-web-components/src/components/ui-shell/header-menu.ts
```diff
@@ -24,7 +24,7 @@ const { prefix } = settings;
  * Header menu.
  *
  * @element bx-header-menu
- * @csspart trigger The trigger button.
+ * @csspart topMenuItem The trigger button.
```
jkaeser, line 27: I think this was referring to `[part="trigger"]`, which hasn't changed, so I don't think we need this change.

jkaeser (changes requested):
Thanks for addressing my feedback! Just one change needs reverting and I'd say we're good to go here.

packages/carbon-web-components/src/components/ui-shell/header-menu.ts
pjudge: Oh! I thought it was referring to _trigger, from line 37, which did change?

pjudge: Discussed in a huddle! I made the change and am resolving this

jkaeser:
@annawen1 Looks like Percy failed to generate some screenshots and is blocking this merge. Are you able to resolve that, or do we need to reach out to the design team?

kennylam:
> Looks like Percy failed to generate some screenshots and is blocking this merge. Are you able to resolve that, or do we need to reach out to the design team?

Fixed!

Issue #10581, opened 2023-06-21
[Locale Cookie Not Being Set In Web Component Locale Selector]: Locales cannont be properly set when using the webcomponents version of the footer/locale selector


whoisbob:
### Description

Unlike it's react version, the web component locale selector (both dropdown and modal) does not support setting the ipcInfo cookie when a user selects a new locale. This cookie is responsible for persisting the set locale across IBM pages. 

### Component(s) impacted

Footer, Locale Modal 

### Browser

Chrome, Firefox

### Carbon for IBM.com version

1.32.0

### Severity

Severity 1 = The design is broken in a critical way that blocks users from completing tasks or damages the brand. Affects major functionality, no workaround.

### Application/website

MyIBM

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://codesandbox.io/s/cool-sun-4ppk7w?from-embed

### Steps to reproduce the issue (if applicable)

 Create a react app that uses react version of the footer container component. Proceed to change the locale with the locale switcher and note in the dev tools an ipcInfo cookie being changed.

Perform the same thing in a webcomponent instead. Note no change to an ipcInfo cookie.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

annawen1:
From what I can see, the language-only dropdown in the footer doesn't set the ipcinfo cookie - that only happens for the locale modal. I don't think there's defined logic for how the ipcinfo cookie (which consists of both language and country code) is set for the language-only since it only provides the language code and there's no country code. I'll open up another ticket for the language-only dropdown with the team to discuss. 

In the meantime, there is a PR for setting the ipcInfo cookie on locale-modal selection.

PR #10587, opened 2023-06-22
fix(locale-modal): set ipcinfo cookie with selected locale

annawen1:
### Related Ticket(s)

[[Locale Cookie Not Being Set In Web Component Locale Selector]: Locales cannont be properly set when using the webcomponents version of the footer/locale selector#10581](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/10581)

### Description

The web components version of the locale modal doesn't set the locale in the `ipcInfo` cookie when user selects a locale from the modal. This cookie ensures that the user's locale preference/selection persists throughout IBM.com.

### Changelog

**Changed**

- set `ipcInfo` cookie on `dds-locale-item` click


IgnacioBecerra (approved):
LGTM!

PR #10601, opened 2023-06-26
fix(card-link): make card link width match card-group

annawen1:
### Related Ticket(s)

[[Card link] grid bug#10580](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/10580)

### Description

Make the `card-link`/`card-link-cta` width match the `card-group` on the page. This is done within the `content-block` component where it takes the `cards-per-row` property and applies the appropriate styles to the `card-link`.

Screenshot 2023-06-26 at 12 21 42 PM

Screenshot 2023-06-26 at 12 21 55 PM

### Changelog

**Changed**

- had logic in the `content-block` component to detect if `card-group` element is a child and apply the same styles to the `card-link` in the footer


packages/web-components/src/components/content-block/content-block.scss
```diff
@@ -21,6 +22,10 @@
   display: block;
 }
 
+:host(#{$dds-prefix}-content-block) ::slotted([slot='footer']) {
+  background-color: pink;
```
emyarod, line 26: is this background color intentional?

emyarod (approved):
looks good to me, just one question about the background-color!

packages/web-components/src/components/content-block/content-block.scss
annawen1: oh crap, was testing if that selector worked! thanks for catching that!

RichKummer (approved):
Perfect, thanks @annawen1 !

Issue #10501, opened 2023-05-23
[Date range picker]: The date range picker resets after choosing the start date and clicking on the 2nd box


AlessandroPomponio:
### Description

When selecting a date range I usually want to select the start date first, and the end date after.

If I have to go back a few pages to pick the start date, however, I’d rather click on the 2nd datepicker box to be prompted with another selector that’s reset on the current date to avoid having to scroll through the pages again.

When I do this, however, the selection gets reset to the first date box, and the only way to do it without having to scroll again seems to be to select the end date first, which feels counterintuitive to me (see video)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/10339005/316e27aa-2c12-45ed-921c-c15944b735b7

### Component(s) impacted

Date picker - range with calendar

### Browser

Microsoft Edge

### Carbon for IBM.com version

v1.27.0

### Severity

Severity 1 = The design is broken in a critical way that blocks users from completing tasks or damages the brand. Affects major functionality, no workaround.

### Application/website

ST4SD

### Package

@carbon/ibmdotcom-web-components, @carbon/web-components

### CodeSandbox example

https://web-components.carbondesignsystem.com/?path=/story/components-date-picker--range-with-calendar

### Steps to reproduce the issue (if applicable)

Refer to video above.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [ ] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

m4olivei:
Adding some notes on the history for flatpickr usage as I investigate:

* Flatpickr added prior to carbon web components. See now archived repo https://github.com/carbon-design-system/carbon-web-components for relevant history.
* Flatpickr was introduced at version 4.6.1 with the initial date picker component implementation here: https://github.com/carbon-design-system/carbon-web-components/pull/132
[...]

PR #10603, opened 2023-06-26
fix(date-picker): prevent range inputs from clearing prematurely

m4olivei:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/10501
https://jsw.ibm.com/browse/ADCMS-3429

Upstream flatpickr tickets / PRs

- https://github.com/flatpickr/flatpickr/pull/2801
- https://github.com/flatpickr/flatpickr/issues/2918

### Description

Fixes bugs with the date picker. Specifically we fix the reported bug where bluring from the input after a range selection would clear the first input. We then end up fixing a couple of other related bugs to get us back to a usable range input. Notably the usability issue reported wherin there ends up being a lot of month paging for far-past or far-future ranges to get a complete range is not addressed here, as that behavior is by design.

Note that, as dicussed with @IgnacioBecerra, @kennylam and @andy-blum  at office hours, the underlying [flatpickr](https://github.com/flatpickr/flatpickr) library is not seeing a concerted effort at maintenance. There are many bug reports piled up in their issue queue ([over 600 at time of writing](https://github.com/flatpickr/flatpickr/issues)) and not much commit activity or late. As a result, here I've updated to the latest tagged version of flatpickr (4.6.13), and then used [`yarn patch`](https://yarnpkg.com/cli/patch) to include necessary bug fixes. Upstream PR's and issues are linked above.

### Changelog

**Changed**

- date picker with range: fix a bug with date picker where bluring away from the first input after having populated a range value, would clear the first input
- date picker with range: fix a bug where loosing focus from the calendar picker after having populated a range value, would clear the first input
- date picker with range: fix a bug where clicking either input after having set a range would not show the selected range set in the calendar


kennylam (approved):
LGTM thanks @m4olivei!

m4olivei:
> @m4olivei Do you think this is fixable here, or should it be a separate issue?

Nice catch @jkaeser! Looks like it's a simple fix here and also removes some overrides we have to allow flatpickr to work the way it wants to, which is a nice win. All is looking OK locally to me. Will give it another check once built, but 🤞 we should be good now.

Issue #9729, opened 2022-11-18
Filterable doesn't work in multi-select


UXsree:
### Description

By adding filterable to <bx-multi-select>, it should ideally list out the items similar to what we type in the input field and hide the remaining items.
This behavior is not happening as expected and all items are listed instead.

### Component(s) impacted

https://web-components.carbondesignsystem.com/?path=/story/components-multi-select--filterable

### Browser

Chrome, Safari, Firefox, Microsoft Edge

### Carbon Web Components version

1.21.0

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

https://web-components.carbondesignsystem.com/?path=/story/components-multi-select--filterable

### CodeSandbox example

https://codesandbox.io/s/falling-sound-i343jh?file=/index.html

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-web-components/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-web-components/issues) for duplicate issues

rccc:
Hello,

This bug is still here ! ("carbon-components": "10.58.4"). Adding "filterable" attribute does not let us filter the option list.

annawen1:
@rccc I see this working in our latest Carbon Web Components version (1.28.0) in our storybook [here](https://web-components.carbondesignsystem.com/?path=/story/components-multi-select--filterable)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/54281166/ad368c99-94fc-48b1-94fa-3571c44cecf9

annawen1:
Actually, testing in codesandbox, I see the issue - I'll open a PR with the fix

PR #10608, opened 2023-06-27
fix(multi-select): move filterable attribute out of scope for styles

annawen1:
### Related Ticket(s)

[Filterable doesn't work in multi-select#9729](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/9729)

### Description

The filterable option for `multi-select` doesn't appear to work because the styles to hide the options are not coming through when pulling in the package externally outside of storybook.

You can see it currently doesn't appear to filter in this codesandbox: https://codesandbox.io/p/sandbox/dazzling-burnell-4my6t8

 Moving the selector out of the scope should fix this. 

### Changelog

**Changed**

- use separate selector rule when referencing the filterable attribute

### DIFF
--- a/packages/carbon-web-components/src/components/multi-select/multi-select.scss
+++ b/packages/carbon-web-components/src/components/multi-select/multi-select.scss
@@ -37,13 +37,12 @@ $css--plex: true !default;
       outline: none;
     }
   }
+}
 
-  &[filterable] {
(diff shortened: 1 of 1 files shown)


IgnacioBecerra (approved):
lgtm, thanks anna!!

emyarod (approved):
looks good to me!

PR #10620, opened 2023-06-29
chore(storybook): added content section playground

IgnacioBecerra:
### Related Ticket(s)

### Description

### Changelog

**New**

- {{new thing}}

**Changed**

- {{changed thing}}

**Removed**

- {{removed thing}}


oliviaflory:
@IgnacioBecerra thank you for this and sorry it's taken me so long to respond 😞 

1. If this is going to be permanent, could we move the `Playground` under the `Content section` story? Though maybe that wouldn't work because we couldn't have the multiple stories that way 🤔 or rename to `Content section playground` with the sub stories under that?
2. I'm going to give you a smaller list to show case within the content section component – some of the current examples are being deprecated, or are sections themselves and squished, or are so small it doesn't make sense on their own. So here's what we want to keep:
- Callout quote
- Callout media
- card group
- card in card
- carousel
- Content block
- Content group
- Content item horizontal
- Content item
- CTA block
- CTA section
- Feature card
- Image
- Lightbox media viewer
- Link list
- logo grid
- Notice Choice
- Pictogram item
- Pricing table
- Quote
- Structured list
- Tabs extended
- Tabes extended with media (without the section heading)
- Tag group
- Video player
3. The Default story is nice to see each component individually. Could you add another story example with all of the components dumped in one content section all together? So I can see how they're stacking and if they're crashing (or not). Imagining something like the example below, but with all the components listed above:
 
Frame 3

IgnacioBecerra:
@oliviaflory Sure thing, I'll get right on that 👍 

As for the first point, this was originally just for audit purposes, but we can definitely get this added under Content Section as its own story variation.

IgnacioBecerra:
@oliviaflory Updated the Default story with only the components you listed above (except Pricing table, there seems to be some weird thing about its story, it thinks its Pictogram item? lol)
Screenshot 2023-07-12 at 5 22 43 PM


But yeah, the new Kitchen Sink story variation has all the components you mentioned (except Lightbox media viewer because thats just a modal). 

Let me know if you want some other variations or etc!

oliviaflory:
@IgnacioBecerra YIKES SO RUDE OF ME – these are perfect 😄

PR #10650, opened 2023-07-10
fix(tabs-extended): ensure accordion doesn't load on desktop, and tabs don't load on mobile

### LINKED ISSUE #10618: [Tabs extended - with media]: optimize component to lower DOM node count
Currently this component has a DOM node count which can be lowered. Solutions to consider are removing duplicate/unnecessary nodes and lazy loading content in hidden tabs.

Node audit

| Nodes| Comments | Inters |
|--------|--------|--------|
| 1682 | 265 | 842 |

pjudge:
### Related Ticket(s)

Closes #10618 

### Description

The extended tabs component was loading both the tabs component (shown on desktop) and the accordion component (shown on mobile) at the time of page load, despite the window size, and hiding one with CSS. This contributes to an excess of nodes loading in the DOM unnecessarily. 

This PR changes the rendering of the tabs extended and tabs extended with media components to look at the window size and load either the accordion OR the tabs appropriately, thus decreasing nodes. 

### Changelog

**Changed**

- Tabs extended and Tabs Extended with Media now look at window size and load either tabs (desktop) or accordion (mobile), not both.

### DIFF
--- a/packages/web-components/src/components/tabs-extended/tabs-extended.ts
+++ b/packages/web-components/src/components/tabs-extended/tabs-extended.ts
@@ -22,6 +22,9 @@ import { carbonElement as customElement } from '../../internal/vendor/@carbon/we
 const { prefix } = settings;
 const { stablePrefix: ddsPrefix } = ddsSettings;
 
+// Magic Number: 1056px matches 'carbon--breakpoint(lg) mixin'.
+const layoutBreakpoint = window.matchMedia(`(max-width: 1056px)`);
+
 /**
  * A component to present content inside a tabbed layout.
  *
@@ -29,6 +32,18 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
  */
 @customElement(`${ddsPrefix}-tabs-extended`)
 class DDSTabsExtended extends StableSelectorMixin(LitElement) {
+  /**
+   * Whether the we're viewing smaller or larger window.
+   */
+  _isMobileVersion = layoutBreakpoint.matches;
+
+  // Allows conditional rendering of tabs/accordion.
+  firstUpdated() {
+    layoutBreakpoint.addEventListener('change', () => {
+      this._isMobileVersion = layoutBreakpoint.matches;
+      this.requestUpdate();
+    });
+  }
   /**
    * Child tab components.
    */
@@ -166,25 +181,28 @@ class DDSTabsExtended extends StableSelectorMixin(LitElement) {
     this._isLTR = window.getComputedStyle(this).direction === 'ltr';
     this._activeTabIndex = parseInt(this._activeTab, 10);
 
-    this._tabItems.map((tab, index) => {
-      (tab as DDSTab).selected = index === this._activeTabIndex;
-      (tab as DDSTab).setIndex(index);
-      const navLink = this.shadowRoot!.querySelectorAll(
-        `.${prefix}--tabs__nav-link`
-      )[index];
-      const navText = navLink!.querySelector('div p');
-      if (navText!.scrollHeight > navText!.clientHeight) {
-        const label = (tab as DDSTab).getAttribute('label');
-        if (label) {
-          navLink!.setAttribute('aria-label', label);
-          navLink!.setAttribute('hasTooltip', label);
+    // Set aria-label on tabs for desktop.
+    if (!this._isMobileVersion) {
+      this._tabItems.forEach((tab, index) => {
+        (tab as DDSTab).selected = index === this._activeTabIndex;
+        (tab as DDSTab).setIndex(index);
+        const navLink = this.shadowRoot!.querySelectorAll(
+          `.${prefix}--tabs__nav-link`
+        )[index];
+        const navText = navLink!.querySelector('div p');
+        if (navText!.scrollHeight > navText!.clientHeight) {
+          const label = (tab as DDSTab).getAttribute('label');
+          if (label) {
+            navLink!.setAttribute('aria-label', label);
+            navLink!.setAttribute('hasTooltip', label);
+          }
         }
-      }
-      return tab;
-    });
+        return tab;
+      });
+    }
   }
 
-  protected _renderAccordion(): TemplateResult | string | void {
+  protected _renderAccordion(): TemplateResult {
     const { _tabItems: tabs } = this;
     return html`
       <ul class="${prefix}--accordion">
@@ -222,7 +240,7 @@ class DDSTabsExtended extends StableSelectorMixin(LitElement) {
     `;
   }
 
-  protected _renderTabs(): TemplateResult | string | void {
+  protected _renderTabs(): TemplateResult {
     const { _tabItems: tabs } = this;
     return html`
       <div class="${prefix}--tabs">
@@ -280,9 +298,13 @@ class DDSTabsExtended extends StableSelectorMixin(LitElement) {
   orientation = ORIENTATION.HORIZONTAL;
 
   render() {
+    const {
+      _isMobileVersion: isMobileVersion,
+    } = this;
+
     return html`
       <div class="${this._getOrientationClass()}">
-        ${this._renderAccordion()} ${this._renderTabs()}
+        ${isMobileVersion ? this._renderAccordion() : this._renderTabs()}
         <div class="${prefix}--tab-content">
           <slot @slotchange="${this._handleSlotChange}"></slot>
         </div>
--- a/packages/web-components/src/components/tabs-extended-media/tabs-extended-media.ts
+++ b/packages/web-components/src/components/tabs-extended-media/tabs-extended-media.ts
@@ -31,6 +31,10 @@ class DDSTabsExtendedMedia extends DDSTabsExtended {
   sectionHeading = 'true';
 
   render() {
+    const {
+      _isMobileVersion: isMobileVersion,
+    } = this;
+
     return html`
       <div class="${prefix}--tabs-extended-media">
         ${this.sectionHeading === 'true'
@@ -41,7 +45,7 @@ class DDSTabsExtendedMedia extends DDSTabsExtended {
             `
           : undefined}
         <div class="${prefix}--tabs-extended">
-          ${this._renderAccordion()} ${this._renderTabs()}
+          ${isMobileVersion ? this._renderAccordion() : this._renderTabs()}
           <div class="${prefix}--tab-content">
             <slot @slotchange="${this._handleSlotChange}"></slot>
           </div>


jkaeser (approved):
Code looks good to me, as does the end result in Storybook. Excellent work!

andy-blum (changes requested):
This is definitely on the right track but there are a few more optimizations available.

1. `.bx--tab-content` is still present on mobile breakpoints. We should avoid rendering this chunk of the shadowroot when `!this._isMobileVersion`

Screenshot 2023-07-11 at 9 37 28 AM

2. The mobile version (the accordion) is still duplicating all content into the shadowDOM, effectively doubling the memory footprint of this component. We should use slots to render the `dds-tab` elements into the accordion to keep that shadowDOM as minimal as possible. This will likely require something similar the following markup:

```html
<ul class="bx--accordion">

  <li class="bx--accordion__item bx--accordion__item--active">
    <button class="bx--accordion__heading">...</button>
    <div class="bx--accordion__content">
      <slot name="tab-1"></slot>
    </div>
  </lI>
  
  <li class="bx--accordion__item">
    <button class="bx--accordion__heading">...</button>
    <div class="bx--accordion__content">
      <slot name="tab-2"></slot>
    </div>
  </lI>
  
  <li class="bx--accordion__item">
    <button class="bx--accordion__heading">...</button>
    <div class="bx--accordion__content">
      <slot name="tab-3"></slot>
    </div>
  </lI>
</ul>
```

To make the `dds-tab` elements slot in appropriately, we'd need to also react in our `updated()` method to add/remove the `slot="..."` values to make each tab slot correctly.

packages/web-components/src/components/tabs-extended/tabs-extended.ts
```diff
@@ -166,25 +181,28 @@ class DDSTabsExtended extends StableSelectorMixin(LitElement) {
...
-      )[index];
-      const navText = navLink!.querySelector('div p');
-      if (navText!.scrollHeight > navText!.clientHeight) {
-        const label = (tab as DDSTab).getAttribute('label');
-        if (label) {
-          navLink!.setAttribute('aria-label', label);
-          navLink!.setAttribute('hasTooltip', label);
+    // Set aria-label on tabs for desktop.
+    if (!this._isMobileVersion) {
+      this._tabItems.forEach((tab, index) => {
```
m4olivei, line 186: I like the switch to a `forEach` since we weren't using the return from `map` anyway ➕ .

packages/web-components/src/components/tabs-extended/tabs-extended.ts
```diff
@@ -166,25 +181,28 @@ class DDSTabsExtended extends StableSelectorMixin(LitElement) {
...
+          const label = (tab as DDSTab).getAttribute('label');
+          if (label) {
+            navLink!.setAttribute('aria-label', label);
+            navLink!.setAttribute('hasTooltip', label);
+          }
         }
-      }
-      return tab;
-    });
+        return tab;
```
m4olivei, line 200: Small suggestion to remove the `return` statement, since the return from the callback of a `forEach` is discarded anyway (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach).

```suggestion
```

m4olivei (commented):
Just a tiny suggestion from a static review. I saw Andy and John took a good look at the integration, so I'll stop there.

packages/web-components/src/components/tabs-extended/tabs-extended.ts
pjudge: Wis I could take credit but it was all Andy :D

packages/web-components/src/component-mixins/media-query/media-query.ts
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * The method that is invoked when crossing over the breakpoint.
+     */
+    abstract mediaQueryCallback(): void;
+
+    firstUpdated() {
+      this._mediaQuery.addEventListener(
+        'change',
+        this.mediaQueryCallback.bind(this)
+      );
```
andy-blum, line 38: Long-term vision: What if we have a component down the road that needs to respond in someway to _multiple_ breakpoints? Should we add all the major carbon breakpoints (sm/md/lg/xlg/max) & their associated callbacks?

andy-blum: This would have the side-benefit of our magic number being less magical

packages/web-components/src/component-mixins/media-query/media-query.ts
```diff
@@ -0,0 +1,49 @@
...
+  abstract class MediaQueryMixinImpl extends Base {
+    /**
+     * A MediaQueryList object that indicates when we've crossed over the
+     * breakpoint threshold.
+     *
+     * Magic Number: The default, 1056px, matches 'carbon--breakpoint(lg)' mixin.
+     *
+     * @see https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList
+     */
+    _mediaQuery: MediaQueryList = window.matchMedia(`(max-width: 1056px)`);
```
andy-blum, line 27: We should get this breakpoint from the @carbon/layout package

packages/web-components/src/components/tabs-extended/tab.ts
```diff
@@ -72,21 +90,56 @@ class DDSTab extends StableSelectorMixin(LitElement) {
...
+            class: `${prefix}--accordion__arrow`,
+          })}
+          <div class="${prefix}--accordion__title">${label}</div>
+        </button>
+        <div id="pane-${index}" class="${prefix}--accordion__content">
+          <slot></slot>
+        </div>
+      </li>
+    `;
+  }
```
andy-blum, line 119: This is a great refactor - I hadn't thought about making the tab responsible for the accordion functionality. 👏

packages/web-components/src/components/tabs-extended/tab.ts
```diff
@@ -21,7 +26,20 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
...
+   * Whether the we're viewing smaller or larger window.
+   */
+  _isMobileVersion = this._mediaQuery.matches;
+
+  /**
+   * @inheritdoc
+   */
+  mediaQueryCallback() {
+    this._isMobileVersion = this._mediaQuery.matches;
+    this.requestUpdate();
```
andy-blum, line 40: Let's use the `@state` decorator on `this._isMobileVersion` to avoid having to manually kick off the render update.

packages/web-components/src/components/tabs-extended/tabs-extended.ts
```diff
@@ -28,7 +27,19 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
...
+   * Whether the we're viewing smaller or larger window.
+   */
+  _isMobileVersion = this._mediaQuery.matches;
+
+  /**
+   * @inheritdoc
+   */
+  mediaQueryCallback() {
+    this._isMobileVersion = this._mediaQuery.matches;
+    this.requestUpdate();
```
andy-blum, line 41: Use the `@state` decorator again

packages/web-components/src/components/tabs-extended/tabs-extended.ts
```diff
@@ -163,66 +174,45 @@ class DDSTabsExtended extends StableSelectorMixin(LitElement) {
...
-      if (navText!.scrollHeight > navText!.clientHeight) {
-        const label = (tab as DDSTab).getAttribute('label');
-        if (label) {
-          navLink!.setAttribute('aria-label', label);
-          navLink!.setAttribute('hasTooltip', label);
+
+      // Attach accordion behavior on mobile.
+      if (_isMobileVersion) {
+        tab.addEventListener('click', (e) => this._handleClick(index, e));
+      }
```
andy-blum, line 188: We should wrap the update stuff here to only happen if the value of `this._isMobileVersion` has changed to minimize blocking the main thread

andy-blum (changes requested):
There's some really great work happening here that will benefit future component optimizations. A few changes are requested.

packages/web-components/src/component-mixins/media-query/media-query.ts
andy-blum: See [the table-of-contents component](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/3c07ce0ce07063bc46b07bae36d287002713a93c/packages/web-components/src/components/table-of-contents/table-of-contents.ts#L42) for an example of this

packages/web-components/src/component-mixins/media-query/media-query.ts
jkaeser: I took a stab at making this work for any Carbon breakpoint, but I'm not 100% sold on my approach. Let me know what you think (and don't hold any punches 😄).

packages/web-components/src/component-mixins/media-query/media-query.ts
```diff
@@ -8,34 +8,49 @@
...
-        'change',
-        this.mediaQueryCallback.bind(this)
-      );
+      Object.keys(breakpoints).forEach((bp) => {
+        const funcName = `mediaQueryCallback${bp.toUpperCase()}`;
+        if (typeof this[funcName] === 'function') {
+          this._mediaQueries[bp].addEventListener(
+            'change',
+            this[funcName].bind(this)
+          );
```
andy-blum, line 51: This works to call each function, but it would be good to add each function to this class as an optional method. For example:

```ts
abstract class MediaQueryMixinImpl extends Base {

  mediaQueryCallbackLG?(event: MediaQueryListEvent): void;

}
```

This helps the IDE with function definition & typing in the components using this mixin, without defining it unnecessarily.

andy-blum (changes requested):

packages/web-components/src/component-mixins/media-query/media-query.ts
andy-blum: I like the approach, but I think we need to do a few things - outlined in lower change requests

packages/web-components/src/component-mixins/media-query/media-query.ts
```diff
@@ -8,34 +8,49 @@
...
 
 import { Constructor } from '../../globals/defs';
+import { breakpoints } from '@carbon/layout';
+
+const queriesByBreakpoint: { [index: string]: MediaQueryList } = {};
+for (let [key, val] of Object.entries(breakpoints)) {
+  queriesByBreakpoint[key] = window.matchMedia(
+    `(max-width: ${(val as any).width})`
+  );
+}
```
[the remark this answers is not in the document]
andy-blum: Each MediaQueryList is on the order of 0.2kb, but I like the idea of generating only the ones needed.

Screenshot 2023-07-13 at 10 32 06 PM

jkaeser:
Analysis of the node count reduction on our Story's demo content based on the latest code:

## Tabs Extended 
### Desktop
| - | Before | After | Reduction |
|---|---|---|---|
| nodes | 1249 | 821 | 34%
| comments | 149 | 82 | 45%
| inters | 481 | 254 | 47%

### Mobile
| - | Before | After | Reduction |
|---|---|---|---|
| nodes | 1249 | 818 | 35%
| comments | 149 | 81 | 46%
| inters | 481 | 249 | 48%

## Tabs Extended (with Media)
### Desktop
| - | Before | After | Reduction |
|---|---|---|---|
| nodes | 2072 | 1265 | 39%
| comments | 287 | 154 | 46%
| inters | 867 | 468 | 46%

### Mobile
| - | Before | After | Reduction |
|---|---|---|---|
| nodes | 2072 | 1261 | 39%
| comments | 287 | 153 | 47%
| inters | 867 | 463 | 47%

packages/web-components/src/components/tabs-extended/tabs-extended.ts
```diff
@@ -162,67 +176,48 @@ class DDSTabsExtended extends StableSelectorMixin(LitElement) {
...
-      if (navText!.scrollHeight > navText!.clientHeight) {
-        const label = (tab as DDSTab).getAttribute('label');
-        if (label) {
-          navLink!.setAttribute('aria-label', label);
-          navLink!.setAttribute('hasTooltip', label);
+
+      if (_changedProperties.has('_isMobileVersion')) {
+        // Attach accordion behavior on mobile.
+        if (_isMobileVersion) {
+          tab.addEventListener('click', (e) => this._handleClick(index, e));
```
m4olivei, line 191: It's a small thing, and admittedly an edge case, but I'm seeing that there is potential for this event handler to be applied multiple times if you cross back and forth over the breakpoint many times.

m4olivei (approved):
Looks good to me. I have a couple of nits, but nothing blocking. Nice work!

packages/web-components/src/components/tabs-extended/tabs-extended.ts
```diff
@@ -162,67 +176,48 @@ class DDSTabsExtended extends StableSelectorMixin(LitElement) {
     return tabItems;
   }
 
-  updated() {
+  updated(_changedProperties) {
```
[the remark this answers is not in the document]
andy-blum: I think the leading `_` here is _required_ by typescript if the property isn't used.

packages/web-components/src/components/tabs-extended/tabs-extended.ts
```diff
@@ -162,67 +176,48 @@ class DDSTabsExtended extends StableSelectorMixin(LitElement) {
...
+    const { _isMobileVersion } = this;
     this._isLTR = window.getComputedStyle(this).direction === 'ltr';
     this._activeTabIndex = parseInt(this._activeTab, 10);
 
-    this._tabItems.map((tab, index) => {
+    this._tabItems.forEach((tab, index) => {
       (tab as DDSTab).selected = index === this._activeTabIndex;
       (tab as DDSTab).setIndex(index);
```
andy-blum, line 186: Is this something we really want to do _every time_ we go through the render lifecycle?

m4olivei: Related: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10650/files#r1265792886

packages/web-components/src/components/tabs-extended/tabs-extended.ts
jkaeser: Good catch! Made some changes so that we attach a named function, which the browser shouldn't re-attach on subsequent `addEventListener` calls.

For posterity, from https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener:
> Note: If a particular anonymous function is in the list of event listeners registered for a certain target, and then later in the code, an identical anonymous function is given in an addEventListener call, the second function will also be added to the list of event listeners for that target.

and

> If the function or object is already in the list of event listeners for this target, the function or object is not added a second time.

packages/web-components/src/components/tabs-extended/tab.ts
```diff
@@ -72,21 +99,56 @@ class DDSTab extends StableSelectorMixin(LitElement) {
     super.updated(_changedProperties);
   }
 
-  render() {
+  protected _renderAccordionItem() {
+    const { label, selected, disabled, _index: index } = this;
+    const classes = classMap({
+      'bx--accordion__item': true,
+      'bx--accordion__item--active': selected,
+      'bx--accordion__item--disabled': disabled,
```
kennylam, line 107:
```suggestion
      `${prefix}--accordion__item`: true,
      `${prefix}--accordion__item--active`: selected,
      `${prefix}--accordion__item--disabled`: disabled,
```

kennylam (approved):
Not much to add since the rest of the reviewers were so thorough. This is really great!

packages/web-components/src/components/tabs-extended/tab.ts
jkaeser: Good catch!

packages/web-components/src/components/tabs-extended/tabs-extended.ts
```diff
@@ -28,7 +30,19 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
  * @element dds-tabs-extended
  */
 @customElement(`${ddsPrefix}-tabs-extended`)
-class DDSTabsExtended extends StableSelectorMixin(LitElement) {
+class DDSTabsExtended extends MediaQueryMixin(StableSelectorMixin(LitElement), {
+  [MQBreakpoints.LG]: MQDirs.MAX,
+}) {
+  /**
+   * Whether the we're viewing smaller or larger window.
```
ariellalgilmore, line 37:
```suggestion
   * Whether we're viewing smaller or larger window.
```

packages/web-components/src/components/tabs-extended/tab.ts
```diff
@@ -21,7 +29,19 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
  * @element dds-tab
  */
 @customElement(`${ddsPrefix}-tab`)
-class DDSTab extends StableSelectorMixin(LitElement) {
+class DDSTab extends MediaQueryMixin(StableSelectorMixin(LitElement), {
+  [MQBreakpoints.LG]: MQDirs.MAX,
+}) {
+  /**
+   * Whether the we're viewing smaller or larger window.
```
ariellalgilmore, line 36:
```suggestion
   * Whether we're viewing smaller or larger window.
```

ariellalgilmore (approved):
LGTM! just small edits

IgnacioBecerra (approved):
LGTM, thanks @pjudge!

andy-blum (approved):

PR #10651, opened 2023-07-10
fix(e2e): clear failure files before running tests

IgnacioBecerra:
### Related Ticket(s)
None.

### Description
This PR should address the weird case where the parallel e2e tests fail even though no other e2e runners failed. 
Essentially just clears the failure file artifacts before running the tests, in case there is a previous failure file.

### Changelog

**New**

- clears failure artifacts before running e2e tests


kennylam (approved):
LGTM!

Issue #10599, opened 2023-06-26
[cwc-v2] `overflow-menu` accessibility issue


abdonrd:
### Description

With the `@carbon/web-components@1`:

<img width="876" alt="Screenshot 2023-06-24 at 7 24 23 PM" src="https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/1007051/b43d6d6f-21a6-4fbb-ad90-b466606eb4c2">

And no Lighthouse errors.

---

With the `@carbon/web-components@2`:

<img width="879" alt="Screenshot 2023-06-24 at 7 24 32 PM" src="https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/1007051/a4bc7c5e-baf3-4c3e-8f65-1f49e6e28db0">
<img width="742" alt="Screenshot 2023-06-24 at 7 25 00 PM" src="https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/1007051/d69ab1a8-f5cc-4f21-a916-65d7a663469e">
<img width="751" alt="Screenshot 2023-06-24 at 7 25 06 PM" src="https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/1007051/62f3c8c8-c46c-4ce5-85ea-2f9ec02ea924">

### Component(s) impacted

overflow-menu

### Carbon for IBM.com version

v2 beta

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

-

### Package

@carbon/web-components

### CodeSandbox example

-

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

abdonrd:
Closed by https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10652

PR #10652, opened 2023-07-10
fix(buttons): added missing aria-labels for tooltip-buttons

ariellalgilmore:
### Related Ticket(s)

#10599 
### Description

The overflow-menu wasn't passing the accessibility checker because of a missing label. Digging deeper most buttons with tooltips were not passing the accessibility checker for the same reason

Updated the following components:
- overflow-menu
- button (icon variation)
- icon-button
- code-snippet
- copy
- copy-button
- tooltip
### Changelog

**New**

- added an aria-label where needed to support the accessibility checker


IgnacioBecerra (approved):
LGTM, thanks @ariellalgilmore !

Issue #10607, opened 2023-06-27
[Web components]: update all component styles to remove FOUC


kennylam:
Currently all web components are prone to FOUC (flash of unstyled content). Until now user would need to address this on the application side. The issue is many users are unaware of the fix, and our documentation is not very well exposed. The solution is to add the fix to all component styles, e.g.

```css
dds-background-media:not(:defined) {
  display: none;
}
```

andy-blum:
This won't actually work, though, will it?

If the styles are encapsulated within the component, they won't be effective until the component has loaded, at which point these styles no longer apply.

PR #10678, opened 2023-07-13
fix(components): add styles to remove fouc

ariellalgilmore:
### Related Ticket(s)

#10607 

### Description

Users should no longer have to add there no styles to prevent fouc with these changes
### Changelog

**New**

- added to the stable-selector a style tag which gets appeneded to the head and used the component tag to display none if not defined


packages/web-components/src/globals/mixins/stable-selector.ts
```diff
@@ -44,6 +44,13 @@ const StableSelectorMixin = <T extends Constructor<HTMLElement>>(Base: T) => {
         .constructor as typeof StableSelectorMixinImpl;
       if (stableSelector) {
         this.dataset.autoid = stableSelector;
+
+        var css = `${this.tagName.toLowerCase()}:not(:defined) { display: none; }`,
+          head = document.head || document.getElementsByTagName('head')[0],
+          style = document.createElement('style');
+
+        head.appendChild(style);
```
kennylam, line 52: Should it check if `<style>` already exists, and if so just append to that? Otherwise there could be lots of lots of style tags in `<head>`

ariellalgilmore: updated!

IgnacioBecerra (approved):
LGTM, glad to see our solution ended up working out! All hail the Stable Selector mixin 🙇

packages/web-components/src/globals/mixins/stable-selector.ts
```diff
@@ -44,6 +44,13 @@ const StableSelectorMixin = <T extends Constructor<HTMLElement>>(Base: T) => {
         .constructor as typeof StableSelectorMixinImpl;
       if (stableSelector) {
         this.dataset.autoid = stableSelector;
+
+        const css = `${this.tagName.toLowerCase()}:not(:defined) { display: none; }`;
+        const head = document.head || document.getElementsByTagName('head')[0];
+        const style = head.querySelector("style") || document.createElement('style');
```
kennylam, line 50:
```suggestion
        const style = head.querySelector('style') || document.createElement('style');
```

Super small thing but after that lgtm!

kennylam: Looks like you need to run a `yarn format` too.

kennylam (approved):
LGTM!

PR #10699, opened 2023-07-17
docs(migration): add migration v2 template

annawen1:
### Description

Add migration template for v2 so we can fill out as we make changes to the components.

### Changelog

**New**

- migration doc


kennylam (approved):
LGTM!

Issue #10682, opened 2023-07-14
[Masthead authentication UI]: Setting the authentication user status programatically causes UI issue


john-maddock:
### Description

 I'm trying in the masthead component to programatically set the authenticated user status. but when the page loads i see that the circle icon appears but then disappears... 
I've included a stackblitz proj with it in action
https://stackblitz.com/edit/github-eezkpg-mb8c4z?file=index.html

refresh the page to see the issue

### Component(s) impacted

dds-masthead-container

### Browser

Chrome

### Carbon for IBM.com version

latest

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

MY IBM

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://stackblitz.com/edit/github-eezkpg-mb8c4z?file=index.html

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #10700, opened 2023-07-17
fix(masthead): firstupdated user-status is authenticated

ariellalgilmore:
### Related Ticket(s)

#10682 
### Description

If a user tried to set user-status to authenticated it would flash the dot as signed in and then it would disappear.  
### Changelog

**Changed**

- updated `firstUpdated` so that it would only call `loadUserStatus` if `user-status` wasn't already set by the adopter


kennylam (approved):
LGTM!

PR #10702, opened 2023-07-18
fix(ui-shell): add missed header-side-nav-items import

abdonrd:
### Related Ticket(s)

None

### Description

Add missed header-side-nav-items import.
Also sort the imports.

### Changelog

**New**

- fix(ui-shell): add missed header-side-nav-items import


annawen1 (approved):
Thanks for catching that!

Issue #10691, opened 2023-07-15
[cwc-v2] Wrong button kind=ghost icon position


abdonrd:
### Description

I take the screenshots with the ghost button with the hover state to show the error more clearly.

Current:
<img width="275" alt="Screenshot 2023-07-14 at 11 26 21 PM" src="https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/1007051/b26a0c38-d868-49ea-a1e1-3e3ec73f663a">

Expected:
(we also hope to see the icon well aligned with the text, but I have not been able to solve it quickly)
<img width="269" alt="Screenshot 2023-07-14 at 11 28 16 PM" src="https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/1007051/cbdd0b63-b257-4d63-a35d-6d7fc96f9714">

### Component(s) impacted

Button

### Carbon for IBM.com version

v2 beta

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

-

### Package

@carbon/web-components

### CodeSandbox example

https://stackblitz.com/edit/vitejs-vite-hno5li?file=package.json,index.html

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

abdonrd:
Fixed by https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10708

PR #10708, opened 2023-07-18
fix(ghost-button): icon alignment for larger sizes

ariellalgilmore:
### Related Ticket(s)

#10691 
### Description

ghost button with icon wasn't aligning correctly 

Showing fixed button with hover state in 2xl size
Screenshot 2023-07-18 at 1 15 25 PM

### Changelog

**New**

- {{new thing}}

**Changed**

- added styles to temporarly resolve this, but eventually should be able to remove once we can upgrade `@carbon/styles`

**Removed**

- {{removed thing}}

### DIFF
--- a/packages/carbon-web-components/src/components/button/button.scss
+++ b/packages/carbon-web-components/src/components/button/button.scss
@@ -110,3 +110,10 @@ $css--plex: true !default;
 :host(#{$prefix}-button-set[stacked]) {
   @extend .#{$prefix}--btn-set--stacked;
 }
+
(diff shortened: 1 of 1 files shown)


abdonrd:
@ariellalgilmore you are right, thanks for the example!

When I looked the Carbon documentation I think it should be attached to the text:
https://carbondesignsystem.com/components/button/usage/#anatomy

ariellalgilmore:
@abdonrd np! Yeah the example is a little confusing since it doesn't show for the larger button types. Thanks for opening the issue 👍

annawen1 (approved):
LGTM!

PR #10729, opened 2023-07-20
fix(fouc): add to documentation

ariellalgilmore:
### Related Ticket(s)

#10607 

### Description

add documentation about FOUC so adopters know to include it within their applications

### Changelog

**New**

- documentation

**Removed**

- FOUC script that did not work


annawen1 (approved):
LGTM

Issue #10657, opened 2023-07-11
[Locale modal]: optimize component to lower DOM node count


kennylam:
[JIRA](https://jsw.ibm.com/browse/ADCMS-3579)

Currently this component has a DOM node count which can be lowered. Solutions to consider are removing duplicate/unnecessary nodes and lazy loading content.

Node audit

| Nodes| Comments | Inters |
|--------|--------|--------|
| 2734 | 770 | 1155 |

ljcarot:
@kennylam might be good time to redesign the locale modal

andy-blum:
@kennylam @ljcarot I think there's work on this underway from @areagan1030

areagan1030:
@kennylam @ljcarot @andy-blum Yes, we do have some work in the pipeline being guided by Joseph Dombroski. Putting a push on getting this to a good spot, so we can set up some time to share with Carbon team and keep this moving forward.

PR #10731, opened 2023-07-21
feat(footer): optimize node counts

### LINKED ISSUE #10656: [Footer]: optimize component to lower DOM node count
[JIRA](https://jsw.ibm.com/browse/ADCMS-3578)

Currently this component has a DOM node count which can be lowered. Solutions to consider are removing duplicate/unnecessary nodes and lazy loading content in hidden tabs. This component utilizes the `accordion` component in mobile breakpoints, which does not need to be rendered in desktop views (and vice versa). A fix has already been implemented for another component [here](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10650).

**Note**: the first 3 variations in the table below also contain the `locale-modal`, which also contributes to the node count.

| Component | Nodes| Comments | Inters |
|--------|--------|--------|--------|
| Footer | 1228 | 367 | 549 |
| Footer short | 1228 | 367 | 549 |
| Footer micro | 1229 | 367 | 550 |
| Footer - language only | 2476 | 711 | 1071 |
| Footer short - language only | 2476 | 711 | 1071 |
| Footer micro - language only | 2477 | 711 | 1072 |

jkaeser:
### Related Ticket(s)

Resolves #10656 

### Description

Reduces the usage of hidden DOM elements by replacing them with conditionally rendered markup based on the viewport width.

### Analysis
Fortunately, the `<dds-footer>` family of components was pretty well optimized aside from the locale modal (which is being handled in #10657). The `<dds-footer-nav-group>` switches between accordion and list displays for mobile and desktop, respectively, but it was already leveraging browser APIs to render just one version of the markup at a time. I did switch this to use our new `MediaQueryMixin` for consistency's sake and reducing the amount of code necessary for the component to function.

The only opportunities for actual node reductions I found were in the `<dds-footer-composite>`'s rendering of language selectors, which appear in Stories such as **[Default language only](https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/10731/index.html?path=/story/components-footer--default-language-only)**.

#### Mobile
| - | Before | After | % Reduction |
|-|-|-|-|
| Nodes | 2521 | 1840 | 27% |
| Comments | 726 | 528 | 27% |
| Inters | 1086 | 802 | 26% |

#### Desktop
| - | Before | After | % Reduction |
|-|-|-|-|
| Nodes | 2476 | 1891 | 24% |
| Comments | 711 | 560 | 21% |
| Inters | 1071 | 827 | 23% |

### Changelog

**Changed**

- Use `MediaQueryMixin` for `DDSFooterNavGroup` headers' conditional rendering as accordion label or static heading.
- Conditionally render `DDSFooterComposite`'s language selector based on viewport size.
- Make mock `MediaQueryList` reusable and update related unit tests.


packages/web-components/src/components/footer/footer-nav-group.ts
```diff
@@ -104,37 +99,17 @@ class DDSFooterNavGroup extends StableSelectorMixin(LitElement) {
       this.setAttribute('role', 'listitem');
     }
     super.connectedCallback();
-    if (this._hChangeMediaQuery) {
-      this._hChangeMediaQuery = this._hChangeMediaQuery.release();
-    }
-    const { mediaStickExpanded } = this.constructor as typeof DDSFooterNavGroup;
-    const mediaQueryList =
```
m4olivei, line 111: Nice cleanup win for the `MediaQueryMixin` 🙌

m4olivei (changes requested):
Looking code. I think I see a bug in the code where right on 1056px, the `lg` breakpoint, the language selector is not displaying at all:

Monosnap screencast 2023-07-25 11-20-00

packages/web-components/src/components/footer/footer.scss
```diff
@@ -397,7 +397,7 @@
     margin-top: -#{$carbon--spacing-05};
   }
 
-  @include carbon--breakpoint('lg') {
+  @include carbon--breakpoint-down('lg') {
```
m4olivei, line 400: Hum, this isn't quite it. I'm now seeing no language selector display for anything under the `lg` breakpoint.

The other issue is that `carbon--breakpoint-down` produces a media query like `(max-width: 65.98rem)`, which leaves out the border pixel. I think either you change the CSS to never include `display: none` b/c the precence of either tag is now controlled in Javascript. Or you have `:host(#{$dds-prefix}-language-selector-desktop)` default to `display: none;` with a media query using `carbon--breakpoint('lg')` to set it to `display: block`. Then do the opposite for `:host(#{$dds-prefix}-language-selector-mobile)`.

m4olivei (changes requested):

packages/web-components/src/components/footer/footer.scss
jkaeser: Oof, sorry @m4olivei. That was sloppy on my end. I removed the CSS responsible for hiding these elements, instead relying on the JS to do that work. I promise I double-checked this time, too 😅

m4olivei: All good! It's looking good to me now. Thanks.

m4olivei (approved):
Works well from my checks. Thanks John.

andy-blum (approved):
Great work on this - we'll see even better improvements once we update the locale switcher to the new designs.

ariellalgilmore:
looking great!

one thing i did notice was the language only footer dropdown has a height of 40px vs 48px:

deploy-preview:
Screenshot 2023-08-08 at 1 17 44 PM

latest:
Screenshot 2023-08-08 at 1 18 34 PM

IgnacioBecerra (approved):
LGTM, thanks @jkaeser!

jkaeser:
Nice catch, @ariellalgilmore! Should be fixed now.

packages/web-components/src/components/footer/footer-composite.ts
```diff
@@ -242,20 +257,81 @@ class DDSFooterComposite extends ModalRenderMixin(
...
+      _isMobile: isMobile,
       langList,
       languageSelectorLabel,
       selectedLanguage,
       size,
+    } = this;
+    return isMobile
+      ? html`
+          <dds-language-selector-mobile
+            size="${size || DROPDOWN_SIZE.EXTRA_LARGE}"
```
ariellalgilmore, line 272: Thanks @jkaeser! I think the size prop is actually unessary here, since it is already being set in the component: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/fcaa459c82b197b760e589c1378e4b3c1041058b/packages/web-components/src/components/footer/language-selector-mobile.ts#L34

```suggestion
```

packages/web-components/src/components/footer/footer-composite.ts
```diff
@@ -242,20 +257,81 @@ class DDSFooterComposite extends ModalRenderMixin(
...
+                  lang="${ifNonNull(language.id)}"
+                  >${ifNonNull(language.text)}</bx-select-item
+                >
+              `
+            )}
+          </dds-language-selector-mobile>
+        `
+      : html`
+          <dds-language-selector-desktop
+            size="${size || DROPDOWN_SIZE.EXTRA_LARGE}"
```
ariellalgilmore, line 290: same here: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/fcaa459c82b197b760e589c1378e4b3c1041058b/packages/web-components/src/components/footer/language-selector-desktop.ts#L131
```suggestion
```

packages/web-components/src/components/footer/footer-composite.ts
```diff
@@ -242,20 +257,81 @@ class DDSFooterComposite extends ModalRenderMixin(
...
     const {
-      buttonLabel,
       clearSelectionLabel,
-      disableLocaleButton,
-      langDisplay,
+      _isMobile: isMobile,
       langList,
       languageSelectorLabel,
       selectedLanguage,
       size,
```
ariellalgilmore, line 267:
```suggestion
```

packages/web-components/src/components/footer/footer-composite.ts
jkaeser: Ah, there's a discrepancy: the language selector components would make it seem like they should always render with the XL size, but that's not actually true. The micro version of the footer uses a sort of hacked "micro" size dropdown. I've pushed changes to hopefully make this a little more clear.

PR #10781, opened 2023-08-02
fix: update the carbon-element decorator

abdonrd:
### Description

With the current decorator, the components can't be imported into the server because of the `window.` reference.
There was also a lot of code that wasn't being used.

Original Lit decorator:
https://github.com/lit/lit/blob/main/packages/reactive-element/src/decorators/custom-element.ts

### Changelog

**Changed**

- fix: update the carbon-element decorator

### DIFF
--- a/packages/carbon-web-components/src/globals/decorators/carbon-element.ts
+++ b/packages/carbon-web-components/src/globals/decorators/carbon-element.ts
@@ -1,69 +1,28 @@
 /**
  * @license
- *
- * Copyright IBM Corp. 2019, 2023
- *
- * This source code is licensed under the Apache-2.0 license found in the
- * LICENSE file in the root directory of this source tree.
- */
-
-/**
- * @license
- * Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
- * This code may only be used under the BSD style license found at
- * http://polymer.github.io/LICENSE.txt
- * The complete set of authors may be found at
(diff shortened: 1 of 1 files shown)


packages/carbon-web-components/src/globals/decorators/carbon-element.ts
```diff
@@ -1,69 +1,28 @@
...
+type CustomElementClass = Omit<typeof HTMLElement, 'new'>;
 
-const legacyCustomElement = (
-  tagName: string,
-  clazz: Constructor<HTMLElement>
-) => {
+const legacyCustomElement = (tagName: string, clazz: CustomElementClass) => {
   try {
-    window.customElements.define(tagName, clazz);
+    customElements.define(tagName, clazz as CustomElementConstructor);
```
abdonrd, line 25: The important change.

packages/carbon-web-components/src/globals/decorators/carbon-element.ts
```diff
@@ -87,7 +46,7 @@ const standardCustomElement = (
     // This callback is called once the class is otherwise fully defined
     finisher(clazz: Constructor<HTMLElement>) {
       try {
-        window.customElements.define(tagName, clazz);
+        customElements.define(tagName, clazz);
```
abdonrd, line 49: The important change.

kennylam (approved):
LGTM thanks @abdonrd!

PR #10800, opened 2023-08-07
Feat(global-styles): Create mechanism for global resets & primitives to be shared between components

### LINKED ISSUE #10794: Create a mechanism/utility for components to request & adopt global/primitive styles

andy-blum:
### Related Ticket(s)

Closes #10794 

### Description

This PR creates two new lower-level concepts, which when combined allow components to adopt one or more "global" stylesheets.

Consider the wide variety of components that contain button styles. Instead of re-defining those button styles in each component's own internal stylesheets, we can define those styles in one place and have all components adopt that single stylesheet.

This should reduce the filesize (and therefore bandwidth used/blocking time) of every single component by stripping out large chunks of the CSS. Additionally, since each component class defines it's own stylesheet, moving those duplicated styles to a single stylesheet that can be adopted many times frees up large chunks of system memory.

To prove the efficacy of this approach, I've also implemented this setup into the `button-group` and `button-group-item` components. To further reduce each component's definition size I've split the suite of styles (`button-group.scss`) into individual stylesheets for each component. (`button-group.scss` and `button-group-item.scss`).

#### CarbonBase class

The `CarbonBase` class is intended to exist between the LitElement framework and all carbon components. As of this PR, its only job is to handle the global style adoption, but it could also take on the role of the StableSelector mixin or act as another entry point to all components.

#### CarbonStyles class

The `CarbonStyles` class acts as a global storage point for all the CSS primitives. Each stylesheet can be requested by name using the `getStyleSheets()` method, which will return an array of `CSSStyleSheet` objects.

#### How it all ties together
1. **Styles** package generates Sass primitives.
2. **Utilities** package imports primitives in `CarbonStyles` class, which converts the imported CSS strings into `CSSStyleSheet` objects, and stores them on the class instance keyed by name (e.g. "reset").
3. **Carbon Web Components** package defines `CarbonBase` class that extends `LitElement`. It provides mechanisms for extending elements to attach the `CSSStyleSheet` objects to the shadow root's `adoptedStylesheets`.
   1. `_requestCarbonStyles()` method defines which primitives should be adopted, by default requesting reset and type primitives. This method can be overridden by extending classes in the event they need to request additional primitives (e.g. button styles) or they don't need one or more of the default primitives.
   2. The requested primitives are adopted in the `connectedCallback()` lifecycle method.

### Changelog

**New**

- Add `CarbonBase` and `CarbonStyles` classes

**Changed**

- split `button-group.scss` files into individual component stylesheets


packages/carbon-web-components/src/globals/CarbonBase.ts
```diff
@@ -0,0 +1,42 @@
+/**
+ * Copyright IBM Corp. 2016, 2023
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { LitElement } from 'lit';
+import CarbonStyles from '../../../utilities/src/utilities/carbonStyles/carbonStyles.js';
```
IgnacioBecerra, line 9:
```suggestion
import carbonStyles from '../../internal/vendor/@carbon/ibmdotcom-utilities/utilities/carbonStyles/carbonStyles.js';
```

packages/web-components/src/components/button-group/button-group.ts
```diff
@@ -13,6 +13,7 @@ import settings from '@carbon/ibmdotcom-utilities/es/utilities/settings/settings
 import styles from './button-group.scss';
 import StableSelectorMixin from '../../globals/mixins/stable-selector';
 import { carbonElement as customElement } from '../../internal/vendor/@carbon/web-components/globals/decorators/carbon-element.js';
+import CarbonBase from '../../../../carbon-web-components/src/globals/CarbonBase';
```
IgnacioBecerra, line 16:
```suggestion
import CarbonBase from '../../internal/vendor/@carbon/web-components/globals/CarbonBase';
```

packages/web-components/src/components/button/button.ts
```diff
@@ -17,6 +17,7 @@ import settings from '../../internal/vendor/@carbon/ibmdotcom-utilities/utilitie
 import styles from './button.scss';
 import StableSelectorMixin from '../../globals/mixins/stable-selector';
 import { carbonElement as customElement } from '../../internal/vendor/@carbon/web-components/globals/decorators/carbon-element.js';
+import CarbonBase from '../../../../carbon-web-components/src/globals/CarbonBase';
```
IgnacioBecerra, line 20:
```suggestion
import CarbonBase from '../../internal/vendor/@carbon/web-components/globals/CarbonBase';
```

packages/web-components/src/components/button-group/button-group.ts
```diff
@@ -22,7 +23,9 @@ const { stablePrefix: ddsPrefix } = settings;
  * @element dds-button-group
  */
 @customElement(`${ddsPrefix}-button-group`)
-class DDSButtonGroup extends StableSelectorMixin(LitElement) {
+class DDSButtonGroup extends StableSelectorMixin(CarbonBase) {
+  carbonStyles = ['resetStyles', 'typeStyles'];
```
IgnacioBecerra, line 27: If every component uses `resetStyles` and `typeStyles`, would it make sense to omit them here and always send them in the `carbonStyle` utility anyway?

IgnacioBecerra (commented):
This is looking to be a great solution! I think we should try sticking with the internal vendor path for imports though. 

One "concern" I have is that we're gonna have to create `*.scss` files under the `utilities/carbonStyles` directory for every component, correct? If so, we might end up with a bunch of small files there, which while it is still better than the current style implementation, might be a lot of "busywork", and we'd need to maintain it if we remove/add components. 

Maybe it might be worth considering a solution that could automate this process on build time, or even better, only import what we need at a given time?

To do during build, maybe iterate through the style directory and create new files based on a template (seen below) for every `.scss` file? 

For the second approach, when listening to the request event, would we be able to dynamically create that `.scss` file with a template based on the component's tag name or data-autoid, and then dynamically import it? 

This could be `carbonStyles/button.scss`
``` carbonStyles/button.scss
// license text here
@use '@carbon/styles/scss/components/${componentName}' as *;
@include ${componentName};
```

After importing it, we could send that stylesheet, plus other ones that were requested as well. 

Not sure how feasible these suggestions could be, but I think it could be worth looking into/refining them, just to further optimize this nice solution :D 

Let me know what you think!

andy-blum:
> One "concern" I have is that we're gonna have to create *.scss files under the utilities/carbonStyles directory for every component, correct? If so, we might end up with a bunch of small files there, which while it is still better than the current style implementation, might be a lot of "busywork", and we'd need to maintain it if we remove/add components.
> 
> Maybe it might be worth considering a solution that could automate this process on build time, or even better, only import what we need at a given time?

Yeah, my thought was that eventually the global styles would get moved over to [here](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/tree/main/packages/styles/scss/globals) with the other scss globals. This was just easier to prototype - same reason for all the relative URLs and avoiding the internal/vendor paths.

> For the second approach, when listening to the request event, would we be able to dynamically create that .scss file with a template based on the component's tag name or data-autoid, and then dynamically import it? We could also just save the new sheet in a variable and add it to a map, sorta like what you did, but only adding things when they are really needed, instead of all the files at a time.

I had a similar thought but my concern here is that you'd have a lot of potential for FOUC if you're on a slower connection and you have to wait for each of the following:
1. Wait for the document to fetch/load
2. Wait for the JS component index files to fetch/load
3. Wait for the component submodules to fetch/load
4. Wait for the CarbonStyles object to fetch/load
5. Wait for all the necessary stylesheets to fetch/load

That's just a _long_ chain of initators between the document and your _global_ styles.

andy-blum:
Latest changes include changes to fix the build steps of the monorepo and include an additional unaffiliated component in both @carbon/web-components and @carbon/ibmdotcom-web-components to force the rollup process to split the CarbonBase & CarbonStyles classes to be spun off into their own files.

Running a local server environment and comparing C4IBM v1's dds-button-group suite to v2's c4d-button-group suite yields the below results:

Screenshot 2023-10-06 at 3 27 12 PM

The v1 button-group-item component is ~130kb, most of which is the CSS style rules string. This component is nearly half the weight of all downloaded resources.

Screenshot 2023-10-06 at 3 26 44 PM

This v2 button-group-item component is ~13kb, approximately 90% lighter than its v1 counterpart. While the sum of all resources has grown some, I think this is due to some differences in v1 and v2 as well as the increased overhead of the CarbonBase class. I think this growth will be drastically outweighed by the savings of the additional components added in the future.

Below is a .zip archive of the CDN links built with `yarn build:components` if anyone would like to try this for themselves.
[cdn-links.zip](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/files/12834907/cdn-links.zip)

packages/carbon-web-components/src/globals/CarbonBase.ts
```diff
@@ -0,0 +1,30 @@
+/**
+ * Copyright IBM Corp. 2016, 2023
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+// @ts-nocheck
+import { LitElement } from 'lit';
+import CarbonStyles from '../internal/vendor/@carbon/ibmdotcom-utilities/utilities/CarbonStyles/CarbonStyles';
```
m4olivei, line 10: Something isn't right about this import and it's causing `yarn start` within `packages/storybook` to fail.

I think it should look closer to the following the way the index.js / folder structures are, this is working for me:

I've been running `yarn build` at the project root, then running `yarn start` within `src/packages` to test. The following clears me of errors.

```suggestion
import { CarbonStyles } from '../internal/vendor/@carbon/ibmdotcom-utilities/utilities/carbonStyles/carbonStyles.js';
```

packages/carbon-web-components/src/globals/CarbonBase.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export default class CarbonBase extends LitElement {
+  carbonStyles?: String[];
```
m4olivei, line 13: It would be neat if we could type this more narrowly, so that when you consume it, you get auto-complete on the ones presently available. Not totally sure, but somewhere in the utilities package it would export a type definition, that we could consume here.

From `carbonStyles.js`

```
const importedStyleSheets = {
  buttonStyles,
  resetStyles,
  typeStyles,
};

type ImportedStyleSheets = keyof importedStyleSheets;
```

Now this line would become:

```
carbonStyles?: ImportedStyleSheets[];
```

Naming sucks, but hoepfully you get the idea?

Definately in the category of "Bonus"

m4olivei (changes requested):
Initial static review pass. Not totally through yet, and still testing locally, but this is looking great so far Andy. I'll also admit to seeing lots of this for the first time (interaction across packages, `internal` imports, etc.), so I'm afraid I'm not super helpful as a reviewer.

Issue #10589, opened 2023-06-22
[Filter Panel]: Consolidated accessibility issues


andy-blum:
## Description

Following a meeting to review the accessibility of the [Filter Panel](https://carbon-design-system.github.io/carbon-for-ibm-dotcom/canary/web-components/iframe.html?args=&id=components-filter-panel--default&viewMode=story), the issues below have been identified.

---

## ❓Sets of checkboxes need to be reachable using arrow keys

<img width="385" alt="Screenshot 2023-06-22 at 12 00 29 PM" src="https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/25532785/48001712-fb2c-4149-913f-947d545d3ce9">

According to Ram, the checkboxes in the filter panel need to be navigable using the arrow keys to move through the set of all options. The [ARIA Authoring Practices Guide](https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/examples/checkbox/), however, does not seem to match this assessment:

| Key   | Function                                               |
|-------|--------------------------------------------------------|
| Tab   | Moves keyboard focus to the   checkbox.                |
| Space | Toggles checkbox between checked and unchecked states. |

It seems @mbgower also recommends against arrow-key navigation in #10132 per Olivia's comment.

---

## ❌ Non-checkbox filter buttons are announced as "frames" by JAWS

I'm unable to test with JAWS and VoiceOver isn't announcing this quite the same, so I'm sort of guessing here, but I _think_ this might be related to the use of `aria-controls`.

On the top-level accordion items (Product Types, Technologies, etc), we have this markup:

<img width="1168" alt="Screenshot 2023-06-22 at 12 11 45 PM" src="https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/25532785/4a5c1886-00c0-4979-a42a-e42acde47a30">

On the second-level non-checkbox items (like those under Technologies), we have this markup:

<img width="1213" alt="Screenshot 2023-06-22 at 12 11 22 PM" src="https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/25532785/815b40ef-0208-435e-b45f-0f8004c9a9dd">

[...]

PR #10812, opened 2023-08-10
fix(filter-panel) Fixing the placement/display of the clear button | …

bruno-amorim:
…Fixing the focus handling of the mobile version's popup window

### Related Ticket(s)
JIRA: https://jsw.ibm.com/browse/ADCMS-3552

Relates to #10589

### Description

First fix was to placement/display of the clear button. As per described in the issue, it would be better if the clear button now sits at the bottom of the filter panel. 

Second fix was to fix the modal focus handling when in mobile resolutions. In mobile, the filter panel becomes a button which opens a modal. I changed it to now use the expressive-modal components  which has a built in logic that traps the focus to the modal when it's opened.
Additionally, I have also implemented some code to set the proper aria attributes to the modal.


andy-blum (changes requested):
You're on the right track here. Keep up the good work and ping me if you have questions.

packages/web-components/src/components/filter-panel/filter-panel-modal.ts
```diff
@@ -33,8 +33,21 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
  */
 @customElement(`${ddsPrefix}-filter-panel-modal`)
 class DDSFilterPanelModal extends HostListenerMixin(
-  StableSelectorMixin(BXModal)
+  StableSelectorMixin(DDSExpressiveModal)
```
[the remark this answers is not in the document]
bruno-amorim: Hi Andy, 
This is fixed and ready for review

packages/web-components/src/components/filter-panel/filter-panel.ts
```diff
@@ -73,14 +73,14 @@ class DDSFilterPanel extends HostListenerMixin(
       <section class="${prefix}--filter-panel__section">
         <div class="${prefix}--heading-clear">
           <div class="${prefix}--filter__heading">${this.heading}</div>
-          <button class="${prefix}--clear" @click=${this._handleClear}>
-            <div class="${prefix}--clear__container">
-              Clear
-              <div class="${prefix}--reset__icon">${Reset()}</div>
-            </div>
-          </button>
```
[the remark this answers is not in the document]
bruno-amorim: Hi Andy,
This is fixed and ready for review

packages/styles/scss/components/filter-panel/_filter-panel.scss
```diff
@@ -134,7 +134,8 @@
 
     &:not([has-selections]) {
       .#{$prefix}--clear {
-        display: none;
+        color: $ibm-color__gray-40;
```
andy-blum, line 137: I think we probably have a better color token to use here. Maybe @oliviaflory knows?

packages/web-components/src/components/filter-panel/filter-panel-modal.ts
```diff
@@ -26,21 +27,81 @@ import { carbonElement as customElement } from '../../internal/vendor/@carbon/we
 const { prefix } = settings;
 const { stablePrefix: ddsPrefix } = ddsSettings;
 
+const WITHIN = Node.DOCUMENT_POSITION_CONTAINED_BY;
```
andy-blum, line 30: This all looks like it's copy pasted over from somewhere else - that's okay, but I imagine there are ways we can bring this code in without duplicating. Lets schedule a time to pair up and see if we can figure out how to keep this DRY.

andy-blum (changes requested):
Fantastic work on this. A couple of minor nits to fix, and then I want to schedule some time to try to avoid duplicating code.

packages/web-components/src/components/filter-panel/filter-panel-modal.ts
```diff
@@ -7,12 +7,13 @@
...
  */
 
-import { html, property } from 'lit-element';
+import { html, property, query } from 'lit-element';
 import settings from 'carbon-components/es/globals/js/settings.js';
 import HostListenerMixin from '../../internal/vendor/@carbon/web-components/globals/mixins/host-listener.js';
 import './filter-group';
 import './filter-modal-button';
 import './filter-modal-heading';
+//import DDSExpressiveModal from '../../internal/vendor/@carbon/web-components/components/modal/modal.js';
```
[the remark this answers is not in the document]
bruno-amorim: Unnecessary comments removed

packages/styles/scss/components/filter-panel/_filter-panel.scss
m4olivei: My guess would be `$disabled-02`. Given that the background of the filter panel is Grey 10 (`$ui-01`), and there are tokens for `$disabled-n`. The guidance states that you need to use `$disabled-02` on top of `$disabled-01`, which is Grey 10 as well. 

https://v10.carbondesignsystem.com/guidelines/color/usage

image

packages/web-components/src/components/filter-panel/filter-panel-modal.ts
```diff
@@ -26,21 +26,81 @@ import { carbonElement as customElement } from '../../internal/vendor/@carbon/we
 const { prefix } = settings;
 const { stablePrefix: ddsPrefix } = ddsSettings;
 
+const WITHIN = Node.DOCUMENT_POSITION_CONTAINED_BY;
+const PRECEDING =
+  Node.DOCUMENT_POSITION_PRECEDING | Node.DOCUMENT_POSITION_CONTAINS;
+const FOLLOWING =
+  Node.DOCUMENT_POSITION_FOLLOWING | Node.DOCUMENT_POSITION_CONTAINED_BY;
+
+function tryFocusElems(
```
m4olivei, line 35: In the interest of keeping this DRY, could we make it a protected method in `DDSExressiveModal`? Then we would inherit it in `DDSFilterPanelModal` and have it handy for use without duplicating it.

packages/web-components/src/components/filter-panel/filter-panel-modal.ts
```diff
@@ -61,6 +121,80 @@ class DDSFilterPanelModal extends HostListenerMixin(
     );
   }
 
+  get focusableElements() {
```
m4olivei, line 124: Do we need to copy pasta this method here? Isn't it inherited now that we extend `DDSExpressiveModal`?

packages/web-components/src/components/filter-panel/filter-panel-modal.ts
```diff
@@ -61,6 +121,80 @@ class DDSFilterPanelModal extends HostListenerMixin(
...
+      ),
+      ...Array.from(
+        this.querySelectorAll(
+          selectorTabbableForModal
+        ) as NodeListOf<HTMLElement>
+      ),
+    ];
+  }
+
+  private get _focusablePanelModalElements() {
```
m4olivei, line 141: Instead of copy pasting this method, could we make the parent method protected or public so that we would inherit it here?

packages/web-components/src/components/filter-panel/filter-panel-modal.ts
```diff
@@ -61,6 +121,80 @@ class DDSFilterPanelModal extends HostListenerMixin(
...
+      if (comparison & FOLLOWING) {
+        return -1;
+      }
+      /* eslint-enable no-bitwise */
+
+      return 0;
+    });
+  }
+
+  private _handlePanelModalFocusIn = ({ target, relatedTarget }) => {
```
m4olivei, line 169: Same comment here, and theme throughout. Where we are copy / pasting code from the parent class b/c it's marked `private` there, could we instead open up the visibility to be protected or public so it could be inherited as is without duplication?

m4olivei (changes requested):
Apologies if my comments are repetitive. Suggesting a way to be DRY here.

packages/web-components/src/components/filter-panel/filter-panel-modal.ts
m4olivei: Sure I can take a crack at it.

m4olivei (approved):
@bruno-amorim Things are looking good to me on Storybook wrt the focus wrapping. I'm happy with the code reuse I was able to achieve for that. I'll step back and let you finish the remaining threads here, namely @andy-blum 's feedback https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10812#discussion_r1303059460. Of course feel free to reach out if you need a hand with anything. Thanks!

packages/styles/scss/components/filter-panel/_filter-panel.scss
bruno-amorim: Hi @andy-blum and @m4olivei 
I agree to use the color as per described in carbon guidelines, thought I find it a bit too light in contrast with the background, I think it works.

kennylam (approved):

PR #10838, opened 2023-08-17
fix(sticky-header): report cumulative height before first scroll

jkaeser:
### Related Ticket(s)

none

### Description

The StickyHeader utility exposes the cumulative height of all currently stuck elements as both a CSS custom property and a class property. The problem is that these are only calculated on scroll, meaning no value was available before a user interacts with the page (e.g. on initial page load). This PR fixes that by running the calculation whenever an element is set.

### Changelog

**Changed**

- Updates the exposed cumulative height value that is exposed for end-users whenever an element is registered within the StickyHeader utility.


andy-blum (approved):
Code and preview both look good. Thanks!

ariellalgilmore (approved):
LGTM! just need to point the branch to the release branch

jkaeser:
Recreated in #10843 to merge into release branch instead of `main`.

PR #10867, opened 2023-08-21
chore(masthead-v2): merge main

m4olivei:
### Related Ticket(s)

https://jsw.ibm.com/browse/ADCMS-3731

### Description

Merges latest changes from `main` into `feat/masthead-v2-dev`.


jkaeser:
> e2e test failure in the l1 tests. Can you run that test locally before we re-run the suite on Jenkins in case it's an issue with the runner infra?

This is the same failure I ran into with https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10706. I believe it only fails when running the tests headlessly.

I circumvented the failures by replacing all this test's instances of `.should('be.visible')` with the admittedly weaker success criterion of `.should('have.length', 1)`. At least I thought I did, but evidently I missed some? Not sure how tests would pass in that PR and then fail here.

m4olivei:
> This is the same failure I ran into with https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10706. I believe it only fails when running the tests headlessly.

I saw it fail once locally running in dev mode, driving my local Chrome browser. It is failing with a `.should('be.visible')` assertion: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/feat/masthead-v2-dev/packages/web-components/tests/e2e-storybook/cypress/integration/masthead/masthead-l1.e2e.js#L79

Odd it's intermittent like that. One thing I noticed is that we're chaining from a `click()` call, which [Cypress recommends not to do](https://docs.cypress.io/api/commands/click). From the docs:

> It is unsafe to chain further commands that rely on the subject after .click().

I'll try keeping the `.should('be.visible')` cut untangling the chain after `click()`, hopefully thats a viable strategy that lets us keep the better `.should('be.visible')` assertion.

m4olivei:
OK odd. Things are looking much better now, however looks like a final job is still failing on the presence of some fail file, even though the jobs that write that file all passed and should have only written the file in case of failure...

I'm going to guess this is b/c I'm just re-running failed jobs? Lets try a close and re-open...

annawen1:
> OK odd. Things are looking much better now, however looks like a final job is still failing on the presence of some fail file, even though the jobs that write that file all passed and should have only written the file in case of failure...
> 
> I'm going to guess this is b/c I'm just re-running failed jobs? Lets try a close and re-open...

I think if all those e2e-tests-parallel wc jobs pass but `wc-tests-completed` is still showing failed - that's fine, we can go ahead and merge manually next time

m4olivei:
Thanks @annawen1. Looks like, after a couple of runs of the whole suite, that's the consistent result. All `e2e-tests-parallel / *` jobs passed except `e2e-tests-parallel / wc-tests-completed`. I suppose that means we can merge this then.

@jkaeser / @andy-blum recent changes, namely in the test, https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10867/commits/616f7f636c53bd98b32ca8f1b3296b02a49f0135, looking good to you?

jkaeser (approved):

andy-blum (approved):
Deploy preview looks good, trusting the tests on the rest. #YOLO

annawen1:
> @annawen1 we're not able to do the merge because the `wc-tests-completed` one is marked as required.

Ok, I'll go ahead and merge it in manually here

PR #10873, opened 2023-08-23
feat(cta): deprecate cta and migrate features to base components

### LINKED ISSUE #10359: [CTA] web component
### User story
We will be deprecating the CTA component in v2

### Additional information
The CTA component is redundant right now, it pulls from other components (link, link with icon, card, card link, button group). 
We plan to deprecate the CTA component and direct users to use the components listed above directly.

### Tasks
- [ ] See if there is a way to automate the conversions from "CTA" to the other components
- [ ] Add deprecation notice to CTA component
- [ ] Remove from v2 storybook
- [ ] Identify breaking changes and migration guidance?

### Acceptance criteria

IgnacioBecerra:
### Related Ticket(s)
Closes #10359

### Description
This PR deprecates the CTA component and all its subcomponents. 

To ensure the CTA functionality remains in the base components, I repurposed the CTAMixin to add basic call-to-action + icon features, plus video-playing features (from CTAVideoMixin), so that any component can easily extend the class and remove component bloat. 

This means that CTA functionality can now be natively seen in:
- Card
- Button
- Link with Icon

Due to Card now having all native CTA features within, this means that components such as:
- Card link
- Card in card

Can be slightly refactored/absorbed into Card in a future effort

### Changelog

**New**

- `dds-button` is now a new component with CTA features
- `dds-card`, and `link-with-icon` extend from the new CTA mixin

**Changed**

- CTAMixin now includes new properties and functions, as well as contents from CTAVideoMixin

**Removed**

- CTA stories


IgnacioBecerra:
This refactoring causes a butterfly effect throughout our component library, though mostly affecting styles and components that we still need to work on (card in card, card link, etc).

I've gone through the storybooks that were directly impacted (button, card, feature-card, link with icon) and they seem to be fine. At the moment I'm still going through documentation and style clean-up (such as moving into styles package and all), but in order to unblock @ariellalgilmore and @annawen1 in Card related work, I'm okay with merging this PR once it's approved, and I'll create another one to finish off the non-feature work there afterwards.

packages/web-components/src/components/button/button.ts
```diff
@@ -246,10 +149,5 @@ class DDSButtonExpressive extends FocusMixin(StableSelectorMixin(LitElement)) {
   static styles = styles; // `styles` here is a `CSSResult` generated by custom WebPack loader
 }
 
-console.warn(
-  'The dds-button-expressive component has been deprecated in favor of the core carbon button component using the isExpressive prop. ' +
-    'See carbon core button documentation for more information.'
-);
-
 /* @__GENERATE_REACT_CUSTOM_ELEMENT_TYPE__ */
-export default DDSButtonExpressive;
```
ariellalgilmore, line 255: looking good! my only concern is that we were planning to deprecate `DDSButtonExpresive` but this would be a breaking change. I don't think many people use `DDSButtonExpressive` on it's own, but might want to check with the team before merging

IgnacioBecerra: You're right, I added back the `DDSButtonExpressive` component along with a deprecation notice 👍

ariellalgilmore (approved):
Looking good!!! Thanks for the updates with the deprication. The only small issue I am seeing is some of the cta-types for button are not coming through on storybook (email, schedule, chat, and call). Otherwise let's merge this in!! nice job!
Screenshot 2023-08-30 at 9 12 20 AM

annawen1 (commented):
I do see some build errors happening from the refactoring

Screenshot 2023-08-30 at 1 08 37 PM

packages/web-components/src/components/cta/card-cta.ts
```diff
@@ -219,5 +221,9 @@ class DDSCardCTA extends VideoCTAMixin(CTAMixin(DDSCard)) {
...
 }
 
-/* @__GENERATE_REACT_CUSTOM_ELEMENT_TYPE__ */
-export default DDSCardCTA;
+export default deprecate(
+  /* @__GENERATE_REACT_CUSTOM_ELEMENT_TYPE__ */
+  DDSCardCTA,
+  `The dds-card-cta component has been deprecated. All its features have been absorbed into 
+  the base dds-card component. See migration guide for more information.`
+);
```
annawen1, line 229: I think @ariellalgilmore and I ran into build errors when we tried to use the `deprecate` utility on web components. We may have done something else instead to avoid the build errors

ariellalgilmore: oh yeah! the `deprecate` utility does not work with react wrappers. We instead just added `console.warn` statements instead like the following: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/feat/carbon-for-ibm-dotcom-v2/packages/web-components/src/components/pictogram-item/pictogram-item.ts#L61-L64

IgnacioBecerra:
@ariellalgilmore Fixed the issue!
@annawen1 Got the console warnings in places over the deprecate function 👍

PR #10878, opened 2023-08-24
fix(lightbox-video-player): use more reliable reference to video player

jkaeser:
### Related Ticket(s)

none

### Description

`<dds-video-cta-container>` can handle multiple Kaltura video embed requests within the same modal/lightbox by slotting a `<div>` per video into the same `<dds-lightbox-video-player>`. When the modal is activated by user interaction, the current video ID is sent along, and the `<dds-lightbox-video-player-composite>` sets `hidden` on the `<div>`s that don't have a matching ID. This works great when there is only one `<dds-video-cta-container>`.

The problems start when multiple video containers are used throughout the document. Each creates some number modal render root elements depending on the structure of the content, and the lightbox composite uses `document.querySelector('dds-lightbox-video-player')` to find the video player when trying to hide embedded Kaltura videos that are not active. As we know, `querySelector` returns the first matching element, and since we've established there may be multiple of these video players in various modal render roots, sometimes it found the wrong one and failed to hide non-active videos.

Luckily, the fix was pretty simple. Lightbox composites already have an internal reference to their video player, so I swapped out the document query to use that reference, which _should_ guarantee the composite is looking in the right place when hiding inactive videos.

### Changelog

**Changed**

- Ensure `<dds-lightbox-video-player-composite>` hides inactive videos by using an internal reference to modal's video player instead of querying for it from the document root.

### DIFF
--- a/packages/web-components/src/components/card-group/__stories__/card-group.stories.react.tsx
+++ b/packages/web-components/src/components/card-group/__stories__/card-group.stories.react.tsx
@@ -1,7 +1,7 @@
 /**
  * @license
  *
- * Copyright IBM Corp. 2020, 2022
+ * Copyright IBM Corp. 2020, 2023
  *
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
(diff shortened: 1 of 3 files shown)


andy-blum (approved):

m4olivei (approved):

jkaeser:
Just need approval from @ariellalgilmore and/or @IgnacioBecerra and we can get this one merged.

ariellalgilmore (commented):
LGTM! but do we need another story? The default story has a media knob which contains more than one video

jkaeser:
> LGTM! but do we need another story? The default story has a media knob which contains more than one video

That story embeds the same Kaltura video in each video card, and I specifically wanted to demonstrate embedding multiple different Kaltura videos. Would you prefer I change the default story to try to embed different videos per card?

ariellalgilmore:
> > LGTM! but do we need another story? The default story has a media knob which contains more than one video
> 
> @ariellalgilmore That story embeds the same Kaltura video in each video card, and I specifically wanted to demonstrate embedding multiple different Kaltura videos. Would you prefer I change the default story to try to embed different videos per card?

oh gotcha. yeah i think that would be best and update the react wrapper stories as well! thank you!!

ariellalgilmore (approved):
LGTM!

PR #10879, opened 2023-08-25
fix(pricing-table):  hide component examples

marcelojcs:
### Related Ticket(s)

Closes # https://jsw.ibm.com/browse/ADCMS-3666

### Description

PR to fix when viewing the Web Components in the Storybook, the Components list does not display "Pricing Table" as one of the listed components. 

### Changelog

**Changed**
Changed the css selector manager.js to also hide the child elements 

- {{removed thing}}

### DIFF
--- a/packages/web-components/.storybook/manager.js
+++ b/packages/web-components/.storybook/manager.js
@@ -24,7 +24,7 @@ addons.setConfig({
  */
 const getCss = (envVar, cssId) => {
   return envVar !== 'true'
-    ? `button[id^="${cssId}"] { display: none !important; }\n`
+    ? `[id^="${cssId}"] { display: none !important; }\n`
     : '';
 };
 


packages/web-components/.storybook/manager.js
```diff
@@ -24,7 +24,7 @@ addons.setConfig({
  */
 const getCss = (envVar, cssId) => {
   return envVar !== 'true'
-    ? `button[id^="${cssId}"] { display: none !important; }\n`
+    ? `[id^="${cssId}"] { display: none !important; }\n`
```
m4olivei, line 27: While this does hide the child elements, we're now left without a good way to get to the child stories of the Pricing Table. How would I navigate to Pricing Table > Without row headers or Pricing Table > With subheaders? Seems the only way is to search for each of those sub-stories by name, which is not great b/c there is overlap with other sub-stories.

I wonder if there is a better approach where a story could be marked Experimental / Unstable with a tag treatment? I've seen this recently on the [Cloud Common UI Storybook](https://pages.github.ibm.com/ibmcloud/cloud-common-ui/?path=/docs/components-animatedbackground--docs) . Or maybe we can just remove line 43 below and let Pricing Table show up by default?

m4olivei (changes requested):

packages/web-components/.storybook/manager.js
andy-blum: Deferring decision to @oliviaflory or @kennylam.

andy-blum:
Closing this one out too in favor of a different approach.

Issue #10298, opened 2023-04-04
[Tabs extended - with media]: Mobile grid alignment is off when other components are being inserted into the tab content area


proeung:
**JIRA ticket:** https://jsw.ibm.com/browse/ADCMS-3155

---------------

### Description

- Within the AEM page template, we're currently inserting different layout components (eg. Card Group, Image with caption, Accordion, etc) into the body content area of the `Tabs extended - with media` component.
- The grid alignment looks correct on larger breakpoints, however, it's off by `-16px` on mobile.
- Currently, we have a stop-gap solution in AEM by overriding styling within the shadow root, which is not recommended, but it's a workaround for now. cc: @KDarch928

![Screen Shot 2023-04-03 at 10 46 58 AM](https://user-images.githubusercontent.com/1815714/229897590-81434101-e7f2-4e17-a22b-0e2cb688202e.png)
![Screen Shot 2023-04-03 at 10 47 24 AM](https://user-images.githubusercontent.com/1815714/229897594-3d8c711b-daae-474c-9513-b0c3655fc83f.png)
![Screen Shot 2023-04-03 at 10 47 13 AM](https://user-images.githubusercontent.com/1815714/229897597-165020b2-a028-4df6-b199-b66cb12b81f2.png)

### Component(s) impacted

Tabs extended - with media

### Browser

Chrome, Safari, Firefox, Microsoft Edge

### Carbon for IBM.com version

v.1.45

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

AEM

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

N/A

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

proeung:
@oliviaflory FYI ^

@KDarch928 Feel free to add more detail for this particular issue.

oliviaflory:
Hey @proeung thanks for bringing this to our attention

I believe what's happening is that the accordion content area is extending 16px into the gutter on either side, this is to allow some components like cards to hang. 

[...]

PR #10353, opened 2023-04-18
fix(tabs-extended): apply correct mobile gutters 

jkaeser:
### Related Ticket(s)

Resolves #10298

### Description

This PR aligns elements in the mobile version of the `<dds-tabs-extended>` component to the Carbon grid. It achieves this by replacing the custom accordion markup with `<bx-accordion>` web components, which has the correct mobile styling.

### Changelog

**Added**
- Adds "With Mixed Content" Story that demonstrates selected web components used within tabs. 

**Changed**
- Replaces custom accordion markup with `@carbon/web-components`' `<bx-accordion>` markup.
- Makes tabs extend the full width available to them.

**Removed**
- Removes styles that were used for custom markup.

PR #10880, opened 2023-08-25
[dds-tabs-extended-media] Add a story for mixed tabs content to highlight grid issues

marcelojcs:
### Related Ticket(s)

[Jira ADCMS-3155](https://jsw.ibm.com/browse/ADCMS-3155)
https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/10298

### Description

The linked issue here reflects a broader issue wherein inline padding for tab / accordion contents varies in such a way that it can hang in the grid gutter, or even indent too far.

This PR adds a story under "Tabs extended - with media" with a range of mixed contents, reflecting those which are commonly used in AEM specfically, but also serve as a good general sample. This was based off a [Figma mockup](https://www.figma.com/file/kyG0InZz0e83hJaWskhG8I/AEM-component-expressions?node-id=42%3A288&mode=dev) of the various allowed components within "Tabs extended - with media" on AEM.

### Changelog

**New**

* added story "Tabs extended - with media / With Mixed Contents" to demonstrate variety of tab contents

**Changed**

* removed styles that no longer apply given a [recent change](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10650) that shuffled accordian markup out of the `<dds-tabs-extended-media>` shadow DOM.


emyarod (changes requested):
it looks like there is a syntax error? missing a closing brace

packages/styles/scss/components/tabs-extended/_tabs-extended.scss
```diff
@@ -206,6 +206,16 @@
       }
     }
   }
+  #{$prefix}--accordion__content {
```
kennylam, line 209: I'm not understanding the logic here. Both breakpoints have the same spacing. Is the intention for `spacing-05` to apply for `640px` and under? Also, breakpoint values are already mapped, and should be used instead of pixel values.

https://github.com/carbon-design-system/carbon/blob/3cfbc24171c351b475b712edc29a7bfd88f3eb76/packages/grid/scss/_config.scss#L39

m4olivei: This was a accidental commit inclusion from another body of work. It's not relevant here and I've taken it out.

m4olivei:
I've brought over the ["Tabs extended - with media / With Mixed Content" ](https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/10880/index.html?path=/story/components-tabs-extended-with-media--with-mixed-content)story from @jkaeser 's PR (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10353) here. There are a few things that we can observe now:

**1. Issue reported in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/10298 has been resolved by recent changes.**

The original issue, is no longer present, none of the contents are hanging in the gutter. All of the tab contents are getting the default padding that comes from [Carbon web component styles](https://github.com/carbon-design-system/carbon/blob/v10.58.0/packages/components/src/components/accordion/_accordion.scss#L135C40-L136). These styles used to be overidden for the `<dds-tabs-extended-media>` component [here](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/styles/scss/components/tabs-extended-media/_tabs-extended-media.scss#L57). Those styles no longer apply since [this change](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10650). That change pushed a lot of the accordion markup from `<dds-tabs-extended-media>` to `<dds-tab>`, such that the styles no longer apply. I've [removed these styles](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10880/commits/dc31c8a94a9a1d59c602ccbcd9d5d9b91605df08).

**Prior to change (v1.49.0):**
image

**After change (Canary)**
image

2. Some tab contents have magin / padding that interfere with us acheiving a clean inline grid layout

[The change](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10650) effectively acheived what @oliviaflory was calling for in [this comment on the original issue](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/10298#issuecomment-1501989629). Although the original specific issue is resolved, we still have the larger issue @oliviaflory pointed out where some tab contents have magin / padding that interfere with us acheiving a clean inline grid layout. 

3. Content item horizontal padding too much within tabs extended

In the original issue, @oliviaflory noted that Content item horizontal had padding to fit well within tab contents. Now that the tab contents container carries the right padding, tabs with content item horizontal have _too much_ overall padding and are now mis-aligned. See screenshot [taken from Canary](https://carbon-design-system.github.io/carbon-for-ibm-dotcom/canary/web-components/iframe.html?args=&id=components-tabs-extended-with-media--default&viewMode=story):

image

4. Mixed content and variable witdth tab contents.

When you have mixed content in the tabs, the tab contents have variable width and switching between the tabs, expands and contracts the overall component. You don't notice this with the existing story examples b/c the tab contents are homogenous (see below from deploy preview). This is b/c of some `display: flex` declarations to accomodate a section heading. AEM gets around this by setting a custom style like this:

```css
.tabs dds-tabs-extended-media::part(tabs-extended) {
    width: 100%;
}
```

I think we can mitigate this issue by reserving some of the `display: flex` style rules to only when there is a section heading in play.

Here is the bug from deploy preview:


https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/191049/7cf71829-9141-4887-94df-bbd669372f2f

m4olivei:
Discussed with @jkaeser and @marcelojcs:

* The inconsistencies in how components do / don't define inline padding probably deserves a discussion, breaking down how we handle fixing on a component by component basis. @oliviaflory had mentioned this was going to be addressed in v2 [way back in April](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/10298#issuecomment-1501989629). Was that done? If so, maybe we consider this lost for v1?
* There is value in the story added here, to at least highlight the various issues and set the goal posts for getting to an eventual fix around various issues noted in my previous comment.
* I've removed styles that no longer apply. That's worthwhile cleanup.
* The linked issues should be closed and various issues with narrower scopes should be opened if there is continued interest in fixing this stuff. Note that from an AEM perspective, there are various (non-ideal) work arounds in place, so it's not an _immediate_ issue as far as we're aware.

cc @andy-blum @kennylam

Issue #10553, opened 2023-06-12
[Pagination selection doesn't match actual rows shown]: pagination items per page selection shows 5 when the actual rows showing are 10


DoaaRadwan:
### Description

In [pagination](https://web-components.carbondesignsystem.com/?path=/docs/components-pagination--default) `bx-pagination` when used with the [data table](https://web-components.carbondesignsystem.com/?path=/docs/components-data-table--sortable-with-pagination) the `:page-size` doesn't match the `bx-page-sizes-select` selected `option` (what is displayed on the screen to the user as the selected value in the drop down menu) on reload of the table. 

![Screenshot 2023-06-12 at 9 58 47 am](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/24454169/06bf6c15-7a6f-4f3c-8c45-ebf3e8b5dd14)

![Screenshot 2023-06-12 at 10 19 38 am](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/24454169/fe6bc40e-c631-49bf-b223-8325ed1658ce)

If the user chooses 10 and the table is reloaded though a `v-if  v-else` the table will show 10 rows (since the variable for `:page-size` doesn't get reseted) while the items per page selection returns back to default to 5. The selection can't be changed in code, so there is no work around.

### Component(s) impacted

1. [Data table](https://web-components.carbondesignsystem.com/?path=/docs/components-data-table--sortable-with-pagination)
2. [Pagination](https://web-components.carbondesignsystem.com/?path=/docs/components-pagination--default) 

### Browser

Chrome

### Carbon for IBM.com version

v1.27.0

### Severity

Severity 1 = The design is broken in a critical way that blocks users from completing tasks or damages the brand. Affects major functionality, no workaround.

### Application/website

https://registry.st4sd.res.ibm.com/registry-ui/

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://codesandbox.io/s/github/carbon-design-system/carbon-for-ibm-dotcom/tree/main/packages/carbon-web-components/examples/codesandbox/basic/components/data-table

### Steps to reproduce the issue (if applicable)

1. Have more than 10 rows in a data table.
2. Have the options for the selection be 5 10 20
3. Have a v-if v-else on the table
4. If user selects 10 rows then v-if v-else triggers 
5. When table renders back the actual rows showing will remain 10
6. But the selection will be 5

### Code of Conduct

[...]

PR #10884, opened 2023-08-25
fix(pagination): mark selected option in page-sizes-select component

m4olivei:
### Related Ticket(s)

Closes https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/10553
https://jsw.ibm.com/browse/ADCMS-3821

### Description

Fixes a bug where the initial value of the `page-size` attribute on the `<bx-pagination>` component previously did not set the correct `<option>` element within the `<bx-page-sizes-select>` component.

### Changelog

**Changed**

- Updated the `slotchange` handler for the `<bx-page-sizes-select>` component to set the `selected` attribute when applicable.

### Testing Instructions

* Navigate to Pagination > Default story
* Enter a different value for "Number of rows per page (page-size)", using one of the other values in the dropdown, eg. 20 or 30
* Click the "Open canvas in new tab" button, this will initiate the story with the set value for `page-size` to test the bug
* Should see the value you set for the items per page drop down.

### DIFF
--- a/packages/carbon-web-components/src/components/pagination/page-sizes-select.ts
+++ b/packages/carbon-web-components/src/components/pagination/page-sizes-select.ts
@@ -54,14 +54,18 @@ class BXPageSizesSelect extends FocusMixin(LitElement) {
    * @param event The event.
    */
   private _handleSlotChange({ target }: Event) {
-    const { _selectNode: selectNode } = this;
+    const { _selectNode: selectNode, value } = this;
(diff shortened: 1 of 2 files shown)


emyarod (approved):
looks good to me!

m4olivei:
Thanks @emyarod ! I've labeled as Ready to merge and updated with the `main` branch.

PR #10892, opened 2023-08-28
feat(locale-modal): only render modal when its called by user click

pjudge:
### Related Ticket(s)

### Description

### Changelog

**New**

- {{new thing}}

**Changed**

- {{changed thing}}

**Removed**

- {{removed thing}}

### DIFF
--- a/packages/web-components/src/components/footer/footer-composite.ts
+++ b/packages/web-components/src/components/footer/footer-composite.ts
@@ -62,6 +62,7 @@ class DDSFooterComposite extends MediaQueryMixin(
    * Handles `click` event on the locale button.
    */
   private _handleClickLocaleButton = () => {
+    console.log("HALP");
     this.openLocaleModal = true;
   };
 
@@ -245,7 +246,8 @@ class DDSFooterComposite extends MediaQueryMixin(
       openLocaleModal,
       _loadLocaleList: loadLocaleList,
     } = this;
-    return html`
+    return openLocaleModal 
+    ? html`
       <dds-locale-modal-composite
         lang-display="${ifNonNull(langDisplay)}"
         language="${ifNonNull(language)}"
@@ -254,7 +256,8 @@ class DDSFooterComposite extends MediaQueryMixin(
         .localeList="${ifNonNull(localeList)}"
         ._loadLocaleList="${ifNonNull(loadLocaleList)}">
       </dds-locale-modal-composite>
-    `;
+    `
+    : html ``;
   }
 
   renderLanguageSelector(slot = 'language-selector') {


pjudge:
Setting this back to draft as I just found some more things I can lazy load.

kennylam (commented):
@pjudge Looks good! The only thing I'm noticing is the fade in/out transitions are missing, probably because of the way it's being added/removed from the DOM?

pjudge:
> The only thing I'm noticing is the fade in/out transitions are missing, probably because of the way it's being added/removed from the DOM?

Great catch @kennylam , I'm looking into this now. Thank you!

packages/web-components/src/components/locale-modal/locale-modal-composite.ts
```diff
@@ -181,32 +192,50 @@ class DDSLocaleModalComposite extends HybridRenderMixin(LitElement) {
...
-            ({ country, href, language, locale, region }) => html`
-              <dds-locale-item
-                country="${country}"
-                href="${href}"
-                language="${language}"
-                locale="${locale}"
-                region="${region}">
-              </dds-locale-item>
+
+        ${this.chosenRegionRender
```
jkaeser, line 201: You might be able to get away with checking for truthiness of `this.chosenRegion` here, at which point you'd no longer need `this.chosenRegionRender` at all.

packages/web-components/src/components/footer/footer-composite.ts
```diff
@@ -48,6 +48,9 @@ import { carbonElement as customElement } from '../../internal/vendor/@carbon/we
 
 const { stablePrefix: ddsPrefix } = ddsSettings;
 
+// This matches the CSS transition timing.
+const animationTiming = 240;
```
jkaeser, line 52: See https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/carousel/carousel.ts#L15 and https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/carousel/carousel.ts#L164 for an example of getting this value from [the relevant Carbon package](https://github.com/carbon-design-system/carbon/blob/main/packages/motion/src/index.js#L11).

packages/web-components/tests/e2e-storybook/cypress/integration/locale-modal/locale-modal.e2e.js
```diff
@@ -77,9 +77,10 @@ describe('dds-locale-modal | default', () => {
     const closeButton = cy
       .get('dds-locale-modal')
       .shadow()
-      .find('dds-expressive-modal-close-button');
-    closeButton
+      .get('dds-expressive-modal-close-button')
```
jkaeser, line 80: I think we do actually want `.find` here. See https://docs.cypress.io/api/commands/get#Get-vs-Find.

jkaeser (changes requested):
So far this is looking great. Left a few notes for your perusal.

packages/web-components/tests/e2e-storybook/cypress/integration/locale-modal/locale-modal.e2e.js
pjudge: well. it wasn't working with find before, but i tried again and it does. <shrug> thanks!

packages/web-components/src/components/locale-modal/locale-modal-composite.ts
pjudge: i originally tried this and for some reason couldn't get it to work, hence the new var. but i will try again! thank you

packages/web-components/src/components/footer/footer-composite.ts
pjudge: Thank you!!! I tried finding this and wasn't able to. So helpful.

packages/web-components/src/components/locale-modal/locale-modal-composite.ts
```diff
@@ -181,32 +192,50 @@ class DDSLocaleModalComposite extends HybridRenderMixin(LitElement) {
...
-        </dds-locale-search>
+          : ``}
       </dds-locale-modal>
     `;
   }
 
+  /**
+   * Handles click on a region.
+   */
+  private _handleRegionClick(target) {
```
m4olivei, line 234: Doesn't make any functional difference, but by convention the argument name for an event handler is typically `event` or `e` or something along those lines. Especially since there is a `target` property on it, it reads better to have it be `event.target`, similarily `event.currentTarget`.

packages/web-components/src/components/locale-modal/locale-modal-composite.ts
```diff
@@ -181,32 +192,50 @@ class DDSLocaleModalComposite extends HybridRenderMixin(LitElement) {
...
+    this.chosenRegion = target.currentTarget.__name;
```
m4olivei, line 235: Doesn't it also work to refer to the `name` property, eg. `event.currentTarget.name`? Where does the `__name` property come from? From what I can tell `event.currentTarget` is going to be a `<dds-region-item>` right, eg.

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/61f2eaa348ee12f1c8074f3242d3ec0055398fd7/packages/web-components/src/components/locale-modal/region-item.ts#L35-L39

m4olivei (approved):
Just a couple small suggestions, nothing blocking. This is great! TIL about [`LitElement.updateComplete`](https://lit.dev/docs/components/lifecycle/#updatecomplete)

jkaeser (approved):

jkaeser:
Noticing some issues with auto-setting focus on elements in response to user actions. Working on it!

m4olivei:
> @m4olivei I noticed what could be considered a design flaw with our implementation, and I had a little time to spare so I went ahead with a light refactor. [31e018b](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/commit/31e018b452bd55696aadb3d1a475ce27aa956011) & [b99ab12](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/commit/b99ab1285ba0b1f0530e2eff3dce850288f92a4b) endeavor to consolidate some logic.
> 
> > Refresher: The user flow is to open the modal, select a **region**, then select a **locale** that is available in that region.
> 
> Instead of the `<dds-locale-modal-composite>` and the `<dds-locale-modal>` each rolling their own method to keep track of the selected region, the modal itself handles that and reports back to the composite via an event. This keeps the two in sync, which empowers the composite to perform re-renders at all the right times. As a result, the foundation is laid to render only the active region's locale options. This is a further DOM node reduction, and obviates the need for the `<dds-locale-search>` element to keep track of the active region and factor that into its functionality.

This is all really nice. Great work.

> Noticing some issues with auto-setting focus on elements in response to user actions. Working on it!

Gotcha! I was just about to give the thumbs up. I do see that now you point it out. Will watch for the fix.

packages/web-components/src/components/expressive-modal/expressive-modal.ts
```diff
@@ -484,10 +484,10 @@ class DDSExpressiveModal extends StableSelectorMixin(
         this.ownerDocument.body.style.overflow = 'hidden';
         this.removeAttribute('aria-hidden');
         this._launcher = this.ownerDocument!.activeElement;
+        await this._waitForTransitionEnd();
```
jkaeser, line 487: Moving this up before the `querySelector` gives elements time to render into the DOM.

Issue #10899, opened 2023-08-31
[Content Section]: transfer playground stories into local instance


IgnacioBecerra:
A follow-up to this issue: [#10454](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/10454)

Due to the way Storybook compiles components and styles, we need to restart the storybook instance every time we make any changes within the `styles` package.

This issue attempts to transfer the playground stories created in the aforementioned issue to a local instance, or stackblitz/codesandbox in order for hot reloading or a similar feature to start fixing the content section and component spacing.

PR #10915, opened 2023-09-05
chore(storybook): script to use local styles in dev

IgnacioBecerra:
### Related Ticket(s)
[#10899](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/10899)

### Description
Due to the way Storybook styles are compiled, and our current `web-components/*.scss` files are importing from the package directly, any update to the files within the `styles` package will only be seen after restarting the storybook instance. In order to circumvent this issue, we could instead import the local file instead of the package styles. However this can be a little tedious, so this PR automates that process. 

This PR introduces a new bash script that will run parallel with our storybook instance (using the `yarn start-dev` command), that will change the import directory to the local one depending on any modifications done to the file.

For example, if we modified the "package/scss/components/card/_card.scss` file, the styles will only be seen after storybook is restarted. The bash script introduced there does the following:

1. Obtains the component name of the modified file (`card` in this case`)
2. In the `web-components/src/components/card` directory, it will look for any `.scss` files and iterate through them (except under the stories' `.scss` files)
3. Finds and replace `@use '@carbon/ibmdotcom-styles/scss/components/card';` to `../../../../styles/scss/components/card` in order to use local styles
4. Storybook will reload with the updated styles without the need to restart
5. Upon exiting the Storybook command (using ctrl + c), the script also makes sure to revert the changes so all the style files under `web-components/src/*.scss` use the package style imports.

### Changelog

**New**

- introduced script to make it easier to work with local styles in our storybook instance


packages/web-components/package.json
```diff
@@ -60,6 +60,7 @@
     "doctoc": "doctoc --title '## Table of contents' docs && doctoc --title '## Table of contents' README.md",
     "postinstall": "carbon-telemetry collect --install",
     "start": "cross-env NODE_OPTIONS=--openssl-legacy-provider yarn storybook",
+    "start-dev": "./update-scss-paths.sh & cross-env NODE_OPTIONS=--openssl-legacy-provider yarn storybook",
```
emyarod, line 63:
```suggestion
    "start-dev": "./update-scss-paths.sh & yarn start",
```

is it possible to reference the existing `start` script?

packages/web-components/update-scss-paths.sh
```diff
@@ -0,0 +1,37 @@
...
+  # search for scss files under directory
+  find "$directory" -type f -name "*.scss" ! -path "*/__stories__/*" | while read -r scss_file; do
+    # search and replace the string in the file
+    sed -i "" "s#$search_string#$replace_string#g" "$scss_file"
+		echo "Replaced imports in $scss_file"
+  done
+}
+
+# monitor for file changes under the `styles` directory
+fswatch -0 ../styles/ | while read -d "" changed_file
```
emyarod, line 20: does `fswatch` need to be installed first? when I run the script the first line of the output is

image

and then I believe the `cleanup` function runs

IgnacioBecerra: Yes, `brew install fswatch` should make it work, forgot to add that in the PR description!

packages/web-components/package.json
IgnacioBecerra: True, will change to that!

packages/web-components/update-scss-paths.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+declare -a component_names
+
```
emyarod, line 4: would it be useful to check for the dependency at the start? something like:

```suggestion

if ! command -v fswatch &>/dev/null; then
  echo "fswatch required (https://github.com/emcrisostomo/fswatch)"
  exit 1
fi

```

but with a more descriptive message

IgnacioBecerra: I like that idea 👍

ariellalgilmore (approved):
LGTM!

PR #10941, opened 2023-09-15
feat(leaving-ibm): v2 style updates

### LINKED ISSUE #10856: [Component] Leaving IBM

sangeethababu9223:
### Related Ticket(s)

Closes #10856

### Description

Style / layout break in Leaving IBM Component in V2

### Changelog

**Changed**

- button element changed
- Styles updated to match the latest


oliviaflory:
@sangeethababu9223 this is looking really good from a design perspective! A few fixes I found: 

1. - [ ] The  focus state for the link is getting cut off when (tabbing into the link) see the screen shot attached
Screenshot 2023-09-19 at 3 53 31 PM


2. - [ ] I cannot interact with the link within the modal, was it like that in the [v1 version](https://www.ibm.com/standards/carbon/web-components/?path=/story/components-leaving-ibm--default)?
3. - [ ] There are no interactive states on the link within the modal (hover, focus, click) they should match the standard link states, you can see them documented in the [Carbon link website docs](https://carbondesignsystem.com/components/link/style#interactive-states) or [CWC](https://carbon-design-system.github.io/carbon-for-ibm-dotcom/beta/carbon-web-components-v2/?path=/story/components-link--default)
4. - [ ] The focus should probably be trapped within the modal as it is within the [v1 version](https://www.ibm.com/standards/carbon/web-components/?path=/story/components-leaving-ibm--default), right now I can use the keyboard to tab outside of the modal and navigate around on the page under the overlay which will be confusing for the users.
5. - [ ] On mobile the heading and close button are getting cut off at the top of the browser, while there is a gap at the bottom of the browser and I can see overlay. I am expecting to see the same behavior as the v1 version where the heading is not cut off and the button is pinned to the bottom and I should not see the overlay.

| **Current mobile** | **Expecting** |
| --- | --- |
|  Screenshot 2023-09-19 at 4 06 01 PM |  Screenshot 2023-09-19 at 4 06 24 PM |

packages/carbon-web-components/src/components/modal/modal.scss
```diff
@@ -18,7 +18,15 @@
 
 :host(#{$prefix}-modal[open]) {
   @extend .#{$prefix}--modal;
-  @extend .is-visible;
+  // @extend .is-visible;
```
IgnacioBecerra, line 21:
```suggestion
```

packages/styles/scss/components/leaving-ibm/_leaving-ibm.scss
```diff
@@ -0,0 +1,56 @@
...
+
+  :host(#{$c4d-prefix}-leaving-ibm-modal) {
+    @extend :host(#{$prefix}-modal);
+
+    max-height: 100%;
+
+    &[open] {
+      @extend :host(#{$prefix}-modal[open]);
+    }
+  }
```
IgnacioBecerra, line 29:
```suggestion
  :host(#{$c4d-prefix}-leaving-ibm-modal) {
    @extend :host(#{$prefix}-modal);

    max-height: 100%;
  }

  :host(#{$c4d-prefix}-leaving-ibm-modal[open]) {
      @extend :host(#{$prefix}-modal[open]);
  }  
```

Due to a previous issue we've encountered, doing `&[open]` might not work in some environments, so we're having these properties be its own declarations within `:host()` .

IgnacioBecerra (commented):
It's looking good! Olivia addressed most of the things I wanted to mention, so just a few comments code-wise, will re-review once the design changes have been implemented :) 👍

sangeethababu9223:
@oliviaflory,
All of the issues except the focus trap not working in modal should be fixed now.
Focus trap seems to be an issue with Modal component. I've created a separate ticket for this: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/10966
Will work on that once this issue is resolved.

oliviaflory (approved):
@sangeethababu9223 Thank you! FYI the focus trap is working for me on Chrome, but it would be good to track for all browsers if you are noticing it not working on another 😄

sangeethababu9223:
Thanks @oliviaflory,
I faced issue with focus trap in in both Safari & Firefox. I'll look into this for the other ticket.

PR #10952, opened 2023-09-18
Feat/merge alpha masthead to v2 main

jkaeser:
### Related Ticket(s)

### Description

### Changelog

**New**

- {{new thing}}

**Changed**

- {{changed thing}}

**Removed**

- {{removed thing}}


packages/carbon-web-components/src/components/tabs/tabs.ts
```diff
@@ -411,42 +411,64 @@ export default class CDSTabs extends HostListenerMixin(CDSContentSwitcher) {
...
+    const {
       _assistiveStatusText: assistiveStatusText,
       _handleSlotchange: handleSlotchange,
     } = this;
 
-    const previousButtonClasses = classMap({
-      [`${prefix}--tab--overflow-nav-button`]: true,
-      [`${prefix}--tabs__nav-caret-left`]: true,
-      [`${prefix}--tab--overflow-nav-button--previous`]: true,
-      [`${prefix}--tab--overflow-nav-button--hidden`]:
```
m4olivei, line 426: This class doesn't get carried over to the `<button>` rendering in `renderPreviousButton`, is that intentional? I see that we render `null` if `isIntersectionLeftTrackerInContent`, so I'm guessing thats why we don't need it? If so, should be clean up styles that correspond to this class?

packages/carbon-web-components/src/components/tabs/tabs.ts
```diff
@@ -411,42 +411,64 @@ export default class CDSTabs extends HostListenerMixin(CDSContentSwitcher) {
...
-        isIntersectionLeftTrackerInContent,
-    });
-    const nextButtonClasses = classMap({
-      [`${prefix}--tab--overflow-nav-button`]: true,
-      [`${prefix}--tabs__nav-caret-right`]: true,
-      [`${prefix}--tab--overflow-nav-button--next`]: true,
-      [`${prefix}--tab--overflow-nav-button--hidden`]:
```
m4olivei, line 433: Simlar note to `previousButtonClasses` note.

m4olivei (changes requested):
This is looking great! Clearly a lot of little things that had to happen to pull it together. Nice work @jkaeser ! Just one issue, and a couple notes.

Been comparing current masthead v2 to the deploy preview here, found one issue:

* borders are missing from the left nav at the top level and second level
image

Otherwise, don't think it's an issue but just wanted to note it:

* `line-height` changes in the `.link-heading` div that is part of `<c4d-megamenu-category-link>`, but that appears to be b/c of upstream changes in the CSS vars. This is probably expected:

image

packages/carbon-web-components/src/components/tabs/tabs.ts
jkaeser: Yep, no need to check for hidden if we don't render it at all. The corresponding styles actually come from the upstream `@carbon/styles` package: https://github.com/carbon-design-system/carbon/blob/main/packages/styles/scss/components/tabs/_tabs.scss#L104-L106

packages/carbon-web-components/src/components/tabs/tabs.ts
jkaeser: Same answer 🙂

m4olivei (approved):
Changes look great!

kennylam (approved):
LGTM thanks @jkaeser!

PR #10976, opened 2023-09-27
Change(storybook) Create storybook experimental tab

marcelojcs:
### Related Ticket(s)

Closes # https://jsw.ibm.com/browse/ADCMS-3995

### Description

Move hidden components to an "experimental" tab on storybook.

### Changelog

**Changed**

Moved the hidden components to an "experimental" tab on  storybook.

**Removed**

Code hiding the components on manager.js


packages/web-components/.storybook/manager.js
```diff
@@ -10,47 +10,4 @@ import yourTheme from './theme';
...
-}
-
-// Inject any CSS rules into the page.
-if (css.length) {
-  const head = document.head || document.getElementsByTagName('head')[0];
-  const style = document.createElement('style');
-  head.appendChild(style);
-  style.appendChild(document.createTextNode(css));
-}
+});
```
m4olivei, line 13: Nit: Add a newline at the end of the file here.

m4olivei (approved):
I like this a lot. Much cleaner. Like the consistency with Carbon.

andy-blum (approved):
👏 much simpler and more maintainable!

packages/web-components/src/components/content-block-card-static/__stories__/content-block-card-static.stories.ts
```diff
@@ -70,7 +70,7 @@ export const Default = (args) => {
 };
 
 export default {
-  title: 'Components/Content block card static',
+  title: 'Experimental/Content block card static',
```
jkaeser, line 73: The e2e test files point to Storybook URLs, and changes like this one have updated those URLs, causing tests to fail. But don't worry, the fixes will be easy! Just find the source tests files and update the paths to match the new pattern. For example, see: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/47a59204d8624c73c2bfdf230e6091d354212cf4/packages/web-components/tests/e2e-storybook/cypress/integration/masthead/cloud/cloud-masthead.e2e.js#L14

jkaeser (changes requested):
@marcelojcs Looks like you need to update some URLs in the tests.

m4olivei:
Just noting that @kennylam and I discussed this at Carbon office hours yesterday, here are my notes:

* I demo'd the solution, grouping experimental stories under a sub-heading in the left sidebar, like [Carbon](https://react.carbondesignsystem.com/?path=/docs/getting-started-welcome--welcome).
* @kennylam would like to get it in. However Storybook, and decisions about the way it's presented to adopters is more left to the designers and leadership. Kenny to check with @oliviaflory and potentially others for their thoughts.

jkaeser (approved):

oliviaflory:
This looks good to me. I would like to see the following two stories removed entirely from SB before approving because we are deprecating the pre-configured content blocks in v2 and I'd rather not expose these random ones, even as experimental, in our v1 package:
- content block card static
- content block headlines

marcelojcs:
@oliviaflory 
[Addresed on the latest commit](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10976#issuecomment-1761827713)

oliviaflory (approved):
Thank you @marcelojcs

PR #10998, opened 2023-10-03
feat(masthead-v2): add masthead to document flow

jkaeser:
### Description

The visible parts of the masthead are `position: fixed;`, which prevents them from being in the document flow. This forces adopters to manually add spacing to their page content so the masthead doesn't overlap. We provide the `--dds-sticky-header-height` variable via the StickyHeader utility as a value to use, but this changes as the user scrolls, creating a potentially unwanted parallax effect.

parallax

To overcome this, we force the masthead's container into the document flow by giving it a height equal to its visible parts' height. This avoids the parallax effect, and adopters should only need to add the spacing they want between their content and the masthead, rather than solving for the masthead's height in some way.

### Testing

Visit deploy preview Stories for the masthead with and without the L1. In each, use the browser dev tools to inspect the element and find the `<dds-masthead-container>` or `<dds-masthead-composite>` (they are mostly analogous). It should have a CSS `height` set on it equal to the `<dds-masthead>` element's height.

### Changelog

**Changed**

- Puts the masthead into the document flow so it naturally takes up space equivalent to its height.

### DIFF
--- a/packages/web-components/src/components/masthead/masthead-composite.ts
+++ b/packages/web-components/src/components/masthead/masthead-composite.ts
@@ -1336,7 +1336,7 @@ class DDSMastheadComposite extends HostListenerMixin(LitElement) {
   /**
    * Mutation observer
    */
-  private _heightMutationObserver = new MutationObserver(
+  private _heightResizeObserver = new ResizeObserver(
     this._setContainerHeight.bind(this)
   );
 
@@ -1345,8 +1345,10 @@ class DDSMastheadComposite extends HostListenerMixin(LitElement) {
    */
   protected _setContainerHeight() {
     const { mastheadRef } = this;
-    this.style.display = 'block';
-    this.style.height = `${mastheadRef.getBoundingClientRect().height}px`;
+    if (mastheadRef) {
+      this.style.display = 'block';
+      this.style.height = `${mastheadRef.getBoundingClientRect().height}px`;
+    }
   }
 
   createRenderRoot() {
@@ -1373,12 +1375,8 @@ class DDSMastheadComposite extends HostListenerMixin(LitElement) {
       this.requestUpdate();
     });
 
-    // Watch for changes to dds-masthead's immediate children.
-    this._heightMutationObserver.observe(this.mastheadRef, {
-      childList: true,
-      subtree: false,
-      attributes: false,
-    });
+    // Keep render root's height in sync with dds-masthead.
+    this._heightResizeObserver.observe(this.mastheadRef);
   }
 
   updated(changedProperties) {
@@ -1396,7 +1394,7 @@ class DDSMastheadComposite extends HostListenerMixin(LitElement) {
 
   disconnectedCallback() {
     super.disconnectedCallback();
-    this._heightMutationObserver.disconnect();
+    this._heightResizeObserver.disconnect();
   }
 
   render() {
--- a/packages/web-components/src/components/masthead/masthead-composite.ts
+++ b/packages/web-components/src/components/masthead/masthead-composite.ts
@@ -14,6 +14,7 @@ import {
   customElement,
   LitElement,
   TemplateResult,
+  query,
 } from 'lit-element';
 import { nothing } from 'lit-html';
 import { ifDefined } from 'lit-html/directives/if-defined';
@@ -1326,6 +1327,28 @@ class DDSMastheadComposite extends HostListenerMixin(LitElement) {
       : unauthenticatedCtaButtons;
   }
 
+  /**
+   * A reference to the dds-masthead element.
+   */
+  @query(`${ddsPrefix}-masthead`)
+  mastheadRef;
+
+  /**
+   * Mutation observer
+   */
+  private _heightMutationObserver = new MutationObserver(
+    this._setContainerHeight.bind(this),
+  );
+
+  /**
+   * Sets root element's height equal to the height of the fixed masthead elements.
+   */
+  protected _setContainerHeight() {
+    const { mastheadRef } = this;
+    this.style.display = 'block';
+    this.style.height = `${mastheadRef.getBoundingClientRect().height}px`;
+  }
+
   createRenderRoot() {
     // We render child elements of `<dds-masthead-container>` by ourselves
     return this;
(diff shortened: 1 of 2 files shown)


packages/web-components/src/components/masthead/masthead-composite.ts
```diff
@@ -1349,6 +1372,13 @@ class DDSMastheadComposite extends HostListenerMixin(LitElement) {
       this._isMobileVersion = layoutBreakpoint.matches;
       this.requestUpdate();
     });
+
+    // Watch for changes to dds-masthead's immediate children.
+    this._heightMutationObserver.observe(this.mastheadRef, {
+      childList: true,
+      subtree: false,
+      attributes: false,
+    });
```
andy-blum, line 1381: I like that we're watching for changes to try to keep the heights in sync, but I wonder if a [resizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) would be the better choice here.

packages/web-components/src/components/masthead/masthead-composite.ts
```diff
@@ -1326,6 +1327,28 @@ class DDSMastheadComposite extends HostListenerMixin(LitElement) {
...
+   */
+  @query(`${ddsPrefix}-masthead`)
+  mastheadRef;
+
+  /**
+   * Mutation observer
+   */
+  private _heightMutationObserver = new MutationObserver(
+    this._setContainerHeight.bind(this)
+  );
```
andy-blum, line 1341: This throws up a red flag for me after @m4olivei discovered some issues in the carousel component defining an intersectionObserver in a property definition.

I think those issues were limited to that specific use case as we were attempting to set the root to something in the shadowroot which hadn't reliably existed, so I think we're ok, but I'd like Matt's input.

packages/web-components/src/components/masthead/masthead-composite.ts
```diff
@@ -1326,6 +1327,28 @@ class DDSMastheadComposite extends HostListenerMixin(LitElement) {
...
+
+  /**
+   * Sets root element's height equal to the height of the fixed masthead elements.
+   */
+  protected _setContainerHeight() {
+    const { mastheadRef } = this;
+    this.style.display = 'block';
```
andy-blum, line 1348: Is there a reason to set this as an inline style as well as in the SCSS file?

packages/web-components/src/components/masthead/masthead-composite.ts
```diff
@@ -1326,6 +1327,28 @@ class DDSMastheadComposite extends HostListenerMixin(LitElement) {
...
+    this.style.height = `${mastheadRef.getBoundingClientRect().height}px`;
```
andy-blum, line 1349: Does this line throw a ts error? `mastheadRef` is potentially null as the result of a querySelector.

andy-blum (changes requested):
This looks pretty good - suggesting a change from `mutationObserver` to `resizeObserver`

packages/web-components/src/components/masthead/masthead-composite.ts
m4olivei: This _should_ be OK. If we were actually calling the `this._setContainerHeight` function at this point, that's where we'd be in trouble, b/c then in that function it references `this.mastheadRef`, which since it's a `@query` decorator, would be `null`. As it is, `this._setContainerHeight.bind(this)` just returns a reference to that function bound to `this`. It will actually be called later, at which time it will pull the reference to `this.mastheadRef`, which should be populated by then (it gets populated [after the first `update` lifecycle call](https://codepen.io/m4olivei/pen/JjwLMZx)).

The situation we ran into before ([detailed in PR description here](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10971#issue-1912120552)), was that a property decorated with `@query` was referenced in a class field initializer, at which point it was `null`.

packages/web-components/src/components/masthead/masthead-composite.ts
jkaeser: The SCSS file addition is kind of a backup. The styles there aren't actually applying.

packages/web-components/src/components/masthead/masthead-composite.ts
jkaeser: I didn't get a TS error, but you're correct that I could write this more defensively. For now, this method is only ever executed as a callback to the mutation observer that's watching the element `mastheadRef` represents, so I don't _think_ there's an opportunity for it to be null. But I'll add a check anyway 😄

packages/web-components/src/components/masthead/masthead-composite.ts
jkaeser: Good call.

packages/web-components/src/components/masthead/masthead-composite.ts
andy-blum: That's odd. Any ideas why?

m4olivei (approved):
Looks great! Just small comment about updating a comment.

packages/web-components/src/components/masthead/masthead-composite.ts
jkaeser: Nope, but I didn't want to spend too much time digging into it, especially since it doesn't appear to be new (no regressions) and this alpha is not long for this world. Can be addressed in the v2 upgrade if still an issue.

andy-blum (changes requested):

packages/web-components/src/components/masthead/masthead-composite.ts
```diff
@@ -1326,6 +1327,49 @@ class DDSMastheadComposite extends HostListenerMixin(LitElement) {
...
+  /**
+   * Throttled callback for _heightResizeObserver.
+   */
+  protected _resizeObserverCallback() {
+    if (!this._resizeObserverThrottle) {
+      this._setContainerHeight();
+      this._resizeObserverThrottle = setTimeout(() => {
+        this._resizeObserverThrottle = undefined;
+      }, 100);
+    }
```
andy-blum, line 1359: I think the logic here might be backwards: "if there isn't a timer, set the height and prevent additional changes for 100ms". This would mean you could have up to 100ms of additional resizes without the setContainerHeight method firing.

we probably want something closer to this:
```ts
clearTimeout(this._resizeObserverThrottle);
setTimeout(this._setContainerHeight.bind(this), 100)
```

Then, in a series of rapidly firing events, we wait for an idle period of 100ms and then adjust.

andy-blum (changes requested):

andy-blum (approved):

kennylam (approved):
LGTM thanks @jkaeser!

PR #11002, opened 2023-10-04
fix(pricing-table): cell items were being horizontally squished into …

bruno-amorim:
…the cell

### Related Ticket(s)

https://jsw.ibm.com/browse/ADCMS-3714

### Description

In a Pricing Table, for the Table Row when the user choses to "Add Cell Content" for cells to contain multiple paragraphs, or combination of paragraph text, Tag Group tags, or Links, when the page is published the Pricing table is displaying this added cell content distributed horizontally squished into the cell. The Added Cell Content is displaying to the right of the initial cell content in the cell.
Screenshot 2023-10-04 at 11 29 52

I've changed the flex direction to column inside of the cell contents so that all items will sit one below the other.


jkaeser (approved):
Love it! Simple and effective.

What do you think about updating the Pricing Table stories to slot multiple elements into the table cells, i.e. reflecting the conditions the bug appeared in? This would make it easier for other reviewers to test (I had to manually make changes to the DOM to see the fix in action), and it would help us catch any regressions in the future.

jkaeser (changes requested):
You know what, @bruno-amorim ? I just remembered why this was made into a flex container to begin with. There's an option to add an icon into the cell content (e.g. the green checkmark in the Storybook examples), and when that's present, user-entered text should appear appear next to the icon. Sorry for not remembering earlier!

I think the fix is going to be a little more complicated. We need to make sure the user-entered content is wrapped in a containing element (i.e. a `<div>` or similar) that isn't set to `display: flex;`, and that containing element needs to be a sibling of the optional icon, both of which need a parent flex container. Something like the following sketch code:
```html
<!-- dds-pricing-table-cell's shadow-root markup -->
<div style="display: flex;">
  <div class="optional-icon"></div>
  <div class="user-content">
    <slot></slot>
  </div>
</div>
```
The tricky part is going to be getting that to work with all the other existing features, such as annotations. Let me know if you want a hand reasoning through it.

packages/styles/scss/components/pricing-table/_pricing-table.scss
```diff
@@ -155,6 +155,7 @@
 
     .#{$prefix}--pricing-table-cell-content {
       display: flex;
+      flex-direction: column;
       align-items: start;
```
jkaeser, line 159: Are these styles doing anything for us anymore? I disabled them in the browser and didn't see any adverse effects, so I think they're safe to remove.

jkaeser (approved):
Nicely done, @bruno-amorim! I left one small note for potential cleanup, but this looks great.

packages/styles/scss/components/pricing-table/_pricing-table.scss
bruno-amorim: Hey @jkaeser, thanks a lot for reviewing this PR 😁
That's the style that's making all flex items to be horizontally organized. Yeah, removing it in the browser's devtool makes no immediate effect but when I remove it from the source code and rerun the storybook them items get squished again:

Screenshot 2023-10-11 at 14 49 56

jkaeser: I mean the entire block. Removing all of these resulted in no visual changes for me:
```css
display: flex;
flex-direction: column;
align-items: start;
```
Note the styles are duplicated because of the way the CSS is bundled, so you need to disable them in two places in dev tools.

oliviaflory (changes requested):
- [ ] 1.  There needs to be some margin between the body & tag and tag & link elements, they should not touch each other as they do in the current build. If you do not have a designer specing these spacings for you, I would suggest:
- 8px top padding above the `tag` component
- 16px top padding above the `link` component

Screenshot 2023-10-13 at 11 56 40 AM

- [ ] 2. The link is too large in hierarchy compared to the rest of the pricing table content. It looks like this is using the Carbon for IBM.com link with icon, but in this case it should use the default Carbon link to result in a 14px type size. See the [CWC storybook example](https://web-components.carbondesignsystem.com/?path=/story/components-link--default). I would also suggest that links _within_ the pricing table such as these should be the `inline` link style instead of the `link with icon` 

Screenshot 2023-10-13 at 12 15 34 PM

packages/styles/scss/components/pricing-table/_pricing-table.scss
bruno-amorim: @jkaeser hey john, yes! You were right, I missed that I needed to disable the properties in another place. I've just submitted another commit removing those unnecessary properties.

jkaeser:
Hey, @oliviaflory! Thanks for the feedback. I asked Bruno to add more representative content in these places to replicate the conditions in which we were seeing a bug, but you're right: that content doesn't present so well. We'll replace it with something simpler that won't force us into a bigger discussion of what kind of content should and shouldn't be put in these cells.

bruno-amorim:
Hi @oliviaflory 
Please refer to John's last message.
I also have added some more padding so the items look better - https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/11002/index.html?path=/story/experimental-pricing-table--default
Would you please review again?

cc @jkaeser

packages/web-components/src/components/pricing-table/__stories__/pricing-table.stories.scss
```diff
@@ -6,3 +6,8 @@
 //
 
 @import '@carbon/ibmdotcom-styles/scss/globals/vars';
+
+span {
+  margin-top: 1rem;
+  display: block;
+}
```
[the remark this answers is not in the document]
bruno-amorim: Hi @jkaeser Ready for your review

Thanks!

oliviaflory (approved):
@bruno-amorim @jkaeser I apologize for the delayed review, I was out sick last week. 

The multiple paragraphs look good, thank you for cleaning up the examples and removing the tags from the inner rows & columns. 

One thing I will point out is that the checkmarks were not originally intended to be used in tandem with text so that the user's eye doesn't have to zig-zag down the page. 

**Current build**
Screenshot 2023-10-30 at 12 06 47 PM

**Intended use**
Screenshot 2023-10-30 at 12 01 43 PM

I am going to approve this so that you can start using the multi-paragraph, but would really prefer to not showcase the checkmark and text together in the storybook examples. If you could make an update to the storybook preview I would appreciate it!

andy-blum:
Please re-target this PR to the `v1` branch

PR #11003, opened 2023-10-04
fix(styles): update var interpolation causing warnings

kennylam:
### Description

Update sass variable interpolation causing compilation warnings

```
WARNING: math.div() will only support number arguments in a future release.
Use list.slash() instead for a slash separator.

math.div(100%, math.div(12, 2)) - math.div(#{$spacing-01}, 2)
```

### Changelog

**New**

- {{new thing}}

**Changed**

- update SASS vars used in `math.div`
- update `autoprefixer` browserlist

**Removed**

- {{removed thing}}


emyarod (approved):
looks good to me!

PR #11006, opened 2023-10-04
fix(carousel): update v2 styles and fix broken video cards

### LINKED ISSUE #10726: [Component] Carousel

emyarod:
### Related Ticket(s)

Closes #10726

### Description

This PR updates the carousel page size to include the proper number of cards at all breakpoints, and resolves a render issue with video cards that was breaking two component stories

### Changelog

**Changed**

- xlg carousel page size
- video card cta thumbnail render logic

### DIFF
--- a/docs/dotcom-v2-migration.md
+++ b/docs/dotcom-v2-migration.md
@@ -3,89 +3,95 @@
 This document will review in detail the changes in Carbon for IBM.com v1 to v2.
 
 ## Global changes
-Carbon for IBM.com v2 uses Carbon v11, which utilizes [Dart Sass](https://sass-lang.com/dart-sass) (`sass`), as opposed to
+
+Carbon for IBM.com v2 uses Carbon v11, which utilizes
+[Dart Sass](https://sass-lang.com/dart-sass) (`sass`), as opposed to
 `node-sass`. This introduces some new Sass features such as
-[`@use`](https://sass-lang.com/documentation/at-rules/use) and [`@forward`](https://sass-lang.com/documentation/at-rules/forward) keywords ([`@import`](https://sass-lang.com/documentation/at-rules/import) is no longer used), and
+[`@use`](https://sass-lang.com/documentation/at-rules/use) and
+[`@forward`](https://sass-lang.com/documentation/at-rules/forward) keywords
+([`@import`](https://sass-lang.com/documentation/at-rules/import) is no longer
+used), and
 [namespaces](https://sass-lang.com/documentation/at-rules/use#choosing-a-namespace).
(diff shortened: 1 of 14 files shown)


oliviaflory:
Hey @emyarod this is looking pretty good! Just a few comments:

- [ ] 1. **Video cards CTA style** I am expecting for the video cards to use the new play icon instead of the arrow icon

| Current | Expecting |
| --- | --- |
| Screenshot 2023-10-13 at 7 33 48 AM | Screenshot 2023-10-13 at 7 34 16 AM |

- [ ] 2. **Video missing heading** The video card is missing a heading in V2 compared to v1

| Current | Expecting |
| --- | --- |
| Screenshot 2023-10-13 at 7 36 23 AM | Screenshot 2023-10-13 at 7 36 00 AM |

- [ ] 3. **Mixed story varying** I expected the image and video cards to match based on the same height utility. Part of this could be caused by the video not pulling in any heading. I would expect both medias items to be the same aspect ratio and the card bodys to be the same. Plus I would expect the video cards to have the CTA pinned to the bottom. 

| Current | Expecting |
| --- | --- |
| Screenshot 2023-10-13 at 7 20 48 AM | Screenshot 2023-10-13 at 7 21 03 AM |


- [ ] 4. **Large breakpoint grid alignment** in the large breakpoint (1056–1311 px) the cards switch to 3 across instead of 4 across, which is different from the v1 story but I'm curious if you had to do that because of the aspect ratios we applied to the cards? Let me know!
Screenshot 2023-10-13 at 7 25 04 AM

emyarod:
- the video card icon should be updated now (point 1)
- I'm not sure why the video titles are no longer being pulled in, will need to dive into that more (points 2 and 3)
- and for point 4 I made that change because I thought it was a natural progression from 2 cards per page on `@md` to 4 cards per page on `@xlg` but I've changed it to be 4 cards per page on `@lg`

ariellalgilmore:
The carousel should work now! When switching in-between knobs in storybook it might look odd, but if you open the story in a new window you should see the correct heights

emyarod:
thanks @ariellalgilmore, migrating the missing parts of the card CTA `renderHeading` was a major thing I overlooked! I can also replicate this locally, but it looks like it may need to run with component updates? because switching between story knobs seems to break card rendering

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/8265238/4b89eca9-f351-4d9a-a2be-1a3d19abc7ea

IgnacioBecerra (approved):
I think this LGTM!

RichKummer (approved):
Seeing all the changes in, LGTM 🚀

PR #11008, opened 2023-10-05
feat(link-list): upgrade to v2 styles

### LINKED ISSUE #10789: [Component] Link list
Image

### Updated default & end of section

Image

### Design changes
- [x] Headings always optional in all variants

**Default variant**
- [ ] Not dependent on card (if it was?) No longer styled like a card
- [ ] Container size change
- [ ] Icon alignment change
- [ ] Interactive state changes (hover & focus)
- [ ] More similar to end of section variant

**Horizontal & vertical**
- [ ] No icon left alignment option

**End of section**
- [ ] End of section could merge with default variant? They will behave very similarly, but end of section would span wider on the grid.
- [ ] Interactive state changes (hover & focus)

[Figma spec](https://www.figma.com/file/0p2umX6w4LdSxfRZvJuy2O/Link-list?type=design&node-id=101%3A8427&mode=design&t=6c6PfElFpcfqmD4y-1)

kennylam:
### Related Ticket(s)

Closes #10789

### Description

Update `link-list` styles and tokens to Carbon v11

### Changelog

**New**

- {{new thing}}

**Changed**

- Use Carbon v11 styles/tokens
- move web component-specific styles to `styles` package

**Removed**

- `dds-link-list-card`
- `dds-link-list-card-cta`
- `end of section` variant
- `link-list-section` story


oliviaflory:
@kennylam looking good! A few comments:

- [ ] 1. Is it possible to remove the **CTA type: none** option? I realize we have it in prod now, but it seems like this might have been a bug that slipped by because it has a slight shift in the default & end of section link lists and no change in the horizontal & vertical variants. Also, **None** and **Local** have the same icon usage, so they seem redundant? We could possibly hide the icon if we want to keep the "none" option.
Screenshot 2023-10-12 at 1 32 08 PM
Screenshot 2023-10-12 at 1 32 19 PM

---

-  [ ] 2. Default link list hover state is a bit off: the underline is a different color than the text somehow and the icon isn't changing color. Everything should be `$link-primary-hover`
Screenshot 2023-10-12 at 1 37 42 PM

---

- [ ] 3. Default link list click state is a bit off: expecting the text and icon to be `$text-primary` on click.
Screenshot 2023-10-12 at 1 33 01 PM

---

- [ ] 4. I may have not made this clear enough in the specs, can the default link list be updated to have the icon top aligned to the container, and we want to have the container be 48px tall like the end of section variant. 
Also the heading, rules, and link text should all align.
Essentially _default_ and _end of section_ are styled the same – we could merge them if you want? But I'm also ok keeping them separate and possibly merging in v3 to make it easier on adopters.

Screenshot 2023-10-12 at 1 30 43 PM

**Example of expected default link list**
Screenshot 2023-10-12 at 1 46 04 PM

---

- [ ] 5. Is it possible for end of section icons to be pinned to the right? or is that part of the problem with taking it out of a card? 🙈 
Screenshot 2023-10-12 at 1 36 04 PM

packages/web-components/src/components/link-list-section/link-list-section.scss
```diff
@@ -1,32 +1,11 @@
...
+/**
+ * @license
+ *
+ * Copyright IBM Corp. 2020, 2022
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+// @use '@carbon/ibmdotcom-styles/scss/components/link-list-section';
```
IgnacioBecerra, line 10:
```suggestion
```

packages/web-components/src/components/link-list/__stories__/link-list.stories.ts
```diff
@@ -276,70 +288,70 @@ Vertical.story = {
...
+//             href="${ifDefined(href)}"
+//             cta-type="${ifDefined(ctaType)}"
+//             download="${ifDefined(download)}">
+//             ${ctaType !== CTA_TYPE.VIDEO
+//               ? html` Microservices and containers `
+//               : null}
+//           </c4d-link-list-item-cta>
+//         </c4d-link-list>
+//       `;
+// };
```
IgnacioBecerra, line 337:
```suggestion
```
I think we can delete this

packages/web-components/src/components/link-list/__stories__/link-list.stories.ts
```diff
@@ -276,70 +288,70 @@ Vertical.story = {
...
+//       default: {
+//         LinkListItem: {
+//           ctaType: null,
+//           download: undefined,
+//           href: 'https://www.example.com',
+//         },
+//       },
+//     },
+//   },
+// };
```
IgnacioBecerra, line 354:
```suggestion
```

This too

packages/styles/scss/components/card/_card.scss
```diff
@@ -166,6 +166,17 @@
     }
   }
 
+  :host(#{$c4d-prefix}-link-list-item-card),
+  :host(#{$c4d-prefix}-link-list-item-card-cta) {
```
IgnacioBecerra, line 170:
```suggestion
  :host(#{$c4d-prefix}-link-list-item-card) {
```
Since `card-cta` has been changed

IgnacioBecerra (commented):
Looking good! Just a couple things:

kennylam:
> Looking good! Just a couple things:

Doh, totally forgot to remove those!

IgnacioBecerra (approved):
LGTM!

kennylam:
> I was looking at Percy since we said no one looks and so many files changed, a few questions:

1. The heading isn't a property or attribute, so it can't really be marked as required or optional in that sense. I think the component page on the website would be the best place to add this guidance.
2. Fixed! I needed to update the `leadspace-block` story
3. Those styles were available to `link-list-section` only and were removed since it was deprecated. Do we want to bring it back?

oliviaflory (approved):
lovely! Thanks @kennylam 🎉 👏

PR #11018, opened 2023-10-09
fix(card): heading/copy spacing

ariellalgilmore:
### Related Ticket(s)

#10047 

### Description

Card heading/copy spacing
Screenshot 2023-10-09 at 12 33 01 PM

before there was additional margin top spacing on the copy

### Changelog

**New**

- {{new thing}}

**Changed**

- added a `pictogram` attribute to be added/removed if `_hasPictogram`, since `pictogram-placement` is always reflected and the `pictogram card` has different spacing for the heading/copy
- updated `color-scheme`: `inverse` border
- updated hovering to show on the image and video-poster

**Removed**

- {{removed thing}}


RichKummer (commented):
@ariellalgilmore – the spacing seems resolved, but in `@sm` it looks like the bottom spacing is off as the screen size expands. The bottom spacing should be consistent at `16px` – if the size changes, that bottom content should stay with the bottom until the spacing would be `<64px` from the headline. 

Screenshot 2023-10-09 at 4 32 11 PM

ariellalgilmore:
@RichKummer ~ updated!

RichKummer:
@ariellalgilmore seeing the fix for the footer which looks good. I'm not 100% if this is a storybook issue or not, but when I open `card static` in a new tab and set to `@sm` or `@md`, there seems to be extra spacing between the tags and the link (it should be `32px`). The body copy never comes in either.

Weirdly, neither of these are an issue when viewing in the Storybook canvas. Can you double check that this is working fine?

**In Storybook canvas:**
Screenshot 2023-10-10 at 1 31 02 PM

**In a new tab:**
Screenshot 2023-10-10 at 1 29 09 PM

**From the design specs:**
Screenshot 2023-10-10 at 1 29 34 PM

ariellalgilmore:
@RichKummer should be fixed now! the copy thing is a weird storybook knob issue :(

RichKummer (approved):
LGTM! Thanks @ariellalgilmore ✨

Issue #10636, opened 2023-07-06
[lightbox-media-viewer]: CDN link returns 404


andy-blum:
JIRA: https://jsw.ibm.com/browse/ADCMS-3976

### Description

[This codepen](https://codepen.io/andy-blum/pen/rNQmYvZ) is attempting to load all C4IBM web components from their CDN links. Viewing the network tab, we can see that the dds-lightbox-media-viewer is 404ing.

[lightbox-media-viewer canary docs](https://carbon-design-system.github.io/carbon-for-ibm-dotcom/canary/web-components/?path=/docs/components-lightbox-media-viewer--default#lightbox-media-viewer)

[1.33.0 CDN link](https://1.www.s81c.com/common/carbon-for-ibm-dotcom/version/v1.33.0/lightbox-media-viewer.min.js)
[Latest CDN link](https://1.www.s81c.com/common/carbon-for-ibm-dotcom/tag/v1/latest/lightbox-media-viewer.min.js)

---

Update 9/20/2023: There are three components defined in C4IBM that appear to be unreachable via CDN imports:

- `<dds-lightbox-media-viewer>`
- `<dds-left-nav-item>`
- `<dds-quote-link-with-icon>`

kennylam:
@andy-blum There is no CDN file for `lightbox-media-player` (not sure why, actually). You need to use 

```
https://1.www.s81c.com/common/carbon-for-ibm-dotcom/tag/v1/latest/lightbox-video-player.min.js
```

Updating your pen with this URI works. You can also see an example of it [here](https://stackblitz.com/github/carbon-design-system/carbon-for-ibm-dotcom/tree/main/packages/web-components/examples/codesandbox/components/lightbox-media-viewer). Check the CDN example directly [here](https://qgnilbjgbgithub-kstn--9000--e809191e.local-credentialless.webcontainer.io/cdn.html).

m4olivei:
> @andy-blum There is no CDN file for lightbox-media-player (not sure why, actually). 

[...]

PR #11020, opened 2023-10-10
fix(quote): make quote-link-with-icon available in quote cdn bundle

m4olivei:
### Related Ticket(s)

Relates to https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/10636

### Description

The `<dds-quote-link-with-icon>` component was not included with the main `<dds-quote>` component, leading to a broken experience were a user to attempt to use it with the main quote CDN bundle like the Story example does, for example:

https://codepen.io/m4olivei/pen/jOXdvjK

It looks like there is just a missing import in the `quote/index.ts` file. We add that here and update docs and Sandbox example to use the `<dds-quote-link-with-icon>` component.

### Changelog

**Changed**

- Import `<dds-quote-link-with-icon>` from the main `<dds-quote>` index.ts


packages/web-components/src/components/quote/index.ts
```diff
@@ -12,3 +12,4 @@ import './quote-source-bottom-copy';
 import './quote-source-copy';
 import './quote-source-heading';
 import '../link-with-icon/link-with-icon';
```
andy-blum, line 14: Is this element import still needed here since `quote-link-with-icon` is extending it?

m4olivei: Ahh yeah quite right. For a second I thought we still needed it for backward compatibility, but we still get it by way ot `quote-link-with-icon`. Will fix.

jkaeser (approved):
Looks like a good change to me. Not seeing any visual regressions when compared with the latest release.

An argument could be made that the `<dds-quote-link-with-icon>` component isn't strictly necessary... looks like it exists only to have a `display: inline-block;` CSS rule set on it. But that's maybe a conversation for v2 😏

emyarod (approved):
looks good to me!

PR #11028, opened 2023-10-11
fix(footer): fix grid alignment when disabling locale

annawen1:
### Related Ticket(s)

Closes # N/A

### Description

Noticed the grid alignment of the footer when disabling the locale-button was off, so this PR fixes that.

### Changelog

**Changed**

- adjust `flex-flow` property for when the locale button is disabled

**Removed**

- react classes for the footer that are no longer needed


IgnacioBecerra (approved):
LGTM!

Issue #10950, opened 2023-09-18
[RTL]: Carbon carousel RTL issue


marianserafimovibm:
[JIRA](https://jsw.ibm.com/browse/ADCMS-4075)

### Description

Hello
I'm working on replacing yourlearning.ibm.com home page carousel with carbon carousel.
Carbon carousel is set up, and it's working right, but we have a problem with RTL styling.
In documentation says that we only have to configure webpack to load rtl.css.js files, but it doesn't work for me. The styles are loaded to the browser.

**NOTE:** carousel pagination carets should be flipped horizontally

### Component(s) impacted

Here is the carousel we have to replace:
![first](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/145344451/5a8b7bfb-fe89-4de2-b509-11ef4a2e5c0c)
This is how it looks like by simply setting up carbon carousel. It works well for LTR version: 
![second](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/145344451/ad38acc9-b815-4b5c-b69a-ae3d5e0c8d43)
When RTL is enabled it looks like this:
![image](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/145344451/5c076c92-7443-438d-b5ad-fe7046f1f336)

Here are the lines of code we are using: 
![image](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/145344451/65f7c5af-faa4-4832-8e84-e34e56f79337)
![image](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/145344451/ccc9a3c0-bd4a-4d1e-a8ca-11ccd4ffce80)

### Browser

Chrome

### Carbon for IBM.com version

"@carbon/ibmdotcom-web-components": "^1.33.0",

### Severity

Severity 1 = The design is broken in a critical way that blocks users from completing tasks or damages the brand. Affects major functionality, no workaround.

### Application/website

yourlearning.ibm.com

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://www.ibm.com/standards/carbon/web-components/?path=/story/components-carousel--default

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
[...]

PR #11030, opened 2023-10-11
fix(carousel): carousel rtl fixes

m4olivei:
### Related Ticket(s)

Closes https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/10950
[Jira Ticket](https://jsw.ibm.com/browse/ADCMS-4075)

### Description

RTL fixes for the carousel.

### Changelog

**Changed**

- Changes carousel CSS to use CSS logical properties to better support RTL.

### Testing

- [ ] Open the [Web Components (RTL)](https://ibmdotcom-web-components-rtl.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/11030/index.html) build of Storybook
- [ ] Navigate to Carousel stories
- [ ] Verify the slide navigation arrows are correct
- [ ] Verify the slider works as expected in RTL
- [ ] Open the [Web Components](https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/11030/index.html) build of Storybook
- [ ] Regression test the carousel in LTR writing mode


packages/web-components/src/components/carousel/carousel.ts
```diff
@@ -722,7 +722,8 @@ class DDSCarousel extends HostListenerMixin(StableSelectorMixin(LitElement)) {
           )}">
           <div
             class="${scrollContentsClasses}"
-            style="left:${(-start * (contentsBaseWidth + gap)) / pageSize}px">
+            style="inset-inline-start:${(-start * (contentsBaseWidth + gap)) /
+            pageSize}px">
```
andy-blum, line 726: Setting styles inline like this has caused problems with content security policies in the past. Can we move these styles to the `updated` hook and set them with using the CSSOM?

andy-blum (changes requested):

andy-blum:
Comparing the LTR and RTL versions by executing `$$('*').forEach(el => el.dir = "rtl")` has shown a couple additional fixes needed:
1. The nav buttons need to be inline-end aligned
2. The nav buttons should be pointing in the direction of "more cards" (make sure correct button is disabled)
3. CTA icons need to point to inline-end (this might be out of scope for this ticket)

ltr
rtl

m4olivei:
> Comparing the LTR and RTL versions by executing `$$('*').forEach(el => el.dir = "rtl")` has shown a couple additional fixes needed:

I've merged @annawen1 's [fix for RTL](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/11031), so that we can just look at the [`Web Components (RTL)` deploy preview](https://ibmdotcom-web-components-rtl.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/11030/index.html) now. We're dependant on that now, which hopefully lands in the not too distant future.

> 1. The nav buttons need to be inline-end aligned

✅ 

> 2. The nav buttons should be pointing in the direction of "more cards" (make sure correct button is disabled)

✅ 

> 3. CTA icons need to point to inline-end (this might be out of scope for this ticket)

Yeah, I agree that it's out of scope. Issue is with the `<dds-card-footer>` component.

packages/styles/scss/components/carousel/_carousel.scss
```diff
@@ -92,36 +93,48 @@
...
+    block-size: $container-03;
+    inline-size: $container-03;
 
     svg {
       margin: auto;
     }
   }
+
+  .#{$prefix}--carousel__navigation__status {
+    direction: ltr;
```
jkaeser, line 111: Is there a reason we're forcing left-to-right here?

packages/web-components/src/components/carousel/carousel.ts
```diff
@@ -670,6 +670,31 @@ class DDSCarousel extends HostListenerMixin(StableSelectorMixin(LitElement)) {
...
+      return;
+    }
+
+    if (!this._contentsNode) {
+      return;
+    }
+
+    const startPosition =
+      (-this.start * (this._contentsBaseWidth + this._gap)) / this.pageSize;
+    this._contentsNode.style.insetInlineStart = `${startPosition}px`;
```
jkaeser, line 697: If you're up for some friendly debate, I'm not sure I understand the value of using the return early pattern here. IMHO it makes this a little less readable and more difficult to add more functionality into the updated hook in the future. Not a blocking comment 🙂

packages/styles/scss/components/carousel/_carousel.scss
```diff
@@ -133,8 +146,8 @@
     flex-flow: row wrap;
     margin: 0;
     padding: 0;
-    max-width: 100%;
-    width: 100%;
+    max-inline-size: 100%;
+    block-size: 100%;
```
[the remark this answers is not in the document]
m4olivei: Yep! Thanks for catching this. My brain still slips on the translation of width -> inline and height -> block

packages/styles/scss/components/carousel/_carousel.scss
m4olivei: Yes. This selector is for the newly created `<span>` that wraps the carousel status, which shows the ratio of current slide to total slides, eg. "1 / 5". Fractions like this should show the same regardless of language direction. At least, this is based on my light research.

> In-line fractions are written with the numerator and denominator on the left and right of the fraction slash respectively, e.g. ٢/٧ 2/7.

[Source](https://en.wikipedia.org/wiki/Modern_Arabic_mathematical_notation).

jkaeser: Oh, smart!

packages/web-components/src/components/carousel/carousel.ts
m4olivei: Always! A couple reasons I like early returns:

* Rather then writing a really long condition, with brackets and potentially a mix or logical OR and AND expressions, or multiple nested statements, I like breaking it up into multiple flat conditions. Totally subjective, but I think this way is easier to read.
* Broken up into multiple early returns, it's easier to document each. I was lazy here, but I've done that in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/11030/commits/64583dfc26d5004c54c97249007895e8e101f9fd
* I like getting the early returns out of the way so the meat of the function is as flat (minimal nesting) as possible.
* Salva Molina indoctrinated me to early returns 😛 

I agree with you about this being difficult to update in the future. I separated this concern into a private method so that `update` has room to grow if needed in future.

packages/web-components/src/components/carousel/carousel.ts
```diff
@@ -571,6 +571,30 @@ class DDSCarousel extends HostListenerMixin(StableSelectorMixin(LitElement)) {
...
+    // _contentsBaseWidth, _gap and pageSize. If none of them changed, we've
+    // got nothing to do.
+    if (
+      !changedProperties.has('start') &&
+      !changedProperties.has('_contentsBaseWidth') &&
+      !changedProperties.has('_gap') &&
+      !changedProperties.has('pageSize')
+    ) {
+      return;
+    }
```
jkaeser, line 585: > If none of them changed...

Does that mean these should be OR operators?

packages/web-components/src/components/carousel/carousel.ts
jkaeser: Love that compromise. Thanks, Matt!

packages/web-components/src/components/carousel/carousel.ts
m4olivei: Nope. Each operand is negated, so it reads "if start hasn't changed, and _contentBaseWidth hasn't changed and _gap hasn't change and pageSize hasn't changed then we've got nothing to do, return.

m4olivei:
Thanks for the review @jkaeser .

This is ready for review by the rest of you all. Please note that I merged in @annawen1 's PR to fix the Storybook RTL build, so this PR effectively depends on that one.

m4olivei:
Things we're getting messy since my branch had merged an older set of commits from @annawen1 's [PR](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/11031). Now that's merged to `main`, I decided to rebase my branch, and picked just my commits representing work on this branch. Much clearner commit history now.

kennylam (approved):
Thanks @m4olivei! Looks like you just need to update snapshots.

m4olivei:
@kennylam I'm not sure what to make of the job failures here. `main` is also in a failed state across many of the CI jobs, so I'm going to assume it's not me. For instance [`ci-check / check-packages` is failing](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/actions/runs/6891696009/job/18747318974?pr=11030) on `yarn install`. All is well on my local. 

Is this a known issue ATM?

kennylam:
@m4olivei Definitely not just you. We're looking into it now.

PR #11031, opened 2023-10-12
fix(storybook): set dir attribute in storybook for rtl

### LINKED ISSUE #10933: [rtl]: broken in dotcom v2 storybook
Storybook `RTL` job build fails.

annawen1:
### Related Ticket(s)

Closes #10933 

### Description

RTL is not working in storybook, the `dir` attribute doesn't get set. Looking at other storybook packages, the `dir` gets set in the storybook `preview.ts` file - implementing the same with the web components package.

### Changelog

**New**

- add check in `preview.ts` file for web components storybook that sets the `dir` attribute when supplied with the env var.

**Changed**

- bump up `postcss-loader` dependency and update config


packages/web-components/.storybook/main.js
```diff
@@ -222,15 +222,17 @@ module.exports = {
...
+              postcssOptions: {
+                plugins: () => {
+                  const hostPseudo =
+                    require('../tools/postcss-fix-host-pseudo')();
+                  const autoPrefixer = require('autoprefixer')({
+                    overrideBrowserslist: ['last 1 version', 'ie >= 11'],
+                  });
+                  return !useRtl
+                    ? [hostPseudo, autoPrefixer]
+                    : [rtlcss, hostPseudo, autoPrefixer];
```
m4olivei, line 234: Just curious, is there a good way to check that we're actually getting `rtlcss` post css processing?

m4olivei (approved):
Looks good to me, thanks!

packages/web-components/.storybook/main.js
annawen1: yeah there's [rtlcss directives ](https://rtlcss.com/learn/getting-started/why-rtlcss/#Processing-Directives) that we used to have...but I don't see them anymore - let me check

annawen1: I'll merge it in for now to unblock this issue, but this would require more investigation to see why rtlcss styles aren't coming through in storybook

PR #11044, opened 2023-10-17
fix(ui-shell): a11y issue with side nav item

### LINKED ISSUE #10918: [ui-shell]: IBM Accessibility violation error
### Description

The bx-side-nav-menu-item tag from the UI Shell component is throwing the following accessibility error:
Violation
```
The tabbable element's role 'none' is not a widget role
A tabbable element must have a valid widget role
```
It seems the problem can be fixed when I assign those tags the role="button".

### Component(s) impacted

UI Shell Component
bx-side-nav-menu-item tag

### Browser

Chrome, Microsoft Edge

v1.31.0

### Severity

### Application/website

Legal Entity Repository, can be found in storybook itself

### Package

@carbon/web-components

### CodeSandbox example

Don't have

1. Open storybook
2. Open the UI shell component
3. Open one of the menus with dropdowns from the example
4. Run the IBM Equal Access Accessibility Checker
5. Identify errors that say "The tabbable element's role 'none' is not a widget role"

_No response_

### Code of Conduct

sangeethababu9223:
### Related Ticket(s)

Closes #10918

### Description

The bx-side-nav-menu-item tag in UI Shell component throws accessibility error.
 
### Changelog

**Changed**

- Added role attribute value button to the element.

Screenshot 2023-10-17 at 3 51 06 PM

### DIFF
--- a/packages/carbon-web-components/src/components/ui-shell/ui-shell-story.ts
+++ b/packages/carbon-web-components/src/components/ui-shell/ui-shell-story.ts
@@ -134,38 +134,39 @@ export const sideNav = (args) => {
       ?expanded=${expanded}>
       <bx-side-nav-items>
         <bx-side-nav-menu title="L0 menu">
-          <bx-side-nav-menu-item href="${ifNonNull(href)}">
+          <bx-side-nav-menu-item href="${ifNonNull(href)}" role="button">
             L0 menu item
           </bx-side-nav-menu-item>
-          <bx-side-nav-menu-item href="${ifNonNull(href)}">
+          <bx-side-nav-menu-item href="${ifNonNull(href)}" role="button">
             L0 menu item
           </bx-side-nav-menu-item>
-          <bx-side-nav-menu-item href="${ifNonNull(href)}">
+          <bx-side-nav-menu-item href="${ifNonNull(href)}" role="button">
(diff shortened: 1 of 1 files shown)


ariellalgilmore (commented):
this looks good! But wondering if we can instead of adding it to the story add the `role="button"` to the `bx-side-nav-menu-item` directly like this: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/317d2bee88f8a704a56a4e10492a936fd29bdfe4/packages/carbon-web-components/src/components/tabs/tab.ts#L41C1-L45

sangeethababu9223:
> this looks good! But wondering if we can instead of adding it to the story add the `role="button"` to the `bx-side-nav-menu-item` directly like this: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/317d2bee88f8a704a56a4e10492a936fd29bdfe4/packages/carbon-web-components/src/components/tabs/tab.ts#L41C1-L45

Thanks @ariellalgilmore, I've updated the role as per your comment.

ariellalgilmore (approved):
Thanks @sangeethababu9223 !! just need to update snapshots here as well in `packages/web-components` since this will update the `masthead`

PR #11051, opened 2023-10-19
fix(card-w-pictogram): card with pictogram was not using a correct se…

### LINKED ISSUE #10960: [Card with pictogram does not open correct right-click menu]
[JIRA](https://jsw.ibm.com/browse/ADCMS-4074)

### Description

When you right-click on a "Card with pictogram", it opens the menu you would get for clicking on a blank part of the screen rather than the one you get when you right click on a link. As a result, users can't easily copy the link, open it in a new tab, etc. 
Carbon-pictogram

### Component(s) impacted

[Card with pictogram](https://www.ibm.com/standards/carbon/web-components/?path=/docs/components-card--pictogram)

### Browser

_No response_

v1.36.0

### Severity

### Application/website

https://www.ibm.com/community/qradar/

### Package

### CodeSandbox example

https://githubbox.com/carbon-design-system/carbon-for-ibm-dotcom/tree/main/packages/web-components/examples/codesandbox/components/card

Right click on the card.

_No response_

### Code of Conduct

bruno-amorim:
…mantic as an anchor element

### Related Ticket(s)

https://jsw.ibm.com/browse/ADCMS-4074

Closes #10960

### Description

When we right-clicked the Card with Pictogram component, it was not showing up options related to an anchor element. 
This was caused by the component semantics not using an anchor tag. 

Screenshot 2023-10-19 at 09 57 17

### Changelog

**New**

Since the Card with Pictogram component was built using a different logic than the other card variations, I added some conditions so that it's now built inside of an anchor element. 

Screenshot 2023-10-19 at 09 59 24


packages/web-components/src/components/card/card.ts
```diff
@@ -117,47 +117,72 @@ class DDSCard extends StableSelectorMixin(BXLink) {
...
+          </div>
+        </a>
+      `;
+    } else {
+      return html`
+        ${this._renderImage()}
+        <div
+          class="${prefix}--card__wrapper ${hasPictogram
+            ? `${prefix}--card__pictogram`
+            : ''} ${hasPictogram && hasCopy ? `${prefix}--card__motion` : ''}">
```
jkaeser, line 168: We know `hasPictogram` is false at point in the logical flow, so we can remove all the conditional logic that looks at that value here.

packages/web-components/src/components/card/card.ts
```diff
@@ -117,47 +117,72 @@ class DDSCard extends StableSelectorMixin(BXLink) {
...
+          <div class="${prefix}--card__content">
+            <slot name="eyebrow"></slot>
+            ${this.pictogramPlacement === PICTOGRAM_PLACEMENT.TOP
+              ? html`
+                  <slot
+                    name="pictogram"
+                    data-pictogram-placement="${PICTOGRAM_PLACEMENT.TOP}"
+                    @slotchange="${handleSlotChange}"></slot>
+                `
+              : ''}
```
jkaeser, line 178: I believe this whole thing can be removed since, like the previous comment, we know we don't have a pictogram at this point.

packages/web-components/src/components/card/card.ts
```diff
@@ -117,47 +117,72 @@ class DDSCard extends StableSelectorMixin(BXLink) {
...
+           ${hasCopy ? `${prefix}--card__motion` : ''}">
+          <div class="${prefix}--card__content">
+            ${this.pictogramPlacement === PICTOGRAM_PLACEMENT.TOP
+              ? html`
+                  <slot
+                    name="pictogram"
+                    data-pictogram-placement="${PICTOGRAM_PLACEMENT.TOP}"
+                    @slotchange="${handleSlotChange}"></slot>
+                `
+              : ''}
```
jkaeser, line 136: We're rendering almost exactly the same markup a few lines down. The only difference is the value of `data-pictogram-placement`. To me, this is a good opportunity to extract this chunk of HTML into it's own render method, something like this:
```js
/**
 * Renders the pictogram slot.
 */
protected _renderPictogram(placement) {
  const { _handleSlotChange: handleSlotChange } = this;
  return html`
    <slot
      name="pictogram"
      data-pictogram-placement="${placement}"
      @slotchange="${handleSlotChange}"></slot>
  `;
}
```
Then you can call that method here and down a few lines, passing the correct placement values as arguments.

This has a few benefits:
1. We're reducing what is essentially duplicate code.
2. We're making it easier for any components that might extend this one to change the way the pictogram slot is rendered.
3. In my opinion, the code becomes easier to read since we're rendering all the pieces of the markup in the same way.

packages/styles/scss/components/card/index.scss
```diff
@@ -684,7 +685,7 @@
         .#{$prefix}--card__copy {
           margin-top: $carbon--spacing-07;
           margin-bottom: 0;
-
+
```
jkaeser, line 688: Nit: can you remove this whitespace?

jkaeser (changes requested):
Hey, @bruno-amorim! Thanks for taking this one on. I really like your approach here, specifically getting rid of the onclick workaround and using native elements instead. 

Looks to me like a little more work is needed to polish this one up, though. The biggest indicator is that there are two focus stages:
1. One for the outermost `<div>` that has a `tabindex` set on it. This triggers the hover state text, but you can't activate to follow the link.
2. One for the new `<a>` tag, which you can activate to follow the link.

Ideally we just have one focusable element that both of those things.

I left a few more comments for cleaning up the code, too. As always, let me know if you have any questions.

kennylam:
@bruno-amorim Nice one! And thanks @jkaeser for the feedback. One thing to check will be screen reader a11y for this. If I remember correctly, some card components required some semantic changes because using native elements would make screen readers unnecessarily verbose. It was the suggestion of the a11y testers that resulted in these changes.

I'll ask for another quick test of this PR to see how it is.

packages/styles/scss/components/card/index.scss
```diff
@@ -638,7 +638,8 @@
     outline: none;
 
     .#{$prefix}--card__pictogram {
-      display: flex;
+      text-decoration: none;
+      color: $black-100;
```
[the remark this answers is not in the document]
bruno-amorim: 1. I fixed the 2 erroneous focus stages. Now we only have 1 focus that does the two actions required: it triggers the hover state and open the link on keyboard "enter/return" click
2. I have fixed the color property in the css file to now be **color: $black-100;** . I have added that piece of code because by wrapping the card around an anchor element made the text to be in blue and underlined

packages/web-components/src/components/card/card.ts
bruno-amorim: I have removed this piece of code. I also noticed that the state `protected _hasPictogram = false;` was ping ponging between true and false in the execution of the code. I have modified the method `_handleSlotChange` a bit to ensure that stops happening.

packages/web-components/src/components/card/card.ts
bruno-amorim: I have created the method `_renderPictogram` 
just a note here - I tried destructuring `handleSlotChange` right away at the beginning of the method, however, I couldn't access it, for some reason at that point of the code it was getting returned as null. I tried a lot of things but the only thing that worked was passing it as a parameter, and then passing an argument into the method down below for  `handleSlotChange`.

Please let me know of your thoughts, if there's something I'm not seeing here.

packages/styles/scss/components/card/index.scss
bruno-amorim: removed!

packages/web-components/src/components/card/card.ts
bruno-amorim: "but the only thing that worked was passing it as a parameter, and then passing an argument into the method down below for handleSlotChange."

 @jkaeser please disconsider this part. I've found what the bug was and it's fixed!

packages/web-components/src/components/card/card.ts
```diff
@@ -112,52 +128,52 @@ class DDSCard extends StableSelectorMixin(BXLink) {
...
+            ${this.pictogramPlacement !== PICTOGRAM_PLACEMENT.TOP ||
+            !hasPictogram
+              ? this._renderHeading()
+              : null}
+            ${this.pictogramPlacement === PICTOGRAM_PLACEMENT.BOTTOM ||
+            !hasPictogram
+              ? this._renderCopy()
+              : ''}
+            ${this.pictogramPlacement === PICTOGRAM_PLACEMENT.BOTTOM
+              ? this._renderPictogram('bottom')
```
jkaeser, line 154: Same thing here about using the enum.
```suggestion
              ? this._renderPictogram(PICTOGRAM_PLACEMENT.BOTTOM)
```

packages/web-components/src/components/card/card.ts
```diff
@@ -65,8 +65,10 @@ class DDSCard extends StableSelectorMixin(BXLink) {
   protected _handleSlotChange({ target }: Event) {
     const { name } = target as HTMLSlotElement;
     const hasContent = Boolean(this.querySelector('p'));
+    const hasPictogram = Boolean(this.querySelector('[slot="pictogram"]'));
```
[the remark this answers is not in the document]
bruno-amorim: Hi @jkaeser thanks for this tip. Yeah! It's way better to use that method. I have changed that to now use [assignedNodes()](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/assignedNodes)

packages/web-components/src/components/card/card.ts
```diff
@@ -98,6 +100,20 @@ class DDSCard extends StableSelectorMixin(BXLink) {
     `;
   }
 
+  /**
+   * renders the pictogram slot.
+   */
+  protected _renderPictogram(
+    placement: string
```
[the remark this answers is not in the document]
bruno-amorim: Done! Using Enum in both places

jkaeser (approved):

bruno-amorim:
Hey @kennylam 
Please let me know when you have that a11y test completed so we can address any possible issue.
Thanks a lot!

bruno-amorim:
Hey @m4olivei @IgnacioBecerra 
I have new resolved the conflicts in this PR

Thanks!

bruno-amorim:
Hi @jkaeser @m4olivei @andy-blum @marcelojcs 

Matt helped me clean up the commit history on this one by using rebase. Would you please review it again and if suitable provide your approvals.

Thanks!

packages/web-components/src/components/card/card.ts
```diff
@@ -65,8 +65,12 @@ class DDSCard extends StableSelectorMixin(BXLink) {
   protected _handleSlotChange({ target }: Event) {
     const { name } = target as HTMLSlotElement;
     const hasContent = Boolean(this.querySelector('p'));
+    const hasPictogram = Boolean(
+      (target as HTMLSlotElement).assignedNodes().length
+    );
     this[slotExistencePropertyNames[name]] = hasContent;
     this._hasCopy = hasContent;
+    this._hasPictogram = hasPictogram;
```
m4olivei, line 73: I'm finding that these changes are triggering an infinite rendering loop somehow and the browser becomes unresponsive. It's affecting any story that has a card in it 😕 . I'm pretty sure the culprit is here and has something to do with the fact that `_handleSlotChange` will fire for any slot changes amongst the default slot, `pictogram` slot, or `image` slot. There is always an initial `slotchange` event for every slot, and `hasPictogram` is being adjusted even when the slot change event is for `image` or the default slot. I found that adjusting things back to what they were and then appending `pictogram` specific code as follows fixes things:

```js
if (name === 'pictogram') {
  this._hasPictogram = Boolean(
    (target as HTMLSlotElement).assignedNodes().length
  );
}
```

For bonus points, I'd go further and suggest that we shouldn't have a shared `_handleSlotChange` event handler. There should be one per `<slot>` and it should be solely concerned with its relevant state variable (eg. `_hasPictogram` for the `pictogram` slot). I'd also get rid of the `slotExistencePropertyNames` abstraction, as it's hard to follow.

m4olivei (changes requested):
I found an issue with the slot change event that causes pages to become unresponsive. Otherwise, a couple of small suggestions.

m4olivei (approved):
Looks good to me. I did leave a small comment. I don't have a strong opinion on it though. Feel free to take it or leave it.

andy-blum:
To @kennylam's point, I think we have some a11y fixes needed here. I just ran through this component using Safari + VO and was unable to get VO to announce the content of the hover/focus text. This is because we're using a _class_ for a target of `aria-describedby`, which requires an _ID_.


Note that when the link has focus, the VO reads out the button's title (which isn't visible)
Screenshot 2023-11-13 at 1 28 05 PM

And that when the element doesn't have focus but is read by the VO cursor, we again get just the title.
Screenshot 2023-11-13 at 1 29 03 PM

andy-blum (changes requested):
We need to update the value of `aria-describedby` to target an ID attribute.

bruno-amorim:
Hi @andy-blum and @m4olivei 
I addressed both of your comments. Would you please review my last commit when you have some time?
Thanks!

andy-blum (changes requested):
ci-check job is failing - looks like you need to update the unit tests

`yarn test:unit:updateSnapshot` should be the command you want.

Also seems like the deploy previews are failing since one of the mergebacks, but the error seems like a problem on the jenkins side to me.

```
FATAL: command execution failed
java.nio.channels.ClosedChannelException
	at ...
Caused: java.io.IOException: Backing channel 'JNLP4-connect connection from 172.30.95.24/172.30.95.24:59754' is disconnected.
	at ...
FATAL: Unable to delete script file /tmp/jenkins9104133333703048549.sh
java.nio.channels.ClosedChannelException
	at ...
Caused: hudson.remoting.ChannelClosedException: Channel "hudson.remoting.Channel@244a3c90:JNLP4-connect connection from 172.30.95.24/172.30.95.24:59754": Remote call on JNLP4-connect connection from 172.30.95.24/172.30.95.24:59754 failed. The channel is closing down or has closed down
	at ...
Build step 'Execute shell' marked build as failure
Setting status of edc6966bcfe4e45a82578f89373ea263e42ddf1f to FAILURE with url https://dbg-digital-design-team-jenkins.swg-devops.com/job/Carbon%20for%20IBM.com%20-%20Deploy%20Preview%20-%20Web%20Components/9968/ and message: 'Build finished. '
Using context: Deploy Preview - Web Components
Finished: FAILURE
```

kennylam:
@bruno-amorim The QA team was able to test this component this week and it does in fact revert all the [a11y fixes](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/6642) applied to the `card` component a while back.

While the absence of right-click functionality is an annoyance, there would be too many a11y regressions re-introduced (no arrow navigation, wrong elements announced by screen reader, screen reader too verbose, etc.) to justify this change. I think a better approach would be to open a feature request to extend `cta` logic to this variation which would allow it to open links in new tabs. Thanks for all your work here!

Issue #9040, opened 2022-06-27
[Filter Panel]: Behavior


cedricpelle:
[JIRA](https://jsw.ibm.com/browse/ADCMS-4084)

### Description

Behavior  when selected and un-selected item , section is opened 

### Component(s) impacted

Filter Panel

### Browser

Chrome, Safari, Firefox, Microsoft Edge

### Carbon for IBM.com version

v1.19.0

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

https://www.ibm.com/support/customer/csol/terms/

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://www.ibm.com/standards/carbon/web-components/?path=/docs/components-filter-panel--default

### Steps to reproduce the issue (if applicable)

- select an item

![image](https://user-images.githubusercontent.com/38075847/175980167-b2942694-8eb2-4515-844a-ed718e1def9b.png)

-  remove and close the section
![image](https://user-images.githubusercontent.com/38075847/175980282-f910c1da-c46e-4349-a143-be127ce07680.png)

- select an item into the second section
![image](https://user-images.githubusercontent.com/38075847/175980372-876011ca-94d0-42e8-b744-2bffd451d42c.png)

==> and the component will reopen the first section  (even nothing is selected )

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #11057, opened 2023-10-20
[Filter panel]: Reduce Filter Panel DOM node counts

### LINKED ISSUE #10669: Filter panel

m4olivei:
### Related Ticket(s)

Closes #10669 
Closes https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/9040
[Jira ADCMS-4089](https://jsw.ibm.com/browse/ADCMS-4089)
[Jira ADCMS-4084](https://jsw.ibm.com/browse/ADCMS-4084)

### Description

Reduces DOM node counts for the filter panel by only rendering the mobile or destop experience according to viewport. In order to acheive this, does a bit of refactoring how the heading is handled. Before the desktop and mobile layouts were coupled by only the mobile layout rendering the heading slot.

### Changelog

**Changed**

- Filter panel only renders desktop / mobile layout per the viewport

### Testing instructions

- [ ] Regression testing the Filter Panel
- [ ] Inspect the DOM. Only markup for desktop or mobile should be present, never both.


packages/web-components/src/components/filter-panel/filter-panel-composite.ts
```diff
@@ -412,25 +426,32 @@ class DDSFilterPanelComposite extends HostListenerMixin(
...
+        name="heading"
+        @slotchange="${this._handleHeadingSlotChange}"></slot>
+      ${this._isMobile
+        ? html`
+            <button class="bx--filter-button" @click=${this._openModal}>
+              <div class="${prefix}--filter__modal__button">
+                ${this._getComposedHeadingFilterCount()} ${Filter()}
+              </div>
+            </button>
+            ${this._renderModal()}
```
jkaeser, line 452: Nit: I wonder if it makes sense to put this all in a `protected _renderMobile()` method, just so things match. Could make this method really tidy:
```js
return html`
  <slot>...</slot>
  ${this._isMobile ? this._renderMobile() : this._renderDesktop()}
`
```

packages/web-components/src/components/filter-panel/filter-panel-composite.ts
```diff
@@ -412,25 +426,32 @@ class DDSFilterPanelComposite extends HostListenerMixin(
...
+      <slot @slotchange="${this._handleSlotChange}"></slot>
     </dds-filter-panel>
   `;
 
   render() {
+    // Note that the <slot name="heading"> contents, intended to be
+    // <dds-filter-panel-heading> are never shown as is. The text contents
+    // are composed, using this._getComposedHeadingFilterCount(), together with
+    // the current filter count, and passed as an attribute to
+    // <dds-filter-panel-modal> and <dds-filter-panel>.
```
jkaeser, line 440: Thinking out loud, are we set up to essentially deprecate the heading slot altogether? We're doing a lot of work to simply harvest the text content of the first element in this slot, and perhaps this is a scenario where it makes more sense to have a `heading` attribute. Users set it, it gets reflected into a property we access to compose the heading + filter count.

jkaeser (changes requested):
Thanks for taking on this beast, @m4olivei. This is a tough component to work on, but all your code changes look good to me, and I'm glad to see the MediaQueryMixin getting some use 😁 Unfortunately, I noticed the filter count doesn't always stay consistent when transitioning across the breakpoint. Steps to reproduce:
1. At desktop, select the "API" checkbox from the first filter group, then select "Analytics" from the second filter group. Note the reported number of active items should be 2.
2. Shrink the viewport until you cross into the mobile breakpoint. Note the filter number has decreased to 1, despite both items still being selected.
3. Deselect "Analytics" and note the number has gone up again to 2.
4. Reselect "Analytics" and note the number has gone down to 2.
5. Cross back into the desktop breakpoint and note the number has gone down to 2 despite both items remaining selected.

packages/web-components/src/components/filter-panel/filter-panel-composite.ts
m4olivei: Totally agree. I only kept the slot b/c getting rid of it would break compatibility. Are you suggesting that we build in support for a `heading` attribute here and mark the slot and `<dds-filter-panel-heading>` as deprecated? I could get behind that.

m4olivei:
> Thanks for taking on this beast, @m4olivei. This is a tough component to work on, but all your code changes look good to me, and I'm glad to see the MediaQueryMixin getting some use 😁 Unfortunately, I noticed the filter count doesn't always stay consistent when transitioning across the breakpoint. Steps to reproduce:

Nice find! Turns out there was a bug living in the slot change event, which had been covered up by some other code. It wasn't counting the select / select-item elements properly. Corrected in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/11057/commits/dd4cba610769927f218cf3731cce16ca8542943b

packages/web-components/src/components/filter-panel/filter-panel-composite.ts
jkaeser: Yes, exactly! Let's do it 💪

jkaeser (approved):

packages/web-components/src/components/filter-panel/filter-panel-composite.ts
m4olivei: Cool, I'll file a separate issue / PR pair for that to follow this up. Feels worth the paper trail.

m4olivei: Filed: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/11062

kennylam (approved):
LGTM thanks @m4olivei!

PR #11071, opened 2023-10-27
fix(link-with-icon): refactor LINK_SIZE enum

andy-blum:
### Related Ticket(s)

[ADCMS-3542](https://jsw.ibm.com/browse/ADCMS-3542)

### Description
We've been digging into a sticky bug that we think we finally have an explanation for. To see the problem in action see [this codepen](https://codepen.io/jakaeser/pen/qBMdgKg).

As we understand it, this issue is:
* originating in link.ts with [the LINK_SIZE enum](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/carbon-web-components/src/components/link/link.ts#L23-L38).
* Then we use that enum to [set a default value](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/link-with-icon/link-with-icon.ts#L47-L53) in link-with-icon.
* Then we [extend link-with-icon to card-footer](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/card/card-footer.ts#L29)
* Then we [extend card-footer to card-cta-footer](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/cta/card-cta-footer.ts#L33)
* Then we [extend DDSCardCTAFooter to CAEMCardCTAFooter](https://github.ibm.com/webstandards/carbon-for-aem/blob/main/src/components/card/card-cta-footer.ts#L16)

There are lots of other places in the code base that we use ENUMs that don’t cause issues, but this one does and as far as we can tell it’s the only one that:
* is defined in CWC
* is used to set a default value in C4IBM

### Changelog

**Changed**

- Refactor `LINK_SIZE` from enum to a union


packages/web-components/src/components/link-with-icon/link-with-icon.ts
```diff
@@ -50,7 +50,7 @@ class DDSLinkWithIcon extends StableSelectorMixin(BXLink) {
    * @internal
    */
   @property()
-  size = LINK_SIZE.LARGE;
+  size = 'lg' as LINK_SIZE;
```
jkaeser, line 53: Hmm, having to cast this string seems like a smell. Does the solution from https://dev.to/ivanzm123/dont-use-enums-in-typescript-they-are-very-dangerous-57bh obviate the need for this?

andy-blum: Updated to this method.

I had originally gone with the union instead to avoid having to import both an object and a type, but it seems that the type cascades through the inheritance chain on its own just fine, and if the object keeps the same name as the enum, we can make future refactors simpler.

packages/carbon-web-components/src/components/link/link-story.ts
```diff
@@ -16,13 +16,12 @@ import { boolean, select } from '@storybook/addon-knobs';
...
 import ifNonNull from '../../globals/directives/if-non-null';
-import { LINK_SIZE } from './link';
 import storyDocs from './link-story.mdx';
 
 const sizes = {
   'Regular size': null,
-  [`Small size (${LINK_SIZE.SMALL})`]: LINK_SIZE.SMALL,
-  [`Large size (${LINK_SIZE.LARGE})`]: LINK_SIZE.LARGE,
+  [`Small size (sm)`]: 'sm',
+  [`Large size (lg)`]: 'lg',
```
jkaeser, line 24: Given https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/11071#discussion_r1376518961, what do you think about reverting this?

packages/carbon-web-components/src/components/link/link.ts
```diff
@@ -190,7 +182,7 @@ class BXLink extends FocusMixin(LitElement) {
    * Link size.
    */
   @property({ reflect: true })
-  size = LINK_SIZE.REGULAR;
+  size: LINK_SIZE_TYPE = '';
```
jkaeser, line 185: Same question here – can you go back to setting this to `LINK_SIZE.REGULAR`?

jkaeser (approved):

IgnacioBecerra (approved):

PR #11085, opened 2023-11-01
feat(masthead-v2): Minimize CMApp on megamenu open

jkaeser:
### Related Ticket(s)

[Jira](https://jsw.ibm.com/browse/ADCMS-4152)

### Description

Using the Contact Module's `init` method as a workaround to minimize it has the potential to completely break its opening & closing. In a discussion with folks from that team, we've agreed to call the `minimize` method, which is new since our workaround was put in place.

### Changelog

**New**

- Minimizes the Contact Module application when opening a megamenu.

**Removed**

- Clicking the contact link in the masthead when the Contact Module is open no longer closes the module.


jkaeser:
**Note:** This doesn't fully work yet. Calling the `minimize` method somehow seems to trigger the Lit lifecycle, resulting in  megamenus closing rapidly after opening if the Contact Module was previously opened.

andy-blum:
> Calling the minimize method somehow seems to trigger the Lit lifecycle

I wonder if the minimize method of the CM_APP is moving focus somewhere

andy-blum:
Confirmed. Until this behavior is fixed from the contact module, this PR is unmergable.

screenshot_2023-11-01_at_4 49 00___pm

m4olivei (dismissed):
I think we can simplify the event handling by using the `@HostListener` decorator with the `document:` prefix for the event name. See comment for details.

andy-blum (approved):

andy-blum:
Verified with @jkaeser and @m4olivei on a call since testing requires network request modifications.

m4olivei (commented):
LGTM! Glad the arrow function approach worked.

PR #11107, opened 2023-11-07
chore(docs): update v2 migration docs

kennylam:
### Description

Update v2 migration docs to be a bit more consistent in wording, formatting, etc. This is not final, just a bit of a clean up.

### Changelog

**New**

- {{new thing}}

**Changed**

- {{changed thing}}

**Removed**

- {{removed thing}}


docs/dotcom-v2-migration.md
```diff
@@ -1,125 +1,124 @@
...
 
-## List of component changes
+## Components
 
 | Component                 | Changes                                                                                                                                        |
 | ------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------- |
 | back-to-top               | No API changes.                                                                                                                                |
 | background-media          | No API changes.                                                                                                                                |
-| button                    | This component is deprecated in v2 in favor for `button-expressive`.                                                                           |
+| button                    | Deprecated in favor of `button-expressive`.                                                                           |
```
ariellalgilmore, line 19: didn't we end up actually keeping button?

ariellalgilmore (approved):
Small comment, but LGTM otherwise!

docs/dotcom-v2-migration.md
kennylam: I didn't work on this so I'm not 100% sure. But I just spoke with @oliviaflory about it and at the very least it seems like the components are swapped (it should be `button-expressive` deprecated in favor of `button`).

ariellalgilmore: ahh yup: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/feat/carbon-for-ibm-dotcom-v2/packages/web-components/src/components/button/button-expressive.ts#L252

kennylam: Ok, got it – `button-expressive` is getting deprecated. Instead, all button usage should come from `@carbon/web-components`.

kennylam: Right.

PR #11111, opened 2023-11-08
fix(Leadspace): dds-leadspace styling of slotted button group

marcelojcs:
### Related Ticket(s)

Closes # https://jsw.ibm.com/browse/ADCMS-4220

### Description

We're mixing and matching DDS and CAEM components and in this case the styles of the dds-leadspace are not able to target the caem-button-group. adding a new selector to Leadspace Styles to address that. 

### Changelog

**Changed**

- added an new selector to packages/web-components/src/components/leadspace/leadspace.scss


packages/styles/scss/components/leadspace/_leadspace.scss
```diff
@@ -246,6 +246,7 @@ $btn-min-width: 26;
 
       ::slotted(#{$dds-prefix}-button-group),
       ::slotted([slot='action']),
+      ::slotted([data-autoid$='button-group']),
```
jkaeser, line 249: Clever! I think this strikes a nice balance between being fuzzy enough to select the appropriate elements, but specific enough that unwanted elements won't inadvertently pick up the styles.

jkaeser (changes requested):
@marcelojcs I really like the fix, but I'm not sure you've covered all the selectors needed to fully resolve the issue. I think there's at least one other selector that targets `dds-button-group` elements you need to apply the same fix to.

packages/styles/scss/components/leadspace/_leadspace.scss
marcelojcs: Thank you so much @jkaeser!!!!
I have addressed the other selector in the latest commit. 
No other occurrences were found within the same folder.

andy-blum: Agreed - especially the use of `$=` to make sure we don't accidentally select `button-group-item` or similar.

What are your thoughts on adding one more selector - `::slotted([slot="action"].#{$prefix}--buttongroup`. This way items that are slotted there other than `<*-button-group>` have the option to opt-in to the padding rules?

jkaeser (approved):
Looks good to me! Just make sure you see Andy's comment – he has an idea for an easy enhancement.

m4olivei:
Looks like all comments have been addressed and @jkaeser left an approval. @marcelojcs is this ready to be marked "Ready for Review" (currently marked as Draft)?

andy-blum:
@oliviaflory / @RichKummer any chance we can get some design review on this ticket?

oliviaflory (approved):
this looks fine 👍

PR #11145, opened 2023-11-14
fix(video-player): v2 alignment

ariellalgilmore:
### Related Ticket(s)

none

### Description

looks like some v2 changes accidentally effected the video-player alignment. This should bring it back centered

before: 
Screenshot 2023-11-14 at 9 32 39 AM
after:
Screenshot 2023-11-14 at 9 38 17 AM

### Changelog

**Changed**

- set padding to 0


kennylam (approved):
Good catch!

PR #11189, opened 2023-12-04
chore(icons): use js icons instead of svg

kennylam:
### Description

This updates web component templates to use JS versions of IBM logo icons instead of the direct SVG. Some frameworks, such as NextJS, can throw errors when using `unsafeSVG()`.

### Changelog

**New**

- {{new thing}}

**Changed**

- change IBM logos to use es modules instead of SVG
- update logo styles in `footer - micro`

**Removed**

- {{removed thing}}

### DIFF
--- a/packages/styles/scss/components/footer/_footer-logo.scss
+++ b/packages/styles/scss/components/footer/_footer-logo.scss
@@ -64,12 +64,19 @@
   }
 
   :host(#{$c4d-prefix}-footer-logo[size='micro']) {
+    position: relative;
+
+    .#{$c4d-prefix}--footer-logo__link {
+      position: absolute;
+      top: -37px;
+
+      @include breakpoint(lg) {
+        position: relative;
+        top: 0;
+      }
+    }
+
     align-self: center;
     margin: 0;
-
-    @include breakpoint-between(md, lg) {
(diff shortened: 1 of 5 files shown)


packages/styles/scss/components/footer/_footer-logo.scss
```diff
@@ -64,12 +64,19 @@
...
   :host(#{$c4d-prefix}-footer-logo[size='micro']) {
-    align-self: center;
-    margin: 0;
+    position: relative;
 
-    @include breakpoint-between(md, lg) {
+    .#{$c4d-prefix}--footer-logo__link {
       position: absolute;
-      bottom: to-rem(58.5px);
+      top: -37px;
```
IgnacioBecerra, line 71:
```suggestion
      top: to-rem(-37px);
```
We could use the `to-rem` function here too

IgnacioBecerra (approved):
Just this little thing, but otherwise LGTM

packages/styles/scss/components/footer/_footer-logo.scss
kennylam: I tried this but the mixin did not like the negative value so I left it like this.

PR #11081, opened 2023-10-31
feat(feature-flags): implement feature flags for cwc

annawen1:
### Related Ticket(s)

Closes #11048

### Description

Add file to house feature flags and set up storybook to hide experimental components from being displayed. 

Once the feature flag is defined in `feature-flags.ts`, we'll use it in the `render()` method of the component

```
import { CDS_EXPERIEMENTAL_COMPONENT_NAME } from '../../globals/internal/feature-flags';

....

render() {
    return CDS_EXPERIEMENTAL_COMPONENT_NAME
      ? html`
          <ol class="${classes}">
            <slot></slot>
          </ol>
        `
      : null;
  }
```

### Changelog

**New**

- `feature-flags.ts` to house the flags
- copied over logic from dotcom web-components storybook to hide experimental components' stories from the storybook UI

PR #11201, opened 2023-12-05
feat(sidepanel): introduce side panel component

lee-chase:
Contributes to: https://github.com/carbon-design-system/carbon-for-ai/issues/16

### Description

Very much a draft version of the SidePanel. 

It is currently based on the branch `clean_main`.

It has a number of missing features 

Questions:
- Button set does not re-order when stacked. I am uncertain if the behavior is possible with Web components
- Story knobs not working as expected even when changing stories. Not found a declaration form that is either succinct or correctly updates even between stories. Ideally it should be possible to declare many stories with a single template and different default knobs.

Review:
- Feedback on the component produced so far would be useful, as it is the fist Web Component I have produced.

For feature parity with the React SidePanel the following are required
 - [ ] button set to re-order when stacked
 - [ ] Selector-initial-focus not functional
 - [ ] Add 2nd step story
 - [ ] Fix action bar refresh bug - this is a circular resize/scroll issue in reality the HTML needs a refactor to correctly use sticky
   but that involves changing the IBM Products code https://github.com/carbon-design-system/ibm-products/issues/4065

NOTES: 
- Storybooks feels a bit unwieldy for WebComponents, but that may just be use of V6 storybook.


mattrosno:
Added `closes` to PR body.

lee-chase:
@annawen1 @ariellalgilmore I realise there are some conflicts to resolve, but would love some feedback on the WebComponent SidePanel implementation so far. Just want to make sure I'm not going in completely the wrong direction with my first WebComponent.

packages/carbon-web-components/src/components/side-panel/side-panel.ts
```diff
@@ -0,0 +1,1085 @@
...
+    if (!open && !this._isOpen) {
+      return html``;
+    }
+
+    const containerClass = this.containerClass
+      .split(' ')
+      .filter(Boolean)
+      .reduce((acc, item) => ({ ...acc, [item]: true }), {});
+
+    const containerClasses = classMap({
```
ariellalgilmore, line 645: one thing different about web components is since now we have `cds-side-panel` we can leverage that and avoid using all of these different classes at the root of the component. Instead we would have props like `size`, and in styles do something like 

```
:host(cds-side-panel[size='lg']){
  @extend .${blockClass}__container--lg;
}
```

accordion to be a good component to look at which also has something similar to opening/closing. In accordion `expanding` / `collapsing` are being used

packages/carbon-web-components/src/components/side-panel/side-panel-header.ts
```diff
@@ -0,0 +1,276 @@
...
+  // /**
+  //  * Provide a `Slug` component to be rendered inside the `SidePanel` component
+  //  */
+  // @property({ reflect: true })
+  // slug;
+
+  /**
+   * Sets the subtitle text
+   */
+  @property({ reflect: true })
```
annawen1, line 100: should add types for all properties

packages/carbon-web-components/src/components/side-panel/side-panel-header.ts
```diff
@@ -0,0 +1,276 @@
...
+      // [`${blockClass}__subtitle-text-is-animating`]:
+      //   !animationComplete && animateTitle,
+      [`${blockClass}__subtitle-without-title`]: !title,
+    });
+
+    //  const titleEl = title && title.length && labelText && labelText.length ?
+    //   html`<p class={`${prefix}__label-text`}>{labelText}</p>` : html``;
+
+    return html`
+      <div class=${titleContainerClasses}>
```
annawen1, line 198: we could do what Ari suggested here, where we check the attributes passed into the components in the scss file and display accordingly by extending the classes

packages/carbon-web-components/src/components/side-panel/side-panel-story.ts
```diff
@@ -0,0 +1,423 @@
...
+  CDSSidePanel.sortActions(actions, actionsStacked);
+
+  const handleStackingChange = ({ detail }) => {
+    actionsStacked = !actionsStacked; // detail.stacked;
+    CDSSidePanel.sortActions(actions, actionsStacked);
+  };
+
+  return html`
+    <div
+      style="position: fixed; top: 0; left: 0; display: grid; grid-template-rows: 3rem 1fr; width: 100vw; height: 100vh;">
```
annawen1, line 281: can we move these inline styles to the `story-styles.scss` file instead?

packages/carbon-web-components/src/components/side-panel/side-panel.ts
```diff
@@ -0,0 +1,1085 @@
...
+      this._actionsCount = actionsCount;
+    }
+  }
+
+  /**
+   * The `ResizeObserver` instance for observing element resizes for re-positioning floating menu position.
+   */
+  // TODO: Wait for `.d.ts` update to support `ResizeObserver`
+  // @ts-ignore
+  private _resizeObserver = new ResizeObserver(() => {
```
annawen1, line 438: Not sure why we need the resize observer for the action bar? Can you explain?

packages/carbon-web-components/src/components/side-panel/side-panel.ts
lee-chase: @ariellalgilmore would you apply that to philosophy to the internals or just the component?

For example, is it preferable to add attributes rather the ibm-products classes, simpler classes to the title-container?

So 
```
.side-panel__on-detail-step <-- added to HTML or 

// Shorter CSS class inside component
.detail-step {
   @extend .side-panel__on-detail-step
}

// ID + attribute
#title-container[detail-step] {
  @extend .side-panel__on-detail-step
}
```

ariellalgilmore: for the internal classes you can keep the same classes as ibm-products, but there shouldn't be any container classes for components. Instead the container classes are usually used as props

packages/carbon-web-components/src/components/side-panel/side-panel-header.ts
lee-chase: > should add types for all properties
Done, although not sure what to do with Enum, will have a look around.

packages/carbon-web-components/src/components/side-panel/side-panel.ts
lee-chase: Resize observer is top cope with the unknown height of the button set, which may depending on side panel width get stacked. I can probably work around its use with a container query (replicating code in ibm-products), but will have to experiment. Hopefully will not discover `:has` is needed as it's not currently in FF ESR. 

NOTE: the stylelint config is very old in the repo.

packages/carbon-web-components/src/components/side-panel/side-panel-header.ts
lee-chase: Done

packages/carbon-web-components/src/components/side-panel/side-panel-story.ts
lee-chase: Done

lee-chase:
Despite the TODO list I think the component is ready to go experimental @jeffchew @annawen1 @ariellalgilmore 

The most obvious issue is the fact that changing some Knobs fails without a refresh or closing and opening the side-panel. This is a bit annoying but not a functional issue as far as I am aware. 

There is too much javascript in the title scroll animation and it feels a little flaky. This does not seem to happen in React, but I would like to revisit the React SidePanel to address and update styles. The 15/16 is almost unmaintainable and it took quite a bit of effort to get the Lit component lifecycle to mimic the React behavior. I suspect nothing more than timing is hiding some of the undesirable behavior (at least one repeating loop).

annawen1:
We do have [logic to create feature flags](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/11081/files) for the experimental components that would prevent it from showing in storybook, but since we now have the "experimental" section in storybook I don't think it's necessary. So no additional steps to do in order to ensure this as an experimental component.

packages/carbon-web-components/src/components/side-panel/side-panel.ts
```diff
@@ -0,0 +1,1118 @@
...
+          composed: true,
+          detail: {
+            triggeredBy,
+          },
+        }
+      )
+    );
+  }
+
+  @state()
```
annawen1, line 321: can we place all the state declarations in the same section (maybe up top) so it's easier to reference and find?

lee-chase:
@annawen1 queries
-  @extend seems to get an extra #side-panel selector in CSS (higher specificity) than following CSS is this normal?

ariellalgilmore:
> @annawen1 queries
> 
> * @extend seems to get an extra #side-panel selector in CSS (higher specificity) than following CSS is this normal?

sometimes that is the case. if we haven't been able to get the `@extend` to specify correctly then we end up just copying and pasting the styles

lee-chase:
Added a number of updates based on the review. 

Additional work is required for full feature parity, mainly re-order of ActionButtons if possible. There is also a fix for the initial select and some refactoring of the Carbon for IBM Products side panel to be accounted for. I would recommend using a separate PR for that work.

packages/carbon-web-components/src/components/side-panel/side-panel-story.ts
```diff
@@ -0,0 +1,433 @@
...
+import '../textarea/index';
+import storyDocs from './side-panel-story.mdx';
+import { SIDE_PANEL_PLACEMENT } from './defs';
+import Settings from '@carbon/icons/lib/settings/16';
+import Trashcan from '@carbon/icons/lib/trash-can/16';
+
+import styles from './story-styles.scss';
+import { BUTTON_KIND } from '../button/button';
+const toggleButton = () => {
+  document.querySelector('cds-side-panel')?.toggleAttribute('open');
```
ariellalgilmore, line 25:
```suggestion
  document.querySelector(`${prefix}-side-panel`)?.toggleAttribute('open');
```

ariellalgilmore (commented):
Left a few comments... the biggest change might be switching from using `id=` to `class=`

packages/carbon-web-components/src/components/side-panel/side-panel.ts
```diff
@@ -0,0 +1,1040 @@
...
+  // eslint-disable-next-line class-methods-use-this
+  private _handleActionToolbarChange(e: Event) {
+    const target = e.target as HTMLSlotElement;
+    const actions = target?.assignedElements();
+
+    this._hasActionToolbar = actions && actions.length > 0;
+
+    if (this._hasActionToolbar) {
+      for (const action of actions) {
+        // action size should always be md
```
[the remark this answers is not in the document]
lee-chase: sm good spot

packages/carbon-web-components/src/components/side-panel/side-panel.scss
```diff
@@ -0,0 +1,385 @@
...
+:host(#{$prefix}-side-panel) {
+  /* Replaces use of Framer for slide in / out animation */
+  --#{$block-class}--displaced: 100%;
+
+  * {
+    /* not sure why things are coming out as content-box */
+    box-sizing: border-box;
+  }
+
+  #side-panel {
```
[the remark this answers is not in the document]
lee-chase: Now using classes

packages/carbon-web-components/src/components/side-panel/side-panel-story.mdx
```diff
@@ -0,0 +1,71 @@
+import { Props, Description } from '@storybook/addon-docs/blocks';
+import { cdnJs, cdnCss } from '../../globals/internal/storybook-cdn';
+
+# Slider
```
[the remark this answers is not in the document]
lee-chase: Updated mdx

packages/carbon-web-components/src/components/side-panel/side-panel-button-set.ts
```diff
@@ -0,0 +1,42 @@
...
+import styles from '../button/button.scss';
+import { prefix } from '../../globals/settings';
+
+/**
+ * Button set.
+ *
+ * @element cds-side-panel-button-set
+ */
+@customElement(`${prefix}-side-panel-button-set`)
+class CDSSidePanelButtonSet extends LitElement {
```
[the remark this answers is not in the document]
lee-chase: Done

packages/carbon-web-components/examples/codesandbox/basic/components/side-panel/cdn.html
```diff
@@ -0,0 +1,35 @@
...
+  <style>
+    /* Suppress custom element until styles are loaded */
+    cds-slider:not(:defined) {
+      display: none;
+    }
+  </style>
+  <script type="module" src="https://1.www.s81c.com/common/carbon/web-components/tag/v2/latest/slider.min.js"></script>
+</head>
+<body class="cds-theme-zone-white">
+  <cds-slider
```
ariellalgilmore, line 27: small fix hopefully just copying the `index.html` `cds-side-panel` in here instead of slider

ariellalgilmore (approved):
LGTM code wise! not sure if we need a design review on this

lee-chase:
> LGTM code wise! not sure if we need a design review on this

Should not need a design review, should happen in @carbon/ibm-products for the side panel component.

packages/carbon-web-components/examples/codesandbox/basic/components/side-panel/cdn.html
lee-chase: Updated CDN @ariellalgilmore.

annawen1 (approved):
I think we can go ahead and merge this in!

PR #11205, opened 2023-12-06
fix(masthead-v2): do not fail if object data does not exist

jkaeser:
### Related Ticket(s)

Closes https://github.ibm.com/webstandards/ibm-dotcom-translations/issues/973

### Description

We weren't conditionally accessing all the data that is sourced from the translations CDN endpoint before mapping it to props. In cases where data is not where expected, it throws an undefined error and prevents the whole L0 from appearing. 

### Changelog

**Changed**

- Prevents L0 from failing to appear if certain nav data is not present where expected.

### To test:

See the L0 doesn't show up in the latest alpha version's Storybook example: https://carbon-design-system.github.io/carbon-for-ibm-dotcom/masthead-v2/web-components/iframe.html?args=&id=components-masthead--with-l-1&viewMode=story&cc=ie&lc=en. You'll see an error in the console, too: `Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'signedin')`.

Visit the deploy preview's masthead with L1 Story with `&cc=ie&lc=en` appended to the URL parameters (https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/11205/iframe.html?args=&id=components-masthead--with-l-1&viewMode=story&cc=ie&lc=en). Verify the L0 shows up.

### DIFF
--- a/packages/web-components/src/components/masthead/masthead-container.ts
+++ b/packages/web-components/src/components/masthead/masthead-container.ts
@@ -100,16 +100,17 @@ export function mapStateToProps(
         : translations?.[language]?.mastheadNav?.links,
       authenticatedProfileItems: !language
         ? undefined
-        : translations?.[language]?.profileMenu.signedin,
+        : translations?.[language]?.profileMenu?.signedin,
       unauthenticatedProfileItems: !language
         ? undefined
-        : translations?.[language]?.profileMenu.signedout,
+        : translations?.[language]?.profileMenu?.signedout,
       authenticatedCtaButtons: !language
         ? undefined
-        : translations?.[language]?.masthead?.profileMenu.signedin.ctaButtons,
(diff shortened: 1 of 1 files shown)


andy-blum (approved):
Tested on [deploy preview](https://carbon-design-system.github.io/carbon-for-ibm-dotcom/masthead-v2/web-components/iframe.html?args=&id=components-masthead--with-l-1&viewMode=story&cc=ie&lc=en) 👍

jkaeser:
@annawen1 I don't think the React deploy preview is ever going to complete. Can you force this merge for us please?

kennylam (approved):

kennylam:
> I don't think the React deploy preview is ever going to complete. Can you force this merge for us please?

@jkaeser that doesn't need to be part of v2 anyway, so I've removed it as a requirement for this branch.

PR #11212, opened 2023-12-08
chore(masthead-v2): merge latest from alpha dev into alpha main

jkaeser:
### Description

Merges bug fix from masthead v2 alpha's dev branch into its release branch.


kennylam (approved):
LGTM!

PR #11225, opened 2023-12-13
feat(card): use link instead of button for pictogram card

### LINKED ISSUE #10960: [Card with pictogram does not open correct right-click menu]

m4olivei:
### Related Ticket(s)

Closes #10960
[Jira ticket](https://jsw.ibm.com/browse/ADCMS-4074)

### Description

Alternate approach to fixing #10960. Previously we had attempted a resolution that wrapped the whole card in `<a></a>`. That solution introduced regressions with accessibiltiy. It's also come to my attention that the approach is a well [known poor approach](https://css-tricks.com/block-links-are-a-pain-and-maybe-just-a-bad-idea/). 

This solution takes the same approach as the Default Card component, which is to wrap the footer icon in a `<a></a>` and then absolute position a pseudo element child to make the whole card clickable still. More on [the pseudo-element  method](https://css-tricks.com/block-links-the-search-for-a-perfect-solution/#aa-method-3-the-good-ol-before-pseudo-element).

### Changelog

**Changed**

- Wrap the footer icon in a link for the card pictogram in the same way as the default card
- Adjustment to Storybook for Card so that the body copy is used on page load

**Removed**

- Remove the `role="button"` wrapper for pictogram card


kennylam (approved):
LGTM thanks @m4olivei!

Issue #11213, opened 2023-12-11
[lightbox-media-viewer]: Auto-open and play media from targeted link


m4olivei:
JIRA: https://jsw.ibm.com/browse/ADCMS-4360

### The problem

As a Site Author, we want to be able to compose a link as part of an email marketing campaign that will take a user to [a page like this one](https://www.ibm.com/info/ai-for-customer-service) and include a URL fragment or query string that indicates to the target page to automatically open a lightbox media viewer that is in the page and begin media playback.

The specific use case here is for "nurture campaigns". They would prefer to send them to this page, in context, and have a specific video play automatically (already embedded in the page) rather than sending users off site to the MediaCenter page.

### The solution

Enhance the Lightbox media viewer to look for a URL fragment with the pattern `lightbox-video-[video-id]`. If the `[video-id]` found corresponds to the `video-id` attribute of the `<dds-lightbox-media-viewer>` component, than automaticaly open the lightbox and auto-play the media when the lightbox is ready.

### Application/website

https://ibm.com

### Business priority

Medium Priority = upcoming release but is not pressing

### What time frame would this ideally be needed by (if applicable)

Ideally early to mid January 2024

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

m4olivei:
Adding feedback from @oliviaflory on accessibilty / feasibility:

Here's is the feedback I got from the accessibility team based on what you've shared, I asked some pointed questions after sharing the context of the use case:
1. Is it OK to auto play a video in the lightbox media player (essentially a giant modal) *IF* there is a pause or stop mechanism?
- [2.2.2 Pause, Stop, Hide](https://www.ibm.com/able/requirements/requirements/?version=v7_2#2_2_2) allows the autoplay as long as there's an accessible pause mechanism.
2. Does the CTA from the marketing email need to be more informative; i.e. "Watch the video" so the user has a better understanding of what will happen when they click it?
 - Informative link text would fall under [2.4.4 Link Purpose](https://www.ibm.com/able/requirements/requirements/?version=v7_2#2_4_4).
[...]

PR #11253, opened 2023-12-19
[lightbox-media-viewer]: Auto-open and play media from targeted link

m4olivei:
### Related Ticket(s)

Closes https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/11213
[Jira ticket](https://jsw.ibm.com/browse/ADCMS-4360)

### Description

Enables the `CTAMixin` to check the URL fragment for a particular pattern, `cta-video-[video-id]`, and if present, kick off the CTA trigger, which in the case of a video CTA, is opening the corresponding lightbox and auto-playing video. The effect is that visiting any page with a video CTA for a specific video embedded in it (eg. `0_ibuqxqbe`), with a URL fragment like `#cta-video-0_ibuqxqbe`, will automatically open the lightbox modal and play the video (provided [browser auto-play policies to not block it](https://developer.chrome.com/blog/autoplay/))

### Testing instructions

 - [ ] Navigate to any page with a CTA component that allows you to configure it as a video component. Examples include: [Link with icon](https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/11253/index.html?path=/story/components-link-with-icon--default)
 - [ ] Set the CTA type (cta-type) to Video (video)
 - [ ] Take note of the configured Video ID (eg. `0_ibuqxqbe`). 
 - [ ] Click Open canvas in new tab
 - [ ] The component should work without any regressions
 - [ ] Append a URL fragment to the URL like `#cta-video-[video_id]`
 - [ ] When loading the page with the URL fragment, the lightbox should automatically open and begin playback of the video (subject to [browser auto-play policies](https://developer.chrome.com/blog/autoplay/)). Note that mosy likely, the video _won't_ autoplay b/c you probably haven't built up enough of Media Engagement Index (Chrome speak) against the Storybook deploy preview URL. We don't have control over this.

### Changelog

**New**

- `CTAMixin`, where `ctaType === CTA_TYPE.VIDEO`,  will check for a specifically crafted URL fragment (`cta-video-[video-id]`) to trigger it's run action.


andy-blum:
I want to test this a little more thoroughly - my main concern at the moment is what happens if multiple CTAs share the same video. Do we get multiple lightboxes?

m4olivei:
> I want to test this a little more thoroughly - my main concern at the moment is what happens if multiple CTAs share the same video. Do we get multiple lightboxes?

That's a good question. I haven't explored that specifically, that may well be the case, unless the modal logic has protections against that 🤞 . I can make a note to test that when I'm back in the office next week.

andy-blum:
It does seem to open all matching CTAs this way

https://sandbox.ddev.site/ibm/auto-open-lightbox/multiple-ctas.html#cta-video-0_ibuqxqbe

m4olivei:
@andy-blum I addressed the multiple lightbox openings by [only allowing the first](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/11253/commits/35f36a6a53950ecfa4e4acc8250496b87f921f00) in the page to open by that URL trigger. Feels like that should be sufficient. 

From a content editing perspective, I would imagine they would want to try to avoid duplicating video use for CTA's, otherwise you'd have really keen users running into the same video more than once. There could be use cases for that I guess, hopefully this heuristic would suffice.

packages/web-components/src/component-mixins/cta/cta.ts
```diff
@@ -322,6 +322,57 @@ const CTAMixin = <T extends Constructor<HTMLElement>>(Base: T) => {
...
+      }
+      // Only trigger for the first CTA with the video id in the page.
+      if (this.ownerDocument.querySelector(`[href='${href}']`) !== this) {
+        return;
+      }
+      const { eventRunAction } = this.constructor as typeof CTAMixinImpl;
+      const hash = window.location.hash;
+      const urlTrigger = `cta-video-${href}`;
+
+      if (hash.includes(urlTrigger)) {
```
andy-blum, line 361: I'm thinking we might be better off doing an equivalence check instead of using String.prototype.includes.

The URL below will still open the video.
https://sandbox.andy-blum.com/ibm/auto-open-lightbox/multiple-ctas.html#cta-video-0_ibuqxqbe-NOT-A-VALID-HASH

packages/web-components/src/component-mixins/cta/cta.ts
```diff
@@ -322,6 +322,57 @@ const CTAMixin = <T extends Constructor<HTMLElement>>(Base: T) => {
...
+      // Once the eventRequestVideoData finishes and has set the videoName,
+      // check for the URL trigger meant to fire eventRunAction.
+      if (
+        ctaType === CTA_TYPE.VIDEO &&
+        href &&
+        changedProperties.has('videoName') &&
+        videoName
+      ) {
+        this._checkUrlVideoTrigger();
+      }
```
andy-blum, line 335: This works and is _probably_ safe, but this setup also allows the possibility of rouge lightboxes if the video name were to be changed for any reason.

1. Open [demo](https://sandbox.andy-blum.com/ibm/auto-open-lightbox/multiple-ctas.html#cta-video-0_ibuqxqbe)
2. Close the lightbox after it opens
3. Inspect the first link-with-icon
4. Manually alter the video-name property

I wonder if this logic would be better in `firstUpdated`?

```ts
class Element {
  firstUpdated(changedProperties) {
    if (document.readyState !== 'complete') {
      document.addEventListener('load', this._checkUrlVideoTrigger.bind(this));
    } else {
      this._checkUrlVideoTrigger();
    }
  }
}
```

andy-blum (changes requested):
One question about how this new behavior fits into the component lifecycle and one minor nit

andy-blum (approved):
Most recent changes look good. Looking forward to seeing @kennylam & co's takes on this.

packages/web-components/src/component-mixins/cta/cta.ts
```diff
@@ -324,6 +332,46 @@ const CTAMixin = <T extends Constructor<HTMLElement>>(Base: T) => {
...
+      // do here.
+      if (ctaType !== CTA_TYPE.VIDEO || !href || disabled) {
+        return;
+      }
+      // Only trigger for the first CTA with the video id in the page.
+      if (this.ownerDocument.querySelector(`[href='${href}']`) !== this) {
+        return;
+      }
+      const { eventRunAction } = this.constructor as typeof CTAMixinImpl;
+      const hash = window.location.hash;
```
kennylam, line 355: Small thing but this will fail in SSR applications (window undefined). Can use [window-or-global](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/masthead/masthead-composite.ts#L16).

m4olivei: Thanks @kennylam . Fixed in recent commit https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/11253/commits/983f98f1c85158b67e1b5d657e2896f8dbf333da. 

Also noticed a JS error while testing. Not related to changes, but felt small enough to take care of here (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/11253/commits/d815657486af3ac10c7e647326d3dfb1c8b2b528).

PR #11276, opened 2023-12-27
feat(date-picker): update story to Storybook v7

### LINKED ISSUE #11275: [carbon-web-components]: datepicker Storybook 7 update

m4olivei:
### Related Ticket(s)

Closes #11275

### Description

Adjustments to date-picker component to be compatible with Storybook v7.

### Changelog

**Changed**

- Updated date-picker Storybook component for Storybook v7


IgnacioBecerra (approved):
LGTM, thanks for the help! The layer stories should be working once we get the layer component to work.

m4olivei:
@IgnacioBecerra my pleasure! 

By the way, I noticed that I unintentionally dropped the `action` for `onChange` in the DatePicker > Playground story here. [It's here on `main`](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/carbon-web-components/src/components/date-picker/date-picker-story.ts#L33). Link component also had [an action for `onClick`](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/carbon-web-components/src/components/link/link-story.ts#L114), which was dropped in the conversion. Should we maintain those actions, or are we intentionally dropping them?

m4olivei:
> By the way, I noticed that I unintentionally dropped the action for onChange in the DatePicker > Playground story here. [It's here on main](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/carbon-web-components/src/components/date-picker/date-picker-story.ts#L33)

I've added the action back. This is what I've been doing with my more recent PRs.

I've also fixed the layers issue here. The problem is that there are some stories that are missing on or both of the required imports:

```
import '../layer/index';
import '../../../.storybook/templates/with-layer';
```

I'll watch out for that and add it from here on in.

PR #11282, opened 2024-01-01
fix(tooltip): tooltip persists issue

### LINKED ISSUE #10471: Carbon Web Components tooltip persists with certain delay durations
### Description

Given certain values of `enter-delay-ms` and `leave-delay-ms`, the tooltip will persist despite no user interaction

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/8265238/5f763344-ec86-458b-a58f-65fd41f70729

### Component(s) impacted

tooltip and any component consuming it

### Browser

Chrome, Safari, Firefox, Microsoft Edge

v2

### Severity

### Application/website

-

### Package

@carbon/web-components

### CodeSandbox example

-

using the v2 tooltip, pass in enter and leave delay durations like [500ms enter, 100ms leave] or [100ms enter, 0ms leave] then interact with the mouse to replicate the issue

_No response_

### Code of Conduct

sangeethababu9223:
### Related Ticket(s)

Closes #10471

### Description

Carbon Web Components tooltip persists with certain delay durations. If the `enter-delay-ms` is higher than the `leave-delay-ms` tooltip doesn't work as expected.

### Changelog

**Changed**

- Added a condition to check if `enter-delay-ms` is higher than `leave-delay-ms`. And if so, the value of `leave-delay-ms` will be increased by the value of `enter-delay-ms`.

### DIFF
--- a/packages/carbon-web-components/src/components/tooltip/tooltip.ts
+++ b/packages/carbon-web-components/src/components/tooltip/tooltip.ts
@@ -53,13 +53,13 @@ class CDSTooltip extends HostListenerMixin(CDSPopover) {
    * Specify the duration in milliseconds to delay before displaying the tooltip
    */
   @property({ attribute: 'enter-delay-ms', type: Number })
-  enterDelayMs = 100;
+  enterDelayMs = 500;
 
   /**
    * Specify the duration in milliseconds to delay before hiding the tooltip
    */
   @property({ attribute: 'leave-delay-ms', type: Number })
-  leaveDelayMs = 300;
+  leaveDelayMs = 100;
 
   /**
    * Specify the size of the tooltip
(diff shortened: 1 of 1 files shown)


IgnacioBecerra:
@sangeethababu9223 Hm, I think making the leave delay match the enter delay might not be what we're looking for, since the component should work with different delay times as seen in the [react story](https://react.carbondesignsystem.com/?path=/story/components-tooltip--playground&args=enterDelayMs:1000) as well.

sangeethababu9223:
> @sangeethababu9223 Hm, I think making the leave delay match the enter delay might not be what we're looking for, since the component should work with different delay times as seen in the [react story](https://react.carbondesignsystem.com/?path=/story/components-tooltip--playground&args=enterDelayMs:1000) as well.

@IgnacioBecerra
Created a new PR with suggested changes from the reference given. 
Closing the current PR since unwanted files had been accidentally pushed.

PR #11286, opened 2024-01-02
feat(cypress): adding more tests to the pricing table cells

bruno-amorim:
### Related Ticket(s)

Closes # [ADCMS-4051](https://jsw.ibm.com/browse/ADCMS-4051)

### Description

We created a regression when implementing the Pricing Table Cell feature to align text next to the optional icons. Then we almost wiped out the side-by-side alignment feature when fixing the regression. This all occurred because our current tests don't adequately reflect our expected functionality.

We need tests for the following cases:

1.     When the icon feature is enabled, user-entered cell content should appear next to the icon.
2.     User-entered content should appear in a vertical column layout whether there's an icon enabled or not.


andy-blum (dismissed):
Reading through the test you added and the logic seems sound, but we're missing the second test condition:

> User-entered content should appear in a vertical column layout whether there's an icon enabled or not.

See @oliviaflory's comment [here](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/11002#pullrequestreview-1704557294) for a screenshot of what problem we're looking to avoid.

packages/web-components/tests/e2e-storybook/cypress/integration/pricing-table/_tests.js
```diff
@@ -46,6 +51,33 @@ export const createTests = path => [
...
+            cy.get($content)
+              .find('.cds--structured-list-cell-icon-text')
+              .should('be.visible')
+              .then(($span) => {
+                const spanRect = $span[0].getBoundingClientRect();
+                
+                // Checking if the elements are rendered side by side by comparing if they have the same vertical position, allowing a tolerance of 15 px of difference.
+                expect(spanRect.top).to.be.closeTo(svgRect.top, 15);
+                // Same logic as above but now to check if all content is vertically aligned.
+                expect(spanRect.left).to.be.closeTo(svgRect.left, 1);
```
jkaeser, line 75: Is this asserting that the icon and the text should be vertically aligned? Because I don't think that's the goal of the second case we want to test for. Rather, we want to make sure that all the text in the cells of a given column are vertically aligned, specifically when only some of them have also are configured to show the icon.

In other words, given the following table we want to make sure all the text in the first column is vertically aligned even though the second cell doesn't use an icon.

| Column 1 | Column 2 | Column 3 |
|-|-|-|
| ✅ Lorem ipsum dolor sit amet | Foo | Bar |
| Consectetur adipiscing elit | Foo | Bar |
| ✅ Cras fermentum luctus magna | Foo | Bar |

jkaeser (changes requested):
Thanks for your work on this one, @bruno-amorim! It's looking great so far. I think the second test case still needs some work, though. See my comment and let me know if you have any questions.

bruno-amorim:
Hey @jkaeser
I've rewritten the logic to test the vertically aligned texts. Would you please check it now?

packages/web-components/tests/e2e-storybook/cypress/integration/pricing-table/_tests.js
jkaeser: After discussing on a Slack huddle earlier today, we're going to remove this test case. The v1 component does not support this vertical alignment feature at this time.

jkaeser:
@bruno-amorim Don't forget to re-target this against the v1 branch.

m4olivei (approved):
Changes look good to me. Tests passing. I just merged `v1` changes, otherwise this should be good to go.

jkaeser (approved):

annawen1 (approved):

PR #11336, opened 2024-01-04
feat(number-input): update story to Storybook v7

### LINKED ISSUE #11332: [carbon-web-components]: number input Storybook 7 update

m4olivei:
### Related Ticket(s)

Closes #11332

### Description

Adjustments to number-input component stories to be compatible with Storybook v7.

### Changelog

**Changed**

- Updated number-input Storybook stories for Storybook v7


oliviaflory (commented):
Hey @m4olivei this is looking pretty good! I have a few comments

1. the `read only` control styling is a little off, the buttons keep the border strong, but we would expect the border to be the same as the border under the number. Please refer to the [number input style page](https://carbondesignsystem.com/components/number-input/style#interactive-states) for the correct token color 
Screenshot 2024-01-08 at 3 38 20 PM

Following comments should be confirmed with @kennylam: are we matching React 1:1 or is CWC slightly different from an implementation standpoint?

2. missing the `disableWheel` control that we see in [React storybook](https://react.carbondesignsystem.com/?path=/story/components-numberinput--playground&args=allowEmpty:!true;iconDescription:;invalid:!false;invalidText:;label:;max:0;min:0;readOnly:!false;size:md;warn:!true;step:0;disableWheel:!false&globals=theme:white)
3. I see CWC has `decrementButtonDescription` and `incrementButtonDescription` are those the equivalent of `IconDescription` in React?
4. React has `onChange`; `onClick`; `onKeyUp` are those equivalent to `onInput` in CWC?

m4olivei:
Thanks @oliviaflory! To summarize below, based on a discussion with Kenny at office hours yesterday, I think the items you raise are out of scope of the present Storybook v7 update. Having said that, I'll leave my notes below on each item. Hope this helps.

> the read only control styling is a little off, the buttons keep the border strong, but we would expect the border to be the same as the border under the number. Please refer to the [number input style page](https://carbondesignsystem.com/components/number-input/style#interactive-states) for the correct token color

We're not touching styles here, this is an issue that exists in CWC already, see the same in [CWC v2 Storybook](https://carbon-design-system.github.io/carbon-for-ibm-dotcom/v2/carbon-web-components-v2/iframe.html?knob-Decrement%20button%20assistive%20description%20(decrement-button-assistive-text)=decrease%20number%20input&knob-Increment%20button%20assistive%20description%20(increment-button-assistive-text)=increase%20number%20input&knob-Helper%20text%20(helper-text)=Optional%20helper%20text&knob-Invalid%20text%20(invalid-text)=Number%20is%20not%20valid&knob-Label%20(label)=number-input%20label&knob-Read%20only%20(readonly)=true&knob-Value%20(value)=50&knob-Warn%20text%20(warn-text)=Warning%20text&knob-Minimum%20value%20(min)=0&knob-Maximum%20value%20(max)=100&knob-Value%20to%20step%20the%20input%20by%20(step)=1&knob-Size%20(size)=md&id=components-number-input--playground&viewMode=story). Seems like this should get files as issue?

> Following comments should be confirmed with @kennylam: are we matching React 1:1 or is CWC slightly different from an implementation standpoint?

I spoke with Kenny at office hours yesterday. My understanding is that over time we aim to match React more closely, however in this present exercise, we may not get all the way there, as there have been new features added and/or changes made to Carbon React since CWC v2 launched. Where we see those differences which are outside the scope of the present excersise, updating to Storybook v7, we would file follow up issues to be handled later.

> missing the disableWheel control that we see in [React storybook](https://react.carbondesignsystem.com/?path=/story/components-numberinput--playground&args=allowEmpty:!true;iconDescription:;invalid:!false;invalidText:;label:;max:0;min:0;readOnly:!false;size:md;warn:!true;step:0;disableWheel:!false&globals=theme:white)

This functionality appears to be missing from CWC. The mouse wheel can't be used to affect the input in the first place, do `disableWheel` doesn't make sense. I imagine this would need to be filed as an issue.

> I see CWC has decrementButtonDescription and incrementButtonDescription are those the equivalent of IconDescription in React?

Yes. In addition though, after looking at this in the React package, I think you uncovered a bug over there. Despite seting an `iconDescription` in React, I'm finding that it's not used at all. Looking at the code, `iconDescription` is written to be used as a fallback for when there are no values otherwise provided. However those values are [provided statically in the React package](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/NumberInput/NumberInput.tsx#L30-L33) (as "Increment number" and "Decrement number") so `iconDescription` never gets used. You can override them independently in the React package, however the way you do that is using the `translateWithId` to provide translations for each or both.

All that is to say, there might be follow up work in the React package to drop `iconDescription` since it effectively doesn't do anything. Nothing to do here.

> React has onChange; onClick; onKeyUp are those equivalent to onInput in CWC?

Out of scope here, but I think this needs some work in CWC to get closer to being equivalent. 

Not equivalent, although you can mostly achieve what you need from an `onChange` in React via an `onInput` in CWC. We don't presently have the equivalent of `onClick`. We have some code that _should_ work via `onInput` to give you what you get from the React `onClick`, however I'm finding it doesn't work. I have some ideas to fix. I can file a follow up issue for this.

kennylam:
@oliviaflory Like Matt said we spoke yesterday about how far to sync with the React stories. In some instances it isn't possible yet since it may be new features that haven't yet been implemented in cwc. Other differences, such as `on*` events, can be added afterwards. I'm willing to take on a little bit of debt to get this work through. Even after the upgrade is complete we still need to go through each component and compare with the React version to ensure parity.

kennylam:
@oliviaflory I created a new issue addressing the `readOnly` styles here https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/11390.

kennylam (approved):

PR #11358, opened 2024-01-08
feat(button): cwc controls

### LINKED ISSUE #11265: [carbon-web-components]: button Storybook 7 update
upgrade to storybook v7

ariellalgilmore:
### Related Ticket(s)

Closes #11265 

### Description

convert to controls
### Changelog

**Changed**

- also formatted some other files


packages/carbon-web-components/src/components/button/button.stories.ts
```diff
@@ -0,0 +1,429 @@
...
+  [`Large size (${BUTTON_SIZE.LARGE})`]: BUTTON_SIZE.LARGE,
+  [`XL size (${BUTTON_SIZE.EXTRA_LARGE})`]: BUTTON_SIZE.EXTRA_LARGE,
+  [`2XL size (${BUTTON_SIZE.EXTRA_EXTRA_LARGE})`]:
+    BUTTON_SIZE.EXTRA_EXTRA_LARGE,
+};
+
+const defaultArgs = {
+  kind: BUTTON_KIND.PRIMARY,
+  tooltipAlignment: BUTTON_TOOLTIP_ALIGNMENT.CENTER,
+  tooltipPosition: BUTTON_TOOLTIP_POSITION.TOP,
```
m4olivei, line 68: Noticing that you left off a number of defaults here, which causes Storybook to render the button, eg. "Set string" for the missing `dangerDescription`. Is that to get us as close to React's Storybook as possible? Should we be doing this for all args in this conversion? If so, I need to go back through my PR's. Happy to do that, let me know.

packages/carbon-web-components/src/components/button/button.stories.ts
```diff
@@ -0,0 +1,429 @@
...
+};
+
+const controls = {
+  buttonClassName: {
```
m4olivei, line 72: Noticing that this is a new control over what we had. Guessing this is again to get us close to React Storybook. I might have missed this, but, are we aiming for 100% parity in this exercise? What's included / not included in that outside of whats already noted in the [migration guide](https://gist.github.com/IgnacioBecerra/0cba18e31c978e5632a9cbd8045e82e2)?

ariellalgilmore: `ButtonClassName` control is to keep parity with React, but the property already existed. Through this exercise, I don't believe we should introduce any new properties, but just try to align what's currently available to the react controls.

packages/carbon-web-components/src/components/button/button.stories.ts
ariellalgilmore: I just left off defaults that weren't directly effecting the button component storybook results

packages/carbon-web-components/src/components/button/button.stories.ts
m4olivei: That makes sense, also spoke with Kenny this morning at office hours on this. Will look out for these types of things. Thanks!

packages/carbon-web-components/src/components/button/button.stories.ts
m4olivei: Got it. I'll adjust and take the same approach moving forward. Thanks!

IgnacioBecerra (approved):
LGTM!! Thanks @ariellalgilmore!

kennylam (approved):
LGTM!

Issue #11256, opened 2023-12-20
[web-components]: Masthead custom search issue


NicoletaBordea:
[JIRA](https://jsw.ibm.com/browse/ADCMS-4402)

### Description

I am using masthead with custom typeahead and redirected search just like in the codepen example provided below. When clicking on the search icon or clicking a suggestion everything work fine, but if I am using keyboard to select a suggestion, or simply typing a text and pressing enter, query is null.

### Component(s) impacted

Masthead search

### Browser

Chrome

### Carbon for IBM.com version

v2.0.0-rc.3

### Severity

Severity 1 = The design is broken in a critical way that blocks users from completing tasks or damages the brand. Affects major functionality, no workaround.

### Application/website

ibm docs

### Package

@carbon/ibmdotcom-web-components

### CodeSandbox example

https://codepen.io/jakaeser/pen/gOqErGE

### Steps to reproduce the issue (if applicable)

1. type in a text in the search field and hit "Enter". The redirect link in the alert message is https://duckduckgo.com/?q=null
2. type in a text, then go down with arrow down key and select a suggestion. Hit "Enter". First it will appear an alert with correct redirect url, followed by a new alert with null query
3. type in a text and click search icon or select a suggestion by clicking on it will display the correct redirect url

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

NicoletaBordea:
Hi, I have installed latest dotcom version in our project (v2.2.0-rc.1), and tested this issue but seems that is not completely fixed. First scenario is still returning null query. When type in a text in the search field and hit "Enter". 
Second scenario it is fixed: when type in a text, then go down with arrow down key and select a suggestion. Hit "Enter". This is returning correct query.
Thank you

PR #11376, opened 2024-01-09
fix(search-with-typeahead): make eventBeforeRedirect return consistent & useful data

jkaeser:
### Related Ticket(s)

Closes https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/11256

### Description

The reporter of the issue was using a demo I'd set up showing how to customize the masthead's search-with-typeahead. While the `null` error is a bug in my demo, it reveals an issue: the `c4d-search-with-typeahead-beingredirected` event can be dispatched from two places, and the `details` returned are not consistent in these two places. Additionally, the same event can be dispatched twice in rapid succession if a listener calls `event.preventDefault()`.

This PR makes the returned details consistent, and it improves them by giving listeners easy access to what they're probably looking for anyway: the search query.

### Changelog

**New**

- Passes the current search query, be that the text input value or a typeahead item selected, to `c4d-search-with-typeahead-beingredirected` event listeners.

**Changed**

- Improves readability of `search-with-typeahead` component by extracting some logic into a standalone method and making consistent use of getters.

### Testing

Masthead deploy preview Story: https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/11376/index.html?path=/story/components-masthead--default

You'll need to verify that both searching by a query string and selecting typeahead result items behave consistently with regards to the `c4d-search-with-typeahead-beingredirected` event.

1. Add an event listener for the event using the browser console, e.g.:
   ```
   document.documentElement.addEventListener(
     'c4d-search-with-typeahead-beingredirected',
     (e) => {
       e.preventDefault();
       const { redirectUrl, searchQueryString } = e.detail;
       const newUrl = new URL(`https://duckduckgo.com/?q=${searchQueryString}`).toString();
       debugger;
       alert(`At this point, you could navigate to ${newUrl}`);
     }
   );
   ```
2. In the debugger, verify the `redirectUrl` and `searchQueryString` event details return the correct information for each of the user inputs.
3. Verify the event is not dispatched twice by confirming only one alert appears per search action.


m4olivei (approved):
Looks great to me! Nice cleanup as well.

PR #11404, opened 2024-01-12
[Locale API]: Adjust fallback layer logic

### LINKED ISSUE #11238: [Locale API]: Add one or more fallback layers before defaulting users to en-us
[JIRA](https://jsw.ibm.com/browse/ADCMS-4381)

### The problem

The Locale API [pulls information from the Digital Data object](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/1ce5f8dc525b47bf48c84971ea08f3bf01e855c1/packages/services/src/services/Locale/Locale.js#L124-L126) to know what locale to use. This, in turn, is used by components like the masthead and footer to provide the correct translation files. If the LocaleAPI does not provide usable values, however, these components immediately fall back to `us-en` as the global default

### The solution

The Locale API should provide an array of options in order of preference for components to have several layers of fallbacks. The DDO should remain the first priority for backwards-compatibility reasons, but there are several other levels of "close enough" between a bad DDO and `us-en`.

Possible additions:
* the values of `document.documentElement.lang` - this would be the locale provided by the author/platform
* the values of `navigator.language` or `navigator.languages` - these values come from the user's browser and reflect language preferences
* Some combination of one or both of the above and the page's alternate languages specified by `<link rel="alternate" hreflang="de-de" href="..."/>`

m4olivei:
### Related Ticket(s)

Closes #11238
[Jira ticket](https://jsw.ibm.com/browse/ADCMS-4381)

### Description

This PR makes adjustments to the fallback logic that defaults users to en-us.

Along the way it became clear that perhaps the existing logic is not behaving as expected. The first step is to make changes to the test suite to more accurately represent use of the API, revealing bugs. Specifically, as I understand it, the test suite and existing `LocalAPI` seems to communicate the following _expected_ fallback sequece for deciding locale:

1. DDO
2. Lang Attribute
3. Cookies
4. Browser (`navigator.language`)
5. Default (us-en)

However, due to a bug, cookies and browser `navigator.language` are completely ignored. Through testing, I believe this is the _actual_ sequence on `main` today:

1. DDO
2. Lang Attribute
3. Default (us-en)

I'm finding that the test suite has dependencies between tests. Not only that but some of the mocked values are not representative of the API, leading to false positives in the tests.

As such the aims of this PR include:

* Test isolation. Stand up and tear down mocks between each test such that we have a clean SUT
* Restore the indended fallback sequence
* Make `LocaleAPI.getLocale` and `LocaleAPI.getLang` behave the same and deprecate `LocaleAPI.getLang`
* Remove most tests for `getLang`, they are effectively the same as `getLocale` now.
* Improve LocaleAPI JSDoc.

### Testing

I've gone over the test suite and hopefully made things more reliable. Any individual test should be able to run in isolation (eg. `it.only`) and it will pass. Test coverage is reporting back at 100%, which isn't everything, but it's something :).

To manually test, there are a couple of things that can be checked. The `LocaleAPI` is used by the footer, masthead 
 and locale modal components. By default, we inject a static DDO object into any Storybook preview, set to `en-US`. You can override that by setting an `lc` and `cc` query string param, eg. append `&lc=pt&cc=br` to the URL. The UI should respond accordingly. This isn't a complete test, b/c it's sets the DDO object, which is the first step in the fallbacks. It will never get to the other fallbacks. If anyone knows of a better way to test, let me know.

### Changelog

**Changed**

- Test isolation. Stand up and tear down mocks between each test such that we have a clean SUT
- Restore the indended fallback sequence
- Make `LocaleAPI.getLocale` and LocaleAPI.getLang` behave the same and deprecate `LocaleAPI.getLang`
- Remove most tests for `getLang`, they are effectively the same as `getLocale` now.
- Improve LocaleAPI JSDoc.


jkaeser (changes requested):
The code changes all look really good to me, but I'm running into an error when executing `yarn test` from the services package. Specifically, the `global` service's tests fail with:
```
TypeError: _DDO.DDOAPI.getLocation is not a function
```

m4olivei:
> The code changes all look really good to me, but I'm running into an error when executing `yarn test` from the services package. Specifically, the `global` service's tests fail with:
> 
> ```
> TypeError: _DDO.DDOAPI.getLocation is not a function
> ```

Thanks @jkaeser . I think I fixed it in the new commits. The issue was that now that the `LocaleAPI` is fixed and incorporates all layers of fallback, the other tests (I also found `SearchTypeahead` had issues) were getting deeper into the `LocaleAPI` and running into API's that hadn't been mocked out for the test.

It does reveal potential issues in `LocalAPI`, where we could be handling edge cases getter, specifically when [our call to `DDOAPI.getLocation`](https://github.com/m4olivei/carbon-for-ibm-dotcom/blob/feat/locale-fallback/packages/services/src/services/Locale/Locale.js#L145) behaves unexpectedly. If you follow the code, it's [possible that call will reject](https://github.com/m4olivei/carbon-for-ibm-dotcom/blob/feat/locale-fallback/packages/services/src/services/DDO/DDO.js#L67). I think we need a test case around that. I'll work on that.

andy-blum:
_clopening_ to trigger the CDN build

m4olivei:
> It does reveal potential issues in `LocaleAPI`, where we could be handling edge cases better, specifically when [our call to `DDOAPI.getLocation`](https://github.com/m4olivei/carbon-for-ibm-dotcom/blob/feat/locale-fallback/packages/services/src/services/Locale/Locale.js#L145) behaves unexpectedly. If you follow the code, it's [possible that call will reject](https://github.com/m4olivei/carbon-for-ibm-dotcom/blob/feat/locale-fallback/packages/services/src/services/DDO/DDO.js#L67). I think we need a test case around that. I'll work on that.

I've added a test for this in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/11404/commits/f0c564f450995cfe8f68d0fde17ee1c1056adc45 . It was indeed an issue. Yay testing. Fixed in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/11404/commits/01c9652186eafc57cc596d38af37af66c77e79d8 .

jkaeser (approved):
Tests are passing now. Excellent work as usual, @m4olivei 😄

PR #10986, opened 2023-09-28
fix(masthead): new property for custom nav links

annawen1:
### Related Ticket(s)

Closes #10984

### Description

The navigation data fetched from the translation files automatically replaces any custom navigation data passed in. This PR creates a new property `customNavLinks` that adopters will set the custom nav data to. 

### Changelog

**New**

- `customNavLinks` property to pass in custom nav links

**Changed**

- add checks to `masthead-composite` that will render the custom nav links over the default fetched data

PR #11420, opened 2024-01-19
fix(masthead): container should respect user's L0 data

jkaeser:
### Related Ticket(s)

Previously fixed in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10986, but that was unwittingly wiped away by https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/11252

### Description

Fixes an issue where `<c4d-masthead-container>` wipes away user-set L0 data with translations fetched from an API.

This PR takes a different approach from the original fix. The root problem is that the masthead container isn't aware of any `l0Data` passed in before setting that property with translation data. I figured we can make the container a little smarter and handle the logic internally rather than adding another property to the masthead composite and complicating the public API.

In order to facilitate this, I needed the `ConnectMixin` to pass a reference to the class instance into the `mapStateToProps` function it expects classes to define. This empowers the container to first check if it has L0 data before resorting to the translations API data.

As a result, users can reliably set the `l0Data` property on either `<c4d-masthead-container>` or `<c4d-masthead-composite>` and expect the same outcome.

### Changelog

**New**

- `ConnectMixin` passes a class instance reference as an argument in the `mapStateToProps` function.

**Changed**

- `<c4d-masthead-container>` will prefer using L0 data passed by user before resorting to automatically fetched translation data.


jkaeser:
@annawen1 I'm curious to hear your thoughts on this one, since it takes a different approach to a fix you added back in September.

annawen1 (approved):

IgnacioBecerra (approved):
LGTM!!

PR #11421, opened 2024-01-19
Fix/combobox accessibility

### LINKED ISSUE #11268: [accessibility]: accessibility issue with combox
### Description

i have some violations when using the IBM Equal Access Accessibility Checker on the combox
image

[Accessibility_Report-test.xlsx](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/files/13751465/Accessibility_Report-test.xlsx)

### Component(s) impacted

combox

### Browser

_No response_

V1

### Severity

### Application/website

IBM Terms

### Package

@carbon/ibmdotcom-web-components, @carbon/web-components

### CodeSandbox example

https://web-components-v1.carbondesignsystem.com/?path=/story/components-combo-box--default

Code used :

<link rel="stylesheet" href="https://1.www.s81c.com/common/carbon-for-ibm-dotcom/version/v1.37.0/plex.css" />
    <link rel="stylesheet" href="https://1.www.s81c.com/common/carbon-for-ibm-dotcom/version/v1.37.0/grid.css" />
    <link rel="stylesheet" href="https://1.www.s81c.com/common/carbon-for-ibm-dotcom/version/v1.37.0/themes.css" />
    

    <script type="module"
    src="https://1.www.s81c.com/common/carbon/web-components/version/v1.35.0/combo-box.min.js"></script>

<bx-combo-box
  helper-text="Optional helper text"
  label-text="Combo box title"
  trigger-content="Filter..."
  aria-label="Combo box"
>
  <bx-combo-box-item value="all">Option 1</bx-combo-box-item>
  <bx-combo-box-item value="cloudFoundry">Option 2</bx-combo-box-item>
  <bx-combo-box-item value="staging">Option 3</bx-combo-box-item>
  <bx-combo-box-item value="dea">Option 4</bx-combo-box-item>
  <bx-combo-box-item value="router">Option 5</bx-combo-box-item>
  <bx-combo-box-item value="support">Option 6</bx-combo-box-item>
  <bx-combo-box-item value="services">Option 7</bx-combo-box-item>
  <bx-combo-box-item value="products">Option 8</bx-combo-box-item>
</bx-combo-box> 

_No response_

### Code of Conduct

m4olivei:
### Related Ticket(s)

Closes #11268
[Jira ticket](https://jsw.ibm.com/browse/ADCMS-4401)

### Description

Fixes accessibility issues with Combo-box, and by extension Dropdown.

Used both React package (which uses [Downshift](https://www.downshift-js.com/)), and [ARIA APG](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/) as references. Tested with VoiceOver on Mac OS.

### Changelog

**Changed**

- Improved dropdown and combo-box accessibility.


packages/carbon-web-components/src/components/dropdown/dropdown.ts
```diff
@@ -129,10 +131,12 @@ class CDSDropdown extends ValidityMixin(
         ),
         (item) => {
           (item as CDSDropdownItem).selected = false;
+          item.setAttribute('aria-selected', 'false');
         }
       );
       itemToSelect.selected = true;
-      this._assistiveStatusText = this.selectedItemAssistiveText;
+      itemToSelect.setAttribute('aria-selected', 'true');
+      this._assistiveStatusText = `${itemToSelect.textContent} has been selected.`;
```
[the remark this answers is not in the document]
m4olivei: I think there are a few issues with this assistive text. Its confusing to follow b/c `this._assistiveText` (and others) are just plain class properties, not `@state` like you might expect, so I think the problem is that it depends on changing the value _and_ calling `this.requestUpdate()` to trigger an update cycle, and sometimes it's missing.

After playing and thinking about it, I actually thing the `_assitiveText` and `aria-live` region handling is getting in the way. I think it was compensating for not having the correct `aria` attributes setup, which we have now. I've been comparing with ARIA APG examples ([dropdown](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/#ex_label), [combobox](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/#ex_label)). I believe we have an experience on par with those. Neither of which manage any `aria-live` region. I've taken the `aria-live` region out in the recent commits.

m4olivei:
Ready for a re-review.

> The options' aria-selected value does not update to "true" when they are selected. They do get a selected attribute, however.

This is fixed in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/11421/commits/6400a4a818bafd875d26d866464ab622500c2e8e

> When selecting an item, the vague "Selected an item" message is announced. This should announce which specific item has been selected, right?

The ARIA APG examples don't manage any `airia-live` regions and allow the screen reader to manage a correcly marked up combobox. I've taken out all the `aria-live` management in favor of the correct `aria` markup for the respective elements. IMO this is better as it lines up with a typical / expected combobox experience.

In either case, I believe we've fixed the original bugs and hopefully improved things significantly here.

kennylam (approved):
LGTM!

m4olivei:
> Oops, missed these [CI errors](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/actions/runs/7751288847/job/21138931114?pr=11421#step:5:256). While looking at these, I realized we can probably remove `footer/dropdown.ts` entirely and just use it directly from `@carbon/web-components`. Just did a quick test locally ([kennylam@44747f6](https://github.com/kennylam/carbon-for-ibm-dotcom/commit/44747f6667c1ddb69e874d6728325efca0d9a592)) and `footer - language only` variants look fine; `search-with-typeahead` is throwing an `axios` error but I didn't spend any time trying to get it to work, I just wanted to replace `footer/dropdown.ts` with the one from Carbon web components so the package would build.

Ahh I see. If I'm following along right, dropdown and combobox were forked under the footer component for accessibility fixes (https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/5658). They've drifted a bit overtime. I'm on board with taking the time to remove those forks, assuming that the accessibility fixes here end up being sufficent. 

> I know this is out of the scope of this PR, so if you want to apply the same `activeDescendant` change to fix the CI errors, we can create separate issues for getting rid of `footer/dropdown.ts`.

I spent some time trying to just add the `activeDescendant` related changes, but it wasn't straightforward to merge those changes, and it felt like it was just delaying the inevitable. 

I don't think there are time pressures to get this fix, so I'll try to carry on where you left off with removing the forked footer and combo-box.

m4olivei:
I've removed `packages/web-components/src/components/footer/dropdown.ts` and made related changes. @kennylam and @jkaeser this is ready for another look. Thanks!

jkaeser (approved):
Thanks for showing me the correct way to interact with these elements using VoiceOver! Previously I was simply using a combination of the arrow keys, tab, and return, but instead you have to use the VoiceOver-specific keybinds.

Now that we've got user error out of the way, this is looking great to me! Confirmed I could navigate both Dropdown and ComboBox components and select an option.

packages/carbon-web-components/src/components/combo-box/combo-box.ts
```diff
@@ -17,6 +17,8 @@ import CDSDropdown, { DROPDOWN_KEYBOARD_ACTION } from '../dropdown/dropdown';
 import CDSComboBoxItem from './combo-box-item';
 import styles from './combo-box.scss';
 import { carbonElement as customElement } from '../../globals/decorators/carbon-element';
+import { ifDefined } from 'lit/directives/if-defined';
+import ifNonEmpty from '../../globals/directives/if-non-empty';
```
kennylam, line 21: I think we can slowly start removing `ifNonEmpty`, since the latest implementation of [ifDefined](https://lit.dev/docs/templates/directives/#ifdefined) covers both `null` and `undefined`? In fact, after double-checking, `ifNonEmpty` is using `ifDefined` under the hood.

kennylam (commented):
This is looking great! Just one small suggestion.

packages/carbon-web-components/src/components/combo-box/combo-box.ts
m4olivei: Hum, I changed it, but then I reverted the change. Here I actually want `ifNonEmpty` since, where it's used, it references a string variable that defaults to the empty string, and I want to leave the attribute out if its set to the empty string. 

There is a difference between the two. `ifNonEmpty` will avoid outputting the attribute if the value is the empty string, in addition to when its `null` or `undefined`. See [this lit playground](https://lit.dev/playground/#project=W3sibmFtZSI6InNpbXBsZS1ncmVldGluZy50cyIsImNvbnRlbnQiOiJpbXBvcnQgeyBodG1sLCBjc3MsIExpdEVsZW1lbnQgfSBmcm9tIFwibGl0XCI7XG5pbXBvcnQgeyBjdXN0b21FbGVtZW50LCBwcm9wZXJ0eSB9IGZyb20gXCJsaXQvZGVjb3JhdG9ycy5qc1wiO1xuaW1wb3J0IHsgaWZEZWZpbmVkIH0gZnJvbSBcImxpdC9kaXJlY3RpdmVzL2lmLWRlZmluZWQuanNcIjtcblxuY29uc3QgaWZOb25FbXB0eSA9ICh2YWx1ZSkgPT5cbiAgaWZEZWZpbmVkKHZhbHVlID09PSBcIlwiID8gdW5kZWZpbmVkIDogdmFsdWUgPz8gdW5kZWZpbmVkKTtcblxuQGN1c3RvbUVsZW1lbnQoXCJzaW1wbGUtZ3JlZXRpbmdcIilcbmV4cG9ydCBjbGFzcyBTaW1wbGVHcmVldGluZyBleHRlbmRzIExpdEVsZW1lbnQge1xuICBzdGF0aWMgc3R5bGVzID0gY3NzYFxuICAgIHAge1xuICAgICAgY29sb3I6IGJsdWU7XG4gICAgfVxuICBgO1xuXG4gIEBwcm9wZXJ0eSgpXG4gIG5hbWUgPSBcIlNvbWVib2R5XCI7XG5cbiAgcHJpdmF0ZSBmb28gPSBcIlwiO1xuXG4gIHByaXZhdGUgYmFyID0gXCJiYXJcIjtcblxuICBwcml2YXRlIGJhejtcblxuICByZW5kZXIoKSB7XG4gICAgY29uc3QgeyBmb28sIGJhciwgYmF6IH0gPSB0aGlzO1xuICAgIHJldHVybiBodG1sYDxwXG4gICAgICBkYXRhLWZvby1pZmRlZmluZWQ9XCIke2lmRGVmaW5lZChmb28pfVwiXG4gICAgICBkYXRhLWJhci1pZmRlZmluZWQ9XCIke2lmRGVmaW5lZChiYXIpfVwiXG4gICAgICBkYXRhLWJhei1pZmRlZmluZWQ9XCIke2lmRGVmaW5lZChiYXopfVwiXG4gICAgICBkYXRhLWZvby1ub25lbXB0eT1cIiR7aWZOb25FbXB0eShmb28pfVwiXG4gICAgICBkYXRhLWJhci1ub25lbXB0eT1cIiR7aWZOb25FbXB0eShiYXIpfVwiXG4gICAgICBkYXRhLWJhei1ub25lbXB0eT1cIiR7aWZOb25FbXB0eShiYXopfVwiXG4gICAgPlxuICAgICAgSGVsbG8sICR7dGhpcy5uYW1lfSFcbiAgICA8L3A-YDtcbiAgfVxufVxuIn0seyJuYW1lIjoiaW5kZXguaHRtbCIsImNvbnRlbnQiOiI8IURPQ1RZUEUgaHRtbD5cbjxoZWFkPlxuICA8c2NyaXB0IHR5cGU9XCJtb2R1bGVcIiBzcmM9XCIuL3NpbXBsZS1ncmVldGluZy5qc1wiPjwvc2NyaXB0PlxuPC9oZWFkPlxuPGJvZHk-XG4gIDxzaW1wbGUtZ3JlZXRpbmcgbmFtZT1cIldvcmxkXCI-PC9zaW1wbGUtZ3JlZXRpbmc-XG48L2JvZHk-XG4ifSx7Im5hbWUiOiJwYWNrYWdlLmpzb24iLCJjb250ZW50Ijoie1xuICBcImRlcGVuZGVuY2llc1wiOiB7XG4gICAgXCJsaXRcIjogXCJeMy4wLjBcIixcbiAgICBcIkBsaXQvcmVhY3RpdmUtZWxlbWVudFwiOiBcIl4yLjAuMFwiLFxuICAgIFwibGl0LWVsZW1lbnRcIjogXCJeNC4wLjBcIixcbiAgICBcImxpdC1odG1sXCI6IFwiXjMuMC4wXCJcbiAgfVxufSIsImhpZGRlbiI6dHJ1ZX1d).

kennylam: Huh that's interesting. I could swear I read in Lit v3 docs that `ifDefined` now includes the behavior of `iFnonEmpty`. Disregard then, my bad!

PR #11447, opened 2024-01-24
fix(tabs-extended): tabs extended not rendering when in 1056 breakpoint

bruno-amorim:
### Related Ticket(s)

Closes # 
[ADCMS-4241](https://jsw.ibm.com/browse/ADCMS-4241)

### Description

Tab extended component is not rendering the tabs section when breakpoint is exactly 1056px, if author sets the browser simulator at 1057 and other then tab extended does render properly the same is true for a breaking point below like 1055 and below then it the component will render properly. 

### Changelog

**Changed**
- Issue was fixed by changing the way of checking the breakpoint to render the mobile or the desktop version

Screenshot 2024-01-24 at 16 11 47


bruno-amorim:
Tabs Extended with Media also fixed! Ready for review

m4olivei (approved):
Looks great! Nice work.

PR #11468, opened 2024-01-31
Fix(datatable): setting default sort column doesn't work

### LINKED ISSUE #11451: Setting default sorted column in data-table doesn't work (and table toolbar disappears if a table title added)
### Description

It doesn't appear to be possible to set a column in a data-table to be sorted by default. adding `sort-direction="ascending"` makes the column unsortable rather than making that column sorted.
```
      <cds-table-head>
        <cds-table-header-row>
          <cds-table-header-cell>Name</cds-table-header-cell>
          <cds-table-header-cell sort-direction="ascending">Status</cds-table-header-cell>
        </cds-table-header-row>
      </cds-table-head>
```

The documentation suggests this to sort the first column. However that just produces an error of `Uncaught TypeError: document.querySelectorAll(...)[0].shadowRoot is null`
```
document
  .querySelectorAll('cds-table-header-cell')[0]
  .shadowRoot.querySelector('button')
  .click();
```

Additionally, if you have a table with a toolbar, the toolbar isn't displayed if you also use the `cds-table-header-title` option. So it's not possible to have a title on a table that has a toolbar.
`<cds-table-header-title slot="title">DataTable</cds-table-header-title>`

### Component(s) impacted

cds-table

### Browser

Firefox

v2/latest

### Severity

### Application/website

Various tools for UK Labs

### Package

@carbon/web-components

### CodeSandbox example

n/a

_No response_

_No response_

### Code of Conduct

sangeethababu9223:
### Related Ticket(s)

Closes #11451

### Description

Setting default sort column for data table using `sort-direction` property doesn't work.

### Changelog

**New**

- Added a condition to check if during initial load `sort-direction` is set for any of the table header cells, and if so sort that column by default.

**Changed**

- Moved sorting into separate function so that it can be called during initial load as well as on click event on sort icons.


IgnacioBecerra (approved):
LGTM!

PR #11486, opened 2024-02-02
chore(css): remove `rtl` artifact builds

kennylam:
### Description

Now that CSS logical properties are used throughout the project, there is no need for `rtl` artifacts to be built. This PR removes those builds. 

### Changelog

**New**

- {{new thing}}

**Changed**

- {{changed thing}}

**Removed**

- remove `rtl` builds and artifacts
- remove Storybook `rtl` documentation


annawen1:
I think now that we're removing `rtlcss` we can probably remove the rtl directives in the stylesheets too like this one:
https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/masthead/masthead.scss#L227

I see a bunch of them in the masthead and leadspace styles

ariellalgilmore:
if we add the `rtl` tag will. it trigger a rtl deploy preview?

kennylam:
> if we add the `rtl` tag will it trigger a rtl deploy preview?

Yep!

ariellalgilmore (approved):
Looks great! Thinking too when we switch to storybook 7 we won't even need the deploy previews since we can u the `text-direction` option.

PR #11532, opened 2024-02-13
fix(date picker): detect change range interval console error

### LINKED ISSUE #11490: [cds-date-picker]: Detecting change on range interval
### Description

Hello,

while playing with cds-date-picker component I have discovered there is a issue with change detection.
In dev tools it can bee seen a error Uncaught TypeError: fp.getDateStr is not a function.

Example is right here: https://web-components.carbondesignsystem.com/?path=/docs/components-date-picker--range-with-calendar

valueChanged = function getDateStrFromInputs(dates) {
                        return dates.filter((function(value) {
                            return value
                        }
                        )).filter((function(d, i, arr) {
                            return "range" !== fp.config.mode || fp.config.enableTime || arr.indexOf(d) === i
                        }
                        )).join("range" !== fp.config.mode ? fp.config.conjunction : fp.l10n.rangeSeparator)
                    }([firstInput.value, secondInput.value]) **!== fp.getDateStr()**

Since there is a no function getDateStr on object fp I would suggest to transform fp.selectedDates array to required date format

### Component(s) impacted

cds-date-picker

image

### Browser

Chrome

v.2.2.0

### Severity

### Application/website

https://web-components.carbondesignsystem.com/?path=/story/components-date-picker--range-with-calendar

### Package

@carbon/web-components

### CodeSandbox example

https://web-components.carbondesignsystem.com/?path=/story/components-date-picker--range-with-calendar

Select start and stop date

_No response_

### Code of Conduct

sangeethababu9223:
### Related Ticket(s)

Closes #11490

### Description

In Date picker element, dev tool console shows error `Uncaught TypeError: fp.getDateStr is not a function`.

### Changelog

**Removed**

- Removed custom event handling function.


m4olivei (changes requested):
Looks like what happened between [when I last worked on this](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10603) up until now is that, v2 was merged and it reverted flatpickr from the version I had updated it to, 4.6.13, [back to 4.1.6](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/carbon-web-components/package.json#L73).

I don't have a strong opinion, ultimately we'll likely want to drop flatpickr eventually b/c it's not well maintained, but another option is to set flatpickr to 4.6.13 (note we have [patches](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.yarn/patches/flatpickr-npm-4.6.13-06399933fc.patch) to make it all work). Once set to 4.6.13, everything should work the same as v1.

If we keep this change as is, which I kinda like b/c less code we're maintaining, don't need patches, and seems to work well, we should remove the [patch for 4.6.13](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.yarn/patches/flatpickr-npm-4.6.13-06399933fc.patch).

PR #11541, opened 2024-02-15
docs(data-table): update data-table docs

kennylam:
### Description

Update usage documentation for `data-table` component.

### Changelog

**Changed**

- update `data-table` docs


sangeethababu9223:
Hi @kennylam,
In the Doc it is said that for a second line to be included in a table cell, `size=“xl”` attribute is to be added to the cds-table. 
But it seems to work without the attribute just fine.
And there seems to be some issue with programmatic sorting code in the doc too.

emyarod (approved):
looks good to me!

PR #11554, opened 2024-02-19
feat(tabs-extended): avoid opening to middle of accordion content

jkaeser:
### Related Ticket(s)

none

### Description

Before:

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/12532727/47a9ac11-2959-48fe-8b78-263ff538ff08

After:

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/12532727/a4a6abfc-ca0c-4c27-aaf9-110417fda231

### Changelog

**Changed**

- Ensures opening a `<dds-tabs-extended>` mobile accordion item brings users to the top of the accordion content.


packages/web-components/src/components/tabs-extended/tabs-extended.ts
```diff
@@ -73,11 +78,9 @@ class DDSTabsExtended extends MediaQueryMixin(StableSelectorMixin(LitElement), {
...
       (node) => node instanceof DDSTab
     ) as DDSTab[];
-    this._tabItems.forEach((tab, index) => {
-      this._activeTabIndex = (tab as DDSTab).selected
-        ? index
-        : this._activeTabIndex;
-    });
+    this._activeTabIndex =
+      this._tabItems.findIndex((tab) => (tab as DDSTab).selected) ??
+      this._activeTabIndex;
```
m4olivei, line 83: Seems like since [`findIndex` returns either the index, or -1](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex#return_value), the right hand side of the [nullish coalescing operator ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) would never get evaluated. Can this be simplified to the following?
```suggestion
      this._activeTabIndex = this._tabItems.findIndex((tab) => (tab as DDSTab).selected);
```

packages/web-components/src/components/tabs-extended/tabs-extended.ts
```diff
@@ -86,49 +89,42 @@ class DDSTabsExtended extends MediaQueryMixin(StableSelectorMixin(LitElement), {
   }
 
   private _handleAccordionClick(e) {
-    const tab = e.target.closest('dds-tab');
+    const tab = e.target.closest(`${ddsPrefix}-tab`);
```
m4olivei, line 92: 👍

m4olivei (approved):
Behavior looks good. I like the cleanup / bonus docs along the way. Just noting one minor thing to review, feel free to take / leave and move forward. Thanks!

packages/web-components/src/components/tabs-extended/tabs-extended.ts
jkaeser: Good catch and good call!

andy-blum (approved):
I haven't gone over the code with a fine-toothed comb, but the end result is what we're looking for. Good work!

ariellalgilmore:
just needs a `yarn format` and we can get into the release

ariellalgilmore (approved):

PR #11574, opened 2024-02-22
fix(masthead): prevent extra space beneath megapanel columns in Safari

jkaeser:
### Related Ticket(s)

none

### Description

Fixes an issue that only presents on Safari where megamenu tabpanels are unnecessarily tall.

**Before:**
Screenshot 2024-02-22 at 3 44 56 PM

**After:**
Screenshot 2024-02-22 at 3 44 47 PM

### Changelog

**Changed**

- Prevents masthead megamenu tabpanels from displaying with excessive vertical space in Safari.


jkaeser:
While working on this I realized that something has gone wrong with the focusout handling for megamenus in Safari 💀 The `event.relatedTarget` is always `null`, whereas we expect it to be the element that is losing focus. Will open a separate ticket for that.

m4olivei (approved):

PR #11577, opened 2024-02-23
fix(chat-button): update to latest with no icon option

ariellalgilmore:
### Related Ticket(s)

follow up to https://github.com/carbon-design-system/carbon/pull/15778#event-11890913271

### Description

chat button did not account for not having an icon before. now it does
### Changelog

**New**

- {{new thing}}

**Changed**

- {{changed thing}}

**Removed**

- {{removed thing}}


annawen1 (approved):
LGTM!

PR #11581, opened 2024-02-26
fix(pricing-table-annotation-toggle): increased the size of the annot…

bruno-amorim:
…ation toggle area to make it easier to hit on mobile

### Related Ticket(s)

[ADCMS-4432](https://jsw.ibm.com/browse/ADCMS-4432)

### Description

In the pricing table, the toggle annotation arrow was too small, making it difficult to be hit on mobile resolutions.
Instead of adding more padding to it, I thought it would be better to simply import a bigger version (/internal/vendor/@carbon/web-components/icons/chevron--down/24.js) 

Screenshot 2024-02-26 at 11 54 57


jkaeser (approved):
I like this solution. It's simple and makes the button easier to press. +1 from me!

andy-blum (approved):

kennylam (approved):

PR #11595, opened 2024-02-28
fix(tabs-extended): confusing behavior when in mobile the accordion c…

bruno-amorim:
…annot be closed

### Related Ticket(s)

[ADCMS-4434](https://jsw.ibm.com/browse/ADCMS-4434)

### Description

The design for the Tabs Extended when in mobile is confusing. It turns into an accordion and there's always one item opened and showing. However, that opened item has an arrow on the right hand side, that deceives the user into thinking it's possible to close that item, when in fact, the accordion mimics the desktop behavior for tabs where we always have one item being shown at a time.

**I'm opening this PR initially as draft with the solution of removing the arrow for opened items so the whole team can discuss the best approach here.**

As per discussion with the team, the suggested changes were:

1.     Added the ability to open and close the accordion items;
2.     When all accordion items are closed, in desktop view the first tab is selected as default.

### DIFF
--- a/packages/web-components/src/components/tabs-extended/tabs-extended.ts
+++ b/packages/web-components/src/components/tabs-extended/tabs-extended.ts
@@ -88,6 +88,8 @@ class DDSTabsExtended extends MediaQueryMixin(
 
   private _handleAccordionClick(e) {
     const tab = e.target.closest(`${ddsPrefix}-tab`);
+    // const allClosed = this._tabItems.every((tab) => !tab.selected);
+
     this._handleClick(tab.getIndex(), e);
   }
 
@@ -208,13 +210,10 @@ class DDSTabsExtended extends MediaQueryMixin(
           `${ddsPrefix}-tab:nth-of-type(${_activeTabIndex + 1})`
         )?.shadowRoot?.querySelector(`.${prefix}--accordion__heading`);
 
-    if (activeItemControl instanceof HTMLElement) {
-      if (!isLargeOrGreater) {
-        // Unset focus so that when element is focused programmatically, the
-        // browser scrolls element into view.
-        activeItemControl.blur();
-      }
-      activeItemControl.focus();
+    if (activeItemControl instanceof HTMLElement && isLargeOrGreater) {
+      // Unset focus so that when element is focused programmatically, the
+      // browser scrolls element into view.
+      activeItemControl.blur();
     }
   };
 
--- a/packages/web-components/src/components/tabs-extended/tabs-extended.ts
+++ b/packages/web-components/src/components/tabs-extended/tabs-extended.ts
@@ -236,9 +236,9 @@ class DDSTabsExtended extends MediaQueryMixin(
         (tab as DDSTab).setIndex(index);
 
         //Checking for the presence of 'dds-processed' attr. Since the logic will only succeed once, attaching only one version of the 'click' event at a time when switching views.
-        if (!tab.hasAttribute('dds-processed') && !_isLargeOrGreater) {
+        if (!tab.hasAttribute(`${ddsPrefix}-processed`) && !_isLargeOrGreater) {
           tab.addEventListener('click', this._handleAccordionClick.bind(this));
-          tab.setAttribute('dds-processed', 'true');
+          tab.setAttribute(`${ddsPrefix}-processed`, '');
         }
       });
     }
--- a/packages/web-components/src/components/tabs-extended/tabs-extended.ts
+++ b/packages/web-components/src/components/tabs-extended/tabs-extended.ts
@@ -93,7 +93,7 @@ class DDSTabsExtended extends MediaQueryMixin(
     // Adding the ability to close the accordion. If the tab that's clicked is already active, we pass a bogus number through _setActiveItem() method, so that all tabs are closed.
     if (tab.getIndex() === currentIndex && !this._isLargeOrGreater) {
       this._setActiveItem(-1);
-      // return;
+      return;
     }
 
     this._handleClick(tab.getIndex(), e);
@@ -232,21 +232,21 @@ class DDSTabsExtended extends MediaQueryMixin(
     this._activeTabIndex = parseInt(this._activeTab, 10);
 
     // Reset state and event listeners when switching views
-    // if (changedProperties.has('_isLargeOrGreater')) {
-    //   // Removing event listener if in mobile view, else adding the event listener to any other view that's not mobile.
-    //   if (!_isLargeOrGreater) {
-    //     _tabItems.forEach((tab) => {
-    //       tab.removeEventListener(
-    //         'click',
-    //         this._handleAccordionClick.bind(this)
-    //       );
-    //     });
-    //   } else {
-    //     _tabItems.forEach((tab) => {
-    //       tab.addEventListener('click', this._handleAccordionClick.bind(this));
-    //     });
-    //   }
-    // }
+    if (changedProperties.has('_isLargeOrGreater')) {
+      // Removing event listener if in mobile view, else adding the event listener to any other view that's not mobile.
+      if (!_isLargeOrGreater) {
+        _tabItems.forEach((tab) => {
+          tab.removeEventListener(
+            'click',
+            this._handleAccordionClick.bind(this)
+          );
+        });
+      } else {
+        _tabItems.forEach((tab) => {
+          tab.addEventListener('click', this._handleAccordionClick.bind(this));
+        });
+      }
+    }
 
     if (changedProperties.has('_tabItems')) {
       _tabItems.forEach((tab, index) => {
--- a/packages/web-components/src/components/tabs-extended/tabs-extended.ts
+++ b/packages/web-components/src/components/tabs-extended/tabs-extended.ts
@@ -90,6 +90,11 @@ class DDSTabsExtended extends MediaQueryMixin(
     const tab = e.target.closest(`${ddsPrefix}-tab`);
     const currentIndex = this._activeTabIndex;
 
+    // Adding the ability to close the accordion. If the tab that's clicked is already active, we pass a bogus number through _setActiveItem() method, so that all tabs are closed.
+    if (tab.getIndex() === currentIndex && !this._isLargeOrGreater) {
+      this._setActiveItem(-1);
+    }
+
     // If the tab is already active and selected when clicked, just return
     if (
       tab &&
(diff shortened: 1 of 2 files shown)


areagan1030:
@bruno-amorim Agree that this is a subpar user experience on mobile devices/sm+md breakpoints. I'm wondering if following the Carbon component pattern would be an improvement by having a dropdown selection tied to initial tab display. This might help solve the weirdness of the tab not appearing properly interactive — and then needing to scan down the content to figure out where the next interaction point might be.

I also noticed there was a full-width tabbed component made for the Consulting work that appears to be using this dropdown pattern: [see Consulting](https://www.ibm.com/consulting). The affordance is rather bad on this given the type sizes.

One addition point of conversation may be CIBM v2 — it appears that the mobile display here keeps the main tab setup and uses the standard overflow mechanism across breakpoints to allow users to access additional tab nav items.

andy-blum:
@areagan1030 The consulting page tabs were built custom by Wipro for AEM, those aren't C4IBM or C4AEM

packages/web-components/src/components/tabs-extended/tabs-extended.ts
```diff
@@ -88,6 +88,8 @@ class DDSTabsExtended extends MediaQueryMixin(
 
   private _handleAccordionClick(e) {
     const tab = e.target.closest(`${ddsPrefix}-tab`);
+    // const allClosed = this._tabItems.every((tab) => !tab.selected);
```
andy-blum, line 91: can we remove this?

andy-blum (changes requested):
I'm not sure if it's a regression or not, but I'm seeing the behavior here that John fixed in #11554

packages/web-components/src/components/tabs-extended/tabs-extended.ts
```diff
@@ -208,13 +210,10 @@ class DDSTabsExtended extends MediaQueryMixin(
           `${ddsPrefix}-tab:nth-of-type(${_activeTabIndex + 1})`
         )?.shadowRoot?.querySelector(`.${prefix}--accordion__heading`);
 
-    if (activeItemControl instanceof HTMLElement) {
-      if (!isLargeOrGreater) {
-        // Unset focus so that when element is focused programmatically, the
-        // browser scrolls element into view.
-        activeItemControl.blur();
-      }
-      activeItemControl.focus();
```
jkaeser, line 217: This change is what causes the regression Andy mentioned, @bruno-amorim . We need to programmatically set focus on the active item so the browser scrolls it into view.

jkaeser (changes requested):
Hey, @bruno-amorim! Thanks for your work on this one. It's a really tricky component!

As we discussed, can you try keeping all the `selected` state logic in the `DDSTabsExtended` component rather than adding logic to `DDSTab` elements? I was thinking you might be able to achieve this by adding some logic to the `_handleAccordionClick()` method. Maybe something like this?

```ts
private _handleAccordionClick(e) {
    const tab = e.target.closest(`${ddsPrefix}-tab`);
    const targetIndex = tab.getIndex() === this._activeTabIndex
        ? -1
        : tab.getIndex();
    this._handleClick(targetIndex, e);
  }
```

Additionally, the auto-scroll behavior isn't working quite right. It appears to scroll to the previous active item when opening an accordion.

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/12532727/ddb2ceaf-663a-4b13-a1dd-0bd9960dadf4

bruno-amorim:
Hi @jkaeser @andy-blum 

All issues pointed out by you are now fixed. @jkaeser I believe the scroll into view issue has been fixed as I've tested it multiple times (even adding more copy to the story tests) and it worked just like the video from your original PR.

packages/web-components/src/components/tabs-extended/tabs-extended.ts
```diff
@@ -223,6 +249,23 @@ class DDSTabsExtended extends MediaQueryMixin(
...
+            'click',
+            this._handleAccordionClick.bind(this)
+          );
+        });
+      } else {
+        _tabItems.forEach((tab) => {
+          tab.addEventListener('click', this._handleAccordionClick.bind(this));
+        });
+      }
+    }
```
jkaeser, line 267: Hrm, can you help me understand why we need this? 

My understanding is that when we cross over the `lg` breakpoint and `this._isLargerOrGreater` gets updated, we end up re-rendering the whole component. As a result, `this._tabItems` gets updated, which causes another update cycle. At that point we have fresh DOM elements, add we only the event listener if we're below the breakpoint.

packages/web-components/src/components/tabs-extended/tabs-extended.ts
```diff
@@ -88,7 +88,33 @@ class DDSTabsExtended extends MediaQueryMixin(
...
+
+    // Checking if the clicked tab is different from the previously active tab, if they're different, it just sets the focus to the new clicked tab
+    if (currentIndex !== tab.getIndex()) {
+      const tabButton = tab.shadowRoot?.querySelector(
+        `.${prefix}--accordion__heading`
+      );
+      if (tabButton instanceof HTMLElement) {
+        tabButton.focus();
+      }
+    }
```
jkaeser, line 117: I'm also not sure you need this! We're still setting focus in the `_setActiveItem` method.

packages/web-components/src/components/tabs-extended/tabs-extended.ts
```diff
@@ -261,6 +304,15 @@ class DDSTabsExtended extends MediaQueryMixin(
         }
       });
     }
+
+    if (changedProperties.has('_isLargeOrGreater')) {
+      const { _isLargeOrGreater, _tabItems } = this;
```
jkaeser, line 309: These variables have already been declared at the beginning of this method, so I think you can remove this line.

jkaeser (changes requested):
@bruno-amorim Nice! Functionally the behavior is working exactly as we want. Please see my notes for some questions and code cleanup suggestions.

Thank you so much for digging into this one!

packages/web-components/src/components/tabs-extended/tabs-extended.ts
```diff
@@ -88,7 +88,33 @@ class DDSTabsExtended extends MediaQueryMixin(
...
+    }
+
+    // If the tab is already active and selected when clicked, just return
+    if (
+      tab &&
+      tab.getIndex() === currentIndex &&
+      this._tabItems[currentIndex].selected
+    ) {
+      return;
+    }
```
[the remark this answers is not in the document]
bruno-amorim: Code's been changed as suggested and it's working! Can you please review it now @jkaeser

packages/web-components/src/components/tabs-extended/tabs-extended.ts
bruno-amorim: Code's been remove as suggested and it's working! Can you please review it now @jkaeser

packages/web-components/src/components/tabs-extended/tabs-extended.ts
bruno-amorim: I have removed that line! Can you please review it now @jkaeser

packages/web-components/src/components/tabs-extended/tabs-extended.ts
bruno-amorim: Hey @jkaeser,
I needed to add this piece of code because the accordion items were working the first time they're rendered in mobile, but if you switch the viewport to desktop and then go back to mobile, it stops working. I think the changes in the viewport were not properly attaching the events to the accordion items. When I wrote this code in the updated() method, it fixed that bug by ensuring the proper Event Listeners were added and removed based on the viewport. I took a look at the code several times and could not find out why changing the viewport made the accordion items not to work. Do you see something I'm not seeing? Thanks a lot in advance for all your help!

jkaeser: Buckle up, this is a lot 😆 

I tracked the event listeners on a `<dds-tab>` element as I crossed back and forth over the breakpoint. What's actually happening is that any time we cross the threshold, a new copy of the `click` event listener is added. None are ever removed.

All of the accumulated event listeners fire off their callbacks in rapid succession any time an accordion item is clicked. When there is only one listener, we get the expected result. When there is a second, the functionality breaks; the first callback fires and changes `this._activeTabIndex`, and then the second fires and sets it back to the clicked tab's index value. Add a third listener, and things work again. It keeps going like this, with even numbers of listeners working and odd numbers failing.

Screenshot 2024-03-15 at 3 01 58 PM

The real fix to all of this is making sure only one copy of the event listener is ever attached. In my opinion, this can be best achieved by removing this new chunk of code and modifying the existing logic. Two potential approaches come to mind:

1. A few lines down in the conditional before adding the event listener, you can check for the absence of an attribute (e.g. `dds-processed`) on the `<dds-tab>` elements, then set that attribute after adding the event listener. This way the conditional will only succeed once.
   - ```ts
      if (!tab.hasAttribute('dds-processed') && !_isLargeOrGreater) {
        tab.addEventListener(...);
        tab.setAttribute('dds-processed', true);
      }
      ```
2. You can remove the existing `tab.addEventListener` and instead use an approach called [Event Delegation](https://javascript.info/event-delegation). You'll attach a single click handler to the `<dds-tabs-extended>` element, check if the element is what we're actually looking for, and only then fire off `this._handleAccordionClick`.
   - ```ts
      @HostListener('click')
      protected _handleClick(event) {
        const { target } = event;
        if (target.matches('button.bx--accordion__heading') {
          this._handleAccordionClick(event);
        }
      }
      ```

Your call on whichever looks better! As always, let me know if you want to go over any of this together.

jkaeser (changes requested):

bruno-amorim:
@jkaeser Thank you so much for your last notes on this, it's ready for your review!

packages/web-components/src/components/tabs-extended/tabs-extended.ts
```diff
@@ -227,8 +235,10 @@ class DDSTabsExtended extends MediaQueryMixin(
       _tabItems.forEach((tab, index) => {
         (tab as DDSTab).setIndex(index);
 
-        if (!_isLargeOrGreater) {
+        //Checking for the presence of 'dds-processed' attr. Since the logic will only succeed once, attaching only one version of the 'click' event at a time when switching views.
+        if (!tab.hasAttribute('dds-processed') && !_isLargeOrGreater) {
```
jkaeser, line 239: Nit: use `ddsPrefix` variable.

packages/web-components/src/components/tabs-extended/tabs-extended.ts
```diff
@@ -227,8 +235,10 @@ class DDSTabsExtended extends MediaQueryMixin(
...
           tab.addEventListener('click', this._handleAccordionClick.bind(this));
+          tab.setAttribute('dds-processed', 'true');
```
jkaeser, line 241: Two nits:
1. Use `ddsPrefix` variable.
2. Since we're using this as a boolean attribute, it's a little cleaner to set the attribute to `''`. See https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute. My mistake for recommending the `true` value before!

jkaeser (approved):
@bruno-amorim Excellent work 🎉 🎉 🎉 

I left some nits for you, but none of them should block this from merging.

packages/web-components/src/components/tabs-extended/tabs-extended.ts
kennylam, line 239:
```suggestion
        if (!tab.hasAttribute(`${ddsPrefix}-processed`) && !_isLargeOrGreater) {
```

packages/web-components/src/components/tabs-extended/tabs-extended.ts
kennylam, line 241:
```suggestion
          tab.setAttribute(`${ddsPrefix}-processed`, '');
```

kennylam (approved):

PR #11608, opened 2024-03-01
fix: a11y - ensure side panel closes properly with reduced motion

lee-chase:
## Description

With reduce motion enabled the side panel was left in a state where it was waiting for a transition to end before completing the close. There is not transition with reduced motion, and as a result a transparent pane was left over the rest of the page.

This PR corrects that to ensure the side panel is correctly closed when reduced motion is enabled.

### Changelog

**Changed**

- packages/carbon-web-components/src/components/side-panel/side-panel.ts


annawen1 (approved):
LGTM!

Issue #11605, opened 2024-02-29
[card-group]: broken setSameHeight logic when tag-groups are added


andy-blum:
[JIRA](https://jsw.ibm.com/browse/ADCMS-4691)

### Description

#7315 introduced this bug in 2021.

[Reproduction](https://codepen.io/andy-blum/pen/ZEZExVy/cb4d86ef5b52fa298885472341fe2c3f?editors=1000)

[This chunk](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1/packages/web-components/src/components/card-group/card-group.ts#L201-L224) of code seems to be calculating sameHeight values based on an assumption that tags will be _above_ the body text as the first element sibling of the dds-card-header. Per discussion with @RichKummer and @annawen1 this is _not_ the correct layout, and tags should be placed between body text and the footer link. Any additional space needed to compensate for tags should be between body text and the tag group.

<img width="1041" alt="Screenshot 2024-02-29 at 2 24 38 PM" src="https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/25532785/e77a3543-e5ab-4ffd-af88-d0cb86e121d9">

### Component(s) impacted

* dds-card-group
* dds-card-group-item
* dds-tag-group

PR #11629, opened 2024-03-11
fix(card-group):change tagGroupHeight target Element

marcelojcs:
### Related Ticket(s)

Closes https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/11605
JIRA  [ADCMS-4691](https://jsw.ibm.com/browse/ADCMS-4691) 

### Description

Fix for the  big extra unexpected space on cards between heading and paragraph.

### Changelog

**Changed**

Changing the card group script so it adds `tagGroupHeight`  as margin-botton to paragraphs instead of  headings

### DIFF
--- a/packages/web-components/src/components/card-group/card-group.ts
+++ b/packages/web-components/src/components/card-group/card-group.ts
@@ -218,7 +218,7 @@ class DDSCardGroup extends StableSelectorMixin(LitElement) {
           (this.constructor as typeof DDSCardGroup).selectorItemTagGroup
         )
       ) {
-        e.style.marginBottom = `${tagGroupHeight + headingBottomMargin}px`;
+        e.style.marginBottom = `${tagGroupHeight}px`;
       } else {
         let childTagGroup = e.nextElementSibling;
         childTagGroup.style.marginTop = `${
--- a/packages/web-components/src/components/card-group/card-group.ts
+++ b/packages/web-components/src/components/card-group/card-group.ts
@@ -219,6 +219,11 @@ class DDSCardGroup extends StableSelectorMixin(LitElement) {
         )
       ) {
         e.style.marginBottom = `${tagGroupHeight + headingBottomMargin}px`;
+      } else {
+        let childTagGroup = e.nextElementSibling;
+        childTagGroup.style.marginTop = `${
+          tagGroupHeight - childTagGroup.offsetHeight
+        }px`;
       }
     });
   };
--- a/packages/web-components/src/components/card-group/card-group.ts
+++ b/packages/web-components/src/components/card-group/card-group.ts
@@ -26,6 +26,9 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
 const gridLgBreakpoint = parseFloat(breakpoints.lg.width) * baseFontSize;
 const gridMdBreakpoint = parseFloat(breakpoints.md.width) * baseFontSize;
 
+// tag constants used for same height calculations
+const headingBottomMargin = 16; 
+
 /**
  * Card Group.
  *
@@ -215,7 +218,7 @@ class DDSCardGroup extends StableSelectorMixin(LitElement) {
           (this.constructor as typeof DDSCardGroup).selectorItemTagGroup
         )
       ) {
-        e.style.marginBottom = `${tagGroupHeight}px`;
+        e.style.marginBottom = `${tagGroupHeight + headingBottomMargin}px`;
       } else {
         let childTagGroup = e.nextElementSibling;
         childTagGroup.style.marginTop = `${
--- a/packages/web-components/src/components/card-group/card-group.ts
+++ b/packages/web-components/src/components/card-group/card-group.ts
@@ -27,7 +27,7 @@ const gridLgBreakpoint = parseFloat(breakpoints.lg.width) * baseFontSize;
 const gridMdBreakpoint = parseFloat(breakpoints.md.width) * baseFontSize;
 
 // tag constants used for same height calculations
-const headingBottomMargin = 16; 
+const headingBottomMargin = 16;
 
 /**
  * Card Group.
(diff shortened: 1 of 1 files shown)


andy-blum (changes requested):
It seems like we will need to add some margin to `.bx--card__copy`

https://cdpn.io/pen/debug/OJGNdLq/81e2ba26f85b6a11ba0374b4aa61a0af

Screenshot 2024-03-11 at 1 13 03 PM

**Correction:**

As indicated in #11605, we want tags to be aligned down near the CTA link.

<img src="https://private-user-images.githubusercontent.com/25532785/309054780-e77a3543-e5ab-4ffd-af88-d0cb86e121d9.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTAxNzc1NDcsIm5iZiI6MTcxMDE3NzI0NywicGF0aCI6Ii8yNTUzMjc4NS8zMDkwNTQ3ODAtZTc3YTM1NDMtZTVhYi00ZmZkLWFmODgtZDBjYjg2ZTEyMWQ5LnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDAzMTElMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwMzExVDE3MTQwN1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTU1M2JiZjY5NTY4MDQzYmFjNDI0Nzc4YTNjYjk3OTk1Y2I1ODcyOGQ1MjA2MjUxNjhlOWViMTg5YWQyNGU1YTQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.px0VPcj1yzN3B7v4E-KYBLTU4mGlz4VQj2vSYsnWOQQ"

marcelojcs:
Thanks @andy-blum!!
Addressed in my latest commit. 
image

andy-blum (changes requested):
I've updated the codepen to remove the tag group on one card and the alignments were thrown off. Looks like we still have some work on this one. I'm also not sure why the spacing under the CTA is impacted here


Screenshot 2024-03-13 at 8 34 44 AM

marcelojcs:
@andy-blum 
We were adding too much to the margin of paragraphs without tag-groups siblings.
Fixed now.

andy-blum:
This is looking better to me. Adding the design review tag and seeing if @oliviaflory or @RichKummer can take a look.

packages/web-components/src/components/card-group/card-group.ts
```diff
@@ -219,6 +219,11 @@ class DDSCardGroup extends StableSelectorMixin(LitElement) {
         )
       ) {
         e.style.marginBottom = `${tagGroupHeight + headingBottomMargin}px`;
```
jkaeser, line 221: This margin assignment makes sense if other cards have tag groups and as a result `tagGroupHeight` is greater than zero. But we don't want or need to do this if there are no tag groups since we can just set that spacing in CSS (as a matter of fact, we do – there is a `gap` rule that spaces each element in the default slot `1rem` apart). 

Similarly, if there are no tag groups and no subsequent sibling, we end up adding an extra `16px` margin to the bottom of the paragraph, resulting in `48px` of space between the paragraph text and the Card's CTA. In this case, we want just `32px`, as confirmed in the latest v1 Storybook: https://www.ibm.com/standards/carbon/v1/web-components/iframe.html?args=&id=components-card-group--default&viewMode=story 

The fix is simple, though! We just need to conditionally set this value if `tagGroupHeight` is greater than zero.

jkaeser (changes requested):
@marcelojcs This is looking good! Just one regression that I found.

packages/web-components/src/components/card-group/card-group.ts
marcelojcs: Hey @jkaeser  Thank you for the Review!!!
Addressed  in my latest commit

jkaeser: Thanks for addressing it, @marcelojcs! I'm seeing something weird now, though, and it might predate your changes. There's a negative margin on the `<dds-card-cta-footer>` in the deploy preview, but it's not there in the latest v1 Storybook. This is making the card text and footer CTA too close to each other. Can you investigate?

 ### Deploy preview
Screenshot 2024-03-14 at 4 20 48 PM

### Latest v1 Storybook
Screenshot 2024-03-14 at 4 21 34 PM

marcelojcs: Hey @jkaeser! Nice catch! 
The code for same height for tag groups was running on the NextSibling of the paragraph (In this case was the footer) 
adding a conditional to ensure the code only runs if the next sibling is indeed a tag group fixed the issue.

jkaeser (approved):

packages/web-components/src/components/card-group/card-group.ts
jkaeser: Looking good now, @marcelojcs! Nice work.

oliviaflory (approved):
Thanks @marcelojcs @jkaeser @andy-blum this looks good from a design perspective. I'd like to get the devs to take a look as well

kennylam (approved):
LGTM!

PR #11658, opened 2024-03-20
fix(content-block): add story and fix issues

ariellalgilmore:
### Related Ticket(s)

#11624 (content-block)

### Description

This creates a story for content-block and updates the spacing for the component.
### Changelog

**New**

- Story and data folder with content
**Changed**
- link-list
   - heading had extra styling indenting it at md breakpoint -> this was removed to stay in grid
   - had padding at the bottom if last-of-type this was removed since not necessary in content-block
- content-block
   - lots of style updates
   - bottom and top spacing converted from padding to margins
   - accounting for different paddings at each breakpoint and with/without border
   - all children components are slotted unless the fall in the gutter (i.e. cards and tabs-)
   - when there is a card cta and the child in content-block is a card-group that card cta will align to the grid-mode the card-group is in
   - headings and copies have a max-inline-size
   - moved most of the content-block styles inside the styles package
   - readme updates to explain to adopters it should be in a 12 column grid
- content-group/item
   - both now only have bottom margins if there are multiple together OR if single have no bottom/top margins
- cta-block
   - there was an extra row-gap causing additional spacing
 - leadspace block
    - because of content-block updates needed to adjust some styling to stay in line

**Removed**

- border from content-items in cta-block


RichKummer:
Just a note from earlier discussion:

- Adopters will need to manually add styling to ensure enough spacing between `tabs` and any nested components. This has been a preexisting need.
- The mixin for `content item` in this PR means adopters will need to manually add side spacing to achieve type alignment

Screenshot 2024-04-04 at 3 26 20 PM

emyarod (approved):
looks good to me!

RichKummer (approved):

PR #11659, opened 2024-03-20
fix(button-group): extend button icon styles for button-group-item

annawen1:
### Related Ticket(s)

### Description

Spacing with the icon within the `button` and `button-group` elements in the `cta-block` story was off.

Swapped `button-cta` for `button` instead in `cta-block` story and extended the button icon styles within `button-group-item`

### Changelog

**Changed**

- use `button` instead of `button-cta` in `cta-block`
- extend the button icon styles within `button-group-item` so the spacing styles are added


ariellalgilmore (approved):
LGTM!

PR #11668, opened 2024-03-25
feat(search-with-typeahead/masthead): support initial search term

### LINKED ISSUE #11617: [Masthead]: Unable to set initial search terms
### Description

In v1 Masthead used to have a property `initialSearchTerm` that set the initial terms for search.
In v2 this behaviour seems to be missing and is not documented anywhere. Adding it as we did in v1 does not set it anymore.

We need this for the scenario where the user refreshes the page or is redirected to with search query already set.

Expected behaviour is that we need to initialise the Search field in masthead with a query.

### Component(s) impacted

Masthead, c4d-masthead-composite

### Browser

_No response_

v2.5.0-rc.0

### Severity

### Application/website

ibm.com/docs

### Package

### CodeSandbox example

https://stackblitz.com/edit/github-amfequ?file=index.html

_No response_

ASAP

### Code of Conduct

jkaeser:
### Related Ticket(s)

Closes #11617

### Description

I couldn't find definitive proof this property ever existed on either the Search with Typeahead or Masthead web components. Maybe it existed on react versions of these?

In any case, I've added the requested attribute to `<c4d-search-with-typeahead>` and `<c4d-masthead-composite/container>`. It will set the `value` attribute of the search's text input when users open the search bar, unless the user has previously opened the search and modified the default value. In that case, we respect the user's changes and don't set the `value`. This brings up a potential improvement to cache the user's input and use that instead if they close & reopen the search bar, but I'll leave that for another issue 🙂 

### Changelog

**New**

- Adds `initial-search-term` attribute on both `<c4d-search-with-typeahead>` and `<c4d-masthead-composite/container>` elements that sets default query in search input.
- Documents `initial-search-term` feature in Storybook.

**Changed**

- Adds Storybook knobs for `<c4d-search-with-typeahead>`'s "open on load" (`active`) and `should-remain-open` features.

### Testing

Test the "initial search term" knob on both the [Search with Typeahead](https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/11668/index.html?path=/story/components-search-with-typeahead--default) and [Masthead](https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/11668/index.html?path=/story/components-masthead--default) deploy preview stories.


kennylam (approved):
Wow, I can't believe this has been missing from the WC version all this time. You're right, looks like it only existed in React. Looks great thanks @jkaeser!

PR #11672, opened 2024-03-27
docs(card): update image default-src attribute

kennylam:
### Description

Update `card` react-wrapper docs to use correct image source attribute

### Changelog

**Changed**

- change `defaultSrc` to `default-src`


emyarod (approved):
looks good to me!

PR #11692, opened 2024-04-02
chore(storybook): content-block-card storybook link fix

kennylam:
### Description

Deprecated component name in Storybook story was causing the components to render incorrectly.

### Changelog

**Changed**

- use `c4d-card-footer` instead of `c4d-card-cta-footer`

### DIFF
--- a/packages/web-components/src/components/content-block-simple/__stories__/content-block-simple.stories.ts
+++ b/packages/web-components/src/components/content-block-simple/__stories__/content-block-simple.stories.ts
@@ -1,7 +1,7 @@
 /**
  * @license
  *
- * Copyright IBM Corp. 2020, 2023
+ * Copyright IBM Corp. 2020, 2024
  *
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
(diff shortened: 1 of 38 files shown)


packages/web-components/src/components/content-block-simple/__stories__/content-block-simple.stories.ts
```diff
@@ -93,11 +93,11 @@ export const Default = (args) => {
               cta-type="${ifDefined(ctaType)}"
               href="${ifDefined(href)}">
               <c4d-card-link-heading>${ctaCopy}</c4d-card-link-heading>
-              <c4d-card-cta-footer>
+              <c4d-card-footer>
                 ${ctaType === 'local' ? ArrowRight20({ slot: 'icon' }) : ''}
```
ariellalgilmore, line 97: wondering if we can also clean up instances like this... we no longer need to add `ArrowRight20({ slot: 'icon' })` or other icons, we can just use the `cta-type` prop `<c4d-card-footer cta-type="local">` or in this instance `<c4d-card-footer cta-type="${ctaType}">` which will handle it for us and remove an extra import in the story file

ariellalgilmore (commented):
Looks good nice cleanup! Just a few comments maybe out of scope for this PR

packages/web-components/src/components/cta/__stories__/cta.stories.ts
```diff
@@ -158,14 +158,14 @@ export const Default = (args) => {
                     ? html` <c4d-card-heading>${heading}</c4d-card-heading> `
                     : ''}
                   ${ctaType !== 'video' ? copy : ''}
-                  <c4d-card-cta-footer
+                  <c4d-card-footer
```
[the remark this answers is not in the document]
kennylam: Oh interesting, good catch. I totally forgot there are sub-components for `cta`. I didn't see this component in `card` so I assumed it was removed for v2. I'll put it back into the `cta` story.

packages/web-components/src/components/content-block-simple/__stories__/content-block-simple.stories.ts
kennylam: Actually, you're right. But it doesn't seem like `type` even needs to be added since it's already in the parent (L93)?

ariellalgilmore (approved):
Thanks! LGTM

PR #11693, opened 2024-04-02
fix(search-with-typeahead): correct RTL result strings

### LINKED ISSUE #11342: [web-components]: RTL issues on masthead
### Description

there are several issues found for RTL languages:
1. user can not see the inserted string in search field and also the placeholder is not entirely visible
2. the suggestions are wrongly displayed (wrong order of letters, I will add a screenshot)
3. the position of x button and search icon is not correct 

image

also another small issue for the custom typeahead (not rtl), it is bolding the first letter in the search text also in the title of the section, which is wrong. For example here I typed in cloud, and is bolding all c in the title of the section in suggestions:

Screenshot 2024-01-04 at 16 51 05

### Component(s) impacted

masthead

### Browser

Chrome

v2.0.0-rc.3

### Severity

### Application/website

ibm docs

### Package

### CodeSandbox example

-

_No response_

_No response_

### Code of Conduct

jkaeser:
### Related Ticket(s)

Closes #11342 

### Description

This fixes two issues:
1. For each search result, we split it on a substring that matches the current search query, bold that part, then reconstruct the string. The issue is that the strings were reconstructed assuming an LTR text display, and we need to reverse them for RTL.
2. Group titles in search results use the same web component as regular search results, and as a result any portion of their text that matched the current query would get highlighted (bolded). We can avoid this by skipping such elements in the highlighting logic.

The related issue also reported the search actions are not aligned properly, but I was not able to reproduce. Looks like we shifted to using CSS logical properties not long before the issue was opened, which I think solved the problem.

### Changelog

**New**

- Adds and configures `storybook-addon-rtl` package, which adds a toggle in Stories to flip to RTL display.

**Changed**

- Ensures search result strings with highlighted portions are assembled in the correct order depending on language direction.
- Prevents search result group titles from being highlighted as if they were query matches.
- Adds knob in Storybook to demonstrate grouped search results.


kennylam (approved):
This is great, thanks @jkaeser!

PR #11696, opened 2024-04-04
fix(leadspace): be explicit about svg closing tags

### LINKED ISSUE #11690: [Leadspace RTL]: The gradient on Lead space component should be mirrored in RTL
### Description

Anytime a gradient is used in the Leadspace in RTL, it's direction is incorrect, such that the text is hard to read against an image background, or the video is obscured by the mostly opaque side of the gradient.

image

image

### Browser

_No response_

latest

### Severity

### Application/website

https://ibm.com

### Package

### CodeSandbox example

https://carbon-design-system.github.io/carbon-for-ibm-dotcom/next/web-components/?path=/story/components-lead-space--tall-with-image

Navigate to any Lead space story that uses a gradient. Not sure if there is a publically hosted RTL version of the Storybook. Locally I've been:

* Checkout the `main` branch locally
* `yarn build-storybook:rtl`
* Start a server for `storybook-static-rtl/index.html`
* Browse to the Lead space > Tall with image Story

_No response_

### Code of Conduct

andy-blum:
### Related Ticket(s)

Closes #11690 

https://jsw.ibm.com/browse/ADCMS-4848

### Description

Updates leadspace background gradient stops to use explicit closing tabs. Previously gradient stops were authored as self-closing tabs which seem to render as nested stops

Screenshot 2024-04-04 at 3 03 43 PM

### Testing Instructions

Replace the tagged releases in [this codepen](https://codepen.io/andy-blum/pen/oNOEgxN/c58d6ab48ee80510e0c8124a535e7e5e) with the preview CDN URLs

### Changelog

**Changed**

- Fixes a bug where leadspace gradients don't render correctly

### DIFF
--- a/packages/web-components/.storybook/main.js
+++ b/packages/web-components/.storybook/main.js
@@ -217,7 +217,6 @@ module.exports = {
         test: /\.scss$/,
         sideEffects: true,
         use: [
-          'cache-loader',
           require.resolve('../tools/css-result-loader'),
           {
             loader: 'postcss-loader',
--- a/packages/web-components/.storybook/main.js
+++ b/packages/web-components/.storybook/main.js
@@ -248,6 +248,7 @@ module.exports = {
               sourceMap: useStyleSourceMap,
               webpackImporter: false,
               sassOptions: {
+                outputStyle: 'expanded',
                 includePaths: [
                   path.resolve(__dirname, '..', 'node_modules'),
                   path.resolve(__dirname, '../../..', 'node_modules'),
--- a/packages/web-components/.storybook/main.js
+++ b/packages/web-components/.storybook/main.js
@@ -217,7 +217,7 @@ module.exports = {
         test: /\.scss$/,
         sideEffects: true,
         use: [
-          'cache-loader',
+          { loader: 'cache-loader', options: { compare: () => false } },
           require.resolve('../tools/css-result-loader'),
           {
             loader: 'postcss-loader',
--- a/packages/web-components/.storybook/main.js
+++ b/packages/web-components/.storybook/main.js
@@ -248,7 +248,6 @@ module.exports = {
               sourceMap: useStyleSourceMap,
               webpackImporter: false,
               sassOptions: {
-                outputStyle: 'expanded',
                 includePaths: [
                   path.resolve(__dirname, '..', 'node_modules'),
                   path.resolve(__dirname, '../../..', 'node_modules'),
--- a/packages/web-components/.storybook/main.js
+++ b/packages/web-components/.storybook/main.js
@@ -217,7 +217,7 @@ module.exports = {
         test: /\.scss$/,
         sideEffects: true,
         use: [
-          { loader: 'cache-loader', options: { compare: () => false } },
+          'cache-loader',
           require.resolve('../tools/css-result-loader'),
           {
             loader: 'postcss-loader',
--- a/packages/web-components/.storybook/main.js
+++ b/packages/web-components/.storybook/main.js
@@ -217,6 +217,7 @@ module.exports = {
         test: /\.scss$/,
         sideEffects: true,
         use: [
+          'cache-loader',
           require.resolve('../tools/css-result-loader'),
           {
             loader: 'postcss-loader',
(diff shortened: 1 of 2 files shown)


kennylam (commented):
Weirdly, I can't reproduce this in FF/Chrome. In both of those browsers the closing `</stop>` gets rendered correctly.

But since I curious, I did a little digging trying to understand why this would happen. It seems we incorrectly used the `svg` tag function for this [entire block](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/b6bafb2f90af4b902fefff6712d3e370290668a8/packages/web-components/src/components/leadspace/leadspace.ts#L163-L194) (probably in other instances too).

According to the [docs](https://lit.dev/docs/api/templates/#svg) (`Details` section), an `svg` element shouldn't be wrapped in the `svg` function at all, only its fragments. Which means it should probably look something like this?

```typescript
  render() {
    const { gradientStyleScheme, type, size } = this;
    const svgGradientType =
      type === LEADSPACE_TYPE.CENTERED
        ? svg`
        <stop offset="0%" />
        <stop offset="54%" />
        <stop offset="77%" />
        <stop offset="100%" />
      `
        : svg`
        <stop offset="0%" />
        <stop offset="25%" />
        <stop offset="50%" />
        <stop offset="75%" />
      `;
    return html`
      <section class="${this._getTypeClass()}" part="section">
        <div class="${prefix}--leadspace__container">
          <div class="${this._getGradientClass()}">
            ${gradientStyleScheme === LEADSPACE_GRADIENT_STYLE_SCHEME.NONE
              ? undefined
              : html`
                  <svg
                    class="${prefix}--leadspace__gradient"
                    viewBox="0 0 100 100"
                    preserveAspectRatio="none"
                    xmlns="http://www.w3.org/2000/svg"
                    xmlns:xlink="http://www.w3.org/1999/xlink">
                    <defs>
                      <linearGradient
                        id="stops"
                        class="${prefix}--leadspace__gradient__stops"
                        gradientTransform="${type === LEADSPACE_TYPE.CENTERED
                          ? 'rotate(90)'
                          : ''}">
                        ${svgGradientType}
                      </linearGradient>
                    </defs>
                    <rect
                      class="${prefix}--leadspace__gradient__rect"
                      width="100"
                      height="100" />
                  </svg>
                `}
```

andy-blum:
Codepen for verification: https://codepen.io/andy-blum/pen/eYoPJOO/26189acfec3414ad0afc9ac6831f7f23

m4olivei:
I had been trying in the last few commits to address the wrong direction on the gradient for the [RTL Storybook build](https://ibmdotcom-web-components-rtl.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/11696/index.html?path=/story/components-lead-space--medium-with-image). Locally, I found the fix was https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/11696/commits/169524a35af7fa4086f8be0e49d12418afad13ce . The core issue there is that in the Webpack config, the `sass-loader` would minify and strip comments by default. `postcss` and `rtlcss` comes after the `sass-loader`, so `rtlcss` couldn't handle the directives b/c they weren't there. By setting the `outputStyle: 'expanded'` option, `sass-loader` would leave the comments in, allowing `rtlcss` to do the proper processing. At least locally, that was my finding. I'm not sure why the RTL Storybook here didn't reflect that. In any case, the primary issue here for us is the CDN builds.

The CDN builds have been fixed by using explicit end tags for the `<stop>` tags in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/11696/commits/24fb70dc8162fcef909f337e923bc2ec3a69c245. That means Roolup parses it correctly and we get proper syntax for the `<stop>` tags, and all is well. Rollup doesn't have the same problem with `rtlcss` as Webpack does, so the CDN builds are looking good, shown by @andy-blum 's CodePen:

https://codepen.io/andy-blum/pen/eYoPJOO/26189acfec3414ad0afc9ac6831f7f23

m4olivei (approved):

IgnacioBecerra (approved):
LGTM!

PR #11734, opened 2024-04-22
fix(masthead): ensure opened L1 mobile dropdown is visible while there is also a ToC

jkaeser:
### Related Ticket(s)

none

### Description

Given the page uses both an L1 menu and a table of contents (ToC),
and the page is displaying at a mobile breakpoint,
and the ToC menu is the sticky element (i.e. stuck to the top of the viewport),
and the user scrolls up slightly to reveal the L1 dropdown,
and the user open the L1's dropdown menu,
then the dropdown menu gets hidden above the viewport, forcing the user to scroll up more to access it.
Instead, the top of L1 dropdown menu should be visible when opened.

This problem is occurring because the dropdown exists in the document flow, and the StickyHeader utility doesn't reset stuck elements' positions when the menu is opened. 

The easiest solution I found was to position the dropdown menu absolutely like most other dropdown menus you might encounter. This bypasses the tricky problem of trying to get the StickyHeader to recalculate, and in my opinion is a slight enhancement in both performance (avoids layout shifts and associated repaint costs) and overall UX (gives dropdown a more expected behavior).

### Changelog

**Changed**

- Prevents mobile L1 dropdown menu from becoming hidden above viewport when opened while there is currently a sticky table of contents menu.
- Prevents L1 from scrolling away when a submenu is open on desktop or the dropdown is open on mobile.
- Consolidates StickyHeader utility tracked data to make debugging easier.
- Refactors StickyHeader logic to support all potentially sticky elements being on page at the same time.

### Testing
To test the main fix:
1. Visit [the Dotcom Shell > With L1 story](https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/11734/index.html?path=/story/components-dotcom-shell--with-l-1) in the deploy preview.
2. Change the preview size to "Large mobile" or "Small mobile"
3. Scroll down the page until the ToC is stuck to the top of the viewport.
4. Scroll back up just enough to see the L1.
5. Open the L1 and verify the top of it doesn't jump back up above the viewport.
6. Scroll down and verify the L1 stays stuck to the top of the viewport.
7. Close the L1 and verify the ToC gets stuck to the top of the viewport as you scroll down.

You should also test the L1 remains stuck when open at desktop widths.

This PR also changes a good amount of the StickyHeader class logic. To support regression testing the existing sticky element behaviors, I've added the [Sticky Element Sandbox story](https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/11734/index.html?path=/story/components-dotcom-shell--sticky-element-sandbox). Test various knob configurations at both desktop and mobile viewport sizes and let me know how you were able to break it 😄 

**Note:** If you find the sticky elements aren't working correctly when switching between stories, try reloading the window.

### DIFF
--- a/packages/utilities/src/utilities/StickyHeader/StickyHeader.js
+++ b/packages/utilities/src/utilities/StickyHeader/StickyHeader.js
@@ -350,7 +350,7 @@ class StickyHeader {
     } else {
       // In cases where we have both an eligible ToC and leadspace search, we want
       // the ToC to take precedence. Scroll away leadspace search.
-      if (tocIsAtSearch && tocShouldStick) {
+      if (searchIsAtTop && tocIsAtSearch && tocShouldStick) {
         this._state.maxScrollaway += leadspaceSearchBar.offsetHeight;
       }
 
--- a/packages/utilities/src/utilities/StickyHeader/StickyHeader.js
+++ b/packages/utilities/src/utilities/StickyHeader/StickyHeader.js
@@ -317,7 +317,7 @@ class StickyHeader {
       tableOfContents,
       tableOfContentsInnerBar,
       leadspaceSearchBar,
-    } = StickyHeader.global._elements;
+    } = this._elements;
 
     // Reset the value before performing any further calculations.
     this._state.maxScrollaway = 0;
@@ -380,8 +380,8 @@ class StickyHeader {
       masthead,
       tableOfContentsInnerBar: tocInner,
       leadspaceSearchBar,
-    } = StickyHeader.global._elements;
-    const { scrollPosPrevious: oldY } = StickyHeader.global._state;
+    } = this._elements;
+    const { scrollPosPrevious: oldY } = this._state;
 
     /**
      * Reset to a value that is equal to the difference between the previous
--- a/packages/utilities/src/utilities/StickyHeader/StickyHeader.js
+++ b/packages/utilities/src/utilities/StickyHeader/StickyHeader.js
@@ -434,8 +434,8 @@ class StickyHeader {
    * down.
    */
   _manageStickyElements() {
-    const { localeModal } = StickyHeader.global._elements;
-    const { scrollPos: scrollPosPrevious } = StickyHeader.global._state;
+    const { localeModal } = this._elements;
+    const { scrollPos: scrollPosPrevious } = this._state;
 
     // Exit early if locale modal is open.
     if (localeModal && localeModal.hasAttribute('open')) return;
--- a/packages/utilities/src/utilities/StickyHeader/StickyHeader.js
+++ b/packages/utilities/src/utilities/StickyHeader/StickyHeader.js
@@ -229,7 +229,8 @@ class StickyHeader {
 
     /**
      * maxScrollaway is a calculated value matching the height of all components
-     * that are allowed to hide above the viewport.
+     * that are allowed to hide above the viewport. I.e., adding an item's height
+     * to this value indicates we expect it to be hidden above the viewport.
      *
      * We should only have one sticky header showing as the page scrolls down.
      *
@@ -253,13 +254,12 @@ class StickyHeader {
       tocShouldStick =
         toc.layout === 'horizontal' || window.innerWidth < gridBreakpoint;
 
-      if (masthead && tocIsAtTop && (tocShouldStick || mastheadL1)) {
+      // If we have a masthead and the TOC should be the stuck element, hide masthead
+      // above viewport. Otherwise, if there is an L1, just hide the L0. If there's only
+      // the L0, do nothing.
+      if (masthead && tocIsAtTop && tocShouldStick) {
         maxScrollaway += masthead.offsetHeight;
-
-        if (mastheadL1 && !tocShouldStick) {
-          maxScrollaway -= mastheadL1.offsetHeight;
-        }
-      } else if (mastheadL0 && mastheadL1) {
+      } else if (mastheadL1) {
         maxScrollaway += mastheadL0.offsetHeight;
       }
     }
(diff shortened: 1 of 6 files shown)


packages/utilities/src/utilities/StickyHeader/StickyHeader.js
```diff
@@ -18,20 +18,29 @@ const gridBreakpoint = parseFloat(breakpoints.lg.width) * baseFontSize;
...
-    this._leadspaceWithSearchStickyThreshold = 0;
-    this._localeModal = undefined;
-    this._masthead = undefined;
-    this._mastheadL0 = undefined;
-    this._mastheadL1 = undefined;
-    this._tableOfContents = undefined;
-    this._tableOfContentsInnerBar = undefined;
-    this._tableOfContentsLayout = undefined;
+
+    this._data = {
```
m4olivei, line 22: I liked your suggestion yesterday that this be named `_state`

packages/utilities/src/utilities/StickyHeader/StickyHeader.js
```diff
@@ -290,76 +382,62 @@ class StickyHeader {
...
-          });
-        }
+  /**
+   * Manages which elements are stuck and where they are positioned. We should
+   * only have one element stuck to the top of the viewport as the page scrolls
+   * down.
+   */
+  _manageStickyElements() {
+    const { localeModal } = StickyHeader.global._elements;
+    const { scrollPos: scrollPosPrevious } = StickyHeader.global._data;
```
m4olivei, line 424: I'm curious why we're using the singleton here and not just `this`, eg. 
```suggestion
    const { localeModal } = this._elements;
    const { scrollPos: scrollPosPrevious } = this._data;
```
Seems like all the callees, including event handlers, are properly bound to `this` no? Same question goes for other uses of `StickyHeader.global`, which look like all stem from this method. The rest of this method uses `this` directly, so seems like a safe change here to be consistent.

m4olivei (commented):
Couple small things noticed from a static review.

Testing the Dotcom Shell > With L1 I noticed a couple things:
* In the gap before we get to Search / TOC, should the L1 be sticky? I'm seeing nothing stick there.
* The Search box is jumpy when it starts sticking on the way down. I'm seeing that we scroll all the way past it before it begins sticking, and when it does it sort of jumps into place. Similar on the way back up the page, it jumps back to it's spot before we get there.  I can't remember, but maybe this was the piece that b/c of content guidance, this isn't a realistic test?
* Here is a video showing both issues: 

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/assets/191049/d378c754-f667-4e71-a559-e0a9b82df8e4

m4olivei:
> Thanks for the review, @m4olivei! I'm seeing the same With L1 issue. Will get a fix up for that one. What do you think about calling the leadspace search animation out of scope for this PR, though?

I'm fine with that if it's an existing issue.

packages/utilities/src/utilities/StickyHeader/StickyHeader.js
```diff
@@ -171,116 +186,207 @@ class StickyHeader {
...
+   */
+  _calculateMaxScrollaway() {
+    const {
+      masthead,
+      mastheadL0,
+      mastheadL1,
+      tableOfContents,
+      tableOfContentsInnerBar,
+      leadspaceSearchBar,
+    } = StickyHeader.global._elements;
```
m4olivei, line 320: Same note here as in `_manageStickyElements`. Since this method is called exclusively from `_manageStickyElements`, should be safe by the same rationale to use `this`.

packages/utilities/src/utilities/StickyHeader/StickyHeader.js
```diff
@@ -171,116 +186,207 @@ class StickyHeader {
...
+   * revealed when scrolling back up.
+   */
+  _positionElements() {
+    const {
+      banner,
+      masthead,
+      tableOfContentsInnerBar: tocInner,
+      leadspaceSearchBar,
+    } = StickyHeader.global._elements;
+    const { scrollPosPrevious: oldY } = StickyHeader.global._state;
```
m4olivei, line 384: Same note here as in `_manageStickyElements`. Since this method is called exclusively from `_manageStickyElements`, should be safe by the same rationale to use `this`.

m4olivei (commented):
Couple other instances of `StickyHeader.global`.

andy-blum:
closing/re-opening to try to update the jenkins jobs

jkaeser:
@m4olivei Fixed the L1 issue in [`5e0077f`](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/11734/commits/5e0077f66980125758ec8c0f19e0feca9e961af8). You can see it working here: https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/11734/iframe.html?knob-Has%20Masthead%20L1=true&knob-Has%20Leadspace%20With%20Search=true&id=components-dotcom-shell--sticky-element-sandbox&viewMode=story&knob-Has%20Universal%20Banner=false.

m4olivei:
> @m4olivei Fixed the L1 issue in [`5e0077f`](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/11734/commits/5e0077f66980125758ec8c0f19e0feca9e961af8). You can see it working here: https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/11734/iframe.html?knob-Has%20Masthead%20L1=true&knob-Has%20Leadspace%20With%20Search=true&id=components-dotcom-shell--sticky-element-sandbox&viewMode=story&knob-Has%20Universal%20Banner=false.

Looks great!

m4olivei (approved):

Issue #11725, opened 2024-04-16
[web components]: CTA buttons display arrow pointing in the wrong direction in RTL languages


m4olivei:
### Description

Any CTA using `cta-type="local"` makes use of a `ArrowRight20` icon. This is fine for LTR languages, but doesn't work in RTL languages and should be flipped in the opposite direction.

### Component(s) impacted

Any CTA using `cta-type="local"`

### Browser

Chrome

### Carbon for IBM.com version

Affects both v1.42.0 and v2.7.0.

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

https://ibm.com

### Package

@carbon/ibmdotcom-web-components, @carbon/web-components

### CodeSandbox example

v1: https://codepen.io/m4olivei/pen/ZEZMrgM
v2: https://codepen.io/m4olivei/pen/VwNGXKa

### Steps to reproduce the issue (if applicable)

See description and CodePen

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues) for duplicate issues

PR #11748, opened 2024-04-26
[v1] feat(rtl-arrow): Fix arrows used for CTA's when in RTL

m4olivei:
### Related Ticket(s)

Closes https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/11725
[Jira Ticket](https://jsw.ibm.com/browse/ADCMS-4922)

### Description

Checks `document.dir` and uses the appropriate arrow icon for the text direction (right in LTR and left in RTL). For many stories involving CTA's, CTA derivatives, or buttons, avoid using the slotted icon when all we're going for is `cta-type="local"`, in other words the arrow icon. In non-CTA cases, there isn't support for a default arrow icon, so we adjust stories to slot the right icon base on `document.dir`.

Note that in v1, only 'CTA' components have the `CTAMixin`, whereas it's more broadly used in v2. This means that for non-CTA components its up to the client to slot the right arrow icon (as evidenced by the many `*.stories.ts` changes). This isn't great, but it felt OK for now 🤷 . Open to ideas however.

### Testing instructions

1. Navigate to any Story making use of a CTA and/or button
2. Compare LTR and RTL
3. In LTR, where the arrow icon is used, the arrow should be pointing right
4. In RTL, where the arrow icon is used, the arrow should be pointing left

### Changelog

**Changed**

- Change the direction of the CTA arrow when in RTL languages.

### DIFF
--- a/packages/web-components/src/components/link-with-icon/link-with-icon.ts
+++ b/packages/web-components/src/components/link-with-icon/link-with-icon.ts
@@ -15,6 +15,8 @@ import { ICON_PLACEMENT } from '../../globals/defs';
 import StableSelectorMixin from '../../globals/mixins/stable-selector';
 import styles from './link-with-icon.scss';
 import { carbonElement as customElement } from '../../internal/vendor/@carbon/web-components/globals/decorators/carbon-element.js';
+import CTAMixin from '../../component-mixins/cta/cta';
+import { CTA_TYPE } from '../cta/defs';
 
 export { ICON_PLACEMENT };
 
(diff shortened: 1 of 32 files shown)


packages/web-components/src/components/link-with-icon/link-with-icon.ts
```diff
@@ -29,7 +31,10 @@ const { stablePrefix: ddsPrefix } = ddsSettings;
  * @slot icon-left - The CTA icon to place at the left.
  */
 @customElement(`${ddsPrefix}-link-with-icon`)
-class DDSLinkWithIcon extends StableSelectorMixin(BXLink) {
+class DDSLinkWithIcon extends CTAMixin(StableSelectorMixin(BXLink)) {
```
m4olivei, line 34: I've added the `CTAMixin` here so that you could set `cta-type` on a `dds-link-with-icon` or it's derrivatives and have the icon rtl logic decide on the right arrow. I _think_ that's OK based on the fact that this [change was made in v2](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/link-with-icon/link-with-icon.ts#L34), but open to thoughts. Alternatively, we could leave it and let the onus be on the client to slot in the right icon, but that's a real pain.

jkaeser: I'm in favor of this change. Makes me wonder if we should give `<dds-card-footer>` the same treatment.

jkaeser (approved):
LGTM!

andy-blum (approved):
LGTM

m4olivei:
@andy-blum I _think_ the jobs that are supposed to be passing are all good now? The `ci-check / a11y` job is also failing on `v1` due to dependency issues looks like.

PR #11759, opened 2024-05-01
fix(parent-visibility): decouple tab from parent visibility

### LINKED ISSUE #11749: [parent-visibility-mixin]: dds-video imports dds-tab component
https://jsw.ibm.com/browse/ADCMS-4975

### Description

Importing the dds-video-player component brings along dds-tab for the ride.

* [DDSVideoPlayer imports ParentVisibilityMixin](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/video-player/video-player.ts#L27)
* [ParentVisibility imports DDSTab](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/component-mixins/parent-visibility/parent-visibility.ts#L12)

### Component(s) impacted

Any component using the parent-visibility mixin or extending/importing one of those components.

### CodeSandbox example

https://codepen.io/andy-blum/pen/abxMKRm/10cd7d5f6087037b036dbb833767de49

m4olivei:
### Related Ticket(s)

Closes #11749

### Description

This decouples `C4DTab` from `ParentVisibilityMixin`, which avoids the sideeffect of registering the `c4d-tab` component, when you're otherwise just making use of `ParentVisibilityMixin`.

### Changelog

**Changed**

- Avoids referencing `C4DTab` from `ParentVisibilityMixin` by duplicating the event name.


packages/web-components/src/component-mixins/parent-visibility/parent-visibility.ts
```diff
@@ -45,15 +44,25 @@ const ParentVisibilityMixin = <T extends Constructor<HTMLElement>>(Base: T) => {
...
       };
     }
+
+    /**
+     * The name of the custom event that fires whenever a tab's selected state
+     * changes.
+     */
+    static get eventTabSelected() {
+      return `${c4dPrefix}-tab-selected`;
+    }
```
jkaeser, line 65: Do you think it would be too over-engineered to define this string in `/src/components/tabs-extended/defs.ts` then import that value into both this file and the tab source file? I'm worried about the event name changing in the future and there being no indication that it needs to change here, too, causing things to silently break.

m4olivei: I like that suggestion, lets do that.

I was going off existing precedant in places like the `video-player` component where event names are repeated like this (see below). But I agree with you that it's better to have it be defined in one place.

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/video-player/video-player.ts#L288-L290
https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/video-player/video-player-composite.ts#L320-L322

packages/web-components/src/components/tabs-extended/tab.ts
```diff
@@ -30,7 +31,7 @@ class C4DTab extends StableSelectorMixin(CDSTab) {
    * An event that fires whenever the tab's selected state changes.
    */
   static get eventTabSelected() {
```
m4olivei, line 33: Note that since we're defining this event name in `defs.ts`, we could _probably_ get rid of this getter. I'm intentionally leaving it here though on the off chance an adopter is referencing it.

jkaeser: From a purely theoretical standpoint, I like having the getter. It allows extending classes to change the event name.

jkaeser (commented):
The code changes here look good to me, but this is failing ci-check. Can you look into that, @m4olivei ?

m4olivei:
> The code changes here look good to me, but this is failing ci-check. Can you look into that, @m4olivei ?

Huh, interesting, appraently there is a rule that requires `static get` returns to _only_ be a string or string literal:

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/tools/babel-plugin-create-react-custom-element-type.js#L186-L190

jkaeser (changes requested):
Okay, now that we have a representative deploy preview, I've found an issue. `<dds-cta-block-item-row>` uses `ParentVisibilityMixin`, and it's not rendering properly within tabs (see https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/11759/index.html?path=/story/components-cta-block--within-tabs).

Screenshot 2024-05-03 at 1 20 39 PM

m4olivei:
> Okay, now that we have a representative deploy preview, I've found an issue. `<dds-cta-block-item-row>` uses `ParentVisibilityMixin`, and it's not rendering properly within tabs (see https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/11759/index.html?path=/story/components-cta-block--within-tabs).
> 
> Screenshot 2024-05-03 at 1 20 39 PM

Thanks @jkaeser . Looks like [`main` has the same problem](https://carbon-design-system.github.io/carbon-for-ibm-dotcom/next/web-components/?path=/story/components-cta-block--within-tabs). I'll look into it.

m4olivei:
It turns out that the Tabs Extended component changed structure such that the tab component is no longer a parent to the contents of each tab panel. As such, the `ParentVisibilityMixinImpl` doesn't work in v2 and a new paradigm is needed for doing the same.

In addition, the story for [CTA block > Within tabs](https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/11759/index.html?path=/story/components-cta-block--within-tabs) was not correctly updated in v2 for the new tabs strucutre alluded to above. 

All that is to say, we're now out of scope for the original ask and we kind of need to take this back to the drawing board for v2, which we'll do in a follow up issue.

Closing for now.

PR #11796, opened 2024-05-17
feat(back-to-top-btn): adding shadow part so that the component can b…

bruno-amorim:
…e styled from outside shadow

### Related Ticket(s)

[ADCMS-5035](https://jsw.ibm.com/browse/ADCMS-5035)

### Description

Adding a shadow "part" attribute so that the Back to Top Component can be styled from outside the shadow tree.

### Changelog

New shadow part attribute added to Back to Top component.
Snapshots updated.

### DIFF
--- a/packages/web-components/src/components/back-to-top/back-to-top.ts
+++ b/packages/web-components/src/components/back-to-top/back-to-top.ts
@@ -169,6 +169,7 @@ class C4DBackToTop extends HostListenerMixin(StableSelectorMixin(LitElement)) {
     const { backToTopAssistiveText, _handleOnClick: handleOnClick } = this;
     return html`
       <button
+        part="back-to-top__button"
         class="${prefix}--btn ${prefix}--btn--secondary ${prefix}--btn--icon-only ${prefix}--back-to-top__btn"
         aria-label="${backToTopAssistiveText}"
         @click="${handleOnClick}">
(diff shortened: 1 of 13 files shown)


jkaeser (approved):
Looks good! If I'm being nitpicky, the part name looks like it wants to match the CSS class name on the same element, but it differs ever so slightly from `__btn` to `__button`. I don't know how much that actually matters, but figured I'd call it out. Not a blocker IMO.

packages/web-components/src/components/back-to-top/back-to-top.ts
```diff
@@ -169,6 +169,7 @@ class C4DBackToTop extends HostListenerMixin(StableSelectorMixin(LitElement)) {
     const { backToTopAssistiveText, _handleOnClick: handleOnClick } = this;
     return html`
       <button
+        part="back-to-top__button"
```
andy-blum, line 172:
```suggestion
        part="button"
```

andy-blum (changes requested):
Let's not put the entire component name in the part, and also avoid underscores.

I think the main thing I'm looking for in these PRs (especially on larger components) is that the part attributes are paralleling the "element" in BEM classes.

packages/web-components/src/components/back-to-top/back-to-top.ts
bruno-amorim: Created a new commit using this cleaner nomenclature. I'm going to update my other PRs as well.

andy-blum (approved):

PR #11802, opened 2024-05-20
 feat(card): add shadow parts

marcelojcs:
### Related Ticket(s)
[JIRA](https://jsw.ibm.com/browse/ADCMS-5041)

### Description

All non-slot elements in the shadow DOM should be given a unique "part" name allowing CSS to target and override component default styles.

targeting card component

### Changelog

**New**

Adding the part attribute to the shadow dom elements for the card component.


packages/web-components/src/components/card/card.ts
```diff
@@ -155,7 +156,7 @@ class C4DCard extends CTAMixin(StableSelectorMixin(CDSLink)) {
   protected _renderDisabledLink() {
     const { _classes: classes } = this;
     return html`
-      <div id="link" class="${classes}">${this._renderInner()}</div>
+      <div id="link" class="${classes}" part="link">${this._renderInner()}</div>
```
jkaeser, line 159: It concerns me that this is getting the same part name as the `<a>` added in the `render` method. Can you change this to something like `part="disabled-link"`?

jkaeser (changes requested):
One small change request, otherwise looks great!

packages/web-components/src/components/card/card.ts
marcelojcs: Hi @jkaeser! 
Thanks for the review!!! 
Addressed in my latest commit!

jkaeser (approved):

PR #11804, opened 2024-05-20
fix (structured-list): Enable custom max-cols values

andy-blum:
### Related Ticket(s)

https://jsw.ibm.com/browse/ADCMS-5014

### Description

This PR moves the `--max-cols` definitions to the top-level `structured-list` and `pricing-table` components. Previously, each row element set this value, leading to `lg` breakpoints always calculating against an assumed value of 16-cols

This change was needed to make the table work more intuitively in situations where it isn't displayed in a full 16-column container, for example within the `table-of-contents` component.

Before this PR, a pricing table 12-cols wide, with three table-columns each set to be 4-cols wide would set each table-column to be 25% width (4/16).
Screenshot 2024-05-17 at 11 14 54 AM

With this PR, the pricing table can set `--max-cols: 12` and have those same table-columns set to 33.333% width (4/12)
Screenshot 2024-05-17 at 11 15 39 AM

### Changelog

**Changed**

- Enable custom `--max-col` values in structured-list and pricing-table components

### DIFF
--- a/packages/styles/scss/components/structured-list/_structured-list.scss
+++ b/packages/styles/scss/components/structured-list/_structured-list.scss
@@ -120,20 +120,10 @@
   :host(#{$dds-prefix}-structured-list-header-row),
   :host(#{$dds-prefix}-structured-list-row),
   :host(#{$dds-prefix}-structured-list-group) tr {
-    --max-cols: 4;
-
     @include carbon--breakpoint('sm') {
       @include carbon--make-row();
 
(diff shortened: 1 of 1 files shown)


jkaeser (approved):
LGTM!

emyarod (approved):
looks good to me!

m4olivei (approved):
Looks good to me as well. Thanks for a great PR description. Logic is solid.

PR #11806, opened 2024-05-21
feat(all-components): add shadowparts

### LINKED ISSUE #11673: Add `part` attributes to elements in the shadow DOM
## The problem

From time to time there is a need to navigate around styles encapsulated within components. For example:

### Components that attempt to enforce spacing between their adjacent elements

Example: [dds-cta-block](https://www.ibm.com/standards/carbon/v1/web-components/?path=/story/components-cta-block--with-content-items)

Screenshot 2024-03-27 at 12 02 28 PM

Screenshot 2024-03-27 at 12 02 38 PM

The dds-cta-block component adds padding to `:host` as well as to `.bx--content-layout`. This ensures a certain amount of space between this component and another component added later. The problem, however, is that this spacing is prescriptive, and cannot possibly take into account every possible implementation of this component with every other component or with custom markup. While the padding on `:host` can be overridden, the padding of `.bx--content-layout` cannot.

### Components that have an ever growing list of variations

Example: cards, links/buttons, tiles

Some of our components are highly atomic. That is, they are rarely used in isolation and frequently present differently in different contexts. An example of this is the `caem-tile` component that we have in Carbon For AEM.

`caem-tile-group` on [ibm.com homepage](https://www.ibm.com)

Screenshot 2024-03-27 at 12 26 28 PM

`50-50-container` on [ibm.com/consulting](https://www.ibm.com/consulting)

Screenshot 2024-03-27 at 12 26 54 PM

`THINK Spotlight` (proposed)

Screenshot 2024-03-27 at 12 30 03 PM

The `caem-tile` component was originally built as a deliberate deviation from `dds-card`, and was designed for the use on the homepage exhibited above. New designs came forward that were _very_ close to the original specs, but expanded the components internal spacing for the use in the consulting leadspace. At this time, we created a "double-tile" variant as a quick solution. Now, there is another proposed deviation from the orginal design for a more compact spacing in the proposed THINK spotlight designs. 

## The solution

To avoid continually adding new variations to account for different sizing and spacing needs, we've opened up the shadowroots of CAEM components to styling from the AEM CSS Authors. This allows _adopters_ to assume the responsibility of using components in a way that complies with the IDL and carbon specifications while using them in ways that original designs may not have accounted for. To see this in action, see this [codepen demo](https://codepen.io/andy-blum/pen/JjVJQKo/acb9c7dfb4e8866649991abf69e38969)

Instead of prescribing a limited set of variations and needing to handle new requests for new contexts, now AEM can maintain its own CSS that modifies the presentation of the components we've built. Instead of maintaining a "double-tile", authors can target `caem-tile` elements with a class or a parent selector and increase the spacing to satisfy their needs.

```css
caem-tile.big {
  &::part(wrapper) {
    padding: 2rem;
    gap: 2rem;
  }

  &::part(label) {
    font-size: 1rem;
  }

  &::part(text) {
    font-size: 1.33rem;
    margin-block-end: 0.5rem;
  }
}
```

Screenshot 2024-03-27 at 12 40 53 PM

<hr/>

Likewise, instead of building a new "compact" variant for THINK, we can simply maintain the base version of a tile and the adopters can modifiy the spacing to satisfy _their_ needs.

```css
caem-tile.small {
  &::part(wrapper) {
    padding: 0.5rem;
    gap: 0.5rem;
  }

  &::part(outer):before {
    display: none;
  }

  &::part(text) {
    margin-block-end: 0.5rem;
  }
}
```

Screenshot 2024-03-27 at 12 41 05 PM

<hr />

One downside of this approach is that the carefully-crafted carbon-compliant styles we've started with can be completely overridden to create new non-compliant interfaces

```css
caem-tile.cyan {
  &::part(wrapper) {
    background: #e5f6ff;
  }

  &:hover::part(wrapper) {
    background: #bae6ff;
  }
}
```

Screenshot 2024-03-27 at 12 46 48 PM

andy-blum:
### Related Ticket(s)

Closes #11673 

### Description

This PR updates all component-internal markup to include `part` attributes in the shadow roots allowing global stylesheets to override internal styles.

### Changelog

**Changed**

- Add CSS Shadowparts


jkaeser:
Closing in preference of https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/11812

PR #11818, opened 2024-05-23
feat(cta-block): add shadow parts

Valentin-Sorin-Nicolae:
[ADCMS-5066](https://jsw.ibm.com/browse/ADCMS-5066)

Description

All non-slot elements in the shadow DOM should be given a unique "part" name allowing CSS to target and override component default styles. This is for the "cta-block" component.

Changelog

New

Adding the shadow parts for the "cta-block" component

### DIFF
--- a/packages/web-components/src/components/cta-block/cta-block.ts
+++ b/packages/web-components/src/components/cta-block/cta-block.ts
@@ -98,7 +98,10 @@ class C4DCTABlock extends StableSelectorMixin(C4DContentBlock) {
   protected _renderActions(): TemplateResult | string | void {
     const { _hasAction: hasAction, _handleSlotChange: handleSlotChange } = this;
     return html`
-      <div ?hidden="${!hasAction}" class="${prefix}--content-layout__cta">
+      <div
+        ?hidden="${!hasAction}"
+        class="${prefix}--content-layout__cta"
+        part="action">
         <slot name="action" @slotchange="${handleSlotChange}"></slot>
       </div>
     `;
@@ -122,7 +125,7 @@ class C4DCTABlock extends StableSelectorMixin(C4DContentBlock) {
     });
     return html`
       <div ?hidden="${!_hasContent}" class="${classes}">
-        <div class="${prefix}--content-item-wrapper">
+        <div class="${prefix}--content-item-wrapper" part="wrapper">
           <slot @slotchange="${_handleSlotChange}"></slot>
         </div>
       </div>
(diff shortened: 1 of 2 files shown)


packages/web-components/src/components/cta-block/cta-block.ts
```diff
@@ -141,7 +144,10 @@ class C4DCTABlock extends StableSelectorMixin(C4DContentBlock) {
     });
 
     return html`
-      <div ?hidden="${!this._hasBodyContent()}" class="${classes}">
+      <div
+        ?hidden="${!this._hasBodyContent()}"
+        class="${classes}"
+        part="content">
```
m4olivei, line 150: Lets go with the suggestion from the class, which would be `body`.

```suggestion
        part="body">
```

packages/web-components/src/components/cta-block/cta-block.ts
```diff
@@ -98,7 +98,10 @@ class C4DCTABlock extends StableSelectorMixin(C4DContentBlock) {
   protected _renderActions(): TemplateResult | string | void {
     const { _hasAction: hasAction, _handleSlotChange: handleSlotChange } = this;
     return html`
-      <div ?hidden="${!hasAction}" class="${prefix}--content-layout__cta">
+      <div
+        ?hidden="${!hasAction}"
+        class="${prefix}--content-layout__cta"
+        part="action">
```
m4olivei, line 104: Suggesting that we stick with the element name from the class eg. `cta`.
```suggestion
        part="cta">
```

packages/web-components/src/components/cta-block/cta-block.ts
```diff
@@ -122,7 +125,7 @@ class C4DCTABlock extends StableSelectorMixin(C4DContentBlock) {
     });
     return html`
       <div ?hidden="${!_hasContent}" class="${classes}">
```
m4olivei, line 127: Should put a `part` attribute here for completeness
```suggestion
      <div ?hidden="${!_hasContent}" class="${classes}" part="helper-wrapper">
```

packages/web-components/src/components/cta-block/cta-block.ts
```diff
@@ -122,7 +125,7 @@ class C4DCTABlock extends StableSelectorMixin(C4DContentBlock) {
...
-        <div class="${prefix}--content-item-wrapper">
+        <div class="${prefix}--content-item-wrapper" part="wrapper">
```
m4olivei, line 128: Since there are two `wrapper` things here, lets be more specific with this one.
```suggestion
        <div class="${prefix}--content-item-wrapper" part="content-item-wrapper">
```

m4olivei (changes requested):
In addition to inline comments, lets merge the latest `feat/v2-shadow-parts` here. `cta-block` inherits from `content-block`, which was updated here: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/11809

m4olivei (approved):

PR #11838, opened 2024-05-29
chore(storybook): update shadow parts documentation

marcelojcs:
### Related Ticket(s)
[Jira](https://jsw.ibm.com/browse/ADCMS-5105)

### Description

Implementing documentation to the shadow parts we are updating

### Changelog

**New**

Added warning to the Getting started document shadow parts section.

**Changed**

Added more detail to the shadow parts documentation on individual elements

### DIFF
--- a/packages/web-components/README.md
+++ b/packages/web-components/README.md
@@ -482,6 +482,9 @@ cds-locale-modal::part(back-button) {
 }
 ```
 
+> **⚠️ Warning**  
+>While shadow parts selectors are available as an option, we strongly advise against using it. As this may cause components to not behave as expected.
+
 #### Advanced usage for IBM.com site owners
 
 There are some other key advanced usage patterns that are suitable for IBM.com
--- a/packages/web-components/README.md
+++ b/packages/web-components/README.md
@@ -482,8 +482,8 @@ cds-locale-modal::part(back-button) {
 }
 ```
 
-> **⚠️ Warning**  
->While shadow parts selectors are available as an option, we strongly advise against using it. As this may cause components to not behave as expected.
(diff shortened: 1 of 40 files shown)


packages/web-components/README.md
```diff
@@ -482,6 +482,9 @@ cds-locale-modal::part(back-button) {
 }
 ```
 
+> **⚠️ Warning**  
+>While shadow parts selectors are available as an option, we strongly advise against using it. As this may cause components to not behave as expected.
```
jkaeser, line 486: This is a great start! I think we should add some wording that indicates adopters are responsible for making sure their styles don't break anything, and that we can't guarantee their styles will always work with future updates to components' code.

```suggestion
> **⚠️ Warning**
> While shadow parts selectors are available as an option, use them at your own risk. Changing component styles may cause components to not behave as expected. You are responsible for ensuring your components remain functional while using shadow parts selectors. We cannot guarantee updates to our library's component styles won't conflict with shadow part modifications.
```

packages/web-components/src/components/button/button.ts
```diff
@@ -23,7 +23,7 @@ const { prefix, stablePrefix: c4dPrefix } = settings;
  * Button.
  *
  * @element c4d-button
- * @csspart button.
+ * @csspart button - The button. Usage `c4d-button::part(button)`.
```
jkaeser, line 26: This seems like a good pattern to go forward with! One minor nit:
```suggestion
 * @csspart button - The button. Usage: `c4d-button::part(button)`.
```

andy-blum (approved):

PR #11849, opened 2024-06-06
feat(quote): add shadow parts

Valentin-Sorin-Nicolae:
[ADCMS-5174](https://jsw.ibm.com/browse/ADCMS-5174)

Description

All non-slot elements in the shadow DOM should be given a unique "part" name allowing CSS to target and override component default styles. This is for the "quote" component.

Changelog

New

Adding the shadow parts for the "quote" component and documentation

### DIFF
--- a/packages/web-components/src/components/quote/quote.ts
+++ b/packages/web-components/src/components/quote/quote.ts
@@ -96,10 +96,14 @@ class C4DQuote extends StableSelectorMixin(LitElement) {
     switch (this.markType) {
       case QUOTE_TYPES.SINGLE_CURVED:
         return html`
-          <span class="${prefix}--quote__mark" part="mark mark--opening">‘</span>
+          <span class="${prefix}--quote__mark" part="mark mark--opening"
+            >‘</span
+          >
           <blockquote class="${prefix}--quote__copy" part="copy">
             <slot></slot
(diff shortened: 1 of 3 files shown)


packages/web-components/src/components/quote/quote.ts
```diff
@@ -89,47 +96,66 @@ class C4DQuote extends StableSelectorMixin(LitElement) {
...
         return html`
-          <span class="${prefix}--quote__mark">‘</span>
-          <blockquote class="${prefix}--quote__copy">
-            <slot></slot><span class="${prefix}--quote__mark-closing">’</span>
+          <span class="${prefix}--quote__mark" part="mark">‘</span>
+          <blockquote class="${prefix}--quote__copy" part="copy">
+            <slot></slot
+            ><span class="${prefix}--quote__mark-closing" part="mark-closing"
+              >’</span
+            >
```
jkaeser, line 104: I wonder if we should make the naming consistent by doing something like this in all the places we render the quotes:
```suggestion
          <span class="${prefix}--quote__mark" part="mark mark--opening">‘</span>
          <blockquote class="${prefix}--quote__copy" part="copy">
            <slot></slot
            ><span class="${prefix}--quote__mark-closing" part="mark mark--closing"
              >’</span
            >
```
What do you think?

m4olivei (approved):

PR #11859, opened 2024-06-11
feat(table-of-content): add shadow parts

Valentin-Sorin-Nicolae:
[ADCMS-5179](https://jsw.ibm.com/browse/ADCMS-5179)

Description

All non-slot elements in the shadow DOM should be given a unique "part" name allowing CSS to target and override component default styles. This is for the "table-of-content" component.

Changelog

New

Adding the shadow parts for the "table-of-content" component and documentation


packages/web-components/src/components/table-of-contents/table-of-contents.ts
```diff
@@ -734,14 +745,16 @@ class C4DTableOfContents extends MediaQueryMixin(
                 return html`
                   <li
                     class="${itemClasses}"
+                    part="list-item list-item--active"
```
m4olivei, line 748: Not all of these list items will be active. You need to condition `list-item--active` on wether or not the item is actually active.

packages/web-components/src/components/table-of-contents/table-of-contents.ts
```diff
@@ -791,20 +805,32 @@ class C4DTableOfContents extends MediaQueryMixin(
             : ``}
           <div
             class="${c4dPrefix}-ce--table-of-contents__items-container"
+            part="container"
```
m4olivei, line 808: We already have a `part="container"` for something else (line 766).
```suggestion
            part="items-container"
```

packages/web-components/src/components/table-of-contents/table-of-contents.ts
```diff
@@ -791,20 +805,32 @@ class C4DTableOfContents extends MediaQueryMixin(
...
             this.layout !== TOC_TYPES.HORIZONTAL
               ? `${stickyOffset}px`
               : 0}">
-            <div class="${prefix}--tableofcontents-container">
+            <div
+              class="${prefix}--tableofcontents-container"
+              part="item-container">
               <div
                 class="${prefix}--tableofcontents"
+                part="content"
```
m4olivei, line 818: `part="content"` is being reused a lot here. It looks to me like each place it's being used, it's a unique element that calls for a unique part name. Let's give a unique name to each.

m4olivei (changes requested):

packages/web-components/src/components/table-of-contents/table-of-contents.ts
Valentin-Sorin-Nicolae: Done, as you suggested.

packages/web-components/src/components/table-of-contents/table-of-contents.ts
Valentin-Sorin-Nicolae: Please check again.

m4olivei (approved):
Looks good! Thanks for all the adjustments here.

PR #11861, opened 2024-06-11
feat(lightbox-media-viewer): add shadow parts

m4olivei:
### Related Ticket(s)

[ADCMS-5163](https://jsw.ibm.com/browse/ADCMS-5163)
#11673

### Description

Add shadow parts to the elements of lightbox-media-viewer and friends.

### Changelog

**New**

- Added shadow parts to the lightbox-media-viewer component and related components.

### DIFF
--- a/packages/web-components/src/components/lightbox-media-viewer/__stories__/README.stories.mdx
+++ b/packages/web-components/src/components/lightbox-media-viewer/__stories__/README.stories.mdx
@@ -110,4 +110,12 @@ to see how Web Components selector and `data-autoid` should be used.
 
 <Props of="c4d-lightbox-media-viewer" />
 
+### `<c4d-lightbox-image-viewer>`
+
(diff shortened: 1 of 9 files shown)


andy-blum (commented):
There are default-slotted element in `lightbox-media-viewer.ts` `lightbox-video-player.ts` in the `_renderTitle()` methods. Do you think we ought to add a part to that element?

m4olivei:
> There are default-slotted element in `lightbox-media-viewer.ts` `lightbox-video-player.ts` in the `_renderTitle()` methods. Do you think we ought to add a part to that element?

I saw those. My reasoning there was there is a path to override already, in that they could slot something in that spot themselves, in which case it's in the light DOM and fully styleable, so I didn't bother. Happy to add it if you like.

andy-blum:
we can do it later if we decide its needed.

PR #11862, opened 2024-06-11
feat(image): add shadow parts

m4olivei:
### Related Ticket(s)

[ADCMS-5150](https://jsw.ibm.com/browse/ADCMS-5150)
#11673

### Description

Add shadow parts to image component.

### Changelog

**New**

- Added shadow parts to the image component.


andy-blum (changes requested):
There's a `renderModal()` method we need to modify here as well, I think.

m4olivei:
> There's a `renderModal()` method we need to modify here as well, I think.

That markup gets rendered in the light DOM of the document, so it should already be styleable. While technically `C4DImage` class _could_ overwrite functions of `ModalRenderMixin` to render it in the shadow DOM, it doesn't, so it doesn't seem necessary to me.

andy-blum (approved):

PR #11866, opened 2024-06-13
feat(video-player): add shadow parts

Valentin-Sorin-Nicolae:
[ADCMS-5184](https://jsw.ibm.com/browse/ADCMS-5184)

Description

All non-slot elements in the shadow DOM should be given a unique "part" name allowing CSS to target and override component default styles. This is for the "video-player" component.

Changelog

New

Adding the shadow parts for the "video-player" component and documentation


packages/web-components/src/components/video-player/video-player.ts
```diff
@@ -226,7 +234,9 @@ class C4DVideoPlayer extends FocusMixin(StableSelectorMixin(LitElement)) {
       ${hideCaption
```
m4olivei, line 234: Just above here, on line 233, we're missing a `part` attribute. Lets add `part="video-container"` there.

m4olivei (changes requested):

PR #11868, opened 2024-06-14
feat(sticky-header): allow mixed prefixes with warning

andy-blum:
### Related Ticket(s)

[ADCMS-5224](https://jsw.ibm.com/browse/ADCMS-5224)

### Description

The StickyHeader utility validates all components it tracks, prior to this PR is expects the provided element's `tagName` to match `${c4dPrefix}-component-name`. This works when all components are using the same prefix, but as we work on upgrading from v1 to v2 we'll need to mix-and-match prefixes. This PR will allow the StickyHeader to continue working despite the prefix differences

### Changelog

**Changed**

- allows StickyHeader to validate (with a warning) components that have unexpected prefixes.


jkaeser:
To test support for v1 & v2 components:

1. Visit a page that uses multiple sticky elements, such as https://www.ibm.com/flashsystem. 
2. Either using Requestly or local overrides, replace the masthead JS bundle with the CDN preview's bundle.
   ```html
   <script type="module" src="https://ibmdotcom-cdn.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/11868/masthead.min.js"></script>
   ```
4. Verify there are no console errors generated by the StickyHeader.
5. Verify each of the sticky elements on the page are tracked by the StickHeader by inspecting `window.stickyHeader._elements` in the browser console. Take care to check for derived elements, such as `mastheadL0` and `tableOfContentsInnerBar`.
6. Verify the sticky behavior still works as expected in both desktop and mobile layouts.

packages/utilities/src/utilities/StickyHeader/StickyHeader.js
```diff
@@ -10,6 +10,10 @@ import root from 'window-or-global';
 import settings from '../settings/settings.js';
 
 const { prefix, stablePrefix: c4dPrefix } = settings;
+
+const ddsPrefix = 'dds';
+const prefixV1 = 'bx';
```
m4olivei, line 15: The name `prefixV1` reads as ambiguous when it's all by itself below. Should it just be `bxPrefix` like the other name above?

m4olivei (approved):
Small comment regarding naming. Not a blocker. Feel free to adjust, or leave it as you like. Otherwise code looks good and testing well. I've been testing on various Dotcom shell stories. Not seeing any regressions.

jkaeser (approved):

PR #11874, opened 2024-06-18
fix(toast-notification): remove divs duplicated on lines 36 and 47

cordesmj:
### Description

This PR removes the duplicate `div` with `class="${prefix}--${type}-notification__details"` that wraps the notification content. The duplication results in the class being applied twice, causing the style `margin-inline-end: 1rem;` to be applied twice.

image

### Changelog

**Removed**

- Duplicate div wrapping notification content


cordesmj:

cordesmj:
recheck

PR #11882, opened 2024-06-20
fix(tabs-extended): prevent stealing focus on page load

jkaeser:
### Related Ticket(s)

none

### Description

The current implementation sets focus to a tab any time it is set as "active". This works great for the intended use case, but the problem is that the first tab is made active when the component mounts. This is not so great on page load, forcing focus to the tab and scrolling the page down to it.

The fix is to only set focus on a user interaction.

### Changelog

**Changed**

- Prevents DDSTabsExtended elements from stealing focus on component mount.

### Testing Instructions

1. Visit the Tabs Extended deploy preview and "open the canvas in a new tab" (https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/11882/iframe.html?args=&id=components-tabs-extended--default&viewMode=story)
2. Using your browser console, check which element has focus on load – `document.activeElement`
   - Recommend using Chrome's live expression feature and refreshing the page
3. Verify the focused element is _not_ the Tabs Extended component.

Additionally, you can use the CDN preview on an affected page on IBM.com, such as https://www.ibm.com/products/openpages. Verify the page doesn't scroll down to the Tabs Extended Media component on load.

### DIFF
--- a/packages/web-components/src/components/tabs-extended/tabs-extended.ts
+++ b/packages/web-components/src/components/tabs-extended/tabs-extended.ts
@@ -11,8 +11,6 @@ import settings from 'carbon-components/es/globals/js/settings.js';
 import { html, state, LitElement, TemplateResult, property } from 'lit-element';
 import { classMap } from 'lit-html/directives/class-map.js';
 import ddsSettings from '../../internal/vendor/@carbon/ibmdotcom-utilities/utilities/settings/settings';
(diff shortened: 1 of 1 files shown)


m4olivei (approved):
Looks great. Thanks @jkaeser.

I've compared Storybooks and all is looking good with expected `document.activeElement`. I've also used Requestly to replace all the C4IBM CDN links for the affected pages, and the bug we're seeing there is no longer reproducible.

PR #11895, opened 2024-06-25
fix(cta-block-item-row): set statistics same height

jkaeser:
### Description

`<dds-cta-block-item-row>` elements use the `sameHeight` utility to equalize the heights of slotted headings and text, but not statistics. This PR adds support for that and adds documentation in the form of a Storybook example.

### Changelog

**Changed**

- Enables `<dds-cta-block-item>`s' slotted statistics to have their height equalized.


jkaeser:
The CTA Block story isn't appearing in the React Wrapper Storybook (see https://www.ibm.com/standards/carbon/v1/web-components/react/?path=/story/overview-getting-started--page). Normally I'd try to fix this, but we're sort of rushing this one, so I'm going to let it go for now.

m4olivei (approved):
Looks great! Thanks for the quick turn around @jkaeser !

PR #11904, opened 2024-07-03
feat(content-switcher): add shadow parts

Valentin-Sorin-Nicolae:
[ADCMS-5315](https://jsw.ibm.com/browse/ADCMS-5315)

Description

All non-slot elements in the shadow DOM should be given a unique "part" name allowing CSS to target and override component default styles. This is for the "content-switcher" component.

Changelog

New

Adding the shadow parts for the "content-switcher" component and documentation.

### DIFF
--- a/packages/carbon-web-components/src/components/content-switcher/content-switcher-item.ts
+++ b/packages/carbon-web-components/src/components/content-switcher/content-switcher-item.ts
@@ -141,7 +141,9 @@ export default class CDSContentSwitcherItem extends FocusMixin(LitElement) {
       tabindex="${selected ? '0' : '-1'}"
       aria-controls="${ifDefined(target)}"
       aria-selected="${Boolean(selected)}">
-      <span class="${prefix}--content-switcher__label"><slot></slot></span>
+      <span class="${prefix}--content-switcher__label" part="label"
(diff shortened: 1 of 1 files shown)


packages/carbon-web-components/src/components/content-switcher/content-switcher-item.ts
```diff
@@ -133,6 +136,7 @@ export default class CDSContentSwitcherItem extends FocusMixin(LitElement) {
       type="button"
       role="tab"
       class="${className}"
+      part="button"
```
m4olivei, line 139: This is great. There is also a child `<span>` element that needs a `part` name as well. Probably `label` makes sense as a name for that one.

m4olivei (changes requested):
One small item missed. See comment.

packages/carbon-web-components/src/components/content-switcher/content-switcher-item.ts
Valentin-Sorin-Nicolae: Oops, thanks!

m4olivei (approved):
Thanks! Just pushed a quick commit to cover the docs for the added `label` shadow part. All set.

PR #11911, opened 2024-07-08
fix(text-cta): prevent order of operations issues

marcelojcs:
### Related Ticket(s)
[Jira](https://jsw.ibm.com/browse/ADCMS-5207)
### Description
An issue where the CTA on cards with videos is not being populated on some pages is being caused  by an order of operation issue where the element is making calls tho an event listener that is not yet defined. 
This PR is to ensure the correct order of execution  

### Changelog

**Changed**
Added `customElements.whenDefined()` to the video mixin  to make sure the event only triggers after all componets needed are loaded.

### DIFF
--- a/packages/web-components/src/component-mixins/cta/video.ts
+++ b/packages/web-components/src/component-mixins/cta/video.ts
@@ -132,49 +132,51 @@ const VideoCTAMixin = <T extends Constructor<HTMLElement>>(Base: T) => {
       // Declaring this mixin as it extends `LitElement` seems to cause a TS error
       // @ts-ignore
       super.updated(changedProperties);
-      const { ctaType, videoName, videoDescription, href, videoDuration } =
-        this;
-      const { eventRequestVideoData } = this
-        .constructor as typeof VideoCTAMixinImpl;
-      if (changedProperties.has('ctaType') && ctaType === CTA_TYPE.VIDEO) {
-        if (typeof videoDuration === 'undefined') {
+      customElements.whenDefined(`${ddsPrefix}-text-cta`).then((response) => {
+        const { ctaType, videoName, videoDescription, href, videoDuration } =
+          this;
+        const { eventRequestVideoData } = this
+          .constructor as typeof VideoCTAMixinImpl;
+        if (changedProperties.has('ctaType') && ctaType === CTA_TYPE.VIDEO) {
+          if (typeof videoDuration === 'undefined') {
+            this.dispatchEvent(
+              new CustomEvent(eventRequestVideoData, {
+                bubbles: true,
+                cancelable: true,
+                composed: true,
+                detail: {
+                  href,
+                  videoName,
+                  videoDescription,
+                },
+              })
+            );
+          }
+        }
+
+        if (
+          (changedProperties.has('videoName') &&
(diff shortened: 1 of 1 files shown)


packages/web-components/src/component-mixins/cta/video.ts
```diff
@@ -132,49 +132,51 @@ const VideoCTAMixin = <T extends Constructor<HTMLElement>>(Base: T) => {
       // Declaring this mixin as it extends `LitElement` seems to cause a TS error
       // @ts-ignore
       super.updated(changedProperties);
-      const { ctaType, videoName, videoDescription, href, videoDuration } =
-        this;
-      const { eventRequestVideoData } = this
-        .constructor as typeof VideoCTAMixinImpl;
-      if (changedProperties.has('ctaType') && ctaType === CTA_TYPE.VIDEO) {
-        if (typeof videoDuration === 'undefined') {
+      customElements.whenDefined(`${ddsPrefix}-text-cta`).then((response) => {
```
m4olivei, line 135: I'm pretty sure that this should be targetting the `dds-video-cta-container`, should it not? 

My reasonsing is as follows. This method is dispatching an event named `dds-cta-request-video-data`. That event is [handled by `DDSVideoCTAComposite`](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/v1/packages/web-components/src/components/cta/video-cta-composite.ts#L88-L119). The relevant element in use on the page is `dds-video-cta-container`, (it extends from `DDSVideoCTAComposite`). Therefore, we should be targetting `dds-video-cta-container`.

m4olivei (changes requested):

packages/web-components/src/component-mixins/cta/video.ts
m4olivei: Also note that there is a Typescript error here. In the `.then()`, you’re declaring a `response` parameter that you never actually use. Just need to remove that `response` parameter, and go with an empty parameter list.

marcelojcs: @m4olivei, Thanks for the review!!.
Yes, yes it should. 
Addressed on my latest commit.

packages/web-components/src/component-mixins/cta/video.ts
```diff
@@ -132,49 +132,51 @@ const VideoCTAMixin = <T extends Constructor<HTMLElement>>(Base: T) => {
...
-              href,
-            },
-          })
-        );
-      }
-
-      if (ctaType === CTA_TYPE.VIDEO && this.offsetWidth > 0) {
-        this._updateVideoThumbnailUrl();
-      }
+        if (ctaType === CTA_TYPE.VIDEO && this.offsetWidth > 0) {
```
m4olivei, line 176: I'm also wondering if this conditional should be left outside the `whenDefined` promise? I don't think the code in here depends on the `dds-video-cta-container`.

m4olivei (approved):
This is looking good to me. Logically it makes sense. I've also tested on the reported pages using Requestly and the CDN links here, it works!

andy-blum (approved):
Verified in codepen reproduction. Thanks!

PR #11916, opened 2024-07-09
fix(fonts): set font path for plex fonts cdn

annawen1:
### Related Ticket(s)

### Description

the plex font paths are defaulting to `~ibm/plex` which ends up 404ing when we use the font CDNs. This PR sets `use-akamai-cdn` flag to true in the carbon font configs.

### Changelog

**New**

- set `use-akamai-cd` flag for plex fonts


emyarod (approved):
looks good to me!

PR #11920, opened 2024-07-12
feat(copy): add shadow parts

Valentin-Sorin-Nicolae:
[ADCMS-5316](https://jsw.ibm.com/browse/ADCMS-5316)

Description

All non-slot elements in the shadow DOM should be given a unique "part" name allowing CSS to target and override component default styles. This is for the "copy" component.

Changelog

New

Adding the shadow parts for the "copy" component and documentation.

### DIFF
--- a/packages/carbon-web-components/src/components/copy/copy.ts
+++ b/packages/carbon-web-components/src/components/copy/copy.ts
@@ -1,7 +1,7 @@
 /**
  * @license
  *
- * Copyright IBM Corp. 2019, 2023
+ * Copyright IBM Corp. 2019, 2024
  *
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
--- a/packages/carbon-web-components/src/components/copy/copy.ts
+++ b/packages/carbon-web-components/src/components/copy/copy.ts
@@ -17,6 +17,8 @@ import CDSIconButton from '../icon-button/icon-button';
  * Copy.
  *
  * @element cds-copy
+ *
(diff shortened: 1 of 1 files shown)


packages/carbon-web-components/src/components/copy/copy.ts
```diff
@@ -17,6 +17,8 @@ import CDSIconButton from '../icon-button/icon-button';
  * Copy.
  *
  * @element cds-copy
+ *
+ * @csspart content - The content. Usage `cds-copy::part(content)`
```
m4olivei, line 21: Sorry to nit pick. I think it will be less ambiguous to call it `tooltip-content`. If I think about what the copy button's `content` is, I might not end up thinking that this is actually referring to the tooltip content.
```suggestion
 * @csspart tooltip-content - The content. Usage `cds-copy::part(tooltip-content)`
```

packages/carbon-web-components/src/components/copy/copy.ts
```diff
@@ -81,7 +83,7 @@ class CDSCopy extends CDSIconButton {
   // eslint-disable-next-line class-methods-use-this
   protected _renderTooltipContent() {
     return html`
-      <cds-tooltip-content>
+      <cds-tooltip-content part="content">
```
m4olivei, line 86:
```suggestion
      <cds-tooltip-content part="tooltip-content">
```

m4olivei (changes requested):

m4olivei (approved):
Thanks for the adjustment!

PR #11927, opened 2024-07-16
feat(tabs): add shadow parts

Valentin-Sorin-Nicolae:
[ADCMS-5352](https://jsw.ibm.com/browse/ADCMS-5352)

Description

All non-slot elements in the shadow DOM should be given a unique "part" name allowing CSS to target and override component default styles. This is for the "tabs" component.

Changelog

New

Adding the shadow parts for the "tabs" component and documentation.


m4olivei (approved):
LGTM!

PR #11931, opened 2024-07-17
feat(structured-list): adding shadow parts

bruno-amorim:
### Related Ticket(s)

[ADCMS-5351](https://jsw.ibm.com/browse/ADCMS-5351)

### Description
Adding shadow parts to the Structured List component so that it can be styled outside of the shadow tree.
Snapshots updated


m4olivei (changes requested):
Everything here looks great to me! Just missed one element in the skeleton component:

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/carbon-web-components/src/components/structured-list/structured-list-header-cell-skeleton.ts#L21

m4olivei (approved):

PR #11935, opened 2024-07-22
feat(toggle): add shadow parts

Valentin-Sorin-Nicolae:
[ADCMS-5358](https://jsw.ibm.com/browse/ADCMS-5358)

Description

All non-slot elements in the shadow DOM should be given a unique "part" name allowing CSS to target and override component default styles. This is for the "toggle" component.

Changelog

New

Adding the shadow parts for the "toggle" component and documentation.

### DIFF
--- a/packages/carbon-web-components/src/components/toggle/toggle.ts
+++ b/packages/carbon-web-components/src/components/toggle/toggle.ts
@@ -1,7 +1,7 @@
 /**
  * @license
  *
- * Copyright IBM Corp. 2019, 2023
+ * Copyright IBM Corp. 2019, 2024
  *
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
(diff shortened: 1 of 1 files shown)


packages/carbon-web-components/src/components/toggle/toggle.ts
```diff
@@ -150,15 +159,21 @@ class CDSToggle extends HostListenerMixin(CDSCheckbox) {
...
-        <div class="${inputClasses}">
-          <div class="${toggleClasses}" @click=${handleChange}>
+      <label for="${id}" class="${prefix}--toggle__label" part="label">
+        <span class="${prefix}--toggle__label-text" part="label-text"
+          >${labelText}</span
+        >
+        <div class="${inputClasses}" part="appearance">
+          <div
+            class="${toggleClasses}"
+            part="switch switch--checked"
```
m4olivei, line 169: We should only add `switch--checked` when `checked` is true.
```suggestion
            part="switch${checked ? ' switch--checked' : ''}"
```

m4olivei (changes requested):

m4olivei (approved):

PR #11945, opened 2024-07-24
[web components]: Remove babel/preset-env from build step

### LINKED ISSUE #11944: [web components]: Remove babel/preset-env from build step
### Description

We recently realized that [`@babel/preset-env` slipped into the build](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/gulp-tasks/build/modules/scripts.js#L39) for C4IBM at some point. Furthermore, it uses `@babel/preset-env` without any [targets config](https://babeljs.io/docs/options#targets), which [according to Babel docs](https://babeljs.io/docs/options#no-targets):

> When no targets are specified: Babel will assume you are targeting the oldest browsers possible.

This means the `es/` folder that ships with the C4IBM package compiles down to commonjs and includes a lot of polyfills. This has the effect of bloating the transpiled modules by more than is necessary.

C4IBM is unique in its use of `@babel/preset-env` among CWC, C4IBM and CAEM. The other packages do not use it.

The history here was that we added it for the [masthead/v2 branch here](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/9602). It then got ported over to main when we made the same [changes for the Cloud Masthead here](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10085). 

It looks like it was added to avoid build issues which can otherwise be resolved by upgrading the `strip-comments` package to ^2.0.1.

The task here is to remove [use of `@babel/preset-env` in the `build:modules:scripts` gulp task in C4IBM](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/gulp-tasks/build/modules/scripts.js#L39).

### Component(s) impacted

All

### Browser

Chrome

latest

### Severity

### Application/website

ibm.com

### Package

### CodeSandbox example

https://ibm.com/

_No response_

_No response_

### Code of Conduct

m4olivei:
### Related Ticket(s)

Closes #11944
[ADCMS-5513](https://jsw.ibm.com/browse/ADCMS-5513)

### Description

Removes `@babel/preset-env` from the babel configuration for the `build:modules:scripts` gulp task that is responsible for transpiling modules into the `es/` directory for distribution. This seems to have slipped in by mistake (see [issue description](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/11944) for details) and leads to bloated transpiled modules, as the targets for `@babel/preset-env` are undefined leading to transpiling for the lowest common denominator of browser support.

The fix here puts the `@carbon/ibmdotcom-web-components` build script inline with the CWC build script, removing `@babel/preset-env`.

### Changelog

**Changed**

- Updated `strip-comments` to `^2.0.1`

**Removed**

- Removed `@babel/preset-env` from the `build:modules:scripts` gulp task


andy-blum (commented):
Should we remove `@babel/preset-env` from the dev-dependencies too?

m4olivei:
> Should we remove `@babel/preset-env` from the dev-dependencies too?

Yeah good question. I left it, b/c [Carbon Web Components still has it as a dependency](https://github.com/m4olivei/carbon-for-ibm-dotcom/blob/chore/remove-present-env/packages/carbon-web-components/package.json#L92), which makes me think that it's still required somewhere deep in the belly of the beast. I suspect it's on the Storybook side of things.

kennylam (approved):
@m4olivei Yep, looks like its main use is in Storybook at the moment, which I think we should be ok with since we only target the latest browsers (one of the package SB configs is still targeting `ie11`, but we can remove that later). Thanks for this!

PR #11976, opened 2024-08-13
fix(tabs): avoid setting focus/active to -1

### LINKED ISSUE #11975: [tabs]: Keypresses cause tabpanel to disappear
### Description

https://jsw.ibm.com/browse/ADCMS-5840

Tabbed content has a bug where keypresses cause the content to disappear. Keyboard navigation using tab-stops as well as any keypress that's not an arrow key cause content to be hidden from the user. Bumping to sev 2 as this makes content in tabs completely inaccessible to SR users

Screenshot 2024-08-13 at 1 01 46 PM

Screenshot 2024-08-13 at 1 02 02 PM

### Component(s) impacted

tabs-extended, tabs-extended-media

### Browser

Firefox

v1.47.0

### Severity

### Application/website

ibm.com

### Package

### CodeSandbox example

https://www.ibm.com/standards/carbon/v1/web-components/?path=/story/components-tabs-extended--default

Navigate to tabbed content with keyboard, press any non-arrow key

_No response_

### Code of Conduct

andy-blum:
### Related Ticket(s)

Closes #11975 
https://jsw.ibm.com/browse/ADCMS-5840

### Description

tabs-extended's _handleTabListKeyDown method sets `targetTab` to a default value of -1 and then modifies that value for arrow keys and home/end.

Prior to this PR, the component sets the active item and focus to targetTab, even if it's still -1. This PR changes that to only set active/focus if the value has been modified away from -1

### Changelog

**Changed**

- fixes a bug in which tab panels could disappear on keypress

### DIFF
--- a/packages/web-components/src/components/tabs-extended/tabs-extended.ts
+++ b/packages/web-components/src/components/tabs-extended/tabs-extended.ts
@@ -138,8 +138,11 @@ class DDSTabsExtended extends MediaQueryMixin(StableSelectorMixin(LitElement), {
       default:
         break;
     }
-    this._setActiveItem(targetTab);
-    this._setFocus(targetTab);
+
+    if (targetTab !== -1) {
+      this._setActiveItem(targetTab);
+      this._setFocus(targetTab);
+    }
   }
 
   /**


m4olivei (approved):
LGTM. Thanks for the quick fix. Also checked v2, but it doesn't suffer from the same issue. Instead it inherits from CWC, which handles this scenario well.

andy-blum:
Re-running the ci-check on web-components. It's passing for me locally.

m4olivei:
Just triggered another re-run. New day, fresh start? 🤞

PR #11993, opened 2024-08-28
fix(callout-quote): forcing sans-serif font in non latin fonts

bruno-amorim:
### Related Ticket(s)

[ADCMS-6085](https://jsw.ibm.com/browse/ADCMS-6085)

### Description

When using the callout-quote component, the shadowroot styles render the blockquote element in a serif font. (e.g. [mainframe application modernization](https://www.ibm.com/mainframe-application-modernization))

In non-latin fonts, however, there isn't a serif version of those typefaces, leading to the components rendering non-plex characters. (e.g. [MAM in Japanese](https://www.ibm.com/jp-ja/mainframe-application-modernization))

With this fixed applied:
image

### DIFF
--- a/packages/styles/scss/components/callout-quote/index.scss
+++ b/packages/styles/scss/components/callout-quote/index.scss
@@ -26,7 +26,7 @@
     }
 
     &[lang="ar"],
-    &[lang="cn"],
+    &[lang="zh"],
     &[lang="jp"],
     &[lang="ko"] {
       .#{$prefix}--quote__copy {
--- a/packages/styles/scss/components/callout-quote/index.scss
+++ b/packages/styles/scss/components/callout-quote/index.scss
@@ -25,12 +25,12 @@
       }
     }
 
-    &[lang="ar"],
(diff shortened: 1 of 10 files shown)


packages/styles/scss/components/callout-quote/index.scss
```diff
@@ -24,6 +24,15 @@
         }
       }
     }
+
+    &[lang="ar"],
+    &[lang="cn"],
```
andy-blum, line 29:
```suggestion
    &[lang="zh"],
```

packages/styles/scss/components/quote/_quote.scss
```diff
@@ -23,6 +23,15 @@
       padding-left: $spacing-05;
       max-width: 80%;
     }
+
+    &[lang="ar"],
+    &[lang="cn"],
```
andy-blum, line 28:
```suggestion
    &[lang="zh"],
```

andy-blum (changes requested):

PR #12001, opened 2024-08-30
feat(multiple-components): adding last round of missing shadow parts

bruno-amorim:
### Related Ticket(s)

[ADCMS-6148](https://jsw.ibm.com/browse/ADCMS-6148)

### Description

I have added the last round of shadow parts that we've missed through our shadow part tickets.

### DIFF
--- a/packages/web-components/src/component-mixins/cta/cta.ts
+++ b/packages/web-components/src/component-mixins/cta/cta.ts
@@ -72,6 +72,7 @@ export const types = CTA_TYPE;
 /**
  * @param Base The base class.
  * @returns A mix-in implementing the logic of handling link for CTA.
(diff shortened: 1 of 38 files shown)


andy-blum (changes requested):
Looks like we're still missing a couple:

left-nav-menu.ts
```ts
render() {

  // ...

    return html`
      <div class="${prefix}--side-nav__item">
```

left-nav-name.ts
```ts
render() {
    const { href, prefix: namePrefix } = this;
    const namePrefixPart = !namePrefix
      ? undefined
      : html`
          <span class="${prefix}--header__name--prefix">${namePrefix}</span
          >&nbsp;
        `;
```

andy-blum:
This one is a tricky one as well - the CTA mixin renders a default SVG element in the shadowroot that we should put a part attribute on.

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/component-mixins/cta/cta.ts#L195-L205

PR #11986, opened 2024-08-21
chore(package): remove `@carbon/web-components` from monorepo

kennylam:
### Related Ticket(s)

Closes https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/12020

This is ready for review, but should not be merged until the package has been fully migrated to the Carbon monorepo (https://github.com/carbon-design-system/carbon/issues/17167).

### Description

Remove `carbon-web-components` package and all related files, build processes, etc from monorepo.

### Changelog

**Changed**

- update `@carbon/ibmdotcom-web-components` components to import directly from `@carbon/web-components` instead of `internal/vendor` (these build artifacts have been removed along with the package)

**Removed**

- `@carbon/web-components`:
  - entire package
  - build workflows, i.e. gulp, rollup, GH actions
  - documentation

PR #12010, opened 2024-09-04
[devops]: Remove references to internal/vendor packages

### LINKED ISSUE #12008: [devops]: Remove references to internal/vendor packages
### The problem

Within C4IBM, there is potentially unneccessary variety in import statements such that the same code is imported via node resolution algorithm, but also by absolute path from the `src/internal/vendor/@carbon` directories. The variety causes build tools to see effectively the same source files as being different, which leads to duplication in the bundled artifacts. It's also subjectively difficult to follow and overly complicated. Yarn workspaces symlinks internal packages together such that we should not have to use this src/internal directory. Let’s attempt to simplify and potentially gain some efficiency in the builds by being able to hoist and share more common dependencies in the builds.

See also some background on the history of the src/internal/vendor directory: https://ibm-marketing-comms.slack.com/archives/G01LKRP5RB6/p1717702105957869

### The solution

This task is to, in C4IBM, refactor import statements of the following form, and remove the build:vendor jobs from packages:

FROM: `import [thing] from '../../internal/vendor/@carbon/ibmdotcom-utilities/[…]';`
TO: `import [thing] from ‘@carbon/ibmdotcom-utilities/[…]';`

**Acceptance Criteria**

* Relevant CI tasks complete successfully
* Check codesandbox examples that bundle from the es/ folder and ensure that they continue to work as expected.

### Application/website

ibm.com

### Business priority

Low Priority = release date is not dependent on fix or not upcoming

N/A

### Examples

_No response_

### Code of Conduct

m4olivei:
### Related Ticket(s)

Closes #12008 

### Description

Adjusts all `internal/vendor` imports to use the package imports instead.

### Changelog

**Changed**

- All imports of the form `[...]/internal/vendor/@carbon/[package][rest]` changed to `@carbon/[package]/es[rest]`

**Removed**

- vendor build steps


packages/web-components/gulp-tasks/vendor.js
```diff
@@ -1,165 +0,0 @@
...
-    .src([`${servicesStoreESSrcDir}/**/*`, '!**/*-{test,story}.js'])
-    .pipe(gulp.dest(servicesStoreVendorESDstDir));
-
-/**
- * Generate `lib/internal/vendor` contents.
- */
-const servicesStoreVendorCJSDst = () =>
-  gulp
-    .src([`${servicesStoreCJSSrcDir}/**/*`, '!**/*-{test,story}.js'])
-    .pipe(gulp.dest(servicesStoreVendorCJSDstDir));
```
m4olivei, line 106: May have gone too far with removing this one? This copies the files from `packages/services-store/lib` into `packages/web-components/lib/internal/vendor/@carbon/ibmdotcom-services-store`. Similar with the one above, but `s/lib/es`. This appears to be a way of exposing the modules from `@carbon/ibmdotcom-services-store` to downstream users of `@carbon/ibmdotcom-web-components`, since the former is a private package that isn't published. 

This change would be a breaking change for any downstream users who might be using those imports.

m4olivei: Yep, will probably want this back. CAEM is case and point on this. Although we only use it once, and only for type info, it does get used: https://github.ibm.com/webstandards/carbon-for-aem/blob/main/src/components/video-player/video-player-container.ts#L22

m4olivei: I've added back the `vendor` build step, and left anything that copies source files into the `es/` or `lib/` directories, since that forms part of the public API of this package.

andy-blum (approved):
Passes all the tests, and I don't see any stragglers still importing from internal/vendor.

Even after `yarn clean` `yarn install` and `yarn build`, I am still seeing built files importing from those paths, but maybe that can be a followup?

m4olivei:
> Passes all the tests, and I don't see any stragglers still importing from internal/vendor.
> 
> Even after `yarn clean` `yarn install` and `yarn build`, I am still seeing built files importing from those paths, but maybe that can be a followup?

I see one place where those are coming from. Fix incoming.

m4olivei:
> > Passes all the tests, and I don't see any stragglers still importing from internal/vendor.
> > Even after `yarn clean` `yarn install` and `yarn build`, I am still seeing built files importing from those paths, but maybe that can be a followup?
> 
> I see one place where those are coming from. Fix incoming.

Fixed in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/12010/commits/86ef4f5dbbdc49119b28beb53eabdb7ba9c90484. There was a build script for icons that was automating creation of modules that used an `internal/vendor` import path.

andy-blum:
Looks good - should we remove the documentation around the vendor dirs?

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/IMPLEMENTATION_NOTES.md#vendor-directory

m4olivei:
> Looks good - should we remove the documentation around the vendor dirs?
> 
> https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/IMPLEMENTATION_NOTES.md#vendor-directory

Addressed in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/12010/commits/e47e8c17a1faef713e3a51384ded928293744149

m4olivei:
@kennylam There was a big merge conflict here after the [CWC removal PR](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/11986) landed. It was caused by a lot of overlap in adjusting import statements. I've addressed the conflicts and this should now be ready to go.

kennylam (approved):
LGTM!

PR #12030, opened 2024-09-16
fix(card): component renders multiple headings

Valentin-Sorin-Nicolae:
### Related Ticket(s)

[ADCMS-6175](https://jsw.ibm.com/browse/ADCMS-6175)

### Description

C4DCard and C4DCardCTA both have a `_renderHeading()` method that returns the following:

```
<slot name="heading"></slot>
<c4d-card-heading>${caption}</c4d-card-heading>
```

If these components are authored with a <c4d-card-heading> element within them, it's automatically assigned to slot="heading". This leads to cards with multiple headings, each with their own minimum of 64px of margin-bottom.

image

**Changed**

image

### DIFF
--- a/packages/web-components/src/components/card/card.ts
+++ b/packages/web-components/src/components/card/card.ts
@@ -108,8 +108,9 @@ class C4DCard extends CTAMixin(StableSelectorMixin(CDSLink)) {
       )
     );
     return html`
-      <slot name="heading"></slot
-      ><c4d-card-heading>${caption}</c4d-card-heading>
+      <slot name="heading"
+        ><c4d-card-heading>${caption}</c4d-card-heading></slot
+      >
     `;
   }
 


andy-blum (approved):

m4olivei:
Hey @RichKummer thanks for the review.

> The original Storybook styles were off on this one – the video caption (if there is one) should really be coming in as body copy in the body copy section of the card, instead of a second header. The intent is that the caption would render as the body copy.

I'm not sure this is the case. What we're calling "caption" here, is really the `videoName` from the Kaltura API. It seems like the intent is for the `videoName` to be used as a fallback when no Custom title is provided. In which case, what we have here should suffice to fix the issue of a double title when a `videoName` is available.

Does that make sense? As I read over the issue and code, it seems to me that the naming of `caption` is causing some confusion. Maybe we should make an adjustment there as well.

RichKummer (approved):
LGTM!

PR #12059, opened 2024-09-27
fix(masthead): hide scroller buttons when there are no menuItems

m4olivei:
### Related Ticket(s)

[ADMCS-6215](https://jsw.ibm.com/browse/ADCMS-6215)

### Description

This bug fix covers the edge case of the `l0Data` on the `<c4d-masthead-l1>` component being undefined or empty.

### Testing instructions

* Open up the Masthead > With L1 story
* Check off the "Use empty data for L1 menu items". This will employ an L1 dataset that does not have any `menuItems` defined.
* The scroller buttons should be hidden

### Changelog

**Changed**

- Handle the edge case of the `l0Data` being `undefined`.


packages/web-components/src/components/masthead/__stories__/masthead.stories.ts
```diff
@@ -17,7 +17,12 @@ import '../masthead-container';
 import { L1_CTA_TYPES } from '../defs';
 import styles from './masthead.stories.scss';
 import { ifDefined } from 'lit/directives/if-defined.js';
-import { mastheadL0Data, mastheadL1Data, mastheadLogoData } from './links';
+import {
+  mastheadL0Data,
+  mastheadL1Data,
+  mastheadL1EmptyMenuItemsData,
```
andy-blum, line 23: This doesn't appear to be an export from `./links.ts`

packages/web-components/src/components/masthead/masthead-l1.ts
```diff
@@ -231,7 +231,12 @@ class C4DMastheadL1 extends StableSelectorMixin(LitElement) {
       } = this;
 
       // Only act if all the needed elements are present.
-      if (
+      const menuItems = this.l1Data?.menuItems ?? [];
+      if (menuItems.length === 0) {
```
andy-blum, line 235: Is `menuItems` a mandatory field? Maybe we should put some optional chaining here just in case?

andy-blum (changes requested):

packages/web-components/src/components/masthead/masthead-l1.ts
m4olivei: It's not mandatory. That's where in the line above, I have the optional chaining for accessing `menuItems`, and use the null coalescing operator to default to an empty array, so that the local variable `menuItems` is always an array.

packages/web-components/src/components/masthead/__stories__/masthead.stories.ts
m4olivei: Ohh boo, I forgot to commit the changes to that. Coming right up.

andy-blum (approved):

emyarod (approved):
looks good to me!

PR #12063, opened 2024-09-30
[card-group]: Refactor to use subgrid in favor of sameHeight

### LINKED ISSUE #11998: [card-group]: Replace JS-based sameHeight functionality with CSS subgrid
### The problem

The c4d-card-group component utilizes sameHeight calculations to set `height` styles on cards' sub-components. This is a lot of JS for basic layout that can be accomplished with CSS, and it fails to account for cards wrapping into new rows.

Additionally, the existing sameheight logic is buggy, adding extra padding/margin in places its not needed.

### The solution

[CSS Subgrid](https://caniuse.com/css-subgrid) has good support and is labeled as baseline 2023. Using a well crafted grid/subgrid pairing we can ensure cards retain alignment between their headings/bodies/tag groups/footers more performantly and only relative to items in the same row.

### Application/website

ibm.com

### Business priority

_No response_

### Examples

An example of the [card group's buggy alignment](https://iciuc--carbon-demos.netlify.app/card-group). Per the component specs, headings should have 64px of space below them, but we have well more than that in this demo

An example of [cards being kept in alignment with subgrid](https://codepen.io/andy-blum/pen/XWGxrbJ)

### Code of Conduct

m4olivei:
### Related Ticket(s)

Closes #11998

### Description

This is a refactoring of card group to make use of CSS subgrid instead of the Javascript `sameHeight` logic on all the pieces of the card group.

### Changelog

**Changed**

- Use CSS subgrid for layout out the card elements within a card group


andy-blum:
I've put together a couple demo pages showing the implications of this change. 
* [using sameHeight](https://card-group-sameheight--carbon-demos.netlify.app/)
* [using subgrid](https://card-group-subgrid--carbon-demos.netlify.app/)

## Card contents width dependent on body text

If we remove the card's body text something is getting squished. Seems like a rule inherited from `card` about `justify-content` causing this.

Screenshot 2024-10-04 at 10 23 30 AM

## I don't know that we need to subgrid all the way down into the main slot

After fixing the card width issue above, it's easier to see that the tags will never be able to line up in this scenario due to the grid lines. This would also likely end up happening if we had multiple `<p>` tags in a card as well.

Screenshot 2024-10-04 at 10 25 30 AM

## We should make sure grid placement is more explicit

Right now all items just get placed into whatever grid cell is open next. If we are more explicit with our placements we'll get a better result

Before:
Screenshot 2024-10-04 at 10 26 29 AM

After:
Screenshot 2024-10-04 at 10 26 35 AM

The easiest way to do this might be to wrap all our slots in additional divs:

Current:
```html
<div part="container" class="cds--tile cds--card cds--tile--clickable cds--card--link">
  <slot name="image"></slot>
  <div part="wrapper" class="cds--card__wrapper ">
    <div part="content" class="cds--card__content">
      <slot name="eyebrow"></slot> 
      <slot name="heading"></slot>    
      <div part="copy" class="cds--card__copy">
        <slot></slot>
      </div>
      <slot name="pictogram"></slot>
      <slot name="footer"></slot>
    </div>
  </div>
</div>
```

Suggested:
```html
<div part="container" class="cds--tile cds--card cds--tile--clickable cds--card--link">
  <div part="image-wrapper"><slot name="image"></slot></div>  
  <div part="wrapper" class="cds--card__wrapper ">
    <div part="content" class="cds--card__content">
      <div part="eyebrow-wrapper"><slot name="eyebrow"></slot></div> 
      <div part="heading-wrapper"><slot name="heading"></slot></div>    
      <div part="copy" class="cds--card__copy">
        <slot></slot>
      </div>
      <div part="pictogram-wrapper"><slot name="pictogram"></slot></div>
      <div part="footer-wrapper"><slot name="footer"></slot></div>
    </div>
  </div>
</div>
```

In this way, the interaction with the grid is limited to the card-group and the card-group-item. Styles on slotted content shouldn't break our grid's alignment. Additionally, we can apply the necessary spacing to the wrapper divs. With spacing applied to those internal elements we can also hide them with use of `:has()`

```scss
:host(c4d-card-group-item) {

  [part="heading-wrapper"] {
    margin-bottom: 64px // Required spacing below heading
  }
  
  :not(:has([slot="heading"])) [part="heading-wrapper"] {
    display: none;
  }

}
```

packages/styles/scss/components/card-group/_card-group.scss
```diff
@@ -56,10 +56,40 @@
...
 
   :host(#{$c4d-prefix}-card-group-item) {
+    display: contents;
+
     .#{$prefix}--card {
+      display: grid;
       border: 0;
+      grid-row: span 10;
+      grid-template-rows: subgrid;
+      margin-block-end: 2rem;
```
kennylam, line 66:
```suggestion
      margin-block-end: $spacing-07;
```

packages/styles/scss/components/card-group/_card-group.scss
```diff
@@ -84,6 +114,23 @@
         outline-offset: -1px;
       }
     }
+
+    &[grid-mode='narrow'] {
+      .#{$prefix}--card {
+        margin-block-end: 1rem;
```
kennylam, line 120:
```suggestion
        margin-block-end: $spacing-05;
```

kennylam (approved):
This is great, thanks @m4olivei! Just some spots to use tokens instead of hard-coded units.

andy-blum:
Thanks for the updates, alignments are looking much better now.

The pictogram one still has some weirdness, though:
* Pictograms seem to disable eyebrows
* Pictograms seem to create their own footer cta icons
* The body text is double-wrapped in p tags

Screenshot 2024-10-08 at 8 54 32 AM

Screenshot 2024-10-08 at 8 55 59 AM

m4olivei:
We should be looking _pretty_ good now. Waiting on the Storybook, but here is a rundown of the feedback so far.

> This is great, thanks @m4olivei! Just some spots to use tokens instead of hard-coded units.

All set there, thanks @kennylam. Fixed in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/12063/commits/2e3e15b9bf03aa45fe62b482fba442a54724045f

> I've put together a couple demo pages showing the implications of this change.

These are great, thanks @andy-blum .

> If we remove the card's body text something is getting squished. Seems like a rule inherited from card about justify-content causing this.

Yep, fixed in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/12063/commits/3ab3d81e1419f0501648c4d4b608171c78ec9a0d

> I don't know that we need to subgrid all the way down into the main slot

This whole thing threw me off at first too. Originally I was trying to declare a subgrid on a grandchild element to the grid container and expecting it to like get placed on the nearest grid ancestor. I realized pretty quickly that it doesn't work that way. The child, which was just a regular old grid item gets placed in the parent grid first, in the first row and column. Then the grandchild getting `grid-template-rows: subgrid` makes no sense, b/c its immediate parent is just a grid item of the grandparent ... not a grid container. As it turns out, you have to have each element in the tree except for the leaves be `display: grid`. And we have to start up at the element that wraps all of our items, then go all the way down. You can effectively take out some child elements using `display: contents`, but that's not always appropriate.

> We should make sure grid placement is more explicit

Resolved in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/12063/commits/29f5b4f34421320fa02553b02e2dd30cfab365ef also in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/12063/commits/b72b45c7d28cb658d67f1eb20405267fe27371b9 . The wrappers work well.

The only thing is, I made changes to the card markup, which is widely used, so we'll need to be careful to regression test.

> The pictogram one still has some weirdness, though:
> Pictograms seem to disable eyebrows

Fixed in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/12063/commits/b72b45c7d28cb658d67f1eb20405267fe27371b9

> The pictogram one still has some weirdness, though:
> Pictograms seem to create their own footer cta icons

I haven't been able to reproduce this one.

> The pictogram one still has some weirdness, though:
> The body text is double-wrapped in p tags

This is also the case on `main`. There is [some code](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/card/card.ts#L347-L353) that is responsible in `updated`. Haven't touched that here, so I want to say it's out of scope for this.

andy-blum:
Looks like the only remaining alignment issue is some extra margin thrown on `.cds--card__copy` when pictograms are involved.

Screenshot 2024-10-08 at 2 49 54 PM

m4olivei:
> Looks like the only remaining alignment issue is some extra margin thrown on `.cds--card__copy` when pictograms are involved.

I'm not sure that's worth fixing TBH. Pictogram margins appear to be intentionally unique. I don't think it's worth trying to make them conform in the mixed card type case, when my read is that Pictograms are meant to appear in the card-group exclusively with other pictograms.

andy-blum:
We do need to make sure the spacing is correct when pictograms are used, however.

Screenshot 2024-10-08 at 5 01 26 PM

Screenshot 2024-10-08 at 5 01 41 PM

andy-blum (approved):
I think we're in a place where we can declare this good enough and an improvement over the current card-group setup that we should merge it in.

PR #12064, opened 2024-10-01
fix(legal-nav): fix adjunct links not showing

### LINKED ISSUE #12037: [footer]: Adjunct links do not appear automatically when set in Javascript via property setter after initial update
https://jsw.ibm.com/browse/ADCMS-6324

### Description

The `<c4d-footer-container>` component defines an `adjunctLinks` property which allows users the ability to add `adjunctLinks` to the footer in Javascript, eg.

```
const adjunctLinks = [
    { "title": "link 1", "url": "abc" }, 
    { "title": "link 2", "url": "abc" }, 
    { "title": "link 3", "url": "abc" },
];
document.querySelector('c4d-footer-container').adjunctLinks = adjunctLinks;
```
However, when done in this way, the adjunct links, although added to the DOM and slotted into the correct place, remain hidden.

The only workaround is to write custom Javascript to reach into the shadow DOM of the `<c4d-legal-nav>` component and remove the `cds--adjunct-links__container--hidden` class.

### Component(s) impacted

`<c4d-footer-container>`
`<c4d-legal-nav>`

### Browser

Chrome

v2.13.0

### Severity

### Application/website

ibm.com

### Package

@carbon/web-components

### CodeSandbox example

https://stackblitz.com/github/m4olivei/carbon-demo/tree/test/footer-adjunct-links?file=main.js

_No response_

_No response_

### Code of Conduct

marcelojcs:
### Related Ticket(s)
Closes #12037
[Jira](https://jsw.ibm.com/browse/ADCMS-6324)

### Description

PR to solve the following: 

The <c4d-footer-container> component defines an adjunctLinks property which allows users the ability to add adjunctLinks to the footer in Javascript, eg.

```javascript
const adjunctLinks = [
  { "title": "link 1", "url": "abc" },
  { "title": "link 2", "url": "abc" },
  { "title": "link 3", "url": "abc" },
];

document.querySelector('c4d-footer-container').adjunctLinks = adjunctLinks;
```
 
However, when done in this way, the adjunct links, although added to the DOM and slotted into the correct place, remain hidden.
 
### Changelog

**Changed**

- Foced the method that determines if the links should be hidden or not to run when the adjunct links are updated.

### Testing instructions 

To test the changes open the default footer `<storybook URL>?path=/story/components-footer--default` and run the snippet from the example above in the devtools console


packages/web-components/src/components/footer/footer-composite.ts
```diff
@@ -406,7 +406,8 @@ class C4DFooterComposite extends MediaQueryMixin(
 
         <c4d-legal-nav
           size="${ifDefined(size)}"
-          navLabel="${ifDefined(navLabel)}">
+          navLabel="${ifDefined(navLabel)}"
+          adjunctLinksLenght="${adjunctLinks.length ?? 0}">
```
m4olivei, line 410: There is a typo here, should be `adjunctLinksLength`. Looks like the typo carries to other places its being used here as well.

packages/web-components/src/components/footer/legal-nav.ts
```diff
@@ -45,6 +45,10 @@ class C4DLegalNav extends StableSelectorMixin(LitElement) {
   /**
    * The adjunct links container
    */
+
+  @property()
+  adjunctLinksLenght = 0;
```
m4olivei, line 50: Hum, doesn't look like this gets used? If not used, lets remove all references to it.

m4olivei (dismissed):
Working for the Storybook test case. Noticed some code clean up potentially needed. I also added the CDN label to be able to test it with the Stackblitz example.

packages/web-components/src/components/footer/footer-composite.ts
marcelojcs: Thanks for the review. 
That was my bad. 
It is fixed in my latest commit.

packages/web-components/src/components/footer/legal-nav.ts
marcelojcs: It is being used to trigger the update.  
The `attributeChangedCallback` function will only trigger if the property is declared.

packages/web-components/src/components/footer/legal-nav.ts
```diff
@@ -119,18 +132,8 @@ class C4DLegalNav extends StableSelectorMixin(LitElement) {
...
-    } = this;
-    const hideAdjunctLinksContainer =
-      adjunctLinksSlot?.assignedNodes().length === 0
-        ? adjunctLinksContainer?.classList.add(
-            `${prefix}--adjunct-links__container--hidden`
-          )
-        : '';
-    return hideAdjunctLinksContainer;
+  updated() {
+    this._handleAdjunctLinksVisibility();
```
andy-blum, line 136: This updated method is actually probably removable now.

andy-blum (changes requested):
One piece we forgot on our syncup

packages/web-components/src/components/footer/legal-nav.ts
marcelojcs: Yes, it is removable. 
Addressed in my latest commit

andy-blum (approved):

IgnacioBecerra (approved):

PR #12070, opened 2024-10-09
feat(video-player): enable autoplay & load as muted

andy-blum:
### Related Ticket(s)

https://jsw.ibm.com/browse/ADCMS-6363

### Description

Fixes the `auto-play` attribute to play video on load and adds a `muted` attribute to load the video with sound fully muted.

### Changelog

**New**

- Adds `muted` attribute to `video-player-container`
- Adds storybook examples for `auto-play` and `muted` video player usage

**Changed**

- Fixes `auto-play` attribute on `video-player-container`

### DIFF
--- a/packages/web-components/src/components/video-player/video-player-container.ts
+++ b/packages/web-components/src/components/video-player/video-player-container.ts
@@ -182,7 +182,9 @@ export const DDSVideoPlayerContainerMixin = <
       const { backgroundMode, autoPlay, muted } =
         this as unknown as DDSVideoPlayerComposite;
       let playerOptions = {};
-      const autoplayPreference = this._getAutoplayPreference();
+      const autoplayPreference = autoPlay
+        ? this._getAutoplayPreference()
+        : false;
 
       if (backgroundMode) {
         playerOptions = {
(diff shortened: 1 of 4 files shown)


packages/web-components/src/components/video-player/video-player-container.ts
```diff
@@ -282,15 +283,15 @@ export const DDSVideoPlayerContainerMixin = <
      * @param videoId The video ID.
      * @internal
      */
-    _embedMedia = async (videoId: string, backgroundMode = false) => {
+    _embedMedia = async (videoId: string) => {
```
m4olivei, line 286: This is probably going to be a [breaking change for CAEM](https://github.ibm.com/webstandards/carbon-for-aem/blob/main/src/components/video-player/video-player-container.ts#L94). CAEM relies on being able to call this method with this argument, as it has added `inersectionMode` property to take into consideration.

m4olivei (changes requested):
Need to look into potential breaking change for CAEM.

packages/web-components/src/components/video-player/video-player-container.ts
andy-blum: It might be a breaking change, but I think it's one worth making and then fixing in CAEM.

In video-player-container.ts:
```ts
 _getPlayerOptions() {
      const { backgroundMode, autoPlay, muted } =
        this as unknown as DDSVideoPlayerComposite;
        // ...
}
```

Since we're passing through `intersectionMode || backgroundMode`, we should just make sure that setting `intersectionMode` to `true` also sets `backgroundMode` to `true`

m4olivei: Sounds good, that should work. We'll need to remember to do that together with, or before the next C4IBM update PR in CAEM.

m4olivei (approved):

PR #12081, opened 2024-10-22
fix(video-player): correct a visual regression with background video in some contexts

m4olivei:
### Related Ticket(s)

### Description

Recent changes caused visual regressions due to styles that relied on `[background-video="true"]`. These updates make `background-video` a pure boolean attribute and simplify the style selectors accordingly.

### Changelog

**Changed**

- Make `background-video` a reflected boolean attribute all the time.
- Fix a regression with auto-play video.

### DIFF
--- a/packages/web-components/src/components/video-player/video-player-composite.ts
+++ b/packages/web-components/src/components/video-player/video-player-composite.ts
@@ -1,7 +1,7 @@
 /**
  * @license
  *
- * Copyright IBM Corp. 2020, 2023
+ * Copyright IBM Corp. 2020, 2024
  *
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
@@ -194,7 +194,7 @@ class DDSVideoPlayerComposite extends HybridRenderMixin(
   /**
    * `true` to autoplay, mute, and hide player UI.
    */
-  @property({ type: Boolean, attribute: 'background-mode' })
+  @property({ type: Boolean, attribute: 'background-mode', reflect: true })
   backgroundMode = false;
 
   /**
--- a/packages/web-components/src/components/video-player/video-player-container.ts
+++ b/packages/web-components/src/components/video-player/video-player-container.ts
@@ -182,33 +182,42 @@ export const DDSVideoPlayerContainerMixin = <
       const { backgroundMode, autoPlay, muted } =
         this as unknown as DDSVideoPlayerComposite;
       let playerOptions = {};
-      const autoplayPreference = autoPlay
-        ? this._getAutoplayPreference()
-        : false;
+      const autoplayPreference = this._getAutoplayPreference();
 
-      if (backgroundMode) {
-        playerOptions = {
-          'topBarContainer.plugin': false,
-          'controlBarContainer.plugin': false,
-          'largePlayBtn.plugin': false,
-          'loadingSpinner.plugin': false,
-          'unMuteOverlayButton.plugin': false,
-          'EmbedPlayer.DisableVideoTagSupport': false,
-          loop: true,
-          autoMute: true,
-          autoPlay: autoplayPreference,
-          // Turn off CTA's including mid-roll card and end cards.
-          'ibm.callToActions': false,
-          // Turn off captions display, background/ambient videos have no
-          // audio.
-          closedCaptions: {
-            plugin: false,
-          },
-        };
-      } else {
-        playerOptions = {
-          autoMute: muted,
-        };
+      switch (true) {
+        case autoPlay:
+          playerOptions = {
+            autoMute: muted,
+            autoPlay: autoplayPreference,
+          };
+          break;
+
+        case backgroundMode:
+          playerOptions = {
+            'topBarContainer.plugin': false,
+            'controlBarContainer.plugin': false,
+            'largePlayBtn.plugin': false,
+            'loadingSpinner.plugin': false,
+            'unMuteOverlayButton.plugin': false,
+            'EmbedPlayer.DisableVideoTagSupport': false,
+            loop: true,
+            autoMute: true,
+            autoPlay: autoplayPreference,
+            // Turn off CTA's including mid-roll card and end cards.
+            'ibm.callToActions': false,
+            // Turn off captions display, background/ambient videos have no
+            // audio.
+            closedCaptions: {
+              plugin: false,
+            },
+          };
+          break;
+
+        default:
+          playerOptions = {
+            autoMute: muted,
+          };
+          break;
       }
 
       return playerOptions;
--- a/packages/web-components/src/components/video-player/video-player-container.ts
+++ b/packages/web-components/src/components/video-player/video-player-container.ts
@@ -208,7 +208,6 @@ export const DDSVideoPlayerContainerMixin = <
       } else {
         playerOptions = {
           autoMute: muted,
-          autoPlay: autoplayPreference,
         };
       }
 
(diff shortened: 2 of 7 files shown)


andy-blum:
The leadspace with video isn't playing the video on page load. If I click pause & play again, it will play, but it won't play on load.

https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/12081/iframe.html?args=&id=components-lead-space--super-with-video&viewMode=story

andy-blum:
Also seems like inline videos need to be clicked on twice to start the video?

https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/12081/iframe.html?args=&id=components-lead-space-block--with-video&viewMode=story

m4olivei:
> The leadspace with video isn't playing the video on page load. If I click pause & play again, it will play, but it won't play on load.
> 
> https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/12081/iframe.html?args=&id=components-lead-space--super-with-video&viewMode=story

> Also seems like inline videos need to be clicked on twice to start the video?
> 
> https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/12081/iframe.html?args=&id=components-lead-space-block--with-video&viewMode=story

On my local I needed to go back to the `v1.64.0` tag (`@carbon/web-components@v1.49.0` ) for this to be working again. Looks like it regressed with https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/12070

andy-blum:
The current video-player suite order of operations

## video-player
  * renders thumbnail overlay OR slot, based on several factors
    * Added `!this.autoplay` in the last PR. This was necessary to cause the iframe to render on load rather than the thumbnail
    ```js
      return contentState === VIDEO_PLAYER_CONTENT_STATE.THUMBNAIL &&
        !backgroundMode &&
        !this.autoplay
        ? html`
            <div class="${prefix}--video-player__video">
              <button
                class="${prefix}--video-player__image-overlay"
                @click="${this._handleClickOverlay}">
                <dds-image default-src="${thumbnailUrl}" alt="${ifNonNull(name)}">
                  ${PlayVideo({ slot: 'icon' })}
                </dds-image>
              </button>
            </div>
          `
        : html` <slot></slot> `;
    ```
  * click button
  * dispatch event
    ```js
      new CustomEvent(`dds-video-player-content-state-changed`, {
        bubbles: true,
        composed: true,
        detail: {
          videoId,
          contentState: VIDEO_PLAYER_CONTENT_STATE.VIDEO,
          playingMode: this.playingMode,
          name,
          customVideoDescription,
        },
      })
    ```

## video-player-composite
  * catch event, call this._embedMedia
    ```js
      protected _handleContentStateChange(event: CustomEvent) {
        const { contentState, playingMode, videoId } = event.detail;
        if (
          contentState === VIDEO_PLAYER_CONTENT_STATE.VIDEO &&
          playingMode === VIDEO_PLAYER_PLAYING_MODE.INLINE &&
          videoId
        ) {
          this._embedMedia?.(videoId, this.backgroundMode);
        }
      }
    ```

## video-player-contaienr
  * defines _embedMedia
    * calls this._embedVideoImpl
      ```js
        const embedVideoHandle = await KalturaPlayerAPI.embedMedia(
          videoId,
          playerId,
          this._getPlayerOptions()
        );
      ```
    * calls this._getPlayerOptions()
      ```js
        if (backgroundMode) {
          playerOptions = {
            'topBarContainer.plugin': false,
            'controlBarContainer.plugin': false,
            'largePlayBtn.plugin': false,
            'loadingSpinner.plugin': false,
            'unMuteOverlayButton.plugin': false,
            'EmbedPlayer.DisableVideoTagSupport': false,
            loop: true,
            autoMute: true,
            autoPlay: autoplayPreference,
            // Turn off CTA's including mid-roll card and end cards.
            'ibm.callToActions': false,
            // Turn off captions display, background/ambient videos have no
            // audio.
            closedCaptions: {
              plugin: false,
            },
          };
        } else {
          playerOptions = {
            autoMute: muted,
            // autoPlay: autoplayPreference,
          };
        }
      ```

      These override the default flashvars:
      ```js
        const defaultFlashVars = {
          autoPlay: true,
          closedCaptions: {
            plugin: true,
          },
          titleLabel: {
            plugin: true,
            align: 'left',
            text: '{mediaProxy.entry.name}',
          },
          ibm: {
            template: 'idl',
          },
        };
      ```
      
      * Setting `autoPlay: autoplayPreference` overrides the default values in the Kaltura API. autoplay there is set to `true` so that injected iframes play immediately. This has been removed in the latest commit

andy-blum:
This does cause an accessibility issue - Since we don't override `autoPlay` on non-BG videos, it's always true even when `prefers-reduced-motion` is `true`. We need to find a way to check `autoplayPreference` for videos with the auto-play attribute but _not_ for videos that load the thumbnail first as a user has to interact with that video first and we don't want to make them click "play" twice.

andy-blum:
Should be fixed now. We need to make sure the below behaviors are all exhibited:
* Videos that load behind a thumbnail first should play after the overlay is clicked
* Videos that are `background-mode` should autoplay, muted, with no Kaltura API
  * autoplay should be overridable by user's OS preference or by pausing a background video in the UI
* Videos that are `autoplay` should autoplay on page load
  * autoplay should be overridable by user's OS preference or by pausing a background video in the UI
* videos that are `muted` should load as muted (regardless of loading method)

packages/web-components/src/components/video-player/video-player-composite.ts
```diff
@@ -194,7 +194,7 @@ class DDSVideoPlayerComposite extends HybridRenderMixin(
   /**
    * `true` to autoplay, mute, and hide player UI.
    */
-  @property({ type: Boolean, attribute: 'background-mode' })
+  @property({ type: Boolean, attribute: 'background-mode', reflect: true })
```
m4olivei, line 197: I think we should also reflect this as a Boolean attribute in the stories. Many instances currently read as `background-mode="true"`, eg. https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/be33a5b6c220e776dc9571dab345bd36043190b7/packages/web-components/src/components/leadspace/__stories__/leadspace.stories.ts#L173

The reflection seems to force the changes from `background-mode="true"` to just `background-mode`, so it's not a huge deal, just code style at this point.

packages/web-components/src/components/video-player/video-player-container.ts
```diff
@@ -182,34 +182,42 @@ export const DDSVideoPlayerContainerMixin = <
...
-        };
-      } else {
-        playerOptions = {
-          autoMute: muted,
-          autoPlay: autoplayPreference,
-        };
+      const autoplayPreference = this._getAutoplayPreference();
+
+      switch (true) {
+        case autoPlay:
```
m4olivei, line 188: Is there ever a scenario where the user would set `<dds-video-player-container auto-play background-mode>`? In that case, we'd land in this case first, and it wouldn't get the options for `background-mode`.

m4olivei (commented):
Couple things to note. Otherwise, nice work sleuthing the bug. I don't find any issues.

packages/web-components/src/components/video-player/video-player-composite.ts
m4olivei: Addressed in recent commit https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/12081/commits/5fc572f5b8c2781e6ff8931ebdd60a8f433a2f44

packages/web-components/src/components/video-player/video-player-container.ts
m4olivei: This is probably only a docs issue, but one to watch out for. `background-mode` is inclusive of `auto-play`, and if you want `background-mode` only use `background-mode`.

andy-blum (approved):

PR #12084, opened 2024-10-28
fix(cta): prevent an infinate loop of fetching video details

m4olivei:
### Related Ticket(s)

[ADCMS-6577](https://jsw.ibm.com/browse/ADCMS-6577)

### Description

Prevents an infinite loop in some situations with vidoe CTA cards.

There are a few key parts to the infinite loop:

1. Both `<c4d-card-group-item>` and `<c4d-card-footer>` use `CTAMixin`
2. `CTAMixin` introduces an `update` hook that watches for changes to `videoDescription`, and if it's changed, the `c4d-cta-request-video-data` event is fired.
3. The `<c4d-video-cta-container>` component listens for the `c4d-cta-request-video-data` event, makes a request to Kaltura, and sets the `videoDescription` property for the component that fired the event in step 2, either `<c4d-card-group-item>` or `<c4d-card-footer>`
4. The `<c4d-card-group-item>` component transposes its attributes, including `videoDescription` onto the `<c4d-card-footer>` component. Restarting the cycle.
5. There is a race condition mixed into this infinite loop, I believe due to the async nature of the Kaltura request. Sometimes, things settle down, but other times, we end up in an infinite loop where the two components descending from `CTAMixin` ping pong firing an event, triggering Kaltura request, triggering update cycle, repeat.

Here is an adjustment to the provided Carbon demo that uses the CDN preview of this PR to show the fix:

https://stackblitz.com/github/m4olivei/carbon-demo/tree/video-cta-lockup-pr-preview?file=scripts%2Fprestart.js

### Changelog

**Changed**

- Adjusted CTA logic to guard against Kaltura fetch when it's already been done. Like the conditional shortly before it, we use `videoDuration` as a guard to know whether we need to query Kaltura or not.


packages/web-components/src/component-mixins/cta/cta.ts
```diff
@@ -325,26 +325,31 @@ const CTAMixin = <T extends Constructor<HTMLElement>>(Base: T) => {
...
+                })
+              );
+            });
+        }
       }
 
-      if (ctaType === CTA_TYPE.VIDEO && this.offsetWidth > 0) {
+      if (
+        ctaType === CTA_TYPE.VIDEO &&
+        this?.shadowRoot?.querySelector(':first-child')?.offsetWidth > 0
```
m4olivei, line 351: Ended up using a simpler selector so we're not hardcoding card-specific things. I also checked for other CTAMixin uses, and of those I couldn't find any that both used a `videoThumbnailUrl` and were not `C4DCard` children. Those are:

* C4DButton
* C4DButtonGroupItem
* C4DButtonCTA
* C4DLinkWithIcon
* C4DCardFooter
* C4DFeatureCardFooter
* C4DFeatureCTAFooter
* C4DTextCTA

The other way to look at is to see what is actually reading from `videoThumbnailUrl`. Those places are:

* [card.ts](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/card/card.ts#L163)
* [card-in-card.ts](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/card-in-card//card-in-card.ts#L31)
* [card-cta.ts](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/cta/card-cta.ts#L74)
* [feature-cta.ts](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/cta/feature-cta.ts#L74)

All of those are descendants of `C4DCard`. Does give a good list to test from.

andy-blum:
The [demo link](https://video-cta-lockup--carbon-demos.netlify.app/) from the issue is updated to use the CDN deploy previews from this PR. While the browser no longer locks up, the modal containing the media isn't popping up either.

I'd expect the card-group-item to open a modal, seed in the iframe from kaltura, and then either play the video or error on a domain restriction.

andy-blum (changes requested):
changes needed in previous comment

m4olivei:
> The [demo link](https://video-cta-lockup--carbon-demos.netlify.app/) from the issue is updated to use the CDN deploy previews from this PR. While the browser no longer locks up, the modal containing the media isn't popping up either.
> 
> I'd expect the card-group-item to open a modal, seed in the iframe from kaltura, and then either play the video or error on a domain restriction.

Isn't that the issue that we're working on in [ADCMS-6596](https://jsw.ibm.com/browse/ADCMS-6596)?

andy-blum (approved):

PR #12085, opened 2024-10-29
[tests] fix e2e-storybook tests

m4olivei:
### Description

The e2e-storybook tests have been failing for quite awhile. The tests do not appear to have been updated for v2. In many cases, the tests are using the old v1 prefixes.

The goal here is:

* Fix test suite so that it will run on CI
* Bulk update prefixes in the tests. In some places fix the tests where its simple to do so.
* Skip tests that are broken for now. To follow up later. Lets get clean CI. Intent long term is to evaluate all skipped tests, update them to be of value, or remove them.
* e2e-tests-parallel job should run green

### Changelog

**Changed**

- Update e2e-storybook integration tests with v2 prefixes
- Update CI scripts / config so that e2e-tests-paralllel job runs green


packages/web-components/tests/e2e-storybook/cypress.json
```diff
@@ -14,8 +14,8 @@
   "screenshotOnRunFailure": false,
   "video": false,
   "testFiles": "**/*.e2e.js",
-  "pageLoadTimeout": 90000,
-  "defaultCommandTimeout": 90000,
+  "pageLoadTimeout": 30000,
+  "defaultCommandTimeout": 4000,
```
m4olivei, line 18: This represents a big change in terms of the performance of our tests. See https://docs.cypress.io/app/references/configuration#Timeouts. The values of 90000 previously meant that errant tests were very slow. Especially for `defaultCommandTimeout`, if say a selector failed to match due to an issue with the test, each time we'd look for that selector, the test would stall for 1.5min before failing and moving on.

packages/web-components/package.json
```diff
@@ -88,7 +88,8 @@
     "typecheck": "tsc --noEmit -p tsconfig.json",
     "upgrade-carbon": "yarn upgrade-interactive @carbon/web-components @carbon/icon-helpers @carbon/icons --latest --exact",
     "visual-snapshot": "yarn percy storybook:start ./storybook-static",
-    "wca": "web-component-analyzer analyze src --outFile custom-elements.json"
+    "wca": "web-component-analyzer analyze src --outFile custom-elements.json",
+    "cypress:verify": "cypress verify"
```
andy-blum, line 92: What does this command do?

packages/web-components/tests/e2e-storybook/cypress/integration/footer/footer-short.e2e.js
```diff
@@ -80,18 +83,17 @@ describe('c4d-footer | Short (desktop)', () => {
     cy.takeSnapshots();
   });
 
-  it('should display interactive search field and with keywords for locations and languages', () => {
+  it.skip('should display interactive search field and with keywords for locations and languages', () => {
```
andy-blum, line 86: Missed `skip`? or is this deliberate?

packages/web-components/tests/e2e-storybook/cypress/integration/carousel/carousel.e2e.js
```diff
@@ -74,39 +74,45 @@ const _tests = {
     });
   },
   screenshotThemes: () => {
-    it('should render correctly in all themes', () => {
+    it.skip('should render correctly in all themes', () => {
```
andy-blum, line 77: skipped?

packages/web-components/tests/e2e-storybook/cypress/integration/carousel/carousel.e2e.js
```diff
@@ -74,39 +74,45 @@ const _tests = {
...
       cy.carbonThemesScreenshot({
         capture: 'viewport',
       });
     });
   },
   checkTextRenders: () => {
-    it('should render card text and arrow icon', () => {
+    it.skip('should render card text and arrow icon', () => {
```
andy-blum, line 84: skipped?

packages/web-components/tests/e2e-storybook/cypress/integration/carousel/carousel.e2e.js
```diff
@@ -122,15 +128,15 @@ const _tests = {
     });
   },
   checkVideoRenders: () => {
-    it('should render the video thumbnail and play button', () => {
+    it.skip('should render the video thumbnail and play button', () => {
```
andy-blum, line 131: skipped?

packages/web-components/tests/e2e-storybook/cypress/integration/carousel/carousel.e2e.js
```diff
@@ -141,10 +147,10 @@ const _tests = {
     });
   },
   checkVideoDurationText: () => {
-    it('should render the video duration in the footer', () => {
+    it.skip('should render the video duration in the footer', () => {
```
andy-blum, line 150: skipped?

andy-blum (approved):
There's a lot of tests we've updated to `skip` - Many seem to be "did it render right?" kinds of tests, which we should probably just delete to let Percy handle them, but the TOC, for example, has some functional tests skipped.

Other than that it's basically all license, formatting and prefix updates.

packages/web-components/package.json
m4olivei: [This command](https://docs.cypress.io/app/references/command-line#cypress-verify) will:

> Verify that Cypress is installed correctly and is executable.

Before we were running into these errors a lot, which were causing the whole test suite to fail:

image

After some googling of the issue, this was the recommended fix. It seems to make the process of finding the cypress executable before the test runs start reliable.

packages/web-components/tests/e2e-storybook/cypress/integration/footer/footer-short.e2e.js
m4olivei: All the `.skip`'s here are intentional. My approach here was not to go through the entire suite and debug and fix each one, but to cover the most obvious fixes (prefix updates), review the result, and for anything more complex needing a deeper understanding of the individual test, mark as skipped for later follow up.

The idea is, even with all the tests marked as skipped, there is still plenty of value here that could be useful while we chip away at the skipped tests.

andy-blum: That works for me. After we finish the v2 upgrade, I can split the remaining `skips` into tickets for the team and we can look at what holes we might have in coverage.

PR #12089, opened 2024-10-31
[feature] Add cart icon to Masthead for SAP Commerce integration

m4olivei:
### Related Ticket(s)

[ADCMS-6463](https://jsw.ibm.com/browse/ADCMS-6463)

### Description

Adds a new `<c4d-masthead-cart>` component to the Masthead. Needs to be opted in using `has-cart` attribute on the `<c4d-masthead-container>` component. When opted in, there is internal logic to read a cookie that will determine if the cart should be shown.

### Testing instructions

The cart icon functionality is controlled by two new knobs on the Masthead > Default story. The "show the cart functionality (has-cart)" checkbox toggles whether or not the cart renders at all. The "mock active cart id" will affect the cookie that the cart is looking for to show / hide. Anytime you change the "mock active cart id", you need to toggle the "show the cart functionality (has-cart)" so that the `<c4d-masthead-card>` component will re-render and check the cookie value. The cart icon should only appear when "show the cart functionality (has-cart)" is set **and** there is a non-empty value for "mock active cart id".

Check visuals against [Figma specs](https://www.figma.com/design/aQ0mMVEnxwDDoupkKeaNHc/Ecommerce-Mid-Fidelity-(delivery-file)?node-id=15928-64507&node-type=instance&t=bHBr0X1ykXryXSFP-0).

### Changelog

**New**

- `<c4d-masthead-cart>` component, automaticallly placed in `<c4d-masthead-global-bar>` when opted into via `has-cart` attribute on `<c4d-masthead-container>`

### DIFF
--- a/packages/web-components/src/components/masthead/masthead-cart.ts
+++ b/packages/web-components/src/components/masthead/masthead-cart.ts
@@ -1,0 +1,68 @@
+/**
+ * @license
+ *
+ * Copyright IBM Corp. 2024
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import StableSelectorMixin from '../../globals/mixins/stable-selector';
+import { html, LitElement } from 'lit';
+import { property } from 'lit/decorators.js';
+import settings from '@carbon/ibmdotcom-utilities/es/utilities/settings/settings.js';
+import { carbonElement as customElement } from '@carbon/web-components/es/globals/decorators/carbon-element';
+import ShoppingCart20 from '@carbon/web-components/es/icons/shopping--cart/20.js';
+import styles from './masthead.scss';
+
+const { prefix, stablePrefix: c4dPrefix } = settings;
+
+/**
+ * The Cart icon in the masthead.
+ *
+ * @element c4d-masthead-cart
+ * @csspart cart-link - The masthead cart link. Usage: `c4d-masthead-cart::part(cart-link)`
+ */
+@customElement(`${c4dPrefix}-masthead-cart`)
+class C4DMastheadCart extends StableSelectorMixin(LitElement) {
+  /**
+   * The `aria-label` attribute for the link.
+   */
+  @property({ attribute: 'link-label' })
+  linkLabel = 'Cart';
+
+  /**
+   * Tracks whether the user has an active cart to control the display.
+   * @todo flesh out the logic for reading the cookie
+   */
+  @property({ attribute: 'has-active-cart', reflect: true, type: Boolean })
+  hasActiveCart = true;
+
+  connectedCallback() {
+    super.connectedCallback();
+    // @todo flesh out the logic for reading the cookie
+  }
+
+  render() {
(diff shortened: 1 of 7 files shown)


packages/web-components/src/components/masthead/masthead-cart.ts
```diff
@@ -0,0 +1,85 @@
...
+   * The `aria-label` attribute for the link.
+   */
+  @property({ attribute: 'link-label' })
+  linkLabel = 'Cart';
+
+  /**
+   * Tracks whether the user has an active cart to control the display.
+   */
+  @property({ attribute: 'has-active-cart', reflect: true, type: Boolean })
+  hasActiveCart = false;
```
andy-blum, line 40: Should this be something configurable by an author? Perhaps we should use state and toggle a class on the rendered link?

packages/web-components/src/components/masthead/masthead-composite.ts
```diff
@@ -1179,6 +1180,11 @@ class C4DMastheadComposite extends HostListenerMixin(LitElement) {
         `;
   }
 
+  protected _renderCart() {
+    const { hasCart } = this;
+    return hasCart ? html`<c4d-masthead-cart></c4d-masthead-cart>` : undefined;
```
andy-blum, line 1185: I think we probably need a way to pass through a value to `link-label` from the `masthead-container`

andy-blum (changes requested):

packages/web-components/src/components/masthead/masthead-cart.ts
m4olivei: Good call. I've refactored based on your suggestion to just toggle a `hidden` attribute when we shouldn't be showing the cart. It simplifies the render method as well as the Sass. Let me know what you think.

packages/web-components/src/components/masthead/masthead-composite.ts
m4olivei: 🤔 , if I look at the other icons, `<c4d-masthead-contact>` [does something similar](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/src/components/masthead/masthead-composite.ts#L1176-L1177) via `@property({ attribute: false })`. Want to do the same, or should be go with a reflected attribute?

andy-blum: Yeah that seems like a good pattern to follow - we just need a way for use of the `masthead-container` to be able to pass in the translated value for the link label.

m4olivei: Might be better as a follow up to address these hidden label translation issues across the <c4d-masthead-global-bar>` items.

andy-blum:
Adding the CDN preview label so we can try to test on prod

m4olivei:
@andy-blum we've got design approval here now. Anything else you need? I was able to test on prod, and it looks good. I left notes in the ticket if you would like to check.

andy-blum (approved):

PR #12097, opened 2024-11-04
(card): Remove class form footer to make component clickable

marcelojcs:
### Related Ticket(s)
 
 [JIRA](https://jsw.ibm.com/browse/ADCMS-6596)
### Description

PR to solve 
> _Card's shadowroot element matching .cds--card__footer has a pseudo-element that blocks click access to the slotted link._

### Changelog
Removed class name from footer wrapper on the `card` component

### DIFF
--- a/packages/web-components/src/components/card/card.ts
+++ b/packages/web-components/src/components/card/card.ts
@@ -258,7 +258,7 @@ class C4DCard extends CTAMixin(StableSelectorMixin(CDSLink)) {
           ${hasPictogram && this.pictogramPlacement === PICTOGRAM_PLACEMENT.TOP
             ? this._renderCopy()
             : ''}
-          <div part="footer-wrapper">
+          <div part="footer-wrapper" class="${prefix}--card__footer-wrapper">
             <slot name="footer"></slot>
           </div>
         </div>
--- a/packages/web-components/src/components/card/card.ts
+++ b/packages/web-components/src/components/card/card.ts
@@ -258,7 +258,7 @@ class C4DCard extends CTAMixin(StableSelectorMixin(CDSLink)) {
           ${hasPictogram && this.pictogramPlacement === PICTOGRAM_PLACEMENT.TOP
             ? this._renderCopy()
             : ''}
-          <div part="footer-wrapper" class="${prefix}--card__footer">
+          <div part="footer-wrapper">
             <slot name="footer"></slot>
           </div>
         </div>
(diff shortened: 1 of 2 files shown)


packages/web-components/src/components/card/card.ts
```diff
@@ -258,7 +258,7 @@ class C4DCard extends CTAMixin(StableSelectorMixin(CDSLink)) {
           ${hasPictogram && this.pictogramPlacement === PICTOGRAM_PLACEMENT.TOP
             ? this._renderCopy()
             : ''}
-          <div part="footer-wrapper" class="${prefix}--card__footer">
+          <div part="footer-wrapper">
```
m4olivei, line 261: Removing this class is good, however, I think we need to keep some class here, suggesting `footer-wrapper`, so that we can hang some styles off of it which are now missing for the footer / link placement.

```suggestion
          <div part="footer-wrapper" class="${prefix}--card__footer-wrapper">
```

I'm _guessing_ it's then this style rule selector that needs updating to `footer-wrapper`:

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/styles/scss/components/card/_card.scss#L463

Maybe also this one:

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/styles/scss/components/card/_card.scss#L534

m4olivei (changes requested):

packages/web-components/src/components/card/card.ts
andy-blum: It's definitely _not_ that first one. That use of `__footer` is combined with all the rules targeting card link elements. We don't want our wrapper to have the same styles as the links that will be inside that wrapper.

I also don't think we need to update the classes in the second linked chunk either. In v1, this div didn't exist, so there shouldn't be any styles targeting it.

I'm comfortable marking this as ready for merge and we'll deal with any issues that pop up if they happen.

andy-blum (approved):

packages/web-components/src/components/card/card.ts
m4olivei: We at least need the `margin-block-start: auto;` from the first one. I think that would be enough to put the arrows in the right place and avoid the visual regressions.

andy-blum: I don't know that we want to remove the __footer class from the code - is it in use anywhere in AEM via the styles package?

PR #12098, opened 2024-11-05
fix(content-block-segmented): remove content-layout class

andy-blum:
### Related Ticket(s)

https://jsw.ibm.com/browse/ICIUC-544

### Description

Removes the `content-layout` class from the shadowroot of the content-block-segmented element. This class added extra margin to the component.

V1
Screenshot 2024-11-05 at 10 19 46 AM

V2 _current_
Screenshot 2024-11-05 at 10 20 23 AM

V2 _with fix_
Screenshot 2024-11-05 at 10 20 05 AM

### Changelog

**Changed**

- Fixed alignment bug in content-block-segmented

### DIFF
--- a/packages/web-components/src/components/content-block-segmented/content-block-segmented.ts
+++ b/packages/web-components/src/components/content-block-segmented/content-block-segmented.ts
@@ -35,7 +35,6 @@ class C4DContentBlockSegmented extends C4DContentBlock {
       _hasMedia: hasMedia,
     } = this;
     return classMap({
-      [`${prefix}--content-layout`]: true,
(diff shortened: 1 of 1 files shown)


m4olivei (approved):
Noticing one visual regression: 

image

However, that should be able to be addressed from global CSS.

Also wondering if we could be cleaning up styles related to `${prefix}--content-layout`, but that can be a follow up.

Nothing blocking, so approving.

andy-blum:
closing this as the class removal has wider implications than just grid alignment. we'll recommend adopters apply their own styles. In the future we should look to deprecate these components that are layout-only.

PR #12099, opened 2024-11-05
feat(countdown): add components

Valentin-Sorin-Nicolae:
### Related Ticket(s)

[ADCMS-6655](https://jsw.ibm.com/browse/ADCMS-6655)

### Description

Migrates the `countdown `component from carbon-for-aem into carbon-for-ibmdotcom

### Changelog

**New**

- add `c4d-countdown`

### DIFF
--- a/packages/web-components/src/components/countdown/__stories__/countdown.stories.ts
+++ b/packages/web-components/src/components/countdown/__stories__/countdown.stories.ts
@@ -1,0 +1,92 @@
+import { html } from 'lit-element';
+import '../index';
+import ifNonEmpty from '@carbon/web-components/es/globals/directives/if-non-empty';
+import { bxGrid16Col } from '../../../globals/internal/storybook-decorators';
+
+import readme from './README.stories.mdx';
+
+const msInDay = 86400000;
+const twoWeeksFromNowTimestamp =
+  Number(new Date().getTime().toString()) + 14 * msInDay;
+const twoWeeksFromNowISO = new Date(twoWeeksFromNowTimestamp).toISOString();
+
+export default {
+  title: 'Components/Countdown',
+  parameters: {
+    ...readme.parameters,
+  },
+  argTypes: {
+    targetDate: {
+      control: { type: 'date' },
+      name: 'Target Date',
+      defaultValue: twoWeeksFromNowISO,
+    },
+    separator: {
+      control: { type: 'text' },
+      name: 'Separator',
+      defaultValue: ', ',
+    },
+    labelType: {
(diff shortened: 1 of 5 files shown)


Valentin-Sorin-Nicolae:
I couldn't add the snapshots, this is not working on my machine.

packages/web-components/src/components/countdown/__stories__/countdown.stories.ts
```diff
@@ -0,0 +1,97 @@
...
+      name: 'Separator',
+      defaultValue: ', ',
+    },
+    labelType: {
+      control: { type: 'select' },
+      options: ['long', 'short', 'narrow', 'none'],
+      name: 'Label Type',
+      defaultValue: 'long',
+    },
+  },
```
m4olivei, line 44: These controls are not showing up in Storybook. I believe you need to convert these to knobs. See other functioning stories with with knobs in the repo.

packages/web-components/src/components/countdown/countdown.ts
```diff
@@ -0,0 +1,229 @@
...
+const getFormatters = (locale: Locale, labelType: UnitDisplay) => {
+  const lc_cc = `${locale.lc}-${locale.cc}`;
+
+  return Object.fromEntries(
+    units.map((unit) => [
+      `to_${unit}s`,
+      new Intl.NumberFormat(lc_cc, {
+        style: 'unit',
+        unit,
+        unitDisplay: labelType,
```
m4olivei, line 39: Typescript is not happy here. I believe it's b/c we may have a value of `none`, which is [not allowed for this option](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#unitdisplay). I think if we adjust the `UnitDisplay` type on line 72 to append on `| undefined`, and then add the following if statement below line 31, we should be good:

```
// The Typescript compiler sees unitDisplay as type
// "short" | "long" | "narrow" | undefined. Convert 'none' to undefined to
// avoid passing an unknown value to the unitDisplay option.
if (labelType === 'none') {
  labelType = undefined;
}
```

m4olivei (changes requested):

m4olivei:
I believe I've resolved all the outstanding issues here. Note that I've merged the [promo-banner migration PR here](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/12100). That PR should be merged first.

m4olivei (approved):
Looks good now.

PR #12091, opened 2024-11-01
feat(tile, tile-group): add components

andy-blum:
### Related Ticket(s)

https://jsw.ibm.com/browse/ADCMS-6691

### Description

Migrates the `tile` and `tile-group` components from `carbon-for-aem` into `carbon-for-ibmdotcom`

### Changelog

**New**

- adds `c4d-tile`
- adds `c4d-tile-group`

PR #12103, opened 2024-11-12
refactor(tile): use relative paths for imports in the same pkg

m4olivei:
### Related Ticket(s)

[ADCMS-6691](https://jsw.ibm.com/browse/ADCMS-6691)
Follow up to https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/12091

### Description

Quick followup to https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/12091. Updates imports for the same package to be relative paths. Can avoid weird local development / bundling bugs and maintains consistency with the rest of the codebase.

### Changelog

**Changed**

- Adjusts `import '@carbon/ibmdotcom-web-components/es*';` imports to relative path imports.

### DIFF
--- a/packages/web-components/src/components/tile/__stories__/tile-group.stories.ts
+++ b/packages/web-components/src/components/tile/__stories__/tile-group.stories.ts
@@ -10,7 +10,7 @@
 import { html } from 'lit';
 import '../index';
 import readme from './README--tile-group.stories.mdx';
-import '@carbon/ibmdotcom-web-components/es/components/image';
+import '../../image';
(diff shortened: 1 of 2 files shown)


andy-blum (approved):

m4olivei:
@andy-blum fyi, I approved the Percy changes reported on the tile component. Second time I've done that. They are reported as different everytime on account of the randomized lorem ipsum. Maybe we need a knob for a seed value somehow to get "consistent" randomness.

PR #12106, opened 2024-11-13
fix(masthead-l1): set default active item based on host & pathname

andy-blum:
### Related Ticket(s)

https://jsw.ibm.com/browse/ADCMS-6689

### Description

Modifies the fallback behavior of marking the "active" L1 items. 

Note how the L1 behaves on https://www.ibm.com/products/watsonx-orchestrate/ai-agent-for-hr
Screenshot 2024-11-13 at 10 04 31 AM

compared to https://www.ibm.com/products/watsonx-orchestrate/ai-agent-for-hr?lnk=hpUSrc4 (same url, with query params)
Screenshot 2024-11-13 at 10 05 12 AM

### Changelog

**Changed**

- updates L1 active item fallback behavior to ignore query parameters


packages/web-components/src/components/masthead/masthead-l1.ts
```diff
@@ -933,9 +933,20 @@ class C4DMastheadL1 extends StableSelectorMixin(LitElement) {
     }
     // Fall back to automated selection based on URL.
     else {
-      this.selectedElements = allLinks.filter(
-        (el) => el.href === currentUrlPath
-      );
+      this.selectedElements = allLinks.filter((el) => {
+        try {
+          el.href === currentUrlPath;
```
m4olivei, line 938: What does this line do? It's a strict equality expression, but we're not doing anything with the result. Is this for purposes of triggering an error condition if one exists here?

m4olivei (approved):
Works as advertised. I'm curious about one line in the code (see comment). Feel free to address/ignore and move forward.

packages/web-components/src/components/masthead/masthead-l1.ts
andy-blum: The `try` block isn't needed for this line, but for the next lines where we set `new URL()`. If a value doesn't parse to a proper URL, it'll throw an exception that we need to catch.

Screenshot 2024-11-19 at 10 36 55 AM

There's a relatively new addition to the URL prototype we could use instead - the `parse` static method, but browser support still isn't quite there.

Screenshot 2024-11-19 at 10 38 21 AM

PR #12108, opened 2024-11-18
feat(masthead): update simple megamenu

andy-blum:
### Related Ticket(s)

https://jsw.ibm.com/browse/ADCMS-6733

### Description

Updates the "simple" megamenu to add a title element to the left-side portion.

### Changelog

**Changed**

- Updated simple megamenu layout


packages/web-components/src/components/masthead/masthead-composite.ts
```diff
@@ -300,20 +300,19 @@ class C4DMastheadComposite extends HostListenerMixin(LitElement) {
...
-                  this._renderMegapanelLinkGroup(group, {
-                    headingLevel: 2,
-                    autoid: `${c4dPrefix}--masthead__l0-nav-list${i}`,
-                  })
-                )}
-              </c4d-megamenu-left-navigation>
-            `
-          : null}
+        <c4d-megamenu-left-navigation>
+          <span>${title}</span>
```
m4olivei, line 308: Is this supposed to be a link? Also, have we considered if this should be a heading?

m4olivei (commented):
I'm not sure what level of design review this needs, but quick look and it seems the design intent for the left indentation of the new title

Figma:
image

Storybook:
image

packages/web-components/src/components/masthead/masthead-composite.ts
andy-blum: It should not be a link, I considered a heading, but wasn't sure if that was the right semantics. Considering also adding `aria-hidden` to make this a presentation-only element.

m4olivei: Right, on closer look that seems fine. Especially considering the `<c4d-megamenu-right-navigation>` can have an `<h2>` (see below). I mean, if design is saying that the repetition is useful for sighted users, it should follow that its useful for non-sighted users, so I'd avoid using `aria-hidden`.

Also we should probably add a `part` attribute to the `<span>`.

PR #12117, opened 2024-11-21
fix(leadspace): in storybook adding long copy to copy knob led to button being clipped away

bruno-amorim:
### Related Ticket(s)

[ADCMS-6827](https://jsw.ibm.com/browse/ADCMS-6827)

### Description

Adding **long** text to the leadspace copy knob leads to clipping the text & CTAs.

Issue:
image

Fixed:
image

### NOTE
There's an apparent bug in all heading/title knobs in all components, when you try to modify it, we get the below error.
I've been trying to debug this error, but with @m4olivei 's help, we found out it's an issue that's been sitting there for a while now, so It would just be very time consuming for the tiny fix that was the original scope of this ticket
example3

### DIFF
--- a/packages/styles/scss/components/leadspace/_leadspace.scss
+++ b/packages/styles/scss/components/leadspace/_leadspace.scss
@@ -528,7 +528,7 @@ $btn-min-width: 26;
   :host(#{$c4d-prefix}-leadspace)[size='medium'] {
     @include breakpoint(lg) {
       .#{$c4d-prefix}--leadspace__overlay {
-        block-size: 30rem;
+        min-block-size: 30rem;
       }
     }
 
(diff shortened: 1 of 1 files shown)


andy-blum (changes requested):
While this fixes the `medium` leadspace, `super`, `tall`, and `short` still have the ability to overflow & clip

bruno-amorim:
@andy-blum This latest commit will address your last comment. Thank you!

m4olivei (changes requested):
I'm seeing a visual regression here for the default values. Most obvious in Super. Comparing current v2 to the deploy preview here:

https://github.com/user-attachments/assets/0bc1374b-0088-4c58-9232-cee07f08af2b

As a sidenote, I was surprised to realize that the Leadspace stories aren't represented in Percy.

andy-blum (approved):

m4olivei (approved):

PR #12128, opened 2024-11-27
fix(card-group): set min height to content

marcelojcs:
### Related Ticket(s)

[JIRA](https://jsw.ibm.com/browse/ADCMS-6840)

### Description

Adding min height to card-group component to ensure the elements have the correct height.

### Changelog

**New**
Adding min height to card-group component to ensure the elements have the correct height.


packages/styles/scss/components/card-group/_card-group.scss
```diff
@@ -80,10 +80,8 @@
       }
 
       .#{$prefix}--card__content {
-        display: grid;
-        grid-row: span 10;
-        grid-template-rows: subgrid;
-        row-gap: 0;
+        display: flex;
+        min-block-size: 10rem;
```
m4olivei, line 84: Hum, this won't work, as it breaks the [fixes for sameheight](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/12063).

image

Compare to: https://carbon-design-system.github.io/carbon-for-ibm-dotcom/next/web-components/?path=/story/components-card-group--default

m4olivei (changes requested):

m4olivei (approved):

PR #12136, opened 2024-11-29
fix(footer): adjustments to adjunct links visibility

m4olivei:
### Related Ticket(s)

[ADCMS-7082](https://jsw.ibm.com/browse/ADCMS-7082)

### Description

Make some adjustments to the logic inside the legal-nav component for adjunct-links visibility. Before the fix here, the `@slotchange` event was not firing as expected when there were no adjunct links to hide them. Further, once the `hidden` class got correctly set on the `adjunct-links` wrapper, needed to fix some Sass to actually hide it.

### Changelog

**Changed**

- Hide adjunct-links container when there are no adjunct links.


bruno-amorim (approved):
Nice work, @m4olivei! Approved on my end. Thank you!

PR #12143, opened 2024-12-03
fix(button): avoid taking any video name bubbled on the page

m4olivei:
### Related Ticket(s)

[ADCMS-7088](https://jsw.ibm.com/browse/ADCMS-7088)

### Description

As it was, button elements that had no label would adopt any video name fetched by the `<c4d-video-cta-container>` component, which may have been triggered by a completely unrelated CTA. This change now guards against that by ignoring events when the button is not a `cta-type="video"` or does not have the same video id as the fetched video.

### Changelog

**Changed**

- Fixes an issue with buttons not using a label placed on a page with other unrelated video CTAs

### DIFF
--- a/packages/web-components/src/components/button/button.ts
+++ b/packages/web-components/src/components/button/button.ts
@@ -7,16 +7,16 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import { LitElement, html } from 'lit';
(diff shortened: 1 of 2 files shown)


andy-blum (approved):
Change seems sound - approving pending tests

PR #12144, opened 2024-12-03
fix(toc): remove implicit top and bottom padding for TOC

m4olivei:
### Related Ticket(s)

[ADCMS-7171](https://jsw.ibm.com/browse/ADCMS-7171)

### Description

Removes the top and bottom implicit padding on the TOC contents in the shadow root.

### Changelog

**Changed**

- Removes the internal top and bottom padding from TOC contents to leave to adopter as in v1


bruno-amorim (approved):
Nice work Matt! Looks good to me!!

andy-blum (approved):

PR #12151, opened 2024-12-06
fix(button): fix disable state of the button

m4olivei:
### Related Ticket(s)

[ADCMS-6961](https://jsw.ibm.com/browse/ADCMS-6961)

### Description

Fixes some odd icon logic left in the button component. The fix allows us to consistently re-render the correct CTA type icon whenever a change in the parent component may mean a different `<slot name="icon">` element is rendered. This allows for consistency across disable/active states.

### Testing instructions

* Toggling disable knob and/or changing the `href` should maintain the expected icon and reflect the relevant state.
* Change CTA type knob state to make sure it reflects the right value

### Changelog

**Changed**

- Fixed icon issues in the button component for disable state

### DIFF
--- a/packages/web-components/src/components/button/button.ts
+++ b/packages/web-components/src/components/button/button.ts
@@ -141,18 +141,20 @@ class C4DButton extends CTAMixin(StableSelectorMixin(CDSButton)) {
       'tooltipText',
       'href',
     ];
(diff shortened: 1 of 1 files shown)


marcelojcs (commented):
All works as expected

andy-blum (approved):

PR #12152, opened 2024-12-06
chore(styles): properly handle to-rem func calls

m4olivei:
### Related Ticket(s)

None. Noticed in a Slack conversation.

### Description

There are many instances of using the `to-rem` utliity funcation that don't include the correct module import, hence `to-rem` makes it into compiled CSS, which invalidates the style rule.

### Changelog

**Changed**

- Corrects `to-rem` usage by `@use` the appropriate module.


m4olivei:
Percy changes here look like they are legitimately fixes and were bugs that have not been noticed.

andy-blum (approved):

PR #12158, opened 2024-12-11
fix(background-media): avoid potential VPC setup race

andy-blum:
### Related Ticket(s)

https://jsw.ibm.com/browse/ADCMS-7249

### Description

Modifies component logic to avoid a possible race condition. Previously we assumed the background media component was setting up after the video player container had created its video player child.

### Changelog

**Changed**

- avoid possible race condition in with background videos

### DIFF
--- a/packages/web-components/src/components/background-media/background-media.ts
+++ b/packages/web-components/src/components/background-media/background-media.ts
@@ -93,12 +93,6 @@ class C4DBackgroundMedia extends C4DImage {
   mobilePosition = MOBILE_POSITION.BOTTOM;
 
   /**
-   * Current state of video playback
-   */
-  @property()
-  videoIsPlaying = false;
-
-  /**
    * Query selector to get the child video player container
    */
   protected get videoPlayerContainer() {
@@ -134,6 +128,8 @@ class C4DBackgroundMedia extends C4DImage {
     } else {
       video?.playAllVideos();
     }
+
+    this.requestUpdate();
   }
 
   renderVideoControls() {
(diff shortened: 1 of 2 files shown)


packages/web-components/src/components/background-media/background-media.ts
```diff
@@ -122,40 +119,36 @@ class C4DBackgroundMedia extends C4DImage {
...
-    ) {
+    if (assignedImages.length && !assignedVideos.length) {
       this.containsOnlyImages = true;
     }
-
-    if (assignedVideos.length) {
-      const [video] = assignedVideos;
-      this.videoId = video.getAttribute('video-id');
-      this.videoPlayer = video.querySelector(`${c4dPrefix}-video-player`);
-      this.videoIsPlaying = (video as C4DVideoPlayerContainer).isPlaying;
```
m4olivei, line 138: Can we remove the `videoIsPlaying` property altogether from this component? Doesn't look like its used anymore.

m4olivei (approved):
Small nit pick. Looks like we can clean up an unused class property. Otherwise looking good. The changes make logical sense, its cleaner and more direct to go through the video player container rather than the video player. 

I wasn't able to reproduce the reported issue in AEM on my local, but using these changes in AEM did not cause any regressions for me, working as expected there.

Feel free to address the nit or not and move forward. Nice work.

m4olivei (changes requested):
Wait sorry, too quick.

In Storybook, (and in local AEM), I'm not seeing the button toggle between play / pause visually along with the video playback itself. Its also not remembering auto-play preferences 😕 .

m4olivei (changes requested):
We're still not getting the autoplay preference behavior happening for bg video. Thats still a regression we should fix. I'm seeing that with the updated approach here, the `
c4d-background-video-prefers-autoplay` localstorage key does not get set, compared to the latest rc3 Storybook instance where it does and is respected on a reload of the Story.

m4olivei (approved):
Looks good now! Thanks for the quick fixes.

PR #12159, opened 2024-12-11
fix(link-with-icon): fix icon placement

m4olivei:
### Related Ticket(s)

[ADCMS-7258](https://jsw.ibm.com/browse/ADCMS-7258)

### Description

This corrects a bad prefix conversion from v1 to v2 that meant that the necessary styles required to correctly reflect the `icon-placement="left"` on a `<c4d-link-with-icon>` component were not present. Note that the way Storybook knobs work was hiding this issue in the Link with Icon story. The bug can be reproduced by going to the [Link with icon story in the v2 Storybook](https://carbon-design-system.github.io/carbon-for-ibm-dotcom/next/web-components/?path=/story/components-link-with-icon--default), setting "Icon Position" to left, then opening in a new tab. Test in the deploy preview in the same way to ensure the bug is fixed.

### Testing instructions

* Open the Link with Icon > Default story
* Set the "Icon Position" knob to left
* Open in a new tab
* Icon should now appear as expected
* Also go through the Footer and Locale Modal stories to check the reported link icon is in the right spot.

### Changelog

**Changed**

- Fixed left icon placement in Link with icon component

### DIFF
--- a/packages/styles/scss/components/link-with-icon/_link-with-icon.scss
+++ b/packages/styles/scss/components/link-with-icon/_link-with-icon.scss
@@ -1,5 +1,5 @@
 /**
- * Copyright IBM Corp. 2016, 2023
+ * Copyright IBM Corp. 2016, 2024
  *
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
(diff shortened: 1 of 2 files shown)


andy-blum (approved):
This also fixes the bug seen in the current storybook where changing the icon position and then adding more items resulted in mixed placements.

PR #11248, opened 2023-12-18
chore(deps): update linting packages

kennylam:
### Related Ticket(s)

Closes #10375

### Description

This removes local linting packages in favor of hosted ones.

### Changelog

**New**

- add `eslint-config-carbon` and `stylelint-config-carbon` linting packages

**Changed**

- update styles to use CSS logical properties
- update lint configs
- `packages/storybook-images` folder moved to `packages/web-components/.storybook/`
- various configs moved to package.json and standalone files removed

**Removed**
- local lint packages removed
  - `eslint-config-ibmdotcom`
  - `eslint-plugin-ibmdotcom-import-rules`
  - `eslint-plugin-react-prop-type-comments`
  - `stylelint-config-ibmdotcom`

PR #12160, opened 2024-12-11
fix(content-group-cards): fix alignment

andy-blum:
### Related Ticket(s)

Closes # 
https://jsw.ibm.com/browse/ADCMS-7030

### Description

In medium resolution, the cards are having an undesired horizontal overflow that was causing a scroll bar to appear.
I made a fix that will make them span full width, that's the behavior we have in the [v2 builds I found](https://ibmdotcom-webcomponents.s3.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/11248/index.html?path=/story/overview-getting-started--page) for example, that came [from this issue](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/11248)

Fixed as per this PR:
image

### Changelog

**New**

setting `margin-inline-start` to be 0 in resolutions lower than LG.

### DIFF
--- a/packages/styles/scss/components/content-group-cards/_content-group-cards.scss
+++ b/packages/styles/scss/components/content-group-cards/_content-group-cards.scss
@@ -38,7 +38,6 @@
 
   .#{$prefix}--content-group-cards__row {
     display: grid;
-    margin-inline: -$spacing-05;
 
     @include breakpoint(md) {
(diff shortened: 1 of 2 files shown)


m4olivei (approved):
Looks good! Cleaner than where we were getting with another approach. cc @bruno-amorim

PR #12164, opened 2024-12-12
fix(callout-quote): fix horizontal scroll for callout quote

m4olivei:
### Related Ticket(s)

[ADCMS-7277](https://jsw.ibm.com/browse/ADCMS-7277)

### Description

Fixes horizontal scroll on the callout quote component at the mobile breakpoint. While here, also make slight adjustments to inline margin for the nested `<c4d-hr>` component within the callout quote component, [per the spec](https://www.figma.com/design/KQdUduecrBf4DCjRZnsl25/Layout-component--callout-quote--visual-specs--2.1.0?node-id=0-106&node-type=frame&t=y1vGtscmEVeFVKKv-0).

### Changelog

**Changed**

- Fixes horizontal scroll on the callout quote component at the mobile breakpoint.

### DIFF
--- a/packages/styles/scss/components/callout-quote/_callout-quote.scss
+++ b/packages/styles/scss/components/callout-quote/_callout-quote.scss
@@ -20,6 +20,10 @@
 @include quote;
 
 @mixin callout-quote {
+  * {
+    box-sizing: border-box;
+  }
+
   :host(#{$c4d-prefix}-callout-quote) {
     @extend :host(#{$c4d-prefix}-callout);
     @extend :host(#{$c4d-prefix}-callout-text);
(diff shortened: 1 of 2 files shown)


packages/styles/scss/components/callout-quote/_callout-quote.scss
```diff
@@ -20,6 +20,10 @@
 @include quote;
 
 @mixin callout-quote {
+  * {
+    box-sizing: border-box;
+  }
```
andy-blum, line 25: I can't beleive this got removed from the reset upstream

m4olivei: For sure. There are quite a few `box-sizing: border-box` rules building up in the styles package, but ship has sailed to try it overall in a reset, until a major version update probably.

andy-blum (approved):

PR #12173, opened 2024-12-17
fix(content-item): fixing cta spacing in relation to the copy

bruno-amorim:
### Related Ticket(s)

Closes # [ADCMS-7259](https://jsw.ibm.com/browse/ADCMS-7259)

### Description

In the `c4d-content-item` component, the CTA should have 2rem of space between the paragraph and the CTA per [specs](https://www.figma.com/design/CVN80lAAGuolpeudRGS4BN/Content-section-v2?node-id=3-2648&t=Iw4WyJUr8nfD8sXD-1)

### Changelog

Adjusted padding for `c4d-link-with-icon`

### DIFF
--- a/packages/styles/scss/internal/content-item/_content-item.scss
+++ b/packages/styles/scss/internal/content-item/_content-item.scss
@@ -101,10 +101,6 @@
     margin-block-end: $spacing-05;
   }
 
-  ::slotted(#{$c4d-prefix}-link-with-icon) {
-    padding: $spacing-05 0 $spacing-07;
-  }
-
   :host(#{$c4d-prefix}-content-item-copy:not([has-cta])) {
     ::slotted(#{$c4d-prefix}-content-item-paragraph) {
       margin-block-end: 0;
@@ -121,5 +117,6 @@
 
   :host(#{$c4d-prefix}-content-item) ::slotted([slot='footer']) {
     margin-inline-start: 0;
+    padding-block: $spacing-05 $spacing-07;
   }
 }
--- a/packages/styles/scss/internal/content-item/_content-item.scss
+++ b/packages/styles/scss/internal/content-item/_content-item.scss
@@ -119,17 +119,7 @@
     margin-inline-start: 0;
   }
 
-  :host(#{$c4d-prefix}-content-item) ::slotted([slot='footer']:first-of-type) {
-    padding-top: $spacing-05;
-  }
-
-  :host(#{$c4d-prefix}-content-item) ::slotted([slot='footer']:last-of-type) {
-    padding-bottom: $spacing-07;
-  }
-
-  :host(#{$c4d-prefix}-content-item)
-    ::slotted([slot='footer']:not(:first-of-type):not(:last-of-type)) {
-    padding-top: 0;
-    padding-bottom: 0;
+  .#{$prefix}--content-item__cta {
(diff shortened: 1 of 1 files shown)


packages/styles/scss/internal/content-item/_content-item.scss
```diff
@@ -101,6 +101,10 @@
     margin-block-end: $spacing-05;
   }
 
+  ::slotted(#{$c4d-prefix}-link-with-icon) {
```
m4olivei, line 104: As [discussed in Slack](https://ibm-marketing-comms.slack.com/archives/C05P50B9UNT/p1734459607142819?thread_ts=1734352965.815569&cid=C05P50B9UNT), this selector is likely to miss the AEM scenario that came up, as originally reported in [ADCMS-6878](https://jsw.ibm.com/browse/ADCMS-6878), on [this test page](https://author-p131558-e1281330.adobeaemcloud.com/content/adobe-cms/us/en/carbon-v2-implementation-qa-devfolder/carbon-v2-implementation-qa-stagefolder/melike-test-stage/card-in-card--12-col.html?wcmmode=disabled).

m4olivei (changes requested):

packages/styles/scss/internal/content-item/_content-item.scss
bruno-amorim: @m4olivei I moved that rule to a selector that was already selecting the slotted footer. Can you please check now?

Thanks!

packages/styles/scss/internal/content-item/_content-item.scss
```diff
@@ -118,4 +118,18 @@
...
+
+  :host(#{$c4d-prefix}-content-item) ::slotted([slot='footer']:last-of-type) {
+    padding-bottom: $spacing-07;
+  }
+
+  :host(#{$c4d-prefix}-content-item)
+    ::slotted([slot='footer']:not(:first-of-type):not(:last-of-type)) {
+    padding-top: 0;
+    padding-bottom: 0;
+  }
```
bruno-amorim, line 134: @m4olivei 

So, it turns out when we have multiple CTAs stacked, it won't multiple <a> elements but the entire parent, in storybook it will be `c4d-link-with-`icon and in AEM it will be `div.cds--footer--cta-link`

Therefore, each CTA will have much space between them. 

In this approach I'm still selecting `[slot='footer']` but applying the margin depending on the slotted element being the first or last in his context. If we only have 1 CTA, the browser seems to be considering it as the first and the last element - applying the proper top and bottom margin. This one will reset padding of CTAs in the middle ` ::slotted([slot='footer']:not(:first-of-type):not(:last-of-type))`

I appreciate your thoughts. Thanks @m4olivei !

m4olivei: That's a clever approach! I wonder if it would be more straightforward to set the padding on the wrapper `<div>`? eg. this one: https://github.com/bruno-amorim/carbon-for-ibm-dotcom/blob/fix/content-item-cta-spacing/packages/web-components/src/components/content-item/content-item.ts#L143-L146

Have you tried that?

bruno-amorim: @m4olivei thanks for your suggestion! Can you please check it now?

PR #10779, opened 2023-08-01
feat(package): update packages to Carbon v11

kennylam:
### Related Ticket(s)

Closes #10373, #10616

### Description

This updates packages to use Carbon v11. This includes updating all token, function, mixin, and variable names to the v11 API.

### Changelog

**New**

- {{new thing}}

**Changed**

- `@carbon/ibmdotcom-styles` and `@carbon/ibmdotcom-web-components` updated to Carbon v11

**Removed**

- {{removed thing}}

PR #12175, opened 2024-12-17
fix(styles): restore use of globals/imports.scss

andy-blum:
### Description

the carbon reset has all items box-sizing set to inherit. this changes that to border box and is intended mainly to see what shifts in Percy

### DIFF
--- a/packages/styles/scss/components/audio-player/_audio-player.scss
+++ b/packages/styles/scss/components/audio-player/_audio-player.scss
@@ -9,6 +9,7 @@
 @use '@carbon/styles/scss/spacing' as *;
 @use '@carbon/styles/scss/theme' as *;
 @use '../../globals/vars' as *;
+@use '../../globals/imports' as *;
 @use '../../globals/utils/ratio-base' as *;
 @use '@carbon/styles/scss/components/slider';
 
--- a/packages/styles/scss/components/back-to-top/_back-to-top.scss
+++ b/packages/styles/scss/components/back-to-top/_back-to-top.scss
@@ -10,6 +10,7 @@
 @use '@carbon/styles/scss/spacing' as *;
 @use '@carbon/styles/scss/theme' as *;
 @use '../../globals/vars' as *;
+@use '../../globals/imports' as *;
 @use '@carbon/styles/scss/components/button' as *;
 
 @mixin back-to-top {
--- a/packages/styles/scss/components/background-media/_background-media.scss
+++ b/packages/styles/scss/components/background-media/_background-media.scss
@@ -12,6 +12,7 @@
 @use '@carbon/styles/scss/utilities' as *;
 @use '@carbon/web-components/scss/components/tooltip/tooltip' as *;
 @use '../../globals/vars' as *;
+@use '../../globals/imports' as *;
 @use '../image';
 
 @mixin background-media {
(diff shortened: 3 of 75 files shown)


andy-blum:
This did nothing.

andy-blum:
The components most impacted in the percy review:

## Locale Modal

Component fails grid alignment in both cases.

Before:
Screenshot 2024-12-18 at 8 44 30 AM

After:
Screenshot 2024-12-18 at 8 44 44 AM

## Expressive Modal

Component fails grid alignment in both cases.

Before:
Screenshot 2024-12-18 at 8 46 03 AM

After:
Screenshot 2024-12-18 at 8 46 08 AM

## Card Section Simple

Component fails grid alignment in both cases.

Before:
Screenshot 2024-12-18 at 8 54 32 AM

After:
Screenshot 2024-12-18 at 8 54 43 AM

## Quote

Component is improved by this PR

Before:
Screenshot 2024-12-18 at 8 57 06 AM

After:
Screenshot 2024-12-18 at 8 57 17 AM

## Link List Section

Component is improved by this PR

Before:
Screenshot 2024-12-18 at 8 58 46 AM

After:
Screenshot 2024-12-18 at 8 58 50 AM

areagan1030:
- **Locale modal** – it appears that the internal spacing is correct, but the outer container is shifting in how it sits on the grid on md to max (sm is as expected). I don't see v2 Figma specs, but I'm looking at v1 and the modal is not sitting on the grid in a normal way here either. So, storybook preview does actually align with the v1 Figma layout.
- **Expressive modal** – same as above. The v1 Figma is looking closer to insetting the modal 4 cols on lg to max, and md is somewhat off grid. While the storybook preview is also off grid, these are both a little wonky in different ways.
- **Card Section Simple** – the leftmost alignment of the cards is looking okay (I'm assuming borders might be set to outside, which makes this look shifted 1px left – but text alignment is correct). It appears that the card width and the spacing between cards is off. This is dependent on whether the section should be using the default or narrow grid mode. If default, card edges align on the grid left and right (4 cols) with 2rem between. If condensed, left edge hangs 1rem in gutter and right edge is on grid with 1rem between cards. I assume default may be the preferred for CIBM here, but worth noting that we're not using default grid in AEM (generally has been condensed or narrow).
- **Quote** – For md to max, the text should be aligning on the grid line with the quote hanging to the gutter at left. Design specs show quote hanging 1 rem to left. It also appears that the rule at the bottom of the component is hanging 1rem into gutter left and right. So this aligns with the quote mark at left. And the rule in storybook preview looks like it's stopping short on the right side.
- **Link list section** – This one seems to be matching the design specs. In AEM, we had been following the narrow grid standards previously with the horizontal rule, but it appears this doesn't hang in the gutter in v2 now. The storybook preview seems to have the correct grid alignment.

packages/styles/scss/components/audio-player/_audio-player.scss
```diff
@@ -9,6 +9,7 @@
 @use '@carbon/styles/scss/spacing' as *;
 @use '@carbon/styles/scss/theme' as *;
 @use '../../globals/vars' as *;
+@use '../../globals/imports' as *;
```
m4olivei, line 12: This import is definitely part of all the components in the `styles` package in v1 (with a couple exceptions). I wonder why it was removed, and/or if it was removed intentionally? Did we ask @kennylam , @jeffchew and friends?

I found the change was made here:

PR: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10779
Commit: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/10779/commits/35bd459da0693b6c7b108f68ea1aa83360bdde07#diff-f1b1620afc656e11801eeb4789dc82e1a97c3199a0626c184802d5e0a72f1a5d

No conversation about the change there. It was a massive PR, sensing it just got missed.

PR #12179, opened 2025-01-06
fix(carousel): Fix gap calculation

marcelojcs:
### Related Ticket(s)

[JIRA](https://jsw.ibm.com/browse/ADCMS-7404)
### Description

PR To address the carousel component gap distance not being calculated properly. causing the slotted components to scroll too much and appear cropped.

### Changelog

**Changed**

changed the `_updateGap` function to calculate the gap value based on the margin of the fist element. 

### Testing instructions 
Open the carousel components on storybook and see that it scrolls properly and there are no regressions.


Valentin-Sorin-Nicolae (approved):
The behavior of carousel is good now and all tests have pass. Great work mate!

PR #12180, opened 2025-01-08
fix(leadspace): wrong gradient display when in RLT mode

bruno-amorim:
### Related Ticket(s)

Closes # [ADCMS-7399](https://jsw.ibm.com/browse/ADCMS-7399)

### Description

In the Leadspace component, when in RTL mode, the text would be displayed from the right to left but the gradient in the image would still be from the left to the right, causing the text to be unreadable in images with low contrast.

I also took the opportunity to fix the leadspace with video background - when in RTL, the video was not spanning the full width of its container.


marcelojcs (approved):
The code Changes look great and is passing all tests
Great job!!

Valentin-Sorin-Nicolae:
The gradient is changing using 'rotate(90)' but the image stays the same. I think the image needs a flip also, right?

PR #12190, opened 2025-01-20
feat(card group): adding left and right gutters to the cards when in …

bruno-amorim:
…mobile

### Related Ticket(s)

Closes [#ADCMS-6599](https://jsw.ibm.com/browse/ADCMS-6599)

### Description

Adding left and right gutters to the Block Card Container component, specifically, so we make sure it's clear that cards are tappable and that they are indeed cards rather than background containers. 
All card groups that are wrapped by `c4d-card-group` will also be benefited from this update.

### Changelog

`16px` inline margin added to the c4d-card-group element when in mobile.

### DIFF
--- a/packages/styles/scss/components/card-group/_card-group.scss
+++ b/packages/styles/scss/components/card-group/_card-group.scss
@@ -42,11 +42,19 @@
         1fr
       );
     }
+
+    @include breakpoint(sm) {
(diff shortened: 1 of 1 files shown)


Valentin-Sorin-Nicolae (approved):
Looks good, Bruno. Approved!

PR #12191, opened 2025-01-22
feat(card-in-card): add gutter space on mobile

Valentin-Sorin-Nicolae:
### Related Ticket(s)

Closes # [ADCMS-6606](https://jsw.ibm.com/browse/ADCMS-6606)

### Description

It's difficult to see that cards are clickable rather than just having differently colored background when they display full-width in the TOC layout. Add proper "gutter" space on the left and right sides of cards to suggest that they are clickable cards.

### Changelog

**New**

Added left and right space using padding around the `card-in-card` component using the $spacing-05 token for mobile screen.


bruno-amorim (approved):
LGTM! Nice job mate!

PR #12214, opened 2025-02-19
fix(content-block-paragraph): fixing the logic to hide the component …

bruno-amorim:
…when no copy

### Description

In this PR https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/12198, we created a logic to hide the component if it has no copy, however, that logic that hides and displays, was only considering direct text nodes, but in AEM we have a div for the RTE and only then the text content. It was resulting in the component to hide even if there's copy 

### Changelog

Modified the logic for the method `toggleVisibility()` to also look for any text or element nodes, in case of element nodes, we'll check if they have text content.


marcelojcs (approved):
Works wonderfully!
Great job!!

PR #12218, opened 2025-02-27
feat(card): replicate v1 hover behavior

marcelojcs:
### Related Ticket(s)

[JIRA](https://jsw.ibm.com/browse/ADCMS-7539)

### Description

PR to replicate the the v1 hover interaction on cards with pictogram.

### Changelog

**Changed**

- V1 hover interaction for cards with pictogram is now preesent on viwewports larger than breakpoint md and with fine pointer

### DIFF
--- a/packages/styles/scss/components/card/_card.scss
+++ b/packages/styles/scss/components/card/_card.scss
@@ -1,5 +1,5 @@
 /**
- * Copyright IBM Corp. 2016, 2024
+ * Copyright IBM Corp. 2016, 2025
  *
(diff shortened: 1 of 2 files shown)


Valentin-Sorin-Nicolae (commented):
The hover behavior is good. What it needs to be different is the title when the pictogram is on top, the title should be on the bottom
image
Like here on the v1 
image

Valentin-Sorin-Nicolae (approved):
Now works good. Approved!

PR #12233, opened 2025-03-18
fix(table-of-contents): fix false positives on error handling

marcelojcs:
### Description

PR to fix false positives on the error handling function making the table of contents not render


bruno-amorim (approved):
LGTM - Nice job mate!

PR #12237, opened 2025-03-19
fix(masthead): globe icon logic fixed to display

bruno-amorim:
### Related Ticket(s)

None - this is a work requested for the L0 work by Nishant and Neha

### Description

Adding a condition through an attribute to render the globe icon for the lang selector. 

By adding `has-language-selector` attribute to `c4d-masthead`, it'll display the globe icon:
image


marcelojcs (approved):
LGTM

PR #12238, opened 2025-03-25
fix(masthead): enabling the icon and using the correct icon

bruno-amorim:
### Related Ticket(s)

none - this is a work for the new L0 that will live inside of AEM

### Description

Enabling the world icon and using the correct one

image


marcelojcs (approved):
LGTM!!!

PR #12249, opened 2025-04-15
fix(pricing-table): enabling overflow hints in mobile

bruno-amorim:
### Related Ticket(s)

https://jsw.ibm.com/browse/ADCMS-6628

### Description

The `Pricing Table` inherits its inner structure from the `Structured List component`. The latter has a built in logic to indicate mobile users that there's scrollable content to the right or to the left. 

This logic was not working due to some CSS constraints.

### Changelog

Added some new CSS rules to enable Overflow Hints as per Structured List component

https://github.com/user-attachments/assets/49e00d3c-0701-43d2-8d17-84adff8cd423


marcelojcs (approved):
Works well on my end!!

PR #12256, opened 2025-04-30
fix(card-group): incorrect mobile gutters for FW layout pages in AEM

bruno-amorim:
### Related Ticket(s)

https://jsw.ibm.com/browse/ADCMS-8167

### Description

[In this ticket](https://jsw.ibm.com/browse/ADCMS-6599) we added inline gutters for the card group in mobile.

But we found out that it would only work correctly in ToC pages, since FW ones already have that gutter in mobile resulting in an ever bigger gutter. 

### Changelog

I added a logic that will check if `body` has the attribute of `data-fullwidthtemplate="true"` - if we're in a FW template, the gutters will not be applied

`  :not(:host(#{$c4d-prefix}-card-group.is-full-width-template)) {`
    `@include breakpoint-down(md) {`
      `padding-inline: $spacing-05;`
   ` }`

### DIFF
--- a/packages/styles/scss/components/card-group/_card-group.scss
+++ b/packages/styles/scss/components/card-group/_card-group.scss
@@ -1,5 +1,5 @@
 /**
- * Copyright IBM Corp. 2016, 2024
+ * Copyright IBM Corp. 2016, 2025
  *
  * This source code is licensed under the Apache-2.0 license found in the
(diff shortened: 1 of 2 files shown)


Valentin-Sorin-Nicolae (approved):
Now is working great on both TOC and FW. Thanks mate!

marcelojcs (approved):

PR #12257, opened 2025-05-02
fix(masthead-l1): mobile dropdown with 0 width on iOS devices

marcelojcs:
### Description
PR to fix l1 dropdown not appearing on IOS devices 

### Changelog
The component is not appearing because it has 0 width on iOS devices.
This PR forces full width on that component.


bruno-amorim (approved):
LGTM!! Nice work mate!!

PR #12274, opened 2025-05-28
fix(background-media): gradient rtl display issue

bruno-amorim:
### Related Ticket(s)

https://jsw.ibm.com/browse/ADCMS-7759

### Description

The background media component was not flipping the gradient to be from the right to the left when in RTL mode.
I've added the RTL logic to make it work in RTL as well:

old:
image

new:
image


Valentin-Sorin-Nicolae (approved):
Working as intended. Great work, Bruno!

PR #12278, opened 2025-06-10
feat(callout-quote): new appearances enabled card and bubble

bruno-amorim:
### Related Ticket(s)

https://jsw.ibm.com/browse/ADCMS-8432

### Description

Enabling new 'appearance' props to now be used along with 'Mark Types'.

image

image

NOTE: the new appearances are mobile only. Default should display on all screens but "card" and "bubble-quote" are mobile only

### Changelog

**New**

- Callout quote component
New prop "appearance" added that enables two new styles: "card" quotes and "bubble" quotes

### DIFF
--- a/packages/styles/scss/components/quote/_quote.scss
+++ b/packages/styles/scss/components/quote/_quote.scss
@@ -191,55 +191,71 @@
     display: inline-block;
   }
 
+  //Hiding bubble-pointer as bubble-quote is mobible only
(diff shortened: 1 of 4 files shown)


Valentin-Sorin-Nicolae (approved):
It's working. I don't know if the knobs from the story can be adjusted not to have that selector on desktop view and being available only on mobile one.

PR #12291, opened 2025-07-17
fix(button-group): remove element selector restrictions

marcelojcs:
### Related Ticket(s)

[jira](https://jsw.ibm.com/browse/ADCMS-7379)

### Description

### Changelog

PR to remove the restrictions on the method that defines the child count on `button-group` 
The restrictions were filtering out valid components and causing the button group to misbehave.   

Before: 
image
(buttons on top of each other)

After:
image
(Buttons side by side)


bruno-amorim (approved):
Working fine to me too! Nice job!

PR #12382, opened 2026-01-07
fix(card-grup): card group item is clickable

marcelojcs:
### Related Ticket(s)

[Jira](https://jsw.ibm.com/browse/ADCMS-10636)

### Description

Adding a negative z-index so the after component on the grid display styles does not impact user interactions.

### Changelog

**New**

Adding a negative z-index so the after component on the grid display styles does not impact user interactions.


bruno-amorim (approved):
LGTM!!

PR #12514, opened 2026-05-14
Fix potential OS command injection in tag-release.js

GJQ-Q:
# [PR Description] Command Injection Reproduction Notes for `tag-release.js`

## Summary
Hello,
This document describes a command injection vulnerability in the following file:
`carbon-design-system/carbon-for-ibm-dotcom/tag-release.js`

The issue exists in the automated tagging and releasing utility. User-controlled input (specifically the `tagName`) is used to construct shell command strings that are executed via `child_process.execSync(...)`. Due to a lack of sanitization, a local user can inject arbitrary OS commands by using shell metacharacters during the interactive prompt.

## Root Cause

The issue is caused by direct string concatenation of user input into a shell command within the createTag function:

```js
function createTag(tagName) {
  console.log(chalk.green(`Tagging the repo with ${tagName}...`));
  child.execSync(`git tag -a ${tagName} -m "Release ${tagName}"`);
  child.execSync('git push --tags');
}
```
Since `execSync(string)` invokes a system shell, characters like `&`, `;`, or `|` allow an attacker to "break out" of the intended `git tag` command.

## Reproduction Material

A professional Proof of Concept script is provided: `poc_tag-release.js`.
[poc_tag-release.js](https://github.com/user-attachments/files/27757404/poc_tag-release.js)

### What the PoC Does

1. Environment Hijacking: It hooks `child_process.execSync` to monitor for the injection signature (e.g., `calc.exe`).
2. Environment Mocking: It intercepts Git commands (like `rev-parse` and `config`) and provides fake return values to ensure the script doesn't crash in a non-git environment.
3. Automated Injection: It monitors the target script's `stdout`. When the "What is the release tag name" prompt appears, it automatically writes the payload: `v1.0.0 & calc.exe &`.
4. Automated Confirmation: It automatically responds with `y\r\n` to all subsequent confirmation prompts to ensure the execution flow reaches the vulnerable "Sink".

## How to Run

**Ensure the target script (tag-release.js) and PoC are in the same directory, then run:**

```bash
node poc_tag-release.js

```

## Expected Output

When the PoC runs, you should see output similar to:

```text
? What is the release tag name to create? (e.g. v0.0.0-rc.0)
[POC]  Payload...
? What is the release tag name to create? (e.g. v0.0.0-rc.0) v1.0.0 & calc.exe &
? Tag and push to Github? Yes
? Create the release from changelogs? (For release candidates only. If this is a
 full release, create the release from the pinned issue from this sprint
instead) Yes
Tagging the repo with v1.0.0 & calc.exe &...
```
In the vulnerable version, after the interactive prompt is processed, the local machine will launch the Calculator application as a benign demonstration effect.

This shows that the attacker-controlled `tagName` value can influence command execution behavior through the vulnerable `createTag()` path.

## Patch Explanation

This branch includes a patched version of `tag-release.js` intended to mitigate the command injection risk described above.

### What the patch changes

The patch introduces strict validation and safer execution methods before user-controllable values are used by the following functions:

- `createTag(tagName)`
- `createRelease(tagName)`

In the vulnerable version, the `tagName` provided via the interactive prompt could be concatenated into shell commands and executed via `child_process.execSync(...)` without any sanitization.

The patched version introduces two layers of protection to ensure that only legitimate tag names can be processed.

### Validation introduced by the patch
The patch implements multiple layers of protection to ensure that only legitimate tag names can be processed:

Native Git Refname Validation: Instead of relying on a custom regular expression, the patch uses `child.execFileSync('git', ['check-ref-format', '--allow-onelevel', input])`. This leverages Git's internal logic to ensure the tagName conforms to official Git reference standards.

Argument Injection Prevention: The patch explicitly checks if the input starts with a dash (`-`). If a user attempts to enter a tag name like `--help` or `-f`, the script rejects it immediately to prevent the input from being misinterpreted as a command-line flag.

Safe Command Execution: By switching to `child_process.execFileSync(file, [args])`, the command and its arguments are passed as a distinct array. This prevents shell interpretation of shell metacharacters (like `&` or `|`) even if they were to bypass the initial validation.


GJQ-Q:

GJQ-Q:

GJQ-Q:

GJQ-Q:
recheck

PR #12531, opened 2026-05-31
fix(Videoplayer): Fix container div to fit its parent height

kohm:
## Problem

When using `c4d-background-media` with a `c4d-video-player-container-v7` in `background-mode`, or placing two or more video player containers on the same page, only the first video would load. The remaining containers appeared but were never populated by the Kaltura player.

### 1. Concurrent `embed()` calls (`KalturaPlayer.js`)
Both containers called `IBM.Mediacenter.player.embed()` simultaneously with the same `entryId`. The IBM Mediacenter library does not reliably handle concurrent embed requests for the same video entry — the second call returns `null`, leaving its container with no player instance.

**Fix:** introduced a module-level `_embedQueue` that chains each `embed()` call onto the previous one, ensuring they run sequentially.

### 2. Shared retry counter (`KalturaPlayer.js`)
The `_attempt` counter used by `_scriptReady()` to wait for the IBM script to load was a shared module-level variable. Two concurrent callers each incremented it on every tick, effectively halving the 5s timeout to ~2.5s and causing one container to give up prematurely on a cold first load.

**Fix:** made `attempt` a per-call parameter passed through recursion instead of a shared module-level variable.

### 3. Wrong `querySelector` scope (`video-player-composite.ts`)
`renderLightDOM()` used `document.querySelector('.c4d--video-player__video')` to set the RTL `dir-mode` attribute, which always targeted the **first** video div in the entire document regardless of which container instance triggered the render.

**Fix:** scoped to `this.querySelector(...)` so each instance only targets its own video div.

### 4. Container div height (`_video-player.scss`)
The inner video container div was not inheriting its parent's height in `background-mode`, causing the video to be invisible even when successfully embedded.

**Fix:** added `block-size: 100%` to the container div.

### DIFF
--- a/packages/services/src/services/KalturaPlayerV7/KalturaPlayer.js
+++ b/packages/services/src/services/KalturaPlayerV7/KalturaPlayer.js
@@ -126,7 +126,12 @@ let _embedQueue = Promise.resolve();
  * @param {number} attempt Per-call retry count
  * @private
  */
-function _scriptReady(resolve, reject, environment = _ibmEnvironment, attempt = 0) {
+function _scriptReady(
+  resolve,
+  reject,
+  environment = _ibmEnvironment,
+  attempt = 0
+) {
   /**
    * @param {object} root?.IBM.Mediacenter.player if exists then resolve
    */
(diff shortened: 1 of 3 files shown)


kohm:

marcelojcs (approved):

PR #12589, opened 2026-07-31
feat(cta): add suppor to chang theme prop

marcelojcs:
### Description

Adding theme as new attribute so we can ensure we style this background correctly from AEM side.


bruno-amorim (approved):
LGTM! Thanks!

PR #12660, opened 2026-09-14
fix(card-group): change logo handling to avoid CLS

marcelojcs:
### Description
Addresses a layout shift issue caused by the logo attribute being injected after page load via JavaScript. The fix ensures the card-group component initializes properly without triggering layout instability.


bruno-amorim (approved):
LGTM!

